PR #9, opened 2018-10-08
execute command on the cnd container

rberrelleza:


cmd/exec.go
```diff
@@ -0,0 +1,66 @@
...
+		return err
+	}
+
+	dev, err := model.ReadDev(devPath)
+	if err != nil {
+		return err
+	}
+
+	pods, err := client.CoreV1().Pods(namespace).List(v1.ListOptions{
+		LabelSelector: fmt.Sprintf("cnd=%s", dev.Name),
```
rberrelleza: @pchico83 can you think of a better selector for this? This won't be unique once we have more than one environment on the same cluster. 

Maybe we need to update deploy to add the unique pod name to the yaml file?

PR #45, opened 2018-10-16
Generate config and launch individual syncthing instance with hardcoded deviceIDs

### LINKED ISSUE #46: Can't run more than one CND environment at a time
When adding a second CND environment, there is a race condition, and only one of the environments is synchronized. 

It seems like there's a race condition on the ksync api, we are into how to mitigate this. For the the time being, please only run one environment at a time.

rberrelleza:
Fixes #46 and #47


cmd/up.go
```diff
@@ -6,26 +6,31 @@ import (
...
 	cmd := &cobra.Command{
 		Use:   "up",
 		Short: "Starts or upgrades a cloud native environment",
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return executeUp(devPath)
+			return executeUp(devPath, remoteAddressStr)
 		},
 	}
 	cmd.Flags().StringVarP(&devPath, "file", "f", "cnd.yml", "manifest file")
+	cmd.Flags().StringVarP(&remoteAddressStr, "address", "a", "localhost:22001", "remote address")
```
pchico83: This PR looks super nice! Just remember to update the README.mc

pchico83:
We can remove `ksync` as a library and the go code to call `ksync create`.

PR #56, opened 2018-10-18
restore the environment before exiting cnd up

rberrelleza:


cmd/exec.go
```diff
@@ -101,3 +105,20 @@ func getCNDPod(c *kubernetes.Clientset, namespace string, dev *model.Dev) (*apiv
 
 	return &pod, nil
 }
+
+func checkForGracefulExit(err error) error {
```
rberrelleza: this is so that a ctrl+c on exec doesn't show as an error

pchico83:
Outdated PR, closing for now...

PR #61, opened 2018-12-04
Replace deployments

rberrelleza:
Instead of launching a new deployment and appending it to the service, we create a new dev deployment based out deployment provided by `dev.Swap.Deployment.File` and launch it to replace the existing one.

When a user runs `cnd down` we do the opposite and relaunch a deployment with the original configuration to replace the dev one.

I propose we remove `rm` in favor of just having `up`(to enable dev mode) and `down` (to restore the original configuration)


model/deployment.go
```diff
@@ -17,39 +17,33 @@ type deployment struct {
...
+	d, err := dev.loadDeployment()
 	if err != nil {
 		return nil, err
 	}
 
-	dec := k8Yaml.NewYAMLOrJSONDecoder(file, 1000)
-	var d appsv1.Deployment
-	dec.Decode(&d)
-
-	d.GetObjectMeta().SetName(dev.Name)
```
rberrelleza: @pchico83 do you think we should remove the name altogether, and just use the name of the deployment instead?

pchico83: Yeap!

model/deployment.go
```diff
@@ -17,39 +17,33 @@ type deployment struct {
...
 	d.GetObjectMeta().SetLabels(labels)
-	if d.Spec.Selector == nil {
-		d.Spec.Selector = &metav1.LabelSelector{
-			MatchLabels: map[string]string{"cnd": dev.Name},
-		}
-	} else {
-		d.Spec.Selector.MatchLabels["cnd"] = dev.Name
-	}
-	d.Spec.Template.GetObjectMeta().SetName(dev.Name)
+
```
pchico83: maybe in another PR, but could we define a flag to deactivate (or not) every kind on healthcheck?

PR #73, opened 2018-12-06
Keep default image command/args if none is specified

pchico83:


pkg/model/dev.go
```diff
@@ -72,12 +72,6 @@ func loadDev(b []byte) (*Dev, error) {
 			Source: ".",
 			Target: "/src",
 		},
-		Swap: swap{
```
rberrelleza: This would make a lot of images instantly exit on development mode (e.g. if they depend on a compiled artifact). I think we need to document that, as it's a pretty common scenario (.e.g you might want to control when to run your webserver, as it might exit when you have a bug, killing the container)

PR #78, opened 2018-12-08
Update the quickstart guide

rberrelleza:


README.md
```diff
@@ -8,12 +8,44 @@ CND helps you achieve this with a mix of kubernetes automation, file synching be
...
 ## Quickstart
 
-- Install `cnd` locally 
-- Deploy a kubernetes service
-- Run `cnd up`
-- 
+Let's start with something simple to show you the power of cloud native development. First, install [cnd locally](#installation).  
 
-For a more advanced scenario [check out our voting app cnd demo](https://github.com/okteto/cnd-voting-demo).
+Clone the `welcome` service.
```
rberrelleza: Should we rename the service to something like `voting`? It's no longer a `welcome web page`

rlamana: Agree, maybe we can call it `voting-app`.
```suggestion
Clone the `voting-app` service.
```

README.md
```diff
@@ -2,18 +2,50 @@
...
```
pchico83: I would move the example here. It is a common practice in other dev tools

rberrelleza: My thought here is that it's kind of odd to clone the repo of a tool just to get a sample of how to use it, so instead we can have a samples repo (maybe put vote the getting started and the more advance one there?) (telepresence and duffle do this)

rberrelleza:
@rlamana @pchico83 are you ok if we do the voting app rename (and if to move it to here or it's own repo) on a different PR?

PR #80, opened 2018-12-08
Run command

rberrelleza:


docs/cnd-file.md
```diff
@@ -40,3 +42,14 @@ The local folder synched to the remote container. (default: the current folder)
 ## mount.target (required)
 
 The remote folder path synched with the local file system.
+
+## swap.scripts (optional)
```
pchico83: is it inside `swap` or not? then, the example above is wrong.

pkg/model/dev_test.go
```diff
@@ -92,7 +93,9 @@ swap:
     container: core
 mount:
   source: /Users/fernandomayofernandez/PycharmProjects/codescope-core
-  target: /app`)
+  target: /app
+scripts:
+  run: "uwsgi --gevent 100 --http-socket 0.0.0.0:8000 --mount /=codescope:app --python-autoreload 1"`)
```
pchico83: can we have another example?

cmd/list.go
```diff
@@ -26,7 +26,7 @@ type Data struct {
 func List() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "list",
-		Short: "lists your active cloud native development environments",
+		Short: "Lists your active cloud native development environments",
```
rlamana: nit: the language we use to describe our commands is not consistent. Most CLIs use imperative mood rather than 3rd person. We should also change this for `down` and `up`.

```suggestion
		Short: "List your active cloud native development environments",
```

pkg/model/dev_test.go
rberrelleza: This is for a test, but what other example did you have in mind?

docs/cnd-file.md
rberrelleza: the example is wrong, I'll fix it. `scripts` should be a first level element

pkg/model/dev_test.go
```diff
@@ -85,24 +86,60 @@ mount:
...
+	}{
+		{
+			"long script",
+			[]byte(`
+            swap:
+              deployment:
+                name: service
+                container: core
+            mount:
+              source: /Users/fernandomayofernandez/PycharmProjects/codescope-core
```
pchico83: better to use a different example here

pkg/model/dev_test.go
```diff
@@ -85,24 +86,60 @@ mount:
...
+              target: /app
+            scripts:
+              run: "uwsgi --gevent 100 --http-socket 0.0.0.0:8000 --mount /=codescope:app --python-autoreload 1"`),
```
pchico83: and here

PR #92, opened 2018-12-20
Reorganize readme in different sections

pchico83:


examples/k8-manifests/README.md
```diff
@@ -0,0 +1,59 @@
+# Voting App deployed with raw kubernetes manifests
```
pchico83: @rberrelleza we could have two more folders here, one for helm and another one for cnab. this way, we provide examples about how to do CND with different deployment technologies.

PR #93, opened 2019-01-04
Update issue templates

rberrelleza:
Use the standard issue templates


rberrelleza:
This is so that we have all the requirements from https://opensource.guide/

PR #99, opened 2019-01-05
Support for resource override

### LINKED ISSUE #97: Support to override resource's value
On production, a kubernetes deployment might have limited memory or CPU resources. On dev mode, I'd like to override these, as I might need to run resource-intensive tasks like compile, run tests, etc...

Add a resources values to `cnd.yml` to allow overrides

rberrelleza:
Fixes #97 

## Proposed changes
- Remove the `resources` element on the container so it doesn't constraint during development.

### DIFF
--- a/pkg/model/deployment.go
+++ b/pkg/model/deployment.go
@@ -4,27 +4,15 @@ import (
 	log "github.com/sirupsen/logrus"
 	appsv1 "k8s.io/api/apps/v1"
 	apiv1 "k8s.io/api/core/v1"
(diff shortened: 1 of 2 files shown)


rberrelleza:
This is an exploratory PR. If approved, I'll add tests and docs.

pchico83:
Should we have a way to unset resource limits/requests?
Or maybe unset by default and set resources if they are specified in `cnd.yml`?

rberrelleza:
I think we should just unset them for now. I think it makes more sense from a dev experience perspective.

PR #101, opened 2019-01-05
Add create command to generate default cnd.yml

rberrelleza:
## Proposed changes
- Add an `cnd create` command that automatically generates  a cnd.yml based on the contents of the folder where the command is being run.  

e.g., if run on example/vote/kubectl/vote, it results in:
```
swap:
  deployment:
    name: vote
    image: python
    command:
    - python
    - app.py
mount:
  source: .
  target: /usr/src/app
scripts:
  hello: echo Your cluster ♥ you
```

### DIFF
--- a/cmd/create.go
+++ b/cmd/create.go
@@ -17,7 +17,7 @@ import (
 func Create() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "create",
-		Short: "Automatically create the cnd manifest",
+		Short: "Automatically create the cnd manifest file",
 		RunE: func(cmd *cobra.Command, args []string) error {
 			return executeCreate()
 		},


pkg/linguist/dev.go
```diff
@@ -0,0 +1,84 @@
...
+	helloCommandName = "hello"
+)
+
+var (
+	lenguageDefaults map[string]lenguageDefault
+	tailCommand      = []string{"tail", "-f", "/dev/null"}
+)
+
+func init() {
+	lenguageDefaults = make(map[string]lenguageDefault)
```
rlamana: typo? "lenguage" -> "language"

pkg/linguist/dev.go
```diff
@@ -0,0 +1,84 @@
...
+var (
+	languageDefaults map[string]languageDefault
+	tailCommand      = []string{"tail", "-f", "/dev/null"}
+)
+
+func init() {
+	languageDefaults = make(map[string]languageDefault)
+	languageDefaults[javascript] = languageDefault{
+		image:   "node",
+		command: tailCommand,
```
rlamana: 100% of node projects will have to do `npm install` first and a very high percentage of them will then do `npm start` which is the default script to execute. I would use `npm install & npm start` as default command here.

`npm` is the official and more established package manager, but `yarn` is the new trend and it is faster. Still not sure which one we should use though.

PR #108, opened 2019-01-06
Adds JSX as recognized language and platform

rlamana:


pkg/linguist/dev.go
```diff
@@ -14,39 +14,39 @@ type languageDefault struct {
...
 var (
-	languageDefaults map[string]languageDefault
+	platformDefaults map[string]languageDefault
 	tailCommand      = []string{"tail", "-f", "/dev/null"}
 )
 
 func init() {
-	languageDefaults = make(map[string]languageDefault)
-	languageDefaults[javascript] = languageDefault{
+	platformDefaults = make(map[string]languageDefault)
```
rberrelleza: `platformDefaults` and  `languageDefault ` are now inconsistent no?

pkg/linguist/dev.go
```diff
@@ -66,13 +66,15 @@ func GetDevConfig(language string) *model.Dev {
 	return dev
 }
 
-func normalizeLanguage(language string) string {
+func getPlatform(language string) string {
```
rberrelleza: None of these but node is a `platform`. I think this is more confusing that the older naming scheme.

PR #120, opened 2019-01-07
Retry log streaming connection if it breaks

pchico83:
Fixes: stream logs stop working if pod is restarted

## Proposed changes
- Retry streaming logs connection
- Use waiting groups to handle the end of the program.

### DIFF
--- a/pkg/k8/logs/logs.go
+++ b/pkg/k8/logs/logs.go
@@ -3,30 +3,68 @@ package logs
 import (
 	"io"
 	"os"
+	"os/signal"
+	"sync"
+	"time"
 
+	log "github.com/sirupsen/logrus"
+
+	"github.com/okteto/cnd/pkg/k8/deployments"
+	appsv1 "k8s.io/api/apps/v1"
 	apiv1 "k8s.io/api/core/v1"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 )
 
-//Logs stremas logs from a container
-func Logs(c *kubernetes.Clientset, config *rest.Config, pod *apiv1.Pod, container string) error {
-	var tailLines int64
-	tailLines = 100
-	req := c.CoreV1().Pods(pod.Namespace).GetLogs(
-		pod.Name,
-		&apiv1.PodLogOptions{
-			Container:  container,
-			Timestamps: true,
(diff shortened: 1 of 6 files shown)


pkg/k8/logs/logs.go
```diff
@@ -3,30 +3,61 @@ package logs
...
-	readCloser, err := req.Stream()
-	if err != nil {
-		return err
+//StreamLogs stremas logs from a container
+func StreamLogs(d *appsv1.Deployment, container string, c *kubernetes.Clientset, config *rest.Config, wg *sync.WaitGroup) {
+	defer wg.Done()
+	var readCloser io.ReadCloser
+	end := false
+
+	channel := make(chan os.Signal, 1)
```
rberrelleza: I think it's a bit confusing to handle an os.Signal here, since it's coming from the user. I think we should have this on the `cmd` package instead, and use another channel or function call to propagate the event "down".

pkg/k8/logs/logs.go
```diff
@@ -3,30 +3,61 @@ package logs
...
+				wait++
+			}
+			continue
+		}
+		wait = 0
+		io.Copy(os.Stdout, readCloser)
 	}
-	defer readCloser.Close()
-	_, err = io.Copy(os.Stdout, readCloser)
-	return err
```
rberrelleza: who's going to handle this error? Can we log it or something?

pkg/k8/logs/logs.go
```diff
@@ -3,30 +3,61 @@ package logs
...
+	go func() {
+		<-channel
+		readCloser.Close()
+		end = true
+	}()
+
+	var wait time.Duration
+	for !end {
+		time.Sleep(wait * time.Second)
+		pod, err := deployments.GetCNDPod(d, c)
```
rberrelleza: Can we only call this once and save the values, since namespace and name are pretty set? Or is that a way to know if the pod is gone?

pkg/k8/logs/logs.go
pchico83: Right, I will do it in a follow-up PR

pkg/k8/logs/logs.go
pchico83: No, the pod name is dynamic.

PR #135, opened 2019-01-10
Show kubernetes errors when creating the cnd container

pchico83:


pkg/k8/deployments/deployments.go
```diff
@@ -109,6 +113,33 @@ func deploy(d *appsv1.Deployment, c *kubernetes.Clientset) error {
 	return nil
 }
 
+//GetPodEvents shows the events of a given pod
+func GetPodEvents(ctx context.Context, wg *sync.WaitGroup, pod *apiv1.Pod, c *kubernetes.Clientset) {
+	defer wg.Done()
```
rberrelleza: I don't think this function should prevent `cnd up` from exiting.

pkg/k8/deployments/deployments.go
```diff
@@ -109,6 +113,33 @@ func deploy(d *appsv1.Deployment, c *kubernetes.Clientset) error {
...
+	field := fields.Set{}
+	field["involvedObject.uid"] = string(pod.GetUID())
+	watch, _ := c.Core().Events(pod.Namespace).Watch(
```
rberrelleza: can we manage the error?

pkg/k8/deployments/deployments.go
```diff
@@ -109,6 +113,33 @@ func deploy(d *appsv1.Deployment, c *kubernetes.Clientset) error {
...
+			FieldSelector: field.AsSelector().String(),
+		},
+	)
+	ch := watch.ResultChan()
+	for {
+		select {
+		case e := <-ch:
+			event := e.Object.(*apiv1.Event)
+			if event.Type == "Normal" {
+				log.Info(event.Message)
```
rberrelleza: can we make this debug?

PR #143, opened 2019-01-17
Don't append volume if one with the same name is already present

pchico83:
This enables two scenarios:
- define the volume type/size/driver without us having to implement nothing in cnd.
- share the volume with other containers in the same pod (useful for running docker compose commands)

### DIFF
--- a/pkg/k8/deployments/translate.go
+++ b/pkg/k8/deployments/translate.go
@@ -165,8 +165,13 @@ func createSyncthingVolume(d *appsv1.Deployment, dev *model.Dev) {
 		d.Spec.Template.Spec.Volumes = []apiv1.Volume{}
 	}
 
-	syncVolume := apiv1.Volume{Name: dev.GetCNDSyncVolume()}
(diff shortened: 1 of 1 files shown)


pkg/k8/deployments/translate.go
```diff
@@ -165,8 +165,13 @@ func createSyncthingVolume(d *appsv1.Deployment, dev *model.Dev) {
...
 	}
 
-	syncVolume := apiv1.Volume{Name: dev.GetCNDSyncVolume()}
+	for _, v := range d.Spec.Template.Spec.Volumes {
+		if v.Name == dev.GetCNDSyncVolume() {
+			return
+		}
+	}
 
+	syncVolume := apiv1.Volume{Name: dev.GetCNDSyncVolume()}
```
rberrelleza: could this affect the scenario where we have two containers in the same deployment, and we turn both on?

pchico83: no, they use different volume names

PR #146, opened 2019-01-20
Allow for custom kubectl instead of the default

rberrelleza:
## Proposed changes
- This is to allows the user to create a `~/.cnd/kubeconfig` file to try cnd without having to mess up the existing configuration

What do you think of this approach? Another option would be to follow `kubectl` example and have a `--kubeconfig` global parameter to allow specifying the kubectl to use. I prefer the one in the PR just because  once set, then the user doesn't have to remember again. 

We could even do both (`--kubeconfig`, if empty do`~/.cnd/kubeconfig` and if empty do `.kube/config`) but it might be overkill.

### DIFF
--- a/pkg/k8/client/client.go
+++ b/pkg/k8/client/client.go
@@ -1,9 +1,6 @@
 package client
 
 import (
-	"os"
-	"path"
-
 	log "github.com/sirupsen/logrus"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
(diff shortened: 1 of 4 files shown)


pkg/k8/client/client.go
```diff
@@ -4,17 +4,25 @@ import (
...
 
-//Get returns a kubernetes client. If namespace is empty, it will use the default namespace configured.
-func Get(namespace string) (string, *kubernetes.Clientset, *rest.Config, error) {
-	home := os.Getenv("HOME")
+//Get returns a kubernetes client.
+// If namespace is empty, it will use the default namespace configured.
+// If path is empty, it will use the default path configuration
+func Get(namespace, configPath string) (string, *kubernetes.Clientset, *rest.Config, error) {
+	if configPath == "" {
+		configPath = path.Join(os.Getenv("HOME"), ".kube/config")
```
pchico83: I would set this value from the caller:
https://github.com/okteto/cnd/pull/146/files#diff-ff7686b39bf90dc2520886fb874371a4R90

PR #151, opened 2019-01-27
fail early if syncthing is not in the PATH

rberrelleza:
## Proposed changes
- fail early if syncthing is not available in the PATH


cmd/up.go
```diff
@@ -93,6 +93,11 @@ func ExecuteUp(ctx context.Context, wg *sync.WaitGroup, dev *model.Dev, namespac
 		return nil, err
 	}
 
+	sy, err := syncthing.NewSyncthing(namespace, d.Name, devList)
```
rberrelleza: I'm moving this up so that we fail if syncthing is missing before modifying the deployment

PR #153, opened 2019-01-27
Reconnect portforward

### LINKED ISSUE #50: Reconnect port forward if connection breaks

rberrelleza:
Fixes #50  

## Proposed changes
- Catch errors in the portforward, and restart the connectivity
-  Monitor the state of the connection with syncthing, and restart when it's marked as disconnected

### DIFF
--- a/cmd/up.go
+++ b/cmd/up.go
@@ -44,11 +44,13 @@ func Up() *cobra.Command {
 			var wg sync.WaitGroup
 			defer shutdown(cancel, &wg)
 
-			reconnectChannel := make(chan struct{})
+			reconnectChannel := make(chan struct{}, 1)
 			runtime.ErrorHandlers = []func(error){
 				func(e error) {
 					if strings.Contains(e.Error(), "an error occurred forwarding") ||
-						strings.Contains(e.Error(), "error creating forwarding stream") {
+						strings.Contains(e.Error(), "error creating forwarding stream") ||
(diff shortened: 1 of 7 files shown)


cmd/up.go
```diff
@@ -38,26 +41,57 @@ func Up() *cobra.Command {
...
+
+			for {
+				select {
+				case <-stopChannel:
+					fmt.Println()
+					return nil
+				case <-reconnectChannel:
+					log.Infof("reconnecting")
+					shutdown(cancel, &wg)
+					storage.Stop(namespace, dev)
```
rberrelleza: I'm calling this because I've seen several times where shutdown will finish before calling storage.Stop, not sure if this is by design

cmd/up.go
```diff
@@ -38,26 +41,57 @@ func Up() *cobra.Command {
...
+
+					ctx, cancel := context.WithCancel(context.Background())
```
rberrelleza: instead of this we could split the up sequence to a separate function, and call it here again. thoughts?

cmd/up.go
```diff
@@ -38,26 +41,57 @@ func Up() *cobra.Command {
...
+					defer shutdown(cancel, &wg)
```
rberrelleza: this might not be needed, need to do more testing

cmd/up.go
rberrelleza: Fixed the root issue on a separate PR

PR #172, opened 2019-01-31
Reconnect port forward when we detect a disconnection

### LINKED ISSUE #50: Reconnect port forward if connection breaks

rberrelleza:
Fixes #50

## Proposed changes
- Stop and start the portforward between the local machine and the remote sync process when the syncthing API detects the remote as 'disconnected'
- Add a monitoring loop to see the state of syncthing, and log when there's a connectivity issue

### DIFF
--- a/cmd/up.go
+++ b/cmd/up.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"os/signal"
 	"sync"
+	"syscall"
 
 	"github.com/cloudnativedevelopment/cnd/pkg/analytics"
 	"github.com/cloudnativedevelopment/cnd/pkg/config"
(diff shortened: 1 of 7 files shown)


cmd/up.go
```diff
@@ -77,84 +80,114 @@ func Up() *cobra.Command {
...
+		return err
+	}
+
+	pod, err := deployments.GetCNDPod(ctx, d, client)
+	if err != nil {
+		return err
+	}
+
+	if err := pf.Start(ctx, wg, client, restConfig, pod); err != nil {
+		return fmt.Errorf("couldn't start the connection to your cluster: %s", err)
```
rlamana: Since this is a reconnection `start` might be misleading. `Establish`? Is the user aware that it was trying to reconnect? What happens after this? Maybe the message should inform about the final result/impact rather of what happens to this particular piece.

rberrelleza: Yes, that's a good point. Let's do a separate PR review of those.

PR #177, opened 2019-02-02
Include the context name in the error when we can't find a deployment

### LINKED ISSUE #155: Mention cluster and namespace on errors
When okteto can't find a deployment, it's not very clear which context it was used:
```console
➜  okteto up                        
Activating your cloud native development environment...
Error: deployments.apps "website" not found
```

It would be useful if the error message included the namespace and cluster / context name it used

This is related to #149

rberrelleza:
Fixes #155

## Proposed changes
-  Include the context name in the error when we can't find a deployment
- Add extra logs for troubleshooting connectivity issues
- Downgrade log.error since we are also sending the error to the caller


cmd/up.go
```diff
@@ -174,20 +220,32 @@ func reconnectPortForward(ctx context.Context, wg *sync.WaitGroup, d *appsv1.Dep
...
-	wg.Wait()
-
-	log.Debugf("completed shutdown sequence")
+	done := make(chan struct{})
+	go func() {
+		wg.Wait()
+		close(done)
+	}()
+
+	select {
```
rberrelleza: there's a race condition in the port forwarding code (I think) that's causing this to block sometimes). I'm going to fix that on a separate PR. Either way, I think we should time out this, there's no reason to block the exit for more than a second or so.

cmd/down.go
```diff
@@ -38,17 +39,22 @@ func executeDown() error {
...
+	_, client, _, k8sContext, err := GetKubernetesClient(namespace)
 	if err != nil {
 		return err
 	}
 
 	d, err := deployments.Get(namespace, deployment, client)
 	if err != nil {
+		if strings.Contains(err.Error(), "not found") {
+			fullname := deployments.GetFullName(namespace, deployment)
+			return fmt.Errorf("deployment %s not found [current context: %s]", fullname, k8sContext)
```
rberrelleza: @rlamana this is the error wheere we include the context. It's the only one that we do it for now.

PR #182, opened 2019-02-04
Init volume from docker image instead of using a tarball

pchico83:


pkg/k8/deployments/translate.go
```diff
@@ -134,15 +136,30 @@ func mergeEnvironmentVariables(current []v1.EnvVar, dev []model.EnvVar, namespac
...
 			apiv1.VolumeMount{
 				Name:      dev.GetCNDSyncVolume(),
-				MountPath: "/src",
+				MountPath: "/cnd/init",
 			},
 		},
+		Command: []string{
+			"sh",
+			"-c",
+			fmt.Sprintf(`[ "$(ls -A /cnd/init)" ] || mv %s /cnd/init || true`, source),
```
rberrelleza: do you think is possible that `ls` or `mv` don't work in a container?

PR #186, opened 2019-02-05
pick up value from KUBECONFIG if available

### LINKED ISSUE #149: CND does not use KUBECONFIG variable
**Describe the bug**
CND does not use the KUBECONFIG environemnt variable and always looks in the user's home folder for the configuration

**To Reproduce**
```
export KUBECONFIG=/home/pau/Desktop/cndtest/test.config
cnd up
    Activating your cloud native development environment...
    Error: stat /home/pau/.kube/config: no such file or directory
```

**Expected behavior**
CDN should read from the environment variable first

 - OS: Ubuntu, Mac OS
 - All versions?

rberrelleza:
Fixes #149 

## Proposed changes
-  pick up value from KUBECONFIG if available, or fallback to $HOME/.kube/config

### DIFF
--- a/pkg/k8/client/client.go
+++ b/pkg/k8/client/client.go
@@ -1,7 +1,11 @@
 package client
 
 import (
-	"github.com/cloudnativedevelopment/cnd/pkg/log"
+	"fmt"
+	"os"
+	"path"
+	"runtime"
+
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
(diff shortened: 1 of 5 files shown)


pkg/k8/client/client.go
```diff
@@ -11,11 +15,27 @@ import (
...
+		}
+	}
+
+	kubeconfig := path.Join(home, ".kube", "config")
+	kubeconfigEnv := os.Getenv("KUBECONFIG")
+	if len(kubeconfigEnv) > 0 {
+		kubeconfig = kubeconfigEnv
+	}
+
+	if len(kubeconfig) == 0 {
```
pchico83: this case is no possible, right?

rberrelleza: yep, changing the check to see if the file exists

PR #199, opened 2019-02-11
Support for "cnd up" to two different containers in the same pod

pchico83:


cmd/up.go
```diff
@@ -249,9 +249,12 @@ func (up *UpContext) Execute(isRetry bool) error {
 		return err
 	}
 
-	up.Sy, err = syncthing.NewSyncthing(up.Namespace, up.Deployment.Name, devList)
-	if err != nil {
-		return err
+	runSync := up.Dev.Swap.Deployment.Container == devList[0].Swap.Deployment.Container
```
rberrelleza: instead of runSync I'd call this: startSyncthing or something, so it's clearer what the outcome will be

cmd/up.go
```diff
@@ -249,9 +249,12 @@ func (up *UpContext) Execute(isRetry bool) error {
...
+	if runSync {
+		up.Sy, err = syncthing.NewSyncthing(up.Namespace, up.Deployment.Name, devList)
```
rberrelleza: this might cause issues below, make sure to always initialize up.Sy

cmd/up.go
```diff
@@ -249,9 +249,12 @@ func (up *UpContext) Execute(isRetry bool) error {
...
+		if err != nil {
+			return err
```
rberrelleza: At this level, I wouldn't return the err directly, since this will be displayed to the user. Maybe return a more generic error, like fmt.Errorf("couldn't start file synchronization, please try again), and log.Info the real error

I'd apply this pattern to any error at this level, since it's the one before the user (in the inner layers it makes sense to just bubble it up).

cmd/up.go
```diff
@@ -266,11 +269,16 @@ func (up *UpContext) Execute(isRetry bool) error {
...
+	if runSync {
+		if err := up.Sy.Run(up.Context, up.WG); err != nil {
+			return err
+		}
+		syncGUIAddress = up.Sy.GUIAddress
+
 	}
 
-	err = storage.Insert(up.Context, up.WG, up.Namespace, up.Dev, up.Sy.GUIAddress, up.Pod.Name)
+	err = storage.Insert(up.Context, up.WG, up.Namespace, up.Dev, syncGUIAddress, up.Pod.Name)
```
rberrelleza: could you just pass up.sy.GUIAddress here?

rberrelleza: If the UI address is empty, this could break how `cnd list` detects stale services. Can you check that scenario?

pkg/syncthing/syncthing.go
```diff
@@ -64,29 +64,40 @@ type Syncthing struct {
 	GUIAddress       string
 	ListenAddress    string
 	Client           *http.Client
+	Master           bool
 }
 
 // NewSyncthing constructs a new Syncthing.
-func NewSyncthing(namespace, deployment string, devList []*model.Dev) (*Syncthing, error) {
+func NewSyncthing(namespace, deployment string, devList []*model.Dev, runSync bool) (*Syncthing, error) {
```
rberrelleza: can you call the paramter `master` instead of `runSync`?

pkg/syncthing/syncthing.go
```diff
@@ -64,29 +64,40 @@ type Syncthing struct {
 	GUIAddress       string
 	ListenAddress    string
 	Client           *http.Client
+	Master           bool
```
rberrelleza: don't use the term `Master`, use `primary` and `secondary` instead :)

PR #208, opened 2019-02-19
Up to diff deployments

rberrelleza:
## Proposed changes
- Allow `cnd up` to target more than one deployment with the same folder at the same time 
-  For the other operations, when there's more than one env for that folder, we'll use the path of the manifest to tell them apart

@rlamana @pchico83 how do you feel about hiding the `file` flag on `down` and `exec` since it's not the main use case (at least for now)?

Another option would be to not ask for the --file, but instead ask the user to pick among the available ones, what do you think? I prefer the `--file` because that way I know my command will just work, instead of requiring an extra prompt.


rberrelleza:
image

image

rberrelleza:
This is how the new params look. We could hide them, or add a note that they are only required if there's more than one active dev env in the same location?
image

rlamana:
I would divide the message into 2 lines: 
1. The error (red): There are more than one active envs...
2. The info/action (white): Use --file to specify which env...

PR #229, opened 2019-06-05
add forward info to the output

### LINKED ISSUE #227: Show forwarded ports at the end of the "okteto up" execution

rberrelleza:
This fixes #227 

This is how it looks for  a single port:
image

And this is how it looks for multiple ports:
image

### DIFF
--- a/cmd/up.go
+++ b/cmd/up.go
@@ -169,7 +169,7 @@ func (up *UpContext) Activate() {
 			log.Green("Reconnected to your cluster.")
 		}
 
-		printDisplayContext("Your Okteto Environment is ready", up.Dev.Name)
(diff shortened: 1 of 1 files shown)


rberrelleza:
@rlamana we could also align the texts, what do you think? (I don't really like it)
image

rlamana:
> @rlamana we could also align the texts, what do you think? (I don't really like it)

I actually like it, but the first port pair should appear on the same line as the label "Forward".

```
   Name:      vote
   Forward:   8090 -> 8090
              8080 -> 8080
```

pchico83:
@rberrelleza if the port cannot be opened, do we show a warning?

rberrelleza:
image
@rlamana thoughts?

rlamana:
> @rlamana thoughts?
It looks perfect to me ;)
Should we place the `name` before the `namespace`?

PR #238, opened 2019-06-05
trim volume name if needed

### LINKED ISSUE #239: Trim volume name if longer than 64 chars
From one of our users
```
Our deployment names are long, and we ran into an issue with the sidecar that Okteto injects: "Invalid value: "cnd-sync-data-jj-wayup-main-react-server-jj-wayup-main-react-server": must be no more than 63 characters."
```
Maybe we should use the same name as the deployment, and add labels to indicate it was created by okteto?

rberrelleza:
fixes #239

### DIFF
--- a/pkg/k8s/volumes/translate.go
+++ b/pkg/k8s/volumes/translate.go
@@ -43,7 +43,7 @@ func GetVolumeName(dev *model.Dev) string {
 
 //GetVolumeDataName returns the okteto volume name for a given dev environment
 func GetVolumeDataName(dev *model.Dev, i int) string {
-	n := fmt.Sprintf(oktetoVolumeDataTemplate, dev.Name, i)
+	n := fmt.Sprintf(oktetoVolumeDataTemplate, i, dev.Name)
 	if len(n) > 63 {
 		n = n[0:63]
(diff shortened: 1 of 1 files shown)


pkg/k8s/volumes/translate.go
```diff
@@ -33,10 +33,20 @@ func translate(name string) *apiv1.PersistentVolumeClaim {
...
+		n = n[0:63]
+	}
+
+	return n
 }
 
 //GetVolumeDataName returns the okteto volume name for a given dev environment
 func GetVolumeDataName(dev *model.Dev, i int) string {
-	return fmt.Sprintf(oktetoVolumeDataTemplate, dev.Name, i)
+	n := fmt.Sprintf(oktetoVolumeDataTemplate, dev.Name, i)
```
pchico83: I wouldn´t have the index at the end if we are trimming the volume name

rberrelleza: ok, would changing this break existing deployments?

pchico83: no. it would recreate the volume, but this feature is not documented yet

PR #245, opened 2019-06-06
Improve error handlers in the up command

pchico83:
## Proposed changes
- execute `kill -- -1` before `exec` to clean previous command executions. It cleans everything but the INIT process.
- restore the terminal when the connection fails.

### DIFF
--- a/cmd/up.go
+++ b/cmd/up.go
@@ -3,18 +3,18 @@ package cmd
 import (
 	"context"
 	"fmt"
-	"net/http"
 	"os"
-	"os/exec"
 	"os/signal"
 	"sync"
 	"time"
 
+	"github.com/docker/docker/pkg/term"
 	"github.com/okteto/okteto/pkg/analytics"
 	"github.com/okteto/okteto/pkg/config"
 	"github.com/okteto/okteto/pkg/errors"
 	k8Client "github.com/okteto/okteto/pkg/k8s/client"
 	"github.com/okteto/okteto/pkg/k8s/deployments"
+	"github.com/okteto/okteto/pkg/k8s/exec"
 	"github.com/okteto/okteto/pkg/k8s/pods"
 	"github.com/okteto/okteto/pkg/k8s/secrets"
 	"github.com/okteto/okteto/pkg/k8s/services"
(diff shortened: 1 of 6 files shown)


cmd/up.go
```diff
@@ -128,19 +125,26 @@ func RunUp(dev *model.Dev) error {
 
 // Activate activates the dev environment
 func (up *UpContext) Activate() {
-	up.WG.Add(1)
```
pchico83: No need to wait for `Activate`, we already have the `Exit` channel for this

cmd/up.go
```diff
@@ -152,67 +156,36 @@ func (up *UpContext) Activate() {
...
-		if err := cmd.Start(); err != nil {
-			log.Infof("Failed to execute okteto exec: %s", err)
-			up.Exit <- err
-			return
-		}
-
-		log.Debugf("started new okteto exec")
 
 		go func() {
-			up.WG.Add(1)
```
pchico83: No need to wait for the exec, this way we do not need to create a new process.
- If the command exists, it is notified via the `Running` channel.
- If not:
  a) the `okteto up` finishes due to connectivity issues and the `exec` is killed.
  b) `okteto up` retries and it will kill the exec with `kill -- -1`

cmd/up.go
```diff
@@ -152,67 +156,36 @@ func (up *UpContext) Activate() {
...
-		execEndpoint := fmt.Sprintf("http://127.0.0.1:%d", port)
-		prevError = up.WaitUntilExitOrInterrupt(execEndpoint)
-		if prevError != nil && (prevError == errors.ErrLostConnection ||
-			prevError == errors.ErrCommandFailed && !up.Sy.IsConnected()) {
-			log.Yellow("\nConnection lost to your Okteto Environment, reconnecting...")
-			fmt.Println()
-			up.shutdown()
-			continue
+		prevError := up.WaitUntilExitOrInterrupt()
+		term.RestoreTerminal(inFd, state)
```
pchico83: Restore the terminal as it was before the `exec` call.

cmd/up.go
```diff
@@ -128,19 +125,26 @@ func RunUp(dev *model.Dev) error {
...
+	state, err := term.SaveState(inFd)
+	if err != nil {
+		up.Exit <- err
+		return
+	}
 
 	for {
 		up.Context, up.Cancel = context.WithCancel(context.Background())
-		err := up.devMode(attach)
+		up.Disconnect = make(chan struct{}, 1)
```
pchico83: Fresh channels for each iteration

pkg/syncthing/monitor.go
```diff
@@ -21,16 +21,20 @@ func (s *Syncthing) IsConnected() bool {
...
-	ticker := time.NewTicker(5 * time.Second)
+	ticker := time.NewTicker(3 * time.Second)
+	connected := true
 	for {
 		select {
 		case <-ticker.C:
-			if !s.IsConnected() {
-				if !s.IsConnected() {
+			if s.IsConnected() {
+				connected = true
```
rberrelleza: what is this for? won't the loop exit the first time is not connected anyway?

pchico83: it is to fail for two consecutive connection issues

PR #249, opened 2019-06-07
Wait for persistent volume claim to be created

pchico83:


pkg/k8s/volumes/crud.go
```diff
@@ -1,32 +1,58 @@
...
+		}
 	}
 
-	log.Infof("creating volume claim '%s'...", pvc.Name)
-	_, err = vClient.Create(pvc)
-	if err != nil {
-		return fmt.Errorf("error creating kubernetes volume claim: %s", err)
+	tries := 0
+	ticker := time.NewTicker(1 * time.Second)
+	for k8Volume.Status.Phase != apiv1.ClaimBound {
```
rberrelleza: would this also catch the situation where the PVC is not scheduled due to lack of storage, or lack of storage class (like what happened in k3s) and stuff like that?

pchico83: I don´t remember the k3s case well. if the claim is not bound, this will fail.

PR #251, opened 2019-06-07
Syncthing running on its own stateful set

### LINKED ISSUE #217: Redirect syncthing and okteto init logs to a file to not pollute app pod logs
**Describe the bug**

**To Reproduce**

**Expected behavior**

 - OS: [e.g. macOS]

**Additional context**
Add any other context about the problem here (e.g cnd.yml, k8s manifests, logs)

pchico83:
Closes #217 #271 #264

### DIFF
--- a/pkg/k8s/code/crud.go
+++ b/pkg/k8s/code/crud.go
@@ -1,0 +1,71 @@
+package code
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/okteto/okteto/pkg/log"
+	"github.com/okteto/okteto/pkg/model"
+
+	appsv1 "k8s.io/api/apps/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+var devTerminationGracePeriodSeconds int64
+
+//Deploy creates or updates a database
+func Deploy(dev *model.Dev, c *kubernetes.Clientset) error {
+	ss := translate(dev)
+
+	if exists(ss, c) {
+		if err := update(ss, c); err != nil {
+			return err
+		}
+	} else {
+		if err := create(ss, c); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func exists(ss *appsv1.StatefulSet, c *kubernetes.Clientset) bool {
+	ss, err := c.AppsV1().StatefulSets(ss.Namespace).Get(ss.Name, metav1.GetOptions{})
+	if err != nil {
+		return false
+	}
+	return ss.Name != ""
+}
+
+func create(ss *appsv1.StatefulSet, c *kubernetes.Clientset) error {
+	log.Infof("creating code statefulset '%s", ss.Name)
+	_, err := c.AppsV1().StatefulSets(ss.Namespace).Create(ss)
+	if err != nil {
+		return fmt.Errorf("error creating kubernetes code statefulset: %s", err)
+	}
+	log.Infof("code statefulset '%s' created", ss.Name)
+	return nil
+}
+
+func update(ss *appsv1.StatefulSet, c *kubernetes.Clientset) error {
+	log.Infof("updating code statefulset '%s'", ss.Name)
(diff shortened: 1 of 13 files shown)


pkg/k8s/code/crud.go
```diff
@@ -0,0 +1,71 @@
+package code
```
rberrelleza: code?

pkg/k8s/code/crud.go
```diff
@@ -0,0 +1,71 @@
...
+	"github.com/okteto/okteto/pkg/model"
+
+	appsv1 "k8s.io/api/apps/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+var devTerminationGracePeriodSeconds int64
+
+//Deploy creates or updates a database
```
rberrelleza: I think this docstring should say something different?

pkg/k8s/code/crud.go
```diff
@@ -0,0 +1,71 @@
...
+		if err := create(ss, c); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func exists(ss *appsv1.StatefulSet, c *kubernetes.Clientset) bool {
+	ss, err := c.AppsV1().StatefulSets(ss.Namespace).Get(ss.Name, metav1.GetOptions{})
+	if err != nil {
```
rberrelleza: I've seen a Get on deployments return a 'not-found' error instead of an empty name. do we need to check for that here as well?

pkg/k8s/code/crud.go
```diff
@@ -0,0 +1,71 @@
...
+	log.Infof("code statefulset '%s' updated", ss.Name)
+	return nil
+}
+
+// Destroy destroys a database
+func Destroy(dev *model.Dev, c *kubernetes.Clientset) error {
+	log.Infof("destroying code statefulset '%s' ...", dev.Name)
+	sfsClient := c.AppsV1().StatefulSets(dev.Namespace)
+	if err := sfsClient.Delete(dev.Name, &metav1.DeleteOptions{GracePeriodSeconds: &devTerminationGracePeriodSeconds}); err != nil {
+		if strings.Contains(err.Error(), "not found") {
```
rberrelleza: I think this is missing a `!` in the condition?

pkg/k8s/code/translate.go
```diff
@@ -0,0 +1,164 @@
...
+)
+
+func translate(dev *model.Dev) *appsv1.StatefulSet {
+	initContainer := translateInitContainer(dev)
+
+	reqMem, _ := resource.ParseQuantity("64Mi")
+	reqCPU, _ := resource.ParseQuantity("50m")
+	limMem, _ := resource.ParseQuantity("256Mi")
+	limCPU, _ := resource.ParseQuantity("200m")
+	ss := &appsv1.StatefulSet{
```
rberrelleza: Why statefulspec vs deployment? (or even a pod directly)

pkg/k8s/code/translate.go
```diff
@@ -0,0 +1,164 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      dev.Name,
```
rberrelleza: I think we should prefix it with okteto, so it's obvious we created it.

pkg/k8s/code/crud.go
pchico83: any other idea? 
it is the package that creates the remote synthing stateful set, and manages it

pkg/k8s/code/translate.go
pchico83: it is for pods that do now play well with restarts, and they usually mount volumes. it is a better fit for this case.

pkg/k8s/code/translate.go
pchico83: ok

pkg/k8s/code/crud.go
pchico83: moved to `/syncthing/k8s/`

PR #265, opened 2019-06-13
fail if there's a quota error

rberrelleza:
image

### DIFF
--- a/pkg/errors/errors.go
+++ b/pkg/errors/errors.go
@@ -26,6 +26,9 @@ var (
 
 	// ErrInternalServerError is raised when an internal server error or similar is received
 	ErrInternalServerError = fmt.Errorf("internal server error, please try again")
+
+	// ErrQuota is returned when there aren't enough resources to enable dev mode
+	ErrQuota = fmt.Errorf("Quota exceeded, please free some resources and try again")
 )
 
 // IsNotFound returns true if err is of the type not found
--- a/pkg/k8s/pods/pod.go
+++ b/pkg/k8s/pods/pod.go
@@ -5,8 +5,13 @@ import (
 	"fmt"
 	"time"
 
+	"strings"
+
+	"github.com/okteto/okteto/pkg/errors"
+	"github.com/okteto/okteto/pkg/k8s/deployments"
 	"github.com/okteto/okteto/pkg/log"
 	"github.com/okteto/okteto/pkg/model"
+	appsv1 "k8s.io/api/apps/v1"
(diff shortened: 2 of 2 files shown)


pkg/errors/errors.go
```diff
@@ -26,6 +26,9 @@ var (
 
 	// ErrInternalServerError is raised when an internal server error or similar is received
 	ErrInternalServerError = fmt.Errorf("internal server error, please try again")
+
+	// ErrQuota is returned when there aren't enough resources to enable dev mode
+	ErrQuota = fmt.Errorf("Exceeded quota, please adjust your limits and resources and try again")
```
pchico83: or release a few resources too?

pkg/k8s/pods/pod.go
```diff
@@ -33,6 +37,13 @@ func GetDevPod(ctx context.Context, dev *model.Dev, c *kubernetes.Clientset) (*a
 			return nil, fmt.Errorf("failed to retrieve dev environment information")
 		}
 
+		if tries%30 == 0 && len(pods.Items) == 0 {
```
pchico83: could we do it every 10 retries?

pkg/k8s/pods/pod.go
```diff
@@ -66,6 +77,23 @@ func GetDevPod(ctx context.Context, dev *model.Dev, c *kubernetes.Clientset) (*a
...
+
+func isDeploymentFailed(namespace, name string, c *kubernetes.Clientset) error {
+	d, err := deployments.Get(name, namespace, c)
+	if err != nil {
+		log.Infof("failed to get deployment information: %s", err)
+		return nil
+	}
+
+	for _, c := range d.Status.Conditions {
+		if c.Type == appsv1.DeploymentReplicaFailure && c.Reason == failedCreateReason && c.Status == apiv1.ConditionTrue {
```
pchico83: is it possible to get a more specific check?
looks like we are not chequing any query specific message.
`kubectl describe`, which query the events, could give you more info.
Also, checking the status and|or events of the pod.

rberrelleza: this is the most specific check we can do. the value of `d.Status.Conditions` is the one you see when you run describe. 

We can't check anything on the pod, because it's not yet scheduled, that's why we have to check at the deployment level.

pkg/errors/errors.go
rberrelleza: @rlamana what do you think it's a good error message here?

PR #272, opened 2019-06-15
Update syncthing version

### LINKED ISSUE #261: Update to latest syncthing

pchico83:
Fixes #261

### DIFF
--- a/pkg/k8s/deployments/translate.go
+++ b/pkg/k8s/deployments/translate.go
@@ -18,13 +18,8 @@ import (
 )
 
 const (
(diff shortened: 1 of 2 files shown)


pchico83:
@rberrelleza This is missing to update the local syncthing binaries.
We should also detect that a new binary must be download. What would you suggest? To store something in the `.okteto` folder telling the syncthing version?

rberrelleza:
We could use the same approach we do for the okteto binary, and check the metadata of the file to see if the version doesn't match the one installed (we store that one in a constant).

PR #273, opened 2019-06-16
Multi service dev mode

### LINKED ISSUE #242: Command to start multiple okteto environments at the same time
Some developers need to start more than one service at a time when developing. Something similar to what 'docker-compose' does for docker would be very useful.

pchico83:
Closes #242

### DIFF
--- a/cmd/up.go
+++ b/cmd/up.go
@@ -26,6 +26,7 @@ import (
 	"github.com/okteto/okteto/pkg/syncthing"
 
 	"github.com/spf13/cobra"
+	appsv1 "k8s.io/api/apps/v1"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 )
@@ -44,7 +45,6 @@ type UpContext struct {
 	RestConfig    *rest.Config
 	DevPod        string
 	SyncPod       string
-	Container     string
 	DevForwarder  *forward.PortForwardManager
 	SyncForwarder *forward.PortForwardManager
 	Disconnect    chan struct{}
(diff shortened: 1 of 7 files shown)


cmd/up.go
```diff
@@ -290,26 +301,36 @@ func (up *UpContext) devMode(isRetry bool) error {
...
-	c, err := deployments.DevModeOn(d, up.Dev, create, up.Client)
-	if err != nil {
+	if err := deployments.TraslateDevMode(deploys, up.Dev); err != nil {
 		return err
 	}
 
-	up.Container = c.Name
+	for name, d := range deploys {
+		forceCreate := false
+		if d.Name == name {
```
rberrelleza: what is this check for?

pchico83: it is to handle de autocreation of the main service. for the other ones, for now, they must be already created

rberrelleza: could we have that separate from the loop so it's more obvious? Or maybe have a function called "isMainService" so it's easier to troubleshoot in the future?

pkg/k8s/deployments/translate.go
```diff
@@ -238,20 +258,3 @@ func translateOktetoVolumes(d *appsv1.Deployment, dev *model.Dev) {
 		d.Spec.Template.Spec.Volumes = append(d.Spec.Template.Spec.Volumes, v)
 	}
 }
-
-func getPullPolicy() apiv1.PullPolicy {
```
rberrelleza: Are we removing the support for this? I thought this was breaking quid's use ase

pchico83: I talked to alberto the other day. he said we should keep the policy defined in the deployment manifest. It is what we are doing for other fields, I think we should follow that approach here too.

PR #319, opened 2019-07-10
Fix translation for multi container development on the same pod

pchico83:


pkg/model/translation.go
```diff
@@ -0,0 +1,33 @@
...
+type Translation struct {
+	Interactive bool               `json:"-" yaml:"-"`
+	Name        string             `json:"-" yaml:"-"`
+	Deployment  *appsv1.Deployment `json:"-" yaml:"-"`
+	Rules       []*TranslationRule `json:"rules" yaml:"rules"`
+}
+
+//TranslationRule represents how to apply a container translation in a deployment
+type TranslationRule struct {
+	Node        string               `json:"node,omitempty" yaml:"node,omitempty"`
```
rberrelleza: can we only add the `json` or `yaml` tag but not both? (not sure which one we're using), it's a bit confusing.

PR #343, opened 2019-08-07
Support for RunAsUser

### LINKED ISSUE #342: Support for RunAsUser
I want to do `okteto up` to an existing deployment that I cannot modify. This deployment does not have security context defined, but the container does not run as root.

Define the security context in the okteto manifest.

**Additional context**
This is required to do `okteto up` to an openfass deployment, for example.

pchico83:
Closes #342

### DIFF
--- a/pkg/k8s/deployments/translate.go
+++ b/pkg/k8s/deployments/translate.go
@@ -91,6 +91,13 @@ func translate(t *model.Translation) error {
 		devContainer := GetDevContainer(&t.Deployment.Spec.Template.Spec, rule.Container)
 		TranslateDevContainer(devContainer, rule)
 		TranslateOktetoVolumes(&t.Deployment.Spec.Template.Spec, rule)
+		if rule.SecurityContext != nil {
+			t.Deployment.Spec.Template.Spec.SecurityContext = &apiv1.PodSecurityContext{
+				RunAsUser:  rule.SecurityContext.RunAsUser,
+				RunAsGroup: rule.SecurityContext.RunAsGroup,
+				FSGroup:    rule.SecurityContext.FSGroup,
+			}
+		}
 	}
 	return nil
 }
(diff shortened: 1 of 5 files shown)


pkg/k8s/deployments/translate.go
```diff
@@ -91,6 +91,13 @@ func translate(t *model.Translation) error {
 		devContainer := GetDevContainer(&t.Deployment.Spec.Template.Spec, rule.Container)
 		TranslateDevContainer(devContainer, rule)
 		TranslateOktetoVolumes(&t.Deployment.Spec.Template.Spec, rule)
+		if rule.RunAsUser != nil {
+			t.Deployment.Spec.Template.Spec.SecurityContext = &apiv1.PodSecurityContext{
+				RunAsUser:  rule.RunAsUser,
+				RunAsGroup: rule.RunAsUser,
```
rberrelleza: can we have runasgroup as different, and not put it if nil?

pkg/k8s/deployments/translate.go
```diff
@@ -91,6 +91,13 @@ func translate(t *model.Translation) error {
...
+				FSGroup:    rule.RunAsUser,
```
rberrelleza: fsgroup is tricky. I'd not put it if we don't have a specific use case for this. If we do, I'd add it as a separate value as well.

pkg/k8s/deployments/translate_test.go
```diff
@@ -11,11 +11,17 @@ import (
...
 func Test_translate(t *testing.T) {
+	var userID int64
+	userID = 100
 	manifest := []byte(`name: web
 container: dev
 image: web:latest
 command: ["./run_web.sh"]
 workdir: /app
+securityContext:
+  runAsUser: 100
```
rberrelleza: I'd use a different value for each of them

rberrelleza: Also, I'd add a diferent test where only one or two values are used

PR #349, opened 2019-08-08
PullAlways by default and support local development with special flag

### LINKED ISSUE #346: Allow to overwrite pullPolicy
I want to pull if not present in my prod manifest, but pull always in dev.

pchico83:
Closes #346


pkg/k8s/deployments/translate_test.go
```diff
@@ -27,6 +27,7 @@ services:
   - name: worker
     container: dev
     image: worker:latest
+    localImage: true
```
rberrelleza: could we instead have an ImagePullPolicy attribute instead, to avoid introducing a new concept on the manifest?

PR #356, opened 2019-08-13
Add CII badge

rberrelleza:
This is displayed by the landscape. We're currently at 18%, we might want to improve our rating before publishing :)


rberrelleza:
Updated the checks, we are at 70% now.

PR #361, opened 2019-08-16
improve okteto down errors

### LINKED ISSUE #307: Improve "okteto down -v" errors
`okteto down -v` blocks if the volume cannot be deleted.

Inform about why the volume cannot be deleted.

**Additional context**
The volume could be still in-use by an `Evicted` pod.

rberrelleza:
Fixes #307 

This change adds a 300s timeout for the volume claim to be deleted. 
It also adds a check to verify that the PVC is not  still attached to other pods.

### DIFF
--- a/pkg/k8s/volumes/crud.go
+++ b/pkg/k8s/volumes/crud.go
@@ -49,8 +49,15 @@ func checkIfAttached(pvc string, dev *model.Dev, c *kubernetes.Clientset) error 
 		return nil
 	}
 
+	prefixes := []string{dev.Name}
+	for _, s := range dev.Services {
+		prefixes = append(prefixes, s.Name)
+	}
+
 	for _, p := range pods.Items {
-		if strings.HasPrefix(p.Name, dev.Name) {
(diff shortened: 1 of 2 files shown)


pkg/k8s/volumes/crud.go
```diff
@@ -13,18 +13,77 @@ import (
...
+		<-ticker.C
+		if i%10 == 5 {
+			log.Infof("waiting for volume claim '%s' to be destroyed...", name)
+		}
+	}
+
+	return fmt.Errorf("volume claim '%s' wasn't destroyed after 300s", name)
+}
+
+func checkIfAttached(pvc string, dev *model.Dev, c *kubernetes.Clientset) error {
```
pchico83: This code does not work with labels. I would check if pods are scheduled for termination:
https://github.com/okteto/okteto/blob/master/pkg/k8s/pods/pod.go#L64

rberrelleza: ah, good point. I'll check that way.

PR #367, opened 2019-08-17
Update sync pods in parallel

### LINKED ISSUE #358: Dev environment doesn't get the right container image
**Describe the bug**

**To Reproduce**
1. Run your deployment
2. `okteto up`
3. exit the dev env
4. Change the image in `okteto.yml`
5. `okteto up`

**Expected behavior**
I'd expect my new image to be applied. Instead, I get the old one. The same thing happened even if I did `okteto down`.  The only way to get a new image was to do `okteto down -v`

 - OS: macos
 - Version 1.3.0

pchico83:
Closes #358

### DIFF
--- a/pkg/syncthing/k8s/crud.go
+++ b/pkg/syncthing/k8s/crud.go
@@ -20,8 +20,9 @@ var devTerminationGracePeriodSeconds int64
 func Deploy(dev *model.Dev, d *appsv1.Deployment, c *apiv1.Container, client *kubernetes.Clientset) error {
 	ss := translate(dev, d, c)
 
-	if exists(ss, client) {
+	if old := exists(ss, client); old != nil {
 		ss.Spec.Template.Spec.NodeName = pods.GetSyncNode(dev, client)
+		ss.Spec.PodManagementPolicy = old.Spec.PodManagementPolicy
 		if err := update(ss, client); err != nil {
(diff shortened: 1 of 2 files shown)


pkg/syncthing/k8s/crud.go
```diff
@@ -20,8 +20,9 @@ var devTerminationGracePeriodSeconds int64
 func Deploy(dev *model.Dev, d *appsv1.Deployment, c *apiv1.Container, client *kubernetes.Clientset) error {
 	ss := translate(dev, d, c)
 
-	if exists(ss, client) {
+	if old := exists(ss, client); old != nil {
 		ss.Spec.Template.Spec.NodeName = pods.GetSyncNode(dev, client)
+		ss.Spec.PodManagementPolicy = old.Spec.PodManagementPolicy
```
pchico83: This is to avoid forcing `okteto down -v` if you already have a dev environment running, since changing `PodManagementPolicy` is an incompatible statefulset change

rberrelleza: When would `okteto down -v` be required?

Is there a way to catch the error message and tell the user when it's needed?

pchico83: It is not required because we only do this for new `okteto up`s

PR #371, opened 2019-08-20
fix spacing and error message

rberrelleza:


main.go
```diff
@@ -62,7 +62,7 @@ func main() {
 		log.Fail(err.Error())
 		if uErr, ok := err.(errors.UserError); ok {
 			if len(uErr.Hint) > 0 {
-				log.Hint(uErr.Hint)
+				log.Hint("    %s", uErr.Hint)
```
rberrelleza: This is so the hint is aligned with the error message, the same way we do it for the success messages.

PR #372, opened 2019-08-20
wait for pod to be ready on up

rberrelleza:


cmd/exec.go
```diff
@@ -72,7 +72,7 @@ func executeExec(ctx context.Context, dev *model.Dev, args []string) error {
 		dev.Namespace = namespace
 	}
 
-	p, err := pods.GetDevPod(ctx, dev, pods.OktetoInteractiveDevLabel, client)
+	p, err := pods.GetByLabel(ctx, dev, pods.OktetoInteractiveDevLabel, client, false)
```
rberrelleza: we need this in exec to fail the command if the pod doesn't exist.

PR #373, opened 2019-08-20
timeout 90s

rberrelleza:


pkg/k8s/volumes/crud.go
```diff
@@ -43,7 +39,12 @@ func Destroy(name string, dev *model.Dev, c *kubernetes.Clientset) error {
 		}
 	}
 
-	return fmt.Errorf("volume claim '%s' wasn't destroyed after 300s", name)
+	if err := checkIfAttached(name, dev, c); err != nil {
```
rberrelleza: we use this now to give the user a clue of why the PVC wasn't deleted

PR #375, opened 2019-08-22
Integrate remote builds

### LINKED ISSUE #257: Enable remote docker builds
I want to build on remote

### LINKED ISSUE #355: Inject docker,kubectl,okteto... binaries if namespace is managed by okteto
Also, set the `DOCKER_HOST` envvar to allow remote builds.

pchico83:
Closes #355
Closes #257

### DIFF
--- a/cmd/up.go
+++ b/cmd/up.go
@@ -437,7 +437,7 @@ func (up *UpContext) runCommand() error {
 		in,
 		os.Stdout,
 		os.Stderr,
-		[]string{"sh", "-c", "trap '' TERM && kill -- -1 && sleep 0.1 & kill -s KILL -- -1 >/dev/null 2>&1"},
+		[]string{"sh", "-c", "((cp -n /var/okteto/bin /usr/local/bin); (trap '' TERM && kill -- -1 && sleep 0.1 & kill -s KILL -- -1 >/dev/null)) 2>&1"},
 	); err != nil {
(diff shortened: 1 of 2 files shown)


cmd/up.go
```diff
@@ -443,6 +443,8 @@ func (up *UpContext) runCommand() error {
 	}
 
 	log.Infof("starting remote command")
+	command := strings.Join(up.Dev.Command, "\" \"")
```
rberrelleza: could we move this to a separate function and add tests?

cmd/up.go
```diff
@@ -443,6 +443,8 @@ func (up *UpContext) runCommand() error {
...
+	command = fmt.Sprintf("PATH=$PATH:/var/okteto/bin \"%s\"", command)
```
rberrelleza: I wonder if this is somethign we should instead do in the dev env container image

pchico83: All the point is to make them available and transparently in every dev image.
I have tried another approach that I like it better.
We run a custom exec command before the user start command to kill previouos command executions. We can add logic there to move the binaries from `/var/okteto/bin` to `/usr/local/bin` if they don't exist in `/usr/local/bin` (`cp -n`).

pchico83:
@rberrelleza merging. `OKTETO_CONTINUOUS_DEVELOPMENT` works as feature flag for remote builds too

PR #380, opened 2019-08-28
add securitycontext

rberrelleza:
Adds the ability to specify capabilities to the manifest. This will only be applied on the dev pod

### DIFF
--- a/pkg/k8s/deployments/translate.go
+++ b/pkg/k8s/deployments/translate.go
@@ -262,11 +262,6 @@ func translateSecurityContext(c *apiv1.Container, s *model.SecurityContext) {
 		c.SecurityContext.Capabilities = &apiv1.Capabilities{}
 	}
 
-	for _, a := range s.Capabilities.Add {
-		c.SecurityContext.Capabilities.Add = append(c.SecurityContext.Capabilities.Add, a)
-	}
-
-	for _, d := range s.Capabilities.Drop {
-		c.SecurityContext.Capabilities.Drop = append(c.SecurityContext.Capabilities.Drop, d)
-	}
+	c.SecurityContext.Capabilities.Add = append(c.SecurityContext.Capabilities.Add, s.Capabilities.Add...)
+	c.SecurityContext.Capabilities.Drop = append(c.SecurityContext.Capabilities.Drop, s.Capabilities.Drop...)
 }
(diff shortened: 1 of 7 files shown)


pkg/k8s/deployments/translate.go
```diff
@@ -246,3 +248,25 @@ func TranslateOktetoVolumes(spec *apiv1.PodSpec, rule *model.TranslationRule) {
...
+
+	if c.SecurityContext == nil {
+		c.SecurityContext = &apiv1.SecurityContext{}
+	}
+
+	if c.SecurityContext.Capabilities == nil {
+		c.SecurityContext.Capabilities = &apiv1.Capabilities{}
+	}
+
+	for _, a := range s.Capabilities.Add {
```
golangcibot: S1011: should replace loop with `c.SecurityContext.Capabilities.Add = append(c.SecurityContext.Capabilities.Add, s.Capabilities.Add...)` (from `gosimple`)

pkg/k8s/deployments/translate.go
```diff
@@ -246,3 +248,25 @@ func TranslateOktetoVolumes(spec *apiv1.PodSpec, rule *model.TranslationRule) {
...
+		c.SecurityContext.Capabilities.Add = append(c.SecurityContext.Capabilities.Add, a)
+	}
+
+	for _, d := range s.Capabilities.Drop {
```
golangcibot: S1011: should replace loop with `c.SecurityContext.Capabilities.Drop = append(c.SecurityContext.Capabilities.Drop, s.Capabilities.Drop...)` (from `gosimple`)

pkg/k8s/deployments/translate.go
```diff
@@ -246,3 +248,25 @@ func TranslateOktetoVolumes(spec *apiv1.PodSpec, rule *model.TranslationRule) {
 		spec.Volumes = append(spec.Volumes, v)
 	}
 }
+
+func translateSecurityContext(c *apiv1.Container, s *model.SecurityContext) {
```
pchico83: We should do something similar here:
https://github.com/okteto/okteto/blob/master/pkg/k8s/deployments/translate.go#L96
Check if it is defined in the pod or not, and only update the values that are not nil in dev.SecurityContext

PR #386, opened 2019-09-02
Reduce rescanIntervals from 1hour to 60s

pchico83:


rberrelleza:
Could we have CI coverage for this class of issues, since it has caused so much pain to our users? At least to catch regressions next time we modify the values, or upgrade syncthing.

PR #388, opened 2019-09-02
Improve synchronization sequence

pchico83:
Fixes use cases where remote files are in conflict with local changes and the initial sync cannot complete. The solution is to avoid waiting for initial sync before overriding changes.

The new flow is:
- run remote syncthing
- wait for remote syncthing to finish the initial scan
- run local syncthing
- wait for local synthing to finish initial scan (this is required before calling override changes).
- override local changes to remote
- wait for remote to complete

### DIFF
--- a/cmd/up.go
+++ b/cmd/up.go
@@ -287,10 +287,6 @@ func (up *UpContext) sync(d *appsv1.Deployment, c *apiv1.Container) error {
 		return err
 	}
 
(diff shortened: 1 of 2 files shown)


rberrelleza:
I can haz tests please?

pchico83:
@rberrelleza syncthings must be started in sequence to check their scans independently, but starting the local one is almost immediately

PR #391, opened 2019-09-03
static build to support alpine

### LINKED ISSUE #382: Build an alpine compatible okteto binary
I want to run okteto inside an alpine container

Build with flags `GO_BUILD_TAGS_linux := "osusergo netgo static_build"`

rberrelleza:
Fixes #382


rberrelleza:
> Are those flags valid in every platform?

The CI is building it in all platforms, so I'm assuming it would fail to build the binary if it couldn't build. We should do a windows test before releasing though.

PR #393, opened 2019-09-04
Sync .stignore to reduce remote scans

pchico83:


pkg/syncthing/syncthing.go
```diff
@@ -267,7 +262,40 @@ func (s *Syncthing) waitForPing(ctx context.Context, wg *sync.WaitGroup, local b
...
+		log.Errorf("'.stignore' not sent to the remote container: %s", err)
+	}
+	_, err = s.APICall("rest/db/ignores", "POST", 200, params, false, body)
+	if err != nil {
+		log.Infof("error posting 'rest/db/ignores' syncthing API: %s", err)
+		log.Errorf("'.stignore' not sent to the remote container: %s", err)
+		return
+	}
+	log.Infof("'.stignore' sent to the remote container")
+	return
```
golangcibot: S1023: redundant `return` statement (from `gosimple`)

pkg/syncthing/syncthing.go
```diff
@@ -267,7 +262,39 @@ func (s *Syncthing) waitForPing(ctx context.Context, wg *sync.WaitGroup, local b
...
-func (s *Syncthing) waitForScanning(ctx context.Context, wg *sync.WaitGroup, dev *model.Dev, local bool) error {
+//SendStignoreFile sends .stignore from local to remote
+func (s *Syncthing) SendStignoreFile(ctx context.Context, wg *sync.WaitGroup, dev *model.Dev) {
+	log.Infof("Sending '.stignore' file to the remote container...")
+	folder := fmt.Sprintf("okteto-%s", dev.Name)
+	params := map[string]string{"folder": folder}
+	ignores := &Ignores{}
+	body, err := s.APICall("rest/db/ignores", "GET", 200, params, true, nil)
+	if err != nil {
+		log.Infof("error getting 'rest/db/ignores' syncthing API: %s", err)
```
rberrelleza: why two logs?

pkg/syncthing/syncthing.go
```diff
@@ -267,7 +262,39 @@ func (s *Syncthing) waitForPing(ctx context.Context, wg *sync.WaitGroup, local b
...
+		log.Infof("error unmarshaling 'rest/db/ignores': %s", err)
+		log.Errorf("'.stignore' not sent to the remote container: %s", err)
+		return
+	}
+	body, err = json.Marshal(ignores)
+	if err != nil {
+		log.Infof("error marshaling 'rest/db/ignores': %s", err)
+		log.Errorf("'.stignore' not sent to the remote container: %s", err)
+	}
+	_, err = s.APICall("rest/db/ignores", "POST", 200, params, false, body)
```
rberrelleza: took me a bit to catch that we were using `true` vs `false`. could you add a comment mentioning that this is to the remote? (or maybe have two functions, `LocalAPICall` and `RemoteAPICall` so is clearer?)

pkg/syncthing/syncthing.go
```diff
@@ -267,7 +262,39 @@ func (s *Syncthing) waitForPing(ctx context.Context, wg *sync.WaitGroup, local b
...
+	if err != nil {
+		log.Infof("error posting 'rest/db/ignores' syncthing API: %s", err)
+		log.Errorf("'.stignore' not sent to the remote container: %s", err)
```
rberrelleza: Error level logs are displayed to the user, is that intended? I prefer the packages to send the error up, and let the caller at the CMD pkg level decide if it should be displayed.

pkg/syncthing/api.go
```diff
@@ -26,14 +27,14 @@ func NewAPIClient() *http.Client {
...
-func (s *Syncthing) APICall(url, method string, code int, params map[string]string, local bool) ([]byte, error) {
+func (s *Syncthing) APICall(url, method string, code int, params map[string]string, local bool, body []byte) ([]byte, error) {
 	var urlPath string
 	if local {
 		urlPath = path.Join(s.GUIAddress, url)
 	} else {
 		urlPath = path.Join(s.RemoteGUIAddress, url)
 	}
-	req, err := http.NewRequest(method, fmt.Sprintf("http://%s", urlPath), nil)
+	req, err := http.NewRequest(method, fmt.Sprintf("http://%s", urlPath), bytes.NewBuffer(body))
```
rberrelleza: should we check if we get a GET/DELETE with a non nil body?

rberrelleza:
I can haz tests?

pkg/syncthing/api.go
pchico83: no, nil argument is fine

PR #396, opened 2019-09-05
Set deployment annotation before setting other annotations/labels

pchico83:
This way, `okteto down` rollbacks to the correct state


pchico83:
@rberrelleza https://github.com/okteto/okteto/issues/400

PR #404, opened 2019-09-06
strip breaklie when checking versiono

rberrelleza:


cmd/version.go
```diff
@@ -43,7 +44,7 @@ func upgradeAvailable() string {
 	if len(v) > 0 {
 		latest, err := semver.NewVersion(v)
 		if err != nil {
-			log.Errorf("failed to parse latest version: %s", err)
+			log.Info("failed to parse latest version '%s': %s", v, err)
```
golangcibot: printf: Info call has possible formatting directive %s (from `govet`)

PR #409, opened 2019-09-10
Run syncthing in-process

rberrelleza:
Instead of starting a separate process, we can run syncthing in a go-routing. This would make it easier to control since we only have one process to deal with. This is particularly usefulwith the VSCode integration, since the lifecycle of `up` becomes simpler.

At this point it's an experiment. I haven't done a lot of tests yet to see if there are downsides to this approach. But I think is worth considering.

One downside is that this makes the binary go from 50MB to 75MB.


go.mod
```diff
@@ -73,6 +59,7 @@ require (
 )
 
 replace (
+	github.com/syncthing/syncthing => github.com/rberrelleza/syncthing v1.2.2-gen1
```
pchico83: why is this?

pkg/syncthing/syncthing.go
```diff
@@ -385,19 +327,21 @@ func (s *Syncthing) WaitForCompletion(ctx context.Context, wg *sync.WaitGroup, d
...
+	status := s.app.Stop(syncthing.ExitSuccess)
+	if status == syncthing.ExitError {
+		log.Infof("failed to restart syncthing")
+		return fmt.Errorf("internal server error")
+	}
 
-	if err := s.cleanupDaemon(pidPath); err != nil {
+	s.app.Wait()
+	log.Infof("syncthing stopped")
+	if err := s.initSyncthingApp(); err != nil {
```
pchico83: why are we restarting here?

rberrelleza: it was already like that before. I believe is when the change in `sendtype` happens.

go.mod
rberrelleza: Forgot to mention on the PR. the syncthing repo generates some source code when building the binary (like the UI's HTML and CSS assets). I forked the repo,checked out the 1.2.2 branch,  generated the assets file, commited it and pushed it to my fork so we can directly import it in our binary.

rberrelleza:
superseded by #466

PR #410, opened 2019-09-10
overwrite the yes/no question via an env-var

rberrelleza:
secret env-var to avoid asking yes/no when running okteto in an integrated environment (like vscode or CI/CD)


cmd/utils.go
```diff
@@ -23,6 +23,10 @@ func loadDev(devPath string) (*model.Dev, error) {
 }
 
 func askYesNo(q string) bool {
+	if len(os.Getenv("OKTETO_YES")) > 0 {
```
pchico83: `OKTETO_ASK_FOR_DEVENV_CREATION_CONFIRMATION`?

rberrelleza: that's way too long even for a copy paste. Maybe  `OKTETO_SILENT` or `OKTETO_AUTODEPLOY` ?

PR #415, opened 2019-09-11
Check quotas before running okteto up

pchico83:


pkg/k8s/namespaces/crud.go
```diff
@@ -21,3 +27,100 @@ func IsOktetoNamespace(ns string, c *kubernetes.Clientset) bool {
 	}
 	return n.Labels[OktetoLabel] == "true"
 }
+
+//CheckAvailableResources checks if therre are available resources for activating the dev env
+func CheckAvailableResources(dev *model.Dev, create bool, d *appsv1.Deployment, c *kubernetes.Clientset) error {
+	needCPU := resource.MustParse("0")
+	needMemory := resource.MustParse("0")
+	if !existsSyncthing(dev, c) {
+		needCPU.Add(resource.MustParse("500m"))
```
rberrelleza: can we get this values from the deployment object so we don't hard code them here?

pkg/k8s/namespaces/crud.go
```diff
@@ -21,3 +27,100 @@ func IsOktetoNamespace(ns string, c *kubernetes.Clientset) bool {
...
+		needMemory.Add(resource.MustParse("256Mi"))
+	}
+	lr, err := c.CoreV1().LimitRanges(dev.Namespace).Get(dev.Namespace, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+	needCPU.Add(matchCPU(dev, true, 1, d.Spec.Template.Spec, lr))
+	needMemory.Add(matchMemory(dev, true, 1, d.Spec.Template.Spec, lr))
+	if !create {
```
rberrelleza: what do the `create` arg means?

pkg/k8s/namespaces/crud.go
```diff
@@ -21,3 +27,100 @@ func IsOktetoNamespace(ns string, c *kubernetes.Clientset) bool {
...
+	if err != nil {
+		return err
+	}
+	usedCPU := rq.Status.Used[apiv1.ResourceLimitsCPU]
+	usedCPU.Add(needCPU)
+	totalCPU := rq.Status.Hard[apiv1.ResourceLimitsCPU]
+	if usedCPU.Value() > totalCPU.Value() {
+		usedCPU.Sub(needCPU)
+		totalCPU.Sub(usedCPU)
+		return fmt.Errorf("Activating your dev environment requires %s cpu and your remaining quota is %s cpu", needCPU.String(), totalCPU.String())
```
rberrelleza: Do you think we should send a more generic error, and then have the UI display this as a `hint`? (the blue text)

pkg/k8s/namespaces/crud.go
```diff
@@ -21,3 +27,100 @@ func IsOktetoNamespace(ns string, c *kubernetes.Clientset) bool {
 	}
 	return n.Labels[OktetoLabel] == "true"
 }
+
+//CheckAvailableResources checks if therre are available resources for activating the dev env
+func CheckAvailableResources(dev *model.Dev, create bool, d *appsv1.Deployment, c *kubernetes.Clientset) error {
```
rberrelleza: is it possible to split the part where we call kubernetes api from the rest? that way we can add unit tests to the limit calculation and validate the different scenarios.

pkg/k8s/namespaces/crud.go
```diff
@@ -21,3 +27,100 @@ func IsOktetoNamespace(ns string, c *kubernetes.Clientset) bool {
...
+		usedMemory.Sub(needMemory)
+		totalMemory.Sub(usedMemory)
+		return fmt.Errorf("Activating your dev environment requires %s of memory and your remaining quota is %s of memory", needMemory.String(), totalMemory.String())
+	}
+	return nil
+}
+
+func existsSyncthing(dev *model.Dev, c *kubernetes.Clientset) bool {
+	ss, err := c.AppsV1().StatefulSets(dev.Namespace).Get(dev.GetStatefulSetName(), metav1.GetOptions{})
+	if err != nil || ss.Name == "" {
```
rberrelleza: can we log the error?

pkg/k8s/namespaces/crud.go
```diff
@@ -21,3 +27,100 @@ func IsOktetoNamespace(ns string, c *kubernetes.Clientset) bool {
...
+	return true
+}
+
+func matchCPU(dev *model.Dev, isDev bool, replicas int32, spec apiv1.PodSpec, lr *apiv1.LimitRange) resource.Quantity {
+	cpuPod := resource.MustParse("0")
+	for _, c := range spec.Containers {
+		v, ok := dev.Resources.Limits[apiv1.ResourceCPU]
+		if !isDev || !ok {
+			if c.Resources.Limits == nil {
+				cpuPod.Add(lr.Spec.Limits[0].Default[apiv1.ResourceCPU])
```
rberrelleza: why do we add a value here if there isn't a limit in the deployment?

pkg/k8s/namespaces/crud.go
```diff
@@ -21,3 +27,100 @@ func IsOktetoNamespace(ns string, c *kubernetes.Clientset) bool {
...
+			} else {
+				cpuPod.Add(c.Resources.Limits[apiv1.ResourceCPU])
+			}
+		} else {
+			cpuPod.Add(v)
+		}
+	}
+	if isDev {
+		return cpuPod
```
rberrelleza: is this branch because we only have a single replica of dev? Could we remove this branch and just let the loop in line 99 take care of it?

pkg/k8s/namespaces/crud.go
```diff
@@ -21,3 +27,100 @@ func IsOktetoNamespace(ns string, c *kubernetes.Clientset) bool {
...
+		return false
+	}
+	return true
+}
+
+func matchCPU(dev *model.Dev, isDev bool, replicas int32, spec apiv1.PodSpec, lr *apiv1.LimitRange) resource.Quantity {
+	cpuPod := resource.MustParse("0")
+	for _, c := range spec.Containers {
+		v, ok := dev.Resources.Limits[apiv1.ResourceCPU]
+		if !isDev || !ok {
```
rberrelleza: I don't get this condition. Why the `!isDev` part?

pkg/k8s/namespaces/crud.go
```diff
@@ -21,3 +27,100 @@ func IsOktetoNamespace(ns string, c *kubernetes.Clientset) bool {
...
+				memoryPod.Add(c.Resources.Limits[apiv1.ResourceMemory])
+			}
+		} else {
+			memoryPod.Add(v)
+		}
+	}
+	if isDev {
+		return memoryPod
+	}
+	meemoryTotal := memoryPod
```
rberrelleza:
```suggestion
	memoryTotal := memoryPod
```

pkg/k8s/namespaces/crud.go
```diff
@@ -21,3 +27,100 @@ func IsOktetoNamespace(ns string, c *kubernetes.Clientset) bool {
...
+		return cpuPod
+	}
+	cpuTotal := cpuPod
+	for i := 1; int32(i) < replicas; i++ {
+		cpuTotal.Add(cpuPod)
+	}
+	return cpuTotal
+}
+
+func matchMemory(dev *model.Dev, isDev bool, replicas int32, spec apiv1.PodSpec, lr *apiv1.LimitRange) resource.Quantity {
```
rberrelleza: why do we call the function `matchMemory`?

cmd/up.go
```diff
@@ -196,6 +197,13 @@ func (up *UpContext) Activate() {
 			create = true
 		}
 
+		if namespaces.IsOktetoNamespace(up.Dev.Namespace, up.Client) {
```
rberrelleza: why do we only do it for okteto namespaces?

pkg/k8s/namespaces/crud.go
rberrelleza: it would also make it easier to follow if we had at least one function where we collect the  information from the namespace, another where we calculate the needs of the `up` and a third one that makes the decision.

PR #417, opened 2019-09-14
Dynamic translation for vscode remote execution integration

pchico83:


cmd/up.go
```diff
@@ -102,6 +104,7 @@ func Up() *cobra.Command {
 
 	cmd.Flags().StringVarP(&devPath, "file", "f", defaultManifest, "path to the manifest file")
 	cmd.Flags().StringVarP(&namespace, "namespace", "n", "", "namespace where the up command is executed")
+	cmd.Flags().BoolVarP(&vscodeRemote, "vscode-remote", "", false, "configures vscode remote development")
```
rberrelleza: `--ssh` or `--remote` maybe?

pchico83: `--remote`? we do more than ssh, we also set capabilities, add volumes...

PR #419, opened 2019-09-16
mout the vscode folder as a volume

rberrelleza:


pkg/model/dev.go
```diff
@@ -254,6 +254,14 @@ func (dev *Dev) LoadRemote(load bool) {
 			Remote: 22000,
 		},
 	)
+
+	if dev.Volumes == nil {
```
pchico83: we init `volumes` when reading the okteto manifest, it is never nil

PR #420, opened 2019-09-16
select a port for remote

rberrelleza:


cmd/up.go
```diff
@@ -102,7 +104,7 @@ func Up() *cobra.Command {
 
 	cmd.Flags().StringVarP(&devPath, "file", "f", defaultManifest, "path to the manifest file")
 	cmd.Flags().StringVarP(&namespace, "namespace", "n", "", "namespace where the up command is executed")
-	cmd.Flags().BoolVarP(&remote, "remote", "r", false, "configures remote execution")
+	cmd.Flags().Int32VarP(&remote, "remote", "r", 0, "configures remote execution on the specified port")
```
pchico83: Default to 22000?
It is weird that a flag is set but it does nothing

PR #422, opened 2019-09-17
Check for duplicated volumes

pchico83:
I want to run `okteto up` and then `okteto up --remote 22000`, but it gives me an error because of adding an extra volume.
This way I can enable that scenario.

### DIFF
--- a/pkg/model/dev.go
+++ b/pkg/model/dev.go
@@ -176,14 +176,7 @@ func (dev *Dev) setDefaults() error {
 	if dev.Labels == nil {
 		dev.Labels = map[string]string{}
 	}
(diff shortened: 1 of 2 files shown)


pkg/model/dev.go
```diff
@@ -176,6 +176,14 @@ func (dev *Dev) setDefaults() error {
 	if dev.Labels == nil {
 		dev.Labels = map[string]string{}
 	}
+	vMap := map[string]bool{}
```
rberrelleza: maybe extrapolate this to a `unique([]string)` function that returns the array without duplicates? we could use that in a few places.

Issue #423, opened 2019-09-18
More information on the sync status


rberrelleza:
**Is your feature request related to a problem? Please describe.**
It's hard to tell if a file has been synchronized or not. I end up having to check the timestamp or the file, or just `cat` the content when I'm not sure if a change has been applied or not. It would be useful to see what was synched, when, and if something went wrong without going to the logs.

rlamana:
Maybe we could add some sort of `status` command to output a diff of the local and remote directories and the current synchronization status.

It could be interesting not only for the case you described but also for other software to consume. For example, our VSCode plugin could use this output to visually mark those files still being synched in the file explorer.

pchico83:
@rlamana what about showing something in the okteto terminal?
We could have a bash plugin or something that shows the status all the time

pchico83:
Closed by #713

PR #427, opened 2019-09-20
Wait for completion based on events

pchico83:
In order to implement #423 I need to check the completion state based on events.
`/rest/db/completion` is high cpu intensive as reported in syncthing documentation:
https://docs.syncthing.net/rest/db-completion-get.html

After this change, I can implement the `okteto status` command using the same logic than here

### DIFF
--- a/pkg/syncthing/syncthing.go
+++ b/pkg/syncthing/syncthing.go
@@ -82,12 +82,15 @@ type Status struct {
 	State string `json:"state"`
 }
 
+//Event represents a syncthing event
+type Event struct {
+	ID   int        `json:"id,omitempty"`
+	Data Completion `json:"data,omitempty"`
+}
+
 // Completion represents the completion status of a syncthing folder.
 type Completion struct {
-	Completion  float64 `json:"completion"`
-	GlobalBytes int64   `json:"globalBytes"`
-	NeedBytes   int64   `json:"needBytes"`
-	NeedDeletes int64   `json:"needDeletes"`
+	Completion float64 `json:"completion"`
 }
 
 // New constructs a new Syncthing.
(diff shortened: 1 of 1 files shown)


pkg/syncthing/syncthing.go
```diff
@@ -329,64 +332,44 @@ func (s *Syncthing) WaitForScanning(ctx context.Context, wg *sync.WaitGroup, dev
...
-			}
-		case <-ctx.Done():
-			log.Debug("cancelling call to 'rest/db/completion'")
-			return ctx.Err()
-		}
-
+	params := map[string]string{
+		"events":  "FolderCompletion",
+		"device":  DefaultRemoteDeviceID,
+		"folder":  fmt.Sprintf("okteto-%s", dev.Name),
```
rberrelleza: Is there a function we can use to get the name  of the folder instead?

pkg/syncthing/syncthing.go
```diff
@@ -329,64 +332,44 @@ func (s *Syncthing) WaitForScanning(ctx context.Context, wg *sync.WaitGroup, dev
...
-			progress,
-			completion.NeedBytes,
-			completion.NeedDeletes,
-		)
-
-		reporter <- progress
-
-		if completion.NeedBytes == 0 {
-			return nil
+		if len(events) > 0 {
```
rberrelleza: If the sync was completed before this code is executed, do we still get the completion events?

pkg/syncthing/syncthing.go
```diff
@@ -329,64 +332,44 @@ func (s *Syncthing) WaitForScanning(ctx context.Context, wg *sync.WaitGroup, dev
...
+		"events":  "FolderCompletion",
+		"device":  DefaultRemoteDeviceID,
+		"folder":  fmt.Sprintf("okteto-%s", dev.Name),
+		"timeout": "0",
+		"since":   "0",
+	}
+	var since int
+	var completion float64
+	log.Info("waiting for synchronization to complete...", completion)
+	for completion != 100 {
```
rberrelleza: At some point we've seen the syncthing return a 100% completion and then go back to 0 and then start increasing (it was a problem that the companion app ran into). I believe we were using this same api. Have you seen anything like it?

pchico83: that was because we were calling completion twice, before and after overwriting local changes.

pkg/syncthing/syncthing.go
pchico83: Yes, we start querying from the begin of the history of the local instance:
https://github.com/okteto/okteto/pull/427/files#diff-989b94ab98dccde0ec44b8855ebd5267R340

chelsealadd349:
👌

PR #432, opened 2019-09-27
add --wait to restart command

### LINKED ISSUE #431: okteto restart should wait until the services are up and running
When I run `okteto restart`, I always have to run `kubectl get pod` a few times to check if the restarted services are ready to be consumed.

`okteto restart --wait` to block until all the services have been restarted and are healthy.

rberrelleza:
Fixes #431

### DIFF
--- a/cmd/restart.go
+++ b/cmd/restart.go
@@ -52,11 +52,13 @@ func executeRestart(dev *model.Dev) error {
 		dev.Namespace = namespace
 	}
 
-	log.Information("Restarting pods")
+	spinner := newSpinner("Restarting your Okteto environment...")
+	spinner.start()
+	defer spinner.stop()
+
 	if err := pods.Restart(dev, client); err != nil {
 		return err
 	}
 
-	log.Information("All pods restarted")
 	return nil
 }
--- a/cmd/restart.go
+++ b/cmd/restart.go
@@ -52,9 +52,11 @@ func executeRestart(dev *model.Dev) error {
 		dev.Namespace = namespace
 	}
 
+	log.Information("Restarting pods")
 	if err := pods.Restart(dev, client); err != nil {
 		return err
 	}
 
+	log.Information("All pods restarted")
 	return nil
 }
--- a/cmd/restart.go
+++ b/cmd/restart.go
@@ -43,6 +43,7 @@ func Restart() *cobra.Command {
 }
 
 func executeRestart(dev *model.Dev) error {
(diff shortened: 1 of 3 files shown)


pkg/k8s/pods/pod.go
```diff
@@ -178,5 +178,47 @@ func Restart(dev *model.Dev, c *kubernetes.Clientset) error {
...
-	return nil
+	if !wait {
+		return nil
+	}
+
+	return waitUntilRunning(dev.Namespace, fmt.Sprintf("%s=%s", OktetoDetachedDevLabel, dev.Name), c)
+}
+
+func waitUntilRunning(namespace, selector string, c *kubernetes.Clientset) error {
+	t := time.NewTicker(1 * time.Second)
```
pchico83: I would add logs telling we are waiting for pods to be ready and a final log telling everything is up and running. Maybe a log for each pod that reaches the running state too.

cmd/restart.go
```diff
@@ -38,11 +39,12 @@ func Restart() *cobra.Command {
 
 	cmd.Flags().StringVarP(&devPath, "file", "f", defaultManifest, "path to the manifest file")
 	cmd.Flags().StringVarP(&namespace, "namespace", "n", "", "namespace where the exec command is executed")
+	cmd.Flags().BoolVarP(&wait, "wait", "w", false, "wait until pods are ready")
```
pchico83: What about doing this the default (and maybe the unique) option?

pkg/k8s/pods/pod.go
rberrelleza: I'm not a big fan of showing too many logs. I like linux'  no news is good news approach. How about we just show a log saying 'waiting for pods to restart', and the success one? @rlamana what do you think?

cmd/restart.go
rberrelleza: Im 50/50 on this one. Do you know if anyone other than us is using this command?

pkg/k8s/pods/pod.go
pchico83: sounds good

rlamana: I also think that way. "Waiting > Success" sounds good to me.

rberrelleza:
Changed it to use a spinner like we do in up and down instead. Thoughts @pchico83 @rlamana ?

restart

rlamana:
@rberrelleza looks really good! and it is consistent with the rest of outputs.

PR #438, opened 2019-10-01
Generate a context per server/namespace

rberrelleza:
Generate a context per server/namespace configuration, so that it's easier to switch namespaces with other kubernetes tools (like the VS Code context switcher).

I also want to remove the redundant postfixes, so that all names are shorter and easier to read. I think it's ok to leave the old ones, but we could have a migration code if you fix strongly about it.

### DIFF
--- a/pkg/okteto/client.go
+++ b/pkg/okteto/client.go
@@ -69,15 +69,14 @@ func isConnectionError(s string) bool {
 }
 
 //SetKubeConfig updates a kubeconfig file with okteto cluster credentials
-func SetKubeConfig(cred *Credential, kubeConfigPath, namespace, userName, host string) error {
-	clusterName := fmt.Sprintf("%s", host)
+func SetKubeConfig(cred *Credential, kubeConfigPath, namespace, userName, clusterName string) error {
 	var contextName = ""
 	if len(namespace) == 0 {
 		// don't include namespace for the personal namespace
-		contextName = fmt.Sprintf(host)
+		contextName = fmt.Sprintf(clusterName)
 		namespace = cred.Namespace
 	} else {
-		contextName = fmt.Sprintf("%s-%s", host, namespace)
+		contextName = fmt.Sprintf("%s-%s", clusterName, namespace)
 	}
 
 	var cfg *clientcmdapi.Config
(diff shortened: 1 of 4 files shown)


pkg/okteto/client.go
```diff
@@ -68,28 +68,27 @@ func isConnectionError(s string) bool {
...
-	userName := GetUserID()
-	contextName := fmt.Sprintf("%s-context", oktetoURL)
-
-	var cfg *clientcmdapi.Config
-	cred, err := GetCredentials(namespace)
-	if err != nil {
-		return err
+//SetKubeConfig updates a kubeconfig file with okteto cluster credentials
+func SetKubeConfig(cred *Credential, kubeConfigPath, namespace, userName, clusterName string) error {
+	var contextName = ""
```
golangcibot: ineffectual assignment to `contextName` (from `ineffassign`)

pkg/okteto/client.go
```diff
@@ -68,28 +68,27 @@ func isConnectionError(s string) bool {
...
+	if len(namespace) == 0 {
+		// don't include namespace for the personal namespace
+		contextName = fmt.Sprintf(clusterName)
```
golangcibot: SA1006: printf-style function with dynamic format string and no further arguments should use print-style function instead (from `staticcheck`)

pkg/okteto/client.go
```diff
@@ -68,28 +68,28 @@ func isConnectionError(s string) bool {
...
-	contextName := fmt.Sprintf("%s-context", oktetoURL)
-
-	var cfg *clientcmdapi.Config
-	cred, err := GetCredentials(namespace)
-	if err != nil {
-		return err
+//SetKubeConfig updates a kubeconfig file with okteto cluster credentials
+func SetKubeConfig(cred *Credential, kubeConfigPath, namespace, userName, host string) error {
+	clusterName := fmt.Sprintf("%s", host)
+	var contextName = ""
```
golangcibot: ineffectual assignment to `contextName` (from `ineffassign`)

pkg/okteto/client.go
```diff
@@ -68,28 +68,28 @@ func isConnectionError(s string) bool {
...
+	if len(namespace) == 0 {
+		// don't include namespace for the personal namespace
+		contextName = fmt.Sprintf(host)
```
golangcibot: SA1006: printf-style function with dynamic format string and no further arguments should use print-style function instead (from `staticcheck`)

rberrelleza:
> You need to update the UI too

opened https://github.com/okteto/app/pull/482/files for that

PR #443, opened 2019-10-04
Init command for gradle and maven

pchico83:


cmd/init.go
```diff
@@ -159,9 +160,21 @@ func getDevelopmentEnvironment(language string) (*model.Dev, string, error) {
...
 func getDeploymentName(name string) string {
-	deploymentName := filepath.Base(name)
-	deploymentName = strings.ToLower(deploymentName)
-	deploymentName = model.ValidKubeNameRegex.ReplaceAllString(deploymentName, "-")
-	log.Infof("deployment name: %s", deploymentName)
+	var deploymentName string
+	defaultDeploymentName := filepath.Base(name)
+	defaultDeploymentName = strings.ToLower(defaultDeploymentName)
+	defaultDeploymentName = model.ValidKubeNameRegex.ReplaceAllString(defaultDeploymentName, "-")
+	log.Debugf("default dployment name: %s", defaultDeploymentName)
```
rberrelleza:
```suggestion
	log.Debugf("default deployment name: %s", defaultDeploymentName)
```

cmd/init.go
```diff
@@ -159,9 +160,21 @@ func getDevelopmentEnvironment(language string) (*model.Dev, string, error) {
 }
 
 func getDeploymentName(name string) string {
```
rberrelleza: It's better to split this into two functions. One for to generate the default name, and another to ask the user.

cmd/init.go
```diff
@@ -159,9 +160,21 @@ func getDevelopmentEnvironment(language string) (*model.Dev, string, error) {
...
+	var deploymentName string
+	defaultDeploymentName := filepath.Base(name)
+	defaultDeploymentName = strings.ToLower(defaultDeploymentName)
+	defaultDeploymentName = model.ValidKubeNameRegex.ReplaceAllString(defaultDeploymentName, "-")
+	log.Debugf("default dployment name: %s", defaultDeploymentName)
+
+	fmt.Printf("Which deployment do you want to use for your development environment? [%s]: ", defaultDeploymentName)
+	_, err := fmt.Scanln(&deploymentName)
+	if err != nil {
+		log.Debugf("Scanln failed to read deployment name: %s", err)
```
rberrelleza: we should fail at this point, since something went wrong.

cmd/init.go
```diff
@@ -159,9 +160,21 @@ func getDevelopmentEnvironment(language string) (*model.Dev, string, error) {
...
-	deploymentName = strings.ToLower(deploymentName)
-	deploymentName = model.ValidKubeNameRegex.ReplaceAllString(deploymentName, "-")
-	log.Infof("deployment name: %s", deploymentName)
+	var deploymentName string
+	defaultDeploymentName := filepath.Base(name)
+	defaultDeploymentName = strings.ToLower(defaultDeploymentName)
+	defaultDeploymentName = model.ValidKubeNameRegex.ReplaceAllString(defaultDeploymentName, "-")
+	log.Debugf("default dployment name: %s", defaultDeploymentName)
+
+	fmt.Printf("Which deployment do you want to use for your development environment? [%s]: ", defaultDeploymentName)
```
rberrelleza: This is very verbose, and I dont think it's clear. I'd just go with `Deployment name [%s]:`.

pkg/linguist/dev.go
```diff
@@ -106,7 +159,10 @@ func normalizeLanguage(language string) string {
 	case "python":
 		return python
 	case "java":
```
rberrelleza: I think the linguist library already returns gradle and/or maven, can you try it out?

rberrelleza:
Split the deployment selection change into a separate PR so we can have 1 PR for feature :)

PR #447, opened 2019-10-07
option to pass the language instead of detecting it

rberrelleza:


pkg/linguist/dev.go
```diff
@@ -20,10 +19,12 @@ type languageDefault struct {
 
 const (
 	javascript = "javascript"
+	node       = "node"
```
golangcibot: `node` is unused (from `varcheck`)

PR #454, opened 2019-10-07
support for csharp

rberrelleza:


pkg/linguist/dev.go
```diff
@@ -110,6 +111,24 @@ func init() {
 		command: []string{"bash"},
 	}
 
+	languageDefaults[csharp] = languageDefault{
+		image:   "mcr.microsoft.com/dotnet/core/sdk",
+		path:    "/usr/src/app",
```
pchico83: can we use the default?

PR #457, opened 2019-10-07
Activate OKTETO_CLIENTSIDE_TRANSLATION in e2e tests

pchico83:


Makefile
```diff
@@ -40,6 +40,7 @@ test:
 
 .PHONY: integration
 integration:
+	 export OKTETO_CLIENTSIDE_TRANSLATION=true
```
rberrelleza: I think we should put this on .circleci/config.yml line 44 instead

PR #471, opened 2019-10-08
Mount syncthing folder as a volume for persistency and to inherit the…

pchico83:


rberrelleza:
will the webhook do this as well?

pchico83:
@rberrelleza yes, everything that changes the translation rule content is done by the webhook too, since it applies the same translation algorithm.
The webhook must be updated when modifying the translation logic itself.

PR #472, opened 2019-10-08
Show proper error when security context is not set

pchico83:
Error management is not a little simpler since there are fewer moving pieces.
This is how it looks like if the user missed to set the security context:

Screenshot 2019-10-08 at 11 53 38

### DIFF
--- a/cmd/up.go
+++ b/cmd/up.go
@@ -401,7 +401,7 @@ func (up *UpContext) startLocalSyncthing() error {
 	if err := up.Sy.WaitForPing(up.Context, up.WG, false); err != nil {
 		return errors.UserError{
 			E:    fmt.Errorf("Failed to connect to the synchronization service"),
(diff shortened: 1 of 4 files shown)


cmd/up.go
```diff
@@ -398,6 +398,12 @@ func (up *UpContext) startLocalSyncthing() error {
 	if err := up.Sy.WaitForPing(up.Context, up.WG, true); err != nil {
 		return err
 	}
+	if err := up.Sy.WaitForPing(up.Context, up.WG, false); err != nil {
+		return errors.UserError{
+			E:    fmt.Errorf("Failed to connect to the synchonization service"),
+			Hint: fmt.Sprintf("This is probably because your development image is not root.\n    Please, add securityContext.runAsUser and securityContext.fsGroup to your okteto manifest.\n    Follow this link for an example: https://github.com/okteto/java-gradle-getting-started/blob/master/okteto.yml"),
```
rberrelleza: I'm working on the docs, let me do this and then we use the docs link.

```suggestion
			Hint: fmt.Sprintf("This is probably because you're using a non-root container for your development environment. \n    If you are using a non-root container, and the runAsUser and runAsGroup values are not specified in your Kubernetes manifest, you need to set these values in your Okteto manifest.\n    Follow this link for more information: https://okteto.com/docs/reference/manifest/index.html#securityContext-object-optional"),
```

PR #475, opened 2019-10-09
add ssh config when using --remote

rberrelleza:
With this, you can do `ssh ${name}.okteto` without having to set keys, hostnames or specify the port


pkg/ssh/ssh.go
```diff
@@ -0,0 +1,129 @@
...
+	"os"
+	"path/filepath"
+	"strconv"
+
+	"github.com/havoc-io/ssh_config"
+	"github.com/okteto/okteto/pkg/config"
+)
+
+const (
+	sshConfigFile = ".ssh/config"
```
golangcibot: `sshConfigFile` is unused (from `deadcode`)

Issue #332, opened 2019-08-01
CLI doesn't detect if the replicaset is not healthy


rberrelleza:
**Describe the bug**
Ran into a situation where, due to GKE issues, the replica set was stuck in pod-initializing, blocking everything. If would be good if the CLI could detect this and inform the user to do `okteto down -v` or something similar. 

Kubectl lots:
```
Name:               okteto-website-0
Namespace:          website-rlamana
Priority:           0
PriorityClassName:  <none>
Node:               gke-cloud-okteto-com-default-pool-2f874076-cdm5/10.128.15.231
Start Time:         Thu, 01 Aug 2019 17:48:16 +0200
Labels:             controller-revision-hash=okteto-website-777669cf7
                    statefulset.kubernetes.io/pod-name=okteto-website-0
                    syncthing.okteto.com=website
Annotations:        cni.projectcalico.org/podIP: 10.32.2.3/32
                    container.apparmor.security.beta.kubernetes.io/okteto: runtime/default
                    container.apparmor.security.beta.kubernetes.io/okteto-init: runtime/default
                    kubernetes.io/psp: cloud-okteto-enterprise-restrictive
                    seccomp.security.alpha.kubernetes.io/pod: docker/default
Status:             Pending
IP:
Controlled By:      StatefulSet/okteto-website
Init Containers:
  okteto-init:
    Container ID:  docker://d5d66a5f48f7cbfd95fb39a2c1435770246f7c71dda3018c1caa2b344d458791
    Image:         okteto/node:10
    Image ID:      docker-pullable://okteto/node@sha256:078916b624ef8be7f619959f32ab44a2a223e6046a6f6de81060563844af1bbe
    Port:          <none>
    Host Port:     <none>
    Command:
      sh
      -c
      (ls -A /okteto/init | grep -v lost+found || cp -Rf /usr/src/app/* /okteto/init); touch /okteto/init/okteto.yml
    State:          Terminated
      Reason:       Completed
      Exit Code:    0
      Started:      Thu, 01 Aug 2019 17:48:19 +0200
      Finished:     Thu, 01 Aug 2019 17:48:19 +0200
    Ready:          True
    Restart Count:  0
    Limits:
      cpu:     50m
      memory:  16Mi
    Requests:
      cpu:        50m
      memory:     16Mi
    Environment:  <none>
    Mounts:
      /okteto/init from pvc-0 (rw)
      /var/run/secrets/kubernetes.io/serviceaccount from default-token-hdklp (ro)
[...]

PR #479, opened 2019-10-10
Stream pod events to the spinner if they last for more than 1s

### LINKED ISSUE #430: Clearly log when `okteto up` fails due to scheduling constraints
`okteto up` fails with the following:

```
okteto up
Deployment [deployment] doesn't exist in namespace [namespace]. Do you want to create a new one? [y/n]: y
⠼ Provisioning your persistent volume...
x  kubernetes is taking too long to create the cloud native environment. Please check for errors and try again
```

Looking at the logs, the issue is related to scheduling issues with the syncthing pod  due to CPU constraints. The log should  be clear about it and propose actions.

pchico83:
Closes #430 #332

### DIFF
--- a/cmd/exec.go
+++ b/cmd/exec.go
@@ -24,7 +24,7 @@ func Exec() *cobra.Command {
 
 	cmd := &cobra.Command{
 		Use:   "exec <command>",
-		Short: "Execute a command in your Okteto Environment",
+		Short: "Execute a command in your development environment",
 		RunE: func(cmd *cobra.Command, args []string) error {
 			ctx, cancel := context.WithCancel(context.Background())
 			defer cancel()
@@ -41,7 +41,7 @@ func Exec() *cobra.Command {
 
 			if errors.IsNotFound(err) {
 				return errors.UserError{
-					E:    fmt.Errorf("Okteto environment not found in namespace %s", dev.Namespace),
+					E:    fmt.Errorf("Development environment not found in namespace %s", dev.Namespace),
 					Hint: "Run `okteto up` to launch it or use `okteto namespace` to select the correct namespace and try again",
 				}
 			}
--- a/cmd/up.go
+++ b/cmd/up.go
@@ -65,7 +65,7 @@ func Up() *cobra.Command {
 	var remote int
 	cmd := &cobra.Command{
 		Use:   "up",
-		Short: "Activates your Okteto Environment",
+		Short: "Activates your development environment",
 		RunE: func(cmd *cobra.Command, args []string) error {
 			log.Debug("starting up command")
 			u := upgradeAvailable()
@@ -185,7 +185,7 @@ func (up *UpContext) Activate() {
 		if !activated {
 			return
 		}
-		log.Success("Okteto Environment activated")
+		log.Success("Development environment activated")
 
 		err = up.sync()
 		if err != nil {
@@ -215,7 +215,7 @@ func (up *UpContext) Activate() {
 
 		if prevError != nil {
 			if prevError == errors.ErrLostConnection || (prevError == errors.ErrCommandFailed && !pods.Exists(up.Pod, up.Dev.Namespace, up.Client)) {
-				log.Yellow("\nConnection lost to your Okteto Environment, reconnecting...\n")
+				log.Yellow("\nConnection lost to your development environment, reconnecting...\n")
 				up.shutdown()
 				continue
 			}
@@ -280,7 +280,7 @@ func (up *UpContext) WaitUntilExitOrInterrupt() error {
 }
 
 func (up *UpContext) devMode(isRetry bool, d *appsv1.Deployment, create bool) (bool, error) {
-	spinner := newSpinner("Activating your Okteto Environment...")
+	spinner := newSpinner("Activating your development environment...")
 	up.updateStateFile(activating)
 	spinner.start()
 	defer spinner.stop()
@@ -314,7 +314,7 @@ func (up *UpContext) devMode(isRetry bool, d *appsv1.Deployment, create bool) (b
 	}
 
 	if isRetry && !deployments.IsDevModeOn(d) {
-		log.Information("Okteto Environment has been deactivated")
+		log.Information("Development environment has been deactivated")
 		return false, nil
 	}
 
@@ -356,12 +356,29 @@ func (up *UpContext) devMode(isRetry bool, d *appsv1.Deployment, create bool) (b
 		}
 	}
 
-	p, err := pods.GetDevPod(up.Context, up.Dev, up.Client, create)
+	pod, err := pods.GetDevPod(up.Context, up.Dev, up.Client, create)
 	if err != nil {
 		return false, err
 	}
 
-	up.Pod = p.Name
+	reporter := make(chan string)
+	defer close(reporter)
+	go func() {
+		message := "Attaching persistent volume"
+		for {
+			spinner.update(fmt.Sprintf("%s...", message))
+			message = <-reporter
+			if message == "" {
+				return
+			}
+		}
+	}()
+	pod, err = pods.MonitorDevPod(up.Context, up.Dev, pod, up.Client, reporter)
+	if err != nil {
+		return false, err
+	}
+
+	up.Pod = pod.Name
 	go up.cleanCommand()
 
 	up.Forwarder = forward.NewPortForwardManager(up.Context, up.RestConfig, up.Client, up.ErrChan)
--- a/pkg/k8s/pods/pod.go
+++ b/pkg/k8s/pods/pod.go
@@ -13,13 +13,14 @@ import (
 	"github.com/okteto/okteto/pkg/model"
 	appsv1 "k8s.io/api/apps/v1"
 	apiv1 "k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 )
 
 const (
 	deploymentRevisionAnnotation = "deployment.kubernetes.io/revision"
-	maxRetriesPodRunning         = 1500 //5min pod is running
+	maxRetriesPodRunning         = 300 //1min pod is created
 )
 
 var (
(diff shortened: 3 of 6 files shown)


cmd/up.go
```diff
@@ -356,7 +356,30 @@ func (up *UpContext) devMode(isRetry bool, d *appsv1.Deployment, create bool) (b
 		}
 	}
 
-	p, err := pods.GetDevPod(up.Context, up.Dev, up.Client, create)
+	reporter := make(chan string)
+	go func() {
+		long := false
+		message := ""
+		ticker := time.NewTicker(1000 * time.Millisecond)
```
```suggestion
		ticker := time.NewTicker(1 * time.Second)
```

cmd/up.go
```diff
@@ -356,7 +356,30 @@ func (up *UpContext) devMode(isRetry bool, d *appsv1.Deployment, create bool) (b
...
+		for {
+			select {
+			case <-ticker.C:
+				if long && message != "" {
```
rberrelleza: I don't understand this part of the loop. Why do we log the message that we received in the other case here?

cmd/up.go
```diff
@@ -356,7 +356,30 @@ func (up *UpContext) devMode(isRetry bool, d *appsv1.Deployment, create bool) (b
...
+					spinner.update(fmt.Sprintf("%s...", message))
+				}
+				long = true
+			case message = <-reporter:
+				if message == "" {
+					return
+				}
+				spinner.update("Activating your development environment...")
```
rberrelleza: Why do we wait until getting a message to show the  `Activating...` message?

pkg/k8s/pods/pod.go
```diff
@@ -76,19 +78,58 @@ func loopGetDevPod(dev *model.Dev, c *kubernetes.Clientset, waitUntilDeployed bo
...
+	watchPodEvents, err := c.CoreV1().Events(dev.Namespace).Watch(opts)
+	if err != nil {
+		return nil, err
+	}
+	for {
+		select {
+		case event := <-watchPod.ResultChan():
+			pod, ok := event.Object.(*v1.Pod)
+			if !ok {
+				return nil, nil
```
rberrelleza: If we can't parse the object what happens when we return from the call? Should we send an error, or at least log what happened?

pkg/k8s/pods/pod.go
```diff
@@ -76,19 +78,58 @@ func loopGetDevPod(dev *model.Dev, c *kubernetes.Clientset, waitUntilDeployed bo
 		return nil, err
 	}
 
-	log.Infof("pod %s with revison %s is progressing", pod.Name, d.Annotations[deploymentRevisionAnnotation])
-	if pod.Status.Phase == apiv1.PodRunning {
-		log.Infof("pod %s with revison %s is running", pod.Name, d.Annotations[deploymentRevisionAnnotation])
-		return pod, nil
+	opts := metav1.ListOptions{
```
rberrelleza: If `pod.Name` is unique, do we also need the okteto label?

pkg/k8s/pods/pod.go
```diff
@@ -76,19 +78,58 @@ func loopGetDevPod(dev *model.Dev, c *kubernetes.Clientset, waitUntilDeployed bo
...
+		select {
+		case event := <-watchPod.ResultChan():
+			pod, ok := event.Object.(*v1.Pod)
+			if !ok {
+				return nil, nil
 			}
+			if pod.Status.Phase == apiv1.PodRunning {
+				return pod, nil
+			}
+			if pod.DeletionTimestamp != nil {
```
rberrelleza: This would mean that the pod is in process of being deleted. What happens in this case? The caller of GetDevPod will most likely get a nullref no? (since there's no error returned)

cmd/up.go
```diff
@@ -356,7 +356,30 @@ func (up *UpContext) devMode(isRetry bool, d *appsv1.Deployment, create bool) (b
...
+		message := ""
+		ticker := time.NewTicker(1000 * time.Millisecond)
+		for {
+			select {
+			case <-ticker.C:
+				if long && message != "" {
+					spinner.update(fmt.Sprintf("%s...", message))
+				}
+				long = true
+			case message = <-reporter:
```
rberrelleza: Since we are waiting for a second, a lot of messages could come through, and we are not going to log them all.  Would it make sense to discard all the messages in the channel except for the last one when assigning to `message`?

cmd/up.go
```diff
@@ -356,7 +356,30 @@ func (up *UpContext) devMode(isRetry bool, d *appsv1.Deployment, create bool) (b
...
+				if message == "" {
+					return
+				}
+				spinner.update("Activating your development environment...")
+				long = false
+			case <-up.Context.Done():
```
rberrelleza: Will this loop also exit when we close the `reporter` channel?

cmd/up.go
```diff
@@ -356,7 +356,30 @@ func (up *UpContext) devMode(isRetry bool, d *appsv1.Deployment, create bool) (b
...
 
-	p, err := pods.GetDevPod(up.Context, up.Dev, up.Client, create)
+	reporter := make(chan string)
+	go func() {
+		long := false
+		message := ""
+		ticker := time.NewTicker(1000 * time.Millisecond)
+		for {
+			select {
+			case <-ticker.C:
```
rberrelleza: Do we need both the ticker and the message case? Maybe simplify it by having a single case that receives the message and then waits a second before checking again?

cmd/up.go
rberrelleza: I think this loop should depend on the lifecycle of `reporter`, and don't care about `up.Context.Done`. Instead, we close the reporter when done, and that can cause this loop to be closed.

cmd/up.go
```diff
@@ -356,7 +356,30 @@ func (up *UpContext) devMode(isRetry bool, d *appsv1.Deployment, create bool) (b
 		}
 	}
 
-	p, err := pods.GetDevPod(up.Context, up.Dev, up.Client, create)
+	reporter := make(chan string)
```
```suggestion
	reporter := make(chan string)
	defer close(reporter)
```

cmd/exec.go
```diff
@@ -72,7 +72,7 @@ func executeExec(ctx context.Context, dev *model.Dev, args []string) error {
 		dev.Namespace = namespace
 	}
 
-	p, err := pods.GetDevPod(ctx, dev, client, false)
+	p, err := pods.GetDevPod(ctx, dev, client, false, nil)
```
rberrelleza: I think calling this `GetDevPod` is confusing, since it calls and monitors. Maybe we can have two functions, one to get the pod and another to start the watcher?

pkg/k8s/pods/pod.go
```diff
@@ -27,19 +28,20 @@ const (
...
 
 // GetDevPod returns the dev pod for a deployment
-func GetDevPod(ctx context.Context, dev *model.Dev, c *kubernetes.Clientset, waitUntilDeployed bool) (*apiv1.Pod, error) {
+func GetDevPod(ctx context.Context, dev *model.Dev, c *kubernetes.Clientset, waitUntilDeployed bool, reporter chan string) (*apiv1.Pod, error) {
+	defer close(reporter)
 	tries := 0
 	ticker := time.NewTicker(200 * time.Millisecond)
 	for tries < maxRetriesPodRunning {
-		pod, err := loopGetDevPod(dev, c, waitUntilDeployed)
+		pod, err := loopGetDevPod(ctx, dev, c, waitUntilDeployed, reporter)
```
rberrelleza: what about the `nil, nil` return case?

pkg/k8s/pods/pod.go
```diff
@@ -27,19 +28,20 @@ const (
...
 	devTerminationGracePeriodSeconds int64
 )
 
 // GetDevPod returns the dev pod for a deployment
-func GetDevPod(ctx context.Context, dev *model.Dev, c *kubernetes.Clientset, waitUntilDeployed bool) (*apiv1.Pod, error) {
+func GetDevPod(ctx context.Context, dev *model.Dev, c *kubernetes.Clientset, waitUntilDeployed bool, reporter chan string) (*apiv1.Pod, error) {
+	defer close(reporter)
 	tries := 0
 	ticker := time.NewTicker(200 * time.Millisecond)
 	for tries < maxRetriesPodRunning {
```
rberrelleza: if we have a watcher pattern in `loopGetDevPod`, I don't think we should have the retries. This would complicate consuming the `reporter`.  Could we simplify everything on a single model?  Get the pod, watch until ready, return?

pkg/k8s/pods/pod.go
```diff
@@ -76,19 +78,58 @@ func loopGetDevPod(dev *model.Dev, c *kubernetes.Clientset, waitUntilDeployed bo
...
+			if pod.DeletionTimestamp != nil {
+				return nil, nil
+			}
+		case event := <-watchPodEvents.ResultChan():
+			if reporter == nil {
+				continue
+			}
+			e, ok := event.Object.(*v1.Event)
+			if !ok {
+				return nil, nil
```
rberrelleza: I think we should log and continue.

pkg/k8s/pods/pod.go
```diff
@@ -76,19 +78,58 @@ func loopGetDevPod(dev *model.Dev, c *kubernetes.Clientset, waitUntilDeployed bo
...
-	log.Infof("pod %s with revison %s is progressing", pod.Name, d.Annotations[deploymentRevisionAnnotation])
-	if pod.Status.Phase == apiv1.PodRunning {
-		log.Infof("pod %s with revison %s is running", pod.Name, d.Annotations[deploymentRevisionAnnotation])
-		return pod, nil
+	opts := metav1.ListOptions{
+		LabelSelector: fmt.Sprintf("%s=%s", OktetoInteractiveDevLabel, dev.Name),
+		FieldSelector: fmt.Sprintf("metadata.name=%s", pod.Name),
+	}
+
+	watchPod, err := c.CoreV1().Pods(dev.Namespace).Watch(opts)
```
rberrelleza: Do we get events even if they were created before the watcher? (e.g. what if the pod is already running by the time we run this code)

PR #489, opened 2019-10-16
Update state file based on events

pchico83:


cmd/up.go
```diff
@@ -312,8 +312,6 @@ func (up *UpContext) devMode(isRetry bool, d *appsv1.Deployment, create bool) (b
 		return false, nil
 	}
 
-	up.updateStateFile(starting)
```
rberrelleza: leave the starting, since a few things happen before the first state change, it helps give a "default state"

PR #492, opened 2019-10-17
exit process when dev mode is turned off

rberrelleza:
There's a bug where, if I run `okteto down`, the `okteto up` hangs forever. 

I'm changing the behavior slightly so that it does exit, and to log it as an error, since it is an unexpected effect.

### DIFF
--- a/cmd/up.go
+++ b/cmd/up.go
@@ -177,14 +177,12 @@ func (up *UpContext) Activate() {
 			return
 		}
 
-		activated, err := up.devMode(retry, d, create)
+		err = up.devMode(retry, d, create)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


cmd/up.go
```diff
@@ -177,14 +177,12 @@ func (up *UpContext) Activate() {
...
 		}
 
-		activated, err := up.devMode(retry, d, create)
+		err = up.devMode(retry, d, create)
 		if err != nil {
 			up.Exit <- err
 			return
 		}
-		if !activated {
-			return
```
rberrelleza: This return made the process hang, since we have a loop that only exits if there's an error, or context.Done() returns

PR #495, opened 2019-10-21
Update golang init command to follow java approach

pchico83:
Update golang init command to follow java approach

### DIFF
--- a/pkg/linguist/dev.go
+++ b/pkg/linguist/dev.go
@@ -5,6 +5,7 @@ import (
 	"strings"
 
 	"github.com/okteto/okteto/pkg/model"
+	apiv1 "k8s.io/api/core/v1"
 )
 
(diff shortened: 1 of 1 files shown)


pkg/linguist/dev.go
```diff
@@ -76,22 +90,12 @@ func init() {
...
-			RunAsGroup: &user1000,
-			FSGroup:    &user1000,
-		},
 	}
 
 	languageDefaults[maven] = languageDefault{
 		image:   "okteto/maven:latest",
+		path:    "/okteto",
 		command: []string{"bash"},
-		environment: []model.EnvVar{
```
rberrelleza: why are we removing this?

pchico83: they don't work. we switched to maven/gradle config files

PR #496, opened 2019-10-22
Add support for pod annotations

AnesBenmerzoug:
## Proposed changes
This PR adds annotations for the deployment Pod.
This is useful in conjunction with [kube2iam](https://github.com/jtblin/kube2iam) in order to assign a role to the okteto pod.

This makes the using the following development manifest possible:

```yaml
name: okteto
image: python:3.5
annotations:
  key1: value1
  key2: value2
command:
- bash
workdir: /okteto
```

```bash
$ okteto up

 ✓  Development environment activated
 ✓  Files synchronized
    Namespace: default
    Name:      okteto
    Forward:   8080 -> 8080
               2345 -> 2345
```

```bash
$ kubectl get pod okteto-7b9fc7c989-k4jkf -o jsonpath='{.metadata.annotations}{"\n"}'
map[key1:value1 key2:value2]
```

### DIFF
--- a/pkg/k8s/deployments/crud.go
+++ b/pkg/k8s/deployments/crud.go
@@ -87,6 +87,7 @@ func GetTranslations(dev *model.Dev, d *appsv1.Deployment, c *kubernetes.Clients
 			Name:        dev.Name,
 			Version:     model.TranslationVersion,
 			Deployment:  d,
+			Annotations: dev.Annotations,
 			Replicas:    *d.Spec.Replicas,
 			Rules:       []*model.TranslationRule{rule},
 		}
@@ -108,6 +109,7 @@ func GetTranslations(dev *model.Dev, d *appsv1.Deployment, c *kubernetes.Clients
 				Interactive: false,
 				Version:     model.TranslationVersion,
 				Deployment:  d,
+				Annotations: dev.Annotations,
 				Replicas:    *d.Spec.Replicas,
 				Rules:       []*model.TranslationRule{rule},
 			}
--- a/pkg/k8s/deployments/crud.go
+++ b/pkg/k8s/deployments/crud.go
@@ -178,6 +178,7 @@ func DevModeOff(d *appsv1.Deployment, c *kubernetes.Clientset) error {
 		delete(annotations, oktetoVersionAnnotation)
 		d.GetObjectMeta().SetAnnotations(annotations)
 		annotations = d.Spec.Template.GetObjectMeta().GetAnnotations()
+		deleteUserAnnotations(annotations)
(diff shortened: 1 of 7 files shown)


pkg/model/dev.go
```diff
@@ -430,6 +437,7 @@ func (dev *Dev) GevSandbox() *appsv1.Deployment {
 					Labels: map[string]string{
 						"app": dev.Name,
 					},
+					Annotations: dev.Annotations,
```
pchico83: This code is only executed when `okteto up` creates a new deployment for scratch.
If you also want to add the annotations when `okteto up` modifies an existing deployment, you should assign the annotations in the `translate` logic, somewhere over here:
https://github.com/okteto/okteto/blob/master/pkg/k8s/deployments/translate.go#L73

AnesBenmerzoug: Thanks for pointing that out.
I wasn't too sure where to put it, but now it makes sense.

pchico83:
I have just realized that we also need to remove the added annotations on `okteto down`:
https://github.com/okteto/okteto/blob/master/pkg/k8s/deployments/crud.go#L179

AnesBenmerzoug:
I will take care of that.

pkg/k8s/deployments/crud.go
```diff
@@ -176,6 +178,7 @@ func DevModeOff(d *appsv1.Deployment, c *kubernetes.Clientset) error {
 		delete(annotations, oktetoVersionAnnotation)
 		d.GetObjectMeta().SetAnnotations(annotations)
 		annotations = d.Spec.Template.GetObjectMeta().GetAnnotations()
+		deleteUserAnnotations(annotations)
```
golangcibot: Error return value of `deleteUserAnnotations` is not checked (from `errcheck`)

pkg/k8s/deployments/utils_test.go
```diff
@@ -0,0 +1,55 @@
...
+		t.Fatal(err)
+	}
+	d := dev.GevSandbox()
+	dev.DevPath = "okteto.yml"
+	translations, err := GetTranslations(dev, d, nil)
+	if err != nil {
+		t.Fatal(err)
+	}
+	tr1 := translations[d.Name]
+	setTranslationAsAnnotation(d.GetObjectMeta(), tr1)
```
golangcibot: Error return value of `setTranslationAsAnnotation` is not checked (from `errcheck`)

pchico83:
@AnesBenmerzoug thanks for contributing to okteto!
We are packaging a release just now, so your changes will be available for everyone today.

PR #523, opened 2019-10-29
Since syncthing runs in the dev container itself, the dev container n…

pchico83:
…eeds to sync the whole local folder

### DIFF
--- a/pkg/model/dev.go
+++ b/pkg/model/dev.go
@@ -235,6 +235,10 @@ func (dev *Dev) validate() error {
 		return errBadName
 	}
 
+	if dev.SubPath != "" {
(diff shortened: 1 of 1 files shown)


pkg/model/dev.go
```diff
@@ -235,6 +235,10 @@ func (dev *Dev) validate() error {
 		return errBadName
 	}
 
+	if dev.SubPath != "" {
```
rberrelleza: can we remove it from the dev struct in this case? Or do we still need it under `services `? Also, could you update the manifest reference?

pchico83: Yes, it is still valid for `services`. I will update docs before the release is done

PR #524, opened 2019-10-31
add machineID to the token file

rberrelleza:
With this, we can use the same machineID on the vscode extensions and the CLI

### DIFF
--- a/pkg/analytics/analytics.go
+++ b/pkg/analytics/analytics.go
@@ -200,34 +200,27 @@ func isEnabled() bool {
 }
 
 func getTrackID() string {
-	if len(trackID) == 0 {
-		uid := okteto.GetUserID()
-		if len(uid) > 0 {
-			trackID = uid
-		} else {
-			trackID = getMachineID()
-		}
+	uid := okteto.GetUserID()
+	if len(uid) > 0 {
+		return uid
 	}
 
-	return trackID
+	return getMachineID()
 }
 
 func getMachineID() string {
-	if len(machineID) > 0 {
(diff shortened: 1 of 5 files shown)


pkg/okteto/auth.go
```diff
@@ -137,7 +148,31 @@ func GetURL() string {
 }
 
 func saveToken(id, token, url string) error {
-	t := Token{Token: token, URL: url, ID: id}
+	t, err := getToken()
+	if err != nil {
+		log.Debugf("bad token, re-initializing: %s")
```
golangcibot: printf: Debugf format %s reads arg #1, but call has 0 args (from `govet`)

pkg/okteto/auth.go
```diff
@@ -137,7 +148,31 @@ func GetURL() string {
...
+	t.Token = token
+	t.URL = url
+	return save(t)
+}
+
+// SaveMachineID updates the token file with the machineID value
+func SaveMachineID(machineID string) error {
+	t, err := getToken()
+	if err != nil {
+		log.Debugf("bad token, re-initializing: %s")
```
golangcibot: printf: Debugf format %s reads arg #1, but call has 0 args (from `govet`)

pkg/analytics/analytics.go
```diff
@@ -32,6 +32,7 @@ const (
 
 var (
 	mixpanelClient mixpanel.Mixpanel
+	trackID        string
```
golangcibot: var `trackID` is unused (from `unused`)

PR #529, opened 2019-11-04
Increase syncthing timeout

pchico83:


xiaods:
let me testing it. thanks for your help. i will update asap.

xiaods:
this patch can't resolve my issue, based in china network.

PR #536, opened 2019-11-06
Support for "okteto down -v"

### LINKED ISSUE #533: Support for "okteto down -v"
Kubernetes does not apply `fsGroup` to existing folders in a PVC. In order to change `fsGroup` in the okteto manifest, the user needs to run `okteto down` for every active dev environment,  delete the okteto volume, and run `okteto up` again.

`okteto volume -v` removes the folder of the current dev environment in the okteto volume.

pchico83:
Fixes #533

### DIFF
--- a/cmd/down.go
+++ b/cmd/down.go
@@ -48,10 +48,11 @@ func Down() *cobra.Command {
 
 			if rm {
 				if err := cleanVolume(dev); err != nil {
-					analytics.TrackDown(config.VersionString, false)
+					analytics.TrackDownVolumes(config.VersionString, false)
 					return err
 				}
 				log.Success("Persistent volume cleaned")
+				analytics.TrackDownVolumes(config.VersionString, true)
 			}
 
 			log.Println()
@@ -160,7 +161,7 @@ func removeVolumes(dev *model.Dev) error {
 }
 
 func cleanVolume(dev *model.Dev) error {
-	spinner := newSpinner("Cleaning volume...")
+	spinner := newSpinner("Cleaning persistent volume...")
 	spinner.start()
 	defer spinner.stop()
 
--- a/cmd/down.go
+++ b/cmd/down.go
@@ -6,6 +6,7 @@ import (
 	"github.com/okteto/okteto/pkg/errors"
 	k8Client "github.com/okteto/okteto/pkg/k8s/client"
 	"github.com/okteto/okteto/pkg/k8s/deployments"
+	"github.com/okteto/okteto/pkg/k8s/pods"
 	"github.com/okteto/okteto/pkg/k8s/secrets"
 	"github.com/okteto/okteto/pkg/k8s/services"
 	"github.com/okteto/okteto/pkg/k8s/statefulsets"
@@ -21,6 +22,7 @@ import (
 func Down() *cobra.Command {
 	var devPath string
 	var namespace string
+	var rm bool
 
 	cmd := &cobra.Command{
 		Use:   "down",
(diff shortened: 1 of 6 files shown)


cmd/down.go
```diff
@@ -44,6 +46,14 @@ func Down() *cobra.Command {
 
 			log.Success("Development environment deactivated")
 
+			if rm {
+				if err := cleanVolume(dev); err != nil {
```
rberrelleza: we should create a separate event for this specifically, so we know if the clean fails vs the down operation.

pkg/k8s/pods/pod.go
```diff
@@ -178,6 +178,54 @@ func Restart(dev *model.Dev, c *kubernetes.Clientset) error {
 	return waitUntilRunning(dev.Namespace, fmt.Sprintf("%s=%s", okLabels.DetachedDevLabel, dev.Name), c)
 }
 
+// RunCleanerPod runs a pod to clean the dev environmentt volume
+func RunCleanerPod(dev *model.Dev, c *kubernetes.Clientset) error {
```
rberrelleza: It feels like a `job` would be more suitable for this than the pod no?

pkg/k8s/pods/translate.go
```diff
@@ -0,0 +1,59 @@
...
+)
+
+const (
+	oktetoVolumeName = "okteto"
+)
+
+func translate(dev *model.Dev) *apiv1.Pod {
+	return &apiv1.Pod{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      fmt.Sprintf("volume-cleaner-%s", dev.Name),
```
rberrelleza: if the user runs `okteto down -v` multiple times could this be problematic?

cmd/down.go
```diff
@@ -148,6 +159,34 @@ func removeVolumes(dev *model.Dev) error {
...
+	if dev.Namespace == "" {
+		dev.Namespace = namespace
+	}
+
+	exists, err := volumes.Exists(dev.Namespace, client)
+	if err != nil {
+		return err
+	}
+
+	if !exists {
```
rberrelleza: in which situations it wouldn't exist?

cmd/up.go
```diff
@@ -434,7 +434,7 @@ func (up *UpContext) startLocalSyncthing() error {
 	if err := up.Sy.WaitForPing(up.Context, false); err != nil {
 		return errors.UserError{
 			E:    fmt.Errorf("Failed to connect to the synchronization service"),
-			Hint: fmt.Sprintf("If you are using a non-root container, you need to set the securityContext.runAsUser and securityContext.runAsGroup values in your Okteto manifest (https://okteto.com/docs/reference/manifest/index.html#securityContext-object-optional)."),
+			Hint: fmt.Sprintf("If you are using a non-root container, run 'okteto down -v' and set the fields securityContext.runAsUser, securityContext.runAsGroup and securityContext.fsGroup in your Okteto manifest (https://okteto.com/docs/reference/manifest/index.html#securityContext-object-optional)."),
```
```suggestion
			Hint: fmt.Sprintf("If you are using a non-root container, run 'okteto down -v' to clean the existing files, set the securityContext.runAsUser, securityContext.runAsGroup and securityContext.fsGroup fields in your Okteto manifest, and try again (https://okteto.com/docs/reference/manifest/index.html#securityContext-object-optional)."),
```

cmd/down.go
pchico83: the user can remove it manually

pkg/k8s/pods/pod.go
pchico83: Not really. The job would create a pod, and we would need to monitor the pod anyway, and clean the pod and the job. jobs are suitable for cronjobs or tasks where you want to run up to 1000 tasks with less than X errors.

PR #540, opened 2019-11-08
Build command for remote builds support

pchico83:


cmd/build.go
```diff
@@ -0,0 +1,197 @@
...
+//Build build and optionally push a Docker image
+func Build() *cobra.Command {
+	var file string
+	var tag string
+	var target string
+	var noCache bool
+
+	cmd := &cobra.Command{
+		Use:   "build",
+		Short: "Build (and optionally push) a Docker image",
```
rberrelleza: should we say Docker even if we are using `Buildkit`?

cmd/build.go
```diff
@@ -0,0 +1,197 @@
...
+		RunE: func(cmd *cobra.Command, args []string) error {
+			log.Debug("starting build command")
```
rberrelleza: can we add analytics? It'd be useful to see if people use this command.

cmd/build.go
```diff
@@ -0,0 +1,197 @@
...
+	return nil
+}
+
+func getBuildKitHost() (string, error) {
+	buildKitHost := os.Getenv("BUILDKIT_HOST")
+	if buildKitHost != "" {
+		return buildKitHost, nil
+	}
+
+	c, restConfig, namespace, err := k8Client.GetLocal()
```
rberrelleza: could we split the logic for a remote buildkit_host in a separate function so it's more obivous?

cmd/build.go
```diff
@@ -0,0 +1,197 @@
...
+		return "", err
+	}
+	if err := forwarder.Start(buildKitContainer, internalNamespace); err != nil {
+		return "", err
+	}
+	return fmt.Sprintf("tcp://localhost:%d", localPort), nil
+}
+
+func getSolveOpt(buildCtx, file, tag, target string, noCache bool) (*client.SolveOpt, error) {
+	if file == "" {
```
rberrelleza: could we make the name of the Dockerfile optional, like docker?

rberrelleza: ah, this is for the empty case. could we instead use the default of the argument to set the default value of `file`?

cmd/build.go
pchico83: it is a docker image what we build. buildkit supports other formats

cmd/build.go
pchico83: Yes, I will add them before merging

cmd/build.go
pchico83: no, it depends on the value of the build context argument

pkg/okteto/credential.go
```diff
@@ -33,3 +39,25 @@ func GetCredentials(ctx context.Context, namespace string) (*Credential, error)
...
+	if err != nil {
+		return nil, nil, "", err
+	}
+	defer os.Remove(file.Name())
+
+	if err := SetKubeConfig(cred, file.Name(), internalNamespace, GetUserID(), "okteto"); err != nil {
+		return nil, nil, "", err
+	}
+	os.Setenv("KUBECONFIG", file.Name())
+	c, restConfig, _, err := client.GetLocal()
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

rberrelleza: @pchico83 do we need to check this error?

PR #542, opened 2019-11-12
stop uploading releases to s3, since we download from github now

rberrelleza:
This way we reduce external dependencies. The only build we upload to s3 is latest from master. That also makes it easier to stop using my personal aws account 😛


rberrelleza:
A few things still depend on that URL, let me fix those first and then I'll resubmit.

PR #549, opened 2019-11-19
Support for gpu resources

### LINKED ISSUE #548: GPUs specified in the resources section of an okteto.yml are ignored
**Describe the bug**
When specifying GPUs in the `resources` section of a `okteto.yml`, this gets ignored in the deployment.

**To Reproduce**
Example manifest:
```yaml
name: demo
image: tensorflow/tensorflow:latest-gpu
command: ["bash"]
resources:
  limits:
    nvidia.com/gpu: 1
```
**Expected behavior**
The resource specification should be translated properly into the corresponding k8s deployment.

pchico83:
Fixes #548


kklemon:
I could only verify it for Nvidia cards but it works perfectly.

PR #556, opened 2019-11-22
Add Development Section to Contributing Documentation

danielhelfand:
This pull request adds a small section on building Okteto locally. If there are any other considerations that are recommended, would be more than happy to include them (e.g. running tests, etc).


rberrelleza:
LGTM. Could you include that you can run the unit tests by running `make test`?

danielhelfand:
@rberrelleza Added golang version and note on running unit tests. Thanks for the feedback!

PR #560, opened 2019-11-29
Analytics review

pchico83:


pkg/analytics/analytics.go
```diff
@@ -51,61 +53,81 @@ func init() {
...
+func TrackSyncError() {
+	track(syncErrorEvent, false, nil)
 }
 
 // TrackUp sends a tracking event to mixpanel when the user activates a development environment
-func TrackUp(image, version string, success bool) {
-	track(upEvent, version, image, success)
+func TrackUp(success bool, dev, clusterType string, single, swap bool) {
+	props := map[string]interface{}{
+		"dev":         dev,
```
rberrelleza:
```suggestion
		"devEnvironmentName":         dev,
```

pkg/analytics/analytics.go
```diff
@@ -51,61 +53,81 @@ func init() {
...
-	track(namespaceDeleteEvent, version, "", success)
+func TrackDeleteNamespace(success bool) {
+	track(namespaceDeleteEvent, success, nil)
+}
+
+// TrackReconnect sends a tracking event to mixpanel when the dev environment reconnect
+func TrackReconnect(success bool, clusterType string, swap bool) {
+	props := map[string]interface{}{
+		"clusterType": clusterType,
+		"swap":        swap,
```
rberrelleza:
```suggestion
		"swap":        strconv.FormatBool(swap),
```

cmd/up.go
```diff
@@ -29,19 +29,27 @@ import (
...
+)
+
 // UpContext is the common context of all operations performed during
 // the up command
 type UpContext struct {
 	Context    context.Context
 	Cancel     context.CancelFunc
 	Dev        *model.Dev
+	Namespace  *apiv1.Namespace
+	isAttach   bool
```
```suggestion
	isSwap   bool
```

cmd/up.go
```diff
@@ -29,19 +29,27 @@ import (
...
+	apiv1 "k8s.io/api/core/v1"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 )
 
 // ReconnectingMessage is the message shown when we are trying to reconnect
 const ReconnectingMessage = "Trying to reconnect to your cluster. File synchronization will automatically resume when the connection improves."
 
+var (
+	localClusters = []string{"127.", "10.", "172.", "192.", "169.", "localhost"}
```
rberrelleza: I'd remove `10.` from this list, since that it's also a typical IP for a private network (that's the default for GKE, EKS, Digital Ocean, etc...), and for the case where you access kubernetes through a VPN instead of through the public internet.

rberrelleza: Also, could the address of a cluster be an ipv6 address? This might be useful: https://stackoverflow.com/questions/41240761/go-check-if-ip-address-is-in-private-network-space

cmd/up.go
```diff
@@ -547,10 +558,31 @@ func (up *UpContext) runCommand() error {
...
+	}
+	return "remote"
+}
+
 // Shutdown runs the cancellation sequence. It will wait for all tasks to finish for up to 500 milliseconds
 func (up *UpContext) shutdown() {
 	log.Debugf("up shutdown")
-	analytics.TrackUp(up.Dev.Image, config.VersionString, up.success)
+	if !up.success {
+		analytics.TrackUp(false, up.Dev.Name, up.getClusterType(), up.isAttach, len(up.Dev.Services) == 0)
```
rberrelleza: I'd send a separate event for up failed so we don't track double ups.

PR #590, opened 2019-12-11
support remote from the manifest

### LINKED ISSUE #570: Define the remote in the okteto manifest
This is so developers won't have to manually add this option when starting development mode and they can just start working.

An option in the manifest to start okteto with remote mode

Every developer needs to remember to start okteto with the `--remote` flag.

rberrelleza:
Fixes #570 

## Proposed changes
- Adds the key `remote` to the manifest, functioning just as the --remote parameter.

### DIFF
--- a/cmd/up.go
+++ b/cmd/up.go
@@ -60,7 +60,6 @@ type UpContext struct {
 	Sy         *syncthing.Syncthing
 	ErrChan    chan error
 	cleaned    chan struct{}
-	remotePort int
 	success    bool
 }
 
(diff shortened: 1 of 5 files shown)


cmd/up.go
```diff
@@ -60,12 +60,15 @@ type UpContext struct {
 	Sy         *syncthing.Syncthing
 	ErrChan    chan error
 	cleaned    chan struct{}
-	remotePort int
 	success    bool
 }
 
 func (up *UpContext) remoteModeEnabled() bool {
-	return up.remotePort > 0
+	if up.Dev == nil {
```
pchico83: This should never happen, right?

rberrelleza: it could if  this is called during a cleanup and  up.Dev wasn't set.

PR #591, opened 2019-12-11
keep the original command while in remote mode

### LINKED ISSUE #572: Run the command when in remote mode
I want okteto to still run the command when running in --remote mode

Run the command manually after

rberrelleza:
Fixes #572 

## Proposed changes
- update start.sh so to use the https://github.com/okteto/supervisor instead of launching processes
- keep the original command for remote


rberrelleza:
@pchico83 do you think we should add a healthcheck, so the container dies if the SSH server dies? Other option could be to use a process manager as the command, and have that launch syncthing and remote (https://github.com/ochinchina/supervisord looks interesting).

pchico83:
@rberrelleza I would go with https://github.com/ochinchina/supervisord, but I am fine with the other approaches for now. The supervisor ensures the pod is not restarted, which cleans all data in memory or not belonging to the docker image. And who knows, we might need more processes in the future.

rberrelleza:
Updated with the supervisor change. @pchico83 can you give it a pass?

pkg/model/dev_test.go
```diff
@@ -298,8 +298,8 @@ func Test_LoadRemote(t *testing.T) {
 
 	dev.LoadRemote()
 
-	if dev.Command[0] != "/var/okteto/bin/remote" {
-		t.Errorf("remote command wasn't set: %s", dev.Command)
+	if dev.Command[0] != "uwsgi" {
```
pchico83: what is this for?

rberrelleza: test to validate that the command is not replaced when remote mode is enabled.

PR #592, opened 2019-12-12
Update how it works image and explanation

pchico83:
https://github.com/okteto/okteto/blob/how-it-works/docs/how-does-it-work.md


rberrelleza:
I think the text makes a lot of assumptions about the user knowing how okteto works. I think we should explain every step more thoroughly, and maybe correlate to the values in the manifest?

docs/how-does-it-work.md
```diff
@@ -1,21 +1,12 @@
 ## How does Okteto work?
 
-This is how a standard dev environment looks like:
+Okteto swaps your applications pods by a development environment:
```
rberrelleza:
```suggestion
Okteto swaps your applications pod with a preconfigured development environment:
```

docs/how-does-it-work.md
```diff
@@ -1,21 +1,12 @@
...
 
-<img align="left" src="env.png">
+<img align="left" src="okteto-architecture.png">
 
-&nbsp;
+The **api-dev** pod duplicates the manifest of the **api** pod, so it is fully integrated with your Kubernetes setup.
```
rberrelleza: this is not obvious a lot. I think you should mention how it's fully integrated.

Issue #574, opened 2019-12-09
Make remote mode compatible with Intellij's SSH plugin


rberrelleza:
**Is your feature request related to a problem? Please describe.**
I want to be able to use `okteto up --remote` with PHPStorm or Pycharm's remote dev features

**Describe the solution you'd like**
I want to run `okteto up --remote` and be able to connect it to pycharm or phpstorm as a remote interpreter. 

**Describe alternatives you've considered**
Creating my own dev image with an SSH server, or manually configuring the whole thing.

RubenAtPA:
Some additional information concerning PHPStorm (and maybe others):

There is a problem with the ssh command `cd` in the okteto ssh-server.
When using the  ssh-server configured directly in the image:
`ssh root@localhost -p 22000 cd /var/www && echo "Hello world"`  => `$ Hello world`
When using the ssh-server configured with okteto remote:
`ssh root@localhost -p 22000 cd /var/www && echo "Hello world"` => `$ internal error`

All other commands seem to work just find, so it looks like it's isolated to the `cd` command which is unfortunately used by default by PHPStorm.

rberrelleza:
@RubenAtPA which dev image are you using for the `--remote` case?

rberrelleza:
fixed by #590

PR #594, opened 2019-12-18
Support for remote forward tunnels

rberrelleza:
Contributes to #574 

## Proposed changes
- Add support for remote forward tunnels, so that debuggers and similar can connect back to localhost

SEE https://github.com/okteto/okteto/pull/594#issuecomment-566965320 for updated information

------ DEPRECATED
This is how the manifest would look like:
```
name: hello-world
image: okteto/golang:1
workdir: /okteto
command: ["go", "run", "main.go"]
volumes:
  - /go/pkg/
  - /root/.cache/go-build/
securityContext:
  capabilities:
    add:
    - SYS_PTRACE
forward:
  - 8080:8080
  - 2345:2345
remoteForward:
  - 9000:9000
  - 9001:9001
```
And this is how the UI looks like:
image

Design decisions:
- I'm using the term Remote Forward Tunnel to be consistent with the terminology used by OpenSSH
- We always open the listeners in localhost in both sides, to simplify the notation. If we want to allow forwarding traffics for servers not on `localhost` we could change the notation to something like: `127.0.0.1:8080:0.0.0:8080`. This is what OpenSSH does (but I find this extremely confusing)
- This feature requires SSH access, if remote is not enabled, it will automatically enable remote in a random port. If there is another remote running on 2222, the tunnel will fail.
- A tunnel failure doesn't interrupt the `up` sequence. It will pause for a second, and it will try to reconnect. All errors are logged to the log file,  and not to the UI.

I'll file a ticket to update our docs once this is fixed.

### DIFF
--- a/pkg/ssh/reverse.go
+++ b/pkg/ssh/reverse.go
@@ -20,20 +20,20 @@ type reverse struct {
 // ReverseManager handles the lifecycle of all the remote forwards
 type ReverseManager struct {
 	reverses map[int]*reverse
-	ctx            context.Context
-	sshUser        string
-	sshHost        string
-	sshPort        int
+	ctx      context.Context
+	sshUser  string
+	sshHost  string
+	sshPort  int
 }
 
 // NewReverseManager returns a newly initialized instance of RemoteReverseManager
 func NewReverseManager(ctx context.Context, sshPort int) *ReverseManager {
 	return &ReverseManager{
-		ctx:            ctx,
+		ctx:      ctx,
 		reverses: make(map[int]*reverse),
-		sshUser:        "root",
-		sshHost:        "localhost",
-		sshPort:        sshPort,
+		sshUser:  "root",
+		sshHost:  "localhost",
+		sshPort:  sshPort,
 	}
 }
 
@@ -95,7 +95,7 @@ func (r *reverse) start(ctx context.Context, c *ssh.ClientConfig, sshAddr string
 	}
 
 	// Listen on remote server port
-	listener, err := serverConn.Listen("tcp", fmt.Sprintf("localhost:%d", r.remotePort))
+	listener, err := serverConn.Listen("tcp", fmt.Sprintf("0.0.0.0:%d", r.remotePort))
 	if err != nil {
 		return fmt.Errorf("failed open remote port %d: %s", r.remotePort, err)
 	}
--- a/pkg/ssh/reverse.go
+++ b/pkg/ssh/reverse.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"github.com/okteto/okteto/pkg/log"
+	"github.com/okteto/okteto/pkg/model"
 	"golang.org/x/crypto/ssh"
 )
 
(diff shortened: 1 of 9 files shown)


pkg/ssh/reverse.go
```diff
@@ -0,0 +1,150 @@
...
+	// Connect to SSH remote server using serverEndpoint
+	c := &ssh.ClientConfig{
+		User:            r.sshUser,
+		HostKeyCallback: ssh.InsecureIgnoreHostKey(),
+	}
+
+	sshAddr := fmt.Sprintf("%s:%d", r.sshHost, r.sshPort)
+
+	for _, rt := range r.remoteForwards {
+		go rt.startWithRetry(r.ctx, c, sshAddr)
```
golangcibot: Error return value of `rt.startWithRetry` is not checked (from `errcheck`)

pchico83:
@rberrelleza Code looks good to me. About your comments:
```
I'm using the term Remote Forward Tunnel to be consistent with the terminology used by OpenSSH
```
We should use `Local Forward Tunnel` for the `forward` field then. I would go with `reverse`, in singular, for this field, as we do with `forward`. 
Also, I would show this info in the UI:
```
Namespace: xxxx
Name:            xxxx
Forward:    8080 <-80
                    8081 <- 81
Reverse:     90 -> 9090
                     91 -> 9091
```
using the arrow between ports to indicate the direction of the tunnel.

-------------

```
We always open the listeners in localhost in both sides, to simplify the notation. If we want to allow forwarding traffics for servers not on localhost we could change the notation to something like: 127.0.0.1:8080:0.0.0:8080. This is what OpenSSH does (but I find this extremely confusing)
```
Lets expose it on 0.0.0.0 on the remote. This is because you might want to attach a process running on a different pod. Think about our php sample, the php process run in a different pod than the dev environment.

-------------

```
This feature requires SSH access, if remote is not enabled, it will automatically enable remote in port 2222. If there is another remote running on 2222, the tunnel will fail.
```
I would expose it to a random port, and I wouldn't show the port in the UI for this case. But I am fine with the current approach if changing it is too much work.

-------------

```
A tunnel failure doesn't interrupt the up sequence. It will pause for a second, and it will try to reconnect. All errors are logged to the log file, and not to the UI.
```
Ideally, I would check the ssh port and then the reverse ports as part of the up sequence, and fail if it does work. And I will restart the `up` sequence if after that, the port is not accessible for a while. But let's keep it like this and open an issue to monitor the ssh port and the reverse tunnels.

rberrelleza:
Based on the comments,  I updated the UI with the arrows to keep local ports in the left and remote ports in the right. The arrow indicates the data flow

image

@rlamana @pchico83 thoughts?

rlamana:
It looks good to me. However I was thinking how would it look to label the columns (remote  local) on the top, like some sort of table, to clarify even more. Wdyt?

rberrelleza:
> It looks good to me. However I was thinking how would it look to label the columns (remote local) on the top, like some sort of table, to clarify even more. Wdyt?

that sounds pretty cool. do you have a sketch of how it would look like? I can do a separate PR for that.

Issue #517, opened 2019-10-28
KIND not supported by "okteto up"


Albertoimpl:
Hi there!
I was trying okteto on my local machine using [KIND](https://github.com/kubernetes-sigs/kind):

```
kind version
v0.5.1
```

```
okteto version
okteto version 1.5.1
```

With the following autogenerated with `okteto init`  okteto.yml:
```
name: app-okteto
image: okteto/gradle:latest
command:
- bash
workdir: /okteto
volumes:
- /home/gradle/.gradle
forward:
- 8080:8080
- 8088:8088
```

After running `okteto up` it errors with the following message:
```
 x  Error: stat /tmp/hostpath_pv/f09519e7-801a-4622-93de-60658c2961f3: no such file or directory
```

Some extra cluster-info dump:
```
            "status": {
                "phase": "Pending",
                "conditions": [
                    {
                        "type": "Initialized",
                        "status": "True",
                        "lastProbeTime": null,
                        "lastTransitionTime": "2019-10-28T17:05:52Z"
                    },
                    {
                        "type": "Ready",
                        "status": "False",
                        "lastProbeTime": null,
                        "lastTransitionTime": "2019-10-28T17:05:39Z",
                        "reason": "ContainersNotReady",
                        "message": "containers with unready status: [dev]"
                    },
                    {
                        "type": "ContainersReady",
                        "status": "False",
                        "lastProbeTime": null,
                        "lastTransitionTime": "2019-10-28T17:05:39Z",
                        "reason": "ContainersNotReady",
                        "message": "containers with unready status: [dev]"
                    },
                    {
                        "type": "PodScheduled",
                        "status": "True",
                        "lastProbeTime": null,
                        "lastTransitionTime": "2019-10-28T17:05:39Z"
                    }
                ],
                "hostIP": "172.17.0.2",
                "podIP": "10.244.0.10",
                "startTime": "2019-10-28T17:05:39Z",
                "initContainerStatuses": [
                    {
                        "name": "okteto-bin",
                        "state": {
                            "terminated": {
                                "exitCode": 0,
                                "reason": "Completed",
                                "startedAt": "2019-10-28T17:05:51Z",
                                "finishedAt": "2019-10-28T17:05:51Z",
                                "containerID": "containerd://ba76084795d46412c57494d8ce7e4d316932987c64cdb480660728f8e461a4ee"
                            }
                        },
                        "lastState": {},
                        "ready": true,
                        "restartCount": 0,
[...]

PR #595, opened 2019-12-18
Switch to disable persistent volumes

### LINKED ISSUE #514: Support for "okteto up" without persistent volumes
PVs make the orchestration of dev envs more complex due to zone/node constraints. There are dev envs where PVCs provide little value and I want to deactivate them.

An option to deactivate PVCs in my dev env.

rberrelleza:
Fixes #514 #517 

## Proposed changes
- Add a key to the manifest to control the usage of PVCs
- By default, it will be enabled. 

Manifest looks like this:
```
name: hello-world
image: okteto/golang:1
workdir: /okteto
command: ["go", "run", "main.go"]
persistentVolume: false
securityContext:
  capabilities:
    add:
    - SYS_PTRACE
```

### DIFF
--- a/pkg/model/dev.go
+++ b/pkg/model/dev.go
@@ -393,11 +393,12 @@ func (dev *Dev) syncthingSubPath() string {
 // ToTranslationRule translates a dev struct into a translation rule
 func (dev *Dev) ToTranslationRule(main *Dev) *TranslationRule {
 	rule := &TranslationRule{
-		Container:       dev.Container,
-		Image:           dev.Image,
-		ImagePullPolicy: dev.ImagePullPolicy,
-		Environment:     dev.Environment,
-		WorkDir:         dev.WorkDir,
+		Container:        dev.Container,
+		Image:            dev.Image,
+		ImagePullPolicy:  dev.ImagePullPolicy,
+		Environment:      dev.Environment,
+		WorkDir:          dev.WorkDir,
+		PersistentVolume: dev.PersistentVolumeEnabled(),
 		Volumes: []VolumeMount{
 			{
 				Name:      OktetoVolumeName,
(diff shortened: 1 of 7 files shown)


pkg/model/dev.go
```diff
@@ -57,31 +57,33 @@ var (
 
 	devReplicas                      int32 = 1
 	devTerminationGracePeriodSeconds int64
+	persistentVolumeEnabled          bool = true
```
golangcibot: var `persistentVolumeEnabled` is unused (from `unused`)

rberrelleza:
I still get the "attaching to volume" event in the UI even when do we don't create the volume. Any ideas why @pchico83 ? the `okteto`volume is not created. maybe it's waiting for the volume with the token? 🤔

pkg/model/dev.go
```diff
@@ -268,6 +270,16 @@ func (dev *Dev) validate() error {
 		return err
 	}
 
+	if !dev.PersistentVolumeEnabled() {
+		if len(dev.Services) > 0 {
+			return fmt.Errorf("'persistentVolume' must be set to true to work with services")
+		}
+
+		if len(dev.Volumes) > 0 {
```
pchico83: Thinking twice about this,  we can enable volumes in this case. The reason would be to mount subpaths of the local folder at a given path in the pod. This is useful to set config files.

pchico83: Although I am finde deactivating them too and only support that use case with persistent volumes

rberrelleza: I think there's value on the subpath case. I'd rather support as many features as possible without persistence enabled. What's needed for that? Just let the validation pass?

pchico83: Right. Just remove the validation code

rberrelleza: > The reason would be to mount subpaths of the local folder at a given path in the pod

rberrelleza: Oh wait, you're talking about the 'volume from local folder' feature right?

pchico83: Right

PR #609, opened 2020-01-03
Forward service ports

### LINKED ISSUE #577: Access services in my namespace directly from my local machine
When I launched my application, it includes a frontend, a backend and a database. When I'm developing, I often want to access the DB via a client to inspect my data.  I'd like okteto to automate the connectivity so that  I can access other services locally, not just the ones I'm developing.

I'd like to specify the services I care about in the manifest, and be able to access them via localhost, just like I do for application ports.

Manually create port-forwards for the services.

rberrelleza:
Fixes #577

## Proposed changes
- `forwards`can now take a string. if a string is given, the forward will go to a service in the namespace instead of to the local dev env

```
name: hello-world
image: okteto/golang:1
workdir: /okteto
command: ["bash"]
forward:
  - 8080:8080
  - 3306:mariadb:3306
```

@rlamana how do you think we should show this in the CLI's UI? Separate from the forward?

This is how it looks right now. It's sorted so the dev env ports are shown first
image


pkg/k8s/pods/pod_test.go
```diff
@@ -0,0 +1,116 @@
...
+				},
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			c := fake.NewSimpleClientset(ns)
+			for _, p := range tt.pods {
+				c.Tracker().Add(&p)
```
golangcibot: Error return value of `(k8s.io/client-go/testing.ObjectTracker).Add` is not checked (from `errcheck`)

rberrelleza:
> I would make the remote port mandatory as kubectl does. I find it quite confusing to make it optional.

Yeah, I was doubtful about that one as well. I'll remove it.

> Also, could you document the new syntax?
Yes, I'll do it once we release the new version.

Issue #602, opened 2019-12-23
Support for okteto PVC size at the okteto manifest level


pchico83:
**Is your feature request related to a problem? Please describe.**
My app needs a few MBs for my okteto folder. Creating a 10GB volume seems overkill

**Describe the solution you'd like**
Define the volume size at the okteto manifest level since this is a setting related to the app.

pchico83:
Closed by #627

Issue #508, opened 2019-10-24
Support for setting the storage class for the okteto PVc


pchico83:
**Is your feature request related to a problem? Please describe.**
Okteto volume is created in a different zone that the pod of my dev environment.

**Describe the solution you'd like**
Set the storage class instead of using the default one. This way I have control to set the volume zone or to use `WaitForFirstConsumer` `volumeBindingMode`.

PR #625, opened 2020-01-10
Move to a single PVC per dev environment to simplify management

pchico83:
I order to support:

- https://github.com/okteto/okteto/issues/508
- https://github.com/okteto/okteto/issues/602

PVCs can not be shared between dev environments.

This PR creates a single PVC for each dev environment.
If `volumes` field is defined, they are created as subpaths of the PVC.


cmd/down.go
```diff
@@ -107,13 +106,11 @@ func runDown(dev *model.Dev) error {
...
 
-	if err := statefulsets.Destroy(dev, client); err != nil {
-		return err
-	}
-
-	if err := removeVolumes(dev); err != nil {
-		return err
-	}
+	//remove deprecated okteto volume
+	client.CoreV1().PersistentVolumeClaims(dev.Namespace).Delete(
```
golangcibot: Error return value of `(k8s.io/client-go/kubernetes/typed/core/v1.PersistentVolumeClaimInterface).Delete` is not checked (from `errcheck`)

Issue #624, opened 2020-01-10
Return a user friendly error when the storage class is not available


rberrelleza:
If persistent volumes are enabled and the cluster doesn't have a default storage class, we should return a user friendly error instead of:
```
 x  Error: stat /var/lib/k8s-pvs/okteto/pvc-faae04be-971a-4904-99a8-a45b85731fa8: no such file or directory
```

rberrelleza:
okteto int takes care of this now

PR #627, opened 2020-01-10
Support for custom PVC storageclass and size

### LINKED ISSUE #508: Support for setting the storage class for the okteto PVc

pchico83:
Fixes #508 #602

### DIFF
--- a/pkg/k8s/volumes/crud.go
+++ b/pkg/k8s/volumes/crud.go
@@ -9,6 +9,8 @@ import (
 	"github.com/okteto/okteto/pkg/log"
 	"github.com/okteto/okteto/pkg/model"
 
+	apiv1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"k8s.io/client-go/kubernetes"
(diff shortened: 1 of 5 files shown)


pkg/model/dev_test.go
```diff
@@ -473,7 +473,8 @@ func Test_validate(t *testing.T) {
       name: deployment
       container: core
       image: code/core:0.1.8
-      persistentVolume: false
+      persistentVolume:
+        enabled: false
```
rberrelleza: can we add a test with a valid size included?

pkg/k8s/volumes/crud.go
```diff
@@ -37,6 +39,31 @@ func Create(ctx context.Context, dev *model.Dev, c *kubernetes.Clientset) error
 	return nil
 }
 
+func checkPVCValues(pvc *apiv1.PersistentVolumeClaim, dev *model.Dev) error {
```
rberrelleza: can we also take care of #624 with this change?

pchico83: Let's keep this PR short and fix #624 on a different PR

PR #648, opened 2020-01-24
fail fast if inside a cluster

### LINKED ISSUE #287: Okteto up inside of a cloud environment enters an error loop
**Describe the bug**
I accidentally run okteto up in my dev environment. 

**Expected behavior**
Okteto up should just fail in this case

rberrelleza:
Fixes #287

## Proposed changes
Prevent up, namespace, login  from working on a remote dev environment

### DIFF
--- a/cmd/down.go
+++ b/cmd/down.go
@@ -14,6 +14,8 @@
 package cmd
 
 import (
+	"strings"
+
 	"github.com/okteto/okteto/pkg/analytics"
 	"github.com/okteto/okteto/pkg/errors"
 	k8Client "github.com/okteto/okteto/pkg/k8s/client"
@@ -27,7 +29,6 @@ import (
 	"github.com/okteto/okteto/pkg/syncthing"
 	"github.com/spf13/cobra"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"strings"
 )
 
 //Down deactivates the development environment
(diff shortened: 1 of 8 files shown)


cmd/down.go
```diff
@@ -42,6 +43,10 @@ func Down() *cobra.Command {
 		RunE: func(cmd *cobra.Command, args []string) error {
 			log.Info("starting down command")
 
+			if k8Client.InCluster() {
```
pchico83: `down` works if your dev environment is in Okteto Cloud... same for `exec`...
Maybe we shouldn't inject the `okteto` and `kubectl` binary if you are not in Okteto Cloud.

rberrelleza: yes, I think that's probably a better approach.

pkg/k8s/client/client.go
```diff
@@ -51,3 +51,13 @@ func GetLocal() (*kubernetes.Clientset, *rest.Config, string, error) {
 	}
 	return client, config, namespace, nil
 }
+
+// InCluster returns true if Okteto is running on a Kubernetes cluster
+func InCluster() bool {
+	_, err := rest.InClusterConfig()
+	if err == nil {
```
golangcibot: S1008: should use 'return <expr>' instead of 'if <expr> { return <bool> }; return <bool>' (from `gosimple`)

PR #679, opened 2020-02-14
get registry and buildkit when login in

rberrelleza:

### DIFF
--- a/pkg/okteto/auth.go
+++ b/pkg/okteto/auth.go
@@ -32,6 +32,10 @@ const (
 
 	// CloudURL is the default URL of okteto
 	CloudURL = "https://cloud.okteto.com"
+	// CloudRegistryURL is the default URL of okteto registry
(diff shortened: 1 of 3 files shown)


pkg/okteto/auth.go
```diff
@@ -163,7 +165,27 @@ func GetURL() string {
...
+	}
+
+	return t.Registry
+}
+
+// GetBuildkit returns the URL of the authenticated user
+func GetBuildkit() string {
+	t, err := GetToken()
+	if err != nil {
+		return "registry.cloud.okteto.net"
```
pchico83: registry -> buildkit

pchico83: I would return and error here, auth will fail otherwise

pchico83: The logic should be return error if no okteto login. Return this default value  if it is empty (needs to login again)

PR #680, opened 2020-02-14
Update kubectl libraries

pchico83:


rberrelleza:
woudl this give us trouble with 1.13 clusters?

PR #682, opened 2020-02-14
Independent build cache by user

pchico83:
This uses a different build-cache by user. More information here:
https://github.com/moby/buildkit/blob/master/frontend/dockerfile/docs/experimental.md#run---mounttypecache

### DIFF
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -14,10 +14,13 @@
 package cmd
 
 import (
+	"bufio"
 	"context"
 	"fmt"
+	"io/ioutil"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strings"
 
 	"github.com/containerd/console"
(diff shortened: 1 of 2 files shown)


cmd/build.go
```diff
@@ -208,3 +222,61 @@ func getRepoNameWithoutTag(name string) string {
...
+func getFileWithCacheHandler(filename string) (string, error) {
+	file, err := os.Open(filename)
+	if err != nil {
+		return "", err
+	}
+	defer file.Close()
+
+	scanner := bufio.NewScanner(file)
+
+	tmpFile, err := ioutil.TempFile(config.GetHome(), "okteto-build")
```
rberrelleza: should we do `./okteto/.build`  to keep everything under the same folder?

pchico83:
@rberrelleza the elegant solution would be to have our own buildkit frontend. But for now, I think this is a quite safe transformation. We have the advantage of not breaking any users at this moment :-)

PR #686, opened 2020-02-15
wrap exec command on a shell to run multiple commands

### LINKED ISSUE #685: `okteto exec` fails to execute chained commands
**Describe the bug**
Running a  command with `&&` or `||` fails in the context of `okteto exec`

**To Reproduce**
1. `okteto exec "mkdir foo && cd foo && ls -la"`

**Expected behavior**
Command should execute successfully

 - Version 1.7.1

rberrelleza:
Fixes #685

### DIFF
--- a/cmd/exec.go
+++ b/cmd/exec.go
@@ -78,9 +78,12 @@ func Exec() *cobra.Command {
 
 func executeExec(ctx context.Context, dev *model.Dev, args []string) error {
 
+	wrapped := []string{"sh", "-c"}
+	wrapped = append(wrapped, args...)
+
 	if dev.ExecuteOverSSHEnabled() {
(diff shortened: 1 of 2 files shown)


cmd/exec.go
```diff
@@ -78,9 +78,12 @@ func Exec() *cobra.Command {
 
 func executeExec(ctx context.Context, dev *model.Dev, args []string) error {
 
+	wrapped := []string{"sh", "-c"}
```
rberrelleza: @pchico83 do you see any potential side effects on this approach?

pchico83: not having `sh`, but we already have that problem everywhere

rberrelleza: true. we need to document that for now.

PR #689, opened 2020-02-16
create buildkit cache folder if missing

rberrelleza:


cmd/build.go
```diff
@@ -103,8 +103,9 @@ func RunBuild(path, file, tag, target string, noCache bool) error {
 	if buildKitHost == okteto.GetBuildKit() {
 		fileWithCacheHandler, err := getFileWithCacheHandler(file)
 		if err != nil {
-			return err
+			return fmt.Errorf("failed to get build cache: %s", err)
```
pchico83: failed to get build temporal folder?

rberrelleza: do we need this folder to be in ~/.okteto? Or can we just create it in the tmp dir entirely?

pchico83: It fails in tmp due to lack of permissions when running buildkit

PR #692, opened 2020-02-17
Retry on download error

rberrelleza:
## Proposed changes
- Retry a few times before failing the download
- Don't fail okteto up on download failure.
- Download syncthing directly from github


cmd/up.go
```diff
@@ -105,7 +105,8 @@ func Up() *cobra.Command {
 			if syncthingUpgradeAvailable() {
 				fmt.Println("Installing dependencies...")
 				if err := downloadSyncthing(); err != nil {
-					return fmt.Errorf("couldn't download syncthing, please try again")
+					log.Yellow("couldn't download syncthing, please try again later")
```
rberrelleza: @pchico83 what do you think about this part?

pchico83: does it fail if a previous version is not downloaded?

rberrelleza: > does it fail if a previous version is not downloaded?

It will fail further down when we try to launch `syncthing`

pchico83: I would fail here in that case

pkg/syncthing/install.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+
+	log.Infof("downloaded syncthing %s to %s", versionRegex, i)
+	return nil
+}
+
+// IsInstalled returns true if syncthing is installed
+func IsInstalled() bool {
+	_, err := os.Stat(getInstallPath())
+	if os.IsNotExist(err) {
```
golangcibot: S1008: should use 'return <expr>' instead of 'if <expr> { return <bool> }; return <bool>' (from `gosimple`)

PR #695, opened 2020-02-18
Keep ssh config until down is run

NerdsvilleCEO:
Fixes ssh config getting removed on exit of up command (as discussed on slack)

### DIFF
--- a/cmd/up.go
+++ b/cmd/up.go
@@ -678,12 +678,6 @@ func (up *UpContext) shutdown() {
 		log.Info("sent cancellation signal")
 	}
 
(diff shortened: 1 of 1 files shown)


pchico83:
@NerdsvilleCEO there are merge conflicts, could you update the PR?

NerdsvilleCEO:
Of course @pchico83... It probably got outdated, I'll add a unit test and rebase and then ping once completed

pchico83:
@NerdsvilleCEO It is probably because we are moving logic from the `cmd` folder to the `pkg` folder:
https://github.com/okteto/okteto/blob/master/pkg/cmd/down/run.go

PR #698, opened 2020-02-19
Buildkit auth

rberrelleza:
Use the token as the authentication mechanism for buildkit


cmd/build.go
```diff
@@ -69,12 +74,28 @@ func Build() *cobra.Command {
...
-	c, err := client.New(ctx, buildKitHost, client.WithFailFast())
+	b, err := url.Parse(buildKitHost)
 	if err != nil {
-		return "", err
+		return "", errors.Wrapf(err, "invalid buildkit host %s", buildKitHost)
+	}
+
+	okToken, err := okteto.GetToken()
+	if err != nil {
+		return "", errors.Wrapf(err, "invalid okteto token, please run `okteto login` again")
```
pchico83: errors.ErrNotLogged?

pkg/okteto/cloud.go
```diff
@@ -0,0 +1,59 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package okteto
+
+const (
```
pchico83: What about forcing `okteto login` to start using `okteto build` and remove this file?

rberrelleza: yeah, I think we should do that in all the commands.  #704 tracks this.

PR #710, opened 2020-02-25
Auto-create deployment on "okteto push" if it does not exist

pchico83:


cmd/push.go
```diff
@@ -87,15 +89,27 @@ func Redeploy() *cobra.Command {
 }
 
 func runPush(dev *model.Dev, imageTag, oktetoRegistryURL string, c *kubernetes.Clientset) error {
+	create := false
 	d, err := deployments.Get(dev, dev.Namespace, c)
 	if err != nil {
-		return err
+		if errors.IsNotFound(err) {
```
rberrelleza: Check first for `!errrors.IsNotFound(err)` to make it easier to read.

cmd/push.go
```diff
@@ -87,15 +89,27 @@ func Redeploy() *cobra.Command {
...
+		} else {
+			return err
+		}
 	}
 
 	buildKitHost, isOktetoCluster, err := build.GetBuildKitHost()
 	if err != nil {
 		return err
 	}
+	if create && imageTag == "" && !isOktetoCluster {
```
rberrelleza: this condition is hard to read, think about splitting it.

cmd/push.go
```diff
@@ -113,9 +127,19 @@ func runPush(dev *model.Dev, imageTag, oktetoRegistryURL string, c *kubernetes.C
 	spinner := newSpinner(fmt.Sprintf("Pushing source code to the development environment '%s'...", dev.Name))
 	spinner.start()
 	defer spinner.stop()
-	err = down.Run(dev, imageTag, d, c)
-	if err != nil {
-		return err
+	if create {
```
rberrelleza: Can we move the create logic to a separate function?

cmd/push.go
```diff
@@ -87,15 +89,27 @@ func Redeploy() *cobra.Command {
 }
 
 func runPush(dev *model.Dev, imageTag, oktetoRegistryURL string, c *kubernetes.Clientset) error {
+	create := false
 	d, err := deployments.Get(dev, dev.Namespace, c)
 	if err != nil {
-		return err
+		if errors.IsNotFound(err) {
+			if err := askIfDeploy(dev.Name, dev.Namespace); err != nil {
```
rberrelleza: can we also add the same `--deploy` param like we did for `up`?

PR #718, opened 2020-02-27
Monitor syncthing folder errors

### LINKED ISSUE #622: Monitor syncthing error events API
**Describe the bug**
`okteto up` is not able to complete because the remote syncthing is not able to write files on disk. Running `okteto down` and `okteto up` completes the initial syncthing because the syncthing database is corrupted, but the files are synched to the remote.

**Expected behavior**
`okteto up` should fail and explain the reason of the failure. Syncthing reports this warning:

```
WARNING: Error on folder \"spend-cloud-ui\" (okteto-spend-cloud-ui): insufficient space on disk for database (/var/syncthing/index-v0.14.0.db): 0.6 % < 1 %" process=syncthing
```
We could check for more warnings and errors.

Also, we should consider to have an option to rescan local/remote files, and maybe, clean the remote database everytime syncthing is restarted to avoid database corruption.

Another useful command would be to show the available size in the okteto volume.

pchico83:
Fixes #622


pkg/syncthing/monitor.go
```diff
@@ -15,58 +15,47 @@ package syncthing
...
-				if !connected {
-					log.Info("not connected to syncthing, sending disconnect signal")
-					disconnect <- struct{}{}
-					return
-				}
-				connected = false
+				continue
+			}
+			if !connected {
+				log.Info("syncthing not working, sending disconnect signal")
```
rberrelleza:
```suggestion
				log.Info("syncthing not connected, sending disconnect signal")
```

pchico83:
@rberrelleza I am not retrying these errors because all I have seen require user intervention, like no space in the remote volume, or syncthing failing to write due to file permission settings.

PR #724, opened 2020-02-28
Make remote SSH server port configurable

### LINKED ISSUE #723: Enable configuration of injected SSH server port

I need to use the injected SSH server in contexts where port 22 is already in use by another process, and in pods where I cannot add the `NET_BIND_SERVICE` capability.

I'd like to configure the injected SSH server port to the value of my choosing.

In the conflicting process case, there is no alternative for me. I can sometimes work around the capability issue when I have the ability to introduce a role/binding for the pod to enable the capability addition.

**Additional context**

The `remote` SSH server binary already supports a `OKTETO_REMOTE_PORT` variable to configure the port — `okteto` could be updated to pass through said configuration.

ironcladlou:
Add a new `sshServerPort` manifest field which allows the injected SSH server
port to be configurable, preserving the exiting default port 22.

This allows use of the injected SSH server in pods without the `NET_BIND_SERVICE`
capability, and in contexts where port 22 is already in use by another process.

Fixes #723


pkg/model/dev.go
```diff
@@ -449,6 +459,10 @@ func (dev *Dev) ToTranslationRule(main *Dev) *TranslationRule {
 				Name:  oktetoMarkerPathVariable,
 				Value: path.Join(dev.MountPath, dev.DevPath),
 			},
+			EnvVar{
```
pchico83: Could you add the envvar only if `SSHServerPort` is different than 22?
This way we modify the original deployment the minimum possible

ironcladlou: Agree — fixed.

ironcladlou:
Feedback addressed, and added unit tests.

PR #739, opened 2020-03-03
Run integration test in windows

### LINKED ISSUE #707: Run integration test in windows
Update the existing end-to-end integration test so that it also runs in windows.

rberrelleza:
Fixes #707


rberrelleza:
@pchico83 do you think we should run both server and clientside modes in windows too?

PR #745, opened 2020-03-04
Reconnect if syncthing is not healthy

rberrelleza:
Try and reconnect if we can't talk to syncthing.

### DIFF
--- a/pkg/syncthing/monitor.go
+++ b/pkg/syncthing/monitor.go
@@ -17,6 +17,7 @@ import (
 	"context"
 	"time"
 
+	"github.com/okteto/okteto/pkg/errors"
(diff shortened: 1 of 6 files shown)


pkg/syncthing/monitor.go
```diff
@@ -51,8 +52,8 @@ func (s *Syncthing) Monitor(ctx context.Context, disconnect chan error) {
 				continue
 			}
 			if !connected {
-				log.Info("syncthing not connected, sending disconnect signal")
-				disconnect <- err
+				log.Info("syncthing not connected, sending disconnect signal: %s", err)
```
golangcibot: printf: Info call has possible formatting directive %s (from `govet`)

PR #750, opened 2020-03-06
Increase required watches for warnings

pchico83:
Openshift comes with 65536by default


.gitignore
```diff
@@ -8,3 +8,4 @@ cli
 .syncthing*
 bin/*
 coverage.txt
+cmd/.stignore
```
rberrelleza: do we need this?

PR #756, opened 2020-03-10
Support for in-cluster login execution

pchico83:
Now we have the `--token` flag, there is no need to restrict the inc-luster execution of the login command. It can also enable different scenarios, like building/pushing from the cluster.

### DIFF
--- a/cmd/login.go
+++ b/cmd/login.go
@@ -20,7 +20,6 @@ import (
 
 	"github.com/okteto/okteto/pkg/analytics"
 	"github.com/okteto/okteto/pkg/cmd/login"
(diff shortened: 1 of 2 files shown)


cmd/login.go
```diff
@@ -49,9 +47,6 @@ to log in to a Okteto Enterprise instance running at okteto.example.com.
 `,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			ctx := context.Background()
-			if k8Client.InCluster() {
```
rberrelleza: we should still fail if `--token` is not used

PR #765, opened 2020-03-16
Support for "--plain" option to show plain build container output

pchico83:
Support to show plain build container output.
Buildkit does not show the output of "RUN" instructions by default. "--plain" is useful for debugging Dockerfile scenarios.
Ideally, we should parse the buildkit output and show the info as we prefer, but that requires a lot of work.

### DIFF
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -28,7 +28,7 @@ func Build() *cobra.Command {
 	var tag string
 	var target string
 	var noCache bool
-	var plain bool
+	var progress string
 	var buildArgs []string
 
 	cmd := &cobra.Command{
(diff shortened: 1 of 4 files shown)


cmd/build.go
```diff
@@ -64,6 +65,7 @@ func Build() *cobra.Command {
 	cmd.Flags().StringVarP(&tag, "tag", "t", "", "name and optionally a tag in the 'name:tag' format (it is automatically pushed)")
 	cmd.Flags().StringVarP(&target, "target", "", "", "set the target build stage to build")
 	cmd.Flags().BoolVarP(&noCache, "no-cache", "", false, "do not use cache when building the image")
+	cmd.Flags().BoolVarP(&plain, "plain", "", false, "show plain build container output")
```
rberrelleza: I'd use the same format as buildkit: `--progress=plain` and `--progress=tty` as the default?

PR #766, opened 2020-03-17
check error when getting the replica set

rberrelleza:


pkg/k8s/pods/pod.go
```diff
@@ -117,7 +117,7 @@ func GetDevPod(ctx context.Context, dev *model.Dev, c *kubernetes.Clientset, wai
 	log.Infof("deployment %s with revision %v is progressing", d.Name, d.Annotations[deploymentRevisionAnnotation])
 
 	rs, err := replicasets.GetReplicaSetByDeployment(dev, d, c)
-	if rs == nil {
+	if err == nil {
```
rberrelleza: @pchico83 I'm seeing a situation where both rs and err is nil, expected?

pchico83: yes, it is. if `nil, nil` is returned, `GetDevPodInLoop` retries. if an error is returned, `GetDevPodInLoop` fails.
`nil, nil` means the RS is not yet created, but it is not an error.

PR #770, opened 2020-03-20
Add a message on "okteto down" for discoverability of "okteto push"

pchico83:
## Proposed changes
- Add a message on "okteto down" for discoverability of "okteto push"

### DIFF
--- a/cmd/down.go
+++ b/cmd/down.go
@@ -53,6 +53,7 @@ func Down() *cobra.Command {
 			}
 
 			log.Success("Development environment deactivated")
+			log.Information("Run 'okteto push' to deploy your code changes to the cluster")
 
 			if rm {
 				if err := removeVolume(dev); err != nil {


rlamana:
@rberrelleza I honestly see a lot of value in providing information and helping the discoverability of other features in the tool, which is hard to achieve in a CLI tool.

As an example of other tools doing something similar, `git` constantly tells you what to do next in some of its commands. For example:

```shell
$ git status
On branch master
Your branch is behind 'origin/master' by 1 commit, and can be fast-forwarded.
  (use "git pull" to update your local branch)

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

	modified:   index.html

no changes added to commit (use "git add" and/or "git commit -a")
```

All that o help users with next steps and feature discoverability.

And that one is pretty verbose and we use it all the time. I think the one @pchico83 proposes is not that intrusive or annoying. And `okteto down` is not something you keep doing all the time.

rlamana:
@rberrelleza we could maybe create a different kind of information output that is not so prominent, instead of the intense blue we use for information. Would that help to make it less annoying?

rberrelleza:
@rlamana  I like the concept of the hint, in git it works pretty well (e.g. the suggesting to create a PR after you push a change). I'm not sure that after `down`,  `push` is the action that should go, but if you and @pchico83  like it, let's go for it.

cmd/down.go
```diff
@@ -53,6 +53,7 @@ func Down() *cobra.Command {
 			}
 
 			log.Success("Development environment deactivated")
+			log.Information("Execute 'okteto push' to permanently apply your changes to your cluster")
```
rberrelleza:
```suggestion
			log.Information("Run 'okteto push' to deploy your local changes in your cluster")
```

PR #774, opened 2020-03-21
Generate ssh key

rberrelleza:
This is part of #475 

## Proposed changes
- Generate an SSH key pair
-  Add it to the SSH configuration

By itself, this doesn't do anything until `remote` accepts public key authentication

I decided to create a separate key to keep it separately from existing setup. This is something we could even use later for access to other services (like buildkit). Another option is to use the users's private/public keys for this.  what do you think?


rberrelleza:
> I wouldn't show any output related to SSH, I think it is an implementation detail, but we can do that later when we consolidate the ssh implementation.

That's somethign that I've been thinking about as well. We need to show output because generating those keys takes about 10 seconds. Maybe change the text to just talk about generating key pair?

pchico83:
@rberrelleza yes, for now I would go with this implementation. in the future, we can abstract SSH away and don't show anything ssh related in the output

rberrelleza:
@pchico83 @rlamana what do you think of the message? 
image

pchico83:
> @pchico83 @rlamana what do you think of the message?
> image

Love it!

PR #786, opened 2020-03-27
check for null before dereferencing

### LINKED ISSUE #785: Changing storageClass in okteto.yml after "okteto up" raises SIGSEGV
**Describe the bug**
The error message raised when a change to `persistentVolume.storageClass` in `okteto.yml` is made, is not very helpful to understand the issue. I assume this is a problem because changing storageClass is not something that is supported on the PVC resource (perhaps it should be destroyed and recreated, or maybe it's enough to catch the error a bit earlier before it raises sigsegv and simply tell the user what's about to go wrong and how they can fix it if possible.)

Edit: I understand that `okteto down -v` is already available that will delete the PVC, so it makes sense that we can point the user to that feature if needed to explain how to recover this error.

**To Reproduce**
1. Get the ruby-getting-started repo and run "okteto up" as per the guide
2. Update okteto.yml to add a setting for `persistentVolume.storageClass` that is something other than the default storage class, and set `persistentVolume.enabled` as `true`
3. run `okteto up` again, or run `okteto down` and `okteto up` again. Same result both ways.
4. See error:

```panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x22a2c5e]

goroutine 67 [running]:
github.com/okteto/okteto/pkg/k8s/volumes.checkPVCValues(0xc0017ba1c0, 0xc000f60000, 0x12, 0x0)
	github.com/okteto/okteto/pkg/k8s/volumes/crud.go:68 +0x21e
github.com/okteto/okteto/pkg/k8s/volumes.Create(0x2fb2d60, 0xc000f70b00, 0xc000f60000, 0xc0014b9600, 0x4454f00, 0x1094300)
	github.com/okteto/okteto/pkg/k8s/volumes/crud.go:45 +0x301
github.com/okteto/okteto/cmd.(*UpContext).devMode(0xc000f6a090, 0xc0003c1200, 0x0, 0x0, 0x0)
	github.com/okteto/okteto/cmd/up.go:386 +0x169
github.com/okteto/okteto/cmd.(*UpContext).Activate(0xc000f6a090, 0x0)
	github.com/okteto/okteto/cmd/up.go:250 +0x353
created by github.com/okteto/okteto/cmd.RunUp
	github.com/okteto/okteto/cmd/up.go:169 +0x1a6
```

**Expected behavior**
Any descriptive error message would be better for UX than sigsegv. My proposed solution is to tell the user that an update to persistentVolume.storageClass is not supported, and perhaps to suggest with a link to some related docs that they delete the PVC so a new one can be created.

It is possible that one day (in a future k8s release?) you will be able to just go ahead and update persistent volume storage class and the right thing happens, like an automatic transparent migration of data from one storage class to another. I haven't thought too hard about that, but it would be good if however the solution to this SIGSEGV turns out, it doesn't block that potential future change. (I don't know if that will ever happen or if it's even possible.)

 - OS: MacOS
 - Version 10.14.6 (Mojave)

**Additional context**
My okteto.yml:

```
name: hello-world
image: okteto/hello-world:ruby-dev
command: [ "bash" ]
workdir: /opt/app/
forward:
  - 8080:8080
  - 1234:1234
persistentVolume:
  enabled: true
  storageClass: local-path
```

kingdonb:
Fixes #785, I think! Not sure if it is needed to check the inverse, if
dev.PersistentVolumeStorageClass is unset I think it will be fine
(because it's not being dereferenced here.)

## Proposed changes
- Check for null value before dereferencing `pvc.Spec.StorageClassName`
- If null is found there, display the same error that would show as if the field is blank rather than missing.

## Resulting output:

First the error, from okteto cli at version 1.8.1, then ~bin/okteto is the updated version at `okteto version a199e14`

```
kbarret8@kbarret8-mbp:~/projects/ruby-getting-started (master * u=)$ okteto up
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x22a2c5e]

goroutine 35 [running]:
github.com/okteto/okteto/pkg/k8s/volumes.checkPVCValues(0xc00020afc0, 0xc0008d2000, 0x12, 0x0)
	github.com/okteto/okteto/pkg/k8s/volumes/crud.go:68 +0x21e
github.com/okteto/okteto/pkg/k8s/volumes.Create(0x2fb2d60, 0xc0004d59c0, 0xc0008d2000, 0xc0014fe000, 0x4454f00, 0x1094300)
	github.com/okteto/okteto/pkg/k8s/volumes/crud.go:45 +0x301
github.com/okteto/okteto/cmd.(*UpContext).devMode(0xc0008dc090, 0xc0002d6900, 0x0, 0x0, 0x0)
	github.com/okteto/okteto/cmd/up.go:386 +0x169
github.com/okteto/okteto/cmd.(*UpContext).Activate(0xc0008dc090, 0x0)
	github.com/okteto/okteto/cmd/up.go:250 +0x353
created by github.com/okteto/okteto/cmd.RunUp
	github.com/okteto/okteto/cmd/up.go:169 +0x1a6
kbarret8@kbarret8-mbp:~/projects/ruby-getting-started (master * u=)$ ~/bin/okteto up
Installing dependencies...
syncthing-macos-amd64-v1.4.0.tar.gz 9.16 MiB / 9.16 MiB [-----------------------------------------------------------------------------------------------------] 100.00% 3.42 MiB p/s
 ✓  Dependencies successfully installed
 x  couldn't activate your development environment: current okteto volume storageclass is '' instead of 'local-path'. Run 'okteto down -v' and try again
kbarret8@kbarret8-mbp:~/projects/ruby-getting-started (master * u=)$ okteto down -v
 ✓  Development environment deactivated
 ✓  Persistent volume removed

kbarret8@kbarret8-mbp:~/projects/ruby-getting-started (master * u=)$ ~/bin/okteto up
 ✓  Development environment activated
 ✓  Files synchronized
    Namespace: kingdonb
    Name:      hello-world
    Forward:   1234 -> 1234
               8080 -> 8080
```


kingdonb:
I just checked what happens if you remove storageClass setting from okteto.yml, it doesn't do anything. Maybe it should complain in a reverse fashion (`current okteto volume storageclass is '%s' instead of ''`)

I'll overwrite the commit that I provided with a signed-off copy and someone can review this

rberrelleza:
Thanks for the contribution @kingdonb!

rberrelleza:
failure is unrelated to this PR, I'll merge and fix the core issue on a separate branch.

PR #788, opened 2020-03-27
Add azure auth provider

### LINKED ISSUE #787: Support "azure" auth provider
Hello Okteto team!

I ran into the following issue when using Okteto when the auth-provider in my kubeconfig was set
to use "azure"
```
okteto up
 x  no Auth Provider found for name "azure"
```
I did some digging, and it seems that including the following in `main.go` imports should do the trick
```
...
_ "k8s.io/client-go/plugin/pkg/client/auth/azure"
...
```

Would love to see support for this in the official build, 

Thank you!

tommyto-whs:
Attempts to fix #787

## Proposed changes
- Adds the azure auth plugin for `k8s.io/client-go` to `main.go`

## Resulting Output
- kubeconfig set to use the azure auth provider
```
okteto up 
... 
expected okteto output
...
```


rberrelleza:
This is great, thanks for the contribution @tommyto-whs !

PR #791, opened 2020-03-27
remove the race detector from CI

rberrelleza:


rberrelleza:
Let's do this while we investigate the CI issues.

PR #805, opened 2020-04-10
e2e test for build, push, stack deploy and stack destroy

### LINKED ISSUE #803: Create e2e test for okteto stack desploy(--build)/destroy
This way we test stacks, build, push, and Okteto Build/Registry

Implement before next release

pchico83:

Fixes #803

### DIFF
--- a/integration/integration_test.go
+++ b/integration/integration_test.go
@@ -59,6 +59,10 @@ type deployment struct {
 }
 
 const (
+	stackGitRepo     = "git@github.com:okteto/stacks-getting-started.git"
+	stackGitFolder   = "stacks-getting-started"
+	stackManifest    = "okteto-stack.yml"
+	stackName        = "voting-app"
 	deploymentFormat = `
 apiVersion: apps/v1
 kind: Deployment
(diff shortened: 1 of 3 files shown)


integration/integration_test.go
```diff
@@ -186,6 +190,106 @@ func TestDownloadSyncthing(t *testing.T) {
 	}
 }
 
+func TestStacks(t *testing.T) {
+	if mode == "client" {
+		return
```
rberrelleza, line 195: Skip the test instead so it doesn't show as passing.

integration/integration_test.go
```diff
@@ -186,6 +190,106 @@ func TestDownloadSyncthing(t *testing.T) {
...
+	name := strings.ToLower(fmt.Sprintf("%s-%d", tName, time.Now().Unix()))
+	namespace := fmt.Sprintf("%s-%s", name, user)
+
+	test := scopeagent.GetTest(t)
+	test.Run(tName, func(t *testing.T) {
+		if err := createNamespace(ctx, oktetoPath, namespace); err != nil {
+			t.Fatal(err)
+		}
+
+		defer deleteNamespace(ctx, oktetoPath, namespace)
```
rberrelleza, line 214: Could we not delete it if it fails? It's helpful for debugging. We can have a once week job that deleted any leftover namespaces.

integration/integration_test.go
```diff
@@ -186,6 +190,106 @@ func TestDownloadSyncthing(t *testing.T) {
 	}
 }
 
+func TestStacks(t *testing.T) {
```
rberrelleza, line 193: Can you put it in a separate file in the same folder for clarity? (Make sure to copy the tags defined at the top of the file)

integration/integration_test.go
```diff
@@ -318,12 +425,10 @@ func waitForDeployment(ctx context.Context, name string, revision, timeout int)
...
 }
 
-func getContent() (string, error) {
-	endpoint := "http://localhost:8080/index.html"
+func getContent(endpoint string) (string, error) {
 	retries := 0
-
 	t := time.NewTicker(1 * time.Second)
-	for i := 0; i < 60; i++ {
+	for i := 0; i < 120; i++ {
```
rberrelleza, line 431: Two minutes to get the content is a lot no?  Maybe pass this as a param so the `up` test fails sooner?

integration/integration_test.go
```diff
@@ -186,6 +190,106 @@ func TestDownloadSyncthing(t *testing.T) {
...
+
+		if err := cloneGitRepo(ctx, stackGitRepo); err != nil {
+			t.Fatal(err)
+		}
+		defer deleteGitRepo(ctx, stackGitFolder)
+		if err := deployStack(ctx, oktetoPath, stackManifest); err != nil {
+			t.Fatal(err)
+		}
+		endpoint := fmt.Sprintf("https://vote-%s.cloud.okteto.net", namespace)
+		content, err := getContent(endpoint)
```
rberrelleza, line 224: What about the rest of the services? I think we should also check that all the resources are created as expected as part of this test.

integration/integration_test.go
```diff
@@ -186,6 +190,106 @@ func TestDownloadSyncthing(t *testing.T) {
...
+		if err != nil {
+			t.Fatalf("failed to get stack content: %s", err)
+		}
+		if !strings.Contains(content, "Cats vs Dogs!") {
+			t.Fatalf("wrong stack content: %s", content)
+		}
+		if err := destroyStack(ctx, oktetoPath, stackName); err != nil {
+			t.Fatal(err)
```
rberrelleza, line 232: The function should also check that the resources are scheduled for deletion.

PR #818, opened 2020-04-14
Support for  auto-login in "okteto build/push/stack deploy"

pchico83:

## Proposed changes
- This makes it simple to use the okteto cli in dev environments and CI jobs


cmd/build.go
```diff
@@ -37,6 +39,10 @@ func Build() *cobra.Command {
 		RunE: func(cmd *cobra.Command, args []string) error {
 			log.Debug("starting build command")
 
+			if err := login.WithEnvVar(ctx); err != nil {
+				return err
+			}
```
rberrelleza, line 44: Add a specif error that if failed when auto-logging in or something?

cmd/build.go
```diff
@@ -37,6 +39,10 @@ func Build() *cobra.Command {
 		RunE: func(cmd *cobra.Command, args []string) error {
 			log.Debug("starting build command")
 
+			if err := login.WithEnvVar(ctx); err != nil {
```
rberrelleza, line 42:
```suggestion
			if err := login.WithEnvVarIfAvailable(ctx); err != nil {
```

PR #821, opened 2020-04-15
Stop scope agent after CLI span is finished

drodriguezhdez:
Signed-off-by: Daniel Rodriguez Hernandez daniel.rguez.hdez.github@gmail.com

## Proposed changes
- The span should be finished before agent stops.


pchico83:
@drodriguezhdez thanks!

PR #828, opened 2020-04-19
Fix "--namespace" option for stacks

### LINKED ISSUE #822: When creating a stack, the  --namespace parameter is not respected
**Describe the bug**
The --namespace parameter in stacks is not respected

**To Reproduce**
1. Setup your context for your personal namespace
1. Create a second namespace (e.g. stack)
1. okteto stack create --namespace=stack

image

 - OS: macOS
 - Version 1.8.4

pchico83:
Fixes #822

### DIFF
--- a/pkg/cmd/stack/deploy.go
+++ b/pkg/cmd/stack/deploy.go
@@ -35,6 +35,9 @@ import (
 
 //Deploy deploys a stack
 func Deploy(ctx context.Context, s *model.Stack, forceBuild, wait bool) error {
+	if s.Namespace != "" {
+		os.Setenv("HELM_NAMESPACE", s.Namespace)
+	}
 	settings := cli.New()
 	if s.Namespace == "" {
 		s.Namespace = settings.Namespace()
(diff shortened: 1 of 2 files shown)


pkg/cmd/stack/deploy.go
```diff
@@ -35,6 +35,9 @@ import (
 
 //Deploy deploys a stack
 func Deploy(ctx context.Context, s *model.Stack, forceBuild, wait bool) error {
+	if s.Namespace != "" {
```
pchico83, line 38: I try to initialize the value in the settings struct, but it is private, and it only gets the value from this env variable:
https://github.com/helm/helm/blob/master/pkg/cli/environment.go#L68

rberrelleza: We are depending on an internal implementation for this. Can we add a test to prevent regressions when updating the helm package?

pchico83:
Fixed by https://github.com/okteto/charts/pull/54

PR #829, opened 2020-04-19
Autogenerate stack name if it is not provided

### LINKED ISSUE #827: Stack name should be optional
Stack name should be optional

The default value can be the folder's name, as docker-compose does.

pchico83:

Fixes #827


pkg/model/utils_test.go
```diff
@@ -80,3 +80,29 @@ func TestFileExists(t *testing.T) {
 		t.Errorf("fail to detect existing file")
 	}
 }
+
+func Test_GetValidNameFromFolder(t *testing.T) {
```
rberrelleza, line 84: Can you test what happens when the folder is `.`, or `../` ?

Issue #700, opened 2020-02-19
Support for --build-arg with okteto build


tullo:
**Is your feature request related to a problem? Please describe.**

Okteto build should accept build args so it can be used as a replacement for docker build

**Describe the solution you'd like**

Support for `--build-arg` with okteto build

**Additional context**
My current way of building images
```
	docker build \
		-f dockerfile.sales-api \
		-t gcr.io/$(PROJECT)/sales-api-amd64:$(VERSION) \
		--build-arg PACKAGE_NAME=sales-api \
		--build-arg VCS_REF=`git rev-parse HEAD` \
		--build-arg BUILD_DATE=`date -u +”%Y-%m-%dT%H:%M:%SZ”` \
		.
	docker image tag \
		gcr.io/$(PROJECT)/sales-api-amd64:$(VERSION) \
		$(REGISTRY_ACCOUNT)/sales-api-amd64:$(VERSION)
```

tullo:
`okteto push` should also accept --build-arg options. Otherwise it is still a two-step process `okteto build` and `kubectl apply` for redeployments.

tullo:
`okteto push` should also accept --build-arg options. Otherwise it is still a two-step process `okteto build` and `kubectl apply` for redeployments.

tullo:
`okteto push` should also accept --build-arg options. Otherwise it is still a two-step process `okteto build` and `kubectl apply` for redeployments.

PR #832, opened 2020-04-22
Support for setting context, dockerfile, target and build args in "ok…

pchico83:
…teto push"

Fixes https://github.com/okteto/okteto/issues/700#issuecomment-617260201

### DIFF
--- a/pkg/model/dev.go
+++ b/pkg/model/dev.go
@@ -114,6 +114,11 @@ type Dev struct {
 
 // BuildInfo represents the build info to generate an image
 type BuildInfo struct {
+	BuildInfoRaw
+}
+
+// BuildInfoRaw represents the build info for serialization
+type BuildInfoRaw struct {
 	Context    string   `yaml:"context,omitempty"`
 	Dockerfile string   `yaml:"dockerfile,omitempty"`
 	Target     string   `yaml:"target,omitempty"`
--- a/pkg/model/dev.go
+++ b/pkg/model/dev.go
@@ -91,6 +91,7 @@ type Dev struct {
 	Namespace            string                `json:"namespace,omitempty" yaml:"namespace,omitempty"`
 	Container            string                `json:"container,omitempty" yaml:"container,omitempty"`
 	Image                string                `json:"image,omitempty" yaml:"image,omitempty"`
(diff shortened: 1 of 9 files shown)


pkg/model/dev.go
```diff
@@ -91,6 +91,7 @@ type Dev struct {
 	Namespace            string                `json:"namespace,omitempty" yaml:"namespace,omitempty"`
 	Container            string                `json:"container,omitempty" yaml:"container,omitempty"`
 	Image                string                `json:"image,omitempty" yaml:"image,omitempty"`
+	Push                 *BuildInfo            `json:"-,omitempty" yaml:"push,omitempty"`
```
rberrelleza, line 94: Yes!!!!

should we call it `build` since it's buildinfo?

pkg/model/dev.go
```diff
@@ -111,6 +112,14 @@ type Dev struct {
 	SSHServerPort        int                   `json:"sshServerPort,omitempty" yaml:"sshServerPort,omitempty"`
 }
 
+// BuildInfo represents the build info to generate an image
+type BuildInfo struct {
```
rberrelleza, line 116: Will this be overridable via the CLI?

pkg/model/dev.go
rberrelleza: and could we also use it in `build`?

pchico83: I talked to @rlamana  and the idea is to have a `build` field for building the dev image. `okteto up --build` would build your dev image on the fly. This is why we chose `push` here.

pkg/model/dev.go
pchico83: I can add it as part of this PR

pkg/model/dev.go
rberrelleza: cool

pkg/model/dev.go
rberrelleza: only if we have a customer ask, if not, let's keep it simple and wait?

pchico83: Yeap, that is why I didn't add them.
Also, those parameters shouldn't change in the okteto workflow

PR #836, opened 2020-04-23
use clean instead of ps

### LINKED ISSUE #835: Remove dependency on "ps"
**Describe the bug**
When "okteto up" reconnects to a running dev environment, it kills previous program executions that might have been left zombie. This login depends on "ps" being installed in the image, which is not always the case.

**Expected behavior**
The logic should work regarding "ps" being installed or not. Let's move this login to our golang portable binaries to remove the dependency on "ps".

rberrelleza:

Fixes #835 

## Proposed changes
- Use https://github.com/okteto/clean instead of `ps` to clean the session on up

### DIFF
--- a/cmd/up.go
+++ b/cmd/up.go
@@ -714,7 +714,7 @@ func (up *UpContext) cleanCommand() {
 	in := strings.NewReader("\n")
 	var out bytes.Buffer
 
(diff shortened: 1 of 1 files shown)


cmd/up.go
```diff
@@ -716,7 +716,7 @@ func (up *UpContext) cleanCommand() {
 		in,
 		&out,
 		os.Stderr,
-		[]string{"sh", "-c", "(((cp /var/okteto/bin/* /usr/local/bin); (ps -ef | grep -v -E '/var/okteto/bin/syncthing|/var/okteto/bin/remote|PPID' | awk '{print $2}' | xargs -r kill -9)) >/dev/null 2>&1); cat /proc/sys/fs/inotify/max_user_watches"},
+		[]string{"sh", "-c", "(((cp /var/okteto/bin/* /usr/local/bin); (/var/okteto/bin/clean)) >/dev/null 2>&1); cat /proc/sys/fs/inotify/max_user_watches"},
```
pchico83, line 719: I would move the full line to `clean` to simplify this code.
Also, if `clean` is not present, we should execute the previous command

PR #839, opened 2020-04-25
Improve error messages for non root containers

### LINKED ISSUE #696: "okteto up" should ignore the some securityGroup fields
**Describe the bug**
My deployment is using:
```
     securityContext:
        runAsUser: 1000
        runAsGroup: 1000
        fsGroup: 1000
```
and my dev image is root. `okteto up` fails with permission errors.

**Expected behavior**
It should work. "okteto up" should unset:
```
     securityContext:
        runAsUser: 1000
        runAsGroup: 1000
        fsGroup: 1000
```

pchico83:
Fixes #696

### DIFF
--- a/pkg/k8s/pods/pod.go
+++ b/pkg/k8s/pods/pod.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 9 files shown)


pkg/k8s/pods/pod.go
```diff
@@ -212,11 +213,58 @@ func Exists(podName, namespace string, c kubernetes.Interface) bool {
 	return pod.GetObjectMeta().GetDeletionTimestamp() == nil
 }
 
-//ContainerLogs returns the logs of a given container in a given pod
-func ContainerLogs(container string, pod *apiv1.Pod, namespace string, c kubernetes.Interface) (string, error) {
+//GetDevPodUserID returns the user id running the dev pod
+func GetDevPodUserID(ctx context.Context, dev *model.Dev, c *kubernetes.Clientset) int64 {
```
rberrelleza, line 217: could we do a `kubectl exec echo $USER` instead?

pkg/k8s/pods/pod.go
```diff
@@ -212,11 +213,58 @@ func Exists(podName, namespace string, c kubernetes.Interface) bool {
...
+	if err != nil {
+		log.Errorf("failed to access development environment logs: %s", err)
+		return -1
+	}
+	return parseUserID(devPodLogs)
+}
+
+func parseUserID(output string) int64 {
+	lines := strings.Split(output, "\n")
+	for _, line := range lines {
```
rberrelleza, line 228: this depends on `start.sh` logging a specific thing. Since that's in a different repo, it's going to be hard to keep them in sync. Could we add a test or something so that if we  remove that log by accident our CI will catch it?

pkg/model/dev.go
```diff
@@ -331,6 +334,24 @@ func (dev *Dev) setDefaults() error {
...
 }
 
+func (dev *Dev) setRunAsUserDefaults(main *Dev) {
+	if !main.PersistentVolumeEnabled() {
+		return
+	}
+	if dev.SecurityContext == nil {
+		dev.SecurityContext = &SecurityContext{}
+	}
+	if dev.SecurityContext.RunAsUser == nil {
```
rberrelleza, line 344: what's the purpose of this? I'd rather not set something that the user is not setting herself, to limit the stuff we change.

pkg/k8s/pods/pod.go
pchico83: no, at that time, the pod might not be running

pkg/k8s/pods/pod.go
pchico83: I think we should move the bin image to this repo, and bin-cloud to okteto app

pkg/model/dev.go
pchico83: it is one of the main things this issue is about. Currently, we document that root images work by default. if the user wants to use a non root dev image, `runAsUser` must be defined.
But the current behavior is broken, if `runAsUser` is set in the k8 manifest, a root dev image does not work. several users have been reported this issue

PR #848, opened 2020-04-30
forward the SSH_AUTH_SOCK in Go client

### LINKED ISSUE #672: Use DockerFile as base image
I gave Okteto a try with the ASPNetCore example. It worked like a charm. Next step was to try it with our own solution. During that attempt I noticed that the OS my microservice was running in, was based upon some Microsoft SDK docker image. We have had issues in the past, where 1 image produced slightly different results than another image.

We want to be as close as possible to the real production environment. This includes having the same docker base image for working with Okteto as the production environment would be great. So if possible I would prefer a way to enrich the DockerFile belonging to the (ASPNetCore) project in such way that it can work with Okteto.

Could not think of an alternative solution to this problem

### LINKED ISSUE #845: the shell started by `okteto up` doesn't forward the SSH agent
**Describe the bug**
The shell started by `okteto up` doesn't forward the SSH agent

**To Reproduce**

1. Ensure you are forwarding keys (run ssh-add -L in macos)
2. Check that your SSH agent is running (printenv SSH_AUTH_SOCK)
3. `export OKTETO_EXECUTE_SSH=1`
4. `okteto up`
5. `okteto>ssh -T git@github.com`

**Expected behavior**
My key should be forwarded in the main session.  

 - OS: macos
 - Version 1.8.4

**Additional context**
Add any other context about the problem here (e.g okteto.yml, k8s manifests, logs)

unthought:
Fixes #845

## Proposed changes
- forward the ssh agent socket found at `SSH_AUTH_SOCK` (if set) to remote, during `okteto up` and `okteto exec`.
- currently if the `SSH_AUTH_SOCK` env var is set, it's forwarded unconditionally. We might want to allow the user to disable this.

### DIFF
--- a/pkg/ssh/exec.go
+++ b/pkg/ssh/exec.go
@@ -17,11 +17,13 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"os"
 	"strings"
 	"time"
 
 	"github.com/okteto/okteto/pkg/log"
 	"golang.org/x/crypto/ssh"
+	"golang.org/x/crypto/ssh/agent"
 	"golang.org/x/crypto/ssh/terminal"
 )
 
@@ -82,6 +84,18 @@ func Exec(ctx context.Context, remotePort int, tty bool, inR io.Reader, outW, er
 
 		if err := session.RequestPty("xterm", height, width, modes); err != nil {
 			return fmt.Errorf("request for pseudo terminal failed: %s", err)
+		}
+	}
+
+	sockEnvVar, ok := os.LookupEnv("SSH_AUTH_SOCK")
+	if !ok {
+		log.Info("SSH_AUTH_SOCK is not set, not forwarding socket")
+	} else {
+		if err := agent.ForwardToRemote(connection, sockEnvVar); err != nil {
+			log.Infof("failed to existing SSH_AUTH_SOCK('%s'): %s", sockEnvVar, err)
+		}
+		if err := agent.RequestAgentForwarding(session); err != nil {
+			log.Infof("failed to forward ssh agent to remote: %s", err)
 		}
 	}
 


pkg/ssh/exec.go
```diff
@@ -17,16 +17,18 @@ import (
...
 
 	"github.com/okteto/okteto/pkg/log"
 	"golang.org/x/crypto/ssh"
+	"golang.org/x/crypto/ssh/agent"
 	"golang.org/x/crypto/ssh/terminal"
 )
 
 // Exec executes the command over SSH
-func Exec(ctx context.Context, remotePort int, tty bool, inR io.Reader, outW, errW io.Writer, command []string) error {
+func Exec(ctx context.Context, remotePort int, tty bool, agentFwd bool, inR io.Reader, outW, errW io.Writer, command []string) error {
```
rberrelleza, line 31: Since both callers of ssh.Exec are passing true, I think it's better to not even have the agentFwd argument here.

pkg/ssh/exec.go
```diff
@@ -85,6 +87,20 @@ func Exec(ctx context.Context, remotePort int, tty bool, inR io.Reader, outW, er
 		}
 	}
 
+	if agentFwd {
+		sockEnvVar, ok := os.LookupEnv("SSH_AUTH_SOCK")
+		if !ok {
+			log.Info("SSH_AUTH_SOCK is not set, not forwarding socket")
+		} else {
+			if err := agent.ForwardToRemote(connection, sockEnvVar); err != nil {
+				log.Error(err)
```
rberrelleza, line 96: since the execution will keep going, could you make this a `log.Infof` instead? And also add a message for context? Maybe something like `log.Infof("failed to forward ssh agent to remote: %w", err)`. Same with the one below.

unthought: If I do this, `go vet` complains with:
```
Infof call has error-wrapping directive %w
```
any suggestions?

rberrelleza: use the `%s` directive instead.

PR #853, opened 2020-05-01
Set healthchecks based on "healthchecks" field

### LINKED ISSUE #849: Disable healthchecks for "services" on "okteto up"
Using `services` with healthchecks is a problem. The pod takes time to pass the healthcheck everytime I run `okteto restart`

Disable healthchecks on `okteto up` for `services` as it is done for the main dev container.

**Additional context**
A field `disableHealthchecks` with default value `true`. There are valid use cases where you want to keep healthchecks, but it does not look as the more general use case.

pchico83:
Fixes #849

### DIFF
--- a/pkg/model/dev.go
+++ b/pkg/model/dev.go
@@ -99,6 +99,7 @@ type Dev struct {
 	Environment          []EnvVar              `json:"environment,omitempty" yaml:"environment,omitempty"`
 	Secrets              []Secret              `json:"secrets,omitempty" yaml:"secrets,omitempty"`
 	Command              []string              `json:"command,omitempty" yaml:"command,omitempty"`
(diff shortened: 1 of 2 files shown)


pkg/model/dev.go
```diff
@@ -99,6 +99,7 @@ type Dev struct {
 	Environment          []EnvVar              `json:"environment,omitempty" yaml:"environment,omitempty"`
 	Secrets              []Secret              `json:"secrets,omitempty" yaml:"secrets,omitempty"`
 	Command              []string              `json:"command,omitempty" yaml:"command,omitempty"`
+	Healthchecks         bool                  `json:"healthchecks,omitempty" yaml:"healthchecks,omitempty"`
```
rberrelleza, line 102: would this make them on or off by default?

pchico83: it disables them by default

PR #859, opened 2020-05-02
Add okteto restart for a single service

### LINKED ISSUE #510: do `okteto restart` on a single service
Sometimes I'd like to restart a single service, instead of all of them

`okteto restart api` 

Right now I do `okteto restart` but it takes longer because it has to restart multiple services

adhaamehab:
Signed-off-by: adhaamehab <adhaamehab.me@gmail.com>

Fixes #510 

## Proposed changes
- Added a new argument to `okteto restart` which represents single service name to restart. 

- Command will fail with error message `Unable to find any service with the provided name` if the provided service name doesn't match any name in the services section in `okteto.yml`

- If no argument provided all pods will be restarted (normal behaviour)

### DIFF
--- a/pkg/k8s/pods/pod.go
+++ b/pkg/k8s/pods/pod.go
@@ -284,7 +284,7 @@ func containerLogs(container string, pod *apiv1.Pod, namespace string, timestamp
 }
 
 // Restart restarts the pods of a deployment
-func Restart(dev *model.Dev, c *kubernetes.Clientset) error {
+func Restart(dev *model.Dev, c *kubernetes.Clientset, sn string) error {
 	pods, err := c.CoreV1().Pods(dev.Namespace).List(
 		metav1.ListOptions{
 			LabelSelector: fmt.Sprintf("%s=%s", okLabels.DetachedDevLabel, dev.Name),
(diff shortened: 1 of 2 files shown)


pkg/k8s/pods/pod.go
```diff
@@ -295,16 +295,24 @@ func Restart(dev *model.Dev, c *kubernetes.Clientset) error {
...
 	}
 
+	found := false
 	for i := range pods.Items {
-		err := c.CoreV1().Pods(dev.Namespace).Delete(pods.Items[i].Name, &metav1.DeleteOptions{GracePeriodSeconds: &devTerminationGracePeriodSeconds})
-		if err != nil {
-			if strings.Contains(err.Error(), "not found") {
-				return nil
+
+		if sn == "" || strings.HasPrefix(pods.Items[i].Name, fmt.Sprintf("%s-", sn)) {
```
rberrelleza, line 301: I'd build the prefix string once before the loop, since it doesn't change.

rberrelleza: For the condition check, I think it's more readable if you check if the pod should be skipped, and do a `continue`, instead of having a longer body inside the `if`.

adhaamehab: Yeah that makes more sense!

rberrelleza:
Thanks for the contribution @adhaamehab!

Issue #874, opened 2020-05-12
Support for env vars with default values in the image


rberrelleza:
**Describe the bug**
I have this definition in my stack

```
db:
    image: postgres:${POSTGRES_VERSION:-9.6.15-alpine}
    environment:
      - POSTGRES_USER=devto
      - POSTGRES_PASSWORD=devto
      - POSTGRES_DB=PracticalDeveloper_development
    ports:
      - 5432
```

When deployed, the image was `postgres:`. It would be nice if to have support for default values in the env vars, like in docker-compose.

**Desktop (please complete the following information):**
 - OS: macos
 - Version: 1.8.8

PR #881, opened 2020-05-18
Support for variable expansion in the "name" field

### LINKED ISSUE #880: Support for generateName
I often have to create multiple development environments at the same time, but have to change the name in the yml to avoid collision

In k8s I use 
```
metadata:
  generateName: basename-
```

Which adds a uuid to the end of basename to avoid collisions

pchico83:
Fixes #880

### DIFF
--- a/pkg/model/dev.go
+++ b/pkg/model/dev.go
@@ -252,8 +252,12 @@ func Read(bytes []byte) (*Dev, error) {
 		}
 	}
 
-	dev.loadImage()
+	dev.loadName()
+	for _, s := range dev.Services {
+		s.loadName()
+	}
 
+	dev.loadImage()
 	for _, s := range dev.Services {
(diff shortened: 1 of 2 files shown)


pkg/model/dev.go
```diff
@@ -273,6 +277,12 @@ func Read(bytes []byte) (*Dev, error) {
 	return dev, nil
 }
 
+func (dev *Dev) loadName() {
+	if len(dev.Name) > 0 {
+		dev.Name = os.ExpandEnv(dev.Name)
```
rberrelleza, line 282: will this support env vars with a  default value ? `${foo-NAME}`  kind of thing.

pchico83:
@rberrelleza let's track that as part of https://github.com/okteto/okteto/issues/874

PR #889, opened 2020-06-03
move okteto/bin and okteto/okteto to the cli repo

rberrelleza:
## Proposed changes
- Move the okteto/bin and okteto/okteto docker files to the CLI repo, since they all depend on each other 

As a follow up, we can update our release job so everything is generated  together.


rberrelleza:
The dockerfile generates okteto/bin and okteto/okteto (which is currently consumed by the github actions), I put it in root so it can add the source. 

We could back to instead building separately and just ADD the binary into the container. what do you think?

pchico83:
@rberrelleza LGTM

PR #893, opened 2020-06-04
Support for setting dev time tolerations

### LINKED ISSUE #888: Support tolerations in okteto manifest

We taint our EKS nodes so that we can minimise the work scheduled on nodes with GPUs to workloads that require GPUs. We are using okteto to set up dev environments for machine learning researchers who sometimes need access to GPUs to train models. 
A deployment for them would require the following toleration:
```
      tolerations:
        - key: nvidia.com/gpu
          operator: Exists
```
it would be much nicer for our users to be able to define this block in the okteto file - they are not regular kubernetes users and just one step to set up their dev environment rather than two would be easier.

okteto file supports tolerations blocks in kubernetes syntax 

We are solving the problem at the moment by creating a deployment with kubernetes manifests, then writing an okteto file to attach to that deployment. As mentioned above, this extra step is potentially confusing for our users.

**Additional context**

pchico83:

Fixes #888

### DIFF
--- a/pkg/k8s/deployments/crud.go
+++ b/pkg/k8s/deployments/crud.go
@@ -103,6 +103,7 @@ func GetTranslations(dev *model.Dev, d *appsv1.Deployment, c *kubernetes.Clients
 			Version:     model.TranslationVersion,
 			Deployment:  d,
 			Annotations: dev.Annotations,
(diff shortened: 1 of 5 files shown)


pchico83:
Also fixes integration tests

PR #899, opened 2020-06-04
Set replicas to 1 on push

rberrelleza:
## Proposed changes
- Set the replicas to 1 when push if they are zero
- Breakup function into smaller bits

### DIFF
--- a/cmd/push.go
+++ b/cmd/push.go
@@ -31,7 +31,6 @@ import (
 	"github.com/okteto/okteto/pkg/model"
 	"github.com/okteto/okteto/pkg/okteto"
 	"github.com/spf13/cobra"
-	appsv1 "k8s.io/api/apps/v1"
 	"k8s.io/client-go/kubernetes"
 )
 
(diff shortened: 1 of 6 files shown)


cmd/push.go
```diff
@@ -158,52 +166,58 @@ func runPush(dev *model.Dev, autoDeploy bool, imageTag, oktetoRegistryURL, progr
...
-	if imageDigest != "" {
-		imageWithoutTag := build.GetRepoNameWithoutTag(imageTag)
-		imageTag = fmt.Sprintf("%s@%s", imageWithoutTag, imageDigest)
-	}
-
 	spinner := utils.NewSpinner(fmt.Sprintf("Pushing source code to the development environment '%s'...", dev.Name))
 	spinner.Start()
 	defer spinner.Stop()
+
 	if create {
```
pchico83, line 178: Is this recreating the service?
I think it is not. It could work if you check the deployment autocreate annotation instead of the `create` envvar.

PR #910, opened 2020-06-12
Rename "development environment" to "development container"

pchico83:

## Proposed changes
- Rename "development environment" to "development container".  The development environment is a wider concept.

### DIFF
--- a/cmd/restart.go
+++ b/cmd/restart.go
@@ -31,7 +31,7 @@ func Restart() *cobra.Command {
 
 	cmd := &cobra.Command{
 		Use:   "restart",
-		Short: "Restarts the pods of your development environment",
+		Short: "Restarts the deployment listed in the services field of the okteto manifest",
 		RunE: func(cmd *cobra.Command, args []string) error {
 			dev, err := utils.LoadDev(devPath)
 			if err != nil {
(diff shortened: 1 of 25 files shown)


cmd/restart.go
```diff
@@ -31,7 +31,7 @@ func Restart() *cobra.Command {
 
 	cmd := &cobra.Command{
 		Use:   "restart",
-		Short: "Restarts the pods of your development environment",
+		Short: "Restarts the services of your okteto manifest",
```
rlamana, line 34: This is the only one that sounds odd. The rest of the changes are much more clear now.
What about `Restarts the pods containing your development container`? or `Restart the pod of your development container` or something like that... I'm not sure if my proposal is accurate enough.

pchico83: This command restarts the pods indicated by the `services` field of the okteto manifest. It does not restart the dev container. What about something like:
`Restarts the pods references by the "services" field of the okteto manifest`

rlamana: Sounds more accurate like that then

rberrelleza: In kubernetes you don't restart services, you recreate (or restart) the pods. I think we should use more  kubernetes-specific in this kind of descriptions.

rberrelleza: Maybe: restarts the deployment listed in the `services` field of the okteto manifest?

PR #914, opened 2020-06-12
move build to github action

rberrelleza:
Remove dependencies on 3rd party CI/CD


rberrelleza:
wanted to test a few things. I don't think we gain much by moving TBH.

PR #916, opened 2020-06-14
Accept single string as command in the okteto manifest

pchico83:
Single strings, when possible, improve readability and usability of okteto manifests

### DIFF
--- a/pkg/model/serializer.go
+++ b/pkg/model/serializer.go
@@ -48,6 +48,38 @@ func (e EnvVar) MarshalYAML() (interface{}, error) {
 }
 
 // UnmarshalYAML Implements the Unmarshaler interface of the yaml pkg.
+func (c *Command) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	var multi []string
+	err := unmarshal(&multi)
+	if err != nil {
+		var single string
+		err := unmarshal(&single)
+		if err != nil {
+			return err
+		}
+		if strings.Contains(single, " ") {
+			c.Values = []string{"sh", "-c", single}
+		} else {
+			c.Values = []string{single}
+		}
+	} else {
+		c.Values = multi
+	}
+	return nil
+}
+
+// MarshalYAML Implements the marshaler interface of the yaml pkg.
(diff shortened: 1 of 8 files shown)


pkg/model/serializer.go
```diff
@@ -47,6 +47,38 @@ func (e EnvVar) MarshalYAML() (interface{}, error) {
...
+func (c *Command) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	var multi []string
+	err := unmarshal(&multi)
+	if err != nil {
+		var single string
+		err := unmarshal(&single)
+		if err != nil {
+			return err
+		}
+		if strings.Contains(single, " ") {
```
rberrelleza, line 60: What is this check for? I think doing `strings.Split` would work better here.

pkg/model/serializer.go
```diff
@@ -47,6 +47,38 @@ func (e EnvVar) MarshalYAML() (interface{}, error) {
...
+	}
+	return nil
+}
+
+// MarshalYAML Implements the marshaler interface of the yaml pkg.
+func (c Command) MarshalYAML() (interface{}, error) {
+	if len(c.Values) == 1 && !strings.Contains(c.Values[0], " ") {
+		return c.Values[0], nil
+	}
+	if len(c.Values) == 3 && c.Values[0] == "sh" && c.Values[1] == "-c" && !strings.Contains(c.Values[2], " ") {
```
rberrelleza, line 76: Having the `sh -c` on the command is odd. I think that's something we should add during the execution, but not the manifest.

rberrelleza:
> Single strings, when possible, improve readability and usability of okteto manifests

Don't this break with how docker and kubernetes takes commands though?

pkg/model/serializer.go
pchico83: I don't want to split, it is just a checking to know if it is a single command or not

pkg/model/serializer.go
pchico83: It is the other way around. It is to remove them when marshaling. Although thinking twice, I don't need this even for the `init` command

PR #918, opened 2020-06-14
"okteto init" asking for target deployment

### LINKED ISSUE #907: Next "okteto init" iteration
I want to clone a git repo, run `okteto init` and have my okteto manifests auto-generated.

- `okteto init` receives an optional parameter with the deployment to put on dev mode. If not specified, the current list of deployments is shown. The user picks one, she can also pick an isolated dev environment.
- If it is a new isolated dev environment, the behavior is the same as for today.
- For existing deployments, if it has labels, select it in the okteto manifests based on labels. If not, select it by name.
- Check if there is only one container in the target deployment.
- Infer the language. For python, node, ruby and PHP, use the deployment image. Try to inspect the deployment image to get the start command and workdir.
- For Go and Java, use the okteto dev image. We could ask for the `Dockerfile` and add the `build` field to use `okteto up --build`.
- Infer port forwards from associated service.
- If deployed by Flux, add flux ignore annotation.
Infer `.stignore` from `.dockerignore` or `.gitignore`. If none is present, use the current defaults.

pchico83:

Fixes #907

### DIFF
--- a/cmd/init.go
+++ b/cmd/init.go
@@ -17,12 +17,17 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
-	"strings"
 
+	"github.com/okteto/okteto/cmd/utils"
 	"github.com/okteto/okteto/pkg/analytics"
+	initCMD "github.com/okteto/okteto/pkg/cmd/init"
+	k8Client "github.com/okteto/okteto/pkg/k8s/client"
+	"github.com/okteto/okteto/pkg/k8s/deployments"
 	"github.com/okteto/okteto/pkg/linguist"
 	"github.com/okteto/okteto/pkg/log"
 	"github.com/okteto/okteto/pkg/model"
+	appsv1 "k8s.io/api/apps/v1"
+	"k8s.io/client-go/kubernetes"
 
 	"github.com/manifoldco/promptui"
 	"github.com/spf13/cobra"
@@ -32,6 +37,7 @@ const (
 	stignore          = ".stignore"
 	defaultManifest   = "okteto.yml"
 	secondaryManifest = "okteto.yaml"
+	createDeployment  = "new deployment"
 )
 
 var wrongImageNames = map[string]bool{
@@ -47,6 +53,7 @@ var wrongImageNames = map[string]bool{
 
 //Init automatically generates the manifest
 func Init() *cobra.Command {
+	var namespace string
 	var devPath string
 	var overwrite bool
 	cmd := &cobra.Command{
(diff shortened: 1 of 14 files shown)


pchico83:
Still requires some more testing, but the code should be ready for review

cmd/init.go
```diff
@@ -59,57 +66,62 @@ func Init() *cobra.Command {
 				return err
 			}
 
-			if err := executeInit(devPath, overwrite, l, workDir); err != nil {
+			if err := executeInit(namespace, devPath, l, workDir, overwrite); err != nil {
 				return err
 			}
 
-			log.Success(fmt.Sprintf("Okteto manifest (%s) created", devPath))
+			log.Success(fmt.Sprintf("Okteto manifest (%s) created. Check its content in case you need to adapt it", devPath))
```
rberrelleza, line 73:
```suggestion
			log.Success(fmt.Sprintf("Okteto manifest (%s) created", devPath))
```

That second sentence sounds bad, let's not have it.

cmd/init.go
```diff
@@ -59,57 +66,62 @@ func Init() *cobra.Command {
...
-	if len(language) > 0 {
-		log.Debugf("generating config for %s", language)
-		dev = linguist.GetDevConfig(language)
-	} else {
-		l, err := linguist.ProcessDirectory(workDir)
-		if err != nil {
-			log.Info(err)
-			return fmt.Errorf("Failed to determine the language of the current directory")
-		}
+	ifAskingForDeployment := language == ""
```
rberrelleza, line 91:
```suggestion
	askForDeployment := language == ""
```

cmd/init.go
```diff
@@ -59,57 +66,62 @@ func Init() *cobra.Command {
...
+		d, container, err := getDeployment(namespace)
+		if err != nil {
+			return err
+		}
+		dev.Container = container
+		if container == "" {
+			container = d.Spec.Template.Spec.Containers[0].Name
+		}
+
+		postfix := fmt.Sprintf("Analyzing deployment '%s'...", d.Name)
```
rberrelleza, line 113: put the spinner before getting the deployment to it's also shown while doing the get

cmd/init.go
```diff
@@ -59,57 +66,62 @@ func Init() *cobra.Command {
...
+		spinner := utils.NewSpinner(postfix)
+		spinner.Start()
+		defer spinner.Stop()
```
rberrelleza, line 116: doing defers with the spinner create odd behaviors in the UI. It's better to start the spinner, call teh function, stop the spinner, and then check for error.

cmd/init.go
```diff
@@ -122,41 +134,122 @@ func executeInit(devPath string, overwrite bool, language string, workDir string
...
 	return nil
 }
 
-func askForImage(language, defaultImage string) string {
-	var image string
-	fmt.Printf("Recommended image for development with %s: %s\n", language, log.BlueString(defaultImage))
-	fmt.Printf("Which docker image do you want to use for your development container? [%s]: ", defaultImage)
-	_, err := fmt.Scanln(&image)
-	fmt.Println()
+func getDeployment(namespace string) (*appsv1.Deployment, string, error) {
```
rberrelleza, line 141: can split this in two functions? one that gets the deployment / container, and another one that asks the user for info?

rberrelleza: Also, you need to stop the spinner before showing text or the terminal can start behaving strangely.

pkg/cmd/init/run.go
```diff
@@ -0,0 +1,159 @@
...
+func getRunningPod(d *appsv1.Deployment, c *kubernetes.Clientset) (*apiv1.Pod, error) {
+	rs, err := replicasets.GetReplicaSetByDeployment(d, "", c)
+	if err != nil {
+		return nil, err
+	}
+	return pods.GetPodByReplicaSet(rs, "", true, c)
+}
+
+func setWorkDirFromPod(ctx context.Context, dev *model.Dev, pod *apiv1.Pod, container string, config *rest.Config, c *kubernetes.Clientset) *model.Dev {
+	if dev.WorkDir != "" {
```
rberrelleza, line 75: why would this be set if we are generating it?

pkg/cmd/init/run.go
```diff
@@ -0,0 +1,159 @@
...
+
+func setForwardsFromPod(ctx context.Context, dev *model.Dev, pod *apiv1.Pod, c *kubernetes.Clientset) (*model.Dev, error) {
+	ports, err := services.GetPortsByPod(pod, c)
+	if err != nil {
+		return nil, err
+	}
+	for _, port := range ports {
+		localPort := port
+		if port <= 1024 {
+			localPort = port + 8000
```
rberrelleza, line 107: also check if there's no collision with other port forwards?

pkg/cmd/init/run.go
```diff
@@ -0,0 +1,159 @@
...
+		dev.Name = component
+		dev.Labels = map[string]string{l: component}
+		return dev
+	}
+	dev.Name = d.Name
+	return dev
+
+}
+
+func setAnnotationsFromDeployment(dev *model.Dev, d *appsv1.Deployment) *model.Dev {
```
rberrelleza, line 136:
```suggestion
func setFluxAnnotationsIfNeeded(dev *model.Dev, d *appsv1.Deployment) *model.Dev {
```

pkg/cmd/init/run.go
```diff
@@ -0,0 +1,159 @@
...
+}
+
+func setResourcesFromPod(dev *model.Dev, pod *apiv1.Pod, container string) *model.Dev {
+	for i := range pod.Spec.Containers {
+		if pod.Spec.Containers[i].Name != container {
+			continue
+		}
+		if pod.Spec.Containers[i].Resources.Limits != nil {
+			dev.Resources = model.ResourceRequirements{
+				Limits: model.ResourceList{
```
rberrelleza, line 150: how about requests?

pkg/cmd/init/run.go
```diff
@@ -0,0 +1,159 @@
...
+					apiv1.ResourceCPU:    resource.MustParse("1"),
```
rberrelleza, line 151: why are you hardcoding them instead of using the ones from the pod?

pkg/k8s/services/crud.go
```diff
@@ -73,3 +73,25 @@ func DestroyDev(dev *model.Dev, c *kubernetes.Clientset) error {
 func Get(namespace, name string, c kubernetes.Interface) (*apiv1.Service, error) {
 	return c.CoreV1().Services(namespace).Get(name, metav1.GetOptions{})
 }
+
+//GetPortsByPod returns the ports exposed via endpoint of a given pod
+func GetPortsByPod(p *apiv1.Pod, c *kubernetes.Clientset) ([]int, error) {
+	eList, err := c.CoreV1().Endpoints(p.Namespace).List(metav1.ListOptions{})
```
rberrelleza, line 79: why endpoints instead of checking the pod directly?

pkg/cmd/init/run.go
pchico83: it depends on the language

pkg/cmd/init/run.go
pchico83: for dev, it is a good practice to leave them as they are. also, they are useless in Okteto Cloud

pkg/cmd/init/run.go
pchico83: the ones on the pod are not valid for dev. thinks about a go binary image using 50m cpus.
I have set defaults we have seen in Okteto Cloud that makes sense. this also helps for the feature discoverability

pkg/k8s/services/crud.go
pchico83: ports at the pod level are only decorative. the endpoints tell exactly which ports are mapped to a kubernetes service (healthy or not).

PR #922, opened 2020-06-15
Return clean on failure

rberrelleza:
If the "clean" command fails, we shouldn't try and get the watches configuration from the remote container.


cmd/up.go
```diff
@@ -806,6 +806,8 @@ func (up *UpContext) cleanCommand() {
 
 	if err != nil {
 		log.Infof("failed to clean session: %s", err)
+		up.cleaned <- struct{}{}
```
rberrelleza, line 809: @pchico83 do you think we should fail the sequence here instead?

PR #925, opened 2020-06-16
Run the "okteto init" command from "okteto up" if okteto manifest does not exist

pchico83:
@rberrelleza @rlamana this PR completes the work needed to make it easier to integrate okteto in new projects.
I think it is a good moment to test it by yourselves and review the texts and the general experience.

### DIFF
--- a/cmd/init.go
+++ b/cmd/init.go
@@ -37,7 +37,7 @@ const (
 	stignore          = ".stignore"
 	defaultManifest   = "okteto.yml"
 	secondaryManifest = "okteto.yaml"
-	createDeployment  = "new deployment"
+	createDeployment  = "create new deployment"
 )
 
 var wrongImageNames = map[string]bool{
@@ -87,6 +87,9 @@ func Init() *cobra.Command {
 }
 
 func executeInit(namespace, devPath, language, workDir string, overwrite bool) error {
+	fmt.Println("This command walks you through creating an okteto manifest.")
+	fmt.Println("It only covers the most common items, and tries to guess sensible defaults.")
+	fmt.Println("See https://okteto.com/docs/reference/manifest for the official documentation about the okteto manifest.")
 	devPath, err := validateDevPath(devPath, overwrite)
 	if err != nil {
 		return err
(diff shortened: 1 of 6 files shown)


cmd/up.go
```diff
@@ -138,7 +138,25 @@ func Up() *cobra.Command {
...
+				if err != nil {
+					return err
+				}
+				if err := executeInit(namespace, devPath, "", workDir, false); err != nil {
+					return err
+				}
+				log.Success(fmt.Sprintf("Okteto manifest (%s) created", devPath))
+				dev, err = utils.LoadDev(devPath)
+				if err != nil {
+					return err
```
rberrelleza, line 158: move this to a separate function so the code is easier to understand

cmd/init.go
```diff
@@ -260,7 +260,7 @@ func askForOptions(options []string, label string) (string, error) {
 		Items: options,
 		Size:  len(options),
 		Templates: &promptui.SelectTemplates{
-			Label:    fmt.Sprintf("%s {{ . }}", log.BlueString("?")),
+			Label:    fmt.Sprintf("{{ . }}"),
```
rberrelleza, line 263: what is this change for?

cmd/up.go
```diff
@@ -138,7 +138,25 @@ func Up() *cobra.Command {
...
+				if !strings.Contains(err.Error(), "okteto init") {
+					return err
+				}
+				if !utils.AskIfOktetoInit(devPath) {
+					return err
+				}
+
+				workDir, err := os.Getwd()
+				if err != nil {
+					return err
```
rberrelleza, line 150: in this ase don't send the raw error, but instead something that explains what happened.

cmd/utils/dev.go
```diff
@@ -80,6 +80,15 @@ func AskYesNo(q string) (bool, error) {
 	return answer == "y", nil
 }
 
+//AskIfOktetoInit asks if okteto init should be executed
+func AskIfOktetoInit(devPath string) bool {
+	result, err := AskYesNo(fmt.Sprintf("Okteto manifest %s doesn't exist. Do you want to create it [y/n]: ", devPath))
```
rberrelleza, line 85:
```suggestion
	result, err := AskYesNo(fmt.Sprintf("okteto manifest %s doesn't exist, do you want to create it? [y/n] ", devPath))
```

cmd/init.go
```diff
@@ -108,7 +108,7 @@ func executeInit(namespace, devPath, language, workDir string, overwrite bool) e
 	}
 
 	if checkForDeployment {
-		log.Information("This command walks you through creating an okteto manifest.")
+		log.Information("This command walks you through creating an okteto manifest defining your development container.")
```
rberrelleza, line 111: I'd remove the part of development container unless you explain what the development container is. 
```suggestion
		log.Information("This command walks you through creating an okteto manifest.")
```

cmd/init.go
```diff
@@ -227,7 +227,7 @@ func askForDeployment(namespace string, c *kubernetes.Clientset) (*appsv1.Deploy
 	}
 	option, err := askForOptions(
 		options,
-		"Pick the deployment target for your okteto manifest from the list below:",
+		"What is the deployment target for your development container?",
```
rberrelleza, line 230: If this is going to be picked from a list, I'd use 'select' or 'pick' rather that 'What'. 
```suggestion
		"Select the deployment you want to replace with your development container:"
```

cmd/init.go
```diff
@@ -250,7 +250,7 @@ func askForContainer(d *appsv1.Deployment) (string, error) {
 	}
 	return askForOptions(
 		options,
-		fmt.Sprintf("The deployment '%s' has %d containers. Pick the container target for your okteto manifest from the list below:", d.Name, len(d.Spec.Template.Spec.Containers)),
+		fmt.Sprintf("The deployment '%s' has %d containers. What is the container target for your development container?", d.Name, len(d.Spec.Template.Spec.Containers)),
```
rberrelleza, line 253:
```suggestion
		fmt.Sprintf("The deployment '%s' has %d containers. Select the container you want to replace with your development container?", d.Name, len(d.Spec.Template.Spec.Containers)),
```

rberrelleza:
> @rberrelleza @rlamana this PR completes the work needed to make it easier to integrate okteto in new projects.
> I think it is a good moment to test it by yourselves and review the texts and the general experience.

What happens with the case where I don't have a kubeconfig just yet, or  a deployment available?

rberrelleza:
I tried it out, feedback:
1. The info stuff should be displayed before we get the deployments, so there's no "wait with not text"

2. I don't think this two options are equivalent. Maybe we should first ask if new or existing? If not, make the 'create new deployment' the last choice, in a different font (and with regular casing, not title)
image

3. The deployment I was using had a custom 'securityContext', but the resulting okteto.yaml didn't got the users (and it has volumes, so it wouldn't work)

4. In one of my deployments, my pod wasn't running. I think we should mention the error but still continue the command
image

pchico83:
@rberrelleza thanks for the feedback. I forgot to infer the user, great catch!
I will update the PR later

rberrelleza:
> @rberrelleza thanks for the feedback. I forgot to infer the user, great catch!
> I will update the PR later

since we are already poking in the pod, can we infer the real user instead of the one in the securityContext, that could help A LOT in those scenarios

rberrelleza:
also, could we add a final confirmation before writing the file, like `npm init` does?

rberrelleza:
okteto init fails to load if there's no kubeconfig. should we mention the error, and skip to only analyzing the code? (e.g. 'skipping deployment analysis since you don't have valid kubeconfig` in yellow or something)

image

cmd/init.go
pchico83: It removes the question mark at the beginning of the label, which looks odd:
Screenshot 2020-06-17 at 11 55 24

PR #937, opened 2020-06-18
move up to its own package for clarity

rberrelleza:

## Proposed changes
- Only save the syncthing.info file on up 
- Don't log out ports and other things when only running syncthing.Stop
- Reduce log frequency
- Move `up` code to it's own package
- Remove unused channels


cmd/up/up.go
```diff
@@ -286,24 +258,25 @@ func (up *UpContext) Activate(autoDeploy, build, resetSyncthing bool) {
...
-		if !up.retry {
-			analytics.TrackUp(true, up.Dev.Name, up.getClusterType(), len(up.Dev.Services) == 0, up.isSwap, up.Dev.RemoteModeEnabled())
+		if !isRetry {
 			if build {
 				if err := up.buildDevImage(d, create); err != nil {
 					up.Exit <- fmt.Errorf("error building dev image: %s", err)
 					return
 				}
 			}
-		} else if !deployments.IsDevModeOn(d) {
```
rberrelleza, line 297: I'm removing this condition because is already checked on line 247

PR #956, opened 2020-06-25
Check storageclass for generating okteto manifests

### LINKED ISSUE #947: "okteto init" checking default storage class
I run `okteto init` for a Go a Java app and the generated okteto manifest does not work on my cluster because my cluster does not have persistent volume support.

Check if the cluster has persistent volume support and adapt the okteto manifest based on it.

pchico83:

Fixes #947

### DIFF
--- a/cmd/init/init.go
+++ b/cmd/init/init.go
@@ -23,6 +23,7 @@ import (
 	initCMD "github.com/okteto/okteto/pkg/cmd/init"
 	k8Client "github.com/okteto/okteto/pkg/k8s/client"
 	"github.com/okteto/okteto/pkg/k8s/deployments"
+	okLabels "github.com/okteto/okteto/pkg/k8s/labels"
 	"github.com/okteto/okteto/pkg/k8s/namespaces"
 	"github.com/okteto/okteto/pkg/linguist"
 	"github.com/okteto/okteto/pkg/log"
@@ -131,6 +132,7 @@ func Run(namespace, devPath, language, workDir string, overwrite bool) error {
 		}
 
 		if !supportsPersistentVolumes(namespace) {
+			log.Yellow("Default storage class not found in your cluster. Persistent volumes not enabled in your okteto manifest")
 			dev.Volumes = nil
 			dev.PersistentVolumeInfo = nil
 		}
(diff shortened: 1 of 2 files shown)


cmd/init/init.go
```diff
@@ -173,6 +180,35 @@ func getDeployment(namespace string) (*appsv1.Deployment, string, error) {
 	return d, container, nil
 }
 
+func supportsPersistentVolumes(namespace string) bool {
+	c, _, currentNamespace, err := k8Client.GetLocal()
+	if err != nil {
+		log.Debugf("deactivating persistent volumes: %s", err.Error())
```
rberrelleza, line 186:
```suggestion
		log.Debugf("couldn't get kubernetes local client: %s", err.Error())
```

cmd/init/init.go
```diff
@@ -173,6 +180,35 @@ func getDeployment(namespace string) (*appsv1.Deployment, string, error) {
...
+		return false
+	}
+	if namespace == "" {
+		namespace = currentNamespace
+	}
+
+	ns, err := namespaces.Get(namespace, c)
+	if err != nil {
+		log.Debugf("deactivating persistent volumes: %s", err.Error())
```
rberrelleza, line 195:
```suggestion
		log.Debugf("failed to get the current namespace: %s", err.Error())
```

cmd/init/init.go
```diff
@@ -173,6 +180,35 @@ func getDeployment(namespace string) (*appsv1.Deployment, string, error) {
...
+		return true
+	}
+
+	stClassList, err := c.StorageV1().StorageClasses().List(metav1.ListOptions{})
+	if err != nil {
+		log.Debugf("deactivating persistent volumes: %s", err.Error())
+		return false
+	}
+
+	return len(stClassList.Items) > 0
```
rberrelleza, line 209: we should also check that there's a default storage class, I think some of the issues come from there. 

Also, what about incompatible storage classes? (e.g. the issue we saw with minikube and the volumes not persisting a restart). Maybe we can identify those by name and log a warning?

cmd/init/init.go
```diff
@@ -122,6 +124,11 @@ func Run(namespace, devPath, language, workDir string, overwrite bool) error {
 				log.Yellow(fmt.Sprintf("Analysis for deployment '%s' failed: %s", d.Name, err))
 			}
 		}
+
+		if !supportsPersistentVolumes(namespace) {
```
rberrelleza, line 128: should we log warnings when there are errors?

PR #961, opened 2020-06-26
Clean syncthing database on pod restarts if PV is not enabled

### LINKED ISSUE #949: okteto up routinely deletes every file in the current directory
**Describe the bug**

`okteto up` routinely deletes every file in the current directory. I'm using Okteto version 1.8.12.
Similar to #522 

**To Reproduce**
1. `okteto up -n <namepsace> -d`
2. Occasionally the Syncthing process hangs, and I need to Ctrl+C to terminate
3. Files are deleted

**Expected behavior**
The files should not be deleted.

 - OS: Centos
 - Version: 7.8.2003 (Core)

**Additional context**
[cli.txt](https://github.com/okteto/okteto/files/4826281/cli.txt)
[syncthing.log](https://github.com/okteto/okteto/files/4826282/syncthing.log)
[okteto.yaml.txt](https://github.com/okteto/okteto/files/4826294/okteto.yaml.txt)

pchico83:

Fixes #949

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,6 +46,10 @@ integration:
 build:
 	 $(BUILDCOMMAND) -o ${BINDIR}/okteto
 
+.PHONY: build-bin-image
+build-bin-image:
+	 okteto build -t okteto/bin:1.1.21 -f images/bin/Dockerfile images/bin
+
 .PHONY: dep
 dep:
 	go mod tidy
--- a/cmd/up/up.go
+++ b/cmd/up/up.go
@@ -352,6 +352,12 @@ func (up *upContext) activate(autoDeploy, build, resetSyncthing bool) {
 
 		if prevError != nil {
 			if up.shouldRetry(prevError) {
+				if !up.Dev.PersistentVolumeEnabled() {
+					if err := pods.Destroy(up.Pod, up.Dev.Namespace, up.Client); err != nil {
+						up.Exit <- err
+						return
+					}
+				}
 				up.shutdown()
 				continue
 			}
--- a/cmd/up/up.go
+++ b/cmd/up/up.go
@@ -314,7 +314,7 @@ func (up *upContext) activate(autoDeploy, build, resetSyncthing bool) {
 
 		if err := up.sync(resetSyncthing && !isRetry); err != nil {
 
-			if !pods.Exists(up.Pod, up.Dev.Namespace, up.Client) {
+			if err == errors.ErrLostSyncthing || !pods.Exists(up.Pod, up.Dev.Namespace, up.Client) {
 				log.Yellow("\nConnection lost to your development container, reconnecting...\n")
 				up.shutdown()
 				continue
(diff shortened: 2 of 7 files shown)


Makefile
```diff
@@ -46,6 +46,10 @@ integration:
 build:
 	 $(BUILDCOMMAND) -o ${BINDIR}/okteto
 
+.PHONY: build-bin-image
+build-bin-image:
```
rberrelleza, line 50: should we add a manual CI job for this?

rberrelleza: or maybe use the CLI version as the tag for `okteto\bin` and build them both during release?

pkg/k8s/deployments/translate_test.go
```diff
@@ -521,6 +521,7 @@ image: web:latest`)
 							Image:           "web:latest",
 							ImagePullPolicy: apiv1.PullAlways,
 							Command:         []string{"/var/okteto/bin/start.sh"},
+							Args:            []string{"-c"},
```
rberrelleza, line 524: do we need this here, since dev.go (line 685) will inject it?

pkg/k8s/deployments/translate.go
```diff
@@ -428,7 +428,7 @@ func TranslateOktetoInitBinContainer(spec *apiv1.PodSpec) {
 	c := apiv1.Container{
 		Name:            oktetoBinName,
 		Image:           oktetoBinImageTag,
-		ImagePullPolicy: apiv1.PullIfNotPresent,
+		ImagePullPolicy: apiv1.PullAlways,
```
rberrelleza, line 431: this is for the bin image right? if so, could we leave it as it was so it reduces the time for the `up`?

images/bin/start.sh
```diff
@@ -51,12 +55,17 @@ while getopts ":s:r" opt; do
 done
 
 syncthingHome=/var/syncthing
+if [ $ephemeral -eq 1 ] && [ -f ${syncthingHome}/executed ]; then
+    echo "fialing: syncthing restarted and persistent volumes are not enabled in the okteto manifest. Run 'okteto down' and try again"
```
rberrelleza, line 59: `fialing`?

What's this warning about? Is it based on something the user did?

cmd/up/up.go
```diff
@@ -352,6 +352,12 @@ func (up *upContext) activate(autoDeploy, build, resetSyncthing bool) {
 
 		if prevError != nil {
 			if up.shouldRetry(prevError) {
```
rberrelleza, line 354: this has a lot of nested conditions, it makes it hard to understand what's happening. maybe moving it a separate function can allow us to simplify it? (like a `retry` function that checks if the pod needs to be destroyed, sets the retry flag, and other things we do in previous lines of this function)

rberrelleza: and maybe add the preverErr != nil condition to `shouldRetry` as well

cmd/up/up.go
```diff
@@ -352,6 +352,12 @@ func (up *upContext) activate(autoDeploy, build, resetSyncthing bool) {
...
+				if !up.Dev.PersistentVolumeEnabled() {
+					if err := pods.Destroy(up.Pod, up.Dev.Namespace, up.Client); err != nil {
```
rberrelleza, line 356: how much time does this add to the retry cycle? If it's more than a couple of seconds (e.g. due to the image pulling) maybe it's better to not support this retry mode for ephermeral dev envs?  (or maybe we can start optimizing so we don't pullAlways the okteto images as part of this change?

images/bin/start.sh
pchico83: at this point, the `okteto up` sequence will retry from the beginning. if not, `okteto down` must be run

PR #983, opened 2020-07-28
Remove "okteto namespace" sequence from "okteto login" command

pchico83:
## Proposed changes
- Don't run the `okteto namespace` sequence as part of `okteto login`


rberrelleza:
why do we want to remove this?

pchico83:
@rberrelleza we moved to login sequence to the namespace command, o we don't need it here anymore. I have found several scenarios where I want to login to run remote docker builds, but keep my current namespace. `okteto login` with auto credential fetching is very confusing on those cases (for example, I couldn't use it in the git installer job)

PR #999, opened 2020-08-17
command to launch dev jobs

### LINKED ISSUE #998: Support for launching Jobs with the latest code
When building a job, it would useful to be able to automatically apply the same dev overrides that happen to services, and for the job to have the latest code

I can manually add the volume and the claim to any job, but this requires understanding a lot of the internals of okteto

rberrelleza:
Fixes #998

## Proposed changes
Add  a "create job" command that copies an existing job and applies all the overrides, similarly to what we do for services. Since Jobs are ephemeral and can't be edited, we have to clone an existing one and launch it, instead of the approach we do with deployments.

I propose we release this on a separate command, as a 'beta', to see if this approach solves issues for the users who have requested this.

### DIFF
--- a/cmd/create.go
+++ b/cmd/create.go
@@ -15,7 +15,6 @@ package cmd
 
 import (
 	"context"
-	"fmt"
 
 	"github.com/okteto/okteto/cmd/job"
 	"github.com/okteto/okteto/cmd/namespace"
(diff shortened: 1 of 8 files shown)


pchico83:
@rberrelleza should we close this one for now?

PR #1003, opened 2020-08-18
Fix .stignore path

pchico83:

## Proposed changes
- `.stignore` must be the same as the okteto manifest. this was generating trash files when running tests

### DIFF
--- a/cmd/init/init.go
+++ b/cmd/init/init.go
@@ -17,6 +17,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
+	"path/filepath"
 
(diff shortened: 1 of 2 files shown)


cmd/init/init.go
```diff
@@ -149,6 +150,12 @@ func Run(namespace, devPath, language, workDir string, overwrite bool) error {
 		return err
 	}
 
+	devDir, err := filepath.Abs(filepath.Dir(devPath))
```
rberrelleza, line 153: can we add a test to catch regresions?

PR #1006, opened 2020-08-19
Support for synchronizing several folders

### LINKED ISSUE #668: Synchronize more than one folder

`okteto up` synchronizes the local folder where the `okteto.yaml` is placed on to the destination path defined by the  `mountpath` or `workdir` field. 

There are several uses cases where you want to synchronize a different folder than the one the `okteto.yaml` is placed on, or where you want to synchronize several folders. The current `okteto up` behavior is not flexible enough.

I'd like to be able to specify multiple synchronization points. 

I propose the field `sync` to specify the synching behavior. `mounthpath` and `subpath` will be deprecated, and `workdir` keeps the current behavior if `sync`  is not defined. In the future, it will also be deprecated.

`sync` is a list of `localPath:remotePath` pairs. Each pair would indicate a local folder to synchronize into a remote destination.

`LocalPath` can be:
- relative path: In which case, the base folder is the location of `okteto.yml`
- absolute path

We can also support `env var` expansion in the paths, in case a user wants to synchronize ssh keys, env files, etc...

**example**

```
name: hello-world
image: okteto/golang:1
sync: 
- code: /code
- assets: /usr/html/public/assets
- $HOME/.npm/secret:/root/.npm/secret
command: bash
```

This gives the ability to organize files as the user prefers.

pchico83:
Fixes #668

### DIFF
--- a/pkg/model/dev.go
+++ b/pkg/model/dev.go
@@ -390,16 +390,16 @@ func (dev *Dev) setDefaults() error {
 
 func (dev *Dev) translateDeprecatedFields() error {
 
-	showWarning := func(field string) {
+	showWarning := func() {
 		if warningHasBeenShown {
 			return
 		}
-		log.Yellow("'%s' is deprecated to define your synched folders. Use the field 'volumes' instead. More info at https://okteto.com/docs/reference/manifest#volumes-string-optional", field)
+		log.Yellow("'mounthpath' is deprecated to define your synched folders. Use the field 'volumes' instead. More info at https://okteto.com/docs/reference/manifest#volumes-string-optional")
 		warningHasBeenShown = true
 	}
 
 	if dev.MountPath != "" {
-		showWarning("mounthpath")
+		showWarning()
 		dev.Volumes = append(
 			dev.Volumes,
 			Volume{
@@ -409,7 +409,6 @@ func (dev *Dev) translateDeprecatedFields() error {
 		)
 	} else if dev.WorkDir != "" && !dev.HasLocalVolumes() {
 		dev.MountPath = dev.WorkDir
-		showWarning("workdir")
 		dev.Volumes = append(
 			dev.Volumes,
 			Volume{
@@ -424,7 +423,7 @@ func (dev *Dev) translateDeprecatedFields() error {
 			if dev.MountPath == "" {
 				return fmt.Errorf("'mountpath' is not supported. Use 'volumes' instead to specify your synched folders (https://okteto.com/docs/reference/manifest#volumes-string-optional)")
 			}
-			showWarning("mounthpath")
+			showWarning()
 			s.Volumes = append(
 				s.Volumes,
 				Volume{
@@ -436,7 +435,6 @@ func (dev *Dev) translateDeprecatedFields() error {
 			if dev.MountPath == "" {
 				return fmt.Errorf("'mountpath' is not supported. Use 'volumes' instead to specify your synched folders (https://okteto.com/docs/reference/manifest#volumes-string-optional)")
 			}
-			showWarning("workdir")
 			s.Volumes = append(
 				s.Volumes,
 				Volume{
--- a/pkg/model/dev.go
+++ b/pkg/model/dev.go
@@ -35,9 +35,6 @@ import (
 )
 
 const (
-	//OktetoBinImageTag image tag with okteto internal binaries
-	OktetoBinImageTag = "okteto/bin:1.2.3"
-
 	oktetoSSHServerPortVariable = "OKTETO_REMOTE_PORT"
 	oktetoDefaultSSHServerPort  = 2222
 	//OktetoDefaultPVSize default volume size
@@ -88,7 +85,7 @@ const (
 
 var (
 	//OktetoBinImageTag image tag with okteto internal binaries
-	OktetoBinImageTag = "okteto/bin:1.2.2"
+	OktetoBinImageTag = "okteto/bin:1.2.3"
 
 	errBadName = fmt.Errorf("Invalid name: must consist of lower case alphanumeric characters or '-', and must start and end with an alphanumeric character")
 
@@ -111,8 +108,8 @@ type Dev struct {
 	Labels               map[string]string     `json:"labels,omitempty" yaml:"labels,omitempty"`
 	Annotations          map[string]string     `json:"annotations,omitempty" yaml:"annotations,omitempty"`
 	Tolerations          []apiv1.Toleration    `json:"tolerations,omitempty" yaml:"tolerations,omitempty"`
-	Namespace            string                `json:"namespace,omitempty" yaml:"namespace,omitempty"`
 	Context              string                `json:"context,omitempty" yaml:"context,omitempty"`
+	Namespace            string                `json:"namespace,omitempty" yaml:"namespace,omitempty"`
 	Container            string                `json:"container,omitempty" yaml:"container,omitempty"`
 	Image                *BuildInfo            `json:"image,omitempty" yaml:"image,omitempty"`
 	Push                 *BuildInfo            `json:"-" yaml:"push,omitempty"`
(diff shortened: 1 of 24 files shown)


pkg/model/dev.go
```diff
@@ -299,6 +296,35 @@ func Read(bytes []byte) (*Dev, error) {
...
+	dev.Image.Context = loadAbsPath(devDir, dev.Image.Context)
+	dev.Image.Dockerfile = loadAbsPath(devDir, dev.Image.Dockerfile)
+	dev.Push.Context = loadAbsPath(devDir, dev.Push.Context)
+	dev.Push.Dockerfile = loadAbsPath(devDir, dev.Push.Dockerfile)
+
+	for i := range dev.Volumes {
+		if dev.Volumes[i].LocalPath == "" {
+			continue
+		}
+		dev.Volumes[i].LocalPath = loadAbsPath(devDir, dev.Volumes[i].LocalPath)
```
rberrelleza, line 322: are we going to support absolute paths on the left? (e.g. to sync a folder that it's not on the repo where I'm running `okteto up` from, like $HOME/.ssh).

pkg/model/dev.go
```diff
@@ -370,6 +383,65 @@ func (dev *Dev) setDefaults() error {
 	return nil
 }
 
+func (dev *Dev) translateDeprecatedFields() error {
+
+	showWarning := func() {
+		if warningHasBeenShown {
```
rberrelleza, line 389: use `sync.once` instead, to be more idiomatic go?

pkg/model/dev.go
```diff
@@ -370,6 +383,65 @@ func (dev *Dev) setDefaults() error {
...
+				LocalPath:  ".",
+				RemotePath: dev.WorkDir,
+			},
+		)
+	}
+
+	for _, s := range dev.Services {
+		if s.MountPath != "" {
+			if dev.MountPath == "" {
+				return fmt.Errorf("'mountpath' is not supported. Use 'volumes' instead to specify your synched folders (https://okteto.com/docs/reference/manifest#volumes-string-optional)")
```
rberrelleza, line 419:
```suggestion
				return fmt.Errorf("'mountpath' is not supported. Use 'volumes' instead to specify your synchronized folders (https://okteto.com/docs/reference/manifest#volumes-string-optional)")
```

pkg/model/dev.go
```diff
@@ -370,6 +383,65 @@ func (dev *Dev) setDefaults() error {
 	return nil
 }
 
+func (dev *Dev) translateDeprecatedFields() error {
+
+	showWarning := func() {
+		if warningHasBeenShown {
+			return
+		}
+		log.Yellow("'mounthpath' is deprecated to define your synched folders. Use the field 'volumes' instead. More info at https://okteto.com/docs/reference/manifest#volumes-string-optional")
```
rberrelleza, line 392:
```suggestion
		log.Yellow("'mounthpath' is deprecated to define your synchronized folders. Use the field 'volumes' instead. More info at https://okteto.com/docs/reference/manifest#volumes-string-optional")
```

pkg/model/dev.go
rberrelleza: is not supported or deprecated?

pkg/model/dev.go
```diff
@@ -619,27 +652,9 @@ func (dev *Dev) ToTranslationRule(main *Dev) *TranslationRule {
...
-				Name:      main.GetVolumeName(),
-				MountPath: dev.MountPath,
-				SubPath:   fullDevSubPath(dev.SubPath),
-			},
-		)
-	}
-
 	if main == dev {
-		rule.Marker = dev.DevPath
+		rule.Marker = OktetoBinImageTag //for backward compatibility
```
rberrelleza, line 656: why do we need the bin image as the marker file?

pkg/model/dev.go
```diff
@@ -700,21 +715,28 @@ func (dev *Dev) ToTranslationRule(main *Dev) *TranslationRule {
...
 
-	for _, v := range dev.Volumes {
-		var devSubPath string
-		if v.SubPath == "" {
-			devSubPath = fullGlobalSubPath(v.MountPath)
-		} else {
-			devSubPath = fullDevSubPath(v.SubPath)
+	if main.PersistentVolumeEnabled() {
+		for _, v := range dev.Volumes {
+			if v.LocalPath == "" {
```
rberrelleza, line 720: can we create the volume and just change the subpath once to avoid declaring the whole thing twice?

pkg/model/volumes.go
```diff
@@ -0,0 +1,217 @@
...
+
+import (
+	"fmt"
+	"path/filepath"
+	"strings"
+
+	"github.com/okteto/okteto/pkg/errors"
+)
+
+// HasLocalVolumes returns true if the manifest has local volumes
```
rberrelleza, line 24: what do you mean by local vs remote volumes?

pkg/model/volumes.go
```diff
@@ -0,0 +1,217 @@
...
+	for _, v := range dev.Volumes {
+		if v.LocalPath == "" {
+			return true
+		}
+	}
+	return false
+}
+
+// IsSyncFolder returns if path must be synched
+func (dev *Dev) IsSyncFolder(path string) (bool, error) {
```
rberrelleza, line 45: for which scenario will this return false?

pkg/model/volumes.go
```diff
@@ -0,0 +1,217 @@
...
+		}
+		found = true
+		if rel != "." {
+			return false, nil
+		}
+	}
+	if found {
+		return true, nil
+	}
+	return false, errors.ErrNotFound
```
rberrelleza, line 66: what does not-found means on this context?

pkg/model/volumes.go
```diff
@@ -0,0 +1,217 @@
...
+	if dev.PersistentVolumeInfo == nil {
+		return true
+	}
+	return dev.PersistentVolumeInfo.Enabled
+}
+
+// PersistentVolumeSize returns the persistent volume size
+func (dev *Dev) PersistentVolumeSize() string {
+	if dev.PersistentVolumeInfo == nil {
+		return OktetoDefaultPVSize
```
rberrelleza, line 80: can we set this when we do the SetDefault or in the deserialization for consistency?

pkg/model/volumes.go
```diff
@@ -0,0 +1,217 @@
...
+	}
+	if found {
+		return true, nil
+	}
+	return false, errors.ErrNotFound
+}
+
+// PersistentVolumeEnabled returns true if persistent volumes are enabled for dev
+func (dev *Dev) PersistentVolumeEnabled() bool {
+	if dev.PersistentVolumeInfo == nil {
```
rberrelleza, line 71: can we set this when we do the SetDefault or in the deserialization for consistency?

pkg/model/volumes.go
```diff
@@ -0,0 +1,217 @@
...
+	if dev.HasRemoteVolumes() {
+		return fmt.Errorf("'persistentVolume.enabled' must be set to true to use remote volumes")
+	}
+	for _, v := range dev.Volumes {
+		result, err := dev.IsSyncFolder(v.LocalPath)
+		if err != nil {
+			return err
+		}
+		if !result {
+			return fmt.Errorf("'persistentVolume.enabled' must be set to true to use subfolders in the 'volumes' field")
```
rberrelleza, line 112: why do we need persistent volumes to do subfolders? now it's only needed if you want to share it with a service no?

pkg/model/volumes.go
```diff
@@ -0,0 +1,217 @@
...
+		if err != nil {
+			return fmt.Errorf("Local volume '%s' in '%s' not defined in the main development container", v.LocalPath, dev.Name)
+		}
+	}
+	return nil
+}
+
+func (dev *Dev) validateVolumes(main *Dev) error {
+	if !dev.HasLocalVolumes() {
+		return fmt.Errorf("the 'volumes' field must define a local folder to be synched. More info at https://okteto.com/docs/reference/manifest#volumes-string-optional")
```
rberrelleza, line 181: on which condition would this error be triggered?

pkg/syncthing/syncthing.go
```diff
@@ -89,12 +88,18 @@ type Syncthing struct {
...
 	LocalPort        int          `yaml:"-"`
-	Source           string       `yaml:"-"`
 	Type             string       `yaml:"-"`
 	IgnoreDelete     bool         `yaml:"-"`
 	pid              int          `yaml:"-"`
 }
 
+//Folder represents a sync folder
+type Folder struct {
+	Name       string `yaml:"name"`
```
rberrelleza, line 98: what's the name of the folder?

pkg/syncthing/syncthing.go
```diff
@@ -317,71 +338,84 @@ func (s *Syncthing) WaitForPing(ctx context.Context, local bool) error {
...
+		err = json.Unmarshal(body, ignores)
+		if err != nil {
+			return fmt.Errorf("error unmarshalling ignore files: %s", err.Error())
+		}
+		for i, line := range ignores.Ignore {
+			line := strings.TrimSpace(line)
+			if line == "" {
+				continue
+			}
+			if strings.Contains(line, "(?d)") {
```
rberrelleza, line 359: can you add a comment on what's this check about?

pkg/model/volumes.go
pchico83: data volumes. or if you have something like this:
```
volumes:
   - .:/app
   - config:/config
```
the second one is not a sync folder, it is a subpath of the first one and it is not managed by synchthing, but just using a k8s volume mount with the proper subpath

pkg/model/volumes.go
pchico83: not related to this PR, I created an issue to handle it:
https://github.com/okteto/okteto/issues/1007

pkg/model/volumes.go
pchico83: https://github.com/okteto/okteto/issues/1007

pkg/model/volumes.go
pchico83: it is also required now, but we are not failing fast. subfolders use k8s volume mounts, and volume mounts require a volume to be defined

pkg/model/volumes.go
pchico83: when there is not `workdir` nor a Synched Volume define. It would mean nothing is synched. We can support this use case in the future

pkg/syncthing/syncthing.go
pchico83: I am just using numbers in sequence. this is not exposed to the user

PR #1008, opened 2020-08-20
Support for setting the kubernetes context in the okteto manifest

### LINKED ISSUE #997: Add `context` attribute to okteto manifest
When working with multiple contexts (development and one or more production contexts) it might happen that the developer forgets to switch to the development context before calling `okteto up`. That can inadvertently cause a development deployment to be booted on a production environment instead, with all side-affects. 

An additional attribute called `allowed-contexts` where one or more contexts can be listed in which running okteto is allowed.

Adding the label `dev.okteto.com/not-allowed: "true"` to the namespace in question (default in this case). Because the infrastructure is managed using terraform it is not possible to manually add a label to the namespace, nor is it possible to create the namespace when it already exists (default namespace is automatically generated and it causes a `namespace default already exists` error when tried).

pchico83:

Fixes #997

Supports variable expansion.

### DIFF
--- a/pkg/cmd/stack/translate.go
+++ b/pkg/cmd/stack/translate.go
@@ -43,7 +43,7 @@ func translate(s *model.Stack, forceBuild, noCache bool) error {
 		return nil
 	}
 
-	c, _, configNamespace, err := k8Client.GetLocal()
+	c, _, configNamespace, err := k8Client.GetLocal("")
 	if err != nil {
 		return err
 	}
--- a/pkg/model/dev.go
+++ b/pkg/model/dev.go
@@ -105,6 +105,7 @@ type Dev struct {
 	Annotations          map[string]string     `json:"annotations,omitempty" yaml:"annotations,omitempty"`
 	Tolerations          []apiv1.Toleration    `json:"tolerations,omitempty" yaml:"tolerations,omitempty"`
 	Namespace            string                `json:"namespace,omitempty" yaml:"namespace,omitempty"`
+	Context              string                `json:"context,omitempty" yaml:"context,omitempty"`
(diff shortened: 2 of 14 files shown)


rberrelleza:
How does this work? Is this to 'force' okteto commands to run on a specific k8s context?

pkg/model/dev.go
```diff
@@ -731,16 +733,20 @@ func (dev *Dev) ToTranslationRule(main *Dev) *TranslationRule {
 	return rule
 }
 
-//UpdateNamespace updates the dev namespace
-func (dev *Dev) UpdateNamespace(namespace string) error {
-	if namespace == "" {
-		return nil
+//UpdateContext updates the dev namespace and context
+func (dev *Dev) UpdateContext(namespace, k8sContext string) {
```
rberrelleza, line 737: Can we call this LoadContext or SetContext instead? Update sounds as if we are writing something in the context or changing it.

pkg/cmd/stack/translate.go
```diff
@@ -43,7 +43,7 @@ func translate(s *model.Stack, forceBuild, noCache bool) error {
 		return nil
 	}
 
-	c, _, configNamespace, err := k8Client.GetLocal()
+	c, _, configNamespace, err := k8Client.GetLocal("")
```
rberrelleza, line 46: why is it empty here? if we are considering the context, don't we need it in all the k8s calls?

rberrelleza:
I think we should make this visible as well, maybe add it to the summary we display in `okteto up` if it's non-empty?

rberrelleza:
> How does this work? Is this to 'force' okteto commands to run on a specific k8s context?

discussed offline: it's going to work by changing the context for the commands invoked by okteto

PR #1015, opened 2020-08-26
Remove context from kubeconfig on `okteto namespace delete`

### LINKED ISSUE #493: `okteto delete namespace` should remove the `Kubeconfig` entry
`okteto delete namespace` should remove the `Kubeconfig` entry, since `okteto create namespace` created it.

Cirrith:
Fixes #493 

Will also remove cluster and user if no longer referenced by contexts.

### DIFF
--- a/cmd/namespace/delete.go
+++ b/cmd/namespace/delete.go
@@ -53,8 +53,8 @@ func executeDeleteNamespace(ctx context.Context, namespace string) error {
 		return fmt.Errorf("failed to delete namespace: %s", err)
 	}
 
-	if err := CleanNamespace(ctx, namespace); err != nil {
-		return fmt.Errorf("failed to clean up namespace: %s", err)
+	if err := RemoveNamespace(ctx, namespace); err != nil {
+		return fmt.Errorf("failed to remove namespace: %s", err)
 	}
 
 	log.Success("Namespace '%s' deleted", namespace)
--- a/cmd/namespace/delete.go
+++ b/cmd/namespace/delete.go
@@ -53,6 +53,10 @@ func executeDeleteNamespace(ctx context.Context, namespace string) error {
 		return fmt.Errorf("failed to delete namespace: %s", err)
 	}
 
+	if err := CleanNamespace(ctx, namespace); err != nil {
+		return fmt.Errorf("failed to clean up namespace: %s", err)
+	}
+
 	log.Success("Namespace '%s' deleted", namespace)
 	return nil
 }
--- a/cmd/namespace/namespace.go
+++ b/cmd/namespace/namespace.go
@@ -85,35 +85,31 @@ func RunNamespace(ctx context.Context, namespace string) error {
 	}
 
 	kubeConfigFile := config.GetKubeConfigFile()
+	clusterHost := getClusterHost()
 
-	u, _ := url.Parse(okteto.GetURL())
-	parsedHost := strings.ReplaceAll(u.Host, ".", "_")
-
-	if err := okteto.SetKubeConfig(cred, kubeConfigFile, namespace, okteto.GetUserID(), parsedHost); err != nil {
+	if err := okteto.SetKubeConfig(cred, kubeConfigFile, namespace, okteto.GetUserID(), clusterHost); err != nil {
 		return err
 	}
 
-	log.Success("Updated context '%s' in '%s'", parsedHost, kubeConfigFile)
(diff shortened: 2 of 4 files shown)


Cirrith:
This my first open-source commit, any recommendations/feedback is welcome!

cmd/namespace/delete.go
```diff
@@ -53,6 +53,10 @@ func executeDeleteNamespace(ctx context.Context, namespace string) error {
 		return fmt.Errorf("failed to delete namespace: %s", err)
 	}
 
+	if err := CleanNamespace(ctx, namespace); err != nil {
```
rberrelleza, line 56: nit: I'd make this lowercase, since it should be private to this package only

pkg/okteto/client.go
```diff
@@ -165,6 +148,50 @@ func SetKubeConfig(cred *Credential, kubeConfigPath, namespace, userName, cluste
...
+//RemoveKubeConfig removes from a kubeconfig fle the okteto cluster credentials
+func RemoveKubeConfig(cred *Credential, kubeConfigPath, namespace, userName, clusterName string) error {
+	contextToRemoveName := getContextName(cred, namespace, clusterName)
+	cfg, err := getOrCreateKubeConfig(kubeConfigPath)
+	if err != nil {
+		return err
+	}
+
+	contextToRemove, ok := cfg.Contexts[contextToRemoveName]
+	if !ok {
```
rberrelleza, line 160: I think it's better not to return an error, since the objective of the call  (to remove the context) has been accomplished).

cmd/namespace/namespace.go
```diff
@@ -97,6 +97,26 @@ func RunNamespace(ctx context.Context, namespace string) error {
...
+func CleanNamespace(ctx context.Context, namespace string) error {
+	cred, err := okteto.GetCredentials(ctx, namespace)
+	if err != nil {
+		return err
+	}
+
+	kubeConfigFile := config.GetKubeConfigFile()
+
+	u, _ := url.Parse(okteto.GetURL())
+	parsedHost := strings.ReplaceAll(u.Host, ".", "_")
```
rberrelleza, line 110: can we make this a common function, since it's also called on `RunNamespace` ?

cmd/namespace/namespace.go
```diff
@@ -97,6 +97,26 @@ func RunNamespace(ctx context.Context, namespace string) error {
...
+
+	if err := okteto.RemoveKubeConfig(cred, kubeConfigFile, namespace, okteto.GetUserID(), parsedHost); err != nil {
+		return err
+	}
+
+	log.Success("Cleaned context '%s' in '%s'", parsedHost, kubeConfigFile)
```
rberrelleza, line 116:
```suggestion
	log.Success("Removed context '%s' from '%s'", parsedHost, kubeConfigFile)
```

pkg/okteto/client.go
```diff
@@ -173,3 +200,33 @@ func InDevContainer() bool {
...
+	}
+	return cfg, nil
+}
+
+func getContextName(cred *Credential, namespace, clusterName string) string {
+	contextName := ""
+	if namespace == "" {
+		// don't include namespace for the personal namespace
+		contextName = clusterName
+		namespace = cred.Namespace
```
rberrelleza, line 227: what's the objective of this assignment? `namespace` doesn't seem to be used anywhere below this.

Cirrith: I copied it from SetKubeConfig which updates namespace. I'll resolve this

cmd/namespace/namespace.go
Cirrith: Should I rename everything to remove instead of clean? That seems to make more sense

cmd/namespace/delete.go
Cirrith: Feels like it should mirror RunNamespace, which is sorta its opposite. Thoughts?

cmd/namespace/namespace.go
rberrelleza: Yeah, that would be great.

cmd/namespace/delete.go
rberrelleza: Ah, good point.

pchico83 (approved):

rberrelleza:
Thanks for the contribution @Cirrith!

PR #1025, opened 2020-09-02
add commands to create and destroy pipeline

rberrelleza:
Fixes https://github.com/okteto/cloud-feedback/issues/48

This PR adds two new commands:
- okteto create pipeline
- okteto delete pipeline

This command is the equivalent of going to okteto cloud/okteto enterprise, clicking on the deploy button and passing the repository and branch. 

This is the first iteration, where I want to focus on core functionality, use cases, and 'shape' of the command. On a next PR, I'd like to add:
- automatically detect the current repository
- automatically detect the current branch
- ensure all commands are optional

Current design decisions:
- This command works by invoking the existing okteto cloud API, rather than implementing the commands locally
- This command doesn't check if the repository is 'okteto pipeline' ready, leaving that job to Okteto Cloud
- The command fails if you don't pass a 'namespace' and the current namespace is not managed by okteto
- `--wait` means to wait until the pipeline is finished, not that the actual application deployed is ready. 
- I'm using the `create/delete` verbs, since that's what we already use for the `create namespace`, `delete namespace` commands. I think that a deploy/destroy verb would be more correct, but I'm not too keen on adding more verbs, specially when create/delete only have 1 command.  Another option to consider is to follow the same pattern as `stacks`, and have a `okteto pipelines deploy` and `okteto pipelines destroy`. 

What do you think?

(I'll add test coverage once we agree on the design of the command)


pchico83:
@rlamana @rberrelleza I love this one!
I would go with the same syntax than stacks for now: 
- `okteto pipeline deploy`
- `okteto pipeline destroy`
It is what feels more auto-explanatory. It is what the UI is using too.
At some point, we will probably need to rethink the general command syntax.

cmd/pipeline/deploy.go
```diff
@@ -0,0 +1,164 @@
...
+			Hint: "Run `okteto namespace`, or use the `--namespace` parameter",
+		}
+	}
+
+	ns, err := namespaces.Get(namespace, c)
+	if err != nil {
+		return namespace, nil
+	}
+
+	if !namespaces.IsOktetoNamespace(ns) {
```
pchico83, line 156: Instead of failing, I would use the default/personal Okteto Cloud namespace here (`space` not defined).
I could be pointing my context to minikube, for example, and run the pipeline on Okteto Cloud on a git hook.

pchico83: Although it could be confusing. Let's go with this for now.
I think we need the concept of okteto context sooner than later.

rberrelleza: yes. I think we could find a way of incorporate the context property of the okteto.yaml as part of this process (on a separate PR). But before that I want to focus on the flow and use cases this enable.

pchico83 (approved):

PR #1047, opened 2020-09-22
Return a friendly error when running okteto exec on a non-dev pod

### LINKED ISSUE #1046: Okteto Exec breaks with non-dev container
**Describe the bug**
`okteto exec` is breaking with an ugly error when I try to run in a non-dev container

**To Reproduce**
`$kubectl apply -f k8s.yaml`
`$okteto exec bash`

**Expected behavior**
```panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x150 pc=0x1995c26]

goroutine 1 [running]:
github.com/okteto/okteto/cmd.executeExec(0x26eb600, 0xc0006caf00, 0xc00021f0e0, 0xc00081aa60, 0x1, 0x1, 0xc0005ce660, 0xc00129fd08)
	/home/circleci/project/cmd/exec.go:105 +0x2e6
github.com/okteto/okteto/cmd.Exec.func1(0xc000840280, 0xc00081aa60, 0x1, 0x1, 0x0, 0x0)
	/home/circleci/project/cmd/exec.go:53 +0x16e
github.com/spf13/cobra.(*Command).execute(0xc000840280, 0xc00081aa40, 0x1, 0x1, 0xc000840280, 0xc00081aa40)
	/go/pkg/mod/github.com/spf13/cobra@v0.0.5/command.go:826 +0x47c
github.com/spf13/cobra.(*Command).ExecuteC(0xc0009a3b80, 0xc00129ff58, 0x1, 0x1)
	/go/pkg/mod/github.com/spf13/cobra@v0.0.5/command.go:914 +0x30b
	/go/pkg/mod/github.com/spf13/cobra@v0.0.5/command.go:864
main.main()
	/home/circleci/project/main.go:94 +0x874
```

 - OS: Ubuntu 20.04
 - Okteto version 1.8.22

rberrelleza:
Fixes #1046

## Proposed changes
- Return a friendly error when running `okteto exec` on a non-dev container
- Check if the pod exists before running okteto exec in remote mode

### DIFF
--- a/cmd/exec.go
+++ b/cmd/exec.go
@@ -82,11 +82,6 @@ func executeExec(ctx context.Context, dev *model.Dev, args []string) error {
 	wrapped := []string{"sh", "-c"}
 	wrapped = append(wrapped, args...)
 
-	if dev.ExecuteOverSSHEnabled() || dev.RemoteModeEnabled() {
-		log.Infof("executing remote command over SSH")
-		return ssh.Exec(ctx, dev.RemotePort, true, os.Stdin, os.Stdout, os.Stderr, wrapped)
-	}
-
 	client, cfg, namespace, err := k8Client.GetLocal(dev.Context)
 	if err != nil {
 		return err
@@ -101,8 +96,21 @@ func executeExec(ctx context.Context, dev *model.Dev, args []string) error {
 		return err
 	}
 
+	if p == nil {
+		return errors.UserError{
+			E:    fmt.Errorf("development mode is not enabled on your deployment"),
+			Hint: "Run `okteto up` to enable it and try again",
+		}
+	}
+
 	if dev.Container == "" {
 		dev.Container = p.Spec.Containers[0].Name
+	}
+
+	if dev.ExecuteOverSSHEnabled() || dev.RemoteModeEnabled() {
+		log.Infof("executing remote command over SSH")
+		dev.LoadRemote(ssh.GetPublicKey())
+		return ssh.Exec(ctx, dev.RemotePort, true, os.Stdin, os.Stdout, os.Stderr, wrapped)
 	}
 
 	return exec.Exec(ctx, client, cfg, dev.Namespace, p.Name, dev.Container, true, os.Stdin, os.Stdout, os.Stderr, wrapped)


pchico83 (approved):

cmd/exec.go
```diff
@@ -101,9 +96,22 @@ func executeExec(ctx context.Context, dev *model.Dev, args []string) error {
 		return err
 	}
 
+	if p == nil {
+		return errors.UserError{
```
pchico83, line 100: @rberrelleza when working with complex apps, it is usually required to run `okteto exec` to finish the app configuration from files that only exist in the local filesystem. When users face this error, they get really confused.
What could we do? should we run `okteto up` + `okteto exec` + `okteto down` for these cases?

rberrelleza: @pchico83  I don't understand what you mean here. This error happens if they are trying to do "okteto exec" on a pod that's not on developer mode.   

Are you suggesting that we should allow "okteto exec" on  non-dev pods?

pchico83: The use case is to be able to run `exec` on non-dev pods, but activating dev mode first.
For example, they have a `.env` file not committed in git. The deploy the app, and they want to run:
```
okteto exec -- make configure
```
which uses `.env`. Now, they need to run `okteto up`, `okteto exec` and `okteto down`. Andif they don't do `up`, we show this error, which it is generating confusion.
Maybe it is a use case for `okteto up -- make configure`

rlamana: I totally see this as a use case for the `okteto up -- make configure` or a similar way to execute the command but implying that it is happing in a dev container. Allowing `exec` and magically doing `up` probably causes more confusion about whether it is happening on a non-dev container or not.

pchico83: @rlamana would you automate `okteto down` after `okteto up -- make configure`?

rlamana: so we are talking about a single execution and then it dies and does `okteto down`?

rlamana: I thought of `okteto up -- command` more of a way to do UP and run a command overriding the one defined in the `okteto.yml`. I would not do the `down`.

rlamana: I understand this is kind of a corner case and not an operation you keep doing over and over again. I would focus on keeping everything coherent and intuitive, even if that means to run 2 commands (up and then down), rather than adding more magic that makes it difficult to understand what exactly is happening.

rberrelleza:
> the use case is to be able to run exec on non-dev pods

cmd/exec.go
rberrelleza: > I understand this is kind of a corner case and not an operation you keep doing over and over again. I would focus on keeping everything coherent and intuitive, even if that means to run 2 commands (up and then down), rather than adding more magic that makes it difficult to understand what exactly is happening.

I agree with @rlamana , for me, an "okteto up -- command" means "do an up, and then run `command` instead of the command defined on okteto.yml"

PR #1052, opened 2020-09-26
Support for expanding default values

pchico83:

## Proposed changes
- Support for expanding default values

### DIFF
--- a/pkg/model/dev.go
+++ b/pkg/model/dev.go
@@ -287,19 +287,13 @@ func Read(bytes []byte) (*Dev, error) {
 		}
 	}
 
-	dev.loadName()
-	for _, s := range dev.Services {
-		s.loadName()
+	if err := dev.expandEnvVars(); err != nil {
+		return nil, err
 	}
-
-	dev.loadLabels()
 	for _, s := range dev.Services {
-		s.loadLabels()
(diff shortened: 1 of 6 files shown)


pkg/model/dev.go
```diff
@@ -823,3 +824,13 @@ func (s *Secret) GetKeyName() string {
 func (s *Secret) GetFileName() string {
 	return filepath.Base(s.RemotePath)
 }
+
+//ExpandEnv expands the environments supporting the notation "${var:-$DEFAULT}"
+func ExpandEnv(value string) string {
+	result, err := envsubst.String(value)
+	if err != nil {
+		log.Errorf("error expanding variables on '%s': %s", value, err.Error())
```
rberrelleza, line 832: Log.error logs to the UI. Instead of this return an error, and the caller can handle ir.

If this fails we shouldn't proceed, since we'd be having bad data.

rberrelleza (changes requested):

rberrelleza (changes requested):
if the expansion fails we should not log an error that deep (since those get displayed in the UI). I think we should surface the error, since that mostly likely happens when there's a parsing situation or things like that.

PR #1061, opened 2020-10-03
Serialize command as single string if possible

pchico83:
## Proposed changes
- Serialize command as single string if possible


pkg/model/serializer.go
```diff
@@ -71,9 +71,9 @@ func (c *Command) UnmarshalYAML(unmarshal func(interface{}) error) error {
 
 // MarshalYAML Implements the marshaler interface of the yaml pkg.
 func (c Command) MarshalYAML() (interface{}, error) {
-	// if len(c.Values) == 1 && !strings.Contains(c.Values[0], " ") {
-	// 	return c.Values[0], nil
-	// }
+	if len(c.Values) == 1 && !strings.Contains(c.Values[0], " ") {
```
rberrelleza, line 74: can we add tests for this branch?

PR #1069, opened 2020-10-03
Don't wait for remote init scan if using persistent volumes

pchico83:

## Proposed changes
- Don't wait for remote init scan if using persistent volumes
- Configurable rescan interval

### DIFF
--- a/pkg/syncthing/api.go
+++ b/pkg/syncthing/api.go
@@ -20,6 +20,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"path"
+	"strings"
 	"time"
 
 	"github.com/okteto/okteto/pkg/log"
@@ -50,7 +51,7 @@ func (s *Syncthing) APICall(ctx context.Context, url, method string, code int, p
 		if err == nil {
 			return result, nil
 		}
-		if retries == 3 {
+		if retries == 3 || strings.Contains(url, "rest/db/ignores") {
 			return nil, err
 		}
 		log.Infof("retrying syncthing call[%s] local=%t: %s", url, local, err.Error())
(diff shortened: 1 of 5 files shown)


pkg/syncthing/api.go
```diff
@@ -50,7 +51,7 @@ func (s *Syncthing) APICall(ctx context.Context, url, method string, code int, p
 		if err == nil {
 			return result, nil
 		}
-		if retries == 3 {
+		if retries == 3 || strings.Contains(url, "rest/db/ignores") {
```
rberrelleza, line 54: why do we need this? could we instead control this on the caller of APICall? (maybe include a param that controls the retries?)

pkg/syncthing/syncthing.go
```diff
@@ -166,6 +167,10 @@ func New(dev *model.Dev) (*Syncthing, error) {
 		hash = []byte("")
 	}
 
+	rescanInterval := os.Getenv("OKTETO_RESCAN_INTERVAL")
+	if rescanInterval == "" {
+		rescanInterval = "3600"
```
rberrelleza, line 172: what's syncthing default?

rberrelleza (commented):
What is this PR fixing?

pkg/syncthing/syncthing.go
pchico83: 3600

pchico83:
> What is this PR fixing?

Scanning takes a lot of time with large folders. When using persistent volumes, there is no need to wait for scanning because:
a) on initial `up`, folder is empty.
b) on future `up`s, syncthing database is persisted

rberrelleza (approved):

PR #1079, opened 2020-10-06
Sort stack envvars before deploying stack services

pchico83:

## Proposed changes
- Kubernetes redeploys pods if envvars have a different order

### DIFF
--- a/pkg/cmd/stack/translate.go
+++ b/pkg/cmd/stack/translate.go
@@ -17,6 +17,8 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"sort"
+	"strings"
 
 	"github.com/okteto/okteto/pkg/cmd/build"
 	k8Client "github.com/okteto/okteto/pkg/k8s/client"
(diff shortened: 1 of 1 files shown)


pkg/cmd/stack/translate.go
```diff
@@ -58,6 +60,9 @@ func translateEnvVars(s *model.Stack) error {
 		for _, envFilepath := range svc.EnvFiles {
 			translateEnvFile(&svc, envFilepath)
 		}
+		sort.SliceStable(svc.Environment, func(i, j int) bool {
```
rberrelleza, line 63: why do we need to do this? To make it easier to read?

rberrelleza (approved):

pkg/cmd/stack/translate.go
pchico83: Kubernetes redeploys pods if envvars have a different order. This is very inconvenient with stefulsets, for example. If my stack yaml does change, it shouldn't be expected that pods are redeployed

PR #1090, opened 2020-10-09
Update auth query to use "externalID"

pchico83:
## Proposed changes
- The Okteto API has renamed "githubid" to "externalid"

### DIFF
--- a/pkg/okteto/auth.go
+++ b/pkg/okteto/auth.go
@@ -47,7 +47,7 @@ type Token struct {
 type User struct {
 	Name        string
 	Email       string
-	GithubID    string
+	ExternalID  string
 	Token       string
 	ID          string
 	New         bool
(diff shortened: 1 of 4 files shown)


pkg/okteto/auth.go
```diff
@@ -130,7 +130,7 @@ func queryUser(ctx context.Context, client *graphql.Client, token string) (*q, e
 	var user q
 	q := fmt.Sprintf(`query {
 		user {
-			id,name,email,githubID,token,new,registry,buildkit,certificate
+			id,name,email,externalID,token,new,registry,buildkit,certificate
```
rberrelleza, line 133: this is going to break on all appliances until we release and upgrade everyone. I think we could catch the error and then do a second query with githubID (I think we already do that for other fields)

rberrelleza (approved):

pchico83:
@rberrelleza I need to test it more, but I think that logic is already in-place

rberrelleza:
> @rberrelleza I need to test it more, but I think that logic is already in-place

yes, I think you're right.

PR #1102, opened 2020-10-15
fix: quote ssh options

### LINKED ISSUE #1101: invalid ssh configuration when $HOME has spaces
**Describe the bug**

if the `$HOME` directory contains spaces then the generated ssh configuration file is invalid,

**To Reproduce**
have a space in your $HOME path
run ` okteto up -r 22100`
try to connect with SSH

**Expected behavior**
The IdentityFile parameter should be wrapped in double quotes

 - OS: Windows

freeman:
closes #1101 

Signed-off-by: Michel Rasschaert <michel.rasschaert@gmail.com>

### DIFF
--- a/pkg/ssh/ssh.go
+++ b/pkg/ssh/ssh.go
@@ -48,7 +48,7 @@ func add(path string, name string, port int) error {
 		newParam(portKeyword, []string{strconv.Itoa(port)}, nil),
 		newParam(strictHostKeyCheckingKeyword, []string{"no"}, nil),
 		newParam(userKnownHostsFileKeyword, []string{"/dev/null"}, nil),
-		newParam(identityFile, []string{privateKey}, nil),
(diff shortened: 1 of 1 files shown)


freeman:
my PR is probably not using the right approach as it will modify existing arguments of the SSH config which breaks the test

should I make the change here instead: https://github.com/okteto/okteto/blob/master/pkg/ssh/ssh.go#L51 ?

pchico83:
> my PR is probably not using the right approach as it will modify existing arguments of the SSH config which breaks the test
> 
> should I make the change here instead: https://github.com/okteto/okteto/blob/master/pkg/ssh/ssh.go#L51 ?

Thanks for your contribution @freeman !

Yes, I think it is better there. We can test on Linux/Mac before merging the PR.

pchico83 (approved):
Awesome @freeman!

PR #1109, opened 2020-10-19
OKTETO_HOME to override .okteto folder not $HOME

### LINKED ISSUE #1104: OKTETO_HOME should point to the okteto folder and not the HOME directory

I would like to relocate my OKTETO folder outside my home directory

Currently if I change the OKTETO_HOME environment variable, this relocates the `.okteto` folder but also makes all the logic to find `.ssh/config` and `.kube/config` inside the OKTETO_HOME folder.

Currently `OKTETO_HOME` is used to determine the HOME directory:
https://github.com/okteto/okteto/blob/master/pkg/config/config.go#L89

I think OKTETO_HOME should instead be used in the `GetOktetoHome` function
https://github.com/okteto/okteto/blob/master/pkg/config/config.go#L49

None

**Additional context**

rberrelleza:
Fixes #1104

## Proposed changes
- OKTETO_HOME now points to the replacement for the .okteto folder
- Added an OKTETO_USERHOME for those who want to replace the $HOME directory (we've run into this in the past with minimalistic setups)

This is a breaking change, so we need to document it in the release notes.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -78,21 +78,6 @@ func GetDeploymentHome(namespace, name string) string {
 	return d
 }
 
-// GetStateFile returns the path to the state file
-func GetStateFile(namespace, name string) string {
-	return filepath.Join(GetDeploymentHome(namespace, name), "okteto.state")
-}
-
-// GetSyncthingInfoFile returns the path to the syncthing info file
-func GetSyncthingInfoFile(namespace, name string) string {
-	return filepath.Join(GetDeploymentHome(namespace, name), "syncthing.info")
-}
-
-// GetSyncthingLogFile returns the path to the syncthing log file
-func GetSyncthingLogFile(namespace, name string) string {
-	return filepath.Join(GetDeploymentHome(namespace, name), "syncthing.log")
-}
-
 // GetUserHomeDir returns the OS home dir
 func GetUserHomeDir() string {
 	if v, ok := os.LookupEnv("OKTETO_USERHOME"); ok {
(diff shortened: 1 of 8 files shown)


pchico83:
@rberrelleza is the vscode plugin using `$OKTETO_HOME`?

pkg/config/config.go
```diff
@@ -86,7 +95,11 @@ func GetSyncthingLogFile(namespace, name string) string {
 
 // GetUserHomeDir returns the OS home dir
 func GetUserHomeDir() string {
-	if v, ok := os.LookupEnv("OKTETO_HOME"); ok {
+	if v, ok := os.LookupEnv("OKTETO_USERHOME"); ok {
```
pchico83, line 98: what is the use case for `OKTETO_USERHOME`?

rberrelleza: when we can't determine the path to the $HOME directory. that was the reason why we put OKTETO_HOME in the first place.

pchico83: but why would a user set `OKTETO_USERHOME` instead of `OKTETO_HOME`?

rberrelleza: we need to know where's the $HOME directory to be able to write the ssh configuration and kubeconfig files

pchico83 (approved):

rberrelleza:
> @rberrelleza is the vscode plugin using `$OKTETO_HOME`?

no, I don't think so

pkg/config/config.go
```diff
@@ -47,6 +48,14 @@ func GetBinaryFullPath() string {
 
 // GetOktetoHome returns the path of the okteto folder
 func GetOktetoHome() string {
+	if v, ok := os.LookupEnv("OKTETO_HOME"); ok {
```
pchico83, line 51: Could we call this one `OKTETO_FOLDER` and keep the current behavior for `OKTETO_HOME` to avoid breaking changes?

rberrelleza: sure

PR #1118, opened 2020-10-21
recreate context on every up iteration

rberrelleza:
Since the kubernetes calls now require a context, we should recreate the context after every `cancel` signal. I think it's better to remove ctx and cancel from the `upContext` struct, so the scope of each is clearer

### DIFF
--- a/cmd/up/up.go
+++ b/cmd/up/up.go
@@ -210,7 +210,7 @@ func (up *upContext) start(autoDeploy, build bool) error {
 		return fmt.Errorf("'okteto up' is not allowed in the current namespace")
 	}
 
-	up.IsOktetoNamespace = namespaces.IsOktetoNamespace(ns)
+	up.isOktetoNamespace = namespaces.IsOktetoNamespace(ns)
 
 	if err := createPIDFile(up.Dev.Namespace, up.Dev.Name); err != nil {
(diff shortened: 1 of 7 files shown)


cmd/up/up.go
```diff
@@ -258,11 +259,15 @@ func (up *upContext) activate(autoDeploy, build bool) {
 	isRetry := false
 
 	for {
+		// create a new context on every iteration
+		ctx, cancel := context.WithCancel(context.Background())
```
rberrelleza, line 263: this is the main change. we need to create the context on every iteration, since it is canceled at this level when there's an error and we retry.

cmd/up/up.go
```diff
@@ -258,11 +259,15 @@ func (up *upContext) activate(autoDeploy, build bool) {
...
+		defer up.shutdown(cancel)
```
pchico83, line 264: this is going to be called twice if there is a retry (one before retrying, and another one when exiting the process). could it be an issue?

rberrelleza: I don't think it gets called twice. In my testing (I did okteto up, and then killed the pod), it only got called once.

rberrelleza: I added a flag to ensure shutdown only runs once, just in case.

pchico83: If I remember well, `shutdown` needs to be called at the end of every loop, to clean the previous loop resources.

pchico83 (approved):

PR #1129, opened 2020-10-22
Remove environment variables from okteto manifest

pchico83:

## Proposed changes
- Information like env vars could be sensitive


rberrelleza:
could we instead remove the values of the env var?

PR #1136, opened 2020-10-26
Support for field "interface"

pchico83:

## Proposed changes
- Improve port error messages


rberrelleza (approved):
In past PRs you mentioned that we have users using this ports. I think that binding to a lower port is still possible on kubectl

PR #1143, opened 2020-10-31
Retry on transient

### LINKED ISSUE #1131: Improve reconnection logic after laptop resumes after sleeping
**Describe the bug**
The `okteto up` sequence is executed before there is internet connection, getting this error:
```
read: connection reset by peer
```

rberrelleza:
Fixes #1131

## Proposed changes
- Add a retry mechanism to get namespace and get deployment 
- Retry up to 3 times (of 30 seconds each) after a disconnect event.

### DIFF
--- a/cmd/up/up.go
+++ b/cmd/up/up.go
@@ -269,7 +269,6 @@ func (up *upContext) activateLoop(autoDeploy, build bool) {
 }
 
 func (up *upContext) activate(isRetry, autoDeploy, build bool) error {
-	// create a new context on every iteration
 	ctx, cancel := context.WithCancel(context.Background())
 	up.Cancel = cancel
 	up.Canceled = false
@@ -278,6 +277,8 @@ func (up *upContext) activate(isRetry, autoDeploy, build bool) error {
 	up.Disconnect = make(chan error, 1)
 	up.CommandResult = make(chan error, 1)
 	up.cleaned = make(chan string, 1)
+
+	// check if I can talk to the cluster
 
 	d, create, err := up.getCurrentDeployment(ctx, autoDeploy, isRetry)
 	if err != nil {
@@ -397,7 +398,7 @@ func (up *upContext) getCurrentDeployment(ctx context.Context, autoDeploy, isRet
 	}
 
 	if !errors.IsNotFound(err) || isRetry {
-		return nil, false, fmt.Errorf("couldn't get deployment %s/%s, please try again: %s", up.Dev.Namespace, up.Dev.Name, err)
+		return nil, false, fmt.Errorf("couldn't get deployment %s/%s, please try again: %w", up.Dev.Namespace, up.Dev.Name, err)
 	}
 
 	if len(up.Dev.Labels) > 0 {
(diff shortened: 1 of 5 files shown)


cmd/up/up.go
```diff
@@ -252,24 +252,36 @@ func (up *upContext) start(autoDeploy, build bool) error {
...
 			if err == errors.ErrLostSyncthing {
 				isRetry = true
 				continue
-			} else {
-				up.Exit <- err
-				return
 			}
+
+			if errors.IsTransient(err) {
+				networkRetries++
```
rberrelleza, line 265: when do you think we should reset this counter? as is, it will do up to 3 times in the entire lifetime of the up. I think it would be better if we reset to 0 after a good connection.

cmd/up/up.go
```diff
@@ -354,9 +366,12 @@ func (up *upContext) activate(isRetry, autoDeploy, build bool) error {
 	}()
 
 	prevError := up.waitUntilExitOrInterrupt()
+	if err != nil {
+		log.Infof("command finished with error: %s", prevError)
+	}
 
 	if up.shouldRetry(ctx, prevError) {
-		log.Yellow("\nConnection lost to your development container, reconnecting...\n")
+		log.Yellow("\nConnection lost to your development container, reconnecting...")
```
rberrelleza, line 374: the console still looks wrong after a disconnect. I'm going to try and  fix that as well.

pkg/errors/errors.go
```diff
@@ -73,6 +73,22 @@ func IsNotFound(err error) bool {
...
 }
 
+// IsTransient returns true if err represents a transient error
+func IsTransient(err error) bool {
+	if err == nil {
+		return false
+	}
+
+	switch {
+	case strings.Contains(err.Error(), "operation time out"),
```
rberrelleza, line 83: these are the errors I've seen when the network is not ready, or offline (e.g. when changing networks)

cmd/up/up.go
pchico83: I would retry forever, and add a warning message telling okteto is waiting for network connectivity. Most app reconnects as soon as network is available

pkg/k8s/namespaces/crud.go
```diff
@@ -43,11 +47,24 @@ func IsOktetoAllowed(ns *apiv1.Namespace) bool {
 }
 
 // Get returns the namespace object of ns
-func Get(ctx context.Context, ns string, c *kubernetes.Clientset) (*apiv1.Namespace, error) {
-	n, err := c.CoreV1().Namespaces().Get(ctx, ns, metav1.GetOptions{})
+func Get(ctx context.Context, name string, c *kubernetes.Clientset) (*apiv1.Namespace, error) {
+
+	var namespace *apiv1.Namespace
+
+	err := retry.OnError(config.DefaultBackoff, errors.IsTransient, func() error {
```
pchico83, line 54: why are we special casing this api call? shouldn't we fail and let the network retry handle this case?

cmd/up/up.go
rberrelleza: Sounds good. Let's put a message on every retry cycle so it doesn't look like it's just frozen.

pkg/k8s/namespaces/crud.go
rberrelleza: I think it's a good idea in general to retry network connections where there's less impact to the overall behavior. Some times the failure is due to a blip in the network, so there's no benefit in having to go all the way back to the top and starting again when a quick second get just fixes it.

 I wanted to start with this one because doing several gets has no side effect.  If we see that this pattern results in a better experience, then we can expand it to other calls. If it makes things more unreliable or complicated, we can go back to the other pattern.

PR #1150, opened 2020-11-04
Wait for shutdown sequence before a new retry

pchico83:

## Proposed changes
- Wait for shutdown sequence before a new retry
- Reduce frequency of reconnecting messages

### DIFF
--- a/cmd/up/up.go
+++ b/cmd/up/up.go
@@ -255,22 +255,37 @@ func (up *upContext) start(autoDeploy, build bool) error {
 // activateLoop activates the development container in a retry loop
 func (up *upContext) activateLoop(autoDeploy, build bool) {
 	isRetry := false
-	t := time.NewTicker(3 * time.Second)
+	isTransientError := false
+	t := time.NewTicker(1 * time.Second)
+	iter := 0
 	defer t.Stop()
 
 	for {
+		if isRetry || isTransientError {
+			log.Infof("waiting for shutdown sequence to finish")
+			<-up.Canceled
+			if iter == 0 {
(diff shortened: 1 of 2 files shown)


cmd/up/up.go
```diff
@@ -255,22 +255,37 @@ func (up *upContext) start(autoDeploy, build bool) error {
...
 	for {
+		if isRetry || isTransientError {
+			log.Infof("waiting for shutdown sequence to finish")
+			<-up.Canceled
+			if iter == 0 {
+				log.Yellow("Connection lost to your development container, reconnecting...")
+			}
+			iter++
+			iter = iter % 10
+			if isTransientError {
```
rberrelleza, line 272: can we also add the log.Yellow here? or maybe use the same animation pattern that we do for the up? I'd like to make it visible that we're trying to reconnect rather than being stuck.

@rlamana what do you think?

cmd/up/up.go
```diff
@@ -957,6 +965,8 @@ func (up *upContext) shutdown() {
 	}
 
 	log.Info("completed shutdown sequence")
+	up.Canceled <- true
```
rberrelleza, line 968: can we rename this to up.ShutdownCompleted (or somethign like that), since this could happen on a regular shutdown as well, and not just when canceling?

rberrelleza (approved):

cmd/up/up.go
pchico83: @rberrelleza the message is shown every 10s

cmd/up/up.go
pchico83: ok

PR #1151, opened 2020-11-05
Default to ssh forward and command execution

pchico83:

## Proposed changes
- Default to ssh forward and command execution

### DIFF
--- a/cmd/up/up.go
+++ b/cmd/up/up.go
@@ -975,9 +975,6 @@ func printDisplayContext(dev *model.Dev) {
 	}
 	log.Println(fmt.Sprintf("    %s %s", log.BlueString("Namespace:"), dev.Namespace))
 	log.Println(fmt.Sprintf("    %s      %s", log.BlueString("Name:"), dev.Name))
-	if dev.RemoteModeEnabled() {
-		log.Println(fmt.Sprintf("    %s       %d -> %d", log.BlueString("SSH:"), dev.RemotePort, dev.SSHServerPort))
-	}
 
 	if len(dev.Forward) > 0 {
 		log.Println(fmt.Sprintf("    %s   %d -> %d", log.BlueString("Forward:"), dev.Forward[0].Local, dev.Forward[0].Remote))
--- a/pkg/k8s/deployments/translate_test.go
+++ b/pkg/k8s/deployments/translate_test.go
@@ -205,7 +205,7 @@ services:
 							Image:           "web:latest",
 							ImagePullPolicy: apiv1.PullAlways,
 							Command:         []string{"/var/okteto/bin/start.sh"},
-							Args:            []string{"-s", "remote:/remote"},
+							Args:            []string{"-r", "-s", "remote:/remote"},
 							WorkingDir:      "/app",
 							Env: []apiv1.EnvVar{
 								{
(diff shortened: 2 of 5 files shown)


cmd/up/up.go
```diff
@@ -975,9 +931,7 @@ func printDisplayContext(dev *model.Dev) {
 	}
 	log.Println(fmt.Sprintf("    %s %s", log.BlueString("Namespace:"), dev.Namespace))
 	log.Println(fmt.Sprintf("    %s      %s", log.BlueString("Name:"), dev.Name))
-	if dev.RemoteModeEnabled() {
-		log.Println(fmt.Sprintf("    %s       %d -> %d", log.BlueString("SSH:"), dev.RemotePort, dev.SSHServerPort))
-	}
+	log.Println(fmt.Sprintf("    %s       %d -> %d", log.BlueString("SSH:"), dev.RemotePort, dev.SSHServerPort))
```
pchico83, line 934: @rberrelleza would you remove this line? at the end, it is more like an implementation detail

rberrelleza: I guess. but we should think about documenting how to access the dev env with SSH tools too (there's an issue opened on showing the path to the SSH key we use for this)

cmd/exec.go
```diff
@@ -107,25 +106,21 @@ func executeExec(ctx context.Context, dev *model.Dev, args []string) error {
...
-		log.Infof("executing remote command over SSH")
-		if dev.RemotePort == 0 {
-			p, err := ssh.GetPort(dev.Name)
-			if err != nil {
-				log.Infof("failed to get the SSH port for %s: %s", dev.Name, err)
-				return errors.UserError{
-					E:    fmt.Errorf("development mode is not enabled on your deployment"),
-					Hint: "Run `okteto up` to enable it and try again",
-				}
+	log.Infof("executing remote command over SSH")
```
rberrelleza, line 109: remove this log since this is always the case.

cmd/exec.go
```diff
@@ -107,25 +106,21 @@ func executeExec(ctx context.Context, dev *model.Dev, args []string) error {
 		dev.Container = p.Spec.Containers[0].Name
 	}
 
-	if dev.RemoteModeEnabled() {
```
rberrelleza, line 110: can we leave a flag (or env var) to enable the old behavior for now? It would be good to check with some of the heavy users to see if they have any issues or concerns with this shift.

pkg/k8s/deployments/translate_test.go
```diff
@@ -542,6 +548,12 @@ persistentVolume:
 									MountPath: "/var/syncthing",
 									SubPath:   model.SyncthingSubPath,
 								},
+								{
+									Name:      dev.GetVolumeName(),
+									ReadOnly:  false,
+									MountPath: model.RemoteMountPath,
+									SubPath:   model.RemoteSubPath,
+								},
```
rberrelleza, line 556: what is this for?

rberrelleza (approved):
I think we should leave the exec code for now in there for now, make `ssh` the default,  and have an env var to swap the exec behavior from `ssh to  `kubectl exec` in case we break someone's use case. In a month or so we can remove it if we haven't seen any issues.

PR #1154, opened 2020-11-05
"okteto stack deploy" and "okteto up" building okteto registry images if they don't exist

### LINKED ISSUE #1098: "okteto stack deploy --build" should be the default behavior
Deploying without building could be supported with the `--no-build` flag.

pchico83:
Fixes #1098


cmd/up/up.go
```diff
@@ -328,6 +329,13 @@ func (up *upContext) activate(isRetry, autoDeploy, build bool) error {
 		}
 	}
 
+	if _, err := registry.GetImageTagWithDigest(ctx, up.Dev.Image.Name); err != nil {
+		log.Infof("error accessing the image %s: %s", up.Dev.Image.Name, err.Error())
```
rberrelleza, line 333: can we log this only if it's not found?

pkg/registry/image.go
```diff
@@ -0,0 +1,64 @@
...
+package registry
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/okteto/okteto/pkg/model"
+)
+
+//GetRepoNameAndTag returns the image name without the tag
```
rberrelleza, line 23: can you document what the two return params are?

pkg/cmd/build/build.go
```diff
@@ -43,15 +39,23 @@ func Run(ctx context.Context, buildKitHost string, isOktetoCluster bool, path, d
 		defer os.Remove(processedDockerfile)
 	}
 
-	tag, err = expandOktetoDevRegistry(ctx, tag)
+	expandedTag, err := registry.ExpandOktetoDevRegistry(ctx, tag)
 	if err != nil {
 		return "", err
 	}
+	if tag != expandedTag {
+		log.Information("'%s' expanded to '%s'.", tag, expandedTag)
```
rberrelleza, line 47: do you think this information is useful to the user? I think it's better to not even show it, since we want to "dissapear" it behind `okteto.dev`

pkg/cmd/stack/translate.go
```diff
@@ -102,7 +102,7 @@ func translateEnvFile(svc *model.Service, filename string) error {
 	return nil
 }
 
-func translateBuildImages(ctx context.Context, s *model.Stack, noCache bool) error {
+func translateBuildImages(ctx context.Context, s *model.Stack, forceBuild, noCache bool) error {
 	c, _, configNamespace, err := k8Client.GetLocal("")
```
rberrelleza, line 106: should we consider the context of `okteto.yml` here?

pkg/cmd/stack/translate.go
```diff
@@ -126,14 +126,28 @@ func translateBuildImages(ctx context.Context, s *model.Stack, noCache bool) err
...
 	for name, svc := range s.Services {
 		if svc.Build == nil {
 			continue
 		}
-		oneBuild = true
-		imageTag := build.GetImageTag(svc.Image, name, s.Namespace, oktetoRegistryURL)
+		mustBuild := forceBuild
+		imageTag, err := registry.GetImageTagWithDigest(ctx, svc.Image)
+		if err != nil {
+			log.Infof("error accessing the image %s: %s", svc.Image, err.Error())
```
rberrelleza, line 138: can we log this only if it's not a not-found error?

pkg/registry/client.go
```diff
@@ -0,0 +1,43 @@
...
+	"strings"
+
+	"github.com/heroku/docker-registry-client/registry"
+	"github.com/okteto/okteto/pkg/log"
+)
+
+// New creates a new Registry with the given URL and credentials, then Ping()s it
+// before returning it to verify that the registry is available.
+func New(registryURL, username, password string) (*registry.Registry, error) {
+	transport := http.DefaultTransport
```
rberrelleza, line 27: can you add timeouts to this? I think the default transport doesn't have any.

pkg/registry/registry.go
```diff
@@ -0,0 +1,92 @@
...
+	expandedTag, err := ExpandOktetoDevRegistry(ctx, imageTag)
+	if err != nil {
+		log.Infof("error expanding okteto registry: %s", err.Error())
+		return imageTag, nil
+	}
+	if !strings.HasPrefix(expandedTag, registryURL) {
+		return imageTag, nil
+	}
+	username := okteto.GetUserID()
+	token, err := okteto.GetToken()
```
rberrelleza, line 44: is the GetImageTagWithDigest function only called for okteto managed images and registry?

pkg/registry/registry.go
```diff
@@ -0,0 +1,92 @@
...
+	if err != nil {
+		log.Infof("error getting token: %s", err.Error())
+		return imageTag, nil
+	}
+	c, err := New("https://"+registryURL, username, token.Token)
```
rberrelleza, line 49: Can you rename the `New` function to `NewRegistryClient`?

rberrelleza: Instead of  `"https://"+registryURL` would it be worth to parse as a URL and set the scheme instead?

pkg/registry/registry.go
```diff
@@ -0,0 +1,92 @@
...
+	if err != nil {
+		log.Infof("error creating registry client: %s", err.Error())
+		return imageTag, nil
+	}
+
+	repoURL, tag := GetRepoNameAndTag(expandedTag)
+	repoName := repoURL[strings.IndexRune(repoURL, '/')+1:]
```
rberrelleza, line 56: could this panic if  the repo URL is malformed?

pkg/registry/registry.go
```diff
@@ -0,0 +1,92 @@
...
+	digest, err := c.ManifestDigest(repoName, tag)
+	if err != nil {
+		if strings.Contains(err.Error(), "status=404") {
+			return "", errors.ErrNotFound
+		}
+		return "", fmt.Errorf("error getting image tag diggest: %s", err.Error())
+	}
+	return repoName + "@" + digest.String(), nil
```
rberrelleza, line 64: can you do fmt.Sprintf instead?

cmd/up/up.go
```diff
@@ -328,6 +329,13 @@ func (up *upContext) activate(isRetry, autoDeploy, build bool) error {
 		}
 	}
 
+	if _, err := registry.GetImageTagWithDigest(ctx, up.Dev.Image.Name); err != nil {
```
rberrelleza, line 332: how much time is the image analysis adding to `okteto up`, for the case where the image does exist?

pchico83: very little, it is just an http call if the image is in the okteto registry

pkg/cmd/stack/translate.go
pchico83: yes, but this is for stacks and they don't support `context` yet

pkg/registry/client.go
pchico83:
```
var DefaultTransport RoundTripper = &Transport{
	Proxy: ProxyFromEnvironment,
	DialContext: (&net.Dialer{
		Timeout:   30 * time.Second,
		KeepAlive: 30 * time.Second,
		DualStack: true,
	}).DialContext,
	ForceAttemptHTTP2:     true,
	MaxIdleConns:          100,
	IdleConnTimeout:       90 * time.Second,
	TLSHandshakeTimeout:   10 * time.Second,
	ExpectContinueTimeout: 1 * time.Second,
}
```

pkg/registry/registry.go
pchico83: no, but it would exit here:
https://github.com/okteto/okteto/pull/1154/files#diff-cb61d8f79451b9541de4a8cc0811523a68d15452b2f5971c7618ea5b423cf4ecR40

PR #1160, opened 2020-11-11
Improve errors if the ssh server is not accessible

pchico83:

## Proposed changes
- Added a new phase to check container connectivity
- Improve error when ssh server is not accessible
- Retry for 5s the first connection to the ssh server

Screenshot 2020-11-11 at 22 25 57


cmd/up/up.go
```diff
@@ -352,14 +352,18 @@ func (up *upContext) activate(isRetry, autoDeploy, build bool) error {
...
 	}
 
+	log.Success("Development container activated")
+
 	if err := up.forwards(ctx); err != nil {
-		return fmt.Errorf("couldn't forward traffic to your development container: %s", err.Error())
+		if err == errors.ErrSSHConnectError {
+			return up.checkOktetoStartError(ctx, "Failed to connect to your development container")
+		}
+		return fmt.Errorf("couldn't connect to your development container: %s", err.Error())
```
rberrelleza, line 361: is the output of `err.Error` useful? we can log just the text and throw the full error in a log.info

cmd/up/up.go
```diff
@@ -906,6 +887,35 @@ func (up *upContext) runCommand(ctx context.Context) error {
...
+	userID := pods.GetDevPodUserID(ctx, up.Dev, up.Client)
+	if up.Dev.PersistentVolumeEnabled() {
+		if userID != -1 && userID != *up.Dev.SecurityContext.RunAsUser {
+			return errors.UserError{
+				E:    fmt.Errorf("User %d doesn't have write permissions for the %s directory", userID, up.Dev.MountPath),
+				Hint: fmt.Sprintf("Set 'securityContext.runAsUser: %d' in your okteto manifest\n    After that, run 'okteto down -v' to reset your development container and run 'okteto up' again", userID),
+			}
+		}
+	} else {
+		if pods.OktetoDevPodMustBeRecreated(ctx, up.Dev, up.Client) {
```
rberrelleza, line 900: it's odd that the function called `checkOktetoStartError` changes the system. can we split the destruction into a separate stage? (or return a speficic error, and then the mail loop does the recreation?

pkg/ssh/pool.go
```diff
@@ -54,6 +56,33 @@ func startPool(ctx context.Context, serverAddr string, config *ssh.ClientConfig)
 	return p, nil
 }
 
+func retryNewClientConn(ctx context.Context, c net.Conn, addr string, conf *ssh.ClientConfig) (ssh.Conn, <-chan ssh.NewChannel, <-chan *ssh.Request, error) {
+	ticker := time.NewTicker(300 * time.Millisecond)
+	to := config.GetTimeout() / 6 // 5 seconds
```
rberrelleza, line 61: could we just keep it for the full 30s ? it's the worst case scenario right?

pkg/ssh/pool.go
```diff
@@ -54,6 +56,33 @@ func startPool(ctx context.Context, serverAddr string, config *ssh.ClientConfig)
...
+
+		if time.Now().After(timeout) {
+			return nil, nil, nil, err
+		}
+
+		select {
+		case <-ticker.C:
+			continue
+		case <-ctx.Done():
+			log.Infof("ssh.retryNewClientConn cancelled")
```
rberrelleza, line 80: done returns an err (or you can get it from the ctx), I find it's useful to log it to know the reason of the cancellation.

rberrelleza (approved):

cmd/up/up.go
pchico83: I find it useful. It is how it was before and people usually report this error without needing to send a doctor file

pkg/ssh/pool.go
pchico83: it is better to fail fast. the `ssh` server starts very fast, but I have found the first connection to fail a few times

pkg/ssh/pool.go
pchico83: ok

cmd/up/up.go
pchico83: ok

PR #1167, opened 2020-11-17
Add a retry error

pchico83:

## Proposed changes
- This handles the case where the ssh command is the one detecting the connectivity issue


rberrelleza (approved):

pkg/errors/errors.go
```diff
@@ -101,6 +101,7 @@ func IsTransient(err error) bool {
 	case strings.Contains(err.Error(), "operation time out"),
 		strings.Contains(err.Error(), "operation timed out"),
 		strings.Contains(err.Error(), "can't assign requested address"),
+		strings.Contains(err.Error(), "command exited without exit status or exit signal"),
```
rberrelleza, line 104: Can we catch this somewhere closer to the ssh connection code, and send back a specific error instead of handling it here? It would make it more readable.

Issue #1165, opened 2020-11-16
Handle syncthing insufficient space


pchico83:
**Is your feature request related to a problem? Please describe.**
`okteto up` hangs for ever, but the remote syncthing is logging this error:
```
2020-11-16T16:02:51.760438455Z time="2020-11-16T16:02:51Z" level=info msg="[ATOPH] 16:02:51 INFO: Error on folder \"1\" (okteto-1) changed: \"insufficient space on disk for database (/var/syncthing/index-v0.14.0.db): 1.0 % < 1 %\" -> \"insufficient space on disk for database (/var/syncthing/index-v0.14.0.db): 0.9 % < 1 %\"" process=syncthing
```

**Describe the solution you'd like**
Get a clear error message and indicate steps to fix the issue (increase persistent volume size).

pchico83:
Closed by #1168

PR #1168, opened 2020-11-17
Populate syncthing errors to the user

### LINKED ISSUE #1127: Syncthing errors not propagated properly
**Describe the bug**
I have an `insufficient space` error in syncthing, but I am getting `syncthing is unresponsive`.

**Expected behavior**
A clear error about the source of the issue.

pchico83:

Fixes #1127 #1165 

## Proposed changes
- Return a custom error with a CTA for insufficient space errors


pkg/syncthing/monitor.go
```diff
@@ -47,7 +47,7 @@ func (s *Syncthing) Monitor(ctx context.Context, disconnect chan error) {
 
 // MonitorStatus will send a message to disconnected if there is a synchronization error
 func (s *Syncthing) MonitorStatus(ctx context.Context, disconnect chan error) {
-	ticker := time.NewTicker(300 * time.Second)
+	ticker := time.NewTicker(30 * time.Second)
```
pchico83, line 50: Since we are not calling `status` API anymore, we can check for errors faster

cmd/up/up.go
```diff
@@ -960,6 +974,24 @@ func (up *upContext) getInteractive() bool {
 	return false
 }
 
+func (up *upContext) getInsufficientSpaceError(err error) error {
+	if up.Dev.PersistentVolumeEnabled() {
+		return errors.UserError{
+			E: err,
+			Hint: `Okteto volume is full
+    Increase your persistent volume size, run 'okteto down -v' and try 'okteto up' again
```
rberrelleza, line 982: are we using punctuation in the multi-line error messages? I think we should, since they are more explanation that error message.

pkg/errors/errors.go
```diff
@@ -52,15 +52,15 @@ var (
...
-	ErrUnknownSyncError = fmt.Errorf("Unknown syncthing error")
-
 	// ErrNotInDevContainer is returned when an unsupported command is invoked from a dev container (e.g. okteto up)
 	ErrNotInDevContainer = fmt.Errorf("this command is not supported from inside an development container")
 
 	// ErrResetSyncthing is raised when syncthing database must be reset
 	ErrResetSyncthing = fmt.Errorf("synchronization database corrupted")
 
+	// ErrInsufficientSpace is raised when syncthing fails with no space available
+	ErrInsufficientSpace = fmt.Errorf("synchronization failure with no space available")
```
rberrelleza, line 62:
```suggestion
	ErrInsufficientSpace = fmt.Errorf("there isn't enough disk space available to synchronize your files")
```

pkg/syncthing/syncthing.go
```diff
@@ -394,7 +393,7 @@ func (s *Syncthing) ResetDatabase(ctx context.Context, dev *model.Dev, local boo
 		if err != nil {
 			log.Infof("error posting 'rest/system/reset' local=%t syncthing API: %s", local, err)
 			if strings.Contains(err.Error(), "Client.Timeout") {
-				return errors.ErrUnknownSyncError
+				return fmt.Errorf("error resetting syncthing database local=%t: %s", local, err.Error())
```
rberrelleza, line 396: the  'local=' makes parsing the error harder. can we instead have a function that returns a message with the "local" or "remote" as part of the message?
```suggestion
				return fmt.Errorf("failed to reset syncthing database local=%t: %s", local, err.Error())
```

cmd/up/up.go
pchico83: We are using both, I am trying to unify that here. Would you add punctuation on every line?

rberrelleza: Yes.

rberrelleza: I'd use punctuation in the hint, and leave the error as is. specially in these cases where the hint is more of a paragraph.

cmd/up/up.go
```diff
@@ -960,6 +974,24 @@ func (up *upContext) getInteractive() bool {
...
+    More information about configuring your persistent volume at https://okteto.com/docs/reference/manifest#persistentvolume-object-optional`,
+		}
+	}
+	return errors.UserError{
+		E: err,
+		Hint: `The synchrinzation service is running out of space
+    Enable persistent volumes in your okteto manifest and try again
```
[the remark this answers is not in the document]
pchico83: I think so. do you have anything in mind?

pkg/syncthing/syncthing.go
pchico83: we follow this pattern in several places. would you mind opening an issue to track the change?

PR #1169, opened 2020-11-17
Automatically create ".stignore" file if does not exist

pchico83:
## Proposed changes
- Also, check if `.git` is defined in the `.stignore` file

### DIFF
--- a/cmd/up/stignore.go
+++ b/cmd/up/stignore.go
@@ -51,7 +51,7 @@ func checkStignoreConfiguration(dev *model.Dev) error {
 }
 
 func askIfCreateStignoreDefaults(folder, stignorePath, gitPath string) error {
-	log.Information("Okteto requires a '.stignore' file containing file patterns the synchrinization service should ignore.")
+	log.Information("Okteto requires a '.stignore' file containing file patterns the synchronization service should ignore.")
 	stignoreDefaults, err := utils.AskYesNo("    Do you want to infer defaults for the '.stignore' file? (otherwise, it will be left blank) [y/n] ")
 	if err != nil {
 		return fmt.Errorf("failed to add '.stignore' to '%s': %s", folder, err.Error())
--- a/cmd/up/stignore.go
+++ b/cmd/up/stignore.go
@@ -1,0 +1,107 @@
+// Copyright 2020 The Okteto Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package up
+
+import (
+	"fmt"
+	"io/ioutil"
+	"path/filepath"
+	"strings"
+
+	initCMD "github.com/okteto/okteto/cmd/init"
+	"github.com/okteto/okteto/cmd/utils"
+	"github.com/okteto/okteto/pkg/linguist"
+	"github.com/okteto/okteto/pkg/log"
+	"github.com/okteto/okteto/pkg/model"
+)
+
+func checkStignoreConfiguration(dev *model.Dev) error {
+	for _, folder := range dev.Syncs {
+		stignorePath := filepath.Join(folder.LocalPath, ".stignore")
+		gitPath := filepath.Join(folder.LocalPath, ".git")
+		if !model.FileExists(stignorePath) {
+			log.Infof("'.stignore' does not exist in folder '%s'", folder.LocalPath)
+			if err := askIfCreatingStignore(folder.LocalPath, stignorePath, gitPath); err != nil {
+				return err
+			}
+			continue
+		}
+
+		log.Infof("'.stignore' exists in folder '%s'", folder.LocalPath)
+		if !model.FileExists(gitPath) {
+			continue
+		}
+
+		if err := askIfUpdatingStignore(folder.LocalPath, stignorePath, gitPath); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func askIfCreatingStignore(folder, stignorePath, gitPath string) error {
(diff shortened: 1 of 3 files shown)


cmd/up/stignore.go
```diff
@@ -0,0 +1,107 @@
...
+		if err := askIfUpdatingStignore(folder.LocalPath, stignorePath, gitPath); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func askIfCreatingStignore(folder, stignorePath, gitPath string) error {
+	log.Information("It is recommended to define a '.stignore' file to avoid synchronizing things like build artifacts, git metadata or third party libraries.")
+	answer, err := utils.AskYesNo("    Do you want to automatically create a '.stignore' file? [y/n] ")
```
rberrelleza, line 55:
```suggestion
	answer, err := utils.AskYesNo("    Do you want to create a '.stignore' file? [y/n] ")
```

cmd/up/stignore.go
```diff
@@ -0,0 +1,107 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to ask for adding '.stignore' to '%s': %s", folder, err.Error())
```
rberrelleza, line 57:
```suggestion
		return fmt.Errorf("failed to add '.stignore' to '%s': %s", folder, err.Error())
```

cmd/up/stignore.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+
+	if !answer {
+		stignoreContent := ""
+		if model.FileExists(gitPath) {
```
rberrelleza, line 62: let's just put it all the time, it's pretty standard to have it as part of an ignore list.

cmd/up/stignore.go
```diff
@@ -0,0 +1,107 @@
...
+			stignoreContent = ".git\n"
+		}
+		if err := ioutil.WriteFile(stignorePath, []byte(stignoreContent), 0644); err != nil {
```
rberrelleza, line 65: if the user chose no as the answer,  why do we still create the ignore file?

cmd/up/stignore.go
```diff
@@ -0,0 +1,107 @@
...
+
+		if err := askIfUpdatingStignore(folder.LocalPath, stignorePath, gitPath); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func askIfCreatingStignore(folder, stignorePath, gitPath string) error {
+	log.Information("It is recommended to define a '.stignore' file to avoid synchronizing things like build artifacts, git metadata or third party libraries.")
```
rberrelleza, line 54:
```suggestion
	log.Information("It is recommended to define a '.stignore' file to avoid synchronizing unnecessary files like build artifacts, git metadata or third party libraries.")
```

cmd/up/stignore.go
```diff
@@ -0,0 +1,107 @@
...
+			return fmt.Errorf("failed to create empty '%s': %s", stignorePath, err.Error())
+		}
+		return nil
+	}
+
+	language, err := initCMD.GetLanguage("", folder)
+	if err != nil {
+		return fmt.Errorf("failed to get language for '%s': %s", folder, err.Error())
+	}
+	c := linguist.GetSTIgnore(language)
```
rberrelleza, line 75: let's make sure all of them include .git

cmd/up/stignore.go
```diff
@@ -0,0 +1,107 @@
...
+			}
+			continue
+		}
+
+		log.Infof("'.stignore' exists in folder '%s'", folder.LocalPath)
+		if !model.FileExists(gitPath) {
+			continue
+		}
+
+		if err := askIfUpdatingStignore(folder.LocalPath, stignorePath, gitPath); err != nil {
```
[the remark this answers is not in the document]
pchico83: I will handle to ask only once. This is important, I see lots of users without `.git` on their `.stignore` file, which makes `okteto up` very inestable.

cmd/up/stignore.go
pchico83: It servers to goals:
- only ask once
- more discoverability of the feature

rberrelleza: but it goes against what you're asking the user. if they are asked "should I create the stignore file", they say no, and the file is still created they are going to think there's a bug. if you want to always create the file, then don't ask them (I don't think this is right approach btw).

cmd/up/stignore.go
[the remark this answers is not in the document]
rberrelleza: do you know how this users are creating the .stignore? are they doing it by hand, or getting it from our samples / init command?

cmd/up/stignore.go
pchico83: what do you suggest?

cmd/up/stignore.go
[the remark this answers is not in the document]
pchico83: must be manually because we always include `.git`

cmd/up/stignore.go
rberrelleza: if the user says no, we don't create the file, and don't ask them again for that repo.

rberrelleza: I understand the painpoint, but we need to find a way to educate the user rather than ignore their answers. I think we should explain why synchronizing .git is a bad idea, but if they want to do it, it's their choice.

pchico83: Right, but the implementation of "don't ask them again for that repo" must live in the repo. If not, every time a user clones the repo will get prompted, and that feels wrong because okteto has been already configured.
I will rephrase the question to make it clear.

cmd/up/stignore.go
```diff
@@ -0,0 +1,107 @@
...
+			return err
+		}
+	}
+	return nil
+}
+
+func askIfCreateStignoreDefaults(folder, stignorePath, gitPath string) error {
+	log.Information("Okteto requires a '.stignore' file containing file patterns the synchrinization service should ignore.")
```
rlamana, line 54: typo:
```suggestion
	log.Information("Okteto requires a '.stignore' file containing file patterns the synchronization service should ignore.")
```

cmd/up/stignore.go
rlamana: It did look like a rephrasing issue. I like it better now, but maybe we should also state the benefits of doing so. What about something like: 
```Okteto requires a '.stignore' file to intentionally ignore file patterns that help optimize the synchronization service.``` ?

PR #1171, opened 2020-11-18
don't check if reverse is taken in the local machine

### LINKED ISSUE #1170: unable to reverse forward
**Describe the bug**
I'm trying to reverse forward traffic from development container localhost:8888 to my local machine 8888.
I already have a server running on my machine's localhost:8888 (which is what I want my development environment to connect to)

**To Reproduce**
1. Do `okteto up` with an okteto.yaml file containing something like:

```
reverse:
  - 8888:8888
```

output is:

```
couldn't forward traffic to your development container: local port 8888 is already in-use in your local machine
```
Maybe I'm miss-understanding the configuration; of course local port 8888 is already in-use which is what I want to connect to.
Is there any way to allow my development container to connect to my localhost's service?

**Expected behavior**
Expect `okteto up` to work and traffic to be forwarded

 - OS: OSX
 - Version: 10.15.2

rberrelleza:
Fixes #1170

## Proposed changes
- don't check if port is taken for the case of reverse forwards, since it's a valid scenario.

### DIFF
--- a/pkg/ssh/manager.go
+++ b/pkg/ssh/manager.go
@@ -57,6 +57,10 @@ func (fm *ForwardManager) canAdd(localPort int, checkAvailable bool) error {
 		return fmt.Errorf("port %d is listed multiple times, please check your forwards configuration", localPort)
 	}
 
+	if !checkAvailable {
+		return nil
+	}
+
 	if !model.IsPortAvailable(fm.localInterface, localPort) {
 		if localPort <= 1024 {
 			os := runtime.GOOS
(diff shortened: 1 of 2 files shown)


pkg/ssh/manager.go
```diff
@@ -48,7 +48,7 @@ func NewForwardManager(ctx context.Context, sshAddr, localInterface, remoteInter
 	}
 }
 
-func (fm *ForwardManager) canAdd(localPort int) error {
+func (fm *ForwardManager) canAdd(localPort int, checkAvailable bool) error {
```
pchico83, line 51: `checkAvailable` is not used, doesn't it?

pchico83 (approved):

pkg/ssh/manager.go
rberrelleza: I didn't push the last commit 🤦‍♂️

PR #1196, opened 2020-12-04
Enable file compression

pchico83:

## Proposed changes
- This can speed up the `up` command 50%, mainly for large folders

### DIFF
--- a/pkg/model/dev.go
+++ b/pkg/model/dev.go
@@ -63,6 +63,8 @@ const (
 	RemoteMountPath = "/var/okteto/remote"
 	//SyncthingSubPath subpath in the development container persistent volume for the syncthing data
 	SyncthingSubPath = "syncthing"
+	//DefaultSyncthingRescanInterval default syncthing re-scan interval
+	DefaultSyncthingRescanInterval = 300
 	//RemoteSubPath subpath in the development container persistent volume for the remote data
 	RemoteSubPath = "okteto-remote"
 	//OktetoAutoCreateAnnotation indicates if the deployment was auto generatted by okteto up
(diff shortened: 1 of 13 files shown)


rberrelleza (commented):
Most repos are going to have lots of small files rather than a few large ones. Will there be a benefit there, or the compress/uncompress time will eat away the benefits?

We could start with this as okteto.yaml parameter, so it's only enabled on repos with big files.

pchico83:
@rberrelleza yes, I was thinking about that too, good idea.
Should we go with `compression: true/false`?

rberrelleza:
> @rberrelleza yes, I was thinking about that too, good idea.
> Should we go with `compression: true/false`?

yeah, but can we put it inside of `sync`? feel like as a standalone it might be confusing. (not sure if you can do list + keys in yaml though).

pchico83:
@rberrelleza it cannot be inside of `sync`. We can do a new one, `syncProperties` or something like that

rberrelleza:
> @rberrelleza it cannot be inside of `sync`. We can do a new one, `syncProperties` or something like that

`syncSettings` and we can put the reload and a few other things there?

rlamana:
Why can't it be inside `sync`? Grouped in an `options` property under the `sync`?

pchico83:
> Why can't it be inside `sync`? Grouped in an `options` property under the `sync`?

that would look like this:
```
sync:
   - .:/app
   options:
      compression: true
```
which is not valid yaml and would be confusing. We would need to group the folders in a new property, like this:
```
sync:
   folders:
      - .:/app
   options:
      compression: true
```
but we would be complicating the syntax for 99% of cases

rlamana:
Ok, makes sense. Then at least `syncOptions` rather than `syncProperties`? 😬

rberrelleza:
could we support both shapes @pchico83 ? (similar to how the build attribute takes a string, or a dict?)

pchico83:
> could we support both shapes @pchico83 ? (similar to how the build attribute takes a string, or a dict?)

@rberrelleza do you mean supporint:
```
sync:
  - .:/app
```
and?
```
sync:
  folders:
    - .:/app
  options:
    compression: true
```

This is doable. I would go with this one.

rlamana:
+1!

pkg/syncthing/syncthing.go
```diff
@@ -168,9 +169,9 @@ func New(dev *model.Dev) (*Syncthing, error) {
 		hash = []byte("")
 	}
 
-	rescanInterval := os.Getenv("OKTETO_RESCAN_INTERVAL")
-	if rescanInterval == "" {
-		rescanInterval = "300"
+	compression := "metadata"
+	if dev.Sync.Compression {
+		compression = "always"
```
rberrelleza, line 174: are there other potential options here? if so, maybe dev.Sync.Compression shouldn't be a bool but an enum?

pkg/model/dev.go
```diff
@@ -462,6 +482,16 @@ func (dev *Dev) setDefaults() error {
 	}
 	dev.setRunAsUserDefaults(dev)
 
+	if os.Getenv("OKTETO_RESCAN_INTERVAL") != "" {
```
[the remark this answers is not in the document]
pchico83: We can deprecate it, but let's keep it for a while first

pkg/syncthing/syncthing.go
pchico83: There are other compression options, but I would keep it as a bool value for now to enable it or disable it

PR #1203, opened 2020-12-11
Propagate errors when translating envvars

pchico83:

## Proposed changes
- This was not intended to be ignored. An error here silently ignores the part where images are built


rberrelleza (approved):
Can we add a test to prevent this from regressing in the future?

PR #1216, opened 2020-12-18
Return all kubernetes available syncthing logs

pchico83:

## Proposed changes
- Return all kubernetes available syncthing logs in `okteto doctor` instead of truncating the last 1200 lines

### DIFF
--- a/pkg/k8s/pods/pod.go
+++ b/pkg/k8s/pods/pod.go
@@ -46,7 +46,7 @@ const (
 
 var (
 	devTerminationGracePeriodSeconds int64
-	tailLines                        int64 = 1200
+	limitBytes                       int64 = 5 * 1024 * 1024 // 5Mb
 )
 
 // GetBySelector returns the first pod that matches the selector or error if not found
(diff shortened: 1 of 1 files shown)


pkg/k8s/pods/pod.go
```diff
@@ -385,7 +384,6 @@ func containerLogs(ctx context.Context, container string, pod *apiv1.Pod, namesp
 	podLogOpts := apiv1.PodLogOptions{
 		Container:  container,
 		Timestamps: timestamps,
-		TailLines:  &tailLines,
```
rberrelleza, line 388: will this do paging behind the scenes?  syncthing can have a lot of lines, so it would be good to have some limit in the request about getting RESPONSE TOO BIG errors, and do a few gets to get it all.

pchico83: The logs api does not have pagination.
Logs size is limited by docker, and it is usually 50MB per container.

Would you limit the response to 50MB for example?

rberrelleza: That's a lot for a response body. that might even cause timeouts or  serialization issues in the API server. 

The problem is that, from a troubleshooting perspective, the first logs are very important (that's where we log the initial commands, the SSH server, syncthing config, etc), even more than the newer ones, which mostly show file sync.   There's no way to get all the logs chunk by chunk? (e.g say, 5mb at a time)

pchico83: No way to do chunks. We can get the first lines and limit the number of bytes.

pchico83: I limited it to 5Mb

PR #1228, opened 2020-12-23
This error is also thrown when the dev container is killed by "okteto…

pchico83:
… down"

### DIFF
--- a/pkg/errors/errors.go
+++ b/pkg/errors/errors.go
@@ -105,7 +105,6 @@ func IsTransient(err error) bool {
 		strings.Contains(err.Error(), "operation timed out"),
 		strings.Contains(err.Error(), "i/o timeout"),
 		strings.Contains(err.Error(), "can't assign requested address"),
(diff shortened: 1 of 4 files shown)


pkg/errors/errors.go
```diff
@@ -105,7 +105,6 @@ func IsTransient(err error) bool {
 		strings.Contains(err.Error(), "operation timed out"),
 		strings.Contains(err.Error(), "i/o timeout"),
 		strings.Contains(err.Error(), "can't assign requested address"),
-		strings.Contains(err.Error(), "command exited without exit status or exit signal"),
```
rberrelleza, line 108: is this the only reason why this error is thrown?

rberrelleza (approved):

pchico83:
@rberrelleza It can also be thrown by a connection glitch. I updated the code to differentiate between them.
`isRetry` was not set to `true` for transient errors.

PR #1232, opened 2020-12-29
Reuse same kubeconfig context for "okteto namespace" and "okteto name…

### LINKED ISSUE #1231: "okteto namespace" and "okteto namespace xxx" don't reuse kubeconfig context
**Describe the bug**
`okteto namespace` and `okteto namespace xxx` don't reuse kubeconfig context. They should create the same context, and just set the `namespace` field of the context as needed.

**Additional context**
After this change, deleting contexts on `okteto namespace delete` doesn't makes sense anymore, since the personal namespace cannot be deleted.

pchico83:
…space xxx"

Fixes #1231

### DIFF
--- a/.typo-ci.yml
+++ b/.typo-ci.yml
@@ -6,6 +6,7 @@ excluded_words:
   - Debugf
   - Warnf
   - Errorf
+  - Fatalf
   - apimachinery
   - unmarshal
   - UnmarshalYAML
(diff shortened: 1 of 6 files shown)


rberrelleza:
The idea with the current implementation is to have separate contexts per namespace, so it's easier to use them with other kubernetes tools without having to run "okteto namespace foo" every time. 

What's the purpose of changing that behavior?

pchico83:
Synched on issue #1231

PR #1233, opened 2020-12-29
Only inject mount key on Okteto Cloud

pchico83:

## Proposed changes
- Injecting a mount cache key degrades the cache performance on OE installations and hard multitenancy is not an issue in dedicated cluster environments.

### DIFF
--- a/pkg/cmd/build/build.go
+++ b/pkg/cmd/build/build.go
@@ -16,8 +16,10 @@ package build
 import (
 	"context"
 	"os"
+	"path/filepath"
 
 	"github.com/okteto/okteto/pkg/log"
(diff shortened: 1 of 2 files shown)


rberrelleza:
I'm not sure that the argument of "hard multi-tenancy is not a concern in dedicated clusters" is necesarily true. What kind of performance degradation are we seeing?

For OE clusters, could this be driven by a configuration setting?

pchico83:
@rberrelleza it cannot be configured on the cluster side, although we should move this logic as part of our buildkit fork (or a buildkit frontend).
Cache stop working between builds from different developers without this fix, which is highly impractical on OR installations.

PR #1255, opened 2021-01-30
Add pod affinities only to dev pods created for "services"

pchico83:
## Proposed changes
`okteto up` uses affinities to enforce all dev pods run on the same node, to make it possible to share a volume between the main dev pod, and the ones defined by "services".
Afinnities are only needed by pods defined by "services". Adding the affinities to the main dev container gives scheduling issues if other affinities are defined in the deployment under development.

### DIFF
--- a/pkg/model/translation.go
+++ b/pkg/model/translation.go
@@ -50,6 +50,11 @@ type TranslationRule struct {
 	Resources         ResourceRequirements `json:"resources,omitempty"`
 }
 
+//IsMainDevContainer returns true if the translation rule applies to the main dev container of the okteto manifest
+func (r *TranslationRule) IsMainDevContainer() bool {
+	return r.OktetoBinImageTag != ""
+}
+
 //VolumeMount represents a volume mount
 type VolumeMount struct {
 	Name      string `json:"name,omitempty"`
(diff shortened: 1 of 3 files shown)


pkg/model/translation.go
```diff
@@ -50,6 +50,11 @@ type TranslationRule struct {
 	Resources         ResourceRequirements `json:"resources,omitempty"`
 }
 
+//IsMainDevContainer returns true if the translation rule applies to the main dev container of the okteto manifest
+func (r *TranslationRule) IsMainDevContainer() bool {
+	return r.OktetoBinImageTag != ""
```
rberrelleza, line 55: Could we use a label to determine this? Checking on a specific string seems brittle and hard to understand for me.

rberrelleza (approved):

pkg/model/translation.go
pchico83: What do you mean by label in this context?

rberrelleza: A kubernetes label. We could also have a flag we set in the rules object to mark it as the 'main container' when we analyze the deployment?

rberrelleza (approved):

pkg/model/translation.go
pchico83: This is why I refactored it as a function. But changing the representation brings issues with backward compatibility. Let's keep it like this for now since it is unrelated to this PR

PR #1258, opened 2021-02-03
use syncthing 1.13.1

rberrelleza:
## Proposed changes
- use syncthing 1.13.1  for both client and bin image

Signed-off-by: Ramiro <rberrelleza@gmail.com>

### DIFF
--- a/images/bin/Makefile
+++ b/images/bin/Makefile
@@ -1,4 +1,4 @@
-tag = 1.2.18
+tag = 1.2.19
 
 .PHONY: push
 push:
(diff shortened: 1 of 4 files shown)


images/bin/Makefile
```diff
@@ -1,4 +1,4 @@
-tag = 1.2.18
+tag = 1.2.19
```
rberrelleza, line 1: I'll build the image  post merge.

pchico83: You are missing the change here:
https://github.com/okteto/okteto/blob/705b52ee2be1ae410cba21df52b7c720d953c03a/pkg/model/dev.go#L97

pchico83 (approved):

PR #1271, opened 2021-02-12
Support for destroying volumes on "okteto pipeline destroy"

pchico83:
## Proposed changes
- This provides feature parity with the Okteto UI

### DIFF
--- a/cmd/pipeline/destroy.go
+++ b/cmd/pipeline/destroy.go
@@ -30,6 +30,7 @@ func destroy(ctx context.Context) *cobra.Command {
 	var name string
 	var namespace string
 	var wait bool
+	var destroyVolumes bool
 	var timeout time.Duration
 
 	cmd := &cobra.Command{
@@ -55,14 +56,14 @@ func destroy(ctx context.Context) *cobra.Command {
 				}
 			}
 
-			if err := deletePipeline(ctx, name, namespace, wait, timeout); err != nil {
+			if err := deletePipeline(ctx, name, namespace, wait, destroyVolumes, timeout); err != nil {
 				return err
 			}
 
 			if wait {
-				log.Success("Pipeline '%s' deleted", name)
+				log.Success("Pipeline '%s' destroyed", name)
 			} else {
-				log.Success("Pipeline '%s' scheduled for deletion", name)
+				log.Success("Pipeline '%s' scheduled for destruction", name)
 			}
 
 			return nil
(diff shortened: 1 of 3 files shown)


cmd/pipeline/destroy.go
```diff
@@ -72,16 +73,17 @@ func destroy(ctx context.Context) *cobra.Command {
 	cmd.Flags().StringVarP(&name, "name", "p", "", "name of the pipeline (defaults to the folder name)")
 	cmd.Flags().StringVarP(&namespace, "namespace", "n", "", "namespace where the up command is executed (defaults to the current namespace)")
 	cmd.Flags().BoolVarP(&wait, "wait", "w", false, "wait until the pipeline finishes (defaults to false)")
+	cmd.Flags().BoolVarP(&destroyVolumes, "volumes", "v", false, "remove persistent volumes created by the pipeline (defaults to false)")
```
rberrelleza, line 76: `-d` or `-r` instead of `-v` since the action is destroy volumes or remove volumes? (not sure if we have something similar in other commands)

rberrelleza, line 76:
```suggestion
	cmd.Flags().BoolVarP(&destroyVolumes, "volumes", "v", false, "destroy persistent volumes created by the pipeline (defaults to false)")
```

pkg/okteto/pipeline.go
```diff
@@ -88,13 +88,13 @@ func GetPipelineByName(ctx context.Context, name, namespace string) (*PipelineRu
 }
 
 // DeletePipeline deletes a pipeline
-func DeletePipeline(ctx context.Context, name, namespace string) (string, error) {
+func DeletePipeline(ctx context.Context, name, namespace string, destroyVolumes bool) (string, error) {
 	log.Infof("delete pipeline: %s/%s", namespace, name)
 	q := fmt.Sprintf(`mutation{
-		destroyGitRepository(name: "%s", space: "%s"){
+		destroyGitRepository(name: "%s", space: "%s", destroyVolumes: %t){
```
rberrelleza, line 94: do we need to care about API compatibility here?

rberrelleza (approved):

cmd/pipeline/destroy.go
pchico83: yes, we are using `-v` in other places.
also, people are already familiar with `-v` usage in docker/docker-compose CLIs

pkg/okteto/pipeline.go
pchico83: Yes, better to be safe

PR #1273, opened 2021-02-12
Unify how sync status is computed between "okteto status" and "okteto…

pchico83:
… up"

People get confused by showing two sync statuses that don't match.

## Proposed changes
- The idea is to compute the status based on one-way sync if dual-sync is not happening

### DIFF
--- a/pkg/cmd/status/run.go
+++ b/pkg/cmd/status/run.go
@@ -37,17 +37,22 @@ func Run(ctx context.Context, dev *model.Dev, sy *syncthing.Syncthing) (float64,
 		log.Infof("error accessing remote syncthing status: %s", err)
 		return 0, fmt.Errorf("error accessing remote syncthing status")
 	}
-	if progressLocal == 100 && progressRemote == 100 {
-		return 100, nil
+
+	return computeProgress(progressLocal, progressRemote), nil
+}
+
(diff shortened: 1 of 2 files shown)


pkg/cmd/status/run.go
```diff
@@ -37,8 +37,22 @@ func Run(ctx context.Context, dev *model.Dev, sy *syncthing.Syncthing) (float64,
...
+
+	return computeProgress(progressLocal, progressRemote), nil
+}
+
+func computeProgress(local, remote float64) float64 {
+	if local == 100 && remote == 100 {
+		return 100
+	}
+
+	if local == 100 {
```
rberrelleza, line 49: shouldn't this be 100 + remote /2?  otherwise you could run into hypothetical cases like goin from 80% (when both local and remote are 40 each) to  40% (when local is 100 and remote is still 40) no?

rberrelleza (approved):

pkg/cmd/status/run.go
pchico83: yes, it is to compute it the same way as the first `up`.  although assuming that either a user is synching in one direction or the other is a good compromise. But you are right, if the user syncs both ways at the same time it could be flacky

PR #1276, opened 2021-02-13
Backward compatibility if none of "workdir" and "sync" is defined

pchico83:
## Proposed changes
Currently, manifests of this kind are broken:
```
name: guestbook
image: okteto/python:3
command: bash
```

### DIFF
--- a/pkg/model/volumes.go
+++ b/pkg/model/volumes.go
@@ -23,6 +23,9 @@ import (
 )
 
 func (dev *Dev) translateDeprecatedVolumeFields() error {
+	if dev.WorkDir == "" && dev.MountPath == "" && len(dev.Sync.Folders) == 0 {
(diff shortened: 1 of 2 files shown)


pkg/model/volumes.go
```diff
@@ -23,6 +23,9 @@ import (
 )
 
 func (dev *Dev) translateDeprecatedVolumeFields() error {
+	if dev.WorkDir == "" && dev.MountPath == "" && len(dev.Sync.Folders) == 0 {
+		dev.WorkDir = "/okteto"
```
rberrelleza, line 27: Will this also make /okteto the sync folder?

rberrelleza (approved):

pkg/model/volumes.go
pchico83: Yeap

PR #1277, opened 2021-02-13
Refresh k8s credentials for okteto managed namespaces

pchico83:

## Proposed changes
This PR implements a k8s credentials refresh when getting a k8s client to a namespace managed by okteto.
It opens the door to implement a rotation of k8s credentials.

The implementation uses a k8s transport that intercept connection errors to the API server and refresh the k8s credentials if the namespace is manged by okteto

### DIFF
--- a/cmd/namespace/namespace.go
+++ b/cmd/namespace/namespace.go
@@ -16,8 +16,6 @@ package namespace
 import (
 	"context"
 	"fmt"
-	"net/url"
-	"strings"
 
 	"github.com/okteto/okteto/cmd/utils"
 	"github.com/okteto/okteto/pkg/analytics"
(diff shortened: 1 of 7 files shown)


rberrelleza:
I don't think introducing a delay in every command for something that is going to be required only once every few months is a good usability pattern.

Could we instead catch an error or something, and then reload the creds? We could also follow the pattern oauth has, and have some sort of expiration on the credentials (a week, a month?) and only reload when they are expired.

pchico83:
> I don't think introducing a delay in every command for something that is going to be required only once every few months is a good usability pattern.
> 
> Could we instead catch an error or something, and then reload the creds? We could also follow the pattern oauth has, and have some sort of expiration on the credentials (a week, a month?) and only reload when they are expired.

Yes, and we already have a few sequential calls before starting the `up` sequence that we should optimize.
I am going to give it a try to the wrapper approach, something similar to this:
https://github.com/topfreegames/go-extensions-k8s-client-go

PR #1281, opened 2021-02-16
Remove syncthing process based on name

### LINKED ISSUE #1204: "okteto up" fails with another syncthing instance running error
**Describe the bug**

`okteto up` fails and the local syncthing logs "Error opening database: resource temporarily unavailable (is another instance of Syncthing running?)".

**Expected behavior**
`okteto up` should be able to recover from this error.

pchico83:
Fixes #1204


pkg/syncthing/process.go
```diff
@@ -16,22 +14,47 @@
...
+	"github.com/shirou/gopsutil/process"
 )
 
-func terminate(pid int, wait bool) error {
-	proc := os.Process{Pid: pid}
-	if err := proc.Signal(os.Interrupt); err != nil {
-		if strings.Contains(err.Error(), "process already finished") {
-			return nil
-		}
+func getParentExe(p *process.Process) string {
```
rberrelleza, line 23: if this just for windows? If this code depends on windows components, you need to put it in the  process_windows.go folder, so it doesn't get compiled on Linux or macos (it's based on build tag at the top of the file)

pkg/syncthing/syncthing.go
```diff
@@ -702,32 +668,34 @@ func (s *Syncthing) Restart(ctx context.Context) error {
...
+	for _, p := range pList {
+		exe, err := p.Exe()
+		if err != nil {
+			log.Infof("error getting procress exec: %s", err.Error())
+			continue
+		}
+		if exe != getInstallPath() {
+			continue
+		}
+		if getParentExe(p) == getInstallPath() {
```
rberrelleza, line 685: what is parentexe on this context?

pkg/syncthing/syncthing.go
```diff
@@ -702,32 +668,34 @@ func (s *Syncthing) Restart(ctx context.Context) error {
...
-	if !force {
-		if pid != s.pid {
-			log.Infof("syncthing pid-%d wasn't created by this command, skipping", pid)
-			return nil
-		}
-	}
-
-	if err := s.cleanupDaemon(pid, force); err != nil {
+func (s *Syncthing) Stop(wait bool) error {
+	pList, err := process.Processes()
```
rberrelleza, line 672: in my experience, listing processes and waiting on them is a bit flaky and time consuming, do you think we should only do this if the pidPath is not valid?

rberrelleza: can you also validate if process.Processes() works when you're not an admin on the machine? (it's something common on windows environments)

pkg/syncthing/process.go
rberrelleza: Is not clear what 'Exe' means in this context, maybe rename to something more explicit? (I believe this is returning the path to the parent of `p` ?

pkg/syncthing/process.go
```diff
@@ -16,22 +14,47 @@
...
+}
 
+func terminate(p *process.Process, wait bool) error {
+	if err := p.Terminate(); err != nil {
 		return err
 	}
 
 	if wait {
-		defer proc.Wait() // nolint: errcheck
+		isRunning, err := p.IsRunning()
```
rberrelleza, line 47: double check, but I believe p.IsRunning doesn't work as expected in windows if the process doesn't exist.

pkg/syncthing/syncthing.go
```diff
@@ -702,32 +668,36 @@ func (s *Syncthing) Restart(ctx context.Context) error {
...
+	if err != nil {
 		return err
 	}
-
-	if err := os.Remove(pidPath); err != nil {
-		if os.IsNotExist(err) {
-			return nil
+	for _, p := range pList {
+		exe, err := p.Exe()
```
rberrelleza, line 677:
```suggestion
		executablePath, err := p.Exe()
```

pkg/syncthing/syncthing.go
```diff
@@ -702,32 +668,36 @@ func (s *Syncthing) Restart(ctx context.Context) error {
...
+		if err != nil {
+			log.Infof("error getting procress exec: %s", err.Error())
```
rberrelleza, line 679: add the pid or process name so it helps troubleshoot?
```suggestion
			log.Infof("error getting exec path for process %s: %s", p.Pid, err.Error())
```

pkg/syncthing/syncthing.go
```diff
@@ -702,32 +668,36 @@ func (s *Syncthing) Restart(ctx context.Context) error {
...
+			continue
 		}
-
-		log.Infof("failed to delete pidfile %s: %s", pidPath, err)
+		if exe != getInstallPath() {
```
rberrelleza, line 682: can you add comments to the intention of the check? It's clear from just looking at the code (I'm guessing you're validating if this is syncthing by checking the path?)

pkg/syncthing/syncthing.go
rberrelleza: why are we skipping if it's the parent?

pkg/syncthing/syncthing.go
```diff
@@ -702,32 +668,36 @@ func (s *Syncthing) Restart(ctx context.Context) error {
...
+			continue
+		}
+		if getParentExe(p) == getInstallPath() {
+			continue
+		}
+		argLine, err := p.Cmdline()
+		if err != nil {
+			return err
+		}
+		if !strings.Contains(argLine, fmt.Sprintf("-home %s", s.Home)) {
```
rberrelleza, line 692: can you add a comment with the expected argLine?

pkg/syncthing/syncthing.go
```diff
@@ -702,32 +668,36 @@ func (s *Syncthing) Restart(ctx context.Context) error {
...
+			continue
+		}
+		log.Infof("terminating syncthing with wait %t", wait)
```
rberrelleza, line 695: can you add which synching process you're terminating?  pid, path, args? that can help us troubleshoot if this doesn't work as expected.

pkg/syncthing/syncthing.go
```diff
@@ -702,32 +668,36 @@ func (s *Syncthing) Restart(ctx context.Context) error {
...
-
-	if !force {
-		if pid != s.pid {
-			log.Infof("syncthing pid-%d wasn't created by this command, skipping", pid)
-			return nil
-		}
-	}
-
-	if err := s.cleanupDaemon(pid, force); err != nil {
+func (s *Syncthing) Stop(wait bool) error {
```
rberrelleza, line 671: we should add tests for this, especially to validate that we're not killing syncthing's owned by a different up, or other processes running in the machine.

rberrelleza (commented):
looks good, we should test in windows.

pkg/syncthing/syncthing.go
pchico83: It is slower. I think I am going to do it just at the beginning of the `up` sequence and in parallel with the rest of the `up` sequence.
For windows, yes, it also works for non-admin users

pkg/syncthing/syncthing.go
pchico83: This is to skip the child of the main syncthing process, which should be terminated by the parent

PR #1288, opened 2021-02-18
Remove syncthing process based on name

pchico83:
Using `p.Name` instead of `p.Exe` solves the cross-compilation issues we have seen before


pkg/syncthing/syncthing.go
```diff
@@ -701,33 +670,51 @@ func (s *Syncthing) Restart(ctx context.Context) error {
...
+			log.Infof("error terminating syncthing %d with wait: %s", p.Pid, err.Error())
 		}
+		log.Infof("terminated syncthing %d with wait: %s", p.Pid, s.Home)
 	}
+	return nil
+}
 
-	if err := s.cleanupDaemon(pid, force); err != nil {
-		return err
+// SoftTerminate best effor to halt the background process
```
rberrelleza, line 704: This doesn't explain what the function does. Try to rephrase it?

pkg/syncthing/syncthing.go
```diff
@@ -701,33 +670,51 @@ func (s *Syncthing) Restart(ctx context.Context) error {
...
+func (s *Syncthing) SoftTerminate() error {
```
rberrelleza, line 705: Both soft and hard Terminate kill the process using the terminate function. What makes one soft vs hard?

rberrelleza (commented):
Tests?

pkg/syncthing/process.go
```diff
@@ -16,22 +14,30 @@
...
+	if !wait {
+		return nil
+	}
+
+	isRunning, err := p.IsRunning()
+	if err != nil {
+		return err
+	}
+	for isRunning {
+		time.Sleep(10 * time.Millisecond)
```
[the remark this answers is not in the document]
pchico83: Oh, good catch. I will also kill it if it takes long

PR #1289, opened 2021-02-18
Add autocreate field to manifest

### LINKED ISSUE #1285: Force "okteto up" to only work on "swap" mode
I want to force `okteto up` to work in swap mode. If not, developers can accidentally create a deployment using `okteto up` and then, it is hard to update the deployment using other cli tools like helm or kubectl.

Have a field `autocreate`, by default to `false`, which enables `okteto up` to create a deployment if it doesn't exist.

**Additional context**
This is a breaking change. A clear message must be shown indicating to define `autocreate` if needed.

jLopezbarb:
Signed-off-by: jLopezbarb <javier@okteto.com>

Fixes #1285 

## Proposed changes
- Creates a autocrate field in manifest to deploy even when there is no application deployed on Okteto Cloud.
- Returns an error if someone tries to create a deployment without adding this field on manifest or adding the --deploy arg

### DIFF
--- a/cmd/up/up.go
+++ b/cmd/up/up.go
@@ -475,9 +475,9 @@ func (up *upContext) getCurrentDeployment(ctx context.Context, autoDeploy bool) 
 		err = errors.UserError{
 			E: fmt.Errorf("Deployment %s not found in namespace '%s'", up.Dev.Name, up.Dev.Namespace),
 			Hint: `Please:
-	- Make sure your application is deployed and ready.
-	- Make sure your context is pointing to the right namespace.
-	- Set the 'autocreate' property in your okteto.yml manifest if you want to automatically create a new development container.
+      - Make sure your application is deployed.
+      - Make sure your context is pointing to the right namespace.
+      - To automatically create a new deployment, set 'autocreate' in your okteto manifest.
     More information is available here: https://okteto.com/docs/reference/cli#up`,
 		}
 		return nil, false, err
--- a/cmd/up/up.go
+++ b/cmd/up/up.go
@@ -467,10 +467,11 @@ func (up *upContext) getCurrentDeployment(ctx context.Context, autoDeploy bool) 
 		return nil, false, err
 	}
 
-	if !autoDeploy {
-		if err := utils.AskIfDeploy(up.Dev.Name, up.Dev.Namespace); err != nil {
-			return nil, false, err
-		}
+	if !autoDeploy && !up.Dev.Autocreate {
+		err = errors.UserError{
+			E:    fmt.Errorf("Can't autocreate deployment"),
+			Hint: "Deploy your app first on okteto cloud or enable it using --deploy arg or setting autocreate variable to true in your manifest"}
+		return nil, false, err
 	}
 
 	return up.Dev.GevSandbox(), true, nil
--- a/cmd/up/up.go
+++ b/cmd/up/up.go
@@ -105,8 +105,8 @@ func Up() *cobra.Command {
 			checkLocalWatchesConfiguration()
 
 			if autoDeploy {
-				log.Yellow(`The use of 'deploy' flag will be deprecated in a future release. 
-Please set 'autocreate' in your okteto manifest to get the same behavior. 
+				log.Yellow(`The 'deploy' flag is deprecated and will be removed in a future release.
+Set the 'autocreate' field in your okteto manifest to get the same behavior.
 More information is available here: https://okteto.com/docs/reference/cli#up`)
 			}
 
@@ -480,11 +480,9 @@ func (up *upContext) getCurrentDeployment(ctx context.Context, autoDeploy bool) 
 
 	if !up.Dev.Autocreate {
 		err = errors.UserError{
-			E: fmt.Errorf("Deployment %s not found in namespace '%s'", up.Dev.Name, up.Dev.Namespace),
-			Hint: `Please:
-      - Make sure your application is deployed.
-      - Make sure your context is pointing to the right namespace.
-      - To automatically create a new deployment, set 'autocreate' in your okteto manifest.
+			E: fmt.Errorf("Deployment '%s' not found in namespace '%s'", up.Dev.Name, up.Dev.Namespace),
+			Hint: `Verify that your application has been deployed and your Kubernetes context is pointing to the right namespace
+    Or set the 'autocreate' field in your okteto manifest if you want to create a standalone development container
     More information is available here: https://okteto.com/docs/reference/cli#up`,
 		}
 		return nil, false, err
--- a/cmd/up/up.go
+++ b/cmd/up/up.go
@@ -105,9 +105,9 @@ func Up() *cobra.Command {
 			checkLocalWatchesConfiguration()
 
 			if autoDeploy {
-				log.Yellow(`Warning: The use of 'deploy' flag will be deprecated in a future release. 
-    	 Please set 'autocreate' in your okteto manifest to get the same behaviour. 
-    	 More information is available here: https://okteto.com/docs/reference/cli#up`)
+				log.Yellow(`The use of 'deploy' flag will be deprecated in a future release. 
+Please set 'autocreate' in your okteto manifest to get the same behavior. 
+More information is available here: https://okteto.com/docs/reference/cli#up`)
 			}
 
 			dev, err := loadDevOrInit(namespace, k8sContext, devPath)
--- a/cmd/up/up.go
+++ b/cmd/up/up.go
@@ -104,6 +104,12 @@ func Up() *cobra.Command {
 
 			checkLocalWatchesConfiguration()
 
+			if autoDeploy {
+				log.Yellow(`Warning: The use of 'deploy' flag will be deprecated in a future release. 
+    	 Please set 'autocreate' in your okteto manifest to get the same behaviour. 
+    	 More information is available here: https://okteto.com/docs/reference/cli#up`)
+			}
+
 			dev, err := loadDevOrInit(namespace, k8sContext, devPath)
 			if err != nil {
 				return err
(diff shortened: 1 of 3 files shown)


cmd/up/up.go
```diff
@@ -467,10 +467,11 @@ func (up *upContext) getCurrentDeployment(ctx context.Context, autoDeploy bool)
...
 	}
 
-	if !autoDeploy {
-		if err := utils.AskIfDeploy(up.Dev.Name, up.Dev.Namespace); err != nil {
-			return nil, false, err
-		}
+	if !autoDeploy && !up.Dev.Autocreate {
+		err = errors.UserError{
+			E:    fmt.Errorf("Can't autocreate deployment"),
+			Hint: "Deploy your app first on okteto cloud or enable it using --deploy arg or setting autocreate variable to true in your manifest"}
```
pchico83, line 473: @rlamana ^^ that's the error if running `okteto up` and the deployment doesn't exist

cmd/up/up.go
```diff
@@ -467,10 +467,11 @@ func (up *upContext) getCurrentDeployment(ctx context.Context, autoDeploy bool)
 		return nil, false, err
 	}
 
-	if !autoDeploy {
-		if err := utils.AskIfDeploy(up.Dev.Name, up.Dev.Namespace); err != nil {
-			return nil, false, err
-		}
+	if !autoDeploy && !up.Dev.Autocreate {
```
pchico83, line 470: Set `dev.Autocreate` to `true` here:
https://github.com/okteto/okteto/blob/master/cmd/up/up.go#L180
and from there, we only care about `dev.Autocreate`.

cmd/up/up.go
rlamana: Here a few suggestions on how we should present this output:

1. Can we be more concise with the error? I wouldn't based it on the "autocreate" or at least I would make it clear that the deployment wasn't found first, and then maybe mention the possibility to add the autocreate in the hint. Maybe something like: `Deployment ${name} not found in namespace '$namespace'`

2. Don't mention `Okteto Cloud` since this could be any cluster, right?

3. In the hint show a clear list of options:
``` 
   - Make sure your application is deployed and ready.
   - Make sure your context is pointing to the right namespace.
   - Set the `autocreate` property in your okteto.yml manifest if you want to automatically create a new development container.
 ```

4. Maybe we could show a link to the docs (are we doing this in other parts of the cli?).

wdyt?

File: pkg/model/dev.go [hunk 3]
```diff
@@ -144,6 +144,7 @@ type Dev struct {
 	Resources            ResourceRequirements  `json:"resources,omitempty" yaml:"resources,omitempty"`
 	Services             []*Dev                `json:"services,omitempty" yaml:"services,omitempty"`
 	PersistentVolumeInfo *PersistentVolumeInfo `json:"persistentVolume,omitempty" yaml:"persistentVolume,omitempty"`
+	Autocreate           bool                  `json:"autocreate,omitempty" yaml:"autocreate,omitempty"`
```
rberrelleza, line 147: Is this the same as if the user passes the `deploy` arg in the CLI? If so, can we use the same name in the manifest to keep it consistent?

cmd/up/up.go
rberrelleza: How will this interact with the CLI param? Will the CLI take precedence?

rberrelleza (commented):
See comments inline on the name of the key in the manifest.

pkg/model/dev.go
pchico83: `deploy` sounds confusing? what about `autodeploy`?
Also, we are keeping the `deploy` arg, but it should be deprecated

rberrelleza: I think calling the same thing two different ways is more confusing. People use the `--deploy` param today, it'll be confusing to them.

both are equally confusing to me TBH, the fact that we call it 'deploy'  or 'autocreate' is leaking that fact that we need a deployment before doing the `up` phase of the process. I'd rather stay with the current 'confusing' term rather than introduce yet another.

pchico83: @rberrelleza do you have numbers on how many are using `--deploy`? The idea is to deprecate the `deploy` arg

rberrelleza: not sure if we are collecting that information. 

maybe we can call it `mode: standalone`? (with the idea that then having `mode:swap` in the future so that users can be more explicit about how they want to use up?)

rberrelleza: if we are deprecating the deploy flag, maybe let's add a message to the help now?

cmd/up/up.go
```diff
@@ -109,7 +109,7 @@ func Up() *cobra.Command {
 				return err
 			}
 
-			if err := loadDevOverrides(dev, namespace, k8sContext, forcePull, remote); err != nil {
+			if err := loadDevOverrides(dev, namespace, k8sContext, forcePull, remote, autoDeploy); err != nil {
```
pchico83, line 112: There is an option (`MarkHidden`) in cobra to not show an arg in the help command. Let add the hidden option to the `autodeploy` flag and also remove `autodeploy` from the docs, but keeping the flag hidden for now

cmd/up/up.go
```diff
@@ -467,10 +471,16 @@ func (up *upContext) getCurrentDeployment(ctx context.Context, autoDeploy bool)
...
 	}
 
-	if !autoDeploy {
-		if err := utils.AskIfDeploy(up.Dev.Name, up.Dev.Namespace); err != nil {
-			return nil, false, err
+	if !up.Dev.Autocreate {
+		err = errors.UserError{
+			E: fmt.Errorf("Deployment %s not found in namespace '%s'", up.Dev.Name, up.Dev.Namespace),
+			Hint: `Please:
+	- Make sure your application is deployed and ready.
```
pchico83, line 478:
```suggestion
- Make sure your application is deployed.
```

cmd/up/up.go
```diff
@@ -467,10 +471,16 @@ func (up *upContext) getCurrentDeployment(ctx context.Context, autoDeploy bool)
...
+	- Make sure your context is pointing to the right namespace.
```
pchico83, line 479:
```suggestion
- Make sure your context is pointing to the right namespace.
```

cmd/up/up.go
```diff
@@ -467,10 +471,16 @@ func (up *upContext) getCurrentDeployment(ctx context.Context, autoDeploy bool)
...
+	- Set the 'autocreate' property in your okteto.yml manifest if you want to automatically create a new development container.
```
pchico83, line 480:
```suggestion
- Set the 'autocreate' field in your okteto manifest if you want to automatically create a new deployment in your current namespace.
```

cmd/up/up.go
```diff
@@ -467,10 +471,16 @@ func (up *upContext) getCurrentDeployment(ctx context.Context, autoDeploy bool)
...
+    More information is available here: https://okteto.com/docs/reference/cli#up`,
```
pchico83, line 481: I think this link should point to the `autocreate` field docs, and add it to the third bullet point.

rlamana: @pchico83 I talked to @jLopezbarb about this. I'd like to start adding links to the full documentation of the commands in case of error. I suggested this so users can read about the general workings of the UP command (including a link to the `autocreate` but from the docs themselves), instead of having many links on the same output.

We should then also add a paragraph explaining the 'autocreate' in general in the UP description and add a link to how to configure it in the manifest from there.

pkg/model/dev.go
rlamana: I thought that the idea of using a `mode` was interesting for a long time. We even discussed this several times in the past, but the truth that if we want to keep it simple that just adds a lot more complexity for a use case that is at its minimum:

According to the data: 

- 95.59% of swap use.
- 4.41% of the "standalone" use. Of which we already know many happened by mistake.

Having a simple flag called `autocreate: true` right now (or even `standalone: true`) allows us to keep this feature for that small percentage who use it and avoids the current mistake of creating an unwanted deployment for some. I don't think we should be blocked by the naming convention for this, as long as is not as confusing as `deploy` or `autodeploy`, which is a term we use for something completely different.

I vote for `autocreate: true` or `standalone: true`

rlamana:
@jLopezbarb can you share a screenshot of how the output looks in the terminal?

pkg/model/dev.go
pchico83: @jLopezbarb let's go with `autocreate` then

jLopezbarb:
This is how the output looks like with the url in the last bullet point. 

Captura de pantalla 2021-02-19 a las 11 03 11

I think something like adding another line to show the url could be more readable for the user.

Captura de pantalla 2021-02-19 a las 11 12 44

rlamana:
> @rlamana this is how the output looks like with the url in the last bullet point.

I thought it was going to look worse :-P We should definitely shorten the last bullet point and take the URL to the next line.

rlamana:
> > @rlamana this is how the output looks like with the url in the last bullet point.

We should also only point to the UP docs. It will make all that shorter too.

cmd/up/up.go
rlamana: What about a shorten version:
```suggestion
	- To automatically create a new deployment, set the 'autocreate' field in your okteto manifest.
```

jLopezbarb:
The warning message looks like this:
Captura de pantalla 2021-02-19 a las 12 51 53

pchico83:
> The warning message looks like this:
> Captura de pantalla 2021-02-19 a las 12 51 53

I would remove the "warning" word, it is already implied by the yellow color. @rlamana thoughts?

cmd/up/up.go
```diff
@@ -104,12 +104,18 @@ func Up() *cobra.Command {
 
 			checkLocalWatchesConfiguration()
 
+			if autoDeploy {
+				log.Yellow(`Warning: The use of 'deploy' flag will be deprecated in a future release. 
+    	 Please set 'autocreate' in your okteto manifest to get the same behaviour.
```
rberrelleza, line 109:
```suggestion
    	 Please set 'autocreate' in your okteto manifest to get the same behavior. 
```

rlamana:
> I would remove the "warning" word, it is already implied by the yellow color. @rlamana thoughts?
@pchico83 @jLopezbarb I would remove it too. But, don't we have a symbol for this type of message, just like we have the [x] for errors, or the [i] for info messages? I think we should use that format too (maybe we can have an [!] exclamation symbol.

pchico83:
> > I would remove the "warning" word, it is already implied by the yellow color. @rlamana thoughts?
> > @pchico83 @jLopezbarb I would remove it too. But, don't we have a symbol for this type of message, just like we have the [x] for errors, or the [i] for info messages? I think we should use that format too (maybe we can have an [!] exclamation symbol.

@rlamana we don't but makes sense. Let's create an issue for it to keep the scope of this issue unaffected.

PR #1291, opened 2021-02-23
Restore previous sleeping state on "okteto down"

pchico83:

## Proposed changes
- `okteto up` checks the replicas before a sleeping operation and annotatess that value to be used by `okteto down`

### DIFF
--- a/pkg/k8s/deployments/utils.go
+++ b/pkg/k8s/deployments/utils.go
@@ -71,7 +71,6 @@ func getTranslationFromAnnotation(annotations map[string]string) (model.Translat
 	return tr, nil
 }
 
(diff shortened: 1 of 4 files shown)


pkg/k8s/deployments/utils.go
```diff
@@ -64,3 +70,18 @@ func getTranslationFromAnnotation(annotations map[string]string) (model.Translat
 	}
 	return tr, nil
 }
+
+//GetTranslations fills all the deployments pointed by a development container
```
ifbyol, line 74: minor thing, comment doesn't seem right for this function

ifbyol (approved):

PR #1295, opened 2021-02-25
Jlopezbarb/logs just for up

### LINKED ISSUE #977: okteto fails to write to the log if more than one `okteto up` is active at the same time
**Describe the bug**
okteto fails to write to the log if more than one `okteto up` is active at the same time

**Error**
`Failed to write to log, can't rename log file: rename W:\.okteto\okteto.log W:\.okteto\okteto-2020-07-20T12-15-47.445.log: The process cannot access the file because it is being used by another process.`

 - OS: Windows

jLopezbarb:
Fixes #977 

## Proposed changes
- Get the logs just for the up command, creating a log file for every dev container 
- Creates a folder with the namespace and name of the dev container for storing the logs of that container.
- When doing Okteto doctor retrieves the logs of the Okteto manifest from the current working directory

### DIFF
--- a/cmd/up/up.go
+++ b/cmd/up/up.go
@@ -190,6 +190,10 @@ func loadDevOverrides(dev *model.Dev, namespace, k8sContext string, forcePull bo
 
 	dev.LoadContext(namespace, k8sContext)
 
+	if dev.Namespace == "" {
+		dev.Namespace, _ = k8Client.GetNamespace(k8sContext)
+	}
+
 	if remote > 0 {
 		dev.RemotePort = remote
 	}
@@ -208,10 +212,6 @@ func loadDevOverrides(dev *model.Dev, namespace, k8sContext string, forcePull bo
 
 	if forcePull {
 		dev.LoadForcePull()
-	}
-
-	if dev.Namespace == "" {
-		dev.Namespace, _ = k8Client.GetNamespace(k8sContext)
 	}
 
 	return nil
--- a/cmd/up/up.go
+++ b/cmd/up/up.go
@@ -119,7 +119,7 @@ More information is available here: https://okteto.com/docs/reference/cli#up`)
 				return err
 			}
 
-			log.CreateLogFile(config.GetDeploymentHome(dev.Namespace, dev.Name), config.VersionString)
+			log.ConfigureFileLogger(config.GetDeploymentHome(dev.Namespace, dev.Name), config.VersionString)
 
 			if err := checkStignoreConfiguration(dev); err != nil {
 				log.Infof("failed to check '.stignore' configuration: %s", err.Error())
--- a/cmd/up/up.go
+++ b/cmd/up/up.go
@@ -44,7 +44,6 @@ import (
 	"github.com/okteto/okteto/pkg/okteto"
 	"github.com/okteto/okteto/pkg/registry"
 	"github.com/okteto/okteto/pkg/ssh"
-	"github.com/sirupsen/logrus"
 
 	"github.com/okteto/okteto/pkg/k8s/forward"
 	"github.com/okteto/okteto/pkg/syncthing"
@@ -120,7 +119,7 @@ More information is available here: https://okteto.com/docs/reference/cli#up`)
 				return err
 			}
 
-			log.Init(logrus.WarnLevel, config.GetDeploymentHome(dev.Namespace, dev.Name), config.VersionString)
+			log.CreateLogFile(config.GetDeploymentHome(dev.Namespace, dev.Name), config.VersionString)
 
 			if err := checkStignoreConfiguration(dev); err != nil {
 				log.Infof("failed to check '.stignore' configuration: %s", err.Error())
--- a/cmd/up/up.go
+++ b/cmd/up/up.go
@@ -120,6 +120,8 @@ More information is available here: https://okteto.com/docs/reference/cli#up`)
 				return err
 			}
 
+			log.Init(logrus.WarnLevel, config.GetDeploymentHome(dev.Namespace, dev.Name), config.VersionString)
+
 			if err := checkStignoreConfiguration(dev); err != nil {
 				log.Infof("failed to check '.stignore' configuration: %s", err.Error())
 			}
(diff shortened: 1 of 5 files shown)


main.go
```diff
@@ -61,7 +60,6 @@ func init() {
 
 func main() {
 	ctx := context.Background()
-	log.Init(logrus.WarnLevel, config.GetOktetoHome(), config.VersionString)
```
pchico83, line 64: Keep this line but just initialize the standard output logger

cmd/up/up.go
```diff
@@ -119,6 +120,8 @@ More information is available here: https://okteto.com/docs/reference/cli#up`)
 				return err
 			}
 
+			log.Init(logrus.WarnLevel, config.GetDeploymentHome(dev.Namespace, dev.Name), config.VersionString)
```
pchico83, line 123: Here, we should initialize the file logger in debug level

pchico83: Double check `dev.Namespace` is always initialized at this point. You might need to run:
```
dev.Namespace, _ = k8Client.GetNamespace(k8sContext)
```

pkg/k8s/client/client.go
```diff
@@ -45,17 +46,8 @@ var (
...
-			clientcmd.NewDefaultClientConfigLoadingRules(),
-			&clientcmd.ConfigOverrides{
-				CurrentContext: k8sContext,
-				ClusterInfo:    clientcmdapi.Cluster{Server: ""},
-			},
-		)
-
-		namespace, _, err = clientConfig.Namespace()
+		clientConfig = GetClientConfig(k8sContext)
+		namespace, err := GetNamespace(k8sContext)
```
pchico83, line 50: Use `=` and you won't need two return statements

cmd/up/up.go
jLopezbarb: It's initialized on loadDevOverrides if its not setted in your Okteto manifest.

pchico83: I think if the user does not set the `namespace` flag and it is not defined in the okteto manifest, it is not initialized until here:
https://github.com/okteto/okteto/blob/master/cmd/up/up.go#L229
Maybe we can refactor that to always initialize it as part of `loadDevOverrides`

pkg/cmd/doctor/run.go
```diff
@@ -88,9 +88,9 @@ func Run(ctx context.Context, dev *model.Dev, devPath string, c *kubernetes.Clie
 	files := []string{summaryFilename}
 	files = append(files, stignoreFilenames...)
 
-	deploymentLogs := filepath.Join(config.GetDeploymentHome(dev.Namespace, dev.Name), "okteto.log")
-	if model.FileExists(deploymentLogs) {
-		files = append(files, deploymentLogs)
+	deploymentLogsPath := filepath.Join(config.GetDeploymentHome(dev.Namespace, dev.Name), "okteto.log")
```
rberrelleza, line 91: With this change, will `okteto doctor` also include the main `okteto.log` file?

pchico83: From https://github.com/natefinch/lumberjack:
```
Lumberjack assumes that only one process is writing to the output files. Using the same lumberjack configuration from multiple processes on the same machine will result in improper behavior.
```
The idea is to log just to the standard output for other commands, and just for `up`, store the logs in the okteto deployment folder to make it available for `okteto doctor`. I think it is a good compromise, thoughts?

rberrelleza: What do you mean by "the standard output of the other commands"?

pchico83: I couldn't find any logging library that supports several processes writing to the same log file, which is what we do right now if you execute `okteto` commands in parallel.
I have checked other CLI tools and they don't usually log to a file. We log things for other commands, but I haven't found that info useful more than for troubleshooting `okteto up`.

rberrelleza: yeah, `okteto up` is the more complex command we have by far.  cool, if we miss the logs we can always bring them back.

cmd/up/up.go
```diff
@@ -119,6 +119,8 @@ More information is available here: https://okteto.com/docs/reference/cli#up`)
 				return err
 			}
 
+			log.CreateLogFile(config.GetDeploymentHome(dev.Namespace, dev.Name), config.VersionString)
```
pchico83, line 122: Let's call it `ConfigureFileLogger`

pkg/k8s/client/client.go
```diff
@@ -46,16 +47,8 @@ var (
...
-			clientcmd.NewDefaultClientConfigLoadingRules(),
-			&clientcmd.ConfigOverrides{
-				CurrentContext: k8sContext,
-				ClusterInfo:    clientcmdapi.Cluster{Server: ""},
-			},
-		)
-
-		namespace, _, err = clientConfig.Namespace()
+		clientConfig = GetClientConfig(k8sContext)
+		namespace, err = GetNamespace(k8sContext)
```
pchico83, line 51: Since we have already initialized the namespace, we can probably remove `namespace` from this function

cmd/up/up.go
```diff
@@ -226,10 +231,6 @@ func (up *upContext) start(autoDeploy, build bool) error {
 		return fmt.Errorf("failed to load your local Kubeconfig: %q context not found in %q", up.Dev.Context, kubecfg)
 	}
 
-	if up.Dev.Namespace == "" {
```
pchico83, line 229: We can probably remove similar lines from other commands, for example here:
https://github.com/okteto/okteto/blob/master/cmd/down.go#L99
if we move the namespace initialization iiside `dev.LoadContext` instead of `loadDevOverrides`

pchico83 (approved):

pkg/k8s/client/client.go
jLopezbarb: I think this change may break other commands, `GetLocal` method is used by all the other commands, each command only needs some of the variables returned by `GetLocal`. It would be convenient to make a refactor of `GetLocal` method so that each command collects only the variables it requires and does not compute the others.

pchico83: Sounds good, we can do it in a different PR

jLopezbarb: I think this change may break other commands, the `GetLocal` method is used by all other commands, each command only needs some of the variables returned by `GetLocal`. It would be convenient to make a refactor of the GetLocal method so that each command collects only the variables it requires and does not compute the others.

cmd/up/up.go
jLopezbarb: The problem of moving the logic to that function is that it creates a cycle import because of client -> analytics -> config -> model -> client. I have to check how can this logic be implemented without changing the main structure of the project.

PR #1300, opened 2021-03-02
Add init container declaration

### LINKED ISSUE #1294: Modify the request/limit for all containers that are in the Sidecar
On the cluster that I am working, we have limitranges configured and the min request is 256Mi.

When I try to make the okteto up I receive and error, because I have the issue as I said with the limitranges.

We able to inject some configuration in order to increase/modify the request/limit.

jLopezbarb:
Signed-off-by: jLopezbarb <javier@okteto.com>

Fixes #1294 

## Proposed changes
- Add initContainer to the manifest declaration
-


pkg/model/dev.go
```diff
@@ -202,6 +203,12 @@ type PersistentVolumeInfo struct {
 	Size         string `json:"size,omitempty" yaml:"size,omitempty"`
 }
 
+// InitContainer represents the initial container
+type InitContainer struct {
```
rberrelleza, line 207: What about also specifying the user/group? Or will this inherit the one from the main config?

pkg/k8s/deployments/translate.go
```diff
@@ -219,6 +220,29 @@ func TranslateDevContainer(c *apiv1.Container, rule *model.TranslationRule) {
 	TranslateContainerSecurityContext(c, rule.SecurityContext)
 }
 
+func TranslateInitContainer(initContainer *model.InitContainer) {
+	if initContainer.Image == "" {
+		initContainer.Image = "okteto/bin:1.2.21"
```
rberrelleza, line 225: Can we put this in a const, so it's easier to find/change when we update the tag?

rberrelleza (commented):
Can we add tests for parsing the manifests, and to validate that `okteto init` doesn't create this section on the generated manifests?

pkg/k8s/deployments/translate.go
```diff
@@ -219,6 +220,29 @@ func TranslateDevContainer(c *apiv1.Container, rule *model.TranslationRule) {
 	TranslateContainerSecurityContext(c, rule.SecurityContext)
 }
 
+func TranslateInitContainer(initContainer *model.InitContainer) {
+	if initContainer.Image == "" {
```
rberrelleza, line 224: What's going to happen if the init container provided by the user doesn't have the binaries we need?  Can we check that we fail gracefully in that scenario?

pkg/model/dev.go
```diff
@@ -202,6 +203,12 @@ type PersistentVolumeInfo struct {
...
+	Image     string               `json:"image,omitempty" yaml:"image,omitempty"`
```
rberrelleza, line 208: For the usecase where the user is going to push our image into their private repository, how are we going to communicate to the user which bin image we're using?

pkg/k8s/deployments/translate.go
rberrelleza: Also, what happens if the user has an older bin image here? How will we troubleshoot that?

pkg/model/dev.go
rberrelleza, line 207: Would it possible to include this config in the `okteto doctor` file?

pkg/k8s/deployments/translate.go
rberrelleza: we might even want to do the same we do for the Version, and make it injectable via our CI/CD process.

pkg/k8s/deployments/translate.go
pchico83: we can add a binary to the `okteto/bin` image that returns the version of the bin image, and validate the version as part of the clean command:
https://github.com/okteto/okteto/blob/master/cmd/up/up.go#L899

pkg/model/dev.go
pchico83: No need to specify the user group

pchico83: It is already done, we include the okteto manifest

pkg/model/dev.go
pchico83: `okteto up` will fail with a clear message

pchico83 (commented):
@jLopezbarb let's add logic to validate that the user is using the right `okteto/bin` image

pkg/k8s/deployments/translate.go
pchico83: @jLopezbarb can we use `model. OktetoBinImageTag` here?

images/bin/Makefile
```diff
@@ -2,4 +2,4 @@ tag = 1.2.21
 
 .PHONY: push
 push:
-	okteto build -t okteto/bin:${tag} .
+	okteto build -t jlopezbarb/bin:${tag} --build-arg OKTETO_BIN_TAG=${tag} .
```
pchico83, line 5:
```suggestion
	okteto build -t okteto/bin:${tag} --build-arg OKTETO_BIN_TAG=okteto/bin:${tag} .
```

cmd/up/up.go
```diff
@@ -422,6 +425,11 @@ func (up *upContext) activate(autoDeploy, build bool) error {
 			}
 		}
 
+		if version != model.GetCurrentImageTagVersion() {
```
pchico83, line 428:
```suggestion
		if version != model.OktetoBinImage {
```

cmd/up/up.go
```diff
@@ -422,6 +425,11 @@ func (up *upContext) activate(autoDeploy, build bool) error {
...
+			log.Yellow("The version of the binary image provided does not match the one used in this version.")
```
pchico83, line 429: The Okteto CLI version {version} uses the init container image { model.OktetoBinImage }.

cmd/up/up.go
```diff
@@ -422,6 +425,11 @@ func (up *upContext) activate(autoDeploy, build bool) error {
...
+			log.Yellow("Please consider upgrading your binary.")
```
pchico83, line 430: Please consider upgrading your init container image { dev.InitContainer.Image} with the content of  { model.OktetoBinImage }

pchico83 (approved):
Minor comments, but good to go!

pchico83 (approved):

PR #1302, opened 2021-03-02
Refactor dev.Namespace and dev.Context initialization in a single place

pchico83:
## Proposed changes

- The logic to initialize `dev.Namespace` and `dev.Context`is different places. This PR unifies their initialization in a single place

### DIFF
--- a/cmd/pipeline/deploy.go
+++ b/cmd/pipeline/deploy.go
@@ -26,8 +26,6 @@ import (
 	"github.com/okteto/okteto/pkg/cmd/login"
 	"github.com/okteto/okteto/pkg/errors"
 	"github.com/okteto/okteto/pkg/k8s/client"
-	k8Client "github.com/okteto/okteto/pkg/k8s/client"
-	"github.com/okteto/okteto/pkg/k8s/namespaces"
 	"github.com/okteto/okteto/pkg/log"
 	"github.com/okteto/okteto/pkg/okteto"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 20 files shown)


cmd/pipeline/deploy.go
```diff
@@ -175,14 +176,15 @@ func waitUntilRunning(ctx context.Context, name, namespace string, timeout time.
 }
 
 func getCurrentNamespace(ctx context.Context) (string, error) {
```
rberrelleza, line 178: should this consider the namespace that's coming from the parameter?

cmd/pipeline/deploy.go
```diff
@@ -175,14 +176,15 @@ func waitUntilRunning(ctx context.Context, name, namespace string, timeout time.
...
-	c, _, namespace, err := k8Client.GetLocal("")
+	c, _, err := k8Client.GetLocal("")
```
rberrelleza, line 179: it seems like half the calls  pass this as an empty string. Could we instead have a `k8Client.GetLocal()` and `k8Client.GetLocalWithContext(context string)`  to make it easier to read?

pkg/k8s/client/client.go
```diff
@@ -35,69 +36,50 @@ const (
...
+	config           *rest.Config
+	clientConfig     clientcmd.ClientConfig
+	currentNamespace string
+	currentContext   string
+	localClusters    = []string{"127.", "172.", "192.", "169.", model.Localhost, "::1", "fe80::", "fc00::"}
 )
 
 //GetLocal returns a kubernetes client with the local configuration. It will detect if KUBECONFIG is defined.
-func GetLocal(k8sContext string) (*kubernetes.Clientset, *rest.Config, string, error) {
+func GetLocal(currentContext string) (*kubernetes.Clientset, *rest.Config, error) {
```
rberrelleza, line 48: I'd also change this to only return the k8s client, and have a separate function to get the REST config, since most of the calls only need the first one.

rberrelleza (approved):

cmd/pipeline/deploy.go
pchico83: yeap, this is not call if namespace is defined

PR #1307, opened 2021-03-03
Remove legacy user queries

pchico83:

### DIFF
--- a/pkg/okteto/auth.go
+++ b/pkg/okteto/auth.go
@@ -21,7 +21,6 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
-	"strings"
 
 	"github.com/machinebox/graphql"
 	"github.com/okteto/okteto/pkg/config"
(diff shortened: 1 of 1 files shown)


jLopezbarb (approved):

pkg/okteto/auth.go
```diff
@@ -136,34 +134,10 @@ func queryUser(ctx context.Context, client *graphql.Client, token string) (*q, e
...
-			}
-
-			return nil, err
-		}
-	}
-
-	return &user, nil
-}
-
-func queryUserLegacy(ctx context.Context, client *graphql.Client, token string) (*q, error) {
```
rberrelleza, line 153: can we catch the error when we hit this (in the odd case that someone is running an older server) and tell them to upgrade the CLI?

rberrelleza (approved):
This is a breaking change, maybe wait until the next major release at the end of march?

pchico83:
Next friday we need a major release

PR #1309, opened 2021-03-03
Retry registry/buildkit transient errors

pchico83:

### DIFF
--- a/pkg/cmd/build/build.go
+++ b/pkg/cmd/build/build.go
@@ -62,7 +62,7 @@ func Run(ctx context.Context, namespace, buildKitHost string, isOktetoCluster bo
 
 	err = solveBuild(ctx, buildkitClient, opt, progress)
 	if oktetoErrors.IsRegistryTransient(err) {
-		log.Yellow("Retrying registry transient error...")
+		log.Yellow("Failed to push  %s to the registry,  retrying ...",  image)
 		return solveBuild(ctx, buildkitClient, opt, progress)
 	}
 	return err
--- a/pkg/cmd/build/build.go
+++ b/pkg/cmd/build/build.go
@@ -18,7 +18,7 @@ import (
 	"os"
 	"path/filepath"
 
-	oktetoErrors "github.com/okteto/okteto/pkg/errors"
+	"github.com/okteto/okteto/pkg/analytics"
 	"github.com/okteto/okteto/pkg/log"
 	"github.com/okteto/okteto/pkg/okteto"
 	"github.com/okteto/okteto/pkg/registry"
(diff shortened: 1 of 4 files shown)


jLopezbarb (approved):

pkg/cmd/build/build.go
```diff
@@ -18,6 +18,7 @@ import (
 	"os"
 	"path/filepath"
 
+	oktetoErrors "github.com/okteto/okteto/pkg/errors"
```
rberrelleza, line 21: we use `okErrors` in other parts of the repo. 

```suggestion
	okErrors "github.com/okteto/okteto/pkg/errors"
```

pkg/errors/errors.go
```diff
@@ -131,6 +131,21 @@ func IsTransient(err error) bool {
 	}
 }
 
+// IsRegistryTransient returns true if err represents a transient registry error
+func IsRegistryTransient(err error) bool {
```
rberrelleza, line 135: can we instead put this in the `registry`  package, so the call is something like `registry.IsTransientError(err error)` or something like that? that way we keep the registry-related business logic in a single package.

rberrelleza (approved):

pkg/cmd/build/build.go
```diff
@@ -59,5 +60,10 @@ func Run(ctx context.Context, namespace, buildKitHost string, isOktetoCluster bo
 		return errors.Wrap(err, "failed to create build solver")
 	}
 
-	return solveBuild(ctx, buildkitClient, opt, progress)
+	err = solveBuild(ctx, buildkitClient, opt, progress)
+	if oktetoErrors.IsRegistryTransient(err) {
+		log.Yellow("Retrying registry transient error...")
```
rberrelleza, line 65:
```suggestion
		log.Yellow("Failed to push  %s to the registry,  retrying ...",  image)
```

pkg/errors/errors.go
```diff
@@ -131,6 +131,21 @@ func IsTransient(err error) bool {
...
+	if err == nil {
+		return false
+	}
+
+	switch {
+	case strings.Contains(err.Error(), "failed commit on ref") && strings.Contains(err.Error(), "500 Internal Server Error"),
```
rberrelleza, line 141: do you need all three messages to be in the error, or are we handling three different errors here?

pkg/cmd/build/build.go
rberrelleza: should we track this event?

rberrelleza (approved):

pkg/errors/errors.go
pchico83: they are two different errors, one of them with two conditions

pkg/registry/registry.go
```diff
@@ -110,3 +110,18 @@ func ExpandOktetoDevRegistry(ctx context.Context, namespace, tag string) (string
...
 }
+
+// IsTransientError returns true if err represents a transient registry error
+func IsTransientError(err error) bool {
+	if err == nil {
+		return false
+	}
+
+	switch {
+	case strings.Contains(err.Error(), "failed commit on ref") && strings.Contains(err.Error(), "500 Internal Server Error"),
```
rberrelleza, line 121: This is not very easy to grasp, can you split into two different `case` statements so it's easier to see the conditions?

PR #1314, opened 2021-03-05
Reconnection logic for exec

### LINKED ISSUE #1313: Implement reconnection logic for "okteto exec"
I usually work with two terminals. One is running `okteto up` and another one is running `okteto exec`.
When there is a connection issue or after resuming my laptop from sleep, the `okteto up` session reconnects, but the `okteto exec` session does not. I would want the `okteto exec` session to reconnect too.

pchico83:

Fixes #1313 

## Proposed changes
- I am finally regenerating the k8s client on each call to `client.GetLocal`, otherwise the reconnection logic is not stable and can generate infinite `write: can't assign requested address` errors.


cmd/up/up.go
```diff
@@ -1034,9 +1038,6 @@ func (up *upContext) shutdown() {
 	if up.Forwarder != nil {
 		up.Forwarder.Stop()
 	}
-	if err := config.DeleteStateFile(up.Dev); err != nil {
```
pchico83, line 1037: Deleting the state file here can generate rc with parallel `exec` commands

rberrelleza: the state file was added so that the vscode plugin can give the users updates as the state of it changes. Not sure about the impact of this change, but keep that in comparison. It's better if the statefile is only modified during the lifecycle of "okteto up"/"okteto down".

PR #1322, opened 2021-03-09
Jlopezbarb/split config healthchecks

### LINKED ISSUE #1296: split configuration of healthchecks into separate fields for liveness and readiness
I would like to run with the readiness probes that my deployment defines.  

This way when I am not running/debugging the app in my container, the deployment pod's endpoints wont be exposed via services.

However I wish to disable the liveness probe, otherwise my container will get restarted over and over again.

Suggest configuring using a struct with fields

```
healthchecks:
  liveness: boolean
  readiness: boolean
```

jLopezbarb:
Fixes #1296

## Proposed changes
- Added a extended notation to configure which healthcheck probes to use
- Still works if a boolean is introduced, setting all the healtcheck probes to true

### DIFF
--- a/pkg/model/serializer.go
+++ b/pkg/model/serializer.go
@@ -47,6 +47,13 @@ type storageResourceRaw struct {
 	Class string   `json:"class,omitempty" yaml:"class,omitempty"`
 }
 
+// healthCheckProbesRaw represents the healthchecks info for serialization
+type healthCheckProbesRaw struct {
+	Liveness  bool `json:"liveness,omitempty" yaml:"liveness,omitempty"`
+	Readiness bool `json:"readiness,omitempty" yaml:"readiness,omitempty"`
+	Startup   bool `json:"startup,omitempty" yaml:"startup,omitempty"`
+}
+
 // UnmarshalYAML Implements the Unmarshaler interface of the yaml pkg.
 func (e *EnvVar) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	var raw string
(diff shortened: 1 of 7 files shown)


pkg/model/serializer.go
```diff
@@ -437,6 +444,34 @@ func (v ExternalVolume) MarshalYAML() (interface{}, error) {
 	return v.Name + ":" + v.SubPath + ":" + v.MountPath, nil
 }
 
+// UnmarshalYAML Implements the Unmarshaler interface of the yaml pkg.
+func (healthcheckProbes *HealthchecksProbes) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	var rawBool bool
+	err := unmarshal(&rawBool)
+	if err == nil {
+		healthcheckProbes.Liveness = true
```
pchico83, line 452: Should it be `rawBool` instead of `true`?

pkg/model/serializer.go
```diff
@@ -437,6 +444,34 @@ func (v ExternalVolume) MarshalYAML() (interface{}, error) {
...
+
+	healthcheckProbes.Liveness = healthCheckProbesRaw.Liveness
+	healthcheckProbes.Startup = healthCheckProbesRaw.Startup
+	healthcheckProbes.Readiness = healthCheckProbesRaw.Readiness
+	return nil
+}
+
+// MarshalYAML Implements the marshaler interface of the yaml pkg.
+func (healthcheckProbes HealthchecksProbes) MarshalYAML() (interface{}, error) {
+	return healthCheckProbesRaw(healthcheckProbes), nil
```
pchico83, line 472: if all are false, we shouldn't print anything
if all are true, a single `true` value should be returned
Also, can we have tests for these cases?

pchico83 (approved):

PR #1323, opened 2021-03-10
Jlopezbarb/add extended forward notation

### LINKED ISSUE #1274: Support label in port forwarding rules that include a service name
okteto supports port forwarding like this -
forward:
- 5012:5012
- 3306:some-service:3306

Here some-service is the name of a service in the same namespace, okteto should support usage of labels also instead of just service names

jLopezbarb:
Fixes #1274

## Proposed changes
- Adds a new notation for forward that supports searching a service by its labels

### DIFF
--- a/pkg/k8s/forward/manager.go
+++ b/pkg/k8s/forward/manager.go
@@ -45,6 +45,7 @@ type PortForwardManager struct {
 	ctx            context.Context
 	restConfig     *rest.Config
 	client         kubernetes.Interface
+	namespace      string
 }
 
 type active struct {
@@ -77,7 +78,7 @@ func (a *active) error() error {
 }
 
 // NewPortForwardManager initializes a new instance
-func NewPortForwardManager(ctx context.Context, iface string, restConfig *rest.Config, c kubernetes.Interface) *PortForwardManager {
+func NewPortForwardManager(ctx context.Context, iface string, restConfig *rest.Config, c kubernetes.Interface, namespace string) *PortForwardManager {
 	return &PortForwardManager{
 		ctx:        ctx,
 		iface:      iface,
@@ -85,6 +86,7 @@ func NewPortForwardManager(ctx context.Context, iface string, restConfig *rest.C
 		services:   make(map[string]struct{}),
 		restConfig: restConfig,
 		client:     c,
+		namespace:  namespace,
 	}
 }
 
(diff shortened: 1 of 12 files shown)


pkg/model/forward_test.go
```diff
@@ -190,3 +190,37 @@ func TestForward_less(t *testing.T) {
...
+		data     Forward
+	}{
+		{
+			name:     "service-name",
+			expected: "8080:svc:9090",
+			data:     Forward{Local: 8080, Remote: 9090, Service: true, ServiceName: "svc"},
+		},
+		{
+			name:     "service-name-and-labels",
+			expected: "8080:svc:5214",
```
rberrelleza, line 207: shouldn't the expected value be different since now there are labels involved now?

jLopezbarb: This will happen only when labels are passed as values and no service name is given (if a user adds labels and a service name it will return an error). The forward manager will look for the service name that has those labels and set the service name to the correct name. Finally you want to show the user which service are you exposing not where it comes from.

pkg/k8s/forward/manager.go
```diff
@@ -282,3 +294,20 @@ func (p *PortForwardManager) forwardService(ctx context.Context, namespace, serv
...
+func (p *PortForwardManager) GetServiceNameByLabel(namespace string, labels map[string]string) (string, error) {
+	labelsString := TransformLabelsToString(labels)
+	serviceName, err := services.GetServiceNameByLabel(p.ctx, namespace, p.client, labelsString)
+	if err != nil {
+		return "", err
+	}
+	return serviceName, nil
+}
+
+func TransformLabelsToString(labels map[string]string) string {
```
pchico83, line 307: `TransformLabelstoSelector`

pkg/k8s/services/crud.go
```diff
@@ -100,3 +100,29 @@ func GetPortsByPod(ctx context.Context, p *apiv1.Pod, c *kubernetes.Clientset) (
 	}
 	return result, nil
 }
+
+//GetServiceNameByLabel returns the name of the service with certain labels
+func GetServiceNameByLabel(ctx context.Context, namespace string, c kubernetes.Interface, labels string) (string, error) {
```
pchico83, line 105: Let's add tests for this with the kubernetes fake client

pchico83 (approved):

PR #1325, opened 2021-03-10
Support for "okteto config view"

pchico83:

## Proposed changes
This is useful for automation scripts that need to create new namespaces based on the githubid of the developer.

### DIFF
--- a/cmd/username.go
+++ b/cmd/username.go
@@ -1,0 +1,56 @@
+// Copyright 2020 The Okteto Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cmd
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/okteto/okteto/pkg/cmd/login"
+	"github.com/okteto/okteto/pkg/errors"
+	"github.com/okteto/okteto/pkg/log"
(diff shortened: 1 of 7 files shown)


cmd/username.go
```diff
@@ -0,0 +1,56 @@
...
+		Use:   "username",
+		Short: "Returns the username of the authenticated user",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			ctx := context.Background()
+			username := okteto.GetUsername()
+			if username != "" {
+				fmt.Println(username)
+				return nil
+			}
+			t, err := okteto.GetToken()
```
rberrelleza, line 39: why is it refreshing the token? if the user is not logged in I think it's better to ask them to log in again.

cmd/username.go
```diff
@@ -0,0 +1,56 @@
...
+				log.Infof("error getting okteto token: %s", err.Error())
+				return errors.ErrNotLogged
+			}
+			log.Info("refreshing okteto token...")
+			u, err := login.WithToken(ctx, t.URL, t.Token)
+			if err != nil {
+				log.Infof("error refreshing okteto token: %s", err.Error())
+				return errors.ErrNotLogged
+			}
+			fmt.Println(u.ExternalID)
```
rberrelleza, line 50: okteto.Username() returns the value of t.Username,  maybe it's better to call that function here?

rberrelleza:
maybe we can extend this to a `okteto config view` , to map what kubectl does? I can see how getting things like the token, okteto URL, etc.. could be useful.

pchico83:
> maybe we can extend this to an `okteto config view` , to map what kubectl does? I can see how getting things like the token, okteto URL, etc.. could be useful.
> 
> it might also be worth it so we can start standardizing our commands on (okteto "subjective" "verb"), instead of the mix of things we have today.

what about?
```
okteto config get external-id
okteto config get url
...
```

cmd/username.go
pchico83: it is because we were not saving the external-id. if the user is logged, but the external-id is not saved locally, we automatically do login to get the external-id

jLopezbarb (approved):

PR #1327, opened 2021-03-11
Jlopezbarb/split config healthchecks

### LINKED ISSUE #1296: split configuration of healthchecks into separate fields for liveness and readiness

jLopezbarb:
Fixes #1296 

## Proposed changes
- Adds new healthchecksprobes field for mutation named probes
-


pkg/model/dev.go
```diff
@@ -128,7 +128,7 @@ type Dev struct {
 	Environment          []EnvVar              `json:"environment,omitempty" yaml:"environment,omitempty"`
 	Secrets              []Secret              `json:"secrets,omitempty" yaml:"secrets,omitempty"`
 	Command              Command               `json:"command,omitempty" yaml:"command,omitempty"`
-	Healthchecks         bool                  `json:"healthchecks,omitempty" yaml:"healthchecks,omitempty"`
```
pchico83, line 131: Keep `Healthchecks` for backward compatibility, and define a new field `Probes`

pkg/model/dev.go
```diff
@@ -248,6 +248,13 @@ type ResourceRequirements struct {
 	Requests ResourceList `json:"requests,omitempty" yaml:"requests,omitempty"`
 }
 
+// HealthchecksProbes defines probes for containers
+type HealthchecksProbes struct {
```
pchico83, line 252: `HealthchecksProbes` -> `Probes`

pkg/model/dev.go
pchico83: If `healthchecks` is true, then set all probes to true.
Also, if all probes are `true`, set `healthchecks` to true.

pkg/model/translation.go
```diff
@@ -32,23 +32,24 @@ type Translation struct {
...
+	Command            []string             `json:"command,omitempty"`
+	Args               []string             `json:"args,omitempty"`
+	WorkDir            string               `json:"workdir"`
+	Healthchecks       bool                 `json:"healthchecks" yaml:"healthchecks"`
+	PersistentVolume   bool                 `json:"persistentVolume" yaml:"persistentVolume"`
+	Volumes            []VolumeMount        `json:"volumes,omitempty"`
+	SecurityContext    *SecurityContext     `json:"securityContext,omitempty"`
+	Resources          ResourceRequirements `json:"resources,omitempty"`
+	InitContainer      InitContainer        `json:"initContainers,omitempty"`
+	HealthchecksProbes *HealthchecksProbes  `json:"probes" yaml:"probes"`
```
pchico83, line 52: `HealthchecksProbes`-> `Probes`

pkg/model/serializer.go
```diff
@@ -437,6 +444,37 @@ func (v ExternalVolume) MarshalYAML() (interface{}, error) {
...
+	}
+
+	healthcheckProbes.Liveness = healthCheckProbesRaw.Liveness
+	healthcheckProbes.Startup = healthCheckProbesRaw.Startup
+	healthcheckProbes.Readiness = healthCheckProbesRaw.Readiness
+	return nil
+}
+
+// MarshalYAML Implements the marshaler interface of the yaml pkg.
+func (healthcheckProbes HealthchecksProbes) MarshalYAML() (interface{}, error) {
```
pchico83, line 471: We should print `probes` if any of them is true. false values won't be printed

pkg/model/dev.go
jLopezbarb: there is a healthcheck field that if you can set a boolean value that sets all the probes to true (backwards compatibility), you can also specify in that healthchecks field the probes that want to be enabled by specifying which one should be enabled.

pkg/model/serializer.go
jLopezbarb: If any of the probes is true it will be printed in the health checks field, if all the probes are enabled the healthchecks field will be set to true. If none of the probes are true then it won't print anything

pchico83 (approved):

Issue #1268, opened 2021-02-11
Okteto Stack hangs when command fails to start


rberrelleza:
**Describe the bug**
I ran an okteto stack, where the container failed with exit code 127 (due to me using an image without `sh` on it).

The cli waited forever, even if the kubernetes resource was already marked in CrashLoopBackoff for over 6 minutes.

```
okteto stack deploy --wait                        
⠇ StatefulSet is not ready: rberrelleza/spago. 0 out of 1 expected pods are ready... 
```

```
kubectl get pod
NAME      READY   STATUS             RESTARTS   AGE
spago-0   0/1     CrashLoopBackOff   6          6m20s
```

pchico83:
Fixed by #1333

Issue #1267, opened 2021-02-11
Can't deploy stack after deleting it


rberrelleza:
**Describe the bug**
`okteto stack deploy --wait` failed to launch my stack after destroying it.

**To Reproduce**
- Deploy stack succesfully.
- Deploy same stack with different command, have it fail due to resources. 
- `okteto stack destroy`
- Try to deploy stack again

Stack failed to be deployed with the error below:
       x  error installing stack: cannot re-use a name that is still in use

I was blocked until I manually deleted the following secrets:

       kubectl delete secret sh.helm.release.v1.spago.v3 
       secret "sh.helm.release.v1.spago.v3" deleted

       kubectl delete secret sh.helm.release.v1.spago.v4 
       secret "sh.helm.release.v1.spago.v4" deleted

I didn't have any pending helm releases
       helm ls
      NAME    NAMESPACE       REVISION        UPDATED STATUS  CHART   APP VERSION

**Desktop (please complete the following information):**
 - OS: macOS
 - Version 1.10.6

rberrelleza:
This might be the same root cause than #1269

pchico83:
Fixed by #1333

PR #1333, opened 2021-03-17
Refactor stacks to remove Helm dependency

### LINKED ISSUE #1269: okteto stack destroy fails to destroy stack if it's in crashloopbackoff
**Describe the bug**

```
okteto stack destroy      
 x  stack spago does not exist
```

```
NAME          READY   STATUS              RESTARTS   AGE
pod/spago-0   0/1     RunContainerError   7          11m

NAME            TYPE        CLUSTER-IP     EXTERNAL-IP   PORT(S)    AGE
service/spago   ClusterIP   10.132.243.1   <none>        1987/TCP   11m

NAME                     READY   AGE
statefulset.apps/spago   0/1     11m
```

```
kubectl get secret
NAME                          TYPE                                  DATA   AGE
basic-auth                    Opaque                                2      12d
default-token-lwj4s           kubernetes.io/service-account-token   3      315d
okteto-regcred                kubernetes.io/dockerconfigjson        2      315d
sh.helm.release.v1.spago.v1   helm.sh/release.v1                    1      12m
```

```
okteto stack deploy       
 x  error installing stack: cannot re-use a name that is still in use
```

pchico83:

Fixes #1269 #1268 #1267

## Proposed changes
- `okteto stack deploy` now creates deployments, statefulsets and services directly talking to the Kubernetes API, instead of using a dynamic Helm chart. This gives us more control on what's going on, improve error messages, simplify automatic testing...


jLopezbarb (approved):

pkg/cmd/stack/deploy.go
```diff
@@ -15,115 +15,170 @@ package stack
...
 		return err
 	}
 
-	dynamicStackFilename, err := saveStackFile(s)
-	if err != nil {
+	cfg := translateConfigMap(s)
+	output := fmt.Sprintf("Deploying stack '%s'...", s.Name)
+	cfg.Data[statusField] = progressingStatus
+	cfg.Data[outputField] = base64.StdEncoding.EncodeToString([]byte(output))
+	if err := configmaps.Deploy(ctx, cfg, s.Namespace, c); err != nil {
```
rberrelleza, line 52: should we put a spinner before this part?

pkg/cmd/stack/deploy.go
```diff
@@ -15,115 +15,170 @@ package stack
...
-	exists, err := helm.ReleaseExist(action.NewList(actionConfig), s.Name)
-	if err != nil {
-		return fmt.Errorf("error listing stacks: %s", err)
+	for name := range s.Services {
+		if len(s.Services[name].Volumes) == 0 {
+			if err := deployDeployment(ctx, name, s, c); err != nil {
+				return err
+			}
+		} else {
+			if err := deployStatefulSet(ctx, name, s, c); err != nil {
```
rberrelleza, line 89: I don't see value in using stateful sets, and I think they limit how we can do updates, as well as our usage of volumes. what if we just support deployments, and we create the PVC?

pkg/cmd/stack/deploy.go
```diff
@@ -15,115 +15,170 @@ package stack
...
+				return err
+			}
+		}
+		if len(s.Services[name].Ports) > 0 {
+			svcK8s := translateService(name, s)
+			if err := services.Create(ctx, svcK8s, c); err != nil {
+				return err
```
rberrelleza, line 96: an important thing of the stack experience is to show clear, non-kubernetes messages for the user when resource creation fails. would it make sense to invest in parsing this type of error and generating more specific messages?

pkg/cmd/stack/deploy.go
```diff
@@ -15,115 +15,170 @@ package stack
...
+			}
+		}
+	}
+
+	if err := destroyServicesNotInStack(ctx, s, c); err != nil {
```
rberrelleza, line 101: what is this for?

pkg/cmd/stack/deploy.go
```diff
@@ -15,115 +15,170 @@ package stack
...
 
-func saveStackFile(s *model.Stack) (string, error) {
-	tmpFile, err := ioutil.TempFile("", "okteto-stack")
-	if err != nil {
-		return "", fmt.Errorf("failed to create dynamic stack manifest file: %s", err)
+func deployDeployment(ctx context.Context, svcName string, s *model.Stack, c *kubernetes.Clientset) error {
+	d := translateDeployment(svcName, s)
+	old, err := c.AppsV1().Deployments(s.Namespace).Get(ctx, svcName, metav1.GetOptions{})
+	if err != nil && !errors.IsNotFound(err) {
+		return fmt.Errorf("error getting kubernetes deployment: %s", err)
```
rberrelleza, line 118: I think this is the kind of error we shouldn't get back to the user, since it breaks the abstraction.  could we change it to be more specific to the context/terminology we use in stack?

pkg/cmd/stack/destroy.go
```diff
@@ -15,27 +15,110 @@ package stack
...
+	}
+
+	if removeVolumes {
+		spinner.Update("Destroying volumes...")
+		if err := destroyStackVolumes(ctx, s, c); err != nil {
+			return err
+		}
+	}
+
+	if err := configmaps.Destroy(ctx, s.GetConfigMapName(), s.Namespace, c); err != nil {
```
rberrelleza, line 96: can we do the volumes at the end instead of the configmap?

pkg/cmd/stack/destroy.go
```diff
@@ -44,17 +127,88 @@ func Destroy(ctx context.Context, s *model.Stack) error {
...
+			}
+		}
+	}
+
+	return nil
+}
+
+func waitForPodsToBeDestroyed(ctx context.Context, s *model.Stack, c *kubernetes.Clientset) error {
+	ticker := time.NewTicker(100 * time.Millisecond)
+	timeout := time.Now().Add(300 * time.Second)
```
rberrelleza, line 185: can we have the timeouts on a single location? (I'm sure at some point we'll want to make them parametric.

pkg/cmd/stack/destroy.go
```diff
@@ -44,17 +127,88 @@ func Destroy(ctx context.Context, s *model.Stack) error {
...
+		<-ticker.C
+		podList, err := pods.ListBySelector(ctx, s.Namespace, selector, c)
+		if err != nil {
+			return err
+		}
+		if len(podList) == 0 {
+			return nil
+		}
+	}
+	return fmt.Errorf("kubernetes is taking too long to destroy your stack. Please check for errors and try again")
```
rberrelleza, line 198: can we give them more information on where to check for errors?

pkg/cmd/stack/deploy.go
pchico83: Yes, but it is the recommended way to use pods with volumes. They also make it simple to handle volumes in multi-zone clusters. We can add logic to regenerate the statefulset when needed

pkg/cmd/stack/deploy.go
pchico83: Yes, agree. In this version I am focusing on feature parity with the previous implementation, but one of the advantages of the new version is to have more control on errors

pkg/cmd/stack/deploy.go
pchico83: it you deploy a stack with 5 services, and then you redeploy with just 3, this code removes the extra 2 services

pkg/cmd/stack/destroy.go
pchico83: the configmap must be the last one, it is the stack metadata

PR #1344, opened 2021-03-23
Allow users to override serviceAccountName

### LINKED ISSUE #867: Feature Request: Allow users to override serviceAccountName on pod spec
I frequently use okteto in an EKS cluster on AWS and would like to make use of IAM Roles for Serivce Accounts on my okteto deployment/pod to authenticate to AWS. Currently, I am unable to change the service account on the pod from the default service account in the namespace okteto is deployed. If I have another service account with the iam role arn annotation, I cannot use it on the okteto deployment.

Allow the user to specify a serviceAccountName in the manifest file to override the default service account on the pod.

One workaround is to manually edit the default service account and add the role arn annotation, however the AWS IAM role has to trust this service account and this means that any pod in my namespace is now able to connect to aws with this role using the default service account. This has been ok for one-off namespaces with no other resources in them but it is still a slight security concern and harder to maintain.

**Additional context**
AWS Doc on this feature of EKS: https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html

While this feature request appears very EKS-specific, i'm sure there are other use cases for changing the service account name, such as setting up custom kubernetes rbac for your okteto pod and granting permissions via in-cluster authentication. This can be used while developing apps that interact with the kubernetes api in-cluster.

marov:
Signed-off-by: marov <mike.arov@gmail.com>

Fixes #867 

## Proposed changes
- Adds a new `serviceAccount` field, that s translated to PodSpec `ServiceAccountName`

### DIFF
--- a/pkg/k8s/deployments/translate.go
+++ b/pkg/k8s/deployments/translate.go
@@ -116,6 +116,7 @@ func translate(t *model.Translation, c *kubernetes.Clientset, isOktetoNamespace 
 		TranslateInitContainer(&rule.InitContainer)
 		TranslateOktetoVolumes(&t.Deployment.Spec.Template.Spec, rule)
 		TranslatePodSecurityContext(&t.Deployment.Spec.Template.Spec, rule.SecurityContext)
+		TranslatePodServiceAccount(&t.Deployment.Spec.Template.Spec, rule.ServiceAccount)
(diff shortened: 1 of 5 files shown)


pchico83 (approved):
Awesome @marov
Thanks a lot for your contribution!

PR #1351, opened 2021-03-26
generate ed25519 keys for SSH access

rberrelleza:
okteto up fails in modern fedora version, because RSA is not enabled by default anymore (see https://github.com/okteto/remote-kubernetes/issues/135 for reference). Changing the algorithm to ed25519. (this is what github also recommends these days)

## Proposed changes
- Generate SSH keys with ed25519

To keep the code simpler, I decided to change the name of the keys so we regenerate them automatically, rather than dealing with legacy and modern keys. What do you think about this? A side effect of this is that you will have to run "okteto down" and "okteto up" on existing dev envs to get the new public key uploaded.


pchico83 (approved):

rberrelleza:
> okteto up fails in modern fedora version, because RSA is not enabled by default anymore (see [okteto/remote-kubernetes#135](https://github.com/okteto/remote-kubernetes/issues/135) for reference). Changing the algorithm to ed25519. (this is what github also recommends these days)
> 
> ## Proposed changes
> * Generate SSH keys with ed25519
> 
> To keep the code simpler, I decided to change the name of the keys so we regenerate them automatically, rather than dealing with legacy and modern keys. What do you think about this? A side effect of this is that you will have to run "okteto down" and "okteto up" on existing dev envs to get the new public key uploaded.

we added a "schema_version" to force redeploy if you run up and you were in an older version.

PR #1352, opened 2021-03-26
add meaningful errors and remove super verbose log

rberrelleza:
## Proposed changes
- Add a message to some errors, rather than just the empty err 
- Remove super verbose event log

### DIFF
--- a/cmd/up/activate.go
+++ b/cmd/up/activate.go
@@ -293,7 +293,7 @@ func (up *upContext) waitUntilDevelopmentContainerIsRunning(ctx context.Context)
 				}
 				continue
 			}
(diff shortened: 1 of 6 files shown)


cmd/up/activate.go
```diff
@@ -293,7 +293,7 @@ func (up *upContext) waitUntilDevelopmentContainerIsRunning(ctx context.Context)
 				}
 				continue
 			}
-			log.Infof("pod %s event: %s", up.Pod, e.Message)
```
rberrelleza, line 296: @jLopezbarb @pchico83 I'm removing this log because it logs the entire pod object a few times. If you feel is important let me know, and maybe we can change it to just the name of the pod, or some key values?

pchico83 (approved):

PR #1360, opened 2021-03-30
support for .env files in okteto up

rberrelleza:
## Proposed changes
- Read an available `.env` file on start, and inject the values to the environment for the rest of the commands

### DIFF
--- a/cmd/push.go
+++ b/cmd/push.go
@@ -51,6 +51,9 @@ func Push(ctx context.Context) *cobra.Command {
 		Short: "Builds, pushes and redeploys source code to the target deployment",
 		RunE: func(cmd *cobra.Command, args []string) error {
 			ctx := context.Background()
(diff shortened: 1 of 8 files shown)


main.go
```diff
@@ -44,6 +46,11 @@ import (
 func init() {
 	log.SetLevel("warn")
 
+	err := godotenv.Load()
```
rberrelleza, line 49: this will load it for all commands, not sure the implications this might have with the stack's env file. 

A second option is to only load it for `okteto up`, but that would give us a different experience than `okteto down` and `okteto exec` when parsing `okteto.yaml`, and I don't think that's a good idea.

pchico83: I would be explicit as we do for okteto stacks:
https://okteto.com/docs/reference/stacks#env_file-string-optional

PR #1371, opened 2021-04-02
define the timeout in the manifest

### LINKED ISSUE #1197: Set timeout in okteto manifest
When working with big images, okteto up times out often, specially if the image is pretty large and not pulled. It would be useful to be able to specify a bigger timeout directly in the manifest, instead of setting an env var.

rberrelleza:
Fixes #1197 

## Proposed changes
- Allows a user to set up the timeout in the manifest, instead of using OKTETO_TIMEOUT

Currently, the timeout is per internal step, instead of for the whole "up" or "stack deploy" command. So specifying "30s" in the timeout really means that it will wait 30 seconds for each of the internal steps (pull image, wait for syncthing, etc..). In the future we should look into either having dedicated timeouts (pull timeout, deploy timeout, delete timeout), or have a single timeout for the entire command ("if up doesn't finish in a minute fail the task"). 

We currently apply the timeout on these steps:
- stack.Destroy
- deployments.UpdateOktetoRevision
- pod.GetDevPodInLoop
- volume.Destroy
- ssh.retryNewClientConn
- syncthing.WaitForPing
- syncthing.waitForFolderScanning


pchico83 (commented):
The method that pulls the dev image doesn't have a timeout:
https://github.com/okteto/okteto/blob/e5184701bd355be2af57614c78cd1d5df24e1b29/cmd/up/activate.go#L252
@rberrelleza do you know where is it timeout when using large images?

rberrelleza:
It times out when waiting for the dev pod to be ready. This isn't a new thing, is the reason why we added the OKTETO_TIMEOUT env var. 

We should rethink that timeout to differentiate between "the pod is stuck" to "the pod is taking a while but it's still going"

pchico83:
`It times out when waiting for the dev pod to be ready` -> could you point to the line where it times out? Looking at the code I still don't understand where it is happening

rberrelleza:
> `It times out when waiting for the dev pod to be ready` -> could you point to the line where it times out? Looking at the code I still don't understand where it is happening

https://github.com/okteto/okteto/pull/1371/files#diff-3702cbbd530526e96e9c30cf2139f248c0072b0bb4b88dd6a28edce5a019dfd8R108

pchico83:
> > `It times out when waiting for the dev pod to be ready` -> could you point to the line where it times out? Looking at the code I still don't understand where it is happening
> 
> https://github.com/okteto/okteto/pull/1371/files#diff-3702cbbd530526e96e9c30cf2139f248c0072b0bb4b88dd6a28edce5a019dfd8R108

That loop waits for the pod to be created, but it does not include the time to pull the image. I am trying to understand how a large image can produce a timeout in `okteto up `

pchico83 (approved):

PR #1375, opened 2021-04-04
Inject remote stignore as a secret instead of using the syncthing API

pchico83:
The send stignore syncthing API timeouts almost every time because it blocks if synchthing is synchronizing files.
Having stignore in the remote syncthing is important to optimize the syncthing of files, reduce scans, ...
This PR injects the stignore file as a secret to make sure the file is available before syncthing starts

### DIFF
--- a/cmd/up/stignore.go
+++ b/cmd/up/stignore.go
@@ -14,17 +14,88 @@
 package up
 
 import (
+	"bufio"
+	"crypto/md5"
 	"fmt"
+	"io"
 	"io/ioutil"
+	"os"
+	"path"
 	"path/filepath"
 	"strings"
 
 	initCMD "github.com/okteto/okteto/cmd/init"
 	"github.com/okteto/okteto/cmd/utils"
+	"github.com/okteto/okteto/pkg/config"
 	"github.com/okteto/okteto/pkg/linguist"
 	"github.com/okteto/okteto/pkg/log"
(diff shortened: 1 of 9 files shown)


cmd/up/stignore.go
```diff
@@ -14,18 +14,85 @@
...
+		infile, err := os.Open(stignorePath)
+		if err != nil {
+			return err
+		}
+		defer infile.Close()
+		reader := bufio.NewReader(infile)
+
+		stignoreName := fmt.Sprintf("stignore-%d", i+1)
+		transformedStignorePath := filepath.Join(config.GetDeploymentHome(dev.Namespace, dev.Name), stignoreName)
+		outfile, err := os.OpenFile(transformedStignorePath, os.O_CREATE|os.O_WRONLY, 0644)
```
rberrelleza, line 49: can we move the transformation to a separate function so it's easier to test and to understand what's happening?

rberrelleza (approved):
What happens when the user changes the .stignore file after doing up the first time? will this cause a dev container redeploy?

jLopezbarb (approved):

pchico83:
> What happens when the user changes the .stignore file after doing up the first time? will this cause a dev container redeploy?

Good point. I added logic to force redeployment in that case. Pods don't get secret updates on the fly

PR #1382, opened 2021-04-07
Allow stack removal by name even if the stack yaml has syntax errors

pchico83:
This is especially relevant for people deploying stacks from pipelines

### DIFF
--- a/cmd/stack/destroy.go
+++ b/cmd/stack/destroy.go
@@ -20,6 +20,7 @@ import (
 	"github.com/okteto/okteto/pkg/analytics"
 	"github.com/okteto/okteto/pkg/cmd/stack"
 	"github.com/okteto/okteto/pkg/log"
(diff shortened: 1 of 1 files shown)


cmd/stack/destroy.go
```diff
@@ -35,12 +36,16 @@ func Destroy(ctx context.Context) *cobra.Command {
 		RunE: func(cmd *cobra.Command, args []string) error {
 			s, err := utils.LoadStack(name, stackPath)
 			if err != nil {
-				return err
+				if name == "" {
+					return err
+				}
+				s = &model.Stack{Name: name}
```
rberrelleza, line 42: do you think we should add a log, and maybe ask for re-confirmation? (not sure if there's a risk of deleting the wrong things if the manifest is malformed).

rberrelleza (approved):

PR #1404, opened 2021-04-12
Set env var if exists on local

### LINKED ISSUE #1399: Environment variables without the = sign are not getting populated
**Describe the bug**
When using the following okteto.yaml:
```
name: hello-world
image: nginx:alpine
environment:
- MY_VAR
```

The $MY_VAR env var is not getting populated when doing `okteto up`. The  `MY_VAR`  variable exists, but the value its empty.

**Expected behavior**
Based on our docs, the value should be populated if `MY_VAR` exists in my environment.

 - OS: macOS
 - Version 1.12.5

jLopezbarb:

Fixes #1399 

## Proposed changes
- If there is no = on an environment variable declaration, set its value if it is declared locally.
- It was affecting `Okteto up` and `Okteto stack` commands

### DIFF
--- a/pkg/model/serializer.go
+++ b/pkg/model/serializer.go
@@ -73,7 +73,11 @@ func (e *EnvVar) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	}
 
 	e.Name, err = ExpandEnv(parts[0])
-	return err
+	if err != nil {
+		return err
+	}
+	e.Value = os.Getenv(e.Name)
+	return nil
 }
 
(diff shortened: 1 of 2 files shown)


pkg/model/serializer.go
```diff
@@ -73,7 +73,11 @@ func (e *EnvVar) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	}
 
 	e.Name, err = ExpandEnv(parts[0])
-	return err
+	if err != nil {
```
pchico83, line 76: Could you add a test for this case?

pchico83 (approved):

PR #1410, opened 2021-04-14
Fix linter issues in bash scripts

rberrelleza:
Fix issues raised by the linter in the bash scripts

### DIFF
--- a/images/bin/start.sh
+++ b/images/bin/start.sh
@@ -1,7 +1,7 @@
 #!/bin/sh
 
 log(){
-  echo $(date +%Y-%m-%dT%H:%M:%S) $1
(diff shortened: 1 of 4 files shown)


images/bin/start.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/sh
```
rberrelleza, line 1: @pchico83 @jLopezbarb what's the best way to test changes on this file?

pchico83: You need to create a new `bin` image, build okteto and run `okteto up` with a manifest with secrets.

rberrelleza: tested!

pchico83 (approved):

PR #1420, opened 2021-04-16
Add support for args

jLopezbarb:

Fixes #

## Proposed changes
- Replace `command` and `args` in favor of `entrypoint` and `command` 
-  Add support for `args` field if there is not `command` or `entrypoint` field (retrocompatibility)
- This PR is a breaking change in `okteto stacks` because it will no longer support `command` field as it was. It will always be the same as previous `args` field.


pkg/model/stack_serializer.go
```diff
@@ -212,7 +213,14 @@ func (serviceRaw *ServiceRaw) ToService(svcName string) (*Service, error) {
 	s.CapDrop = serviceRaw.CapDrop
 
 	s.Command.Values = serviceRaw.Command.Values
-	s.Entrypoint.Values = serviceRaw.Entrypoint.Values
+	if len(s.Command.Values) == 0 {
+		s.Command.Values = serviceRaw.Args.Values
+	}
+	if len(serviceRaw.Entrypoint.Values) > 0 {
```
pchico83, line 219: why do you need this `if`?

cmd/utils/stack.go
```diff
@@ -30,16 +31,19 @@ var (
 //LoadStack loads an okteto stack manifest checking "yml" and "yaml"
 func LoadStack(name, stackPath string) (*model.Stack, error) {
 	if model.FileExists(stackPath) {
-		return model.GetStack(name, stackPath)
+		return model.GetStack(name, stackPath, false)
```
pchico83, line 34: what is the name is `docker-compose.yml`?

pchico83 (approved):

PR #1421, opened 2021-04-16
Jlopezbarb/compose fix ports

jLopezbarb:
Fixes #

## Proposed changes
-  Add tests for when to set a service to public
-  Set service to public when the port is in `xxxx:xxxx` notation

### DIFF
--- a/pkg/model/stack_serializer.go
+++ b/pkg/model/stack_serializer.go
@@ -224,7 +224,7 @@ func (serviceRaw *ServiceRaw) ToService(svcName string) (*Service, error) {
 	s.Public = serviceRaw.Public
 
 	for _, p := range serviceRaw.Ports {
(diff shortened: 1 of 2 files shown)


pkg/model/stack_serializer.go
```diff
@@ -224,7 +224,7 @@ func (serviceRaw *ServiceRaw) ToService(svcName string) (*Service, error) {
 	s.Public = serviceRaw.Public
 
 	for _, p := range serviceRaw.Ports {
-		if !s.Public && isPublicPort(p.ContainerPort) {
+		if (!s.Public && isPublicPort(p.ContainerPort)) || p.HostPort != 0 {
```
pchico83, line 227:
```suggestion
		if p.HostPort != 0 {
```

pchico83 (approved):

PR #1422, opened 2021-04-16
moar linting issues

rberrelleza:
## Proposed changes
- cleaning up code


rberrelleza:
I tested the changes by doing "okteto up" on an oketo.yaml that has secrets, and everything worked fine.

PR #1428, opened 2021-04-19
Add error if user has not the permissions for ns

### LINKED ISSUE #1417: okteto namespace should log an error if I don't have access to the namespace
**Describe the bug**
When I run  "okteto namespace foo", the call succeeds. It's only when I try to access a resource of the namespace that I get an error message.

**To Reproduce**
1. okteto login
2. okteto namespace not-mine
3. kubectl get pod

**Expected behavior**
I expect okteto namespace to fail if the namespace doesn't exist, or if I don't have access to ti.

 - OS: macos
 - Version 1.12.6

jLopezbarb:

Fixes #1417 

## Proposed changes
-  Checks if user can get the namespace before updating context. If the user has not access to the namespace it throws an error
-

### DIFF
--- a/cmd/namespace/namespace.go
+++ b/cmd/namespace/namespace.go
@@ -22,6 +22,8 @@ import (
 	"github.com/okteto/okteto/pkg/cmd/login"
 	"github.com/okteto/okteto/pkg/config"
 	"github.com/okteto/okteto/pkg/errors"
+	k8Client "github.com/okteto/okteto/pkg/k8s/client"
+	"github.com/okteto/okteto/pkg/k8s/namespaces"
 	"github.com/okteto/okteto/pkg/log"
 	"github.com/okteto/okteto/pkg/okteto"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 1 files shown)


cmd/namespace/namespace.go
```diff
@@ -120,3 +126,16 @@ func askOktetoURL() (string, error) {
...
+
+func isNamespaceAvailable(ctx context.Context, namespace string) error {
+	client, _, err := k8Client.GetLocal()
+	if err != nil {
+		return err
+	}
+
+	_, err = namespaces.Get(ctx, namespace, client)
+	if err != nil {
+		return fmt.Errorf("Cannot get '%s' namespace. Please make sure it's created and you have the right permissions.", namespace)
```
rberrelleza, line 138: Could we add the second sentence as a `hint` instead?

```suggestion
		return fmt.Errorf("Namespace '%s' not found. Please verify that the namespace exists and that you have access to it.", namespace)
```

pchico83 (approved):

PR #1429, opened 2021-04-19
Remove warning message if restart is always

### LINKED ISSUE #1425: Support for "restart" in Okteto Stacks

Okteto Stacks defaults to `restart: always`. If this is the value, we shouldn't show a warning. Keep the warning for other values and, in the future, translate those options to the equivalent job definitionjobs.

jLopezbarb:

Fixes #1425

## Proposed changes
-  If restart field is set to 'always' not show the warning at the end
-

### DIFF
--- a/pkg/model/stack_serializer.go
+++ b/pkg/model/stack_serializer.go
@@ -112,7 +112,7 @@ type ServiceRaw struct {
 	Profiles          *WarningType `yaml:"profiles,omitempty"`
 	PullPolicy        *WarningType `yaml:"pull_policy,omitempty"`
 	ReadOnly          *WarningType `yaml:"read_only,omitempty"`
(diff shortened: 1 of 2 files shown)


pkg/model/stack_serializer.go
```diff
@@ -732,7 +732,9 @@ func getServiceNotSupportedFields(svcName string, svcInfo *ServiceRaw) []string
 		notSupported = append(notSupported, fmt.Sprintf("services[%s].read_only", svcName))
 	}
 	if svcInfo.Restart != nil {
-		notSupported = append(notSupported, fmt.Sprintf("services[%s].restart", svcName))
```
pchico83, line 735: let's add a test for this case

pchico83 (approved):

PR #1430, opened 2021-04-19
Move display field warning after build images

jLopezbarb:

Fixes #

## Proposed changes
-  This PR shows the field not supported warning after building all the images, otherwise the user may not see the warning because of buildkit output.
-


pkg/model/stack.go
```diff
@@ -373,3 +368,11 @@ func GroupWarningsBySvc(fields []string) []string {
 	}
 	return result
 }
+
+func (s *Stack) DisplayWarnings() {
```
rberrelleza, line 372: could we keep this function on the  `cmd` package, since they are used to display info in the UI, and they are not part of the data model?

pkg/model/stack.go
```diff
@@ -373,3 +368,11 @@ func GroupWarningsBySvc(fields []string) []string {
...
+	if len(s.Warnings) > 0 {
+		notSupportedFields := strings.Join(GroupWarningsBySvc(s.Warnings), "\n  - ")
+		log.Warning("The following fields are not supported in this version and will be omitted: \n  - %s", notSupportedFields)
```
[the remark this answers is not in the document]
rberrelleza: When there's only a single attribute, the list looks weird. Could we have a different message for when there's only one unrecognized field VS two or more?

pchico83 (approved):

pkg/cmd/stack/deploy.go
```diff
@@ -236,3 +237,15 @@ func waitForPodsToBeRunning(ctx context.Context, s *model.Stack, c *kubernetes.C
 	}
 	return fmt.Errorf("kubernetes is taking too long to create your stack. Please check for errors and try again")
 }
+
+func DisplayWarnings(s *model.Stack) {
+	if len(s.Warnings) > 0 {
+		if len(s.Warnings) == 1 {
+			log.Warning("'%s' field is not currently supported and will be ignored.", s.Warnings[0])
+		} else {
+			notSupportedFields := strings.Join(model.GroupWarningsBySvc(s.Warnings), "\n  - ")
```
pchico83, line 246: Let's create a new field for volume mount warning, so they are also displayed after building the images

PR #1431, opened 2021-04-19
Add isCompose file metrics to analytics

jLopezbarb:

Fixes #

## Proposed changes
-  Add metrics to know if the user is using an Okteto stack or a docker compose file
-

### DIFF
--- a/pkg/analytics/analytics.go
+++ b/pkg/analytics/analytics.go
@@ -58,7 +58,6 @@ const (
 	signupEvent              = "Signup"
 	disableEvent             = "Disable Analytics"
 	stackNotSupportedField   = "Stack Field Not Supported"
-	stackIsComposeFile       = "Stack Is compose file"
 )
 
 var (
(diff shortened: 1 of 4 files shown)


pkg/analytics/analytics.go
```diff
@@ -58,6 +58,7 @@ const (
 	signupEvent              = "Signup"
 	disableEvent             = "Disable Analytics"
 	stackNotSupportedField   = "Stack Field Not Supported"
+	stackIsComposeFile       = "Stack Is compose file"
```
rberrelleza, line 61:
```suggestion
	stackIsComposeFile       = "Stack Is Compose File"
```

rberrelleza: cc @rlamana

pchico83: I would add an extra property to the `Stack Deploy` event

pchico83 (approved):

PR #1437, opened 2021-04-21
Support --redeploy=false to deploy a pipeline only if the repo is not…

pchico83:
… already deployed

This makes it possible to define cyclic dependencies between pipelines

### DIFF
--- a/cmd/pipeline/deploy.go
+++ b/cmd/pipeline/deploy.go
@@ -37,7 +37,7 @@ func deploy(ctx context.Context) *cobra.Command {
 	var name string
 	var namespace string
 	var wait bool
-	var redeploy bool
+	var skipIfExists bool
 	var timeout time.Duration
 
 	cmd := &cobra.Command{
@@ -95,7 +95,7 @@ func deploy(ctx context.Context) *cobra.Command {
 				log.Information("Pipeline context: %s", okteto.GetURL())
 			}
 
-			if !redeploy {
+			if skipIfExists {
 				_, err := okteto.GetPipelineByRepository(ctx, namespace, repository)
 				if err == nil {
 					log.Success("Pipeline '%s' was already deployed", name)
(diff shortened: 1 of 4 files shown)


jLopezbarb (approved):

rberrelleza:
What do you mean by cyclic dependencies @pchico83 ?

rberrelleza:
I think it's odd that `--redeploy=true` is the default, given that the whole purpose of the command is to deploy the pipeline. 

Have we considered instead having a flag to not run the pipeline if already exists? `--skip-if-exists`

rlamana:
@rberrelleza @pchico83 I don't think it is odd for default behavior, but I do like `--skip-if-exists`. Searching for it I actually found many CLIs using a flag with the same name, so I guess that the rule of the "most common use" wins here.

pchico83:
@rlamana @rberrelleza I will switch to `--skip-if-exists`

pchico83:
> What do you mean by cyclic dependencies @pchico83 ? Could you explain the use case this enables a bit more?

@rberrelleza when you have pipelines that call other pipelines, it is sometimes easier to define the pipeline with cyclic references, especially when deploying preview environments. For example, let's say you have 5 repositories, and there is a PR on repo A.
It is easy to say deploy the pipeline of repo A from current branch, and then, call a metapipeline that deploys the 5 repos. If the metapipeline deploys repo A with `--skip-if-exists` everything is fine.
We should think about a more declarative definition for this, but this unblocks several scenarios for now.

rberrelleza:
> > What do you mean by cyclic dependencies @pchico83 ? Could you explain the use case this enables a bit more?
> 
> @rberrelleza when you have pipelines that call other pipelines, it is sometimes easier to define the pipeline with cyclic references, especially when deploying preview environments. For example, let's say you have 5 repositories, and there is a PR on repo A.
> It is easy to say deploy the pipeline of repo A from current branch, and then, call a metapipeline that deploys the 5 repos. If the metapipeline deploys repo A with `--skip-if-exists` everything is fine.
> We should think about a more declarative definition for this, but this unblocks several scenarios for now.

The part I don't get is why do you need skip-if-exists for that scenario? Don't we want all the dependencies to be redeployed together?

pchico83:
> > > What do you mean by cyclic dependencies @pchico83 ? Could you explain the use case this enables a bit more?
> > 
> > 
> > @rberrelleza when you have pipelines that call other pipelines, it is sometimes easier to define the pipeline with cyclic references, especially when deploying preview environments. For example, let's say you have 5 repositories, and there is a PR on repo A.
> > It is easy to say deploy the pipeline of repo A from current branch, and then, call a metapipeline that deploys the 5 repos. If the metapipeline deploys repo A with `--skip-if-exists` everything is fine.
> > We should think about a more declarative definition for this, but this unblocks several scenarios for now.
> 
> The part I don't get is why do you need skip-if-exists for that scenario? Don't we want all the dependencies to be redeployed together?

@rberrelleza it makes the preview pipeline logic harder. For example, you deploy repo A in branch `test`, and then, if you deploy the metapipeline, the metapipeline will redeploy repo A in branch `master`. This is not the case if the metapipeline does `okteto pipeline deploy -r repoA --if-not-exists`
Another setup is when you have repo A and B. Pipeline A deploys A and B (if it doesn't exist), and pipeline B deploys B and A (if it doesn't exist). Then, deploying A or B will deploys all your needed components, without entering cyclic loops.
All this should be handled in more declarative way by application templates, but for now, it unblocks several scenarios that customers are asking for

PR #1444, opened 2021-04-22
Refactor for envs vars, labels and annotations

### LINKED ISSUE #1415: Replace envvars in labels and annotations
I want to add an annotation/label based on envvars. For example:
```
dev.okteto.com/from-snapshot-name: ${SNAPSHOT_NAME}
```
This should be possible for both, okteto manifest and okteto stack manifests.

jLopezbarb:
Fixes #1415

## Proposed changes
-  Now these three fields works in the same way when unmarshalling them.
-  They can accept map and list syntax.
-  When unmarshalling you can use env vars as key or value

### DIFF
--- a/pkg/model/dev.go
+++ b/pkg/model/dev.go
@@ -127,7 +127,7 @@ type Dev struct {
 	Image                *BuildInfo            `json:"image,omitempty" yaml:"image,omitempty"`
 	Push                 *BuildInfo            `json:"-" yaml:"push,omitempty"`
 	ImagePullPolicy      apiv1.PullPolicy      `json:"imagePullPolicy,omitempty" yaml:"imagePullPolicy,omitempty"`
-	Environment          Environments          `json:"environment,omitempty" yaml:"environment,omitempty"`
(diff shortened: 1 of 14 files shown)


pkg/model/dev.go
```diff
@@ -127,7 +127,7 @@ type Dev struct {
 	Image                *BuildInfo            `json:"image,omitempty" yaml:"image,omitempty"`
 	Push                 *BuildInfo            `json:"-" yaml:"push,omitempty"`
 	ImagePullPolicy      apiv1.PullPolicy      `json:"imagePullPolicy,omitempty" yaml:"imagePullPolicy,omitempty"`
-	Environment          []EnvVar              `json:"environment,omitempty" yaml:"environment,omitempty"`
+	Environment          Environments          `json:"environment,omitempty" yaml:"environment,omitempty"`
```
pchico83, line 130:
```suggestion
	Environment          Environment          `json:"environment,omitempty" yaml:"environment,omitempty"`
```

pchico83 (approved):

PR #1446, opened 2021-04-23
Skip well known port to be public if not specified

### LINKED ISSUE #1442: Omit well known ports from being public
I want to avoid well known ports like the ones from mongodb, mysql, rabbit... to expose a service.

If the service has one of these ports it should be always private and don't expose any service.

jLopezbarb:

Fixes #1442 

## Proposed changes
-  If a well known port(DB ports) is written with the syntax (xxxx:xxxx) it will not set the public to public

### DIFF
--- a/pkg/model/stack_serializer.go
+++ b/pkg/model/stack_serializer.go
@@ -348,7 +348,9 @@ func (p *PortRaw) UnmarshalYAML(unmarshal func(interface{}) error) error {
 			return fmt.Errorf("Can not convert %s to a port.", hostString)
 		}
 		p.HostPort = int32(port)
-
+		if IsSkippablePort(p.HostPort) {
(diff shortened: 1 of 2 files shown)


pkg/model/stack_serializer.go
```diff
@@ -373,6 +375,20 @@ func (p *PortRaw) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	return nil
 }
 
+func IsSkippablePort(port int32) bool {
+	relationalDBPortList := []int32{7210, 3306, 1521, 1830, 5432, 1433, 3050}
```
pchico83, line 379: Let's use a map instead of a list so it makes the check operation a constant time operation.

pchico83 (approved):

PR #1449, opened 2021-04-23
Split command and entrypoint without "sh -c"

### LINKED ISSUE #1441: Error to deploy voting app sample
**Describe the bug**
I deployed the repo https://github.com/dockersamples/example-voting-app to Okteto Cloud, but `vote` and `result` failed to start. Looks like the issue is that we are not splitting the `command` in different string when using the `command` single string syntax.

**To Reproduce**
Deploy https://github.com/dockersamples/example-voting-app from Okteto Cloud

**Expected behavior**
It works

**Additional context**
We can use a library like https://github.com/kballard/go-shellquote to split the string

jLopezbarb:

Fixes #1441 

## Proposed changes
- Split entrypoint, command and args instead of using the prefix "sh -c"

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -27,6 +27,7 @@ require (
 	github.com/google/uuid v1.1.2
 	github.com/hashicorp/go-getter v1.5.0
 	github.com/heroku/docker-registry-client v0.0.0-20190909225348-afc9e1acc3d5
+	github.com/kballard/go-shellquote v0.0.0-20180428030007-95032a82bc51
 	github.com/machinebox/graphql v0.2.2
 	github.com/manifoldco/promptui v0.3.2
 	github.com/matryer/is v1.2.0 // indirect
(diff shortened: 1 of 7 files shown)


pchico83 (approved):

rberrelleza:
Since now we don't be starting a shell, could this impact existing strict, or env var interpolation?

pchico83:
> Since now we don't be starting a shell, could this impact existing strict, or env var interpolation?

I think the most relevant affected scenario is when using `command: a && b`, but I think we should tell the user to use `sh` in the command for these cases because longer term it is very confusing that the behavior of `command` changes if you use a single string or a list of strings. Also, we are adding a dependency on `sh` (for jib based images, for example, this is problematic).

@jLopezbarb Let's do this logic just for èntrypoint`, `args` and for `command` (is `command` is a compose file) for now, and we will do `command` for stacks later after talking tocurrent users.

PR #1453, opened 2021-04-26
Support for enable/disable pod lifecycle events

pchico83:
The implementation is similar to the `probes` support.

### DIFF
--- a/pkg/cmd/init/run.go
+++ b/pkg/cmd/init/run.go
@@ -17,12 +17,14 @@ import (
 	"context"
 	"fmt"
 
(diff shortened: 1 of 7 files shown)


rberrelleza:
@pchico83 could you add more information on what use case is this enabling?

pchico83:
> @pchico83 could you add more information on what use case is this enabling?

a pod does not start if `postStart` fails. Sometimes, `postStart` needs the code to be synched in the container. When using the okteto persistent volume this is an issue, because `postStart` fails and the pod fails to run.

rberrelleza (approved):

PR #1454, opened 2021-04-26
Add simplified endpoint declaration

### LINKED ISSUE #1450: Support reduced notation for "endpoints"
Apps usually only have an endpoint. The endpoint name seems redundant in this case, we could get it from the stack name.

For this case:
```
name: movies
endpoints:
   movies:
       - path: /
          service: frontend
          port: 80
       - path: /api
          service: api
          port: 8080
```
supports this notation (which should be the notation documented in the first place): 
```
name: movies
endpoints:
     - path: /
        service: frontend
        port: 80
     - path: /api
        service: api
        port: 8080
```
and:
```
name: movies
endpoints:
   annotations:
      - kubernetes.io/ingress.class: "nginx"
   rules:
       - path: /
          service: frontend
          port: 80
       - path: /api
          service: api
          port: 8080
```

jLopezbarb:

Fixes #1450

## Proposed changes
-  Declare one endpoint directly, the name of the endpoint will be the name of the stack
-

### DIFF
--- a/pkg/model/stack_serializer_test.go
+++ b/pkg/model/stack_serializer_test.go
@@ -651,6 +651,37 @@ endpoints:
 			},
 		},
 		{
+			name: "direct rules with labels/annotations",
+			manifest: []byte(`name: test
+services:
+  app:
+    ports:
+    - 9213
+    image: okteto/vote:1
+endpoints:
+  annotations:
+    key: value
+  labels:
+    key: value
+  rules:
+  - path: /
+    service: app
+    port: 9213`),
+			expected: EndpointSpec{
+				"test": Endpoint{
+					Annotations: Annotations{"key": "value"},
+					Labels:      Labels{"key": "value"},
+					Rules: []EndpointRule{
+						{
+							Service: "app",
+							Path:    "/",
+							Port:    9213,
+						},
+					},
+				},
+			},
+		},
+		{
 			name: "direct rules",
 			manifest: []byte(`name: test
 services:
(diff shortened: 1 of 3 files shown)


pkg/model/stack_serializer_test.go
```diff
@@ -584,3 +584,108 @@ func Test_restartFile(t *testing.T) {
...
+		},
+		{
+			name: "direct rules",
+			manifest: []byte(`name: test
+services:
+  app:
+    ports:
+    - 9213
+    image: okteto/vote:1
+endpoints:
```
pchico83, line 661: is this supported:
```
endpoints:
    annotations:
      key: value
    labels:
      key: value
    rules:
    - path: /
      service: app
      port: 9213`),
```

pkg/model/stack_serializer_test.go
```diff
@@ -584,3 +584,108 @@ func Test_restartFile(t *testing.T) {
...
+  - path: /
+    service: app
```
rberrelleza, line 663: for the simple mode, do we need to specify the service too? since there's only one, maybe we can assume it?

pkg/model/stack_serializer_test.go
jLopezbarb: Yes, I have covered that on a new test

pkg/model/stack_serializer_test.go
jLopezbarb: I think that if you just want to create an endpoint for a stack with one service, setting the service to public will be the easiest way to have access to that.

pchico83: @jLopezbarb I agree

pchico83 (approved):

PR #1457, opened 2021-04-27
Simplify image name from registry

jLopezbarb:

Fixes #

## Proposed changes
-  When the image that is pulling comes from the Okteto Registry, show okteto.dev/{image name} instead of full name registry

### DIFF
--- a/cmd/up/activate.go
+++ b/cmd/up/activate.go
@@ -3,6 +3,7 @@ package up
 import (
 	"context"
 	"fmt"
+	"regexp"
 	"strings"
 
 	"github.com/okteto/okteto/cmd/utils"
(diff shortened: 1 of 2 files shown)


cmd/up/activate.go
```diff
@@ -340,3 +341,9 @@ func (up *upContext) waitUntilDevelopmentContainerIsRunning(ctx context.Context)
 		}
 	}
 }
+
+func getPullingMessage(message string) string {
+	var oktetoRegistryRegex = regexp.MustCompile(`registry\..*\.okteto\.net\/.*\/`)
```
pchico83, line 346: https://github.com/okteto/okteto/blob/master/pkg/okteto/auth.go#L230 gives you the url of the registry, and `up.Dev.Namesapace` gives you the current namespace. If you append both values it gives you what should be replaced by `okteto.dev`

pchico83 (approved):

PR #1460, opened 2021-04-28
Jlopezbarb/fix named volumes permissions

jLopezbarb:
Fixes #

## Proposed changes
-  Run folder permission on the folder created instead of the root folder for volumes
-

### DIFF
--- a/pkg/cmd/stack/translate.go
+++ b/pkg/cmd/stack/translate.go
@@ -327,10 +327,10 @@ func getInitContainerCommandAndVolumeMounts(svc model.Service) ([]string, []apiv
 			if !addedDataVolume {
 				volumeMounts = append(volumeMounts, apiv1.VolumeMount{Name: volumeName, MountPath: "/data"})
 				if command == "" {
-					command = "chmod 777 /data/*"
+					command = "chmod 777 /data/"
 					addedDataVolume = true
 				} else {
-					command += " && chmod 777 /data/*"
+					command += " && chmod 777 /data/"
 				}
 			}
 		}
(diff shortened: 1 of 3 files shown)


pkg/cmd/stack/translate.go
```diff
@@ -310,18 +310,16 @@ func getInitContainerCommandAndVolumeMounts(svc model.Service) ([]string, []apiv
...
-			if !addedVolumesVolume {
-				if command == "" {
-					command = "chmod 777 /volumes/"
-					addedVolumesVolume = true
-				} else {
-					command += " && chmod 777 /volumes/"
-				}
+
+			if command == "" {
+				command = fmt.Sprintf("chmod 777 /volumes/%s", volumeName)
```
pchico83, line 320: Would a single `chmod 777 /volumes/*` work?

pkg/cmd/stack/translate.go
```diff
@@ -310,26 +310,27 @@ func getInitContainerCommandAndVolumeMounts(svc model.Service) ([]string, []apiv
...
+					command += " && chmod 777 /volumes/*"
+				}
 			}
-
 		} else {
 			if !addedDataVolume {
 				volumeMounts = append(volumeMounts, apiv1.VolumeMount{Name: volumeName, MountPath: "/data"})
 				if command == "" {
-					command = "chmod 777 /data"
+					command = "chmod 777 /data/*"
```
pchico83, line 330: in this case `/data` is already the mount point, I think we don't need `/*` here

pchico83 (approved):

PR #1462, opened 2021-04-29
Added support for variables on pipeline deploy command

ifbyol:

Fixes #

## Proposed changes
-  Add support for variables on pipeline deploy command

### DIFF
--- a/cmd/pipeline/deploy.go
+++ b/cmd/pipeline/deploy.go
@@ -128,7 +128,7 @@ func deploy(ctx context.Context) *cobra.Command {
 	cmd.Flags().BoolVarP(&wait, "wait", "w", false, "wait until the pipeline finishes (defaults to false)")
 	cmd.Flags().BoolVarP(&skipIfExists, "skip-if-exists", "", false, "skip the pipeline deployment if the pipeline already exists in the namespace (defaults to false)")
 	cmd.Flags().DurationVarP(&timeout, "timeout", "t", (5 * time.Minute), "the length of time to wait for completion, zero means never. Any other values should contain a corresponding time unit e.g. 1s, 2m, 3h ")
-	cmd.Flags().StringArrayVarP(&variables, "var", "v", []string{}, "set a pipeline variable (can be used multiple times)")
+	cmd.Flags().StringArrayVarP(&variables, "var", "v", []string{}, "set a pipeline variable (can be set more than once)")
 	return cmd
 }
 
--- a/cmd/pipeline/deploy.go
+++ b/cmd/pipeline/deploy.go
@@ -39,6 +39,7 @@ func deploy(ctx context.Context) *cobra.Command {
 	var wait bool
 	var skipIfExists bool
 	var timeout time.Duration
+	var variables []string
 
 	cmd := &cobra.Command{
 		Use:   "deploy",
@@ -106,7 +107,7 @@ func deploy(ctx context.Context) *cobra.Command {
 				}
 			}
 
-			if err := deployPipeline(ctx, name, namespace, repository, branch, wait, timeout); err != nil {
+			if err := deployPipeline(ctx, name, namespace, repository, branch, wait, timeout, variables); err != nil {
 				return err
 			}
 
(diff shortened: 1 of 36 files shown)


pchico83 (approved):

cmd/pipeline/deploy.go
```diff
@@ -127,16 +128,28 @@ func deploy(ctx context.Context) *cobra.Command {
 	cmd.Flags().BoolVarP(&wait, "wait", "w", false, "wait until the pipeline finishes (defaults to false)")
 	cmd.Flags().BoolVarP(&skipIfExists, "skip-if-exists", "", false, "skip the pipeline deployment if the pipeline already exists in the namespace (defaults to false)")
 	cmd.Flags().DurationVarP(&timeout, "timeout", "t", (5 * time.Minute), "the length of time to wait for completion, zero means never. Any other values should contain a corresponding time unit e.g. 1s, 2m, 3h ")
+	cmd.Flags().StringArrayVarP(&variables, "var", "v", []string{}, "set a pipeline variable (can be used multiple times)")
```
rberrelleza, line 131:
```suggestion
	cmd.Flags().StringArrayVarP(&variables, "var", "v", []string{}, "set a pipeline variable (can be set more than once)")
```

cmd/pipeline/deploy.go
```diff
@@ -127,16 +128,28 @@ func deploy(ctx context.Context) *cobra.Command {
...
 }
 
-func deployPipeline(ctx context.Context, name, namespace, repository, branch string, wait bool, timeout time.Duration) error {
+func deployPipeline(ctx context.Context, name, namespace, repository, branch string, wait bool, timeout time.Duration, variables []string) error {
 	spinner := utils.NewSpinner("Deploying your pipeline...")
 	spinner.Start()
 	defer spinner.Stop()
 
+	varList := []okteto.Variable{}
+	for _, v := range variables {
```
rberrelleza, line 141: could we make this a common function? (or even reuse the code we have to parse the env vars, it looks similar)

cmd/pipeline/deploy.go
```diff
@@ -127,16 +128,28 @@ func deploy(ctx context.Context) *cobra.Command {
...
+		kv := strings.SplitN(v, "=", 2)
+		if len(kv) != 2 {
```
rberrelleza, line 143: would it make sense to support flag-like variables?

pkg/okteto/client.go
```diff
@@ -85,6 +85,30 @@ func query(ctx context.Context, query string, result interface{}) error {
 	return nil
 }
 
+func queryWithVariables(ctx context.Context, query string, variables map[string]interface{}, result interface{}) error {
```
rberrelleza, line 88: could we add a test for this? not end to end, but just to make sure we are sending the right thing to the graphql endpoint? I'm concerned that, since this could be any value, that we might need to escape things.

rberrelleza: Should we also check for command injection, since this is a free-formed value?

ifbyol: We could, but we are delegating that logic into the third party library [`machinebox/graphql`](https://github.com/machinebox/graphql), so adding tests for that would mean to add tests to their library.

It would probably need to be an end to end test (maybe with fake API server like they are doing on their tests), because the library receives a Query object and not the actual string representation that will be sent to the graphql endpoint.

cmd/pipeline/deploy.go
ifbyol: What would be the value of the variable? `true` or a behavior similar to `-e`  on docker compose. https://docs.docker.com/compose/environment-variables/#pass-environment-variables-to-containers

`The value of the DEBUG variable in the container is taken from the value for the same variable in the shell in which Compose is run.`

cmd/pipeline/deploy.go
ifbyol: Which logic we have to parse env vars? I saw a couple of places with similar code, but at the end, they are doing different things with the variables once they are extracted, so the only common logic is to split a string into var name and value

pkg/okteto/client.go
ifbyol: At the end, I've changed this function. Since variables are defined in the query string, it makes sense that variables are set in the same place where the query is defined. In this case, `pipeline.go`. The idea is that all variables should be set using their types instead of using free-formed types like `interface{}`

pchico83 (approved):

PR #1465, opened 2021-04-30
Add volumes to images

### LINKED ISSUE #1439: Add mount volumes to the Dockerfile in Okteto Stacks
It is quite common to use data volumes in docker-compose to modify an official image. For example, to add a static file to nginx, you can do:

```
  nginx:
    image: nginx
    ports:
      - 80:80
    volumes:
      - ./public:/usr/share/nginx/html
```

This doesn't work in Okteto Stacks because volume mounts are ignored.

When there is a mount volume as the sample above, automatically add the Dockerfile instruction `COPY ./public /usr/share/nginx/html` to the Dockerfile, or autogenerate a Dockerfile if it doesn't exist. This way, this docker-compose file will work in Okteto Stacks.

jLopezbarb:

Fixes #1439

## Proposed changes
-  Build an image if there are volumes that are necessary to run the application (nginx config file)
-

### DIFF
--- a/pkg/cmd/stack/translate.go
+++ b/pkg/cmd/stack/translate.go
@@ -122,7 +122,7 @@ func translateBuildImages(ctx context.Context, s *model.Stack, forceBuild, noCac
 	if err != nil {
 		return err
 	}
-	hasAddedAnyVolumeMounts, err := addVolumeMountsToBuiltImage(ctx, s, buildKitHost, isOktetoCluster, forceBuild, noCache)
+	hasAddedAnyVolumeMounts, err := addVolumeMountsToBuiltImage(ctx, s, buildKitHost, isOktetoCluster, forceBuild, noCache, hasBuiltSomething)
 	if err != nil {
 		return err
 	}
@@ -168,12 +168,12 @@ func buildServices(ctx context.Context, s *model.Stack, buildKitHost string, isO
 	return hasBuiltSomething, nil
 }
 
-func addVolumeMountsToBuiltImage(ctx context.Context, s *model.Stack, buildKitHost string, isOktetoCluster, forceBuild, noCache bool) (bool, error) {
+func addVolumeMountsToBuiltImage(ctx context.Context, s *model.Stack, buildKitHost string, isOktetoCluster, forceBuild, noCache, hasBuiltSomething bool) (bool, error) {
 	hasAddedAnyVolumeMounts := false
 	var err error
 	for name, svc := range s.Services {
 		if len(svc.VolumeMounts) != 0 {
-			if !hasAddedAnyVolumeMounts {
+			if !hasBuiltSomething && !hasAddedAnyVolumeMounts {
 				hasAddedAnyVolumeMounts = true
 				log.Information("Running your build in %s...", buildKitHost)
 			}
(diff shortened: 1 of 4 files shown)


pkg/cmd/stack/translate.go
```diff
@@ -137,25 +152,58 @@ func translateBuildImages(ctx context.Context, s *model.Stack, forceBuild, noCac
...
 		svc.SetLastBuiltAnnotation()
 		s.Services[name] = svc
 		log.Success("Image for service '%s' successfully pushed", name)
 	}
+	return hasBuiltSomething, nil
+}
 
-	if !building && forceBuild {
-		log.Warning("Ignoring '--build' argument. There are not 'build' primitives in your stack")
+func addVolumeMountsToSvcs(ctx context.Context, s *model.Stack, buildKitHost string, isOktetoCluster, forceBuild, noCache bool) (bool, error) {
```
pchico83, line 171:
```suggestion
func addVolumeMountsToBuiltImage(ctx context.Context, s *model.Stack, buildKitHost string, isOktetoCluster, forceBuild, noCache bool) (bool, error) {
```

pkg/cmd/stack/translate.go
```diff
@@ -137,25 +152,58 @@ func translateBuildImages(ctx context.Context, s *model.Stack, forceBuild, noCac
...
+	hasAddedAnyVolumeMounts := false
+	var err error
+	for name, svc := range s.Services {
+		if len(svc.VolumeMounts) != 0 {
+			if !hasAddedAnyVolumeMounts {
+				hasAddedAnyVolumeMounts = true
+				log.Information("Running your build in %s...", buildKitHost)
```
pchico83, line 178: check that `hasBuiltSomething` is false?

pkg/cmd/stack/translate.go
```diff
@@ -137,25 +152,58 @@ func translateBuildImages(ctx context.Context, s *model.Stack, forceBuild, noCac
...
+					return hasAddedAnyVolumeMounts, err
+				}
+			}
+			svcBuild, err := registry.CreateDockerfileWithVolumeMounts(fromImage, svc.VolumeMounts)
+			if err != nil {
+				return hasAddedAnyVolumeMounts, err
+			}
+			svc.Build = svcBuild
+			if isOktetoCluster && !strings.HasPrefix(svc.Image, "okteto.dev") {
+				tag := strings.Replace(svc.Image, ":", "-", 1)
```
pchico83, line 193: let's always create a new image tag for this, for example:
`okteto.dev/{stack_name}-{service_name}:okteto-with-volume-mounts`

pkg/cmd/stack/translate.go
```diff
@@ -137,25 +152,58 @@ func translateBuildImages(ctx context.Context, s *model.Stack, forceBuild, noCac
...
+				svc.Image = fmt.Sprintf("okteto.dev/%s:okteto", tag)
+			}
+			log.Information("Building image for service '%s'...", name)
```
pchico83, line 196:
```suggestion
			log.Information("Building image for service '%s' to include host volumes...", name)
```

pkg/cmd/stack/translate.go
jLopezbarb: That checks if a previous service has created any mount volume on other image, so it only gets into it the first time

pchico83: yeap, but if `hasBuiltSomething` is true, this has been already printed, not?
the idea is to show the logs only on the first image been built

pchico83 (approved):

PR #1467, opened 2021-04-30
Add exposed ports from images without exposed port

jLopezbarb:

Fixes: When you upload a docker-compose with a service without exposed ports, none of the ports are available in the cluster

## Proposed changes
-  Check docker registry to get the default exposed ports that an image must have
-

### DIFF
--- a/pkg/registry/registry.go
+++ b/pkg/registry/registry.go
@@ -155,7 +155,7 @@ func GetRegistryAndRepo(tag string) (string, string) {
 	if len(splittedImage) == 1 {
 		imageTag = splittedImage[0]
 	} else if len(splittedImage) == 2 {
-		if splittedImage[0] == okteto.DevRegistry {
+		if strings.Contains(splittedImage[0], ".") {
 			return splittedImage[0], splittedImage[1]
 		}
 		imageTag = strings.Join(splittedImage[len(splittedImage)-2:], "/")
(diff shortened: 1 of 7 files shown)


pkg/registry/registry.go
```diff
@@ -152,3 +162,48 @@ func GetRegistryAndRepo(tag string) (string, string) {
 	}
 	return registryTag, imageTag
 }
+
+func GetHiddenExposePorts(image string) []int32 {
+	exposedPorts := make([]int32, 0)
+	c, err := NewRegistryClient("https://registry.hub.docker.com", "", "")
```
rberrelleza, line 168: what if the image is not in docker hub? (e.g. in our own)

jLopezbarb: For now it will only inspect docker hub public images. I think this will cover most of the compose files. When using their own image the user usually set the ports in the expose field. This port injection is costly, so it is good to perform it as few times as possible.

pchico83: Would this work with images with these formats?
- docker.io/bitnami/mongodb
- mongodb

jLopezbarb: New commits adds support for public images from other repositories. It also adds support for private images on our own registry.

jLopezbarb: This will work with both formats. It will have issues when the image is a official one and adds the registry. For example, `docker.io/mongo`.

pchico83 (approved):

PR #1468, opened 2021-05-03
Fixes for 1.12.9 release

pchico83:
## Proposed changes
-  Rename stack field `workingDir` to `workdir` to match the okteto manifest field.
-  Remove redundant stack volume field `name`. Just keep it for compose compatibility.
-  Remove `storage` indirection from stack volumes.

### DIFF
--- a/pkg/cmd/stack/translate.go
+++ b/pkg/cmd/stack/translate.go
@@ -204,7 +204,7 @@ func translateDeployment(svcName string, s *model.Stack) *appsv1.Deployment {
 							Ports:           translateContainerPorts(svc),
 							SecurityContext: translateSecurityContext(svc),
 							Resources:       translateResources(svc),
-							WorkingDir:      svc.WorkingDir,
+							WorkingDir:      svc.Workdir,
 						},
 					},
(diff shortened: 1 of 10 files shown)


pkg/cmd/stack/translate.go
```diff
@@ -327,10 +327,10 @@ func getInitContainerCommandAndVolumeMounts(svc model.Service) ([]string, []apiv
 			if !addedDataVolume {
 				volumeMounts = append(volumeMounts, apiv1.VolumeMount{Name: volumeName, MountPath: "/data"})
 				if command == "" {
-					command = "chmod 777 /data/"
+					command = "chmod 777 /data/*"
```
rberrelleza, line 330: do we need to do `chmod +R` so it's applied to the subdirectories and files?

pchico83: No, we want to avoid that. If you modify the permissions of files created by the container, the container might fail. We do `*` because okteto creates first-level subfolders at this level

jLopezbarb (approved):

PR #1484, opened 2021-05-06
Skip well known ports

### LINKED ISSUE #1480: Ignore host ports of well known remote debuggers
I don't want public endpoints for my PHP debugger when deploying a compose file.

jLopezbarb:

Fixes #1480 

## Proposed changes
-  Add debugger ports to be ignored when exposing the ports


pkg/model/stack_serializer.go
```diff
@@ -630,8 +630,8 @@ func getProtocol(protocolName string) (apiv1.Protocol, error) {
 }
 
 func sanitizeName(name string) string {
```
pchico83, line 632: who is calling this function? why are we changing it?

pchico83: @jLopezbarb ^^

PR #1486, opened 2021-05-06
Jlopezbarb/autogenerate env file

jLopezbarb:
Fixes #

## Proposed changes
-  If an env file named '.env'(maybe we will need to do this for all the files independently of its name) is detected and is in Okteto namespace it will omit the content of the file and will get the environment from the Okteto secrets.
-  This behavior is good for when redeploying an Okteto stack/Docker compose from the UI where we don't have access to the files that the user may have when deploying it from the cli

### DIFF
--- a/pkg/cmd/stack/translate.go
+++ b/pkg/cmd/stack/translate.go
@@ -67,13 +67,13 @@ func translate(ctx context.Context, s *model.Stack, forceBuild, noCache bool) er
 func translateStackEnvVars(ctx context.Context, s *model.Stack) error {
 	var err error
 	isOktetoNamespace := namespaces.IsOktetoNamespaceFromName(ctx, s.Namespace)
-	for _, svc := range s.Services {
+	for svcName, svc := range s.Services {
 		svc.Image, err = model.ExpandEnv(svc.Image)
 		if err != nil {
 			return err
 		}
 		for _, envFilepath := range svc.EnvFiles {
-			if err := translateServiceEnvFile(ctx, svc, envFilepath, isOktetoNamespace); err != nil {
+			if err := translateServiceEnvFile(ctx, svc, svcName, envFilepath, isOktetoNamespace); err != nil {
 				return err
 			}
 		}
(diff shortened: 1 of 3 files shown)


rberrelleza:
This is a great feature!

pkg/cmd/stack/translate.go
```diff
@@ -83,17 +85,41 @@ func translateStackEnvVars(s *model.Stack) error {
...
-func translateServiceEnvFile(svc *model.Service, filename string) error {
+func translateServiceEnvFile(ctx context.Context, svc *model.Service, svcName, filename string, isOktetoNamespace bool) error {
 	var err error
 	filename, err = model.ExpandEnv(filename)
 	if err != nil {
 		return err
 	}
 
+	if filename == ".env" && isOktetoNamespace {
+		log.Information("[%s]: Detected '.env' file but will be skipped. You can set environment variables from https://cloud.okteto.com/#/settings/secrets.", svcName)
```
rberrelleza, line 96: why do we log this? since it's pretty common to have an .env file, this might look like we are telling users not to use that feature no?

rberrelleza: What happens if user has a local .env, but it's deploying on a namespace managed by okteto? maybe the .env generation is something that doesn't happen when deploying local, but only when inside the pipeline?

pchico83: @jLopezbarb let's do as we do for variable interpolation. Instead of omitting the content of `.env`, let's merge its content with the Okteto Secrets with more priority to the content of `.env`:
https://okteto.com/docs/reference/stacks#environment-variables

pchico83: And we can remove the log information message then

pkg/cmd/stack/translate.go
```diff
@@ -83,16 +85,39 @@ func translateStackEnvVars(s *model.Stack) error {
 	return nil
 }
 
-func translateServiceEnvFile(svc *model.Service, filename string) error {
+func translateServiceEnvFile(ctx context.Context, svc *model.Service, svcName, filename string, isOktetoNamespace bool) error {
 	var err error
 	filename, err = model.ExpandEnv(filename)
```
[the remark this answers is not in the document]
jLopezbarb: Right now we are only failing when the file does not exist and we couldn't get any secrets. It could be that the user is not on a Okteto cluster or because we could not get any secret because a failure on the API.

pchico83 (approved):

PR #1493, opened 2021-05-07
Sanitize service names

### LINKED ISSUE #1419: Accept "_" in stack service names
**Describe the bug**
`_` should be sanitized to `-`

jLopezbarb:

Fixes #1419

## Proposed changes
-  Services with underscore or whitespace are sanitized replacing those characters into -
-  Groups stack warnings(not supported fields, sanitized names and volume mount) into one struct instead of having three different fields for each warning

### DIFF
--- a/cmd/stack/deploy.go
+++ b/cmd/stack/deploy.go
@@ -50,7 +50,7 @@ func Deploy(ctx context.Context) *cobra.Command {
 			if err != nil {
 				return err
 			}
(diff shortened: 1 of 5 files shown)


rberrelleza:
Changing the name of the service will affect the DNS name too, which will have runtime implications if the user is hardcoding host names in code (e.g. if my code is calling redis_replica:1234 it will now fail). Maybe we should fail instead?

pchico83 (approved):

PR #1500, opened 2021-05-10
Fix stack volume mounting error on private images

jLopezbarb:

Fixes When a private image had a volume mounting was creating a image name with / which causes permission error

## Proposed changes
-  replace all / with -


pkg/cmd/stack/translate.go
```diff
@@ -192,7 +192,7 @@ func addVolumeMountsToBuiltImage(ctx context.Context, s *model.Stack, buildKitHo
 			}
 			svc.Build = svcBuild
 			if isOktetoCluster && !strings.HasPrefix(svc.Image, "okteto.dev") {
-				tag := strings.Replace(svc.Image, ":", "-", 1)
+				tag := getNewTagFromImage(svc.Image)
```
pchico83, line 195: could we  create stackname + servicename as the image name?

jLopezbarb: Yes, I think it's a better approach 👍

pchico83 (approved):

PR #1502, opened 2021-05-10
skip volume mounts with not found paths

jLopezbarb:
Fixes #

## Proposed changes
-  If a volume mount can't be mounted because the relative path is missing adds a warning and skip it.
-  This will be useful for when deploying an Okteto stack/docker-compose from the UI from a git URL


pkg/cmd/stack/translate.go
```diff
@@ -207,6 +208,20 @@ func addVolumeMountsToBuiltImage(ctx context.Context, s *model.Stack, buildKitHo
 	return hasAddedAnyVolumeMounts, nil
 }
 
+func getAccessibleVolumeMounts(stack *model.Stack, svcName string) []model.StackVolume {
```
pchico83, line 211: can we have a unit test for this function?

pchico83 (approved):

PR #1519, opened 2021-05-12
Jlopezbarb/fix permission error on volume mount

jLopezbarb:
Fixes #

## Proposed changes
-  Add -R flag to init container command, otherwise it fails in some cases
-


pkg/cmd/stack/translate.go
```diff
@@ -431,10 +431,10 @@ func getInitContainerCommandAndVolumeMounts(svc model.Service) ([]string, []apiv
 			if !addedDataVolume {
 				volumeMounts = append(volumeMounts, apiv1.VolumeMount{Name: volumeName, MountPath: "/data"})
 				if command == "" {
-					command = "chmod 777 /data/*"
+					command = "chmod 777 -R /data"
```
pchico83, line 434: it should be `chmod 777 /data`, because it is just that folder in this case. 
`chmod 777 /data/*` is not chmoding `/data/.` which is the one we want to change

pchico83 (approved):

PR #1523, opened 2021-05-12
Support for okteto up with divert

### LINKED ISSUE #933: Add labels and annotations to the service and deployment created by okteto
I have a running application. Instead of "swapping" a deployment, I want to be able to spin up a "dev" copy of a deployment and have isitio redirect traffic. "okteto up" already creates the svc and deployment for me, but istio requires specific labels to be applied.

I'd like to be able to tell to okteto to add certain labels/annotations to the service and deployments it creates, so it can integrate with other tools.

I can run a script post okteto up, but it adds more complexity when compared to just okteto up.

pchico83:
Fixes #933

This PR adds support to divert a deployment on `okteto up`. Instead of replacing the original deployment, divert will create a new ingress, a new service, a service mesh layer, and a cloned deployment to develop without affecting the original deployment.

This feature requires CRDs and operators only available in Okteto Cloud/Enterprise.

### DIFF
--- a/cmd/down.go
+++ b/cmd/down.go
@@ -23,6 +23,7 @@ import (
 	"github.com/okteto/okteto/pkg/errors"
 	k8Client "github.com/okteto/okteto/pkg/k8s/client"
 	"github.com/okteto/okteto/pkg/k8s/deployments"
+	"github.com/okteto/okteto/pkg/k8s/diverts"
 	"github.com/okteto/okteto/pkg/k8s/volumes"
 	"github.com/okteto/okteto/pkg/log"
 	"github.com/okteto/okteto/pkg/model"
@@ -91,6 +92,10 @@ func runDown(ctx context.Context, dev *model.Dev) error {
 
 	client, _, err := k8Client.GetLocalWithContext(dev.Context)
 	if err != nil {
+		return err
+	}
+
+	if err := diverts.Delete(ctx, dev, client); err != nil {
 		return err
 	}
 
--- a/cmd/down.go
+++ b/cmd/down.go
@@ -95,8 +95,10 @@ func runDown(ctx context.Context, dev *model.Dev) error {
 		return err
 	}
 
-	if err := diverts.Delete(ctx, dev, client); err != nil {
-		return err
+	if dev.Divert != nil {
+		if err := diverts.Delete(ctx, dev, client); err != nil {
+			return err
+		}
 	}
 
 	d, err := deployments.Get(ctx, dev, dev.Namespace, client)
--- a/cmd/up/up.go
+++ b/cmd/up/up.go
@@ -29,19 +29,21 @@ import (
 	buildCMD "github.com/okteto/okteto/pkg/cmd/build"
 	"github.com/okteto/okteto/pkg/config"
 	"github.com/okteto/okteto/pkg/errors"
-	k8Client "github.com/okteto/okteto/pkg/k8s/client"
+	k8sClient "github.com/okteto/okteto/pkg/k8s/client"
 	"github.com/okteto/okteto/pkg/k8s/deployments"
+	"github.com/okteto/okteto/pkg/k8s/diverts"
+	"github.com/okteto/okteto/pkg/k8s/ingressesv1"
 	"github.com/okteto/okteto/pkg/k8s/namespaces"
 	"github.com/okteto/okteto/pkg/log"
 	"github.com/okteto/okteto/pkg/model"
 	"github.com/okteto/okteto/pkg/okteto"
 	"github.com/okteto/okteto/pkg/registry"
 	"github.com/okteto/okteto/pkg/ssh"
-
 	"github.com/okteto/okteto/pkg/syncthing"
 
 	"github.com/spf13/cobra"
 	appsv1 "k8s.io/api/apps/v1"
+	"k8s.io/client-go/kubernetes"
 )
 
 // ReconnectingMessage is the message shown when we are trying to reconnect
@@ -56,12 +58,11 @@ func Up() *cobra.Command {
 	var autoDeploy bool
 	var build bool
 	var forcePull bool
-	var resetSyncthing bool
+	var reset bool
 	cmd := &cobra.Command{
 		Use:   "up",
 		Short: "Activates your development container",
 		RunE: func(cmd *cobra.Command, args []string) error {
-
 			if okteto.InDevContainer() {
 				return errors.ErrNotInDevContainer
 			}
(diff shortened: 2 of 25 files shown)


cmd/down.go
```diff
@@ -94,6 +95,10 @@ func runDown(ctx context.Context, dev *model.Dev) error {
 		return err
 	}
 
+	if err := diverts.Delete(ctx, dev, client); err != nil {
```
rberrelleza, line 98: can we rename to diverts.DeleteIfNeeded (or something like that) so it's more obvious that the method will be a no-op if there's no divert? the other option is to have the caller do that check. it's more verbose, but I think it helps with code readability.

cmd/up/up.go
```diff
@@ -247,7 +250,7 @@ func (up *upContext) start(autoDeploy, build bool) error {
 	stop := make(chan os.Signal, 1)
 	signal.Notify(stop, os.Interrupt)
 
-	analytics.TrackUp(true, up.Dev.Name, up.getInteractive(), len(up.Dev.Services) == 0, up.isSwap, up.Dev.RemoteModeEnabled())
+	analytics.TrackUp(true, up.Dev.Name, up.getInteractive(), len(up.Dev.Services) == 0, up.isSwap, up.Dev.Divert != nil)
```
rberrelleza, line 253: could we have a separate event for divert?

cmd/up/up.go
```diff
@@ -503,27 +506,38 @@ func (up *upContext) shutdown() {
...
 		for i := 1; i < len(dev.Reverse); i++ {
-			log.Println(fmt.Sprintf("               %d <- %d", dev.Reverse[i].Local, dev.Reverse[i].Remote))
+			log.Println(fmt.Sprintf("                %d <- %d", dev.Reverse[i].Local, dev.Reverse[i].Remote))
+		}
+	}
+
+	if dev.Divert != nil {
+		username := okteto.GetSanitizedUsername()
+		name := diverts.DivertName(username, dev.Divert.Ingress)
+		i, err := ingressesv1.Get(ctx, name, dev.Namespace, c)
```
rberrelleza, line 536: can we get the ingresses on a separate step, and keep this function just to display. that way there's no wait between printing the forward and then the divert information.

cmd/up/up.go
```diff
@@ -503,27 +506,38 @@ func (up *upContext) shutdown() {
...
+		if err != nil {
+			log.Infof("error getting diverted ingress %s: %s", name, err.Error())
```
rberrelleza, line 538: this should be an error no?

cmd/up/up.go
```diff
@@ -503,27 +506,38 @@ func (up *upContext) shutdown() {
...
+		} else if len(i.Spec.Rules) > 0 {
+			log.Println(fmt.Sprintf("    %s %s", log.BlueString("Divert URL:"), i.Spec.Rules[0].Host))
```
rberrelleza, line 540: I don't think we should use the name `Divert` in the messages,  since it's not a name that tells the user anything about the feature or the usage. Could we instead just say it's the application or deve environment URL?

pkg/errors/errors.go
```diff
@@ -83,6 +83,9 @@ var (
 
 	// ErrDevPodDeleted raised if dev pod is deleted in the middle of the "okteto up" sequence
 	ErrDevPodDeleted = fmt.Errorf("development container has been removed")
+
+	//ErrDivertNotSupported raised if the divert feature is not supported in the current cluster
+	ErrDivertNotSupported = fmt.Errorf("the 'divert' field is only supported in namespaces managed by Okteto")
```
rberrelleza, line 88: can we add a hint to send the user to a page where she can learn more about this?

pkg/okteto/auth.go
```diff
@@ -206,6 +210,16 @@ func GetUsername() string {
 	return t.Username
 }
 
+// GetSanitizedUsername returns the username of the authenticated user sanitized to be DNS compatible
+func GetSanitizedUsername() string {
```
rberrelleza, line 214: what do we need this for? Should we instead get the API to return this in a kube-friendly way? I feel like this is server-side behavior that the CLI shouldn't have to know about.

rberrelleza: if not, is there kube library we can use for this? I think there are other rules we might be missing here (e.g. using dashes, underscores, or that a name can't end on an underscore)

pkg/model/dev.go
```diff
@@ -79,6 +79,10 @@ const (
 	OktetoRestartAnnotation = "dev.okteto.com/restart"
 	//OktetoStignoreAnnotation indicates the hash of the stignore files to force redeployment
 	OktetoStignoreAnnotation = "dev.okteto.com/stignore"
+	//OktetoDivertLabel indicates the object is a diverted version
+	OktetoDivertLabel = "dev.okteto.com/divert"
+	//OktetoDivertServiceModificationAnnotation indicates the service modification done by diverting a service
+	OktetoDivertServiceModificationAnnotation = "divert.okteto.com/modification"
```
rberrelleza, line 85: can you think of a way to make sure that this are consistent between cloud and cli?

pkg/k8s/diverts/types.go
```diff
@@ -0,0 +1,142 @@
...
+		in, out := &in.Items, &out.Items
+		*out = make([]Divert, len(*in))
+		for i := range *in {
+			(*in)[i].DeepCopyInto(&(*out)[i])
+		}
+	}
+}
+
+// DeepCopy a deepcopy function, copying the receiver, creating a new DivertList.
+func (in *DivertList) DeepCopy() *DivertList {
```
rberrelleza, line 97: can we add a unit test for this part? deepcopy might not do what we expect it to do in all the use cases we are handling.

cmd/up/up.go
pchico83: it is just another field in the okteto manifest to configure the `up` behavior. it makes it simple to compare the usage of divert against other `up` executions

pkg/errors/errors.go
pchico83: yes, but we don't have it yet. I will add it when we have docs about divert

pkg/model/dev.go
jmacelroy: I've created an issue for this. I don't know exactly which code will end up in each spot but some of these things are shared by the cli and the operator and will end up in the weaver repo and imported as a dep into cli. There is another set of code that builds on top of that shared code and makes additions to it that is just for cli. 

I don't think it needs to happen for the first iteration but it needs to be addressed before we build more on top of it or it will get harder to fix. https://github.com/okteto/app/issues/2333

PR #1526, opened 2021-05-13
Add new e2e test

jLopezbarb:

## Proposed changes
Adds e2e tests for compose that test:  
 - mounted volumes
 - named volumes relative path
 - non named volumes
 - Port inference from docker hub and external registry(ECR)
 - container name
 - Command split
 - service connection

### DIFF
--- a/integration/stacks_test.go
+++ b/integration/stacks_test.go
@@ -17,6 +17,7 @@ package integration
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"log"
 	"os"
@@ -30,9 +31,11 @@ import (
 )
 
 const (
-	stackGitRepo   = "git@github.com:okteto/stacks-getting-started.git"
-	stackGitFolder = "stacks-getting-started"
-	stackManifest  = "okteto-stack.yml"
+	stackGitRepo     = "git@github.com:okteto/stacks-getting-started.git"
+	stackGitFolder   = "stacks-getting-started"
+	stackManifest    = "okteto-stack.yml"
+	composeGitRepo   = "git@github.com:okteto/flask-producer-consumer.git"
+	composeGitFolder = "flask-producer-consumer"
 )
 
 func TestStacks(t *testing.T) {
@@ -68,7 +71,7 @@ func TestStacks(t *testing.T) {
 
 		defer deleteGitRepo(ctx, stackGitFolder)
 
-		if err := deployStack(ctx, oktetoPath, stackManifest); err != nil {
+		if err := deployStack(ctx, oktetoPath, stackManifest, stackGitFolder); err != nil {
 			t.Fatal(err)
 		}
 
@@ -83,7 +86,7 @@ func TestStacks(t *testing.T) {
 		if !strings.Contains(content, "Cats vs Dogs!") {
 			t.Fatalf("wrong stack content: %s", content)
 		}
-		if err := destroyStack(ctx, oktetoPath, stackManifest); err != nil {
+		if err := destroyStack(ctx, oktetoPath, stackManifest, stackGitFolder); err != nil {
 			t.Fatal(err)
 		}
 
@@ -127,11 +130,11 @@ func deleteGitRepo(ctx context.Context, path string) error {
 	return nil
 }
 
-func deployStack(ctx context.Context, oktetoPath, stackPath string) error {
+func deployStack(ctx context.Context, oktetoPath, stackPath, dir string) error {
 	log.Printf("okteto stack deploy %s", stackPath)
 	cmd := exec.Command(oktetoPath, "stack", "deploy", "-f", stackPath, "--build", "--wait")
 	cmd.Env = os.Environ()
-	cmd.Dir = stackGitFolder
+	cmd.Dir = dir
 	o, err := cmd.CombinedOutput()
 	if err != nil {
 		return fmt.Errorf("okteto stack deploy failed: %s - %s", string(o), err)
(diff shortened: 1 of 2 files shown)


integration/stacks_test.go
```diff
@@ -30,9 +31,11 @@ import (
...
-	stackGitFolder = "stacks-getting-started"
-	stackManifest  = "okteto-stack.yml"
+	stackGitRepo     = "git@github.com:okteto/stacks-getting-started.git"
+	stackGitFolder   = "stacks-getting-started"
+	stackManifest    = "okteto-stack.yml"
+	composeGitRepo   = "git@github.com:okteto/flask-producer-consumer.git"
+	composeGitFolder = "flask-producer-consumer"
 )
 
 func TestStacks(t *testing.T) {
```
rberrelleza, line 41: do you think is worth unifying both tests to save time? the integration job already takes a long time.

integration/stacks_test.go
```diff
@@ -140,15 +143,123 @@ func deployStack(ctx context.Context, oktetoPath, stackPath string) error {
...
+	}
+
+	tName := fmt.Sprintf("TestStacks-%s", runtime.GOOS)
+	name := strings.ToLower(fmt.Sprintf("%s-%d", tName, time.Now().Unix()))
+	namespace := fmt.Sprintf("%s-%s", name, user)
+
+	t.Run(tName, func(t *testing.T) {
+		log.Printf("running %s \n", tName)
+		k8Client.Reset()
+	})
```
rberrelleza, line 178: since we are not running this test in two modes, you don't need to wrap the code in a t.Run(XXX  function.

integration/stacks_test.go
```diff
@@ -140,15 +143,123 @@ func deployStack(ctx context.Context, oktetoPath, stackPath string) error {
...
+	log.Printf("created namespace %s \n", namespace)
+
+	if err := cloneGitRepo(ctx, composeGitRepo); err != nil {
+		t.Fatal(err)
+	}
+
+	if err := deployStack(ctx, oktetoPath, "docker-compose.yml", composeGitFolder); err != nil {
+		t.Fatal(err)
+	}
+	defer deleteGitRepo(ctx, composeGitFolder)
```
rberrelleza, line 192: is composeGitFolder under a temp folder?

integration/stacks_test.go
```diff
@@ -140,15 +143,123 @@ func deployStack(ctx context.Context, oktetoPath, stackPath string) error {
...
+
+	log.Printf("deployed stack using %s \n", "docker-compose.yml")
+
+	endpoint := fmt.Sprintf("https://nginx-%s.cloud.okteto.net/db", namespace)
```
rberrelleza, line 196: could we make the url parametric, so we can also run these tests on our dev environments?

integration/stacks_test.go
```diff
@@ -140,15 +143,123 @@ func deployStack(ctx context.Context, oktetoPath, stackPath string) error {
...
+	}
+	_, err = getDeployment(ctx, namespace, "web-svc")
+	if err == nil {
+		t.Fatalf("'web-svc' deployment not deleted after 'okteto stack destroy'")
+	}
+	if !strings.Contains(err.Error(), "not found") {
+		t.Fatalf("error getting deployment 'web-svc': %s", err.Error())
+	}
+
+	_, err = getStatefulset(ctx, namespace, "mongodb")
```
rberrelleza, line 244: would it make sense to also check that the volume is destroyed?

rberrelleza (approved):

integration/stacks_test.go
jLopezbarb: If I unify them I will need to modify the test to cover all the things that we support now in compose

integration/stacks_test.go
jLopezbarb: yes it is, It will clone the git repo into a temp folder and after the test finish removes it

integration/stacks_test.go
jLopezbarb: The destroy command doesn't destroy volumes by default. You have to specify it by adding `-v` flag

integration/stacks_test.go
jLopezbarb: I will do two separate test. One for compose checking that is working and the other to test our own fields like endpoints

integration/stacks_test.go
```diff
@@ -140,15 +142,165 @@ func deployStack(ctx context.Context, oktetoPath, stackPath string) error {
...
+	log.Println("Stack has been redeployed succesfully")
+
+	if err := destroyStack(ctx, oktetoPath, "docker-compose.yml", composeGitFolder); err != nil {
+		t.Fatal(err)
+	}
+
+	log.Println("destroyed stack")
+
+	time.Sleep(5 * time.Second)
+	_, err = getDeployment(ctx, namespace, "api")
```
pchico83, line 229: You could list by the stack label and check the response is empty

integration/stacks_test.go
```diff
@@ -140,15 +142,165 @@ func deployStack(ctx context.Context, oktetoPath, stackPath string) error {
...
+
+	_, err = getDeployment(ctx, namespace, "web-svc")
+	if err == nil {
+		t.Fatalf("'web-svc' deployment not deleted after 'okteto stack destroy'")
+	}
+	if !strings.Contains(err.Error(), "not found") {
+		t.Fatalf("error getting deployment 'web-svc': %s", err.Error())
+	}
+
+	_, err = getStatefulset(ctx, namespace, "mongodb")
```
pchico83, line 261: same

integration/stacks_test.go
```diff
@@ -140,15 +142,165 @@ func deployStack(ctx context.Context, oktetoPath, stackPath string) error {
...
+	if !strings.Contains(err.Error(), "not found") {
+		t.Fatalf("error getting statefulset 'rabbitmq': %s", err.Error())
+	}
+
+	_, err = getVolume(ctx, namespace, "rabbitmq-data")
+	if err != nil {
+		t.Fatalf("'rabbitmq-data' volume deleted after 'okteto stack destroy'")
+	}
+
+	_, err = getVolume(ctx, namespace, "mongodb-data")
```
pchico83, line 282: same

integration/stacks_test.go
pchico83: it will also make the logic more reusable between compose files

pchico83: also, we should check services and ingress are also deleted

PR #1527, opened 2021-05-14
Better error messages for Docker support

pchico83:

@rberrelleza @melissa-okteto could you review the texts?

### DIFF
--- a/pkg/model/dev.go
+++ b/pkg/model/dev.go
@@ -73,6 +73,8 @@ const (
 	DefaultSyncthingRescanInterval = 300
 	//RemoteSubPath subpath in the development container persistent volume for the remote data
 	RemoteSubPath = "okteto-remote"
+	//OktetoURLAnnotation indicates the okteto cluster public url
+	OktetoURLAnnotation = "dev.okteto.com/url"
 	//OktetoAutoCreateAnnotation indicates if the deployment was auto generatted by okteto up
 	OktetoAutoCreateAnnotation = "dev.okteto.com/auto-create"
 	//OktetoRestartAnnotation indicates the dev pod must be recreated to pull the latest version of its image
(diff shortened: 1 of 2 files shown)


pkg/model/dev.go
```diff
@@ -707,7 +709,7 @@ func (dev *Dev) validate() error {
 	}
 
 	if dev.Docker.Enabled && !dev.PersistentVolumeEnabled() {
-		return fmt.Errorf("Persistent volume is required to enable Docker support")
+		return fmt.Errorf("Docker support requires the okteto persistent volume to be enabled")
```
rberrelleza, line 712:
```suggestion
		return fmt.Errorf("Docker support requires persistent volume to be enabled")
```

rberrelleza: could we add a hint and point them to the docs on this?

cmd/up/up.go
```diff
@@ -232,6 +232,10 @@ func (up *upContext) start(autoDeploy, build bool) error {
 		return fmt.Errorf("'okteto up' is not allowed in the current namespace")
 	}
 
+	if up.Dev.Docker.Enabled && ns.Annotations[model.OktetoURLAnnotation] == okteto.CloudURL {
+		return fmt.Errorf("Docker support requires privileged containers and it's not supported in Okteto Cloud")
```
rberrelleza, line 236: do you think it's important to talk about privileged containers? Maybe we just mention that this feature is not available on Cloud?

rberrelleza: Also, what's going to happen on OE if privileged containers are not enabled? do we handle that well?

pchico83: You get the standard error, something like this:
```
 x  couldn't activate your development container
    pods "api-57879cc69c-" is forbidden: PodSecurityPolicy: unable to admit pod: [spec.containers[1].securityContext.privileged: Invalid value: true: Privileged containers are not allowed]
```

rberrelleza: could catch that error for DinD and show something more user-friendly? (since it's not the user the one that's adding the privileged container).

pchico83: which message would you show?

PR #1533, opened 2021-05-14
Create volumes one by one and log it to the user

### LINKED ISSUE #1528: Refactor "translatePersistentVolumeClaims" to list `s.Volumes` and create them one by one
We should follow the same flow that we do for services and endpoint, and should a message for each volume created:
   https://github.com/okteto/okteto/blob/master/pkg/cmd/stack/deploy.go#L120

Also, for services, endpoints and volumes we should have a divverent message for `create` and `update` operations

jLopezbarb:

Fixes #1528

## Proposed changes
- Creates volumes one by one, after one is deployed it notificates the creation to the user 
-


pchico83 (approved):
@jLopezbarb can you also update the `log.Success` of endpoints from "Deployed endpoint..." to "Created endpoint..."'

PR #1538, opened 2021-05-17
Fix forbidden error when updating volumes stack

jLopezbarb:

Fixes: When there is a volume deployed and you want to update resources (Okteto stack deploy without running Okteto stack destroy), it gives you a forbidden error, because the volume name and storage class name are not set initially. These fields are given by Okteto cluster.

## Proposed changes
- This PR checks if the requested resources are the same from the previous instance.
- If the resources are the same, it doesn't update the volume, but if they are not equal updates the volume with the volume name and storage class name set instead of the new one

### DIFF
--- a/pkg/cmd/stack/deploy.go
+++ b/pkg/cmd/stack/deploy.go
@@ -229,8 +229,11 @@ func deployVolume(ctx context.Context, volumeName string, s *model.Stack, c *kub
 		if old.Labels[okLabels.StackNameLabel] != s.Name {
 			return fmt.Errorf("name collision: the volume '%s' belongs to the stack '%s'", pvc.Name, old.Labels[okLabels.StackNameLabel])
 		}
-		if err := volumes.Update(ctx, &pvc, c); err != nil {
-			return fmt.Errorf("error updating volume of service '%s': %s", pvc.Name, err.Error())
+		if !volumes.IsEqual(&pvc, old) {
+			old.Spec.Resources.Requests["storage"] = pvc.Spec.Resources.Requests["storage"]
+			if err := volumes.Update(ctx, old, c); err != nil {
+				return fmt.Errorf("error updating volume '%s': %s", old.Name, err.Error())
+			}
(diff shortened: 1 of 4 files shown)


pkg/cmd/stack/deploy.go
```diff
@@ -229,8 +229,11 @@ func deployVolume(ctx context.Context, volumeName string, s *model.Stack, c *kub
 		if old.Labels[okLabels.StackNameLabel] != s.Name {
 			return fmt.Errorf("name collision: the volume '%s' belongs to the stack '%s'", pvc.Name, old.Labels[okLabels.StackNameLabel])
 		}
-		if err := volumes.Update(ctx, &pvc, c); err != nil {
-			return fmt.Errorf("error updating volume of service '%s': %s", pvc.Name, err.Error())
+		if !volumes.IsEqual(&pvc, old) {
```
pchico83, line 232: Can we do this in every case? if they are equal, it should do nothing.
Also, can you try that it is possible to change the volume size and the change takes effect in the volume?

pkg/cmd/stack/deploy.go
```diff
@@ -229,8 +229,21 @@ func deployVolume(ctx context.Context, volumeName string, s *model.Stack, c *kub
 		if old.Labels[okLabels.StackNameLabel] != s.Name {
 			return fmt.Errorf("name collision: the volume '%s' belongs to the stack '%s'", pvc.Name, old.Labels[okLabels.StackNameLabel])
 		}
-		if err := volumes.Update(ctx, &pvc, c); err != nil {
-			return fmt.Errorf("error updating volume of service '%s': %s", pvc.Name, err.Error())
+
+		old.Spec.Resources.Requests["storage"] = pvc.Spec.Resources.Requests["storage"]
+		for key, value := range pvc.Labels {
+			old.Labels[key] = value
```
pchico83, line 235: should we remove the `old` labels?

PR #1541, opened 2021-05-17
Init okteto persistent volume from image content

pchico83:

Initializing the volume from the image speeds up the first synchronization. It can also be used to reused the `node_modules` dependencies installed in the dev image


pkg/k8s/deployments/translate.go
```diff
@@ -37,8 +37,8 @@ const (
 	revisionAnnotation         = "deployment.kubernetes.io/revision"
 	//OktetoBinName name of the okteto bin init container
 	OktetoBinName = "okteto-bin"
-	//OktetoInitFromImage name of the okteto init from image container
-	OktetoInitDataName = "okteto-init-data"
+	//OktetoInitVolumeName name of the okteto init container that initializes the persistent colume from image content
+	OktetoInitVolumeName = "okteto-init-volume"
```
rberrelleza, line 41: if it's already a volume, do you think it might be redundant to call it `-volume`?

rberrelleza (approved):

jLopezbarb (approved):

PR #1550, opened 2021-05-19
Improve "okteto init" logic for interpreted languages

pchico83:

For python, nodejs, php and ruby it is better to use the deployment image for development

### DIFF
--- a/pkg/cmd/init/run.go
+++ b/pkg/cmd/init/run.go
@@ -23,12 +23,15 @@ import (
 	"github.com/okteto/okteto/pkg/k8s/pods"
 	"github.com/okteto/okteto/pkg/k8s/replicasets"
 	"github.com/okteto/okteto/pkg/k8s/services"
+	"github.com/okteto/okteto/pkg/linguist"
+	"github.com/okteto/okteto/pkg/log"
 	"github.com/okteto/okteto/pkg/model"
 	"github.com/okteto/okteto/pkg/okteto"
 	appsv1 "k8s.io/api/apps/v1"
 	apiv1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
 )
 
 var (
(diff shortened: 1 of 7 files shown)


rberrelleza:
would the user still get the custom bash prompt?  this helps people distinguish between their local prompt and the remote terminal.

pkg/cmd/init/run.go
```diff
@@ -23,33 +23,52 @@ import (
...
 	pod, err := getRunningPod(ctx, d, container, c)
 	if err != nil {
 		return err
 	}
 
+	language = linguist.NormalizeLanguage(language)
+	updateImageFromPod := false
+	switch language {
+	case linguist.Javascript:
+		updateImageFromPod = pods.HasNodeModules(ctx, pod, container, config, c)
```
rberrelleza, line 57: for node we should also do the `init from volume` thing you implemented the other day, so they don't get replaced.

pkg/k8s/pods/pod.go
```diff
@@ -177,6 +177,16 @@ func GetUserByPod(ctx context.Context, p *apiv1.Pod, container string, config *r
...
 }
 
+//HasNodeModules returns if the container has node_modules
+func HasNodeModules(ctx context.Context, p *apiv1.Pod, container string, config *rest.Config, c *kubernetes.Clientset) bool {
+	cmd := []string{"sh", "-c", "[ -d 'node_modules' ] && echo 'node_modules exists'"}
+	out, err := execCommandInPod(ctx, p, container, cmd, config, c)
+	if err != nil {
+		return false
+	}
+	return strings.Contains(out, "ode_modules exists")
```
rberrelleza, line 187: is the `n` on `node_modules` missing o purpose? Also, could we just check the exit code instead of depending on  string?

pkg/linguist/dev.go
```diff
@@ -55,24 +57,27 @@ var (
 func init() {
 	languageDefaults = make(map[string]languageDefault)
 	forwardDefaults = make(map[string][]model.Forward)
-	languageDefaults[javascript] = languageDefault{
-		image: "okteto/node:12",
-		forward: []model.Forward{
+	languageDefaults[Javascript] = languageDefault{
+		image:   "okteto/node:12",
```
rberrelleza, line 61: can we update the default to node:14?

rberrelleza (approved):

jLopezbarb (approved):

pkg/cmd/init/run.go
pchico83: it is already the default behavior

PR #1552, opened 2021-05-19
Execute start hook if defined in cloudbin image

pchico83:

This makes it possible to run a start.sh hook on every dev container in the cluster

### DIFF
--- a/cmd/up/activate.go
+++ b/cmd/up/activate.go
@@ -127,8 +127,9 @@ func (up *upContext) activate(autoDeploy, build bool) error {
 		log.Debugf("clean command output: %s", output)
 
 		outByCommand := strings.Split(strings.TrimSpace(output), "\n")
-		if len(outByCommand) >= 2 {
-			version, watches := outByCommand[0], outByCommand[1]
+		var version, watches, hook string
+		if len(outByCommand) >= 3 {
+			version, watches, hook = outByCommand[0], outByCommand[1], outByCommand[2]
 
 			if isWatchesConfigurationTooLow(watches) {
 				folder := config.GetNamespaceHome(up.Dev.Namespace)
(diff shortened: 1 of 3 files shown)


images/bin/start.sh
```diff
@@ -14,6 +14,10 @@ if [ -d "/var/okteto/cloudbin" ]; then
         fi
 fi
 
+if [ -d "/var/okteto/hooks/start.sh" ]; then
+  /var/okteto/hooks/start.sh
```
rberrelleza, line 18: start.sh might not be executable. I think it's easier to just run `sh /var/okteto/hooks/start.sh` or source i

rberrelleza: if this command fails, will up fail?

pchico83: I didn't use `sh` to avoid a level of indirection, but I guess it is fine.
About the script failing, I would fail `okteto up`, it would be an unexpected behavior otherwise. If the user wants to ignore errors, she can handle them on their script

rberrelleza: Cool, let's just document that we expect start to be an executable in that case.

pchico83: @rberrelleza also, we are not showing the output, the user will need to do a `kubectl logs podname`.
I am expecting the hook to be fast, if it does things like installations, I would run it via exec (as we do for the cleanup) so we can provide interactive output

pchico83: I moved the logic to run the script client-side when needed

rberrelleza: What motivated the change? The log?

Also, is the command running after all the files are synchronized or before all that?

pchico83: Yes, it is to have more control over the logs and errors. It is executed right after the synchronization is finished

rberrelleza (approved):

PR #1554, opened 2021-05-20
Inject only secrets without dots in key

jLopezbarb:

Fixes #

## Proposed changes
-  When injecting an env file it inject Okteto env vars like the tokens. All these vars have dots in its name.
-  If a secret name has a dot in it it will not be injected into the container

### DIFF
--- a/pkg/cmd/stack/translate.go
+++ b/pkg/cmd/stack/translate.go
@@ -101,7 +101,9 @@ func translateServiceEnvFile(ctx context.Context, svc *model.Service, svcName, f
 		}
 
 		for _, e := range envList {
-			secrets[e.Name] = e.Value
+			if !strings.Contains(e.Name, ".") {
+				secrets[e.Name] = e.Value
+			}
 		}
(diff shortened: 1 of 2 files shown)


pkg/cmd/stack/translate.go
```diff
@@ -101,7 +101,9 @@ func translateServiceEnvFile(ctx context.Context, svc *model.Service, svcName, f
 		}
 
 		for _, e := range envList {
-			secrets[e.Name] = e.Value
+			if !strings.Contains(e.Name, ".") {
```
pchico83, line 104: Let's implement it in `okteto.GetSecrets`

pchico83 (approved):

rberrelleza:
What is the use case that this is solving? And is it common for people to have a dot on their secrets?

pchico83:
@rberrelleza "." secret in an internal okteto secret, like the github token. "." is not allowed for user defined secrets

rberrelleza (approved):

PR #1555, opened 2021-05-20
Avoid infinite loop in EOF connections

pchico83:

Users have reported infinite loops in this code if the error is `EOF`

### DIFF
--- a/pkg/ssh/forward.go
+++ b/pkg/ssh/forward.go
@@ -19,6 +19,7 @@ import (
 	"io"
 	"net"
 	"sync"
+	"time"
 
 	"github.com/okteto/okteto/pkg/errors"
 	"github.com/okteto/okteto/pkg/log"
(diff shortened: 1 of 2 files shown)


pkg/ssh/forward.go
```diff
@@ -77,6 +79,7 @@ func (f *forward) start(ctx context.Context) {
 			}
 
 			log.Infof("%s -> failed to accept connection: %v", f.String(), err)
+			<-tick.C
```
rberrelleza, line 82: If we've only seen this on the reverse, I think we should only do it there for now.

rberrelleza (approved):
For future context: the infinite loop is by design, since we don't have a way of knowing from inside the container is there is a process to reverse forward to, we simply keep trying. This is consuming a lot of CPU, thus the ticket to wait a little bit in between retries.

pkg/ssh/forward.go
pchico83: Loops without a delay scare me

rberrelleza: Ok

PR #1556, opened 2021-05-20
Allow to create endpoints from outside stack

jLopezbarb:

## Proposed changes
-  Allows stack.endpoints be created with services that are not defined in the stack but are already running in the cluster. Returns error when:
    - Stack service and cluster service name not found
    - Cluster service found but not exporting the defined port 
    - Stack service found but it's not exporting the defined port


pkg/cmd/stack/deploy.go
```diff
@@ -350,3 +354,40 @@ func addHiddenExposedPorts(ctx context.Context, s *model.Stack) {
...
 
 }
+
+func validateEndpoints(ctx context.Context, s *model.Stack, c kubernetes.Interface) error {
+
+	for endpointName, endpoint := range s.Endpoints {
+		for _, endpointRule := range endpoint.Rules {
+			svc, _ := services.Get(ctx, endpointRule.Service, s.Namespace, c)
+			if service, ok := s.Services[endpointRule.Service]; !ok && svc == nil {
+				return fmt.Errorf("Invalid endpoint '%s': service '%s' does not exist.", endpointName, endpointRule.Service)
```
pchico83, line 364: Let's no check if it exists. In this case, both stacks are deployed in parallel and the service might not be there

pchico83 (approved):

pchico83 (approved):

PR #1558, opened 2021-05-21
Fix syncthing zombie instance in windows10

### LINKED ISSUE #1405: Syncthing zombie instances in W1O Pro 10.0.19042
**Describe the bug**
I killed the `okteto up` session in W1O Pro 10.0.19042. Now `okteto up` is failing to connect to local syncthing process. 

**To Reproduce**
Not sure about reproducible steps, but running `okteto up` in one terminal, and running `okteto up` in a second terminal, and removing the call to https://github.com/okteto/okteto/blob/master/pkg/syncthing/syncthing.go#L658 might do it.

**Expected behavior**
The second `okteto up` should clean all previous running instances of syncthing.

jLopezbarb:

Fixes #1405 

If you run `Okteto down` inside an okteto container terminal on windows, It doesn't kill children process. This PR remove all synching instances before exiting to the user terminal.

## Proposed changes
-  When the activate loop finish, remove all synching instance that are still running.

### DIFF
--- a/cmd/up/up.go
+++ b/cmd/up/up.go
@@ -308,6 +308,8 @@ func (up *upContext) activateLoop(autoDeploy, build bool) {
 			up.Exit <- err
 			return
 		}
+		up.Sy = syncthing.NewSyncthingWithHome(up.Dev)
+		up.Sy.HardTerminate()
 		up.Exit <- nil
 		return
 	}
--- a/pkg/syncthing/syncthing.go
+++ b/pkg/syncthing/syncthing.go
@@ -235,6 +235,13 @@ func New(dev *model.Dev) (*Syncthing, error) {
 	return s, nil
 }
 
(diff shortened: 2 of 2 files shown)


rberrelleza (approved):

cmd/up/up.go
```diff
@@ -308,6 +308,8 @@ func (up *upContext) activateLoop(autoDeploy, build bool) {
 			up.Exit <- err
 			return
 		}
+		up.Sy = syncthing.NewSyncthingWithHome(up.Dev)
+		up.Sy.HardTerminate()
```
pchico83, line 312: `HardTerminate` is run on `okteto up` before creating new syncthing processes.
Why wouldn't it clean zombie instances on the next `okteto up`?

pchico83:
@jLopezbarb we can close this one, right?

jLopezbarb:
Yes, this should not be merged

PR #1565, opened 2021-05-25
Terminate kills the proccess and subproccesses

jLopezbarb:

Fixes Soft terminate wasn't killing all the sub children created by the process that have to terminate

## Proposed changes
-  Adds a new logic in which before killing the parent process, the children are extracted to kill them once the parent has finished. 
-

### DIFF
--- a/pkg/syncthing/process.go
+++ b/pkg/syncthing/process.go
@@ -14,12 +14,37 @@
 package syncthing
 
 import (
+	"runtime"
 	"time"
 
 	"github.com/shirou/gopsutil/process"
 )
 
 func terminate(p *process.Process, wait bool) error {
+	if runtime.GOOS == "windows" {
+		children, err := p.Children()
+		if err != nil {
+			return err
+		}
+		err = terminateProccess(p, wait)
+		if err != nil {
+			return err
+		}
+		for _, child := range children {
+			err = terminateProccess(child, wait)
+			if err != nil {
+				return err
+			}
+		}
+		return err
+	}
+	err := terminateProccess(p, wait)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+func terminateProccess(p *process.Process, wait bool) error {
 	if err := p.Terminate(); err != nil {
 		return err
 	}
--- a/pkg/syncthing/process.go
+++ b/pkg/syncthing/process.go
@@ -21,29 +21,31 @@ import (
 )
 
 func terminate(p *process.Process, wait bool) error {
-	if runtime.GOOS == "windows" {
-		children, err := p.Children()
-		if err != nil {
-			return err
-		}
-		err = terminateProccess(p, wait)
-		if err != nil {
-			return err
-		}
-		for _, child := range children {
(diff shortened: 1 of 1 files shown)


pkg/syncthing/process.go
```diff
@@ -14,12 +14,37 @@
...
 import (
+	"runtime"
 	"time"
 
 	"github.com/shirou/gopsutil/process"
 )
 
 func terminate(p *process.Process, wait bool) error {
+	if runtime.GOOS == "windows" {
+		children, err := p.Children()
```
pchico83, line 25: how long does it take this call?

pkg/syncthing/process.go
```diff
@@ -14,12 +14,37 @@
...
+		}
+		return err
+	}
+	err := terminateProccess(p, wait)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+func terminateProccess(p *process.Process, wait bool) error {
```
pchico83, line 47: I think you still need to phases in hard terminate, or add this logic there too. If not, is there are zombie syncthing instances for what ever reason, hardTerminate wouldn't kill all of them

pkg/syncthing/process.go
```diff
@@ -14,12 +14,37 @@
...
 
 import (
+	"runtime"
 	"time"
 
 	"github.com/shirou/gopsutil/process"
 )
 
 func terminate(p *process.Process, wait bool) error {
+	if runtime.GOOS == "windows" {
```
pchico83, line 24: I think the logic is cleaner if there is a function to get the children, then kill the parent, then kill the children.
If `runtime.GOOS != "windows"`, children is `[]`

pchico83 (approved):

pkg/syncthing/process.go
jLopezbarb: I have run the code for 5 times and the average time to call for the children was 10.2ms (been 3.8 the fastest and 14.9 the slowest)

pkg/syncthing/process.go
jLopezbarb: This code it's also affecting to hard terminate. It's the code that you call [here](https://github.com/okteto/okteto/blob/96eed3f63d5be1b8fab99cc2f21efe2c5fd753fd/pkg/syncthing/syncthing.go#L688)

pkg/syncthing/process.go
jLopezbarb: If we call it on linux children won't be `[]`, it will be the same as it's on windows and you will have one more call to make (the call to get children). I also thought to be OS agnostic but having one more process call is not very suitable

pkg/syncthing/process.go
pchico83: Ah, perfect!

pkg/syncthing/process.go
pchico83: What I mean is to artificially return an empty list if it is not windows. We can call it `getZombieChildren` to make it clearer

PR #1570, opened 2021-05-26
Add support for restart field

### LINKED ISSUE #1056: Support for "jobs" on Okteto Stacks
When I deploy my stack, some of my services need data initialization. This is usually implemented in Kubernetes using jobs.

A friendly way to define Jobs using a `jobs` section.

jLopezbarb:

Fixes #1056 

## Proposed changes
-  Add support for restart field. 
-  With this change you can create jobs in docker-compose

### DIFF
--- a/pkg/cmd/stack/deploy.go
+++ b/pkg/cmd/stack/deploy.go
@@ -26,6 +26,7 @@ import (
 	"github.com/okteto/okteto/pkg/k8s/configmaps"
 	"github.com/okteto/okteto/pkg/k8s/deployments"
 	"github.com/okteto/okteto/pkg/k8s/ingresses"
+	"github.com/okteto/okteto/pkg/k8s/jobs"
 	okLabels "github.com/okteto/okteto/pkg/k8s/labels"
 	"github.com/okteto/okteto/pkg/k8s/pods"
 	"github.com/okteto/okteto/pkg/k8s/services"
@@ -107,7 +108,11 @@ func deploy(ctx context.Context, s *model.Stack, wait bool, c *kubernetes.Client
 	}
 
 	for name := range s.Services {
-		if len(s.Services[name].Volumes) == 0 {
+		if s.Services[name].RestartPolicy != apiv1.RestartPolicyAlways {
+			if err := deployJob(ctx, name, s, c); err != nil {
+				return err
+			}
+		} else if len(s.Services[name].Volumes) == 0 {
 			if err := deployDeployment(ctx, name, s, c); err != nil {
 				return err
 			}
@@ -144,7 +149,6 @@ func deploy(ctx context.Context, s *model.Stack, wait bool, c *kubernetes.Client
 
 	spinner.Update("Waiting for services to be ready...")
 	return waitForPodsToBeRunning(ctx, s, c)
-
 }
 
 func deployDeployment(ctx context.Context, svcName string, s *model.Stack, c *kubernetes.Clientset) error {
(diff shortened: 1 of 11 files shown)


pkg/cmd/stack/translate.go
```diff
@@ -388,6 +389,54 @@ func translateStatefulSet(svcName string, s *model.Stack) *appsv1.StatefulSet {
...
+
+	return &batchv1.Job{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:        svcName,
+			Namespace:   s.Namespace,
+			Labels:      translateLabels(svcName, s),
+			Annotations: translateAnnotations(svc),
+		},
+		Spec: batchv1.JobSpec{
+			Parallelism: pointer.Int32Ptr(svc.Replicas),
```
pchico83, line 405: I think `Replicas` should translate to `Completions`

pkg/cmd/stack/translate.go
```diff
@@ -388,6 +389,54 @@ func translateStatefulSet(svcName string, s *model.Stack) *appsv1.StatefulSet {
...
+			Selector: &metav1.LabelSelector{
+				MatchLabels: translateLabelSelector(svcName, s),
+			},
+			ManualSelector: pointer.BoolPtr(true),
```
pchico83, line 409: should we leave this empty?

pkg/model/stack_serializer.go
```diff
@@ -138,20 +138,27 @@ type ServiceRaw struct {
...
-	RestartPolicy  *WarningType `yaml:"restart_policy,omitempty"`
 	RollbackConfig *WarningType `yaml:"rollback_config,omitempty"`
 	UpdateConfig   *WarningType `yaml:"update_config,omitempty"`
 }
 
+type RestartPolicyRaw struct {
+	Condition string `yaml:"condition,omitempty"`
+
+	Delay       *WarningType `yaml:"delay,omitempty"`
+	MaxAttempts *WarningType `yaml:"max_attempts,omitempty"`
```
pchico83, line 159: This could translate to `BackoffLimit`

pkg/cmd/stack/translate.go
pchico83: `Parallelism` should be always 1

pkg/cmd/stack/translate.go
```diff
@@ -388,6 +389,54 @@ func translateStatefulSet(svcName string, s *model.Stack) *appsv1.StatefulSet {
...
+			Template: apiv1.PodTemplateSpec{
+				ObjectMeta: metav1.ObjectMeta{
+					Labels:      translateLabels(svcName, s),
+					Annotations: translateAnnotations(svc),
+				},
+				Spec: apiv1.PodSpec{
+					RestartPolicy:                 svc.RestartPolicy,
```
pchico83, line 416: This should be `Never`.  The job will create another pod if this one fails

pchico83:
@jLopezbarb could we create an e2e test for this?

pkg/cmd/stack/translate.go
jLopezbarb: If you leave `ManualSelector` empty you can not add any selector, and when listing the jobs we can not see if it's from another stack

pchico83: From the docs:
```
	// manualSelector controls generation of pod labels and pod selectors.
	// Leave `manualSelector` unset unless you are certain what you are doing.
	// When false or unset, the system pick labels unique to this job
	// and appends those labels to the pod template.  When true,
	// the user is responsible for picking unique labels and specifying
	// the selector.  Failure to pick a unique label may cause this
	// and other jobs to not function correctly.  However, You may see
	// `manualSelector=true` in jobs that were created with the old `extensions/v1beta1`
	// API.
	// More info: https://kubernetes.io/docs/concepts/workloads/controllers/jobs-run-to-completion/#specifying-your-own-pod-selector
	// +optional
	ManualSelector *bool `json:"manualSelector,omitempty" protobuf:"varint,5,opt,name=manualSelector"`
```
We don't need control on the pod labels, no? we only need control on the job labels to do queries, right?

jLopezbarb:
I updated the e2e test for compose adding an initialization for the queue

pchico83:
> I updated the e2e test for compose adding an initialization for the queue

is the e2e test ready to test the job is created/and deleted? would the test fail if the queue is not properly initialized?

pchico83 (approved):
I think we should update the compose test to check that jobs are deleted:
https://github.com/okteto/okteto/blob/master/integration/stacks_test.go#L237

pkg/cmd/stack/deploy.go
```diff
@@ -296,7 +328,7 @@ func waitForPodsToBeRunning(ctx context.Context, s *model.Stack, c *kubernetes.C
 			return err
 		}
 		for i := range podList {
-			if podList[i].Status.Phase == apiv1.PodRunning {
+			if podList[i].Status.Phase == apiv1.PodRunning || podList[i].Status.Phase == apiv1.PodSucceeded {
```
pchico83, line 331: If a pod is in `PodFailed` we should fail and return an error.

pchico83 (approved):

PR #1578, opened 2021-05-28
Adds support for depends on field

### LINKED ISSUE #1386: Supports for "depends_on" in stacks
I want to define dependencies between my services in Okteto Stacks. It a dependency is to a `service`, stacks should wait for the ports of the service to be available. We can do that with a similar approach to `wait-for it.sh` https://github.com/vishnubob/wait-for-it/blob/master/wait-for-it.sh. If the dependency is in a job, stacks should wait for the job to finish. 

Define dependencies between services/jobs using the `depends_on` field, which should have an optional timeout parameter.

**Additional context**
One option is to implement the wait loop as an init container.

jLopezbarb:
Fixes #1386 

## Proposed changes
-  Add support for `depends_on` field in stack and compose.
-  Checks if it has a cyclic dependency before deploying all the stack
-  Waits for the dependent service to be running, available or if its a job checks if it has finished.


pkg/cmd/stack/translate.go
```diff
@@ -459,6 +477,54 @@ func getAddPermissionsInitContainer(svcName string, svc *model.Service) apiv1.Co
 	return initContainer
 }
 
+func getWaitForSvcsInitContainer(svcName string, s *model.Stack) *apiv1.Container {
+	command := "echo waiting for dependent services..."
+	for dependentSvc, condition := range s.Services[svcName].DependsOn {
+		if condition.Condition == model.DependsOnServiceHealthy {
```
pchico83, line 483: Should this be part of the stack serialization?

pkg/cmd/stack/translate.go
```diff
@@ -459,6 +477,54 @@ func getAddPermissionsInitContainer(svcName string, svc *model.Service) apiv1.Co
...
+		}
+		return initContainer
+	}
+	return nil
+}
+
+func getDependentCommand(condition model.DependsOnConditionSpec, dependentSvcName string, s *model.Stack) string {
+	switch condition.Condition {
+	case model.DependsOnServiceCompleted:
+		return fmt.Sprintf(" && kubectl wait --for=condition=complete job/%s", dependentSvcName)
```
pchico83, line 505: If it fails, will we wait forever?

pkg/cmd/stack/translate.go
```diff
@@ -459,6 +477,54 @@ func getAddPermissionsInitContainer(svcName string, svc *model.Service) apiv1.Co
...
+	case model.DependsOnServiceRunning:
+		if isDeployment(s, dependentSvcName) {
+			return fmt.Sprintf(" && kubectl wait --for=condition=available deployment/%s", dependentSvcName)
+		} else {
+			// Workaround until https://github.com/kubernetes/kubernetes/issues/79606
+			return fmt.Sprintf("kubectl rollout status --watch --timeout=600s statefulset/%s", dependentSvcName)
```
pchico83, line 511: What is this for?

pkg/cmd/stack/translate.go
```diff
@@ -459,6 +477,54 @@ func getAddPermissionsInitContainer(svcName string, svc *model.Service) apiv1.Co
...
+				condition.Condition = model.DependsOnServiceRunning
+			}
+		}
+		command += getDependentCommand(condition, dependentSvc, s)
+	}
+
+	if command != "echo waiting for dependent services..." {
+		initContainer := &apiv1.Container{
+			Name:    "wait-for-svcs",
+			Image:   "bitnami/kubectl",
```
pchico83, line 494: I think it is better if we build our own image with a small go program that does this logic.

pchico83 (approved):

pkg/cmd/stack/wait.go
```diff
@@ -0,0 +1,191 @@
...
+		fmt.Printf("Service '%s' is ready\n", dependentSvc)
+	}
+	return nil
+}
+
+func waitForSvcRunning(ctx context.Context, stack *model.Stack, svcName string, c kubernetes.Interface) error {
+	svcToWaitFor := stack.Services[svcName]
+
+	ticker := time.NewTicker(100 * time.Millisecond)
+	timeout := time.Now().Add(300 * time.Second)
```
pchico83, line 127: One option is to print a log a fail. Kubernetes will retry the init container with the right backlogs

pkg/cmd/stack/wait.go
```diff
@@ -0,0 +1,191 @@
...
+
+	for time.Now().Before(timeout) {
+		<-ticker.C
+		for _, p := range svcToWaitFor.Ports {
+			url := fmt.Sprintf("http://%s:%d/", svcName, p.ContainerPort)
+			resp, err := http.Get(url)
```
pchico83, line 168: is this how `wait-for` is implemented? what if it is not an HTTP port?

pkg/k8s/statefulsets/crud.go
```diff
@@ -79,3 +79,11 @@ func Destroy(ctx context.Context, name, namespace string, c kubernetes.Interface
 	log.Infof("statefulset '%s' deleted", name)
 	return nil
 }
+
+func IsRunning(ctx context.Context, namespace, svcName string, c kubernetes.Interface) bool {
+	d, err := c.AppsV1().StatefulSets(namespace).Get(ctx, svcName, metav1.GetOptions{})
```
pchico83, line 84:
```suggestion
	sts, err := c.AppsV1().StatefulSets(namespace).Get(ctx, svcName, metav1.GetOptions{})
```

pkg/model/utils.go
```diff
@@ -121,3 +121,39 @@ func GetRepositoryURL(path string) (string, error) {
 
 	return remotes[0].Config().URLs[0], nil
 }
+
+func getDependentCyclic(s *Stack) []string {
```
pchico83, line 125: what is this for?

pchico83 (commented):
This is looking great!

pkg/model/utils.go
jLopezbarb: This checks if the stack has any cycle in its dependencies, like 
```
services:
  app:
    depends_on:
      - db
  db:
    depends_on:
      - job
  job:
    depends_on:
      - app
```

pkg/cmd/stack/wait.go
jLopezbarb: I can make a callback to https and ftp, but I think it could be very useful if we support docker-compose `healthcheck` field which is the command that must be ran to know if the  service is healthy instead of checking all the ports with all the protocols.

pkg/cmd/stack/deploy.go
```diff
@@ -80,49 +83,34 @@ func Deploy(ctx context.Context, s *model.Stack, forceBuild, wait, noCache bool)
...
-
-	for name := range s.Volumes {
-		if err := deployVolume(ctx, name, s, c); err != nil {
-			return err
-		}
-		spinner.Stop()
-		log.Success("Created volume '%s'", name)
-		spinner.Start()
-	}
+	deployedSvcs := make([]string, 0)
```
pchico83, line 94: would be better to use a `map`?

pkg/cmd/stack/deploy.go
```diff
@@ -80,49 +83,34 @@ func Deploy(ctx context.Context, s *model.Stack, forceBuild, wait, noCache bool)
...
+	deployedVolumes := make([]string, 0)
```
pchico83, line 95: why do you need to track `deployedVolumes`? We could deploy all volumes before entering the services loop, no?

pkg/cmd/stack/deploy.go
```diff
@@ -150,13 +138,156 @@ func deploy(ctx context.Context, s *model.Stack, wait bool, c *kubernetes.Client
...
+func isVolumeDeployed(volumeName string, deployedVolumes []string) bool {
+	for _, deployedVolumeName := range deployedVolumes {
+		if deployedVolumeName == volumeName {
+			return true
+		}
+	}
+	return false
+}
+
+func isSvcDeployed(svcName string, deployedSvcs []string) bool {
```
pchico83, line 187: This is cleaner with a map

pkg/cmd/stack/deploy.go
```diff
@@ -150,13 +138,156 @@ func deploy(ctx context.Context, s *model.Stack, wait bool, c *kubernetes.Client
...
+				resp, err := http.Get(url)
+				if err != nil {
+					url := fmt.Sprintf("https://%s:%d/", model.Localhost, port)
+					resp, err = http.Get(url)
+					if err != nil {
+						continue
+					}
+				}
+				if resp.StatusCode >= 200 && resp.StatusCode <= 299 {
+					forwarder.Stop()
```
pchico83, line 241: Do `defer forwarder.Stop()` after:
https://github.com/okteto/okteto/pull/1578/files#diff-6f7aa5096f3b87ea04156075f6cd70f48809defab4d8fb2a76206ca9328822d8R229

pkg/cmd/stack/deploy.go
```diff
@@ -150,13 +138,156 @@ func deploy(ctx context.Context, s *model.Stack, wait bool, c *kubernetes.Client
...
+	return true
+}
+
+func isSvcReady(ctx context.Context, stack *model.Stack, dependentSvcName string, condition model.DependsOnConditionSpec, client kubernetes.Interface, config *rest.Config) bool {
+	svc := stack.Services[dependentSvcName]
+
+	switch condition.Condition {
+	case model.DependsOnServiceRunning:
+		return isSvcRunning(ctx, svc, stack.Namespace, dependentSvcName, client)
+	case model.DependsOnServiceHealthy:
```
pchico83, line 211: I would split this logic in different functions

pkg/cmd/stack/deploy.go
jLopezbarb: Yes, it will be faster to check if they have already been deployed

pkg/cmd/stack/deploy.go
```diff
@@ -106,23 +109,30 @@ func deploy(ctx context.Context, s *model.Stack, wait bool, c *kubernetes.Client
...
+				continue
 			}
-		} else if len(s.Services[name].Volumes) == 0 {
-			if err := deployDeployment(ctx, name, s, c); err != nil {
-				return err
+
+			if !canSvcBeDeployed(ctx, s, svcName, c, config) {
+				if failedJobs := getDependingFailedJobs(ctx, s, svcName, c, config); len(failedJobs) > 0 {
+					if len(failedJobs) == 1 {
+						return fmt.Errorf("Can not start %s: %s has failed", svcName, failedJobs[0])
```
pchico83, line 122:
```suggestion
						return fmt.Errorf("Service '%s' dependency '%s' failed", svcName, failedJobs[0])
```

pkg/cmd/stack/deploy.go
```diff
@@ -106,23 +109,30 @@ func deploy(ctx context.Context, s *model.Stack, wait bool, c *kubernetes.Client
...
+					}
+					return fmt.Errorf("Can not start %s: %s have failed", svcName, strings.Join(failedJobs, ", "))
```
pchico83, line 124:
```suggestion
					return fmt.Errorf("Service '%s' dependencies '%s' failed", svcName, strings.Join(failedJobs, ", "))
```

pkg/cmd/stack/deploy.go
```diff
@@ -150,13 +160,156 @@ func deploy(ctx context.Context, s *model.Stack, wait bool, c *kubernetes.Client
...
+			return err
+		}
+	}
+	spinner.Stop()
+	log.Success("Deployed service '%s'", svcName)
+	spinner.Start()
+	return nil
+}
+
+func isVolumeDeployed(volumeName string, deployedVolumes []string) bool {
```
pchico83, line 183: remove?

pkg/cmd/stack/deploy.go
```diff
@@ -150,13 +160,156 @@ func deploy(ctx context.Context, s *model.Stack, wait bool, c *kubernetes.Client
...
+	for _, deployedVolumeName := range deployedVolumes {
+		if deployedVolumeName == volumeName {
+			return true
+		}
+	}
+	return false
+}
+
+func isSvcDeployed(svcName string, deployedSvcs map[string]bool) bool {
```
pchico83, line 192: I don't think we need a function for this

pkg/cmd/stack/deploy.go
```diff
@@ -150,13 +160,156 @@ func deploy(ctx context.Context, s *model.Stack, wait bool, c *kubernetes.Client
...
+	case svc.IsJob():
+		if jobs.IsRunning(ctx, namespace, svcName, client) {
+			return true
+		}
+	}
+	return false
+}
+
+func isSvcHealthy(ctx context.Context, svc *model.Service, stack *model.Stack, svcName string, client kubernetes.Interface, config *rest.Config) bool {
+	if isSvcRunning(ctx, svc, stack.Namespace, svcName, client) {
```
pchico83, line 268:
```suggestion
	if !isSvcRunning(ctx, svc, stack.Namespace, svcName, client) {
	    return false
	}
```
This way the overal indentation is reduced

PR #1603, opened 2021-06-07
Fix empty endpoint name

jLopezbarb:

Fixes a bug when adding non named endpoints on a stack file

## Proposed changes
-  Set the name of the endpoint after setting the name of the stack
-

### DIFF
--- a/pkg/model/stack.go
+++ b/pkg/model/stack.go
@@ -302,6 +302,13 @@ func (s *Stack) validate() error {
 		svc.IgnoreSyncVolumes(s)
 	}
 
+	for name, endpoint := range s.Endpoints {
+		if name == "" {
+			s.Endpoints[s.Name] = endpoint
+			delete(s.Endpoints, "")
+		}
+	}
+
 	return nil
 }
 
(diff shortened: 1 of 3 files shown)


pkg/model/stack_serializer.go
```diff
@@ -208,8 +208,8 @@ func (s *Stack) UnmarshalYAML(unmarshal func(interface{}) error) error {
 
 	s.Endpoints = stackRaw.Endpoints
 	if endpoint, ok := s.Endpoints[""]; ok {
-		s.Endpoints[s.Name] = endpoint
```
pchico83, line 211: Let's move all this after we have the final stack name

pkg/model/stack.go
```diff
@@ -302,6 +302,13 @@ func (s *Stack) validate() error {
 		svc.IgnoreSyncVolumes(s)
 	}
 
+	for name, endpoint := range s.Endpoints {
```
pchico83, line 305: no need for a loop, you can just query `s.Endpoints[""]`

pchico83 (approved):

pkg/model/stack.go
pchico83: also, can we move this code here?
https://github.com/okteto/okteto/blob/master/pkg/model/stack.go#L174

we could move all that logic to anew function that computes the stack name, but this is not part of the validation

PR #1610, opened 2021-06-11
Show endpoint in okteto up output

### LINKED ISSUE #1607: Show endpoint in okteto up output
We currently support this for divert, we should add the endpoints as part of the output of every `okteto up`:

Screenshot 2021-06-10 at 10 31 46

We need however to show the full URL with the protocol so most terminals can parse the output, recognize the URL and make it clickable.

jLopezbarb:

Fixes #1607

## Proposed changes
-  Search for services that are linked to the deployment that is on dev mode and show the urls that are on it
-


pchico83:
@jLopezbarb let's close this one for now, we can get this data from the Okteto API, but we need to wait for a PR to get merged

PR #1611, opened 2021-06-11
Update integration tests timeout to 15 min

jLopezbarb:

Fixes Integration tests

## Proposed changes
-  Update integration tests timeout to 15 min instead of 10min
-

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,7 +40,7 @@ test:
 
 .PHONY: integration
 integration:
-	 go test github.com/okteto/okteto/integration -tags=integration --count=1 -v
+	 go test github.com/okteto/okteto/integration -tags=integration --count=1 -v -timeout 15m
 
 .PHONY: build
 build:


Makefile
```diff
@@ -40,7 +40,7 @@ test:
 
 .PHONY: integration
 integration:
-	 go test github.com/okteto/okteto/integration -tags=integration --count=1 -v
+	 go test github.com/okteto/okteto/integration -tags=integration --count=1 -v -timeout 15m
```
rberrelleza, line 43: Maybe we should remove it from here, and instead have circleci time it out?

jLopezbarb: I think its better to set it here because you can use `make integration` in your local environment without worrying if the tests take more than 10 minute. 
We have to set it also on CircleCI because we are having issues on windows tests where the test is not showing any output for 10 minutes and can be solve by setting the CircleCI variable `no_output_timeout`

pchico83: @jLopezbarb could you coordinate with @irespaldiza for the circleci changes?

pchico83 (approved):

PR #1617, opened 2021-06-16
Fix first character lost when reconnect on ssh

### LINKED ISSUE #1455: First input char is lost after "okteto up" reconnects
**Describe the bug**
I am running `okteto up` and there is a reconnection even. For example, the pod is deleted. When my terminal is reconnected, the first input char is lost.

**Expected behavior**
No input char is lost

**Additional context**
We might need to inject a char in stdout when reconnecting.

jLopezbarb:

Fixes #1455

## Proposed changes
-  Checks if there is another process trying to copy from stdin to remote server and wait for it to finish to start copying.
-  Refactor copy to get the character lost
-  Copy that character into a channel and inject it into remote stdin

### DIFF
--- a/pkg/ssh/exec.go
+++ b/pkg/ssh/exec.go
@@ -31,11 +31,11 @@ import (
 	"golang.org/x/crypto/ssh/terminal"
 )
 
-var lostCharacter = make(chan []byte, 1)
-var isAnotherCopyRunning = make(chan bool, 1)
+var lostCharacter = make(chan []byte, 10)
+var iterInLoop = make(chan int, 10)
 
 // Exec executes the command over SSH
-func Exec(ctx context.Context, iface string, remotePort int, tty bool, inR io.Reader, outW, errW io.Writer, command []string) error {
+func Exec(ctx context.Context, iface string, remotePort int, tty bool, inR io.Reader, outW, errW io.Writer, command []string, retry int) error {
 	sshConfig, err := getSSHClientConfig()
 	if err != nil {
 		return fmt.Errorf("failed to get SSH configuration: %s", err)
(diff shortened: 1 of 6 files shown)


pkg/ssh/exec.go
```diff
@@ -32,6 +31,9 @@ import (
 	"golang.org/x/crypto/ssh/terminal"
 )
 
+var lostCharacter = make(chan []byte, 1)
```
pchico83, line 34: could we have a single channel?
if there are several reconnections, there could be several lost chars

pkg/ssh/exec.go
```diff
@@ -184,6 +187,60 @@ More information is available here: https://okteto.com/docs/reference/manifest#r
 	return err
 }
 
+func waitForOtherCopyToFinish() {
+	wasRunning := false
+Loop:
```
pchico83, line 192: is it possible to avoid this syntax and use a structured loop instead?

pkg/ssh/exec.go
```diff
@@ -184,6 +187,60 @@ More information is available here: https://okteto.com/docs/reference/manifest#r
...
+	case char := <-lostCharacter:
+		remoteStdin.Write(char)
+	default:
+		break
+	}
+}
+
+func copyFromLocalToRemote(local io.Reader, remote io.WriteCloser, buf []byte) {
+	for {
+		nr, er := local.Read(buf)
```
pchico83, line 220: could we do a `select`here between `Read` and the `lostCharacter` channel?

jLopezbarb: no, I tried that but it will get blocked anyways on the read until you press a key

pkg/ssh/exec.go
jLopezbarb: yes, we could get out of the loop with a return statement and not use the tag

pkg/ssh/exec.go
jLopezbarb: I will try these user cases and try to have just one channel

PR #1618, opened 2021-06-16
Trim long messages while spinning

### LINKED ISSUE #1596: If the image name is longer than the terminal, the image name is printed over and over again
**Describe the bug**
see https://www.loom.com/share/4d5b6068ff8f4d97b2430e7aba94aaec

**Expected behavior**
I think we should either clip the image name or handle the scenario when the text is longer than the terminal width.

jLopezbarb:

Fixes #1596

spinner

## Proposed changes
-  Use spinner final message as a variable to trim or expand the suffix, taking in count the terminal width
-  It works well if you don't resize the terminal in the process of running the command
-  It also works if you make the terminal larger
- When making the terminal smaller it will create the same behavior as its today because it doesn't have time to trim before adding the new line.

(PD. This PR needs to be test on windows)

### DIFF
--- a/cmd/utils/spinner.go
+++ b/cmd/utils/spinner.go
@@ -23,6 +23,7 @@ import (
 
 	sp "github.com/briandowns/spinner"
 	"github.com/okteto/okteto/pkg/log"
+	"golang.org/x/crypto/ssh/terminal"
 )
 
 var spinnerSupport bool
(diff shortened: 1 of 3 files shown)


pchico83 (approved):

jLopezbarb:
It's not working on windows, it seems that [this method](https://github.com/okteto/okteto/pull/1618/files#diff-2eec6f917c46483a46a7727ac46d550dfb4200a8276164eb8211d596c64c744bR44) is only working for linux.

PR #1623, opened 2021-06-22
Add update command

### LINKED ISSUE #1608: Okteto update
When there's a new update for the Okteto CLI, a user has no way to be notified except the user is on the slack channel or visits our media outlets frequently.

I will like that an `update` option is added so users can run `okteto update` to check for new version of the CLIs and update their tools.

None.

**Additional context**
It'll be great to notify users of a new CLI version from the okteto CLI itself. An example is when a user runs a command using an old version, the user is notified that the cli is an outdated one.

jLopezbarb:

Fixes #1608

## Proposed changes
- Adds a new command where the users can update their okteto binary with just typing okteto update
- It's not compatible with windows
- Updates progress bar library to the latest version

### DIFF
--- a/cmd/update.go
+++ b/cmd/update.go
@@ -15,20 +15,13 @@ package cmd
 
 import (
 	"fmt"
-	"io"
-	"io/ioutil"
-	"net/http"
-	"os"
 	"runtime"
-	"time"
 
 	"github.com/Masterminds/semver/v3"
 	"github.com/okteto/okteto/cmd/utils"
 	"github.com/okteto/okteto/pkg/config"
 	"github.com/okteto/okteto/pkg/log"
 	"github.com/spf13/cobra"
-	"github.com/vbauerster/mpb/v7"
-	"github.com/vbauerster/mpb/v7/decor"
 )
 
 const (
@@ -43,11 +36,7 @@ func Update() *cobra.Command {
 		Short: "Updates okteto version",
 		RunE: func(cmd *cobra.Command, args []string) error {
 			if isUpdateAvailable() {
-				err := updateVersion()
-				if err != nil {
-					return err
-				}
-				log.Success("The latest okteto version has been installed")
+				displayUpdateSteps()
 			} else {
 				log.Success("The latest okteto version is already installed")
 			}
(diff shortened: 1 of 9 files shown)


pchico83 (commented):
This is problematic. We would need to handle different OS, like windows. And the user might want to use brew, scoop. I think it is better to leave the upgrade responsability to the user

rberrelleza:
> This is problematic. We would need to handle different OS, like windows. And the user might want to use brew, scoop.
> I think it is better to leave the upgrade responsibility to the user.
> Do you know what other well-known CLIs do for this?

arkade has  a command that prints out the update instructions
image

pchico83:
> > This is problematic. We would need to handle different OS, like windows. And the user might want to use brew, scoop.
> > I think it is better to leave the upgrade responsibility to the user.
> > Do you know what other well-known CLIs do for this?
> 
> arkade has a command that prints out the update instructions
> image

I like it!

cmd/update.go
```diff
@@ -0,0 +1,95 @@
...
+}
+
+func displayUpdateSteps() {
+	fmt.Println("You can update okteto with the following:")
+	switch {
+	case runtime.GOOS == "darwin" || runtime.GOOS == "linux":
+		fmt.Print(`# Using installation script:
+curl https://get.okteto.com -sSfL | sh
+
+# Using brew:
```
pchico83, line 85: this is only for `darwin`, no?

pchico83 (approved):

cmd/update.go
jLopezbarb: You can install Linuxbrew which is the same as Homebrew but for Linux. It can also be installed on Windows WSL which is pretty useful but it's not very common to see it on Windows.

pchico83: Yes, but it is not usually documented for Linux. For example:
https://kubernetes.io/docs/tasks/tools/install-kubectl-linux/#install-kubectl-on-linux
https://github.com/cli/cli#installation

rberrelleza: If it works, I'd leave it, it's a nice option.

cmd/update.go
```diff
@@ -0,0 +1,95 @@
+// Copyright 2020 The Okteto Authors
```
rberrelleza, line 1:
```suggestion
// Copyright 2021 The Okteto Authors
```

rberrelleza (approved):

cmd/update.go
jLopezbarb: Should we change all the copyright from 2020 to 2021?

rberrelleza: yeah, but it's probably better to do it on a separate PR?

jLopezbarb: I think its going to be better to have a single PR for that specific task

PR #1627, opened 2021-06-23
Jlopezbarb/developer level file

### LINKED ISSUE #1557: Developer-level overrides to okteto.yaml
From an okteto user:

> I think an .oktetorc with the ability to sync extra files would be really useful for thinks like developer-specific dotfiles. Maybe just merge it with the current okteto.yml so you can specify environment variables like PS1. There are things that I want to sync to all the projects I work on (e.g. .vimrc, but having a project specific override could be nice for things like different commands - perhaps I want to pass different flags to the application or load different data)

One option I thought was that we could have a `$HOME/.okteto.yaml` file; whatever is defined gets merged with the local `okteto.yml`, with $HOME/.okteto.yaml taking precedence. This is similar to how things like `.bashrc.  work.

jLopezbarb:
Fixes #1557

## Proposed changes
- Adds a file at developer level to inject common behavior to `okteto up` commands
-

### DIFF
--- a/cmd/utils/dev.go
+++ b/cmd/utils/dev.go
@@ -67,12 +67,12 @@ func loadDevRc(dev *model.Dev) {
 	if model.FileExists(defaultDevRcPath) {
 		devRc, err = model.GetRc(defaultDevRcPath)
 		if err != nil {
-			log.Infof("error while reading .dev file: %s", err.Error())
+			log.Infof("error while reading %s file: %s", defaultDevRcPath, err.Error())
 		}
 	} else if model.FileExists(secondaryDevRcPath) {
 		devRc, err = model.GetRc(secondaryDevRcPath)
 		if err != nil {
-			log.Infof("error while reading .dev file: %s", err.Error())
+			log.Infof("error while reading %s file: %s", secondaryDevRcPath, err.Error())
 		}
 	}
 
--- a/cmd/utils/dev.go
+++ b/cmd/utils/dev.go
@@ -60,8 +60,8 @@ func LoadDev(devPath, namespace, k8sContext string) (*model.Dev, error) {
 }
 
 func loadDevRc(dev *model.Dev) {
-	defaultDevRcPath := filepath.Join(config.GetOktetoHome(), ".okteto.yml")
-	secondaryDevRcPath := filepath.Join(config.GetOktetoHome(), ".okteto.yaml")
+	defaultDevRcPath := filepath.Join(config.GetOktetoHome(), "okteto.yml")
+	secondaryDevRcPath := filepath.Join(config.GetOktetoHome(), "okteto.yaml")
 	var devRc *model.DevRC
 	var err error
 	if model.FileExists(defaultDevRcPath) {
--- a/cmd/utils/dev.go
+++ b/cmd/utils/dev.go
@@ -74,8 +74,6 @@ func loadDevRc(dev *model.Dev) {
 		if err != nil {
 			log.Infof("error while reading .dev file: %s", err.Error())
 		}
-	} else {
-		log.Info("developer level dev file not found")
 	}
 
 	if devRc != nil {
--- a/cmd/utils/dev.go
+++ b/cmd/utils/dev.go
@@ -70,7 +70,7 @@ func loadDevRc(dev *model.Dev) {
 			log.Infof("error while reading .dev file: %s", err.Error())
 		}
 	} else if model.FileExists(secondaryDevRcPath) {
-		devRc, err = model.GetRc(defaultDevRcPath)
+		devRc, err = model.GetRc(secondaryDevRcPath)
 		if err != nil {
 			log.Infof("error while reading .dev file: %s", err.Error())
 		}
--- a/cmd/utils/dev.go
+++ b/cmd/utils/dev.go
@@ -53,13 +53,15 @@ func LoadDev(devPath, namespace, k8sContext string) (*model.Dev, error) {
 		return nil, err
 	}
 
-	loadDevRc(dev)
+	if err := loadDevRc(dev); err != nil {
+		return nil, err
+	}
 	loadContext(dev, k8sContext)
 	loadNamespace(dev, namespace)
 	return dev, nil
 }
 
-func loadDevRc(dev *model.Dev) {
+func loadDevRc(dev *model.Dev) error {
 	defaultDevRcPath := filepath.Join(config.GetOktetoHome(), "okteto.yml")
 	secondaryDevRcPath := filepath.Join(config.GetOktetoHome(), "okteto.yaml")
 	var devRc *model.DevRC
@@ -67,18 +69,19 @@ func loadDevRc(dev *model.Dev) {
 	if model.FileExists(defaultDevRcPath) {
 		devRc, err = model.GetRc(defaultDevRcPath)
 		if err != nil {
-			log.Infof("error while reading %s file: %s", defaultDevRcPath, err.Error())
+			return fmt.Errorf("error while reading %s file: %s", defaultDevRcPath, err.Error())
 		}
 	} else if model.FileExists(secondaryDevRcPath) {
 		devRc, err = model.GetRc(secondaryDevRcPath)
 		if err != nil {
-			log.Infof("error while reading %s file: %s", secondaryDevRcPath, err.Error())
+			return fmt.Errorf("error while reading %s file: %s", defaultDevRcPath, err.Error())
 		}
 	}
 
 	if devRc != nil {
 		model.MergeDevWithDevRc(dev, devRc)
 	}
+	return nil
 }
 
 func loadContext(dev *model.Dev, k8sContext string) {
--- a/cmd/utils/dev.go
+++ b/cmd/utils/dev.go
@@ -18,9 +18,11 @@ import (
 	"fmt"
 	"net/url"
 	"os"
+	"path/filepath"
 	"strings"
 
 	"github.com/joho/godotenv"
+	"github.com/okteto/okteto/pkg/config"
 	"github.com/okteto/okteto/pkg/errors"
 	"github.com/okteto/okteto/pkg/k8s/client"
 	"github.com/okteto/okteto/pkg/log"
@@ -50,9 +52,35 @@ func LoadDev(devPath, namespace, k8sContext string) (*model.Dev, error) {
 	if err != nil {
 		return nil, err
 	}
+
+	loadDevRc(dev)
 	loadContext(dev, k8sContext)
 	loadNamespace(dev, namespace)
 	return dev, nil
+}
+
+func loadDevRc(dev *model.Dev) {
+	defaultDevRcPath := filepath.Join(config.GetOktetoHome(), ".okteto.yml")
+	secondaryDevRcPath := filepath.Join(config.GetOktetoHome(), ".okteto.yaml")
+	var devRc *model.DevRC
+	var err error
+	if model.FileExists(defaultDevRcPath) {
+		devRc, err = model.GetRc(defaultDevRcPath)
+		if err != nil {
+			log.Infof("error while reading .dev file: %s", err.Error())
+		}
+	} else if model.FileExists(secondaryDevRcPath) {
+		devRc, err = model.GetRc(defaultDevRcPath)
+		if err != nil {
+			log.Infof("error while reading .dev file: %s", err.Error())
+		}
+	} else {
+		log.Info("developer level dev file not found")
+	}
+
+	if devRc != nil {
+		model.MergeDevWithDevRc(dev, devRc)
+	}
 }
 
 func loadContext(dev *model.Dev, k8sContext string) {
--- a/pkg/model/devrc.go
+++ b/pkg/model/devrc.go
@@ -14,22 +14,15 @@ import (
 
 // Dev represents a development container
 type DevRC struct {
-	Autocreate           *bool                 `json:"autocreate,omitempty" yaml:"autocreate,omitempty"`
 	Labels               Labels                `json:"labels,omitempty" yaml:"labels,omitempty"`
 	Annotations          Annotations           `json:"annotations,omitempty" yaml:"annotations,omitempty"`
 	Context              string                `json:"context,omitempty" yaml:"context,omitempty"`
 	Namespace            string                `json:"namespace,omitempty" yaml:"namespace,omitempty"`
-	ImagePullPolicy      apiv1.PullPolicy      `json:"imagePullPolicy,omitempty" yaml:"imagePullPolicy,omitempty"`
 	Environment          Environment           `json:"environment,omitempty" yaml:"environment,omitempty"`
 	Secrets              []Secret              `json:"secrets,omitempty" yaml:"secrets,omitempty"`
-	Healthchecks         *bool                 `json:"healthchecks,omitempty" yaml:"healthchecks,omitempty"`
-	SecurityContext      *SecurityContext      `json:"securityContext,omitempty" yaml:"securityContext,omitempty"`
-	ServiceAccount       string                `json:"serviceAccount,omitempty" yaml:"serviceAccount,omitempty"`
 	Sync                 Sync                  `json:"sync,omitempty" yaml:"sync,omitempty"`
-	Interface            string                `json:"interface,omitempty" yaml:"interface,omitempty"`
 	Resources            ResourceRequirements  `json:"resources,omitempty" yaml:"resources,omitempty"`
 	PersistentVolumeInfo *PersistentVolumeInfo `json:"persistentVolume,omitempty" yaml:"persistentVolume,omitempty"`
-	InitFromImage        *bool                 `json:"initFromImage,omitempty" yaml:"initFromImage,omitempty"`
 	Timeout              time.Duration         `json:"timeout,omitempty" yaml:"timeout,omitempty"`
 }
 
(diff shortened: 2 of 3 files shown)


pkg/model/devrc.go
```diff
@@ -0,0 +1,246 @@
...
+	"time"
+
+	"github.com/okteto/okteto/pkg/log"
+	yaml "gopkg.in/yaml.v2"
+	apiv1 "k8s.io/api/core/v1"
+)
+
+// Dev represents a development container
+type DevRC struct {
+	Autocreate           *bool                 `json:"autocreate,omitempty" yaml:"autocreate,omitempty"`
```
pchico83, line 17: Let's just support labels, annotations, context, namespace, environment, secrets, syncs, resources, PersistentVolumeInfo and timeout for now

pkg/model/devrc.go
```diff
@@ -0,0 +1,246 @@
...
+// Get returns a Dev object from a given file
+func GetRc(devPath string) (*DevRC, error) {
+	b, err := ioutil.ReadFile(devPath)
+	if err != nil {
+		return nil, err
+	}
+
+	dev, err := ReadRC(b)
+	if err != nil {
+		log.Warning("there was an error with your developer level 'okteto.yml' and it will be ignored: %s", err.Error())
```
pchico83, line 45:
```suggestion
		log.Warning("ignoring developer overwirites defined in %s: %s", devPath, rr.Error())
```

pchico83 (approved):

pchico83:
@rberrelleza could you review this one?

pchico83 (approved):
This needs docs

cmd/utils/dev.go
```diff
@@ -50,11 +52,37 @@ func LoadDev(devPath, namespace, k8sContext string) (*model.Dev, error) {
...
+
+	loadDevRc(dev)
 	loadContext(dev, k8sContext)
 	loadNamespace(dev, namespace)
 	return dev, nil
 }
 
+func loadDevRc(dev *model.Dev) {
+	defaultDevRcPath := filepath.Join(config.GetOktetoHome(), ".okteto.yml")
+	secondaryDevRcPath := filepath.Join(config.GetOktetoHome(), ".okteto.yaml")
```
rberrelleza, line 64: can we call it `okteto.yaml` without the dot, since this is a file that's meant to be edited by the user?

cmd/utils/dev.go
```diff
@@ -50,11 +52,37 @@ func LoadDev(devPath, namespace, k8sContext string) (*model.Dev, error) {
...
+	var devRc *model.DevRC
+	var err error
+	if model.FileExists(defaultDevRcPath) {
+		devRc, err = model.GetRc(defaultDevRcPath)
+		if err != nil {
+			log.Infof("error while reading .dev file: %s", err.Error())
```
rberrelleza, line 70: can you print the full path instead of `.dev` in all of these errors ?

cmd/utils/dev.go
```diff
@@ -50,11 +52,37 @@ func LoadDev(devPath, namespace, k8sContext string) (*model.Dev, error) {
...
+		}
+	} else if model.FileExists(secondaryDevRcPath) {
+		devRc, err = model.GetRc(secondaryDevRcPath)
+		if err != nil {
+			log.Infof("error while reading .dev file: %s", err.Error())
+		}
+	} else {
+		log.Info("developer level dev file not found")
```
rberrelleza, line 78: since the default is that this file won't exist, do you think we need to log it every time? maybe only log it when it does exist?

pkg/model/devrc.go
```diff
@@ -0,0 +1,187 @@
...
+	"io/ioutil"
+	"strings"
+	"time"
+
+	"github.com/okteto/okteto/pkg/log"
+	yaml "gopkg.in/yaml.v2"
+	apiv1 "k8s.io/api/core/v1"
+)
+
+// Dev represents a development container
```
rberrelleza, line 15: I  think we should mention that this represents overrides

pkg/model/devrc.go
```diff
@@ -0,0 +1,187 @@
...
+type DevRC struct {
```
rberrelleza, line 16: how about the `docker` key?

pkg/model/devrc.go
```diff
@@ -0,0 +1,187 @@
...
+// Get returns a Dev object from a given file
+func GetRc(devPath string) (*DevRC, error) {
+	b, err := ioutil.ReadFile(devPath)
+	if err != nil {
+		return nil, err
+	}
+
+	dev, err := ReadRC(b)
+	if err != nil {
+		log.Warning("ignoring developer overwrites defined in %s: %s", devPath, err.Error())
```
rberrelleza, line 38: I think we should fail this instead of ignoring, since this will result in an `okteto up` with unexpected values.

pkg/model/devrc.go
```diff
@@ -0,0 +1,187 @@
...
+	dev := &DevRC{}
+
+	if bytes != nil {
+		if err := yaml.UnmarshalStrict(bytes, dev); err != nil {
+			if strings.HasPrefix(err.Error(), "yaml: unmarshal errors:") {
+				var sb strings.Builder
+				_, _ = sb.WriteString("Invalid manifest:\n")
+				l := strings.Split(err.Error(), "\n")
+				for i := 1; i < len(l); i++ {
+					e := strings.TrimSuffix(l[i], "in type model.Dev")
```
rberrelleza, line 56: how does this look like? I don't think we should include `model.Dev` in the error, since the typical user doesn't understand the internals of the CLI. I think it's better to talk about the file, which line is wrong, and point them to the docs.

rberrelleza: We have similar code for model.dev, could we reuse it, since this is a subset?

pkg/model/devrc.go
```diff
@@ -0,0 +1,187 @@
...
+		}
+	}
+
+	for annotationKey, annotationValue := range devRc.Annotations {
+		if _, ok := dev.Annotations[annotationKey]; !ok {
+			dev.Annotations[annotationKey] = annotationValue
+		}
+	}
+
+	if devRc.Context != "" && dev.Context == "" {
```
rberrelleza, line 87: isn't devRC supposed to overwrite the content of the local okteto.yaml? I think we should be consistent with this, so it's easier to understand.

pkg/model/devrc.go
```diff
@@ -0,0 +1,187 @@
...
+		dev.Context = devRc.Context
+	}
+
+	if devRc.Namespace != "" && dev.Namespace == "" {
+		dev.Namespace = devRc.Namespace
+	}
+
+	for _, env := range devRc.Environment {
+		if !isEnvOnDev(dev, env) {
```
rberrelleza, line 96: why do we skip the env var if it's not on the okteto.yaml? I think it's easier if we just add them, so that the developer doesn't have to add it on every single manifest.

rberrelleza:
I'm confused about the expectations of the dev file. Which one has priority? I originally through that $HOME/.okteto/okteto.yaml would override $PROJECT/okteto.yaml, since that's in the user' scope. What do you think @jLopezbarb ?

pchico83:
> I'm confused about the expectations of the dev file. Which one has priority? I originally through that $HOME/.okteto/okteto.yaml would override $PROJECT/okteto.yaml, since that's in the user' scope. What do you think 
 @jLopezbarb ?

I think it should be the other way around, or you can break things with your `$HOME/.okteto/okteto.yaml` file.
I find this useful to define common configuration across your projects, like an `NPM_TOKEN`, but feels more natural to merge values giving priority to the project configuration.
@rberrelleza which use case do you have in mind for prioritizing `$HOME/.okteto/okteto.yaml`?

jLopezbarb:
> > I'm confused about the expectations of the dev file. Which one has priority? I originally through that $HOME/.okteto/okteto.yaml would override $PROJECT/okteto.yaml, since that's in the user' scope. What do you think
> > @jLopezbarb ?
> 
> I think it should be the other way around, or you can break things with your `$HOME/.okteto/okteto.yaml` file.
> I find this useful to define common configuration across your projects, like an `NPM_TOKEN`, but feels more natural to merge values giving priority to the project configuration.
> @rberrelleza which use case do you have in mind for prioritizing `$HOME/.okteto/okteto.yaml`?

I think the same as Pablo, I see it more intuitive to have a general behavior and change it per project. If you set something on the project and forgot about the `$HOME/.okteto/okteto.yaml` it can lead into some headaches to the users until he/she realizes its because of the user level file.

rberrelleza:
> > I'm confused about the expectations of the dev file. Which one has priority? I originally through that $HOME/.okteto/okteto.yaml would override $PROJECT/okteto.yaml, since that's in the user' scope. What do you think
> > @jLopezbarb ?
> 
> I think it should be the other way around, or you can break things with your `$HOME/.okteto/okteto.yaml` file.
> I find this useful to define common configuration across your projects, like an `NPM_TOKEN`, but feels more natural to merge values giving priority to the project configuration.
> @rberrelleza which use case do you have in mind for prioritizing `$HOME/.okteto/okteto.yaml`?

In my mind, the whole point of having a developer-level file is to give the developer the flexibility to override whatever they need from the okteto.yaml that comes from the repository. I don't think we need to restrict it. If something fails, as long as we give the developer an actionable error message, they should be ok (especially considering that most developers will not be using this). 

 I can imagine someone needing to overwrite things like:
1. the context name
2. a personal token
3. change the log level 
4. inject Docker in Docker configuration

rberrelleza:
> > > I'm confused about the expectations of the dev file. Which one has priority? I originally through that $HOME/.okteto/okteto.yaml would override $PROJECT/okteto.yaml, since that's in the user' scope. What do you think
> > > @jLopezbarb ?
> > 
> > 
> > I think it should be the other way around, or you can break things with your `$HOME/.okteto/okteto.yaml` file.
> > I find this useful to define common configuration across your projects, like an `NPM_TOKEN`, but feels more natural to merge values giving priority to the project configuration.
> > @rberrelleza which use case do you have in mind for prioritizing `$HOME/.okteto/okteto.yaml`?
> 
> I think the same as Pablo, I see it more intuitive to have a general behavior and change it per project. If you set something on the project and forgot about the `$HOME/.okteto/okteto.yaml` it can lead into some headaches to the users until he/she realizes its because of the user level file.

I see it the other way around. Someone sets up the okteto.yaml in the project, and everyone uses that. Some developers might need something extra, so they can create a personal override to apply the settings they like. I'd keep it as flexible as possible to keep things simple, rather than having complex rules about what takes precedence for every field.

pchico83:
> > > > I'm confused about the expectations of the dev file. Which one has priority? I originally through that $HOME/.okteto/okteto.yaml would override $PROJECT/okteto.yaml, since that's in the user' scope. What do you think
> > > > @jLopezbarb ?
> > > 
> > > 
> > > I think it should be the other way around, or you can break things with your `$HOME/.okteto/okteto.yaml` file.
> > > I find this useful to define common configuration across your projects, like an `NPM_TOKEN`, but feels more natural to merge values giving priority to the project configuration.
> > > @rberrelleza which use case do you have in mind for prioritizing `$HOME/.okteto/okteto.yaml`?
> > 
> > 
> > I think the same as Pablo, I see it more intuitive to have a general behavior and change it per project. If you set something on the project and forgot about the `$HOME/.okteto/okteto.yaml` it can lead into some headaches to the users until he/she realizes its because of the user level file.
> 
> I see it the other way around. Someone sets up the okteto.yaml in the project, and everyone uses that. Some developers might need something extra, so they can create a personal override to apply the settings they like. I'd keep it as flexible as possible to keep things simple, rather than having complex rules about what takes precedence for every field.

@rberrelleza could you provide a real use case for it? In configuration management, the more specific usually takes precedence over the more general configuration

rberrelleza:
> > > > > I'm confused about the expectations of the dev file. Which one has priority? I originally through that $HOME/.okteto/okteto.yaml would override $PROJECT/okteto.yaml, since that's in the user' scope. What do you think
> > > > > @jLopezbarb ?
> > > > 
> > > > 
> > > > I think it should be the other way around, or you can break things with your `$HOME/.okteto/okteto.yaml` file.
> > > > I find this useful to define common configuration across your projects, like an `NPM_TOKEN`, but feels more natural to merge values giving priority to the project configuration.
> > > > @rberrelleza which use case do you have in mind for prioritizing `$HOME/.okteto/okteto.yaml`?
> > > 
> > > 
> > > I think the same as Pablo, I see it more intuitive to have a general behavior and change it per project. If you set something on the project and forgot about the `$HOME/.okteto/okteto.yaml` it can lead into some headaches to the users until he/she realizes its because of the user level file.
> > 
> > 
> > I see it the other way around. Someone sets up the okteto.yaml in the project, and everyone uses that. Some developers might need something extra, so they can create a personal override to apply the settings they like. I'd keep it as flexible as possible to keep things simple, rather than having complex rules about what takes precedence for every field.
> 
> @rberrelleza could you provide a real use case for it? In configuration management, the more specific usually takes precedence over the more general configuration

These where ideas that checker mentioned:

- a personal token
- change start command
- custom bin image
- inject Docker in Docker configuration
- sync user-specific files (e.g. my ssh key) -> this was checker's initial requirement btw
- set a fixed SSH port (for pycharm remote dev configuration)

I think that the developer override is the most specific one. You have the one at the repo, that's for everyone, and then the one on $HOME/.okteto/okteto.yaml that's just for you.

pchico83:
> > > > > > I'm confused about the expectations of the dev file. Which one has priority? I originally through that $HOME/.okteto/okteto.yaml would override $PROJECT/okteto.yaml, since that's in the user' scope. What do you think
> > > > > > @jLopezbarb ?
> > > > > 
> > > > > 
> > > > > I think it should be the other way around, or you can break things with your `$HOME/.okteto/okteto.yaml` file.
> > > > > I find this useful to define common configuration across your projects, like an `NPM_TOKEN`, but feels more natural to merge values giving priority to the project configuration.
> > > > > @rberrelleza which use case do you have in mind for prioritizing `$HOME/.okteto/okteto.yaml`?
> > > > 
> > > > 
> > > > I think the same as Pablo, I see it more intuitive to have a general behavior and change it per project. If you set something on the project and forgot about the `$HOME/.okteto/okteto.yaml` it can lead into some headaches to the users until he/she realizes its because of the user level file.
> > > 
> > > 
> > > I see it the other way around. Someone sets up the okteto.yaml in the project, and everyone uses that. Some developers might need something extra, so they can create a personal override to apply the settings they like. I'd keep it as flexible as possible to keep things simple, rather than having complex rules about what takes precedence for every field.
> > 
> > 
> > @rberrelleza could you provide a real use case for it? In configuration management, the more specific usually takes precedence over the more general configuration
> 
> These where ideas that checker mentioned:
> 
> * a personal token
> * change start command
> * custom bin image
> * inject Docker in Docker configuration
> * sync user-specific files (e.g. my ssh key) -> this was checker's initial requirement btw
> * set a fixed SSH port (for pycharm remote dev configuration)
> 
> I think that the developer override is the most specific one. You have the one at the repo, that's for everyone, and then the one on $HOME/.okteto/okteto.yaml that's just for you.

Sounds good. Let's go with that approach and get feedback

jLopezbarb:
> > > > > > > I'm confused about the expectations of the dev file. Which one has priority? I originally through that $HOME/.okteto/okteto.yaml would override $PROJECT/okteto.yaml, since that's in the user' scope. What do you think
> > > > > > > @jLopezbarb ?
> > > > > > 
> > > > > > 
> > > > > > I think it should be the other way around, or you can break things with your `$HOME/.okteto/okteto.yaml` file.
> > > > > > I find this useful to define common configuration across your projects, like an `NPM_TOKEN`, but feels more natural to merge values giving priority to the project configuration.
> > > > > > @rberrelleza which use case do you have in mind for prioritizing `$HOME/.okteto/okteto.yaml`?
> > > > > 
> > > > > 
> > > > > I think the same as Pablo, I see it more intuitive to have a general behavior and change it per project. If you set something on the project and forgot about the `$HOME/.okteto/okteto.yaml` it can lead into some headaches to the users until he/she realizes its because of the user level file.
> > > > 
> > > > 
> > > > I see it the other way around. Someone sets up the okteto.yaml in the project, and everyone uses that. Some developers might need something extra, so they can create a personal override to apply the settings they like. I'd keep it as flexible as possible to keep things simple, rather than having complex rules about what takes precedence for every field.
> > > 
> > > 
> > > @rberrelleza could you provide a real use case for it? In configuration management, the more specific usually takes precedence over the more general configuration
> > 
> > 
> > These where ideas that checker mentioned:
> > 
> > * a personal token
> > * change start command
> > * custom bin image
> > * inject Docker in Docker configuration
> > * sync user-specific files (e.g. my ssh key) -> this was checker's initial requirement btw
> > * set a fixed SSH port (for pycharm remote dev configuration)
> > 
> > I think that the developer override is the most specific one. You have the one at the repo, that's for everyone, and then the one on $HOME/.okteto/okteto.yaml that's just for you.
> 
> Sounds good. Let's go with that approach and get feedback

I understand what you are saying and I think it's great for the fields that you are going to use always the same value, but what if a developer works in two projects and in one he needs to change the project start command but not in the other one. Does he have to change his developer level depending on the project he is working on at that moment?

rberrelleza:
> I understand what you are saying and I think it's great for the fields that you are going to use always the same value, but what if a developer works in two projects and in one he needs to change the project start command but not in the other one. Does he have to change his developer level depending on the project he is working on at that moment?

Yeah, I'd leave that for the developer to decide, instead of us enforcing one way or another. I think having a clear messaging when an override is applied should be enough in case something goes wrong. If this becomes a problem for our users, we can always add more restrictions later.

PR #1632, opened 2021-06-29
Validate number of args in command

### LINKED ISSUE #1628: analytics command accepts an argument but expects a flag
**Describe the bug**
The analytics command accepts an argument (like disable) but expects only flags.

**To Reproduce**

1. Run `okteto analytics disable`
2. See ` ✓  Analytics have been enabled`

**Expected behavior**
I would expect the command to exit with a non-zero status code and a message along the lines of "unexpected argument" along with the help being printed out

```console
➜  ~ okteto version  
okteto version 1.13.0
```

jLopezbarb:

Fixes #1628

## Proposed changes
- Validates the number of arguments allowed to enter a command and returns an error with the help of the command.
-

### DIFF
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -15,6 +15,7 @@ package cmd
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"path/filepath"
 
@@ -38,7 +39,13 @@ func Build(ctx context.Context) *cobra.Command {
 	var secrets []string
 
 	cmd := &cobra.Command{
-		Use:   "build [PATH]",
+		Use: "build [PATH]",
+		Args: func(cmd *cobra.Command, args []string) error {
+			if len(args) >= 2 {
+				return errors.New("okteto build requires one or no argument")
+			}
+			return nil
+		},
 		Short: "Build (and optionally push) a Docker image",
 		RunE: func(cmd *cobra.Command, args []string) error {
 
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -39,7 +39,7 @@ func Build(ctx context.Context) *cobra.Command {
 
 	cmd := &cobra.Command{
 		Use:   "build [PATH]",
-		Args:  cobra.MaximumNArgs(1),
+		Args:  utils.MaximumNArgsAccepted(1, "https://okteto.com/docs/reference/cli/index.html#build"),
 		Short: "Build (and optionally push) a Docker image",
 		RunE: func(cmd *cobra.Command, args []string) error {
 
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -15,7 +15,6 @@ package cmd
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"path/filepath"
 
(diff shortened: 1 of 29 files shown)


cmd/build.go
```diff
@@ -38,7 +39,13 @@ func Build(ctx context.Context) *cobra.Command {
 	var secrets []string
 
 	cmd := &cobra.Command{
-		Use:   "build [PATH]",
+		Use: "build [PATH]",
+		Args: func(cmd *cobra.Command, args []string) error {
+			if len(args) >= 2 {
+				return errors.New("okteto build requires one or no argument")
```
pchico83, line 45: Can we have the error in `errors.go` to make it easier to review CLI error messages by other teams?

cmd/login.go
```diff
@@ -31,6 +31,7 @@ func Login() *cobra.Command {
 	token := ""
 	cmd := &cobra.Command{
 		Use:   "login [url]",
+		Args:  cobra.MaximumNArgs(1),
```
pchico83, line 34: how does the error look like in this case?

cmd/build.go
pchico83: could we use `cobra.MaximumNArgs(1)`?

cmd/login.go
jLopezbarb: Captura de pantalla 2021-06-30 a las 17 23 14

pchico83 (approved):

rberrelleza:
That error is not very actionable, is it possible to customize the text per command to give a hint to the user? Specially for commands that don't take arguments.

jLopezbarb:
@rberrelleza @melissa-okteto I'm not satisfied with the hint that we can give to the users. The error messages given are as follows:
- For no args command:
Captura de pantalla 2021-07-01 a las 15 53 14
- For commands with maximum number of args:
Captura de pantalla 2021-07-01 a las 16 03 22
- For command with minimum number of args: 
Captura de pantalla 2021-07-01 a las 16 04 19
- For commands with a specified number of args:
Captura de pantalla 2021-07-01 a las 16 05 29

For no args command without documentation page the hint is hidden which is the part that I don't like because it's not consistent to show the documentation on some parts and not in others, so if you can think of what to show on the hint, any suggestion will be welcome .

melissa-okteto:
@jLopezbarb, when the hint is hidden, is there anything shows up? The messages above seem clear.

> For no args command without documentation page the hint is hidden which is the part that I don't like because it's not consistent to show the documentation on some parts and not in others, so if you can think of what to show on the hint, any suggestion will be welcome .

jLopezbarb:
if the hint is hidden we don't show any text other than the error in red

rberrelleza:
> @rberrelleza @melissa-okteto I'm not satisfied with the hint that we can give to the users. The error messages given are as follows:
> 
> * For no args command:
> 
> Captura de pantalla 2021-07-01 a las 15 53 14
> 
> - For commands with maximum number of args:
> Captura de pantalla 2021-07-01 a las 16 03 22
> 
> - For command with minimum number of args:
> Captura de pantalla 2021-07-01 a las 16 04 19
> 
> - For commands with a specified number of args:
> Captura de pantalla 2021-07-01 a las 16 05 29
> 
> For no args command without documentation page the hint is hidden which is the part that I don't like because it's not consistent to show the documentation on some parts and not in others, so if you can think of what to show on the hint, any suggestion will be welcome .

What if we just show "Invalid command" or "Invalid arguments" and show the help of the command?

rberrelleza:
Could we just apply this to ` okteto analytics`? I feel like for all the other commands the error we already display when you put the number of arguments is good enough.

pchico83:
> Could we just apply this to ` okteto analytics`? I feel like for all the other commands the error we already display when you put the number of arguments is good enough.

We are not validating arguments for most commands. For example, you can do `okteto push xxx` and it doesn't fail

PR #1636, opened 2021-07-01
Check if port is open on tcp protocol instead of path

jLopezbarb:

Fixes A infinite loop in cases where `ip:port/` is not accepting requests

## Proposed changes

- check if the port is open in server side
-

### DIFF
--- a/pkg/cmd/stack/deploy.go
+++ b/pkg/cmd/stack/deploy.go
@@ -17,7 +17,7 @@ import (
 	"context"
 	"encoding/base64"
 	"fmt"
-	"net/http"
+	"net"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


pchico83 (approved):

pkg/cmd/stack/deploy.go
```diff
@@ -274,18 +274,12 @@ func isSvcHealthy(ctx context.Context, svc *model.Service, stack *model.Stack, s
 	forwarder.Start(podName, stack.Namespace)
 	defer forwarder.Stop()
 	for _, port := range portsToTest {
-		url := fmt.Sprintf("http://%s:%d/", model.Localhost, port)
-		resp, err := http.Get(url)
+		url := fmt.Sprintf("%s:%d", model.Localhost, port)
+		_, err := net.Dial("tcp", url)
```
rberrelleza, line 278: we have similar code in the up sequence, which I think also does retries and other stuff. do you think we should use that here as well?

jLopezbarb: It's already retrying but it will launch other services that are ready to be launch before retrying instead of waiting for a particular service to be available

PR #1638, opened 2021-07-01
Ignore build argument on up if there is nothing to build

jLopezbarb:

Fixes If the Dockerfile does not exists and the user has set `--build` flag it will be ignored instead of failing.

## Proposed changes

- Ignore build argument if there is nothing to build
-

### DIFF
--- a/cmd/up/up.go
+++ b/cmd/up/up.go
@@ -385,7 +385,7 @@ func (up *upContext) waitUntilExitOrInterrupt() error {
 func (up *upContext) buildDevImage(ctx context.Context, d *appsv1.Deployment, create bool) error {
 	if _, err := os.Stat(up.Dev.Image.Dockerfile); err != nil {
 		return errors.UserError{
-			E:    fmt.Errorf("'--build' argument given but there is no Dockerfile on cwd"),
+			E:    fmt.Errorf("'--build' argument given but there is no Dockerfile"),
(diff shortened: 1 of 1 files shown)


cmd/up/up.go
```diff
@@ -383,6 +383,11 @@ func (up *upContext) waitUntilExitOrInterrupt() error {
 }
 
 func (up *upContext) buildDevImage(ctx context.Context, d *appsv1.Deployment, create bool) error {
+	if _, err := os.Stat(up.Dev.Image.Dockerfile); err != nil {
+		log.Warning("Ignoring '--build' argument. There is not 'build' primitives in your manifest")
```
pchico83, line 387: I think it is better to fail with a clear message

cmd/up/up.go
```diff
@@ -383,6 +383,13 @@ func (up *upContext) waitUntilExitOrInterrupt() error {
 }
 
 func (up *upContext) buildDevImage(ctx context.Context, d *appsv1.Deployment, create bool) error {
+	if _, err := os.Stat(up.Dev.Image.Dockerfile); err != nil {
+		return errors.UserError{
+			E:    fmt.Errorf("'--build' argument given but there is no Dockerfile on cwd"),
```
pchico83, line 388: `up.Dev.Image.Dockerfile` could be different then `Dockerfile`

pchico83 (approved):

PR #1641, opened 2021-07-02
fix missconfigured terminal after ctrl +c

jLopezbarb:
Fixes Missconfigured terminal after ctrl+c for all commands

## Proposed changes
- Return exit code 130(128+2) when stopped by CTRL+C signal.
- Remove spinner text after CTRL+C, that's because besides returning the cursor to the terminal, the spinner checks other properties of the terminal that should be returned to the original state


cmd/down.go
```diff
@@ -89,6 +89,7 @@ func runDown(ctx context.Context, dev *model.Dev) error {
 	spinner := utils.NewSpinner("Deactivating your development container...")
 	spinner.Start()
 	defer spinner.Stop()
+	go utils.StopSpinnerIfInterruptSignal(spinner)
```
pchico83, line 92: could we execute this as part of `spinner.Stop()`?

jLopezbarb: If we do that we could not run the shutdown sequence on `okteto up` command, because we need to return signal exit code, otherwise what `CTRL+C` will do is just running `spinner.Stop()`

pchico83: Can we have two `spinner.Stop` functions and call one or the other depending on the command?

cmd/utils/spinner.go
```diff
@@ -105,3 +106,11 @@ func ucFirst(str string) string {
 	}
 	return ""
 }
+
+func StopSpinnerIfInterruptSignal(spinner *Spinner) {
+	stopChannel := make(chan os.Signal, 1)
```
pchico83, line 111: Let's move this logic to the main goroutine, and wait for CTRL+C or the end of the execution in the main goroutine.

cmd/down.go
```diff
@@ -91,32 +92,50 @@ func runDown(ctx context.Context, dev *model.Dev) error {
...
+		if err != nil {
+			exit <- err
+		}
+		exit <- nil
+	}()
+
+	select {
+	case <-stop:
+		log.Infof("CTRL+C received, starting shutdown sequence")
+		spinner.Stop()
```
pchico83, line 131: Do we need this call? or would the `defer` make the job?
https://github.com/okteto/okteto/pull/1641/files#diff-41918fbb88ea02aac469b573629ce8a0dc3010817fbb0b50e5e6509f22165f43R93
Same for other commands

pchico83 (approved):

PR #1645, opened 2021-07-07
Jlopezbarb/wait for autoescaler creating new nodes

### LINKED ISSUE #1622: Wait for available AKS resources (Okteto CLI)
I'm using the Okteto CLI in combination with AKS. 

When there are no sufficient resources available in AKS and I run Okteto up, the deployment fails with an "insufficient resources" error. In the meantime k8s/aks figured out it needs to scale out and additional nodes are provisioned. After about 4 minutes an extra node is ready and the Okteto deployment (was still trying to deploy) succeeds. However the Okteto cli already backed off with an error.

A similar option as timeout would be great where we can specify the amount of minutes a deployment can wait or "look" for available resources.

example:
resource_timeout: 6m

When this option is used, instead of an immediate error it would be nice to see an info/warning message saying it's will look for resources for $resource_timeout more minutes.

Two alternatives considered: running Okteto up again after a couple of minutes and see if it works then, or running more default resources which means overpaying on nodesmost of the time.

jLopezbarb:
Fixes #1622

## Proposed changes
- Add extended timeout notation
- Wait for autoscaler to create new nodes until timeout of resources expires

### DIFF
--- a/cmd/up/activate.go
+++ b/cmd/up/activate.go
@@ -91,6 +91,9 @@ func (up *upContext) activate(autoDeploy, build bool) error {
 		if strings.Contains(err.Error(), "Privileged containers are not allowed") && up.Dev.Docker.Enabled {
 			return fmt.Errorf("Docker support requires privileged containers. Privileged containers are not allowed in your current cluster")
 		}
+		if _, ok := err.(errors.UserError); ok {
+			return err
+		}
 		return fmt.Errorf("couldn't activate your development container\n    %s", err.Error())
 	}
 
@@ -313,7 +316,8 @@ func (up *upContext) waitUntilDevelopmentContainerIsRunning(ctx context.Context)
 	var insufficientResourcesErr error
 	for {
 		if time.Now().After(to) && insufficientResourcesErr != nil {
-			return insufficientResourcesErr
+			return errors.UserError{E: fmt.Errorf("Insufficient resources."),
+				Hint: "Increase cluster resources or timeout of resources. More information is available here: https://okteto.com/docs/reference/manifest#timeout"}
 		}
 		select {
 		case event := <-watcherEvents.ResultChan():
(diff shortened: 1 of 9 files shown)


cmd/up/activate.go
```diff
@@ -328,6 +334,11 @@ func (up *upContext) waitUntilDevelopmentContainerIsRunning(ctx context.Context)
 				if strings.Contains(e.Message, "pod has unbound immediate PersistentVolumeClaims") {
 					continue
 				}
+				if strings.Contains(e.Message, "Insufficient cpu") || strings.Contains(e.Message, "Insufficient memory") {
+					insufficientResourcesErr = fmt.Errorf(e.Message)
+					spinner.Update("Waiting for new resources to be available...")
```
pchico83, line 339: Insufficient cpu/memory in the cluster. Waiting for new nodes to come up...

pkg/model/dev.go
```diff
@@ -576,7 +585,10 @@ func (dev *Dev) setRunAsUserDefaults(main *Dev) {
 }
 
 func (dev *Dev) setTimeout() error {
-	if dev.Timeout != 0 {
+	if dev.Timeout.Resources == 0 {
+		dev.Timeout.Resources = dev.Timeout.Default * 4
```
pchico83, line 589: let's use 120

cmd/up/activate.go
```diff
@@ -308,7 +309,12 @@ func (up *upContext) waitUntilDevelopmentContainerIsRunning(ctx context.Context)
 		return err
 	}
 
+	to := time.Now().Add(up.Dev.Timeout.Resources)
+	var insufficientResourcesErr error
 	for {
+		if time.Now().After(to) && insufficientResourcesErr != nil {
+			return insufficientResourcesErr
```
pchico83, line 316: I would put a link to the docs to configure the timeout and explain what happened and how the user can configure a longer timeout if needed

pchico83 (approved):

PR #1649, opened 2021-07-12
Create only token if already set user id

### LINKED ISSUE #1630: ".token.json" is empty
**Describe the bug**
Doing an `okteto create namespace <name>` without any kind of credential creates an empty `.token.json` file.
Consecutive executions of the same command result in an `internal-server-error`.
If I execute an `okteto login` with the right credentials, it starts working.

**Solution**
HAndle the error better. Never create an empty token

jLopezbarb:

Fixes #1630

## Proposed changes
- Not create token.json if there is no user on it already
-

### DIFF
--- a/pkg/analytics/analytics.go
+++ b/pkg/analytics/analytics.go
@@ -363,11 +363,9 @@ func getMachineID() string {
 	}
 
 	mid = generateMachineID()
-	if okteto.GetUserID() != "" {
-		if err := okteto.SaveMachineID(mid); err != nil {
-			log.Info("failed to save the machine id")
-			mid = "na"
-		}
+	if err := okteto.SaveMachineID(mid); err != nil {
+		log.Info("failed to save the machine id")
+		mid = "na"
 	}
 
 	return mid
--- a/pkg/analytics/analytics.go
+++ b/pkg/analytics/analytics.go
@@ -363,9 +363,11 @@ func getMachineID() string {
 	}
 
 	mid = generateMachineID()
-	if err := okteto.SaveMachineID(mid); err != nil {
-		log.Info("failed to save the machine id")
-		mid = "na"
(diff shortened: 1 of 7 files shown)


pkg/analytics/analytics.go
```diff
@@ -363,9 +363,11 @@ func getMachineID() string {
 	}
 
 	mid = generateMachineID()
-	if err := okteto.SaveMachineID(mid); err != nil {
-		log.Info("failed to save the machine id")
-		mid = "na"
+	if okteto.GetUserID() != "" {
```
rberrelleza, line 366: will this mean that, for users who never log in, we'll never store the machineID?

jLopezbarb: yes, we will not store their machineID but we will retrieve it every time we need to use it.

rberrelleza: I don't get why we can't store the machineID though. it will a) save time (not sure how long the calculation takes) and b) ensure that it's always the same (in case something changes in the machineID calculation).

pchico83: Let's store the machineID in a different file

rberrelleza: That will require changes in the vscode plugin. Why do we need to store it somewhere else?

pchico83: Why does it require those changes? Creating a file named `.token.json` with no token doesn't seem right

rberrelleza: The file is called token.JSON but it includes more things (machineID, userID, urls, etc), so it won't be empty, it will always have the machine id.

The vscode plug-in uses the machineID  for analytics. It can read it from somewhere else, but it sounds unnecessary to have two files for this info no?

pchico83 (approved):

PR #1651, opened 2021-07-13
Add login url confirmation

### LINKED ISSUE #1629: "okteto login" should default to the current login URL
When running `okteto login` without a URL, it should default to the current login URL.
If not URL is provided, we should ask for confirmation and provide an argument to skip confirmation.

jLopezbarb:

Fixes #1629

## Proposed changes
- Adds confirmation for default url
- Adds a force command to skip confirmation

### DIFF
--- a/cmd/login.go
+++ b/cmd/login.go
@@ -29,7 +29,6 @@ import (
 // Login starts the login handshake with GitHub and okteto
 func Login() *cobra.Command {
 	token := ""
-	force := false
 	cmd := &cobra.Command{
 		Use:   "login [url]",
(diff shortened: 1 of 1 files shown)


cmd/login.go
```diff
@@ -29,6 +29,7 @@ import (
 // Login starts the login handshake with GitHub and okteto
 func Login() *cobra.Command {
 	token := ""
+	force := false
```
pchico83, line 32: what's the use case for this? if you pass the url, there is no ask for confirmation, right?

pchico83 (approved):

PR #1657, opened 2021-07-15
Reduce the amount of logs during init container from image

aguthrie:
We are using okteto with Docker containers that contain all our app code, including a PHP vendor/ directory with contains >100,000 files. Using the verbose copy flag is causing every one of those files to be logged when it's copied and overwhelming our log processor.

This removes logging every file in favor of only logging the commands run. This should hopefully still provide enough info to debug if something goes wrong during the copy and still show what top-level folders/files are copied.

## Proposed changes
- disable logging every file
- only log which commands run


pchico83 (approved):
Awesome PR @aguthrie
We were logging everything to help to troubleshoot this feature in case the copy of files was introducing issues, but I think it is a good time to remove all those logs.
Thanks for your contribution!

PR #1684, opened 2021-08-04
Deprecate environment in favour of variables

jLopezbarb:

Fixes #

## Proposed changes
- Deprecate the use of environment in favour of variables field
-

### DIFF
--- a/pkg/model/dev.go
+++ b/pkg/model/dev.go
@@ -293,13 +293,13 @@ func Get(devPath string) (*Dev, error) {
 // Read reads an okteto manifests
 func Read(bytes []byte) (*Dev, error) {
 	dev := &Dev{
-		Image: &BuildInfo{},
-		Push:  &BuildInfo{},
-		// Environment: make(Environment, 0),
-		Variables: make(Variables, 0),
-		Secrets:   make([]Secret, 0),
-		Forward:   make([]Forward, 0),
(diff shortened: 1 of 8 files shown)


pkg/model/dev.go
```diff
@@ -289,12 +293,13 @@ func Get(devPath string) (*Dev, error) {
...
 	dev := &Dev{
-		Image:       &BuildInfo{},
-		Push:        &BuildInfo{},
-		Environment: make(Environment, 0),
-		Secrets:     make([]Secret, 0),
-		Forward:     make([]Forward, 0),
-		Volumes:     make([]Volume, 0),
+		Image: &BuildInfo{},
+		Push:  &BuildInfo{},
+		// Environment: make(Environment, 0),
```
ifbyol, line 298: I guess it can be deleted, right?

pkg/model/dev.go
```diff
@@ -254,6 +255,9 @@ type Annotations map[string]string
 // Environment is a list of environment variables (key, value pairs).
 type Environment []EnvVar
```
ifbyol, line 256: is this type still being used?

jLopezbarb: Yes, we need to keep it to support retrocompatibility for a time

ifbyol (approved):

pchico83 (approved):

PR #1694, opened 2021-08-12
Interpolate "OKTETO_USERNAME" in preview cmd

### LINKED ISSUE #1689: Interpolate "OKTETO_USERNAME" in the name of the preview environment
**Describe the bug**
I'd like to be able to create a preview env like this: `okteto preview deploy review-mybranch-$OKTETO_USERNAME` and have okteto automatically fill in $OKTETO_USERNAME

jLopezbarb:

Fixes #1689

## Proposed changes

- Adds the option to use OKTETO_USERNAME variable in command
-

### DIFF
--- a/cmd/preview/deploy.go
+++ b/cmd/preview/deploy.go
@@ -74,7 +74,7 @@ func Deploy(ctx context.Context) *cobra.Command {
 			if len(args) == 0 {
 				name = getRandomName(ctx, scope)
 			} else {
-				name = args[0]
+				name = getExpandedName(args[0])
 			}
 
 			varList := []okteto.Variable{}
@@ -262,4 +262,14 @@ func waitForResourcesToBeRunning(ctx context.Context, name, namespace string, t 
 		}
 	}
 	return nil
+}
+
+func getExpandedName(name string) string {
+	username := okteto.GetUsername()
+	os.Setenv("OKTETO_USERNAME", username)
+	expandedName, err := model.ExpandEnv(name)
+	if err != nil {
+		return name
+	}
+	return expandedName
 }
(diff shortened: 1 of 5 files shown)


ifbyol (approved):

cmd/preview/deploy.go
```diff
@@ -263,3 +263,13 @@ func waitForResourcesToBeRunning(ctx context.Context, name, namespace string, t
 	}
 	return nil
 }
+
+func getExpandedName(name string) string {
+	username := okteto.GetUsername()
+	os.Setenv("OKTETO_USERNAME", username)
```
rberrelleza, line 269: this will set OKTETO_USERNAME in the env for the entire execution. Is that what we want? (if so, then I'd suggest we do that earlier, on a separate function, so we can use it in other places as well). 

IF not,  maybe we can use a library that only does the expansion for a string, without requiring it in the global env? (I think we use envsubst somewhere else?)

pchico83: I would go with option a)

cmd/preview/deploy.go
```diff
@@ -57,6 +57,10 @@ func Deploy(ctx context.Context) *cobra.Command {
 				return okErrors.ErrNotLogged
 			}
 
+			if err := utils.SetOktetoUsernameEnv(); err != nil {
```
pchico83, line 60: I would do it as part of the global init function:
https://github.com/okteto/okteto/blob/master/main.go#L45

cmd/utils/okteto.go
```diff
@@ -0,0 +1,28 @@
...
+
+import (
+	"os"
+
+	"github.com/okteto/okteto/pkg/okteto"
+)
+
+func SetOktetoUsernameEnv() error {
+	username := okteto.GetUsername()
+	if err := os.Setenv("OKTETO_USERNAME", username); err != nil {
```
pchico83, line 24: Let's check if `OKTETO_USERNAME` has a value. If so, I wouldn't overwrite it

pchico83 (approved):

pchico83 (approved):

PR #1697, opened 2021-08-12
Add preview endpoints command

### LINKED ISSUE #1665: Support for "okteto preview endpoints"
I want to see the endpoints of my preview environment.

jLopezbarb:

Fixes #1665

## Proposed changes

- Adds okteto preview endpoints command
-


ifbyol (approved):

rberrelleza (commented):
one common use case I can see for this command is to use it as part of an automated script (e.g. get the endpoint and then run tests, or send it to slack). have we considered an option for having a different format? 

Also, could we integrate this with our current github action for preview envs? I think it would be better if the action used this instead of the custom go code (from a maintainability perspective)

pchico83 (approved):

jLopezbarb:
using this code instead of the one we have in the action will result in not being able to create a hyperlink to each of them since you need to use markdown syntax

rberrelleza:
Then what is the use case for this command? Is this for people to run it locally on their own machines?

jLopezbarb:
I think it can be useful for Gitlab actions where we don't write any comment on the PR

cmd/preview/preview.go
```diff
@@ -34,8 +34,12 @@ func Preview(ctx context.Context) *cobra.Command {
 	listCommand := List(ctx)
 	listCommand.Hidden = true
 
+	endpointsCommand := Endpoints(ctx)
+	endpointsCommand.Hidden = true
```
pchico83, line 38: Let's show all info about `previews`

pkg/okteto/preview.go
```diff
@@ -136,6 +143,42 @@ func ListPreviews(ctx context.Context) ([]PreviewEnv, error) {
 	return body.Previews, nil
 }
 
+func ListPreviewsEndpoints(ctx context.Context, previewName string) ([]string, error) {
+	endpoints := make([]string, 0)
+
+	q := fmt.Sprintf(`query{
+		preview(id: "%s"){
+			deployments{
```
pchico83, line 151: You need to query `gitDeploys`, `stacks` and `apps` too, and for each of them, `deployments` and `statefulsets` endpoints:
https://github.com/okteto/app/blob/master/backend/pkg/graphql/types.go#L277

pkg/okteto/preview.go
```diff
@@ -136,6 +143,42 @@ func ListPreviews(ctx context.Context) ([]PreviewEnv, error) {
...
+		}
+	}`, previewName)
+	var body PreviewBody
+	if err := query(ctx, q, &body); err != nil {
+		return nil, err
+	}
+
+	for _, d := range body.Preview.Deployments {
+		for _, endpoint := range d.Endpoints {
+			endpoints = append(endpoints, endpoint.URL)
```
pchico83, line 170: @maroshii how is the Okteto UI sorting the endpoints?
We should follow the same logic here

pchico83 (commented):
@jLopezbarb Let's have a `-o=json` flag to show the endpoints. This way we can use that option to parse the output in the `deploy-preview` action. This is something that people might need to parse the output in other CI tools

pkg/okteto/preview.go
ifbyol: @pchico83  In the UI endpoints endpoints are sorted by length. Shorter endpoints first. I implemented that logic on the deploy preview github action: https://github.com/okteto/deploy-preview/blob/master/message.go#L174

pchico83: @jLopezbarb let's add that logic here too

PR #1701, opened 2021-08-13
Improve e2e tests

jLopezbarb:
improves the robustness of end-to-end testing

## Proposed changes
- Increase the number of retries 
-


pchico83:
@jLopezbarb if windows is giving issues, maybe playing with this field helps:
https://github.com/okteto/okteto/blob/master/.circleci/config.yml#L109

PR #1707, opened 2021-08-16
Jlopezbarb/fix e2e win

jLopezbarb:
Test windows e2e tests

## Proposed changes

-
-


cmd/up/activate.go
```diff
@@ -402,6 +402,9 @@ func (up *upContext) waitUntilDevelopmentContainerIsRunning(ctx context.Context,
 				}
 				continue
 			}
+			if up.Pod.UID != pod.UID {
```
pchico83, line 405: Can we do this in the watch events?
We should ignore events from other pods in the `event := <-watcherEvents.ResultChan()` case too

pkg/k8s/apps/translate.go
```diff
@@ -704,18 +707,23 @@ func TranslateOktetoDevSecret(spec *apiv1.PodSpec, secret string, secrets []mode
...
+				log.Infof("%s/%s is not a development container", k8sObject.Namespace, k8sObject.Name)
+				return k8sObject, nil
+			}
 			dOrig := &appsv1.Deployment{}
 			if err := json.Unmarshal([]byte(dManifest), dOrig); err != nil {
 				return nil, fmt.Errorf("malformed manifest: %s", err)
 			}
 			k8sObject.UpdateDeployment(dOrig)
 		} else {
+			dManifest := annotations.Get(k8sObject.GetObjectMeta(), model.StatefulsetAnnotation)
```
pchico83, line 722: rename `d` variables tp `sfs` names

cmd/up/activate.go
```diff
@@ -356,7 +356,18 @@ func (up *upContext) waitUntilDevelopmentContainerIsRunning(ctx context.Context,
 				}
 				continue
 			}
-
+			pod, ok := event.Object.(*apiv1.Pod)
+			if !ok {
```
pchico83, line 360: what is this loop for? I would just ignore the event if we cannot get the pod of the event

pkg/k8s/pods/pod.go
```diff
@@ -178,9 +179,14 @@ func GetPodByStatefulSet(ctx context.Context, sfs *appsv1.StatefulSet, labels st
 		return nil, err
 	}
 	for i := range podList.Items {
-		for _, or := range podList.Items[i].OwnerReferences {
-			if or.UID == sfs.UID {
-				return &podList.Items[i], nil
+		if podList.Items[i].DeletionTimestamp != nil {
+			return nil, nil
```
pchico83, line 183: would be better a `continue`?

pchico83 (approved):

cmd/up/activate.go
jLopezbarb: That's for StatefulSet. We can get events from the previous pod and it will say that's its pulling images but it's still attaching pvs. This will check that the pod of the event is the one that we need

PR #1725, opened 2021-08-31
Set pipeline name from git config

jLopezbarb:

Fixes #

## Proposed changes
- Get pipeline name from git config. If there is no git config get it from dir name
-

### DIFF
--- a/cmd/pipeline/deploy.go
+++ b/cmd/pipeline/deploy.go
@@ -55,14 +55,6 @@ func deploy(ctx context.Context) *cobra.Command {
 				return errors.ErrNotLogged
 			}
 
-			var err error
-			if name == "" {
-				name, err = getPipelineName(repository)
-				if err != nil {
-					return err
-				}
-			}
-
 			cwd, err := os.Getwd()
 			if err != nil {
 				return fmt.Errorf("failed to get the current working directory: %w", err)
@@ -71,14 +63,17 @@ func deploy(ctx context.Context) *cobra.Command {
 			if repository == "" {
 				log.Info("inferring git repository URL")
 
-				r, err := model.GetRepositoryURL(cwd)
-
+				repository, err = model.GetRepositoryURL(cwd)
 				if err != nil {
 					return err
 				}
+			}
 
-				repository = r
-
+			if name == "" {
+				name, err = getPipelineName(repository)
+				if err != nil {
+					return err
+				}
 			}
 
 			if branch == "" {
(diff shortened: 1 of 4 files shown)


cmd/pipeline/deploy.go
```diff
@@ -186,8 +185,14 @@ func getPipelineName() (string, error) {
 	if err != nil {
 		return "", err
 	}
-
-	return filepath.Base(workDir), nil
+	name, err := model.GetValidNameFromGitRepo(workDir)
+	if err != nil {
```
pchico83, line 189: if it is not a git repo, we cannot deploy it, right?

cmd/pipeline/deploy.go
```diff
@@ -186,8 +185,14 @@ func getPipelineName() (string, error) {
 	if err != nil {
 		return "", err
 	}
-
-	return filepath.Base(workDir), nil
+	name, err := model.GetValidNameFromGitRepo(workDir)
```
pchico83, line 188: I think we need to do the same for destroy

jLopezbarb: This method is also called from destroy

cmd/pipeline/deploy.go
jLopezbarb: If you set the right repository and branch you can deploy it like:
`okteto pipeline deploy --repository https://github.com/okteto/movies --branch master`
It will create the pipeline with the name of the folder that you are currently in

pchico83: What I would do is to infer the repository first, and then, use the repo to infer the name.
That way we have consistent naming even if using the `okteto pipeline deploy --repository https://github.com/okteto/movies` syntax

cmd/pipeline/deploy.go
```diff
@@ -58,7 +57,7 @@ func deploy(ctx context.Context) *cobra.Command {
 
 			var err error
 			if name == "" {
-				name, err = getPipelineName()
+				name, err = getPipelineName(repository)
```
pchico83, line 60: Where are we infering the repo? could we infer the repo before this point?
that way, we will always infer the name from the repo

cmd/pipeline/destroy.go
```diff
@@ -51,7 +51,7 @@ func destroy(ctx context.Context) *cobra.Command {
 
 			var err error
 			if name == "" {
-				name, err = getPipelineName()
+				name, err = getPipelineName("")
```
pchico83, line 54: here, if it is not a git repo and name is not provided, we should fail

cmd/pipeline/deploy.go
```diff
@@ -181,13 +175,8 @@ func deployPipeline(ctx context.Context, name, namespace, repository, branch, fi
...
 }
 
-func getPipelineName() (string, error) {
-	workDir, err := os.Getwd()
-	if err != nil {
-		return "", err
-	}
-
-	return filepath.Base(workDir), nil
+func getPipelineName(repository string) (string, error) {
```
pchico83, line 178: if error is always `nil`, we can just return a string

pchico83 (approved):

PR #1730, opened 2021-09-01
Add local build

jLopezbarb:

Fixes #

## Proposed changes
- Adds support for building images with docker daemon locally

### DIFF
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -28,15 +28,8 @@ import (
 
 //Build build and optionally push a Docker image
 func Build(ctx context.Context) *cobra.Command {
-	var file string
-	var tag string
-	var target string
-	var noCache bool
-	var cacheFrom []string
-	var progress string
-	var buildArgs []string
-	var secrets []string
 
+	options := build.BuildOptions{}
 	cmd := &cobra.Command{
 		Use:   "build [PATH]",
 		Args:  utils.MaximumNArgsAccepted(1, "https://okteto.com/docs/reference/cli/#build"),
@@ -55,12 +48,13 @@ func Build(ctx context.Context) *cobra.Command {
 			if err := utils.CheckIfDirectory(path); err != nil {
 				return fmt.Errorf("invalid build context: %s", err.Error())
 			}
+			options.Path = path
 
-			if file == "" {
-				file = filepath.Join(path, "Dockerfile")
+			if options.File == "" {
+				options.File = filepath.Join(path, "Dockerfile")
 			}
 
-			if err := utils.CheckIfRegularFile(file); err != nil {
+			if err := utils.CheckIfRegularFile(options.File); err != nil {
 				return fmt.Errorf("invalid Dockerfile: %s", err.Error())
 			}
 
@@ -68,19 +62,23 @@ func Build(ctx context.Context) *cobra.Command {
 			if err != nil {
 				return err
 			}
-			log.Information("Running your build in %s...", buildKitHost)
+			if buildKitHost == "" || options.BuildLocally {
+				log.Information("Running your build in docker...")
+			} else {
+				log.Information("Running your build in %s...", buildKitHost)
+			}
 
 			ctx := context.Background()
-			if err := build.Run(ctx, "", buildKitHost, isOktetoCluster, path, file, tag, target, noCache, cacheFrom, buildArgs, secrets, progress); err != nil {
+			if err := build.Run(ctx, "", buildKitHost, isOktetoCluster, options); err != nil {
 				analytics.TrackBuild(buildKitHost, false)
 				return err
 			}
 
-			if tag == "" {
+			if options.Tag == "" {
 				log.Success("Build succeeded")
 				log.Information("Your image won't be pushed. To push your image specify the flag '-t'.")
 			} else {
-				log.Success(fmt.Sprintf("Image '%s' successfully pushed", tag))
+				log.Success(fmt.Sprintf("Image '%s' successfully pushed", options.Tag))
 			}
 
 			analytics.TrackBuild(buildKitHost, true)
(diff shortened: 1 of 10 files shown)


cmd/build.go
```diff
@@ -55,46 +48,52 @@ func Build(ctx context.Context) *cobra.Command {
...
 
 			buildKitHost, isOktetoCluster, err := build.GetBuildKitHost()
 			if err != nil {
 				return err
 			}
-			log.Information("Running your build in %s...", buildKitHost)
+			if buildKitHost == "" || options.BuildLocally {
+				log.Information("Running your build in docker...")
+			} else {
+				log.Information("Running your build in %s...", buildKitHost)
```
rberrelleza, line 68: let's only log the non-default option to keep it less disruptive.

cmd/build.go
```diff
@@ -55,46 +48,52 @@ func Build(ctx context.Context) *cobra.Command {
...
 				return fmt.Errorf("invalid Dockerfile: %s", err.Error())
 			}
 
 			buildKitHost, isOktetoCluster, err := build.GetBuildKitHost()
 			if err != nil {
 				return err
 			}
-			log.Information("Running your build in %s...", buildKitHost)
+			if buildKitHost == "" || options.BuildLocally {
+				log.Information("Running your build in docker...")
```
rberrelleza, line 66: what would happen if the user doesn't have a docker daemon running? (or is using something else like podman, ko, etc...)

rberrelleza, line 66: "running your build" sounds odd. "Building your image using your local docker daemon" maybe? (not sure what docker displays here)
```suggestion
				log.Information("Running your build in docker...")
```

cmd/build.go
```diff
@@ -55,46 +48,52 @@ func Build(ctx context.Context) *cobra.Command {
...
 				return err
 			}
 
-			if tag == "" {
+			if options.Tag == "" {
 				log.Success("Build succeeded")
 				log.Information("Your image won't be pushed. To push your image specify the flag '-t'.")
 			} else {
-				log.Success(fmt.Sprintf("Image '%s' successfully pushed", tag))
+				log.Success(fmt.Sprintf("Image '%s' successfully pushed", options.Tag))
```
rberrelleza, line 81: will the tag show the entire thing? (e.g. okteto.dev/foo/bar:123)

cmd/build.go
rberrelleza: Also, what if the docker daemon is in a different machine? are we support that as well?

cmd/build.go
jLopezbarb: This is to show the user where it will be build on blue: 
Captura de pantalla 2021-09-01 a las 18 33 57

cmd/build.go
jLopezbarb: If the user does not have docker running it will fail when trying to reach the api endpoint, the same as docker does.
At the moment we are not supporting that the docker daemon is in another machine. I think this is a first step to make `okteto stack/up/push` be available on most clusters

cmd/build.go
jLopezbarb: It will show the tag that you passed as a flag, that makes the user know that what he has introduced as a flag is actually pushed

cmd/build.go
rberrelleza: ah cool 👍 , nevermind then.

cmd/build.go
rberrelleza: cool, let's make sure we have a clear error for that case  then?

pkg/cmd/build/build_docker.go
```diff
@@ -0,0 +1,315 @@
...
+	"github.com/moby/buildkit/session/filesync"
+	"github.com/moby/buildkit/util/progress/progressui"
+	"github.com/moby/buildkit/util/progress/progresswriter"
+	"github.com/moby/term"
+	"github.com/okteto/okteto/pkg/log"
+	"github.com/pkg/errors"
+	"golang.org/x/sync/errgroup"
+)
+
+func buildWithDockerDaemonBuildkit(ctx context.Context, buildOptions BuildOptions, cli *client.Client) error {
```
pchico83, line 48: Would be possible to reuse more code from moby or buildkit libraries here?

jLopezbarb: The code was made with private functions, so it is not possible to use more functions.

pchico83:
@jLopezbarb this needs rebase. Once it is merge, we should also enable `okteto stack` is clusters not managed by okteto

cmd/build.go
```diff
@@ -55,46 +48,52 @@ func Build(ctx context.Context) *cobra.Command {
...
 				return fmt.Errorf("invalid Dockerfile: %s", err.Error())
 			}
 
 			buildKitHost, isOktetoCluster, err := build.GetBuildKitHost()
 			if err != nil {
 				return err
 			}
-			log.Information("Running your build in %s...", buildKitHost)
+			if buildKitHost == "" || options.BuildLocally {
+				log.Information("Building your image using your local docker daemon")
```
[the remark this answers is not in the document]
pchico83: we can now remove the `options.BuildLocally` option, no?

cmd/build.go
```diff
@@ -55,46 +48,52 @@ func Build(ctx context.Context) *cobra.Command {
...
-	cmd.Flags().StringArrayVar(&secrets, "secret", nil, "secret files exposed to the build. Format: id=mysecret,src=/local/secret")
+	cmd.Flags().StringVarP(&options.File, "file", "f", "", "name of the Dockerfile (Default is 'PATH/Dockerfile')")
+	cmd.Flags().StringVarP(&options.Tag, "tag", "t", "", "name and optionally a tag in the 'name:tag' format (it is automatically pushed)")
+	cmd.Flags().StringVarP(&options.Target, "target", "", "", "set the target build stage to build")
+	cmd.Flags().BoolVarP(&options.NoCache, "no-cache", "", false, "do not use cache when building the image")
+	cmd.Flags().StringArrayVar(&options.CacheFrom, "cache-from", nil, "cache source images")
+	cmd.Flags().StringVarP(&options.OutputMode, "progress", "", "tty", "show plain/tty build output")
+	cmd.Flags().StringArrayVar(&options.BuildArgs, "build-arg", nil, "set build-time variables")
+	cmd.Flags().StringArrayVar(&options.Secrets, "secret", nil, "secret files exposed to the build. Format: id=mysecret,src=/local/secret")
+	cmd.Flags().BoolVarP(&options.BuildLocally, "local", "", false, "build the image on your local machine")
```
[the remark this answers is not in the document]
pchico83: remove this option?

pkg/cmd/build/build.go
```diff
@@ -27,64 +29,116 @@ import (
...
+	Path         string
+	Secrets      []string
+	Tag          string
+	Target       string
+}
+
 // Run runs the build sequence
-func Run(ctx context.Context, path, dockerFile, tag, target string, noCache bool, cacheFrom, buildArgs, secrets []string, progress string) error {
+func Run(ctx context.Context, namespace string, buildOptions BuildOptions) error {
+	if (!okteto.IsOktetoContext() && okteto.Context().Buildkit == "") || buildOptions.BuildLocally {
```
pchico83, line 48:
```suggestion
	if okteto.Context().Buildkit == "") {
```

pkg/cmd/stack/translate.go
```diff
@@ -149,11 +153,26 @@ func buildServices(ctx context.Context, s *model.Stack, options *StackDeployOpti
 		}
 		if !hasBuiltSomething {
 			hasBuiltSomething = true
-			log.Information("Running your build in %s...", okteto.Context().Buildkit)
+			if okteto.Context().Buildkit != "" {
+				log.Information("Running your build in %s...", okteto.Context().Buildkit)
+			} else {
+				log.Information("Running your build in your locally")
```
pchico83, line 159:
```suggestion
				log.Information("Running your build in docker")
```

pkg/okteto/context.go
```diff
@@ -406,6 +407,9 @@ func encodeOktetoKubeconfig(cfg *clientcmdapi.Config) string {
 	for name, cluster := range cfg.Clusters {
 		if name == currentCluster {
 			cluster.LocationOfOrigin = ""
+			if len(cluster.Extensions) > 0 {
+				cluster.Extensions = make(map[string]runtime.Object)
```
pchico83, line 411: is it possible to keep them? they might be needed to access the cluster. for example, telepresence adds an extension to specify on which namespace the telepresence server is running

pchico83 (approved):

PR #1739, opened 2021-09-03
Handle new/old Okteto API on pipeline/preview deploy/destroy commands

pchico83:
If `okteto pipeline desploy/destroy` returns a job ID, we wait for the job to finish if `--wait` is set.
Same for `okteto preview deploy`.

### DIFF
--- a/pkg/okteto/installer.go
+++ b/pkg/okteto/installer.go
@@ -1,0 +1,63 @@
+package okteto
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/okteto/okteto/pkg/log"
+)
+
+// InstallerJobBody top body answer
+type InstallerJobBody struct {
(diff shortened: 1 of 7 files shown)


pkg/okteto/installer.go
```diff
@@ -0,0 +1,63 @@
...
+}
+
+func WaitforInstallerJobToFinish(ctx context.Context, name, jobName, namespace string, timeout time.Duration) error {
+	t := time.NewTicker(1 * time.Second)
+	to := time.NewTicker(timeout)
+
+	for {
+		select {
+		case <-to.C:
+			return fmt.Errorf("pipeline '%s' didn't finish after %s", name, timeout.String())
```
jLopezbarb, line 46: This can also be a preview env

jLopezbarb (approved):

PR #1740, opened 2021-09-03
Skip windows actions tests

jLopezbarb:
Fixes e2e tests on windows

## Proposed changes
- Skip e2e actions tests on windows
-

### DIFF
--- a/integration/actions_test.go
+++ b/integration/actions_test.go
@@ -88,6 +88,10 @@ var (
 )
 
 func TestApplyPipeline(t *testing.T) {
+	if runtime.GOOS == "windows" {
+		t.Skip("this test is not required for windows e2e tests")
+		return
+	}
 	if mode == "client" {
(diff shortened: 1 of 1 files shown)


integration/actions_test.go
```diff
@@ -88,6 +88,10 @@ var (
 )
 
 func TestApplyPipeline(t *testing.T) {
+	if runtime.GOOS == "windows" {
```
pchico83, line 91: Could have a method `TestAllActions` that calls the other action tests?
That way we only need the check on that method

pchico83 (approved):

integration/actions_test.go
jLopezbarb: I don't think that's a good idea because when any of the actions fail, we will have all the logs together instead of each one its own logs.

pchico83: ok, good to go then!

PR #1741, opened 2021-09-03
Jlopezbarb/fix stignore

### LINKED ISSUE #1720: Single .stignore file is used for all sync folders
**Describe the bug**
When multiple sync folders are specified, each with a unique `.stignore` file, the `.stignore` file of the last sync folder specified in the manifest is used for all folders. 

**To Reproduce**

1. Create file `/path/to/folder1/.stignore` with some contents
2. Create file `/path/to/folder2/.stignore` with some contents that are different from those in step 1.
3. Create a manifest `okteto.yaml` that contains the following:
```
sync:
  - /path/to/folder1:/path/to/folder1
  - /path/to/folder2:/path/to/folder2
```
4. Run `okteto up` and note that in the pod, both `cat /path/to/folder1/.stignore` and `cat /path/to/folder2/.stignore` show the same contents: that of the local machine's `/path/to/folder2/.stignore`. Also, the contents of `/path/to/folder1` in the pod follow the filtering specified by `/path/to/folder2/.stignore` rather than `/path/to/folder1/.stignore`.
5. Optional: Edit `okteto.yaml` to switch the order of `folder1` and `folder2` in the `sync` list and note that the opposite situation occurs. (`/path/to/folder1/.stignore` is used for everything.)

**Expected behavior**
I expected both `/path/to/folder1/.stignore` and `/path/to/folder2/.stignore` to be copied to their respective locations in the pod and those appropriate filters would be applied to the appropriate folders.

- OS: Ubuntu 18.04
- Okteto Version 1.13.6

**Additional context**
None

jLopezbarb:
Fixes #1720

## Proposed changes
- Create different stignore files to not overwrite previous ones
-

### DIFF
--- a/pkg/k8s/apps/translate.go
+++ b/pkg/k8s/apps/translate.go
@@ -698,12 +698,20 @@ func TranslateOktetoDevSecret(spec *apiv1.PodSpec, secret string, secrets []mode
 			},
 		},
 	}
+	idx := 0
 	for i, s := range secrets {
+		key := s.GetKeyName()
+		path := s.GetFileName()
+		if strings.Contains(key, ".stignore") {
+			idx++
+			key = fmt.Sprintf("%s-%d", key, idx)
+			path = fmt.Sprintf("%s-%d", path, idx)
+		}
 		v.VolumeSource.Secret.Items = append(
 			v.VolumeSource.Secret.Items,
 			apiv1.KeyToPath{
-				Key:  s.GetKeyName(),
-				Path: s.GetFileName(),
+				Key:  key,
+				Path: path,
 				Mode: &secrets[i].Mode,
 			},
 		)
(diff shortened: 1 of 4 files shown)


pkg/k8s/apps/translate.go
```diff
@@ -698,12 +698,20 @@ func TranslateOktetoDevSecret(spec *apiv1.PodSpec, secret string, secrets []mode
...
 		},
 	}
+	idx := 0
 	for i, s := range secrets {
+		key := s.GetKeyName()
+		path := s.GetFileName()
+		if strings.Contains(key, ".stignore") {
+			idx++
+			key = fmt.Sprintf("%s-%d", key, idx)
+			path = fmt.Sprintf("%s-%d", path, idx)
```
pchico83, line 708: shouldn't the path be based on the destination mount path?

jLopezbarb: We create a new file based on the one provided by the user here:
https://github.com/okteto/okteto/blob/4bba2288cc517499bf7788357520a7ff585990d0/cmd/up/stignore.go#L50

jLopezbarb: That path is getting the name of the secret, which was the same for all the `.stignore` files and caused the error

pchico83:
@jLopezbarb I am working on removing the webhook dependency, I would close the PR until that is ready (probably next monday)

pchico83:
@jLopezbarb now we have removed the dependency on the mutation webhook, you can resume the work on this one

PR #1749, opened 2021-09-08
Add latest release to all actions

jLopezbarb:

Adds latest release to all actions and updates it with every CLI release

### DIFF
--- a/update_actions.sh
+++ b/update_actions.sh
@@ -38,6 +38,7 @@ for repo in "${actionsRepos[@]}"; do
                 git --no-pager log -1
         fi
         ghr -token "$GITHUB_TOKEN" -replace "$VERSION"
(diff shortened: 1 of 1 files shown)


pchico83:
Talked offline with @jLopezbarb . `latest` is already generated automatically

pchico83 (approved):

rberrelleza:
give how we create a release on every CLI release automatically, what is the difference between using `latest` vs directly using `master`? The latest release is still going to be a moving target no, so it's not really pinning a version.

pchico83:
@rberrelleza `master` means the current state in the action repo. `latest` is the latest release. And we could switch latest to a different version if something is broken in the latest release

pchico83 (approved):

Issue #1687, opened 2021-08-06
docker-compose with extensions fails to be deployed


rberrelleza:
**Describe the bug**
If a compose manifest has extensions, our parsing fails with an error, instead of telling the user that we don't support the field.

**To Reproduce**

1. Deploy https://github.com/NYULibraries/library-guest-passes/blob/main/docker-compose.yml

**Expected behavior**
A user friendly error that we don't support extensions.

PR #1756, opened 2021-09-13
Add extension to compose

### LINKED ISSUE #1755: Support for docker compose extensions ("x-nnnn" fields)
My compose file has an extension like this:
```
x-env: &env
  environment:
    - SOME_ENV_VAR=123
  command: bash -c 'echo $$SOME_ENV_VAR'

services:
  first:
    <<: *env
    image: bash
  second:
    <<: *env
    image: bash  
```

and it fails with:
```
➜ okteto stack deploy
 x  Invalid stack manifest:
    - line 1: field x-env not found in type model.StackRaw
```

It works, although the `x-nnnn` fields will be ignored, as indicated in the [Compose Spec](https://github.com/compose-spec/compose-spec/blob/master/spec.md#extension).

jLopezbarb:

Fixes #1755, #1687

## Proposed changes
- Add support for extension
- Fail if a field does not start with 'x-'

### DIFF
--- a/pkg/model/stack_serializer.go
+++ b/pkg/model/stack_serializer.go
@@ -1444,7 +1444,9 @@ func (a *ArgsStack) UnmarshalYAML(unmarshal func(interface{}) error) error {
 func validateExtensions(stack StackRaw) error {
 	nonValidFields := make([]string, 0)
 	for extension := range stack.Extensions {
-		nonValidFields = append(nonValidFields, extension)
+		if !strings.HasPrefix(extension, "x-") {
+			nonValidFields = append(nonValidFields, extension)
+		}
 	}
 
 	for svcName, svc := range stack.Services {
--- a/pkg/model/stack_serializer.go
+++ b/pkg/model/stack_serializer.go
@@ -39,7 +39,6 @@ type StackRaw struct {
 	Volumes   map[string]*VolumeTopLevel `yaml:"volumes,omitempty"`
 
 	// Extensions
-
 	Extensions map[string]interface{} `yaml:",inline" json:"-"`
 
 	// Docker-compose not implemented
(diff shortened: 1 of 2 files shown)


rberrelleza:
What happens with the extension fields in okteto stacks? Are we just parsing them, or will they impact the deployment somehow?

jLopezbarb:
We are just parsing them for all the users that uses 'x-nnn' fields for yaml expansion which we were supporting but using our fields as the key. This allows users to have their own keys and a more structured file.

pkg/model/stack_serializer.go
```diff
@@ -38,6 +38,10 @@ type StackRaw struct {
 	Endpoints EndpointSpec               `yaml:"endpoints,omitempty"`
 	Volumes   map[string]*VolumeTopLevel `yaml:"volumes,omitempty"`
 
+	// Extensions
+
+	Extensions map[string]interface{} `yaml:",inline" json:"-"`
```
pchico83, line 43: If I get it correctly, this will parse anything not parsed by any other field.
Could you compare the output we are giving for current not supported fields with the output generated by this version? Right now we are using strict yaml parsing. To be honest, our current parsing errors are quite hard to understand, maybe we can do it better with this approach

jLopezbarb: The only problem I see with the new errors management is that we can not show the line where the error is. But not showing model.StackRaw it's better because we are not giving any clue of what the error is
Previous version error:
 ```
x  Invalid stack manifest:
    - line 3: field random-field not found in type model.StackRaw
    See https://okteto.com/docs/reference/stacks/ for details
```

New error:
```
 x  random-field is not supported. More information is available here: https://okteto.com/docs/reference/stacks/
```

Comment (pchico83) [reply pkg/model/stack_serializer.go:43]: @rberrelleza @melissa-okteto what do you think about the new error message for syntax errors in the okteto stack manifests?

Comment (rberrelleza) [reply pkg/model/stack_serializer.go:43]: how does this compare with the errors you get in `docker compose up`?

Comment (jLopezbarb) [reply pkg/model/stack_serializer.go:43]: On compose the error they give is not very informative either:
```
ERROR: The Compose file './docker-compose.yml' is invalid because:
'random-field' does not match any of the regexes: '^x-'
```

jLopezbarb:
I have implemented this error for all the other fields  so we can unify all errors in one:
This is how it looks if only one field is supported
Captura de pantalla 2021-09-15 a las 13 51 58
This is how it looks if more than one field is unsupported
Captura de pantalla 2021-09-15 a las 13 53 04

pchico83 (approved):

PR #1757, opened 2021-09-13
Add context command

jLopezbarb:

Adds context command.

Context comes to replace okteto login.

## Proposed changes
- It will ask you to run okteto context the first time you run okteto 
- It will create a new kubeconfig file  which will be the one used by okteto. This way you can set a context for okteto without downloading credentials and operate with your usual context using `kubectl`. This prevent to run okteto commands on another cluster (like prod cluster)


cmd/kubeconfig.go
```diff
@@ -0,0 +1,69 @@
...
+// Kubeconfig fetch credentials for a cluster namespace
+func Kubeconfig(ctx context.Context) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "kubeconfig",
+		Short: "Downloads k8s credentials from current okteto context",
+		Args:  utils.NoArgsAccepted("https://okteto.com/docs/reference/cli/#kubeconfig"),
+		RunE: func(cmd *cobra.Command, args []string) error {
+
+			namespace := ""
+			if len(args) > 0 {
```
pchico83, line 36: I wouldn't accept a namespace here, but at the context configuration level.

cmd/context/context.go
```diff
@@ -0,0 +1,136 @@
...
+		RunE: func(cmd *cobra.Command, args []string) error {
+			ctx := context.Background()
+			if ctxOptions.Token == "" && k8Client.InCluster() {
+				return fmt.Errorf("this command is not supported without the '--token' flag from inside a pod")
+			}
+
+			var err error
+			if len(args) == 0 {
+				log.Infof("authenticating without context")
+				err = runInteractiveContext(ctx, ctxOptions)
```
pchico83, line 73: is it implemented support for `OKTETO_URL` in `okteto context`?

cmd/namespace/namespace.go
```diff
@@ -44,6 +44,10 @@ func Namespace(ctx context.Context) *cobra.Command {
 				return err
 			}
 
+			if !okteto.IsOktetoCluster() {
+				return errors.ErrNotOktetoCluster
+			}
+
 			err := RunNamespace(ctx, namespace)
```
pchico83, line 51: This is also deprecated by `okteto kubeconfig`. We could also call the `Kubeconfig` logic from here

cmd/up/up.go
```diff
@@ -221,12 +221,12 @@ func (up *upContext) start(autoDeploy, build bool) error {
 	var err error
 	up.Client, up.RestConfig, err = k8sClient.GetLocalWithContext(up.Dev.Context)
 	if err != nil {
-		kubecfg := config.GetKubeConfigFile()
-		log.Infof("failed to load local Kubeconfig: %s", err)
+		kubecfg := config.GetContextKubeconfigPath()
+		log.Infof("failed to load okteto Kubeconfig: %s", err)
 		if up.Dev.Context == "" {
```
pchico83, line 226: if `up.Dev.Context` is defined, we are always using that context, right?

pkg/analytics/analytics.go
```diff
@@ -60,6 +60,7 @@ const (
 	previewDestroyEvent      = "DestroyPreview"
 	execEvent                = "Exec"
 	signupEvent              = "Signup"
+	setContextEvent          = "SetContext"
```
pchico83, line 63:
```suggestion
	contextEvent          = "Context"
```
We are using the command name for other metrics

pkg/analytics/analytics.go
```diff
@@ -283,6 +284,17 @@ func TrackSignup(success bool, userID string) {
 	track(signupEvent, success, nil)
 }
 
+// TrackContext sends a tracking event to mixpanel when the user use context in
+func TrackContextLogin(success, isOktetoCluster bool) {
```
pchico83, line 288:
```suggestion
func TrackContext(success, isOktetoCluster bool) {
```

pkg/errors/errors.go
```diff
@@ -46,7 +46,10 @@ var (
 	ErrCommandFailed = errors.New("Command execution failed")
 
 	// ErrNotLogged is raised when we can't get the user token
-	ErrNotLogged = fmt.Errorf("please run 'okteto login [URL]' and try again")
+	ErrNotLogged = fmt.Errorf("Context is not set. Please run 'okteto ctx' and try again")
```
pchico83, line 49:
```suggestion
	ErrNotLogged = fmt.Errorf("Okteto context isn't configured. Please run 'okteto context' and try again")
```

pkg/errors/errors.go
```diff
@@ -46,7 +46,10 @@ var (
...
+
+	// ErrNotOktetoCluster is raised when we a command is only available on an okteto cluster
+	ErrNotOktetoCluster = fmt.Errorf("User is not logged on okteto cluster. Please run 'okteto ctx' and try again")
```
pchico83, line 52:
```suggestion
	ErrNotOktetoCluster = fmt.Errorf("User is not logged on okteto cluster. Please run 'okteto context' and try again")
```

pkg/errors/errors.go
```diff
@@ -86,6 +89,9 @@ var (
 
 	//ErrDivertNotSupported raised if the divert feature is not supported in the current cluster
 	ErrDivertNotSupported = fmt.Errorf("the 'divert' field is only supported in namespaces managed by Okteto")
+
+	//ContextIsNotOktetoCluster raised if the cluster connected is not managed by okteto
+	ContextIsNotOktetoCluster = fmt.Errorf("okteto context is not managed by okteto")
```
pchico83, line 94: What's the difference with?
https://github.com/okteto/okteto/pull/1757/files#diff-ab01ccb939ab896a48c5445d1d755bf861ad2f337d98013132834c6d88641c8aR52

pkg/k8s/namespaces/crud.go
```diff
@@ -53,6 +53,16 @@ func Get(ctx context.Context, ns string, c *kubernetes.Clientset) (*apiv1.Namesp
 	return n, nil
 }
 
+// List returns the namespace object of ns
+func List(ctx context.Context, c *kubernetes.Clientset) ([]apiv1.Namespace, error) {
```
pchico83, line 57: Why do we need this method? I think we are not listing namespaces anywhere

pkg/log/log.go
```diff
@@ -178,6 +178,11 @@ func BlueString(format string, args ...interface{}) string {
 	return blueString(format, args...)
 }
 
+// YellowString returns a string in blue
+func YellowString(format string, args ...interface{}) string {
```
pchico83, line 182: is this the same as `log.Warning`?

pkg/okteto/auth.go
```diff
@@ -39,13 +39,14 @@ var reg = regexp.MustCompile("[^A-Za-z0-9]+")
...
-	Token     string `json:"Token"`
-	MachineID string `json:"MachineID"`
+	URL             string `json:"URL"`
+	Buildkit        string `json:"Buildkit"`
+	Registry        string `json:"Registry"`
+	ID              string `json:"ID"`
+	Username        string `json:"Username"`
+	Token           string `json:"Token"`
+	MachineID       string `json:"MachineID"`
+	IsOktetoCluster bool   `json:"IsOktetoCluster"`
```
pchico83, line 49: could we infer this from `URL`?

cmd/context/context.go
```diff
@@ -0,0 +1,136 @@
...
+	"github.com/okteto/okteto/pkg/analytics"
+	okContext "github.com/okteto/okteto/pkg/cmd/context"
+	k8Client "github.com/okteto/okteto/pkg/k8s/client"
+	"github.com/okteto/okteto/pkg/log"
+	"github.com/spf13/cobra"
+)
+
+type ContextOptions struct {
+	Token           string
+	isOktetoCluster bool
```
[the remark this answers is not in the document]
rberrelleza: Keep in mind that something like "localhost" or "foo" are valid DNS values as well as a valid URLs, and it can be used as okteto endpoint. So we might want to consider a way for the user to specify if the value passed is a URL or a kubernetes context.

cmd/context/context.go
```diff
@@ -0,0 +1,136 @@
...
+		Use:     "context [url|k8s-context]",
+		Aliases: []string{"ctx"},
+		Args:    utils.MaximumNArgsAccepted(1, "https://okteto.com/docs/reference/cli/#context"),
+		Short:   "Points okteto to a cluster",
+		Long: `Points okteto to a cluster
+
+Run
+    $ okteto context
+
+and this command will ask you for the cluster you want to operate in okteto.
```
rberrelleza, line 47: what do we mean by "the cluster you want to operate"?  This sounds more like a "cluster-management" operation.  Also, this is not asking for the cluster, but an existing kubernetes context isn't it?

cmd/context/context.go
```diff
@@ -0,0 +1,136 @@
...
+	ctxOptions := &ContextOptions{}
+	cmd := &cobra.Command{
+		Use:     "context [url|k8s-context]",
+		Aliases: []string{"ctx"},
+		Args:    utils.MaximumNArgsAccepted(1, "https://okteto.com/docs/reference/cli/#context"),
+		Short:   "Points okteto to a cluster",
+		Long: `Points okteto to a cluster
+
+Run
+    $ okteto context
```
rberrelleza, line 45: `login` is a very common action in CLIs. pulumi, aws, docker, heroku, etc have a login command. I personally find very confusing that we're mixing 'login to your okteto cluster'  with the well-known concept of 'use a kubernetes context' for the operation. 

Have we considered using a different term that doesn't already have a meaning in the world of kubernetes?

rberrelleza: maybe login still exists as a command (only usable in okteto clusters), and it sets the context to the one that we create?  That way,  developers who don't use okteto cloud/enterprise (which is like a third of our user base I think), can still use this to set the kubernetes context?

rberrelleza: another option could be to make this only work for okteto clusters. if you're on a vanilla kube cluster, you use standard  tooling for this (e.g. kubie, kubectl config set-conext, etc..)

cmd/kubeconfig.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/okteto/okteto/pkg/cmd/login"
+	"github.com/okteto/okteto/pkg/config"
+	"github.com/okteto/okteto/pkg/okteto"
+	"github.com/spf13/cobra"
+)
+
+// Kubeconfig fetch credentials for a cluster namespace
+func Kubeconfig(ctx context.Context) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "kubeconfig",
```
rberrelleza, line 30: what's the objective of this command standalone? won't you want this to happen every time you login or set okteto cloud/enterprise as your context?

cmd/kubeconfig.go
```diff
@@ -0,0 +1,69 @@
...
+		Short: "Downloads k8s credentials from current okteto context",
```
rberrelleza, line 31:
```suggestion
		Short: "Downloads the k8s credentials of the current okteto context",
```

cmd/login.go
```diff
@@ -47,68 +41,15 @@ By default, this will log into cloud.okteto.com. If you want to log into your Ok
...
-				analytics.TrackSignup(true, u.ID)
-			}
-			analytics.TrackLogin(true, u.Name, u.Email, u.ID, u.ExternalID)
-			return nil
+			log.Warning(`The 'login' command is deprecated and will be removed in a future release.
+    The 'login' command has been replaced by 'context' command.
+    More information is available here: https://okteto.com/docs/reference/cli#context`)
+			okCtx := contextCMD.Context()
+			okCtx.Flags().Set("token", token)
+			return okCtx.RunE(cmd, args)
```
rberrelleza, line 49: will okteto context take the exact same parameters and arguments than okteto login? if not, we might need to keep it for a while, and do this change on a major release.

cmd/login.go
```diff
@@ -47,68 +41,15 @@ By default, this will log into cloud.okteto.com. If you want to log into your Ok
...
-			} else {
-				log.Success("Logged in as %s @ %s", u.ExternalID, oktetoURL)
-			}
-			log.Hint("    Run `okteto namespace` to switch your context and download your Kubernetes credentials.")
-			if u.New {
-				analytics.TrackSignup(true, u.ID)
-			}
-			analytics.TrackLogin(true, u.Name, u.Email, u.ID, u.ExternalID)
-			return nil
+			log.Warning(`The 'login' command is deprecated and will be removed in a future release.
```
rberrelleza, line 44: let's first release a version where this is silently changed (assuming the command will be compatible) and then add the deprecation after a version or to. I think it's very aggressive to show a warning so soon.

cmd/context/context.go
```diff
@@ -0,0 +1,136 @@
...
+
+type ContextOptions struct {
+	Token           string
+	isOktetoCluster bool
+}
+
+// Context points okteto to a cluster.
+func Context() *cobra.Command {
+	ctxOptions := &ContextOptions{}
+	cmd := &cobra.Command{
```
rberrelleza, line 37: this is a big change in how you use okteto, customer education, samples, github actions, etc. Can we first introduce this as a hidden command, test it ourselves for a few versions, show it to some users, get feedback, and then make the big change?

cmd/context/context.go
```diff
@@ -0,0 +1,136 @@
...
+			}
+
+			if err != nil {
+				analytics.TrackContextLogin(false, ctxOptions.isOktetoCluster)
+				analytics.TrackLogin(false, "", "", "", "")
+				return err
+			}
+
+			analytics.TrackContextLogin(true, ctxOptions.isOktetoCluster)
+			log.Success("Your context have been updated")
```
rberrelleza, line 87: what does this mean? are we updating the current kube context? or something else?

cmd/context/utils.go
```diff
@@ -0,0 +1,215 @@
...
+		clusterOptions = append(clusterOptions, SelectItem{
+			Name:   k8sCluster,
+			Enable: true,
+		})
+	}
+	return clusterOptions
+}
+
+func isOptionAnOktetoCluster(option string) bool {
+	return option == "Okteto cloud" || option == "Okteto enterprise" || isURL(option)
```
rberrelleza, line 182: let's make this check case insensitive?

cmd/context/utils.go
```diff
@@ -0,0 +1,215 @@
...
+
+func getOktetoClusterUrl(ctx context.Context, option string) string {
+	if option == "Okteto cloud" {
+		return okteto.CloudURL
+	}
+
+	return AskForLoginURL(ctx)
+}
+
+func isURL(okContext string) bool {
```
rberrelleza, line 193: pretty much anything can be a valid URL (e.g "localhost" is a valid URL). If what we need to know is if "this URL belongs to an okteto instance" I think we should do something more robust. maybe there's an 'info' API we can use for this?

rberrelleza:
what's the experience going to be if I only have 1 cluster in my context, and I run 'okteto up'? do I still need to go through this flow?

jLopezbarb:
Yes, you will need to go through this command because the context that you have may not be an okteto cluster, and if you want to test okteto on our cloud, setting the url for a newcomer may not be very intuitive

cmd/context/context.go
jLopezbarb: This method comes to set okteto to always point to the same k8s context allowing you to have another one used for your production cluster. Using `kubectl config set-conext` will set their context on their local kubeconfig but not on okteto

cmd/kubeconfig.go
jLopezbarb: We are storing the context the user is using in a different file than your local kubeconfig. This command retrieves that kubernetes context and set it in your local kubeconfig

cmd/up/up.go
jLopezbarb: if `up.Dev.Context` is defined and you have that context on your `.context/kubeconfig` file it will use that value. This means that you have to run at least one `okteto ctx` against that cluster to configure it and be able to use it

cmd/context/context.go
```diff
@@ -0,0 +1,136 @@
...
+If you want to log into your own cluster instance, specify a kubernetes context. For example, run
+
+    $ okteto context my_cluster
+
+to point okteto to 'mycluster'.
+`,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			ctx := context.Background()
+			if ctxOptions.Token == "" && k8Client.InCluster() {
+				return fmt.Errorf("this command is not supported without the '--token' flag from inside a pod")
```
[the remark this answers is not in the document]
jLopezbarb: yes it is ready to run `okteto ctx minikube` as long as the k8s context is defined on the local kubeconfig

cmd/up/up.go
pchico83: if `up.Dev.Context` is defined, and the context accessible, I would use that context even if you never did, for example, `okteto context minikube`

cmd/context/context.go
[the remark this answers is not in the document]
rlamana: We might want to add a way to specify it for those corner cases where a URL === kubernetes context name, but the default way should be simple and it can be inferred for most cases. Are we taking the kubernetes context as the preference? If a context `foo` doesn't exist, it's treated as a URL, for example.

cmd/context/context.go
rlamana: As we mention in the spec, it "points okteto to a cluster". Whether is Okteto Cloud, an instance of Okteto Enterprise, or **maybe** a Kubernetes context/namespace defined in the local `.kubeconfig` file

cmd/context/context.go
rlamana: I think that making the CLI work the same way in an Okteto Instance or your own cluster is key here. Otherwise, the developer will have to think: Am I using an Okteto instance? Then I have to do `login` too, but when I work with my own cluster I don't... 

The idea behind this is that the developer only has to type **one command** to point to the desired cluster, no matter if it is their own or an Okteto instance.

cmd/context/context.go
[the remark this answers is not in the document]
rberrelleza: yeah, I think that's a good approach @rlamana .

cmd/context/context.go
rberrelleza: @rlamana could you link to the spec?

cmd/context/context.go
rberrelleza: > I think that making the CLI work the same way in an Okteto Instance or your own cluster is key here

cmd/context/context.go
rberrelleza: My comment here is on the text we display. Asking for "a cluster" is different from asking from "a context name" vs "a URL to the okteto instance". If someone asks me for a cluster, my first reaction would be to put the cluster name, or maybe the API. 

Given that this is a multi-shape field (it can be a context name or the URL of an okteto instance), I think we need to be very careful of the terms we use to avoid confusion.

cmd/context/context.go
rberrelleza: > the developer will have to think: Am I using an Okteto instance?

Running `okteto context $URL` still involves that decision no?  You have to know you're on okteto cloud/enterprise AND the URL to provide. 

One thing to consider is that standard for kubernetes tooling directly works with your current kubernetes context without you requiring any extra commands (helm and kubectl works like that). So non-cloud users need to run two commands now (okteto context + okteto up) instead of just one.  This might confuse the non-cloud users.

rberrelleza:
how does `okteto namespace` fit in this new world? Will it change the name in the okteto context AND the kube context?

cmd/context/context.go
rlamana: Here: https://github.com/okteto/wiki/blob/main/specs/okteto-dev.md#context

cmd/context/context.go
```diff
@@ -0,0 +1,155 @@
...
+    $ okteto context
+
+and this command this will ask you for the name of your kubernetes context.
+If an okteto cluster is selected it will open your browser to ask your authentication details and retrieve your API token. You can script it by using the --token parameter.
+
+By default, this will ask you for the kubernetes context you want to operate.
+
+If you want to log into your Okteto Enterprise instance, specify a URL. For example, run
+
+    $ okteto context https://okteto.example.com
```
jmacelroy, line 57: I may be saying the same thing as Ramiro but I am finding this command a little confusing. This may be due to the overloaded term `context` and that there are multiple types of parameters that I think won't work in cases I hit commonly. 

If my current `kube context` is `cluster-a` and I have a namespace `jacob`. What happens when I run `kube context jacob` and I wanted it to be `cluster-b` with namespace `jacob`? We can't guarantee uniqueness and I imagine I'll mess it up?

cmd/context/context.go
```diff
@@ -0,0 +1,155 @@
...
+
+// Context points okteto to a cluster.
+func Context() *cobra.Command {
+	ctxOptions := &ContextOptions{}
+	cmd := &cobra.Command{
+		Use:     "context [url|k8s-context]",
+		Aliases: []string{"ctx"},
+		Args:    utils.MaximumNArgsAccepted(1, "https://okteto.com/docs/reference/cli/#context"),
+		Short:   "Points okteto to a cluster",
+		Long: `Points okteto to a cluster
```
jmacelroy, line 45: I think `Points okteto to a cluster` may be too vague. I think the vagueness may be due to the overloaded nature of the command. 

What does it mean to point to a cluster? If I don't have kubectl configuration setup then I need that first. If I do then I just need to use the "name" from the kubectl configuration, which may have been change out of band?

cmd/context/context.go
pchico83: In your example, do you mean `kube context`, or `okteto context`?
In other CLI tools in the space, like kubectl or docker, `context` sets your working cluster. Namespaces are a secondary attribute.

cmd/context/context.go
pchico83: This is the definition of a kubernetes context, which mostly apply to an okteto context too:
```
A kubernetes context is just a set of access parameters that contains a Kubernetes cluster, a user, and a namespace. kubernetes Context is essentially the configuration that you use to access a particular cluster & namespace with a user account.
```
I would go with something like:
```
Short: "Manage your okteto context"
Long: "Manage your okteto context

Explain what is an okteto context. Explain how it works when it is based on a kubernetes context, and explain how it works when it is based on an Okteto cluster
```

cmd/context/context.go
jmacelroy: I mean okteto context. As far as I can tell an okteto context = cluster+namespace and but it seems like the namespace is used to define a context and I'm wondering about possible collisions?

PR #1761, opened 2021-09-14
Fix wrong init container added for jobs without volumes

pchico83:

Looks like if a job doesn't have volumes, we are injecting a failing init container:
```
      initContainers:
      - command:
        - sh
        - -c
        - ""
        image: busybox
        imagePullPolicy: Always
        name: init-seed
        resources: {}
        terminationMessagePath: /dev/termination-log
        terminationMessagePolicy: File
```

### DIFF
--- a/pkg/cmd/stack/translate_test.go
+++ b/pkg/cmd/stack/translate_test.go
@@ -433,7 +433,7 @@ func Test_translateStatefulSet(t *testing.T) {
 	}
 }
 
-func Test_translateJob(t *testing.T) {
+func Test_translateJobWithoutVolumes(t *testing.T) {
 	s := &model.Stack{
 		Name: "stackName",
 		Services: map[string]*model.Service{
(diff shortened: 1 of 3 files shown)


pkg/cmd/stack/translate_test.go
```diff
@@ -466,7 +466,6 @@ func Test_translateJob(t *testing.T) {
 				CapDrop:       []apiv1.Capability{apiv1.Capability("CAP_DROP")},
 				RestartPolicy: apiv1.RestartPolicyNever,
 				BackOffLimit:  5,
-				Volumes:       []model.StackVolume{{RemotePath: "/volume1"}, {RemotePath: "/volume2"}},
```
rberrelleza, line 469: can we add a test for the two variants? (init container with and without volume)

rberrelleza (approved):

PR #1764, opened 2021-09-19
Add functionality to set runAsNonRoot to disable securityContext defaults

### LINKED ISSUE #1658: Feature Request: Add the ability to disable the security context

Provide a mechanism to disable security context in order support OpenShift non-root containers .

**Additional context**

By default OpenShift limits containers to run as a non-root user that has a randomly generated UID. 

This UID falls within a set range that is allocated to the project.  You can find this range by running the following command:

```
oc describe project <project name>

....
Annotations:    openshift.io/description=
                        openshift.io/display-name=
                        openshift.io/requester=developer
                        openshift.io/sa.scc.mcs=s0:c25,c10
                        openshift.io/sa.scc.supplemental-groups=1000620000/10000
                        openshift.io/sa.scc.uid-range=1000620000/10000
```
I really like Okteto and I want to be able to use it with OpenShift 4, however this means that we have to explicitly set the `runAsUser` parameter to a UID in these ranges. This is painful in a namespace per developer setup, especially when storing the manifest in Git.

It would be awesome if there was a way we could tell the tool to disable the security context  as Okteto currently defaults to the root user. 

I firmly believe in the value this project provides, and I wouldn't mind having a crack at contributing if this is something that you guys don't have time to look at.

**A potential solution**

A lot of helm charts allow you to disable pod and container security contexts. This could be used to resolve this problem as not setting them would allow OpenShift to automatically set them.

The Okteto manifest could look like this:

```
...
securityContext:
    enabled: false
```
This would prevent the security context from being set.

It does not really follow the Kubernetes specification closely though.

mrpresidentjk:
Fixes #1658

## Proposed changes

Expose the `runAsNonRoot` option to the `securityContext` configuration of the manifest.

If this is set to true, Okteto will not set the defaults for `fsGroups`, `runAsUser`, `runAsGroup` instead delegating this to be by the container platform.

OpenShift automatically sets `fsGroup`, `runAsUser` and `runAsGroup` to a randomly generated UID if they are not given. By default, the UID is chosen from a range unique to the OpenShift project/namespace. Admins can also change this behaviour via [securityContextConstraints](https://docs.openshift.com/container-platform/3.11/admin_guide/manage_scc.html). 

Okteto currently defaults to root unless specified which does not play well in a project/namespace per developer setup where manifests are stored in Git. You have to hard code the `fsGroups`, `runAsUser`, and `runAsGroup` fields with a valid user, which can be different across projects/namespaces. 

With this change, you could use the following manifest for development with non-root containers in OpenShift:

 ```
name: my-service
labels:
  app: my-service
workdir: /usr/src/app
securityContext:
  runAsNonRoot: true
remote: 2222
sync:
 - ./:/usr/src/app
 ```

Additionally, Kubernetes can default these variables for you based off the user in the docker image. This means that you could theoretically have your init and dev container image run as a specific UID or GID, and have Okteto run as that user without having to define it in your manifest.

By default, the option won't be set and Okteto will behave the same way as it did before this change.

To see how this change works, consider the following the manifests:

1. `runAsNonRoot=true` with no other `securityContext` overrides:
    ```
      name: non-root-user-without-overrides
      securityContext:
          runAsNonRoot: true
    ```
      In this case the container is run with `runAsNonRoot=true`, and the variables `runAsUser`, `runAsGroup`, and `fsGroup` are left to be set by the container platform. It's possible that container will not start correctly due to permissions though this a user configuration error that can only be detected at run time.
 
 2. `runAsNonRoot=false` with no other `securityContext` overrides:
     ```
        name: root-user-without-overrides
        securityContext:
           runAsNonRoot: false
    ```
    In this case Okteto will default to the root user.

    
3.  `runAsNonRoot=true` with `securityContext` overrides:
    ```
    name: non-root-user-with-overrides
    securityContext:
       runAsUser: 100
       runAsGroup: 101
       fsGroup: 102
       runAsNonRoot: true
    ```
    In this case Okteto will still set the values for `fsGroups`, `runAsUser`, `runAsGroup` from the manifest, and run the container with `runAsNonRoot=true`.

4.  `runAsNonRoot=false` with `securityContext` overrides:
    ```
     name: root-user-with-overrides
      securityContext:
         runAsUser: 100
         runAsGroup: 101
         fsGroup: 102
         runAsNonRoot: false
    ```
    This is actually a valid configuration and Okteto will still set the values for `fsGroups`, `runAsUser`, `runAsGroup` from the manifest while running with `runAsNonRoot=false` .

If you try to explicitly set `runAsNonRoot=true` and use the root user, you will get an error on `okteto up`.

As a side note, you will notice that I have also made it so you can set `runAsNonRoot=true`  and `runAsGroup=0`. This is also valid, and it is how all containers in OpenShift run by default.

It is my first time contributing to this project, so please let me know if there is anything I need to consider/fix. Let me know if you do not agree with this change, I am always open to feedback.

### DIFF
--- a/pkg/k8s/apps/translate.go
+++ b/pkg/k8s/apps/translate.go
@@ -50,7 +50,6 @@ const (
 var (
 	devReplicas                      int32 = 1
 	devTerminationGracePeriodSeconds int64
-	falseBoolean                     = false
 )
 
 func translate(t *model.Translation, c *kubernetes.Clientset, isOktetoNamespace bool) error {
@@ -532,16 +531,14 @@ func TranslateContainerSecurityContext(c *apiv1.Container, s *model.SecurityCont
 
 	if s.RunAsUser != nil {
 		c.SecurityContext.RunAsUser = s.RunAsUser
-		if *s.RunAsUser == 0 {
-			c.SecurityContext.RunAsNonRoot = &falseBoolean
-		}
 	}
 
 	if s.RunAsGroup != nil {
 		c.SecurityContext.RunAsGroup = s.RunAsGroup
-		if *s.RunAsGroup == 0 {
-			c.SecurityContext.RunAsNonRoot = &falseBoolean
-		}
+	}
+
+	if s.RunAsNonRoot != nil {
+		c.SecurityContext.RunAsNonRoot = s.RunAsNonRoot
 	}
 
 	if s.Capabilities == nil {
--- a/pkg/k8s/apps/translate_test.go
+++ b/pkg/k8s/apps/translate_test.go
@@ -463,9 +463,8 @@ services:
 							Command:         []string{"./run_worker.sh"},
 							Args:            []string{},
 							SecurityContext: &apiv1.SecurityContext{
-								RunAsUser:    &rootUser,
-								RunAsGroup:   &rootUser,
-								RunAsNonRoot: &falseBoolean,
+								RunAsUser:  &rootUser,
+								RunAsGroup: &rootUser,
 							},
 							VolumeMounts: []apiv1.VolumeMount{
 								{
@@ -748,9 +747,8 @@ docker:
 							ImagePullPolicy: apiv1.PullIfNotPresent,
 							Command:         []string{"sh", "-c", "cp /usr/local/bin/* /okteto/bin"},
 							SecurityContext: &apiv1.SecurityContext{
-								RunAsUser:    &rootUser,
-								RunAsGroup:   &rootUser,
-								RunAsNonRoot: pointer.BoolPtr(false),
+								RunAsUser:  &rootUser,
+								RunAsGroup: &rootUser,
 							},
 							VolumeMounts: []apiv1.VolumeMount{
 								{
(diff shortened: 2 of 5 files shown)


pchico83 (approved):
@mrpresidentjk awesome PR, explanation and code coverage :-)
Thanks for this great contribution to okteto!

pkg/k8s/apps/translate.go
```diff
@@ -532,16 +532,14 @@ func TranslateContainerSecurityContext(c *apiv1.Container, s *model.SecurityCont
 
 	if s.RunAsUser != nil {
 		c.SecurityContext.RunAsUser = s.RunAsUser
-		if *s.RunAsUser == 0 {
-			c.SecurityContext.RunAsNonRoot = &falseBoolean
```
pchico83, line 536: I think you can remove the `falseBoolean` variable:
https://deepsource.io/gh/okteto/okteto/run/69a9ca14-39ce-4acf-82a8-f5c59085ba3c/go/SCC-U1000

mrpresidentjk: Thanks for approval! I have just updated this with the variable removed and an extra test case for coverage. I will try find some time tomorrow to fix the last coverage warning.

pchico83:
@mrpresidentjk merging like it is. we are working on a refactor around this code and I don't want you to deal with merging hell :-)

PR #1765, opened 2021-09-19
Refactor to use App interface

pchico83:
RefactorK8sObject into an App interface. Code cleanup


cmd/push.go
```diff
@@ -155,46 +153,42 @@ func runPush(ctx context.Context, dev *model.Dev, autoDeploy bool, imageTag, okt
...
 	for _, tr := range trList {
-		if tr.K8sObject == nil {
-			continue
-		}
-
 		if len(dev.Services) == 0 {
-			if tr.K8sObject.GetAnnotation(model.OktetoAutoCreateAnnotation) == model.OktetoUpCmd || tr.K8sObject.PodTemplateSpec.Spec.Containers[0].Name == "dev" {
-				tr.K8sObject.SetAnnotation(model.OktetoAutoCreateAnnotation, model.OktetoPushCmd)
+			if tr.App.Annotations()[model.OktetoAutoCreateAnnotation] == model.OktetoUpCmd || tr.App.PodSpec().Containers[0].Name == "dev" {
+				tr.App.Annotations()[model.OktetoAutoCreateAnnotation] = model.OktetoPushCmd
```
ifbyol, line 164: Minor thing: This syntax to set the value of an annotation could be a bit confusing. Probably is for coming from Java, but if I see `App.Annotations()` I would expect to have a local copy of the annotations and adding a new annotation wouldn't modify the original annotations map.

I see more intuitive something like:

```
// Convert function into a property
tr.App.Annotations[model.OktetoAutoCreateAnnotation] = model.OktetoPushCmd
``` 

or 

```
tr.App.SetAnnotation(model.OktetoAutoCreateAnnotation, model.OktetoPushCmd)
```

Anyways, just a minor comment, not a strong opinion, you can leave it as it is

ifbyol (approved):

cmd/push.go
pchico83: Great point!
I want to do another pass to the code and do final testing, so I will go with the `SetAnnotation` and `GetAnnotation` approach, I agree it is clearer

pkg/k8s/apps/interface.go
```diff
@@ -0,0 +1,69 @@
...
+import (
+	"context"
+
+	"github.com/okteto/okteto/pkg/model"
+	appsv1 "k8s.io/api/apps/v1"
+	apiv1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+type App interface {
```
jmacelroy, line 25: This is a fairly large interface. Can it be broken into smaller interfaces and possibly combined into smaller interfaces is my thought. I need to look more at the various use cases and implementations but it seems like we through all requirements into the interface and we may be able to leave some functionality out completely and be more flexible.

pkg/k8s/apps/deployments.go
```diff
@@ -0,0 +1,268 @@
...
+	"github.com/okteto/okteto/pkg/log"
+	"github.com/okteto/okteto/pkg/model"
+	appsv1 "k8s.io/api/apps/v1"
+	apiv1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/utils/pointer"
+)
+
+type DeploymentApp struct {
```
jmacelroy, line 34: Some comments on these new implementations and interfaces may help. I also find it great to mention when something is made and we know we want it to implement some interface.

jmacelroy: It may allow you to remove some of the receivers or at least simplify them a tiny bit if embedding is used, https://golang.org/doc/effective_go#embedding

```suggestion
type DeploymentApp struct {
        *appsv1.Deployment
}
```

Then things like Name could be accessed directly, foo := NewDeploymentApp(d); fmt.Println(foo.Name)

cmd/init/init.go
```diff
@@ -189,31 +187,31 @@ func getResource(ctx context.Context, namespace, k8sContext string) (*model.K8sO
 		namespace = client.GetContextNamespace(k8sContext)
 	}
 
-	r, err := askForResource(ctx, namespace, c)
+	app, err := askForRunningApp(ctx, namespace, c)
```
jmacelroy, line 190: I'm looking at this from the viewpoint of how large the interface is. In this case it looks like we need a much smaller set of functionality then is defined by the interface. I believe we simply need something that implements `PodSpec()` if embedding is used.

cmd/init/init.go
```diff
@@ -117,29 +115,29 @@ func Run(namespace, k8sContext, devPath, language, workDir string, overwrite boo
...
-				container = r.PodTemplateSpec.Spec.Containers[0].Name
+				container = app.PodSpec().Containers[0].Name
 			}
 
-			suffix := fmt.Sprintf("Analyzing %s '%s'...", strings.ToLower(string(r.ObjectType)), r.Name)
+			suffix := fmt.Sprintf("Analyzing %s '%s'...", app.Kind(), app.Name())
 			spinner := utils.NewSpinner(suffix)
 			spinner.Start()
-			err = initCMD.SetDevDefaultsFromResource(ctx, dev, r, container, language)
+			err = initCMD.SetDevDefaultsFromApp(ctx, dev, app, container, language)
```
jmacelroy, line 135: At this point you may use another smaller interface as the parameter to this function to something that simply provides labels and annotations. I believe doing it this way will make things a little simpler at the start and possibly allow for some cleaner factoring. Also, in the long run will provide a stronger abstraction allowing for further cleanup when/if needed.

pkg/k8s/apps/interface.go
pchico83: Yes, I am planing to simplify it after this iteration. Once the webhook dependency is removed, many of these methods won't be needed

pkg/k8s/apps/deployments.go
pchico83: I tried embedding first, but had problems with methods like:
https://github.com/okteto/okteto/pull/1765/files#diff-b97445ee5f3a92ab765be2bc8402d8291b41a4acd087fa16e09d159c79c4a367R262

I will give it another thought for the next iteration

PR #1768, opened 2021-09-21
Return interrupt signal error and set cfg to error

### LINKED ISSUE #1706: Handle stack status in a stack command is canceled
**Describe the bug**
If I cancel my `okteto stack deploy` command, the Okteto UI shows `processing` state.

**Additional context**
We can handle this as a termination handler of the command to set an `error` state.

jLopezbarb:

Fixes #1706

## Proposed changes

-
-


pkg/cmd/stack/deploy.go
```diff
@@ -165,7 +165,7 @@ func deploy(ctx context.Context, s *model.Stack, c *kubernetes.Clientset, config
 	case <-stop:
 		log.Infof("CTRL+C received, starting shutdown sequence")
 		spinner.Stop()
-		os.Exit(130)
+		return fmt.Errorf("interrupt signal received")
```
pchico83, line 168: I think we do `os.Exit(130)` in many other places to handle "CTRL+C". Should we update it everywhere?

jLopezbarb: I thought is was better to use `os.Exit(130)` to follow terminal rules but returning an error is better for debugging on remotes

pchico83 (approved):

PR #1776, opened 2021-09-22
Eliminate App interface getter/setter methods for labels and annotations

pchico83:

## Proposed changes

Implement a method that returns `ObjectMeta` to reduce the number of getter/setter methods


pkg/k8s/apps/interface.go
```diff
@@ -19,21 +19,16 @@ import (
...
 
 type App interface {
-	Name() string
-	Kind() string
+	TypeMeta() metav1.TypeMeta
+	ObjectMeta() metav1.ObjectMeta
+	TemplateObjectMeta() metav1.ObjectMeta
+
 	Replicas() int32
-	GetLabel(key string) string
```
jmacelroy, line 29: I think it's a good improvement. I like that the interface is smaller and we rely on the Meta abstraction which is stronger due to it being part of all the underlying types we will probably use or easily implementable .

jmacelroy (approved):

PR #1778, opened 2021-09-23
Get global namespace

teresaromero:
## Proposed changes

- Add to the Token the globalnamespace value in order to be able to push images to the global repository

### DIFF
--- a/pkg/okteto/auth.go
+++ b/pkg/okteto/auth.go
@@ -145,7 +145,27 @@ func queryUser(ctx context.Context, client *graphql.Client, token string) (*q, e
 	var user q
 	q := `query {
 		user {
-			id,name,email,externalID,token,new,registry,buildkit,certificate,globalnamespace
+			id,name,email,externalID,token,new,registry,buildkit,certificate,globalNamespace
+		}}`
+
+	req := getRequest(q, token)
+
+	if err := client.Run(ctx, req, &user); err != nil {
+		if strings.Contains(err.Error(), "Cannot query field \"globalNamespace\" on type \"me\"") {
+			return deprecatedQueryUser(ctx, client, token)
+		}
+		return nil, err
+	}
+
+	return &user, nil
+}
+
+//TODO: remove when all users are in Okteto Enterprise which supports globalNamespace
+func deprecatedQueryUser(ctx context.Context, client *graphql.Client, token string) (*q, error) {
+	var user q
+	q := `query {
+		user {
+			id,name,email,externalID,token,new,registry,buildkit,certificate
 		}}`
 
 	req := getRequest(q, token)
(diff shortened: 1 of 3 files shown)


pkg/okteto/auth.go
```diff
@@ -143,7 +145,7 @@ func queryUser(ctx context.Context, client *graphql.Client, token string) (*q, e
 	var user q
 	q := `query {
 		user {
-			id,name,email,externalID,token,new,registry,buildkit,certificate
+			id,name,email,externalID,token,new,registry,buildkit,certificate,globalnamespace
```
pchico83, line 148: If you run this against Cloud, it will fail, because we don't support this field yet.
We should catch the error in that case, and do the query without `globalnamespace`.
There is similar code when deploying a pipeline, for example.

pchico83 (approved):

pkg/okteto/auth.go
```diff
@@ -140,6 +142,26 @@ func saveAuthData(user *User, url string) error {
...
+	var user q
+	q := `query {
+		user {
+			id,name,email,externalID,token,new,registry,buildkit,certificate,globalNamespace
+		}}`
+
+	req := getRequest(q, token)
+
+	if err := client.Run(ctx, req, &user); err != nil {
+		if strings.Contains(err.Error(), "Cannot query field \"globalNamespace\" on type \"me\"") {
```
ifbyol, line 154: Maybe it is out of the scope of this PR, but maybe we can start using IntrospectionQueries to this kind of things: https://graphql.org/learn/introspection/.

Executing this query against my dev cluster

```
query IntrospectionQuery {
  __type(name: "me"){
    name
    fields {
      name
    }
  }
}
``` 

I got

```
{
  "data": {
    "__type": {
      "fields": [
        {
          "name": "avatar"
        },
        {
          "name": "buildkit"
        },
        {
          "name": "capabilities"
        },
        {
          "name": "certificate"
        },
        {
          "name": "config"
        },
        {
          "name": "email"
        },
        {
          "name": "externalID"
        },
        {
          "name": "gitSSHSupport"
        },
        {
          "name": "githubID"
        },
        {
          "name": "id"
        },
        {
          "name": "integrations"
        },
        {
          "name": "name"
        },
        {
          "name": "namespace"
        },
        {
          "name": "new"
        },
        {
          "name": "onboarded"
        },
        {
          "name": "personalAccessTokens"
        },
        {
          "name": "plan"
        },
        {
          "name": "planSubscribed"
        },
        {
          "name": "privateEndpointToken"
        },
        {
          "name": "quickstarts"
        },
        {
          "name": "quotaPlan"
        },
        {
          "name": "registry"
        },
        {
          "name": "secrets"
        },
        {
          "name": "super"
        },
        {
          "name": "token"
        },
        {
          "name": "trialExpiration"
        },
        {
          "name": "userDefinedNamespaces"
        }
      ],
      "name": "me"
    }
  }
}
```

I think this is most robust since it is based on the graphql schema defined on the server

teresaromero: Should we open a ticket for this?

pchico83: Yeap, I would implement it right after this is merge

ifbyol: Yes, we can merge this PR as it is now and include the query instrospection later if we see the value it could provide, but do not block this PR for that

ifbyol (approved):

PR #1783, opened 2021-09-23
Remove several methods from App interface

pchico83:
This is just a cleanup of the App interface refactor

### DIFF
--- a/cmd/up/up.go
+++ b/cmd/up/up.go
@@ -257,7 +257,7 @@ func (up *upContext) start(build bool) error {
 	stop := make(chan os.Signal, 1)
 	signal.Notify(stop, os.Interrupt)
 
-	analytics.TrackUp(true, up.Dev.Name, up.getInteractive(), len(up.Dev.Services) == 0, up.isSwap, up.Dev.Divert != nil)
+	analytics.TrackUp(true, up.Dev.Name, up.getInteractive(), len(up.Dev.Services) == 0, up.Dev.Divert != nil)
 
 	go up.activateLoop(build)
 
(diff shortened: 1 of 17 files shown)


cmd/up/up.go
```diff
@@ -327,9 +327,6 @@ func (up *upContext) getApp(ctx context.Context) (apps.App, bool, error) {
 		return apps.NewDeploymentApp(apps.GetDeploymentSandbox(up.Dev)), true, nil
 	}
 	if err == nil {
-		if app.ObjectMeta().Annotations[model.OktetoAutoCreateAnnotation] != model.OktetoUpCmd {
-			up.isSwap = true
```
ifbyol, line 331: Just curiosity, what was the meaning of this flag?

pkg/k8s/apps/crud.go
```diff
@@ -50,6 +50,41 @@ func Get(ctx context.Context, dev *model.Dev, namespace string, c kubernetes.Int
 	return &StatefulSetApp{sfs: sfs}, nil
 }
 
+//IsDevModeOn returns if a statefulset is in devmode
+func IsDevModeOn(app App) bool {
+	return app.ObjectMeta().Labels[model.DevLabel] == "true"
+}
+
+func (t *Translation) DevModeOff() {
+	t.App.DevModeOff(t)
```
ifbyol, line 59: This generates a weird circular dependency. If `App.DevModeOff` would accidentally call to `t.DevModeOff()`, we would have an infinite loop. Is there a way we could avoid it? Would it make more complex the code?

ifbyol (approved):

cmd/up/up.go
pchico83: it was related to `autocreate`, before existing the `autocreate` field

pkg/k8s/apps/crud.go
pchico83: I don't like it either :-)
But next refactor will eliminate the object `Translation` as it is, I think it is better to fix it on that iteration

ifbyol: Sounds good!

PR #1785, opened 2021-09-24
Graphql introspection field validation

teresaromero:
## Proposed changes

- instead of catching an error after a graphql query fails for field validation, retrieve prior to the query, the fields that are available from the list of fields required.
- use graphql introspection for this purpose

While developing this solution i noticed the introspection query for getting the fields that a schema has, is not protected by Bearer token, so this is available for any request of type introspection for a given model. Is this something we want? Reading about this, is not recommended to have the introspection queries open.

### DIFF
--- a/pkg/okteto/auth.go
+++ b/pkg/okteto/auth.go
@@ -142,32 +142,13 @@ func saveAuthData(user *User, url string) error {
 }
 
 func queryUser(ctx context.Context, client *graphql.Client, token string) (*q, error) {
+	queryFields := queryAvailableFields(ctx, client, "me", `id,name,email,externalID,token,new,registry,buildkit,certificate,globalNamespace`)
+
 	var user q
-	q := `query {
+	q := fmt.Sprintf(`query {
 		user {
-			id,name,email,externalID,token,new,registry,buildkit,certificate,globalNamespace
-		}}`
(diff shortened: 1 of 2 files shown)


pkg/okteto/auth.go
```diff
@@ -142,32 +142,13 @@ func saveAuthData(user *User, url string) error {
...
-			return deprecatedQueryUser(ctx, client, token)
-		}
-		return nil, err
-	}
-
-	return &user, nil
-}
-
-//TODO: remove when all users are in Okteto Enterprise which supports globalNamespace
-func deprecatedQueryUser(ctx context.Context, client *graphql.Client, token string) (*q, error) {
```
pchico83, line 164: we have other deprecated queries. at least, we have one for pipelines

rberrelleza:
This looks great! 

One potential downside I see here is that we'll be making twice as many queries (one to get the fields, and then one to do the actual query). Could that be a problem from a scalability or performance perspective? The advantage of catching the error is that you only "pay the extra request" if you're in an older version of the client. But I do understand that the current approach has a big downside when it comes to the maintainability of that approach.

teresaromero:
We won't use this feature for the moment. 
Closing the issue

PR #1787, opened 2021-09-27
Add okteto_namespace envvar to okteto stack deploy

jLopezbarb:

## Proposed changes

- Adds the option to deploy the stack from env var OKTETO_NAMESPACE
-

### DIFF
--- a/pkg/model/stack.go
+++ b/pkg/model/stack.go
@@ -371,7 +371,7 @@ func validateStackName(name string) error {
 func (s *Stack) UpdateNamespace(namespace string) error {
 	if namespace == "" {
 		envNamespace := os.Getenv("OKTETO_NAMESPACE")
-		if envNamespace != "" && s.Namespace != envNamespace {
+		if s.Namespace != "" && s.Namespace != envNamespace {
 			return fmt.Errorf("the namespace in the okteto stack manifest '%s' does not match the environment variable OKTETO_NAMESPACE '%s'", s.Namespace, namespace)
 		}
 		return nil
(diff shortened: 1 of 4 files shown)


pkg/model/stack.go
```diff
@@ -369,6 +370,10 @@ func validateStackName(name string) error {
 //UpdateNamespace updates the dev namespace
 func (s *Stack) UpdateNamespace(namespace string) error {
 	if namespace == "" {
+		envNamespace := os.Getenv("OKTETO_NAMESPACE")
+		if s.Namespace != "" && s.Namespace != envNamespace {
+			return fmt.Errorf("the namespace in the okteto stack manifest '%s' does not match the environment variable OKTETO_NAMESPACE '%s'", s.Namespace, namespace)
+		}
```
pchico83, line 376: should we do something like `os.SetEnv("OKTETO_NAMESPACE", s.Namespace)` at some point?

jLopezbarb: I do't think so, we don't get `OKTETO_NAMESPACE` from other place

pkg/model/stack.go
```diff
@@ -246,6 +247,14 @@ func getStackName(name, stackPath, actualStackName string) (string, error) {
 
 // ReadStack reads an okteto stack
 func ReadStack(bytes []byte, isCompose bool) (*Stack, error) {
+	sNamespace := &StackNamespace{}
```
pchico83, line 250: Could we do the same for okteto manifests?

pchico83:
@jLopezbarb this has been already implemented as part of `okteto context`

PR #1788, opened 2021-09-28
Translate okteto registry at Dockerfiles

teresaromero:
## Proposed changes

- Expand registry when okteto.dev or okteto.global are in the tags of the Dockerfile base images


pkg/cmd/build/build.go
```diff
@@ -35,8 +35,8 @@ func Run(ctx context.Context, namespace, buildKitHost string, isOktetoCluster bo
 		return err
 	}
 
-	if buildKitHost == okteto.CloudBuildKitURL && dockerFile != "" {
-		dockerFile, err = registry.GetDockerfile(path, dockerFile)
+	if dockerFile != "" {
+		dockerFile, err = registry.GetDockerfile(path, dockerFile, ctx)
```
pchico83, line 39: the convention is to use `ctx` as the first argument of any function

pkg/registry/file.go
```diff
@@ -67,6 +68,7 @@ func getDockerfileWithCacheHandler(filename string) (string, error) {
 	for scanner.Scan() {
 		line := scanner.Text()
 		translatedLine := translateCacheHandler(line, userID)
```
pchico83, line 70: this should only be done for Okteto Cloud, I think you missed that check

pkg/registry/file.go
```diff
@@ -103,6 +105,32 @@ func translateCacheHandler(input, userID string) string {
 	return input
 }
 
+func translateOktetoRegistry(input string, ctx context.Context) string {
```
pchico83, line 108: I think for this case, we can just replace strings.
something like `okteto.dev` -> `registryurl/namespace` and `okteto.global` -> `registryurl/okteto` is a single

pchico83: Also, could you add unit tests for this function?

pkg/registry/file.go
```diff
@@ -103,6 +109,20 @@ func translateCacheHandler(input, userID string) string {
 	return input
 }
 
+func translateOktetoRegistryImage(input, oktetoRegistryURL, namespace string) string {
+
+	if strings.Contains(input, okteto.DevRegistry) {
+		return replaceRegistry(input, okteto.DevRegistry, oktetoRegistryURL, namespace)
```
pchico83, line 115: edge case, but you might also need to replace `okteto.GlobalRegistry`.

jLopezbarb (approved):

pkg/okteto/utils.go
```diff
@@ -0,0 +1,26 @@
+package okteto
+
+import (
+	"io/ioutil"
+	"os"
+)
+
+// SetToken is used in test to setup token params, returns tempDir and error
+func SetToken(t *Token) (string, error) {
```
ifbyol, line 9: If this is only used by tests I would have it in an internal package: https://golang.org/doc/go1.4#internalpackages.

I think having the function in a file like `pkg/internal/testutils/utils.go` would let you use the `SetToken` function in the places where it is being used and let others know that it is an utility only for testing purposes

ifbyol (approved):

pkg/okteto/utils.go
teresaromero: I've removed this, as with the new Context is no longer needed :)

pchico83 (approved):

ifbyol (approved):

PR #1799, opened 2021-10-01
refactor: move from io/ioutil to io and os package

Juneezee:
## Proposed changes

- The `io/ioutil` package has been deprecated in Go 1.16 (See https://golang.org/doc/go1.16#ioutil). Since okteto has upgraded to Go 1.17 (#1734), this PR replaces the existing `io/ioutil` functions with their new definitions in `io` and `os` packages.

### DIFF
--- a/cmd/init/init.go
+++ b/cmd/init/init.go
@@ -16,7 +16,6 @@ package init
 import (
 	"context"
 	"fmt"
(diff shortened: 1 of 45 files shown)


pchico83 (approved):
@Juneezee thanks for taking care of this!!!

PR #1800, opened 2021-10-01
Jlopezbarb/fix-ctx-e2e

jLopezbarb:
Fixes errors from ctx 

## Proposed changes
- Adds a new selector for choosing context
- Fix e2e test
- Inject secrets and update context on every command

### DIFF
--- a/cmd/context/context.go
+++ b/cmd/context/context.go
@@ -245,10 +245,7 @@ func updateContext(cred okteto.Credential) error {
 	kubeconfigFile := config.GetKubeconfigPath()
 	cfg := client.GetKubeconfig(kubeconfigFile)
 	u := octx.ToUser()
-	if err := okteto.SaveOktetoClusterContext(okteto.Context().Name, u, okteto.Context().Namespace, cfg); err != nil {
-		return err
-	}
-	return nil
+	return okteto.SaveOktetoClusterContext(okteto.Context().Name, u, okteto.Context().Namespace, cfg)
 }
 
 func getSecretsAndCredentials(ctx context.Context) (*okteto.SecretsAndCredentialToken, error) {
--- a/cmd/context/context.go
+++ b/cmd/context/context.go
@@ -193,13 +193,8 @@ func runContext(ctx context.Context, oktetoContext string, ctxOptions *ContextOp
 }
 
 func getContext(ctxOptions *ContextOptions) (string, error) {
-	clusters := []string{"Okteto Cloud", "Okteto Enterprise"}
-	if !ctxOptions.OnlyOkteto {
-		k8sClusters := getKubernetesContextList()
-		clusters = append(clusters, k8sClusters...)
-	}
-	oktetoContext, err := utils.AskForOptions(clusters, "Select the context you want to activate:")
-
+	ctxs := getContextsSelection(ctxOptions)
+	oktetoContext, err := AskForOptions(ctxs, "Select the context you want to activate:")
 	if err != nil {
 		return "", err
 	}
--- a/cmd/context/context.go
+++ b/cmd/context/context.go
@@ -218,15 +218,11 @@ func Init(ctx context.Context) error {
 		}
 	}
 	if okteto.IsOktetoContext() {
-		client, err := okteto.NewOktetoClient()
-		if err != nil {
-			return err
-		}
-		secretsAndKubeCredentials, err := client.GetSecretsAndKubeCredentials(ctx)
-		if err != nil {
-			return err
-		}
 
+		secretsAndKubeCredentials, err := getSecretsAndCredentials(ctx)
+		if err != nil {
+			return err
+		}
 		if err := updateContext(secretsAndKubeCredentials.Credentials); err != nil {
 			return err
 		}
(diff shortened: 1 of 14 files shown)


cmd/context/context.go
```diff
@@ -222,7 +217,48 @@ func Init(ctx context.Context) error {
...
+	for _, secret := range secrets {
+		os.Setenv(secret.Name, secret.Value)
+	}
+}
+
+func updateContext(cred okteto.Credential) error {
+	octx := okteto.Context()
+	kubeconfigFile := config.GetKubeconfigPath()
+	cfg := client.GetKubeconfig(kubeconfigFile)
+	u := &okteto.User{
```
pchico83, line 251: We have this logic in a few places, we could have an `okteto.Context().ToUser` function

cmd/context/selector.go
```diff
@@ -0,0 +1,402 @@
...
+	optionSelected, err := prompt.Run()
+	if err != nil {
+		log.Infof("invalid init option: %s", err)
+		return "", fmt.Errorf("invalid option")
+	}
+
+	return optionSelected, nil
+}
+
+func (s OktetoSelector) Run() (string, error) {
```
pchico83, line 122: Did you test it on windows?

pkg/k8s/diverts/client.go
```diff
@@ -53,6 +54,9 @@ func NewForConfig(cfg *rest.Config) (*DivertV1Client, error) {
 }
 
 func GetClient(thisContext string) (*DivertV1Client, error) {
+	if okteto.IsOktetoContext() {
```
pchico83, line 57: I think you don't need this. The divert client should always read from `$HOME/.okteto/contexts/.kubeconfig`

pkg/registry/registry.go
```diff
@@ -114,7 +114,7 @@ func GetHiddenExposePorts(image string) []model.Port {
 	var err error
 	var username string
 	var token string
-	if okteto.IsOktetoContext() {
+	if okteto.IsOktetoContext() && strings.HasPrefix(image, okteto.DevRegistry) {
```
pchico83, line 117: why do we need this?

pchico83 (commented):
Looking great!

pkg/registry/registry.go
jLopezbarb: We need that for getting expose fields from public docker hub images. Compose test was failing because we couldn't retrieve mongo and rabbit ports and we were not creating any service for those services

jLopezbarb: I just realized that we need to add that it can start with the global tag

cmd/context/selector.go
jLopezbarb: I will try it and send some screenshots from different windows terminals

jLopezbarb: From left to right (powershell, cmd, git cmd and windows terminal) :
 Captura de pantalla 2021-10-01 175818

pkg/k8s/diverts/client.go
```diff
@@ -53,6 +54,8 @@ func NewForConfig(cfg *rest.Config) (*DivertV1Client, error) {
 }
 
 func GetClient(thisContext string) (*DivertV1Client, error) {
+	thisContext = okteto.UrlToContext(thisContext)
```
pchico83, line 57: what I mean is that `thisContext` shoul be always "", and use:
https://github.com/okteto/okteto/blob/master/pkg/k8s/client/client.go#L54

to point to our copy of kubeconfig in `$HOME/.okteto/contexts/kubeconfig`

pkg/registry/registry.go
```diff
@@ -114,7 +114,7 @@ func GetHiddenExposePorts(image string) []model.Port {
 	var err error
 	var username string
 	var token string
-	if okteto.IsOktetoContext() {
+	if okteto.IsOktetoContext() && (strings.HasPrefix(image, okteto.DevRegistry) || strings.HasPrefix(image, okteto.GlobalRegistry)) {
```
pchico83, line 117: `image = ExpandOktetoDevRegistry(image)` doesn't change `image` if  `okteto.dev` is not in the string, right?
I think you can remove `image = ExpandOktetoDevRegistry(image)` and leave:
```
		image = ExpandOktetoDevRegistry(image)
		image = ExpandOktetoGlobalRegistry(image)
		username = okteto.Context().UserID
		token = okteto.Context().Token
```

cmd/context/context.go
```diff
@@ -222,7 +217,40 @@ func Init(ctx context.Context) error {
 			return err
 		}
 	}
+	if okteto.IsOktetoContext() {
+		client, err := okteto.NewOktetoClient()
+		if err != nil {
+			return err
+		}
+		secretsAndKubeCredentials, err := client.GetSecretsAndKubeCredentials(ctx)
+		if err != nil {
```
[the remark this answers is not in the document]
pchico83: also, we tell the user to run `okteto context url` again if we get a 401 response

cmd/context/context.go
```diff
@@ -222,7 +217,40 @@ func Init(ctx context.Context) error {
...
+			return err
+		}
+
+		if err := updateContext(secretsAndKubeCredentials.Credentials); err != nil {
+			return err
+		}
+		setSecrets(secretsAndKubeCredentials.Secrets)
+	}
+
 	os.Setenv("OKTETO_USERNAME", okteto.Context().Username)
```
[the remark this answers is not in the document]
jLopezbarb: I wouldn't do that because we call `okteto.IsOktetoContext()` more than once for some commands, I think injecting them at the beginning of the command and not calling the api for retrieving the secrets again should be the path to go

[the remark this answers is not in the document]
pchico83: I mean to the `if` section here:
https://github.com/okteto/okteto/pull/1800/files#diff-7ba016dd78fdd00cce76cf854dcc18447a4228dc8dc8bcd660fcdd68876d38e9R229

cmd/context/context.go
```diff
@@ -222,7 +218,66 @@ func Init(ctx context.Context) error {
...
+		if err != nil {
+			return err
+		}
+		if err := updateContext(secretsAndKubeCredentials.Credentials); err != nil {
+			log.Info(err)
+		}
+		setSecrets(secretsAndKubeCredentials.Secrets)
+
+		os.Setenv("OKTETO_USERNAME", okteto.Context().Username)
+		os.Setenv("OKTETO_NAMESPACE", okteto.Context().Namespace)
```
pchico83, line 233: I would keep `OKTETO_NAMESPACE` outside of the if block, it is valid for non okteto clusters

pchico83 (approved):

PR #1805, opened 2021-10-04
Sync Analytics with Telemetry Param

teresaromero:
## Proposed changes

- if Telemetry is disabled by config, Analytics in CLI are also disabled


pkg/okteto/auth.go
```diff
@@ -97,36 +99,44 @@ func Auth(ctx context.Context, code, url string) (*User, error) {
...
+			TelemetryEnabled graphql.Boolean `graphql:"telemetryEnabled"`
 		} `graphql:"user"`
 	}
 	err := c.client.Query(ctx, &query, nil)
 	if err != nil {
 		if strings.Contains(err.Error(), "Cannot query field \"globalNamespace\" on type \"me\"") {
 			return c.deprecatedQueryUser(ctx)
 		}
+		if strings.Contains(err.Error(), "Cannot query field \"telemetryEnabled\" on type \"me\"") {
+			return c.deprecatedQueryUser(ctx)
```
pchico83, line 121: In this case, the default value of `telemetryEnabled` should be `true`, to keep the current behavior.

pkg/okteto/context.go
```diff
@@ -467,5 +473,12 @@ func (okctx *OktetoContext) ToUser() *User {
 		GlobalNamespace: okctx.GlobalNamespace,
 	}
 	return u
+}
 
+func GetTelemetryEnabled() bool {
+	t, err := strconv.ParseBool(Context().TelemetryEnabled)
```
pchico83, line 479: For clusters not managed by okteto, telemetry should be always `true`

pkg/analytics/config.go
```diff
@@ -107,7 +110,9 @@ func get() *Analytics {
 		log.Debugf("error unmarshaling analytics: %s", err)
 		return &Analytics{Enabled: false, MachineID: ""}
 	}
-
+	if !okteto.GetTelemetryEnabled() {
+		result.Enabled = false
+	}
```
pchico83, line 115: not related to this PR, but we should set `currentAnalytics = result` here

pkg/analytics/config.go
```diff
@@ -89,6 +89,9 @@ func deprecatedFileExists() bool {
 
 func get() *Analytics {
 	if currentAnalytics != nil {
+		if !okteto.GetTelemetryEnabled() {
```
pchico83, line 92: I think you don't need this, is already done at the end of this function

pkg/analytics/config.go
```diff
@@ -107,7 +110,9 @@ func get() *Analytics {
 		log.Debugf("error unmarshaling analytics: %s", err)
 		return &Analytics{Enabled: false, MachineID: ""}
 	}
-
+	if !okteto.GetTelemetryEnabled() {
```
pchico83, line 113: Maybe it is better to not initialize this here. And add and extra check if it is an okteto context and telemetry is disabled here:
https://github.com/okteto/okteto/blob/master/pkg/analytics/track.go#L286

pchico83:
@teresaromero we can query the current user in this logic:
https://github.com/okteto/okteto/blob/21dd0d77ed088d84320acb8cfb1d5c7b927b8b7c/cmd/context/context.go#L223

This way, the value of `telemetryEnabled` will be updated if the admin disabled the setting in OE without needing to relogin

rberrelleza:
> @teresaromero we can query the current user in this logic:
> 
> https://github.com/okteto/okteto/blob/21dd0d77ed088d84320acb8cfb1d5c7b927b8b7c/cmd/context/context.go#L223
> 
> This way, the value of `telemetryEnabled` will be updated if the admin disabled the setting in OE without needing to relogin

As we add more queries to the `okteto up` sequence, things will slow down. Can we have a metric or something so we can track how much the 'time to terminal' is affected as we make more of these changes?

pchico83:
> > @teresaromero we can query the current user in this logic:
> > https://github.com/okteto/okteto/blob/21dd0d77ed088d84320acb8cfb1d5c7b927b8b7c/cmd/context/context.go#L223
> > 
> > This way, the value of `telemetryEnabled` will be updated if the admin disabled the setting in OE without needing to relogin
> 
> As we add more queries to the `okteto up` sequence, things will slow down. Can we have a metric or something so we can track how much the 'time to terminal' is affected as we make more of these changes?

@rberrelleza we are not adding queries, it can be done in a single graphql query. In this iteration we have also deleted a k8s query because we don't need to check the `dev.okteto.com` label on the namespace (we infer that info from the context type). Also, we are already getting metrics for the time taken by the `okteto up` sequence

rberrelleza:
> telemetryEnabled

@pchico83  requesting the telemetry data requires more processing and adds more bytes to the payload we return (not a lot, but it keeps adding, and we've been adding more properties and queries lately).  Are we going to be able to tell how long the requests are taking VS the entire duration of the up? (since up has other factors like how big the image is, if the cluster is busy, etc... it might not be super useful as a baseline).

pkg/okteto/context.go
teresaromero: Changed this to IsTelemetryEnabled where the okteto cluster is checked and true is return in case is not

pkg/okteto/auth.go
teresaromero: Changed the user.TelemetryEnabled type to string, in case the query/mutation does not return this data, empty string will be saved in the user, and when saving this value in the context, if this is empty string, the default value true will be saved into the context

jLopezbarb (approved):

ifbyol (approved):

pchico83 (approved):

PR #1809, opened 2021-10-04
Initial implementation for deploy command

ifbyol:

## Proposed changes

- First implementation of `deploy` command


cmd/deploy/proxy.go
```diff
@@ -0,0 +1,182 @@
...
+
+		// Set the right bearer token based on the original kubeconfig
+		r.Header.Set("Authorization", fmt.Sprintf("Bearer %s", clusterConfig.BearerToken))
+
+		// Modify all resources updated or created to include the label.
+		if r.Method == "PUT" || r.Method == "POST" {
+			b, err := io.ReadAll(r.Body)
+			if err != nil {
+				log.Errorf("could not read the request body: %s", err)
+				rw.WriteHeader(500)
```
ifbyol, line 107: What do we want to do in case of some errors processing the body request? For now I'm failing with a 500 error but this would prevent any kubernetes command to be executed (just in the scope of the `okteto deploy` command. User's kubeconfig won't be modified)

cmd/deploy/deploy.go
```diff
@@ -0,0 +1,173 @@
...
+					token:        sessionToken,
+				}),
+			}
+			return c.runDeploy(ctx, cwd, name, filename, variables)
+		},
+	}
+
+	cmd.Flags().StringVarP(&name, "name", "a", "", "application name")
+	cmd.Flags().StringVarP(&filename, "filename", "f", "", "path to the manifest file")
+	cmd.Flags().StringArrayVarP(&variables, "var", "v", []string{}, "set a pipeline variable (can be set more than once)")
```
ifbyol, line 106: Not sure if this is needed or would be read from the manifest

cmd/deploy/deploy.go
```diff
@@ -0,0 +1,173 @@
...
+		log.Errorf("could not find manifest file to be executed: %s", err)
+		return err
+	}
+
+	// Get secrets from API
+	sList, err := dc.getSecrets(ctx)
+	if err != nil {
+		log.Errorf("could not load Okteto Secrets: %s", err.Error())
+		sList = []string{}
+	}
```
ifbyol, line 135: For the pipeline-installer-image I think this wouldn't be necessary since the installer would inject them when execute the command, but I guess that for the rest of cases we want to include secrets on the execution as we do for other commands

cmd/deploy/deploy.go
```diff
@@ -0,0 +1,173 @@
...
+	"github.com/google/uuid"
+	contextCMD "github.com/okteto/okteto/cmd/context"
+	"github.com/okteto/okteto/cmd/utils"
+	"github.com/okteto/okteto/pkg/log"
+	"github.com/okteto/okteto/pkg/model"
+	"github.com/spf13/cobra"
+	"k8s.io/client-go/rest"
+)
+
+var tempKubeConfig = "/var/tmp/okteto.kubeconfig"
```
pchico83, line 19: I would use a file in the okteto home folder

cmd/deploy/deploy.go
```diff
@@ -0,0 +1,173 @@
...
+			if err != nil {
+				log.Errorf("could not read certificate: %s", err)
+				return err
+			}
+
+			// Generate a token for the requests done to the proxy
+			sessionToken := uuid.NewString()
+
+			c := &deployCommand{
+				getSecrets:  getOktetoSecretsAsEnvironmenVariables,
```
pchico83, line 89: this is already done when initializing the context

cmd/deploy/deploy.go
pchico83: It is a nice addition. Not needed for the first version, but if it is already implemented I would keep it

cmd/deploy/deploy.go
pchico83: It is already done when initializing the context, you can remove it

cmd/deploy/deploy.go
```diff
@@ -0,0 +1,173 @@
...
+
+	if err := dc.proxy.Start(ctx, name, clusterConfig); err != nil {
+		log.Errorf("could not start proxy server %s", err)
+		return err
+	}
+
+	// Include secrets as part of variables list
+	env := append(variables, sList...)
+	// Set KUBECONFIG environment variable to make sure it executes kubectl commands against the proxy
+	env = append(env, fmt.Sprintf("KUBECONFIG=%s", tempKubeConfig))
```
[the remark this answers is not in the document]
pchico83: I didn't investigate it, but I am not familiar with it

cmd/deploy/deploy.go
```diff
@@ -0,0 +1,173 @@
...
+	var commandErr error
+	for _, command := range m.Deploy {
+		if err := dc.executor.Execute(command, env); err != nil {
+			log.Errorf("error executing command '%s': %s", command, err.Error())
+			commandErr = err
+			break
+		}
+	}
+
+	if err := dc.proxy.Shutdown(ctx); err != nil {
```
pchico83, line 156: should it be a `defer` after calling `proxy.Start`?

cmd/deploy/deploy.go
```diff
@@ -0,0 +1,173 @@
...
+		log.Errorf("could not stop local server: %s", err)
+		return err
+	}
+
+	return commandErr
+}
+
+func getName(cwd string) string {
```
pchico83, line 164: @jLopezbarb do we have a function to infer this name in pipelines?

cmd/deploy/executor.go
```diff
@@ -0,0 +1,49 @@
...
+	}
+
+	cmd.Stderr = cmd.Stdout
+	done := make(chan struct{})
+	scanner := bufio.NewScanner(r)
+
+	go func() {
+		for scanner.Scan() {
+			line := scanner.Text()
+			log.Information(line)
```
pchico83, line 35: I would print is as normal text

cmd/deploy/proxy.go
pchico83: Yeap, I think it is the best we can do

main.go
```diff
@@ -110,6 +111,11 @@ func main() {
 	root.AddCommand(cmd.Restart())
 	root.AddCommand(cmd.Update())
 
+	deployCmd := deploy.Deploy(ctx)
+	deployCmd.Hidden = true
```
pchico83, line 115: I think you can define this in the `deploy.Deploy` function itself

main.go
```diff
@@ -110,6 +111,11 @@ func main() {
...
+
+	root.AddCommand(deployCmd)
```
pchico83, line 117: I would add `okteto destroy` as part of this PR

cmd/deploy/deploy.go
jmacelroy: Could this be a const?

cmd/deploy/deploy.go
```diff
@@ -0,0 +1,173 @@
...
+	getManifest func(cwd, name, filename string) (*Manifest, error)
+
+	proxy      proxyInterface
+	kubeconfig kubeConfigHandler
+	executor   manifestExecutor
+}
+
+//Deploy deploys the okteto manifest
+func Deploy(ctx context.Context) *cobra.Command {
+	var filename string
```
jmacelroy, line 48: style nit, use a `var {}` block.

cmd/deploy/deploy.go
```diff
@@ -0,0 +1,173 @@
...
+		return err
+	}
+
+	if err := dc.kubeconfig.Modify(ctx, dc.proxy.GetPort(), dc.proxy.GetToken()); err != nil {
+		log.Errorf("could not create temporal kubeconfig %s", err)
+		return err
+	}
+
+	// Read manifest file with the commands to be executed
+	m, err := dc.getManifest(cwd, name, filename)
```
jmacelroy, line 124: We should probably use a more descriptive variable name when the usage is so far away.

cmd/deploy/proxy.go
```diff
@@ -0,0 +1,182 @@
...
+	if err != nil {
+		return err
+	}
+
+	p.createAndStartProxyServer(ctx, handler, p.proxyConfig.port, p.proxyConfig.certificates)
+
+	return nil
+}
+
+func (p *proxy) createAndStartProxyServer(ctx context.Context, handler http.Handler, port int, certs []tls.Certificate) {
```
jmacelroy, line 51: It's probably easier long-term to just do this in main. You could keep the goroutine in Start if you like but I think it's cleane to configure the server, register the handler, then pass the config and server into a New*. This way all the config is closer to where cli would configure it and this already longer list of config won't grow when we need to change timeouts.

cmd/deploy/proxy.go
```diff
@@ -0,0 +1,182 @@
...
+}
+
+func (p *proxy) createProxyHandler(ctx context.Context, name string, clusterConfig *rest.Config) (http.Handler, error) {
+	trans, err := rest.TransportFor(clusterConfig)
+	if err != nil {
+		log.Errorf("could not get transport from config: %s", err)
+		return nil, err
+	}
+
+	handler := http.DefaultServeMux
```
jmacelroy, line 80: I'd create a new mux with https://pkg.go.dev/net/http#NewServeMux. This is a global of sorts that other packages can register things to in their `init()`

cmd/deploy/proxy.go
```diff
@@ -0,0 +1,182 @@
...
+	p.createAndStartProxyServer(ctx, handler, p.proxyConfig.port, p.proxyConfig.certificates)
+
+	return nil
+}
+
+func (p *proxy) createAndStartProxyServer(ctx context.Context, handler http.Handler, port int, certs []tls.Certificate) {
+	s := &http.Server{
+		Addr:         fmt.Sprintf(":%d", port),
+		Handler:      handler,
+		ReadTimeout:  3600 * time.Second,
```
jmacelroy, line 55: I think these are too long and will probably cascading issues if something bad happens. I don't think the proxy will be open to much abuse but I figure the timeouts from https://blog.cloudflare.com/exposing-go-on-the-internet/ would be a good start.

cmd/deploy/proxy.go
```diff
@@ -0,0 +1,182 @@
...
+		Host:   strings.TrimPrefix(clusterConfig.Host, "https://"),
+		Scheme: "https",
+	})
+	proxy.Transport = trans
+
+	log.Debugf("forwarding host: %s", clusterConfig.Host)
+
+	handler.HandleFunc("/", func(rw http.ResponseWriter, r *http.Request) {
+		token := r.Header.Get("Authorization")
+		expectedToken := fmt.Sprintf("Bearer %s", p.proxyConfig.token)
```
jmacelroy, line 92: What happens to the next hop if the actual token is bad? Is this place better to do some further validation or does the token need to make it all the way?

PR #1815, opened 2021-10-05
Retry API calls if is transient error

### LINKED ISSUE #1813: Retry TLS errors to the okteto api
**Describe the bug**
My `okteto pipeline deploy` commands fails with TLS errors

jLopezbarb:

Fixes #1813

## Proposed changes

- Retry on TLS issues
-

### DIFF
--- a/pkg/okteto/client.go
+++ b/pkg/okteto/client.go
@@ -137,3 +137,25 @@ func InDevContainer() bool {
 
 	return false
 }
+
+func (c *OktetoClient) Query(ctx context.Context, query interface{}, variables map[string]interface{}) error {
+	err := c.client.Query(ctx, query, variables)
+	if err != nil {
+		if errors.IsTransient(err) {
+			err = c.client.Query(ctx, query, variables)
+		}
+		return translateAPIErr(err)
+	}
+	return nil
+}
+
+func (c *OktetoClient) Mutate(ctx context.Context, mutation interface{}, variables map[string]interface{}) error {
+	err := c.client.Mutate(ctx, mutation, variables)
+	if err != nil {
+		if errors.IsTransient(err) {
+			err = c.client.Mutate(ctx, mutation, variables)
+		}
+		return translateAPIErr(err)
+	}
+	return nil
+}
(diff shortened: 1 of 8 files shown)


pkg/okteto/client.go
```diff
@@ -137,3 +137,25 @@ func InDevContainer() bool {
...
+		return translateAPIErr(err)
+	}
+	return nil
+}
+
+func (c *OktetoClient) Mutate(ctx context.Context, mutation interface{}, variables map[string]interface{}) error {
+	err := c.client.Mutate(ctx, mutation, variables)
+	if err != nil {
+		if errors.IsTransient(err) {
+			err = c.client.Mutate(ctx, mutation, variables)
```
ifbyol, line 156: Checking the `IsTransient` logic, it seems that timeouts are there

```
strings.Contains(err.Error(), "operation time out"),
strings.Contains(err.Error(), "i/o timeout"),
```

Couldn't this generate issues? In a timeout scenario, the mutation could be executed, so this could generate false errors or unexpected results.

For example, in case of creating a namespace, first one would fail with a timeout (but the namespace could be created in the backend), so we retry again after the error and it fails because the namespace already exists. Other scenarios could be deployed twice same stack/pipeline, etc...

I had problems in the past having this kind of autoretry logic when operations are not idempotent and they are really hard to debug if you don't recall in this logic.

pchico83: This is a great point @ifbyol 
@jLopezbarb @ifbyol what about implementing the retry just for queries? and not for mutations for now?
Also, I wouldn't reuse `errors.IsTransient`, it is designed for k8s api errors.
I would create a different function for transient okteto api errors

pchico83 (approved):

pkg/okteto/client.go
ifbyol: @pchico83 @jLopezbarb  I see 2 quick options we may do:

1. Use this retry logic only for queries as you suggest. I think this is the safest one.
2. Avoid timeout errors, at least for mutations. The rest of errors shouldn't have that effect, but we can always have a surprise there.

I would go with 1 and if we still see issues where the retry could help, add it later for mutations

PR #1837, opened 2021-10-13
Add condition to check global registry and skip build

teresaromero:
Re-opening PR due to branch name change.. (closed #1828)

Fixes https://github.com/okteto/app/issues/2812

Proposed changes 
- When okteto build if the image is already pushed to the global registry, the build is skiped

### DIFF
--- a/pkg/registry/registry.go
+++ b/pkg/registry/registry.go
@@ -221,13 +221,11 @@ func IsImageAtRegistry(image string) (ok bool) {
 			globalRegistryTag = TransformOktetoDevToGlobalRegistry(image)
 		}
 		if _, err := GetImageTagWithDigest(globalRegistryTag); err == nil {
-			log.Success("Skipping build: image is already built at the global registry")
-			log.Information("Force the build by using the flag --no-cache")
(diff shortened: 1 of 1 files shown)


ifbyol:
If we skip the build of an image `okteto.dev/golang:latest` because it is already present on `okteto.global/golang:latest` we need to translate all the references to the first one to the second one. I guess all that will be done in the webhook, but I think there is a case it wouldn't be covered, at least if we remove the `$OKTETO_GIT_COMMIT` check:
* `okteto up` using Okteto build service but deploying the pod on my own vanilla cluster. Since my cluster does not have the webhook, the image wouldn't be translated, so the pod would fail because the specified image does not exist, right? Adding the `$OKTETO_GIT_COMMIT` check would prevent this issue, but we have to bear it mind if we remove the check in the future.

teresaromero:
Refactored the buildWithOkteto so there is a flag wich indicates if the build has been skipped to show the logs accordingly. 
Current is showing like so:

For a skipped build
Screenshot 2021-10-19 at 10 52 53

For a not skipped build
Screenshot 2021-10-19 at 10 53 31

PR #1866, opened 2021-10-21
Add stack endpoints

jLopezbarb:

## Proposed changes
- Adds a command to show the endpoints generated by a stack
- Show the endpoints at the end of `okteto stack deploy --wait`


jLopezbarb:
@rlamana
Captura de pantalla 2021-11-03 a las 18 11 49

rberrelleza:
> @rlamana Captura de pantalla 2021-11-03 a las 18 11 49

I feel like that last part is too verbose. We show the stack name in both messages. Maybe the second one could just say "Endpoints available:"

rlamana:
@jLopezbarb Agree with Ramiro. Update the screenshot when ready.

jLopezbarb:
Captura de pantalla 2021-11-08 a las 12 47 33

rlamana:
@rberrelleza isn't it more correct to say "Available endpoints"?

rlamana (approved):

rberrelleza:
> @rberrelleza isn't it more correct to say "Available endpoints"?

Not sure tbh. Maybe we should just use `Endpoints:`?

jmacelroy:
> > @rberrelleza isn't it more correct to say "Available endpoints"?
> 
> Not sure tbh. Maybe we should just use `Endpoints:`?

They are both correct but I prefer `Endpoints available:` and think it's more common.

It's also listed on wikipedia as an example of postpositive adjective use :) It's one of those things I've never thought about but once I do I have no clue how English works, https://en.wikipedia.org/wiki/Postpositive_adjective, `Certain adjectives are used fairly commonly in postpositive position. Present and past participles exhibit this behavior, as in all those entering should ..., one of the men executed was ..., but at will this can be considered to be a verbal rather than adjectival use (a kind of reduced relative clause). Similar behavior is displayed by many adjectives with the suffix -able or -ible (e.g. the best room available, the only decision possible, the worst choice imaginable, the persons liable). `

jmacelroy (approved):

rlamana:
@jmacelroy English can be a bit of a nightmare when it comes to exceptions or special cases... 

In that case, I like the `Endpoints available: ` label.

pkg/cmd/stack/endpoints.go
```diff
@@ -0,0 +1,60 @@
...
+	"sort"
+	"strings"
+
+	"github.com/okteto/okteto/pkg/log"
+	"github.com/okteto/okteto/pkg/model"
+	"github.com/okteto/okteto/pkg/okteto"
+)
+
+func ListEndpoints(ctx context.Context, stack *model.Stack, output string) error {
+	oktetoClient, err := okteto.NewOktetoClient()
```
pchico83, line 29: @jLopezbarb this is wrong. We could list the ingresses instead and get the URL from there. That wway endpoints are available in vanilla kubernetes clusters too

PR #1905, opened 2021-11-02
Add context unit tests

jLopezbarb:

Adds unit tests for okteto contexts

## Proposed changes
- Add tests for okteto contexts
-


cmd/context/create_test.go
```diff
@@ -0,0 +1,248 @@
...
+	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/kubernetes/fake"
+	"k8s.io/client-go/rest"
+	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
+)
+
+type fakeK8sProvider struct {
```
jLopezbarb, line 32: Where should we put these structs? I've seen that other projects have their interface with their fake implementations on another package like `testing`

pkg/okteto/context_test.go
```diff
@@ -42,7 +45,8 @@ func Test_K8sContextToOktetoUrl(t *testing.T) {
 	ctx := context.Background()
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			if result := K8sContextToOktetoUrl(ctx, tt.in, "namespace"); result != tt.want {
+			fakeK8sProvider := test.NewFakeK8sProvider([]runtime.Object{})
```
ifbyol, line 48: Now that you are mocking the kubernetes client, you can add a suite case to cover the case in which the URL is extracted from the namespace annotation https://github.com/okteto/okteto/blob/master/pkg/okteto/context.go#L172

pkg/okteto/k8s.go
```diff
@@ -17,6 +17,20 @@ import (
...
+	Provide(clientApiConfig *clientcmdapi.Config) (kubernetes.Interface, *rest.Config, error)
+}
+
+type K8sClient struct{}
+
+func NewK8sClientProvider() *K8sClient {
+	return &K8sClient{}
+}
+
+func (c *K8sClient) Provide(clientApiConfig *clientcmdapi.Config) (kubernetes.Interface, *rest.Config, error) {
```
ifbyol, line 30:
```suggestion
func (_ *K8sClient) Provide(clientApiConfig *clientcmdapi.Config) (kubernetes.Interface, *rest.Config, error) {
```

cmd/context/create.go
```diff
@@ -132,3 +150,63 @@ func UseContext(ctx context.Context, ctxOptions *ContextOptions) error {
...
+	okteto.Context().IsOkteto = true
+
+	setSecrets(userContext.Secrets)
+
+	os.Setenv("OKTETO_USERNAME", okteto.Context().Username)
+
+	return nil
+}
+
+func (c *ContextUse) initKubernetesContext(ctxOptions *ContextOptions) error {
```
ifbyol, line 185: I think `c` is not being used

```suggestion
func (_ *ContextUse) initKubernetesContext(ctxOptions *ContextOptions) error {
```

pkg/okteto/k8s.go
```diff
@@ -38,7 +52,7 @@ func getKubernetesTimeout() time.Duration {
 	return timeout
 }
 
-func getK8sClient(clientApiConfig *clientcmdapi.Config) (*kubernetes.Clientset, *rest.Config, error) {
+func GetK8sClientWithApiConfig(clientApiConfig *clientcmdapi.Config) (*kubernetes.Clientset, *rest.Config, error) {
```
ifbyol, line 55: Is this really needed to be public? I think it can be `getK8sClient` is being called always internally I think

pkg/types/okteto_client.go
```diff
@@ -0,0 +1,21 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package types
+
+import "github.com/shurcooL/graphql"
+
```
ifbyol, line 17: Would it make sense to include also here the OktetoClientProvider?

ifbyol (approved):

pkg/okteto/context_test.go
jLopezbarb: I think this is the test you are referring:
https://github.com/okteto/okteto/pull/1905/files#diff-d20b02a443589d3c70663887f0ad6516dac079cb9baed19c795faecce59ecbd7R44

pkg/types/okteto_client.go
jLopezbarb: Yes, I think it's better to have it there 👍

pkg/okteto/context_test.go
ifbyol: oh!, yes, that is the test I was meaning, but as the logic belongs to this function, if we are adding a test here, probably makes sense to have that test also here. Anyways, it is fine if that

pchico83 (approved):

PR #1928, opened 2021-11-03
Jlopezbarb/fix depends on error

### LINKED ISSUE #1891: `depends_on` fails across stack manifests
**Describe the bug**
In a stack defined with multiple docker-compose manifests, `depends_on` fails if a service in one manifest depends on a service in a different manifest.

This seems to be because `depends_on` is validated during unmarshalling: https://github.com/okteto/okteto/blob/1474c7a166e9415db76d7a9486477530e6a6ccaa/pkg/model/stack_serializer.go#L273

**To Reproduce**
```
db.yml:
  services:
    db:
      image: mariadb

app.yml:
  services:
    app:
      image: nginx
      depends_on:
      - db
```

```
$ okteto stack deploy -f db.yml -f app.yml
 x   Service 'app' depends on service 'db' which is undefined.
```
 
**Expected behavior**
`depends_on` should work across files.

jLopezbarb:
Fixes #1891

## Proposed changes
- acídate depends on after merge all the files
-


pkg/model/stack.go
```diff
@@ -366,6 +369,29 @@ func validateStackName(name string) error {
...
+	for svcName, svc := range s.Services {
+		for dependentSvc, condition := range svc.DependsOn {
+			if svcName == dependentSvc {
+				return fmt.Errorf(" Service '%s' depends can not depend of itself.", svcName)
+			}
+			if _, ok := s.Services[dependentSvc]; !ok {
+				return fmt.Errorf(" Service '%s' depends on service '%s' which is undefined.", svcName, dependentSvc)
+			}
+			if condition.Condition == DependsOnServiceCompleted && !s.Services[dependentSvc].IsJob() {
+				return fmt.Errorf(" Service '%s' is not a job. Please change the reset policy so that it is not always in service '%s' ", dependentSvc, dependentSvc)
```
pchico83, line 382: I don't understand this sentence:
```
Please change the reset policy so that it is not always in service
```

jLopezbarb: It means that they have set a `DependsOnServiceCompleted` with a service with a restart policy set to `always`. We only support that kind of condition with jobs where the restart policy can not be `always`.
It was the error that we had but it's not explained properly. Any suggestion?

pchico83: @rberrelleza ?? It should say `restart policy`, not `reset policy`

pchico83 (approved):

PR #1934, opened 2021-11-04
Save only on ctx commands

jLopezbarb:

## Proposed changes
- Save context only when its a context command
-


pchico83:
I would also save here:
https://github.com/okteto/okteto/blob/master/cmd/context/use.go#L104

Because it means the user is selecting the current context

pchico83 (approved):

PR #1939, opened 2021-11-04
fix compat issue with Okteto Enterprise

rberrelleza:
Fix compatibility issues with Okteto Enterprise

### DIFF
--- a/pkg/okteto/secrets.go
+++ b/pkg/okteto/secrets.go
@@ -91,6 +91,12 @@ func (c *OktetoClient) GetUserContext(ctx context.Context) (*UserContext, error)
 	}
 	err := c.Query(ctx, &query, variables)
 	if err != nil {
+		if strings.Contains(err.Error(), "Cannot query field \"globalNamespace\" on type \"me\"") {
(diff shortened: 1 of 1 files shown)


pkg/okteto/secrets.go
```diff
@@ -91,6 +91,12 @@ func (c *OktetoClient) GetUserContext(ctx context.Context) (*UserContext, error)
 	}
 	err := c.Query(ctx, &query, variables)
 	if err != nil {
+		if strings.Contains(err.Error(), "Cannot query field \"globalNamespace\" on type \"me\"") {
```
jmacelroy, line 94: I don't like using error strings, even within our own codebase, for logic redirection or any logic in that case. They should be considered opaque. 

As a hot patch it's fine though. I just mention it because I keep seeing the pattern come up in code review and have commented about it in those locations too.

jmacelroy (approved):

pkg/okteto/secrets.go
rberrelleza: agree. we use this pattern in a couple of other places (I copied it from there).

PR #1943, opened 2021-11-05
Fix action release

jLopezbarb:

Fixes: context action was not updating properly because not having master branch

## Proposed changes
- Get the branch you should push to for every action
-

### DIFF
--- a/update_actions.sh
+++ b/update_actions.sh
@@ -8,24 +8,24 @@ if [ -z "$VERSION" ]; then
         exit 1
 fi
 
-actionsRepos=(delete-namespace_master
-        build_master
-        destroy-preview_master
-        deploy-preview_master
-        deploy-stack_master
-        namespace_master
-        pipeline_master
-        push_master
-        create-namespace_master
-        destroy-pipeline_master
-        login_master
-        destroy-stack_master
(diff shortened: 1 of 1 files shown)


update_actions.sh
```diff
@@ -8,23 +8,25 @@ if [ -z "$VERSION" ]; then
...
-        namespace
-        pipeline
-        push
-        create-namespace
-        destroy-pipeline
-        login
-        destroy-stack
-        apply
-        context)
+actionsRepos=(delete-namespace_master
```
ifbyol, line 11: I think  having repo names with `-` and separating the branch with `_`  could be error prone and probably we would not realize easily. Can we use maybe a tab or another character to separate repo name and branch? I would try to use any char that cannot be "valid" for a repo name

jLopezbarb: yep, we can't use tab because it will detect as another element, but we can use `/`. What do you think?

pchico83 (approved):

update_actions.sh
ifbyol: yeah, I don't think we create a repository with `/`

ifbyol (approved):

PR #1947, opened 2021-11-05
Make the OKTETO_URL configurable in integration tests

pchico83:

## Proposed changes

This makes it possible to run Okteto CLI integration tests on different environments


ifbyol (approved):

jLopezbarb:
I think this tests will fail when getting urls like in push action. https://github.com/okteto/okteto/blob/master/integration/push_test.go#L71

pchico83:
> I think this tests will fail when getting urls like in push action. https://github.com/okteto/okteto/blob/master/integration/push_test.go#L71

Good catch! I will update it

jLopezbarb (approved):

PR #1948, opened 2021-11-05
Add make integration-actions command

jLopezbarb:

Adds a make command to run only actions integration tests

## Proposed changes
- Created a new build tag for initialising user
- Add those tags when running tests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,7 +40,11 @@ test:
 
 .PHONY: integration
 integration:
-	 go test github.com/okteto/okteto/integration -tags=integration --count=1 -v -timeout 45m
+	 go test github.com/okteto/okteto/integration -tags="common integration actions" --count=1 -v -timeout 45m
+
+.PHONY: integration-actions
+integration-actions:
+	 go test github.com/okteto/okteto/integration -tags="common actions" --count=1 -v -timeout 45m
 
 .PHONY: build
 build:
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -40,7 +40,11 @@ test:
 
 .PHONY: integration
 integration:
-	 go test github.com/okteto/okteto/integration -tags=integration --count=1 -v -timeout 45m
+	 go test github.com/okteto/okteto/integration -tags="common integration actions" --count=1 -v -timeout 45m
```
pchico83, line 43: I think it should be fine with 15m timeout

pchico83 (approved):

integration/actions_test.go
```diff
@@ -1,5 +1,5 @@
-//go:build integration
-// +build integration
+//go:build actions
```
pchico83, line 1: could you add an envvar to skip the execution of the `context` action?
that why we can automate the execution of that tests only if the CLI version under test support the `context` command

integration/actions_test.go
```diff
@@ -205,6 +205,10 @@ func TestContextAction(t *testing.T) {
 		t.Skip("this test is not required for windows e2e tests")
 		return
 	}
+	if os.Getenv("OKTETO_SKIP_CTX_TEST") != "" {
```
pchico83, line 208: Can we call it `OKTETO_SKIP_CONTEXT_TEST` to make it more verbouse?

PR #1953, opened 2021-11-08
Adds support for private endpoint annotation

### LINKED ISSUE #1895: Support the "dev.okteto.com/private: true" annotation for stack services
I want to be able to mark one or more services in my stack as private, without having to define an endpoint object, or create any kubernetes resources. 

Something like this would be ideal, since it makes my manifest compliant with compose, but I can still take advantage of an okteto feature:

```
name: hello-world
services:
  hello-world:
    image: okteto.dev/hello-world
    labels:
      dev.okteto.com/private: "true"
    build: .
    ports:
    - 8080:8080
```

Today, this is possible by using the 'dev.okteto.com/auto-ingress: true' label, but I think that is a leaky abstraction, since ingress is a kubernetes thing.

jLopezbarb:

Fixes #1895 

## Proposed changes
-Adds support for okteto dev private annotation
-

### DIFF
--- a/pkg/cmd/stack/translate.go
+++ b/pkg/cmd/stack/translate.go
@@ -561,6 +561,9 @@ func translateService(svcName string, s *model.Stack) *apiv1.Service {
 	annotations := translateAnnotations(svc)
 	if s.Services[svcName].Public && annotations[model.OktetoAutoIngressAnnotation] == "" {
 		annotations[model.OktetoAutoIngressAnnotation] = "true"
+		if annotations[model.OktetoPrivateSvcAnnotation] == "true" {
+			annotations[model.OktetoAutoIngressAnnotation] = "private"
+		}
 	}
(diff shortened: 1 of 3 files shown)


pkg/cmd/stack/translate.go
```diff
@@ -561,6 +561,9 @@ func translateService(svcName string, s *model.Stack) *apiv1.Service {
 	annotations := translateAnnotations(svc)
 	if s.Services[svcName].Public && annotations[model.OktetoAutoIngressAnnotation] == "" {
```
pchico83, line 562: This should work even if `s.Services[svcName].Public` is false, no?

jLopezbarb: We set it to true when there are exposed ports. This way we could support okteto-stack previous implementation

pchico83 (approved):

PR #1957, opened 2021-11-09
Improved error message for invalid sync path

### LINKED ISSUE #1817: Clear error message and hint when a single file is specified in sync section
[`sync` section](https://okteto.com/docs/reference/manifest/#sync-string-required) for Okteto manifest only supports folders but in the documentation example is also specifying  single file to synchronize. The error I got is a bit confusing `open /home/user/.ssh/id_rsa/.stignore: not a directory`

I would like to receive a clear error message indicating that single files are not supported, only folders, and other fields like `secrets` or `volumes` could fit better

**Additional context**
There was a user with this issue in the #okteto slack channel

Youngestdev:
Fixes #1817

## Proposed changes

- Return an improved message to notify the user to crosscheck and update the directory path in the `sync` field in the okteto manifest.


cmd/up/stignore.go
```diff
@@ -27,6 +27,7 @@ import (
 	initCMD "github.com/okteto/okteto/cmd/init"
 	"github.com/okteto/okteto/cmd/utils"
 	"github.com/okteto/okteto/pkg/config"
+	o_error "github.com/okteto/okteto/pkg/errors"
```
jmacelroy, line 30:
```suggestion
	oktetoError "github.com/okteto/okteto/pkg/errors"
```

Convention for Go is to not have underscores in variable names.

Youngestdev: Ah, that's true 😅. Thank you!

cmd/up/stignore.go
```diff
@@ -41,7 +42,10 @@ func addStignoreSecrets(dev *model.Dev) error {
 		}
 		infile, err := os.Open(stignorePath)
 		if err != nil {
-			return err
+			return o_error.UserError{
```
jmacelroy, line 45: I didn't provide a suggestion change for this and now it's broken. You'll want to test this after applying.

```suggestion
			return oktetoError.UserError{
```

Youngestdev:
Thank you @jmacelroy, all set!

jLopezbarb:
Can you check if it's a directory creating a function here: https://github.com/okteto/okteto/blob/a01729016ce8c6ee803081f5080b3c0efc228335/pkg/model/dev.go#L638
That way we can test the function and have all the validation functions grouped

Youngestdev:
> Can you check if it's a directory creating a function here:
> 
> https://github.com/okteto/okteto/blob/a01729016ce8c6ee803081f5080b3c0efc228335/pkg/model/dev.go#L638
> 
> 
> That way we can test the function and have all the validation functions grouped

This?

```go
	file, err := os.Stat(path)
	if err != nil {
		return false, err
	}
	return file.IsDir(), nil
```

When I did, there were a couple of errors at test time. I'll look into it again.

pkg/model/dev.go
```diff
@@ -635,6 +639,17 @@ func (dev *Dev) setTimeout() error {
 	return nil
 }
 
+func (dev *Dev) validateSync() error {
+	for _, folder := range dev.Sync.Folders {
+		validPath, err := os.Stat(folder.LocalPath)
+		if !validPath.IsDir() && err != nil {
+			return err
```
jLopezbarb, line 646: If err == nil and !validatePath.isDir() there is no error but it should fail because if it's a file its forbidden.
I would return a user error saying that you can only use folders in sync and a hint to https://okteto.com/docs/reference/manifest/#sync-string-required

Youngestdev: So this instead:

```go
if !valudPath.IsDir() && err == nil {}
```

pkg/model/dev.go
```diff
@@ -635,6 +639,17 @@ func (dev *Dev) setTimeout() error {
 	return nil
 }
 
+func (dev *Dev) validateSync() error {
```
jmacelroy, line 642: There should be some tests added for this function. It would cover the case mentioned by Javi.

Youngestdev: Sure. Tests will be added after I fix an underlying issue. 

Javier & I are on that!

Youngestdev: This function is now outdated. I have added the validation to the `validat()` itself and modified the tests.

Youngestdev:
Thanks for the help via the 1:1 @jLopezbarb !

pkg/model/dev.go
```diff
@@ -695,6 +700,32 @@ func (dev *Dev) validate() error {
...
+		if err == nil {
+			if !validPath.IsDir() {
+				return oktetoError.UserError{
+					E:    fmt.Errorf("File paths are not supported on sync fields"),
+					Hint: "Update the `sync` field in your okteto manifest file to a valid directory path.",
+				}
+			}
+		} else if errors.Is(err, os.ErrNotExist) {
+			return oktetoError.UserError{
+				E:    fmt.Errorf("`stignore` path does not exist"),
```
jLopezbarb, line 716: I think this can be any kind of path. It's not stignore path
```suggestion
				E:    fmt.Errorf("path '%s' does not exist", folder.LocalPath),
```

jLopezbarb (approved):

jmacelroy (approved):

pchico83 (approved):

PR #1962, opened 2021-11-11
Add user id to context config

jLopezbarb:

## Proposed changes
- Adds user id from deprecated token
- After upgrading it will create the required field on the first okteto command

### DIFF
--- a/cmd/context/use.go
+++ b/cmd/context/use.go
@@ -153,6 +153,8 @@ func getUserContext(ctx context.Context) (*okteto.UserContext, error) {
 	for retries <= 3 {
 		userContext, err := client.GetUserContext(ctx)
 
+		// If userID is not on context config file we add it and save it.
+		// this prevents from relogin to actual users
 		if okteto.Context().UserID == "" {
 			okteto.Context().UserID = userContext.User.ID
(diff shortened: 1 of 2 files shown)


pchico83 (approved):

cmd/context/use.go
```diff
@@ -152,6 +152,14 @@ func getUserContext(ctx context.Context) (*okteto.UserContext, error) {
 	retries := 0
 	for retries <= 3 {
 		userContext, err := client.GetUserContext(ctx)
+
+		if okteto.Context().UserID == "" {
```
pchico83, line 156: could we add a comment explaining why we need this code?

ifbyol (approved):

cmd/context/use.go
```diff
@@ -152,6 +152,17 @@ func getUserContext(ctx context.Context) (*okteto.UserContext, error) {
 	retries := 0
 	for retries <= 3 {
 		userContext, err := client.GetUserContext(ctx)
+
+		// If userID is not on context config file we add it and save it.
+		// this prevents from relogin to actual users
+		if okteto.Context().UserID == "" {
```
rberrelleza, line 158: Should we only run this if it's an okteto context?

PR #1967, opened 2021-11-12
Solves #1966. Disable HTTP/2 when a SPDY request is detected in the okteto deploy proxy

### LINKED ISSUE #1966: Portforward request with empty body
**Describe the bug**
While using `okteto deploy` with a okteto-stack.yaml the pipeline fails when deploying the mongo instance.

**To Reproduce**

1. using this example https://github.com/teresaromero/movies/tree/pipelines-optimization
2. run okteto deploy 
3. the command starts to build and deploy services but when the mongodb service is started, it throws and error repetly 
```
REQUEST /api/v1/namespaces/teresa/pods/mongodb-0/portforward
METHOD POST
ERRO[0183] could not unmarshal request: unexpected end of JSON input
```

**Expected behavior**
The service is deployed and the pipeline ends correctly, the same as `okteto stack deploy` is run.

- OS: macOS
- Version Monterrey 12.0.1

**Additional context**

ifbyol:

Fixes #1966 

## Proposed changes

- When a SPDY request is detected in the proxy, I create a reverse proxy with HTTP/2 disabled and forward the request using that proxy. By default, the TLS config for *rest.Config uses HTTP/2 first which is incompatible with SPDY

### DIFF
--- a/pkg/k8s/namespaces/namespace.go
+++ b/pkg/k8s/namespaces/namespace.go
@@ -32,6 +32,7 @@ import (
 const (
 	parallelism int64 = 10
 	volumeKind        = "PersistentVolumeClaim"
+	jobKind           = "Job"
 )
 
 // DeleteAllOptions options for delete all operation
@@ -103,6 +104,13 @@ func (n *Namespaces) DestroyWithLabel(ctx context.Context, ns string, opts Delet
 			return err
 		}
 		deleteOpts := metav1.DeleteOptions{}
+
+		// It seems that by default, client-go don't delete pods scheduled by jobs, so we need to set the propation policy
+		if gvk.Kind == jobKind {
+			deletePropagation := metav1.DeletePropagationBackground
+			deleteOpts.PropagationPolicy = &deletePropagation
+		}
+
 		err = n.dynClient.
 			Resource(mapping.Resource).
 			Namespace(ns).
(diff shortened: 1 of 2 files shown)


pkg/k8s/namespaces/namespace.go
```diff
@@ -103,6 +104,13 @@ func (n *Namespaces) DestroyWithLabel(ctx context.Context, ns string, opts Delet
 			return err
 		}
 		deleteOpts := metav1.DeleteOptions{}
+
+		// It seems that by default, client-go don't delete pods scheduled by jobs, so we need to set the propation policy
+		if gvk.Kind == jobKind {
+			deletePropagation := metav1.DeletePropagationBackground
+			deleteOpts.PropagationPolicy = &deletePropagation
+		}
```
ifbyol, line 112: @pchico83 with the example provided by Tere in the issue, the pod generated by a Job is not being deleted. As it is created by the job controller, it does not have the `deployed-by` label, and it seems that by default, client-go don't include the propagation policy (while `kubectl` does). So I'm including it in case of jobs (I see we already have this propagation policy on [our code](https://github.com/okteto/okteto/blob/master/pkg/k8s/jobs/crud.go#L60)). For now I just added it for Jobs, should I include it for other Kinds?

pchico83: @ifbyol nothing else comes to my mind

pchico83 (approved):

jLopezbarb (approved):

PR #1987, opened 2021-11-17
Add wake and autocreate e2e

jLopezbarb:
## Proposed changes
- Adds an e2e test for auto wake feature
- Adds an e2e test for autocreate

### DIFF
--- a/cmd/utils/dev.go
+++ b/cmd/utils/dev.go
@@ -253,7 +253,7 @@ func GetApp(ctx context.Context, dev *model.Dev, c kubernetes.Interface, isRetry
 			return nil, false, err
 		}
 		if dev.Autocreate {
-			if isRetry && isAppDown(ctx, dev, c) {
+			if isRetry && !doesAppExist(ctx, dev, c) {
 				return nil, false, fmt.Errorf("Development container has been deactivated")
 			}
 			return apps.NewDeploymentApp(deployments.Sandbox(dev)), true, nil
@@ -280,7 +280,7 @@ func GetApp(ctx context.Context, dev *model.Dev, c kubernetes.Interface, isRetry
 	return app, false, nil
 }
 
-func isAppDown(ctx context.Context, dev *model.Dev, c kubernetes.Interface) bool {
+func doesAppExist(ctx context.Context, dev *model.Dev, c kubernetes.Interface) bool {
 	autocreateDev := *dev
 	autocreateDev.Name = model.DevCloneName(dev.Name)
 	_, err := apps.Get(ctx, &autocreateDev, dev.Namespace, c)
--- a/cmd/utils/dev.go
+++ b/cmd/utils/dev.go
@@ -284,5 +284,5 @@ func doesAppExist(ctx context.Context, dev *model.Dev, c kubernetes.Interface) b
 	autocreateDev := *dev
 	autocreateDev.Name = model.DevCloneName(dev.Name)
 	_, err := apps.Get(ctx, &autocreateDev, dev.Namespace, c)
-	return err != nil
+	return err == nil
 }
--- a/cmd/utils/dev.go
+++ b/cmd/utils/dev.go
@@ -253,7 +253,7 @@ func GetApp(ctx context.Context, dev *model.Dev, c kubernetes.Interface, isRetry
 			return nil, false, err
 		}
 		if dev.Autocreate {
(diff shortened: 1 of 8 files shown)


integration/auto-wake_test.go
```diff
@@ -0,0 +1,209 @@
...
+	}
+
+	var wg sync.WaitGroup
+	upErrorChannel := make(chan error, 1)
+	p, err := up(ctx, &wg, namespace, name, manifestPath, oktetoPath, upErrorChannel)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	content, err = getContent(endpoint, 150, nil)
```
pchico83, line 146: I wouldn't call either `getContent` or `down`. Just up should wake up everything

cmd/utils/dev.go
```diff
@@ -246,13 +246,16 @@ func GetDownCommand(devPath string) string {
...
 
-func GetApp(ctx context.Context, dev *model.Dev, c kubernetes.Interface) (apps.App, bool, error) {
+func GetApp(ctx context.Context, dev *model.Dev, c kubernetes.Interface, isRetry bool) (apps.App, bool, error) {
 	app, err := apps.Get(ctx, dev, dev.Namespace, c)
 	if err != nil {
 		if !errors.IsNotFound(err) {
 			return nil, false, err
 		}
 		if dev.Autocreate {
+			if isRetry && isAppDown(ctx, dev, c) {
```
pchico83, line 256: why do we need to check this here?

rberrelleza:
111

cmd/utils/dev.go
jLopezbarb: That checks that we didn't lose the connection to the development environment.
For example if an user deletes the pod we have to recreate we have the deployment up but if an user runs `okteto down` we don't have the deployment up so we check that we don't have to recreate the deployment

integration/auto-wake_test.go
jLopezbarb: +1 to not call getContent, but if you don't call down there will be a zombie process running on background that is using local port 8080

pchico83: we can call it after checking the namespace is active

cmd/utils/dev.go
pchico83: should it be `doesntexist` then? instead of `appdown`?

integration/auto-wake_test.go
```diff
@@ -0,0 +1,234 @@
...
+	for i := 0; i < timeout; i++ {
+		<-t.C
+		var d *appsv1.Deployment
+		var err error
+		if isDevMode {
+			d, err = getDeployment(ctx, namespace, model.DevCloneName(name))
+			if err != nil {
+				log.Printf("error getting deployment: %s", err.Error())
+				continue
+			}
```
ifbyol, line 206: minor thing. To simplify the if/else branch logic, you could move this after the if/else as err it is already defined outside the conditional

pkg/okteto/namespace.go
```diff
@@ -133,3 +133,21 @@ func validateNamespace(namespace, object string) error {
 	}
 	return nil
 }
+
+// SleepNamespace sleeps a namespace
+func (c *OktetoClient) SleepNamespace(ctx context.Context, namespace string) error {
```
ifbyol, line 138: I would try to avoid as much as possible to include functions that are only for tests in the "production" code.

cmd/utils/dev.go
```diff
@@ -276,3 +279,10 @@ func GetApp(ctx context.Context, dev *model.Dev, c kubernetes.Interface) (apps.A
 	}
 	return app, false, nil
 }
+
+func doesAutocreateAppExist(ctx context.Context, dev *model.Dev, c kubernetes.Interface) bool {
+	autocreateDev := *dev
+	autocreateDev.Name = model.DevCloneName(dev.Name)
+	_, err := apps.Get(ctx, &autocreateDev, dev.Namespace, c)
+	return err == nil
```
ifbyol, line 287: This is a general assumption I see we are doing in several places, even in the backend. The error could be related to any Kubernetes API error o timeout. A temporal error in K8s could make the tests fail. Should we retry if the error != nil and it isn't a not found error? I think that would make a bit more fault-tolerant the test

ifbyol (approved):

Issue #1998, opened 2021-11-18
Add autocomplete commands for fig


rberrelleza:
**Is your feature request related to a problem? Please describe.**
[Fig](https://fig.io) is a really cool CLI tool that makes autocompleting commands a lot easier. It would be great way for our community to discover new commands. e.g. we could make is to that running `okteto context use` and then pressing `tab` shows you all the contexts available, or things like that. 

Check out their kubernetes integration, I think following that path would be great.

rberrelleza:
Missing work:
- Add autocomplete for context lists and namespaces
- Automatically send a PR to fig on every release

RinkiyaKeDad:
@rberrelleza does this look good? (we can add the Okteto logo too for the results once we upload it to fig)


https://user-images.githubusercontent.com/56963264/151990180-df0a3ebb-6c46-4fdf-932d-2588ac042209.mov

rberrelleza:
Yeah, that looks pretty good!  One thing to keep in mind is if we can do some sort of caching so that the result shows up faster (since people don't create namespaces that often)

RinkiyaKeDad:
Yup, I couldn't find anything related to caching on their docs - let me raise a PR and ask them their.

PR #1999, opened 2021-11-19
add command to generate fig descriptor

rberrelleza:
This gets the first step for  #1998


ifbyol (approved):

pchico83 (approved):
We should make it intelligent to autocomplete things like contexts or okteto manifest files

rberrelleza:
Yes, that can be phase 2 if we see adoption there (no idea how complex that would be). The next release of fig will include support for okteto out of the box.

PR #2009, opened 2021-11-23
Deploy with different outputs

jLopezbarb:

Fixes #

## Proposed changes
- Shows last 10 lines of the current command
- When the command is finished check it with success message
- Add the option to show in plain mode

## Screenshots:
- TTY deploying
Captura de pantalla 2021-11-23 a las 15 38 38
- TTY successful
Captura de pantalla 2021-11-23 a las 15 40 39

- Plain
Captura de pantalla 2021-11-23 a las 15 40 01


cmd/deploy/deploy.go
```diff
@@ -160,6 +161,7 @@ func Deploy(ctx context.Context) *cobra.Command {
 	cmd.Flags().StringVar(&options.Name, "name", "", "application name")
 	cmd.Flags().StringVarP(&options.ManifestPath, "file", "f", "", "path to the manifest file")
 	cmd.Flags().StringArrayVarP(&options.Variables, "var", "v", []string{}, "set a variable (can be set more than once)")
+	cmd.Flags().StringVarP(&options.OutputMode, "output", "", "tty", "show plain/tty deploy output")
```
ifbyol, line 164: If we would deploy first the installer using the `okteto deploy` command and then, we merge this and update the cloudbin image, the installer could use tty output, which I think is not the best. We have to coordinate how to do this

cmd/utils/manifest.go
```diff
@@ -268,17 +276,103 @@ func (*Executor) Execute(command string, env []string) error {
 	done := make(chan struct{})
```
ifbyol, line 276: I think this `done` channel can be deleted. It is not being used to check when the result is added to it. Probably I left it behind when I moved this logic from the installer

cmd/utils/manifest.go
```diff
@@ -268,17 +276,103 @@ func (*Executor) Execute(command string, env []string) error {
...
+	}
 
 	if err := cmd.Start(); err != nil {
 		return err
 	}
 
 	return cmd.Wait()
 }
+
+func displayPlainOutput(scanner *bufio.Scanner, done chan struct{}) {
```
ifbyol, line 292: as the executor logic is growing with these changes, would it make sense to create a `executor.go` file in the same package moving all the executor logic there?

I didn't do it when I created this file because the logic was too small, but probably it will grow more

maroshii:
We need to send output in json for the installer which is related to:
- https://github.com/okteto/pipeline-installer-image/pull/19
- https://github.com/okteto/pipeline-installer-image/pull/17

Is that something we should add here as `--output json`? Maybe we are already considering this but just to bring it to your attention.

ifbyol:
> We need to send output in json for the installer as well. Is that something we should add here as `--output json`? Maybe we are already considering this but just to bring it to your attention.

Yes, `json` would be another output format to read by the installer in the future

rlamana:
How is this aligned with the deploy stages that @maroshii is working on in the UI? I think the copy and grouping should be the same in the CLI and the UI.

jLopezbarb:
I have implemented the json output so the UI can do the same thing that the CLI is doing

rlamana:
Why is the output of the command being run not preserving its form and colors? are we removing the escape chars? Can we preserve it?

jLopezbarb:
> Why is the output of the command being run not preserving its form and colors? are we removing the escape chars? Can we preserve it?

It's running on another shell and exporting the messages without any color. We can not get the original colours from there, but we could change the colours based on our logs, so if we encounter a line starting with " i " set it on blue, " x " set it red and " ✔ " set it green

rlamana:
I don't like that workaround, it would be hard to maintain and won't work for other outputs. When running a pipeline the developer could launch any command. Any way we can preserve the output intact? @jLopezbarb @pchico83

jLopezbarb:
we could add colors if we fake that the terminal where it is executed supports colors. By doing this we should remove the collapsing when the command is finished or making the build not run in tty mode as it makes everything fall out of place. 
- Build tty
Captura de pantalla 2021-11-24 a las 12 53 31
- Build plain 
Captura de pantalla 2021-11-24 a las 12 57 59
- Build plain after finishing all builds
Captura de pantalla 2021-11-24 a las 12 59 01

cmd/utils/executor.go
```diff
@@ -0,0 +1,188 @@
...
+
+type Executor struct {
+	outputMode string
+}
+
+type jsonMessage struct {
+	Level     string    `json:"level"`
+	Stage     string    `json:"stage"`
+	Message   string    `json:"message"`
+	Timestamp time.Time `json:"timestamp"`
```
maroshii, line 45: this should be `int64` converted from `.Unix()`

cmd/utils/executor.go
```diff
@@ -58,8 +58,10 @@ func NewExecutor(output string) *Executor {
 }
 
 func IsSupportForTTY() bool {
-	_, err := unix.IoctlGetWinsize(int(os.Stdout.Fd()), unix.TIOCGWINSZ)
-	return err == nil
+	if runtime.GOOS != "windows" {
```
ifbyol, line 61: Isn't this checked in the file compiled only for windows? Why would we need to add also this check here?

cmd/deploy/deploy.go
ifbyol: talked offline. We'll change the installer to send the right output option when we update the CLI in the backend

cmd/utils/executor.go
```diff
@@ -0,0 +1,236 @@
...
+	cmd.Env = append(os.Environ(), env...)
+
+	var scanner *bufio.Scanner
+
+	if e.outputMode == "tty" {
+		var (
+			reader io.Reader
+			err    error
+		)
+		if runtime.GOOS != "windows" {
```
ifbyol, line 80: What would happen if someone in windows executes the deploy command which by default is having `tty` option? If I understood it correctly, command wouldn't be executed, right? Probably I'm missing something

cmd/utils/executor.go
```diff
@@ -0,0 +1,236 @@
...
+	return err
+}
+
+func displayPlainOutput(scanner *bufio.Scanner) {
+	for scanner.Scan() {
+		line := scanner.Text()
+		fmt.Println(line)
+	}
+}
+func displayJSONOutput(scanner *bufio.Scanner, command string) {
```
ifbyol, line 121: @maroshii FYI

cmd/utils/executor.go
```diff
@@ -0,0 +1,236 @@
...
+
+func IsSupportForTTY() bool {
+	if runtime.GOOS != "windows" {
+		return isSupportForTTY()
+	}
+	return false
+}
+
+// Execute executes the specified command adding `env` to the execution environment
+func (e *Executor) Execute(command string, env []string) error {
```
ifbyol, line 68: Would it make sense to create different executors depending on the output? I think it would make the logic on each executor cleaner that having several ifs in the execute code. They would have a common interface but some auxiliar func like `displayTTYOutput` would be only present for the TTY executor.

When a new executor is requested, we can create the one according to the output mode specified by the user

cmd/utils/executor.go
jLopezbarb: they will have the same experience as in Linux but they won't have any color from the output

ifbyol: But, I mean, If the option is `tty` and the OS is windows, we are not calling to the command execution, right?. We skip the call to `startCommandTTY` and we are not calling to `cmd.Start` because that is done in the else branch. Am I missing something?

cmd/utils/executor-unix.go
```diff
@@ -0,0 +1,39 @@
+//go:build !windows
```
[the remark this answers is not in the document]
jLopezbarb: TTY is supported on windows but the colours are not shown because the terminal where we are running the commands does not support the colours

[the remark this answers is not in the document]
jLopezbarb: That means that we can display the last 10 lines but without colour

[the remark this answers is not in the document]
ifbyol: I meant in the code, to remember why this is not for windows in the future

cmd/utils/executor.go
```diff
@@ -0,0 +1,247 @@
...
+func (jsonExecutorDisplayer) startCommand(cmd *exec.Cmd) (io.Reader, error) {
+	reader, err := cmd.StdoutPipe()
+	if err != nil {
+		return nil, err
+	}
+
+	if err := cmd.Start(); err != nil {
+		return nil, err
+	}
+	return reader, nil
```
ifbyol, line 130: This logic can be extracted to a common function to be shared by `jsonExecutorDisplayer`, `plainExecutorDisplayer` and `ttyExecutorDisplayer` for windows

cmd/utils/executor.go
```diff
@@ -0,0 +1,243 @@
...
+	if output == "tty" && !isSupportForTTY() {
+		output = "plain"
+	}
+	var displayer executorDisplayer
+	switch output {
+	case "tty":
+		displayer = ttyExecutorDisplayer{}
+	case "plain":
+		displayer = plainExecutorDisplayer{}
+	case "json":
```
pchico83, line 70: is `json` supported already? I think we need to update the help messages then

cmd/utils/executor.go
```diff
@@ -0,0 +1,243 @@
...
+	}
+}
+
+func (ttyExecutorDisplayer) display(scanner *bufio.Scanner, command string, sb *screenbuf.ScreenBuf) {
+	queue := []string{}
+	for scanner.Scan() {
+		commandLine := renderCommand(command)
+		sb.Write(commandLine)
+		line := scanner.Text()
+		if len(queue) == 10 {
```
pchico83, line 153: can we do this value configurable via a command argument and envvar? also, is i possible to default to 25?

cmd/deploy/deploy.go
```diff
@@ -160,6 +161,7 @@ func Deploy(ctx context.Context) *cobra.Command {
 	cmd.Flags().StringVar(&options.Name, "name", "", "application name")
 	cmd.Flags().StringVarP(&options.ManifestPath, "file", "f", "", "path to the manifest file")
 	cmd.Flags().StringArrayVarP(&options.Variables, "var", "v", []string{}, "set a variable (can be set more than once)")
+	cmd.Flags().StringVarP(&options.OutputMode, "output", "", "plain", "show plain/json deploy output")
```
pchico83, line 164:
```suggestion
	cmd.Flags().StringVarP(&options.OutputMode, "output", "o", "plain", "show plain/json deploy output")
```

cmd/destroy/destroy.go
```diff
@@ -121,6 +122,7 @@ func Destroy(ctx context.Context) *cobra.Command {
 
 	cmd.Flags().StringVar(&options.Name, "name", "", "application name")
 	cmd.Flags().StringVarP(&options.ManifestPath, "file", "f", "", "path to the manifest file")
+	cmd.Flags().StringVarP(&options.OutputMode, "output", "", "tty", "show plain/tty deploy output")
```
pchico83, line 125:
```suggestion
	cmd.Flags().StringVarP(&options.OutputMode, "output", "o", "tty", "show plain/tty deploy output")
```

pchico83 (approved):
Not sure if it's related to this PR, but I see two issues:
- CTRL+C shows an error, it should just exit with no error
- The current context is not shown at the beginning of the `deploy/destroy` command:
```
 i  Using pchico83 @ cloud.okteto.com as context
```
As we for other commands like `okteto stack deploy`. Btw, `CTRL+C` also shows an error in `okteto stack deploy`

ifbyol (approved):

jLopezbarb:
> Not sure if it's related to this PR, but I see two issues:
> 
> * CTRL+C shows an error, it should just exit with no error
> * The current context is not shown at the beginning of the `deploy/destroy` command:
> 
> ```
>  i  Using pchico83 @ cloud.okteto.com as context
> ```
> 
> As we for other commands like `okteto stack deploy`. Btw, `CTRL+C` also shows an error in `okteto stack deploy`

Both has been fixed in master, I've tested after I did the merge and was working properly

PR #2013, opened 2021-11-24
Move envvar to constants

jLopezbarb:
## Proposed changes
- Moves all env var to constant file so we can check all of them by looking at just one file
-

### DIFF
--- a/pkg/model/const.go
+++ b/pkg/model/const.go
@@ -180,4 +180,106 @@ const (
 
 	// HelmSecretType indicates the type for secrets created by Helm
 	HelmSecretType = "helm.sh/release.v1"
+
+	// OktetoBinEnvVar defines the okteto binary that should be used
+	OktetoBinEnvVar = "OKTETO_BIN"
+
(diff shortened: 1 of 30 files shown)


pkg/model/const.go
```diff
@@ -180,4 +180,106 @@ const (
...
 	HelmSecretType = "helm.sh/release.v1"
+
+	// OktetoBinEnvVar defines the okteto binary that should be used
+	OktetoBinEnvVar = "OKTETO_BIN"
+
+	// OktetoSkipCleanupEnvVar defines the okteto binary that should be used
+	OktetoSkipCleanupEnvVar = "OKTETO_SKIP_CLEANUP"
+
+	// OktetoUserEnvVar defines the user is using okteto
+	OktetoUserEnvVar = "OKTETO_USER"
```
jLopezbarb, line 191: Is this one the same as `okteto_username`?

pchico83: I think this is only used for defining the user of our integration tests

pchico83 (approved):

PR #2016, opened 2021-11-25
Rewrite docker daemon not running error

### LINKED ISSUE #1990: Show a user friendly error when docker is not running
**Describe the bug**
It would be nice if we can show a more user-friendly error when docker is not running, and I'm running a local "okteto build".

Currently we log:
```
Failed to dial gRPC: cannot connect to the Docker daemon. Is 'docker daemon' running on this host?: dial unix /var/run/docker.sock: connect: no such file or directory
```

Could we parse the error and simply mention something like "docker is not running on your host", and maybe add a hint on what to do?

- Version 1.14.6

jLopezbarb:

Fixes #1990

## Proposed changes
- Change the error to a user friendly error
-

### DIFF
--- a/pkg/cmd/build/build.go
+++ b/pkg/cmd/build/build.go
@@ -141,12 +141,12 @@ func buildWithDocker(ctx context.Context, buildOptions BuildOptions) error {
 	if versions.GreaterThanOrEqualTo(cli.ClientVersion(), "1.39") {
 		err = buildWithDockerDaemonBuildkit(ctx, buildOptions, cli)
 		if err != nil {
-			return err
+			return translateDockerErr(err)
 		}
 	} else {
 		err = buildWithDockerDaemon(ctx, buildOptions, cli)
 		if err != nil {
-			return err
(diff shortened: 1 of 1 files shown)


pkg/cmd/build/build.go
```diff
@@ -168,3 +168,16 @@ func validateImage(imageTag string) error {
...
 	return nil
 }
+
+func translateDockerErr(err error) error {
+	if err == nil {
+		return nil
+	}
+	if strings.HasPrefix(err.Error(), "failed to dial gRPC: cannot connect to the Docker daemon") {
+		return okErrors.UserError{
+			E:    fmt.Errorf("cannot connect to Docker Daemon"),
```
rberrelleza, line 178: could we add the address of the daemon? (or is this always for the local one?)

jLopezbarb: This is just for the local docker daemon

rberrelleza (approved):

PR #2020, opened 2021-11-25
Fix okteto dev registry replacement

jLopezbarb:

Fixes okteto was failing to build when Okteto dev label was in the registry url

## Proposed changes
- Check if it's already a registry URL or not
- If it's not do the replacement

### DIFF
--- a/pkg/registry/registry.go
+++ b/pkg/registry/registry.go
@@ -17,6 +17,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/url"
+	"regexp"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


pchico83 (approved):

ifbyol (approved):

pkg/registry/registry.go
```diff
@@ -196,5 +197,10 @@ func IsOktetoRegistry(tag string) bool {
 }
 
 func replaceRegistry(input, registryType, namespace string) string {
-	return strings.Replace(input, registryType, fmt.Sprintf("%s/%s", okteto.Context().Registry, namespace), 1)
+	// Check if the registryType is the start of the sentence or has a whitespace before it
+	var re = regexp.MustCompile(fmt.Sprintf(`(^|\s)(%s)`, registryType))
```
jmacelroy, line 201: The regex compilation should probably be moved to package level. This will both cause the panic before the function is called, making detecting an issue easier, and save some performance by only compiling the regex once.

PR #2027, opened 2021-11-26
Check syncthing status using the "rest/events" API

pchico83:

## Proposed changes

`/rest/db/status` is reported as an expensive API call:
https://docs.syncthing.net/rest/db-status-get.html

### DIFF
--- a/pkg/syncthing/syncthing.go
+++ b/pkg/syncthing/syncthing.go
@@ -104,7 +104,6 @@ type Folder struct {
 	Name        string `yaml:"name"`
 	LocalPath   string `yaml:"localPath"`
 	RemotePath  string `yaml:"remotePath"`
-	Retries     int    `yaml:"-"`
 	Overwritten bool   `yaml:"-"`
 }
 
--- a/pkg/syncthing/syncthing.go
+++ b/pkg/syncthing/syncthing.go
@@ -42,7 +42,8 @@ import (
 )
 
 var (
-	configTemplate = template.Must(template.New("syncthingConfig").Parse(configXML))
+	configTemplate   = template.Must(template.New("syncthingConfig").Parse(configXML))
+	isHealthyRetries = 0
 )
 
 const (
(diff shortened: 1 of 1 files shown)


pkg/syncthing/syncthing.go
```diff
@@ -486,68 +507,108 @@ func (s *Syncthing) GetCompletion(ctx context.Context, local bool, device string
...
 // GetFolderErrors returns the last folder errors
-func (s *Syncthing) GetFolderErrors(ctx context.Context, folder *Folder, local bool) error {
-	params := getFolderParameter(folder)
-	params["since"] = "0"
-	params["limit"] = "1"
-	params["timeout"] = "0"
-	params["events"] = "FolderErrors"
-	folderErrorsList := []FolderErrors{}
+func (s *Syncthing) GetFolderErrors(ctx context.Context, local bool) error {
+	params := map[string]string{
```
ifbyol, line 605: Is it not needed anymore the folder parameter?

pkg/syncthing/syncthing.go
```diff
@@ -486,68 +507,108 @@ func (s *Syncthing) GetCompletion(ctx context.Context, local bool, device string
...
 
-		folder.Retries++
-		err = s.GetFolderErrors(ctx, folder, false)
-		if err != nil {
-			log.Infof("syncthing error in folder '%s' local=%t retry %d: %s", folder.RemotePath, local, folder.Retries, err.Error())
-		}
-		if err == errors.ErrInsufficientSpace {
-			return err
-		}
+	isHealthyRetries++
```
ifbyol, line 522: Why is it global now? Probably I'm missing context, I've never worked with syncthing

jLopezbarb (approved):

pkg/syncthing/syncthing.go
pchico83: Before, for each folder in sync, we were counting the number of retries for that folder. It was a global state, but now we can use a single variable because we are checking errors on any folder in a single query.

pkg/syncthing/syncthing.go
pchico83: With the events API we can query errors on any folder in a single query

PR #2035, opened 2021-11-30
mentioning logs' location when okteto up fails

### LINKED ISSUE #1004: Mention the location of the logs when okteto up fails
**Describe the bug**
Making it more visible  would help troubleshooting issues 

We could add them as a hint to syncthing-related errors.

RinkiyaKeDad:
Signed-off-by: RinkiyaKeDad <arshsharma461@gmail.com>

Fixes #1004 

## Proposed changes

- In the error which gets returned when `okteto up` fails, I've appended a message which points to the location of `okteto.log`

## Tests

I wasn't able to test this locally. Is there a way to force `okteto up` to fail?

### DIFF
--- a/cmd/up/up.go
+++ b/cmd/up/up.go
@@ -181,7 +181,7 @@ func Up() *cobra.Command {
 				log.Infof("error deleting deprecated volume: %v", err)
 			}
 
-			err = fmt.Errorf("%w; Additional logs can be found at: $HOME/.okteto/okteto.log", err)
+			err = fmt.Errorf("%w\n    Find additional logs at: %s/okteto.log", err, config.GetOktetoHome())
 
 			return err
 		},
--- a/cmd/up/up.go
+++ b/cmd/up/up.go
@@ -181,6 +181,8 @@ func Up() *cobra.Command {
 				log.Infof("error deleting deprecated volume: %v", err)
 			}
 
+			err = fmt.Errorf("%w; Additional logs can be found at: $HOME/.okteto/okteto.log", err)
+
 			return err
 		},
 	}
--- a/cmd/up/up.go
+++ b/cmd/up/up.go
@@ -181,7 +181,7 @@ func Up() *cobra.Command {
 				log.Infof("error deleting deprecated volume: %v", err)
 			}
 
-			err = fmt.Errorf("%w\n    Find additional logs at: %s/okteto.log", err, config.GetOktetoHome())
+			err = fmt.Errorf("%w\n    Find additional logs at: %s/okteto.log", err, config.GetAppHome(dev.Namespace, dev.Name))
 
 			return err
 		},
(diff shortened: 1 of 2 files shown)


pchico83:
@RinkiyaKeDad  could you attach a screenshot of how it looks like in the terminal?

RinkiyaKeDad:
I didn't know how to make the `okteto up` command fail on purpose so I changed the code to return an error always and this is how the final output looked in the terminal:

Screen Shot 2021-12-01 at 12 45 33 PM

pchico83:
> I didn't know how to make the `okteto up` command fail on purpose so I changed the code to return an error always and this is how the final output looked in the terminal:
> 
> Screen Shot 2021-12-01 at 12 45 33 PM

@rlamana @rberrelleza how would you show this tip?
I would go with an `log.Information` message in a new line. Also, I would expand `$HOME`  to its real value:
https://github.com/okteto/okteto/blob/master/pkg/config/config.go#L109

Also, I would do the same for `okteto down`

rlamana:
> > I didn't know how to make the `okteto up` command fail on purpose so I changed the code to return an error always and this is how the final output looked in the terminal:
> @rlamana @rberrelleza how would you show this tip? I would go with an `log.Information` message in a new line. Also, I would expand `$HOME` to its real value: https://github.com/okteto/okteto/blob/master/pkg/config/config.go#L109
> 
> And I would do the same for `okteto down`

I'd go for a new line too, but as part of the error message, and properly indented. Something like:

```
  x  Failed to load okteto context 'xxxx'.
     Find additional logs at: ~/.okteto/okteto.log
 ```

Trying to keep the message shorter. I'd also expand $HOME to its real value.

**Comment by RinkiyaKeDad** [comment]
Is this better? 🙂
image

File: cmd/up/up.go [hunk 1]
```diff
@@ -181,6 +181,8 @@ func Up() *cobra.Command {
 				log.Infof("error deleting deprecated volume: %v", err)
 			}
 
+			err = fmt.Errorf("%w\n    Find additional logs at: %s/okteto.log", err, config.GetOktetoHome())
```
RinkiyaKeDad, line 184: Kept 4 spaces here since the icon takes 3 and 1 extra before beginning the sentence.

RinkiyaKeDad: Also should I switch to `log.Information` or would this be fine?

pchico83:
> Is this better? 🙂 image

I like it! But I think the location is wrong, it should be `/Users/arsh/.okteto/<your-namespace>/<your-okteto-manifest-name>/okteto.log`

pchico83 (approved):

RinkiyaKeDad:
> I like it! But I think the location is wrong, it should be /Users/arsh/.okteto/<your-namespace>/<your-okteto-manifest-name>/okteto.log

Oh yes! I realized I had used the wrong function. Fixed it now :)

image

pchico83:
@RinkiyaKeDad could you add it `okteto down` too?

rlamana (approved):
Looks great!

RinkiyaKeDad:
> @RinkiyaKeDad could you add it okteto down too?


@pchico83 that had totally skipped my mind - added it now! :)

Screen Shot 2021-12-10 at 8 04 03 PM

PR #2037, opened 2021-12-01
Fix fluxCD annotations

### LINKED ISSUE #2029: Okteto behaviour of cloning the deployment (instead of modifying it) problematic with FluxCD
**Describe the bug**
Somewhere between version 1.12 and current, Okteto has changed how the development container is started. Now the original deployment is cloned with `-okteto` suffix. This causes severe trouble with GitOps approach (at least as used by FluxCD v1) - both the original deployment and development deployment serve incoming requests after `okteto up`.

**To Reproduce**

1. Run Kubernetes cluster with FluxCD v1
2. Deploy an application with FluxCD in a developer's namespace
3. Use `okteto up` with file like:
```
---
name: refrigerator-api-deployment
autocreate: false
namespace: rkujawa
container: refrigerator-api
metadata:
  annotations:
    fluxcd.io/ignore: "true"
workdir: /app
command: /usr/bin/flask run -h 0.0.0.0
environment:
  FLASK_ENV: development
remote: 2222
forward:
  - 5000:5000
sync:
  - src:/app
persistentVolume:
  enabled: false
```
4. Deployment that is watched by FluxCD gets cloned. The cloned (`-okteto`) deployment has `fluxcd.ignore` annotation set, as specified in `okteto.yml`. The original deployment does not.
5. Okteto scales down the original deployment.
6. FluxCD sees that original deployment state is different than within Git, and scales up the deployment again.
7. I end up with a service for this application that has two deployments as backends (original and one cloned by Okteto), so any requests that come into the service, gets distributed in a round-robin fashion between the original and development container.

**Expected behavior**
`okteto up` starts a development container that *replaces* the original container. Current behaviour breaks this with FluxCD.

- OS: Linux 
- Version: 1.14.8

**Additional context**
This makes Okteto unusable with Flux, I had to revert previously used version 1.12.15.

Is it possible to revert to the old behaviour of modifying existing deployment instead of cloning it (by means of a setting in `okteto.yml` or whatever other option)?

If not, is it possible to modify the original deployment, while doing `okteto up`, so that it gets `fluxcd.ignore` set automatically?

jLopezbarb:

Fixes #2029

## Proposed changes
- Set previous annotations on app
- Remove annotations on devModeOff
- Restore previous annotations on devModeOff
- Add test

### DIFF
--- a/pkg/k8s/apps/translate.go
+++ b/pkg/k8s/apps/translate.go
@@ -14,6 +14,7 @@
 package apps
 
 import (
+	"encoding/json"
 	"fmt"
 	"path"
 	"strconv"
@@ -56,12 +57,15 @@ func (tr *Translation) translate() error {
 	delete(tr.App.ObjectMeta().Annotations, model.StateBeforeSleepingAnnontation)
 
 	tr.DevApp = tr.App.DevClone()
+	annotationsString, _ := json.Marshal(tr.App.ObjectMeta().Annotations)
 
 	tr.App.ObjectMeta().Annotations[model.AppReplicasAnnotation] = strconv.Itoa(int(replicas))
+	tr.App.ObjectMeta().Annotations[model.AppAnnotationsAnnotation] = string(annotationsString)
 	tr.App.ObjectMeta().Labels[model.DevLabel] = "true"
 	tr.App.SetReplicas(0)
 
 	for k, v := range tr.Dev.Metadata.Annotations {
+		tr.App.ObjectMeta().Annotations[k] = v
 		tr.DevApp.ObjectMeta().Annotations[k] = v
 		tr.DevApp.TemplateObjectMeta().Annotations[k] = v
 	}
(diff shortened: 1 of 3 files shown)


pkg/k8s/apps/translate.go
```diff
@@ -56,12 +57,15 @@ func (tr *Translation) translate() error {
...
 	tr.DevApp = tr.App.DevClone()
+	annotationsString, _ := json.Marshal(tr.App.ObjectMeta().Annotations)
 
 	tr.App.ObjectMeta().Annotations[model.AppReplicasAnnotation] = strconv.Itoa(int(replicas))
+	tr.App.ObjectMeta().Annotations[model.AppAnnotationsAnnotation] = string(annotationsString)
 	tr.App.ObjectMeta().Labels[model.DevLabel] = "true"
 	tr.App.SetReplicas(0)
 
 	for k, v := range tr.Dev.Metadata.Annotations {
+		tr.App.ObjectMeta().Annotations[k] = v
```
pchico83, line 68: Let's do `tr.App.TemplateObjectMeta().ObjectMeta().Annotations[k] = v` too

pkg/k8s/apps/translate.go
```diff
@@ -56,12 +57,15 @@ func (tr *Translation) translate() error {
 	delete(tr.App.ObjectMeta().Annotations, model.StateBeforeSleepingAnnontation)
 
 	tr.DevApp = tr.App.DevClone()
+	annotationsString, _ := json.Marshal(tr.App.ObjectMeta().Annotations)
 
 	tr.App.ObjectMeta().Annotations[model.AppReplicasAnnotation] = strconv.Itoa(int(replicas))
+	tr.App.ObjectMeta().Annotations[model.AppAnnotationsAnnotation] = string(annotationsString)
```
pchico83, line 63: For now, let's not store them. On `okteto down` we just delete the ones in `tr.Dev.Metadata.Annotations`

pchico83 (approved):

pkg/k8s/apps/translate.go
jLopezbarb: What if you modify one environment from the original deployment? This would cause a variable to be deleted from the original deployment which may result in the application being left in an undesired status

pchico83: Do you have a use case in mind? I think is very unlikely, but if you want to handle it, I would add the previous annotation state just if needed

jLopezbarb: I was thinking about connecting to a different database from the one in staging, for example I could connect to my local db instead of using a shared db

pchico83: You would use envvars for that, no?

jLopezbarb: Yes, you're right, I was mixing up both. I won't store them for now

ifbyol (approved):

PR #2044, opened 2021-12-02
Youngestdev/compose env

### LINKED ISSUE #1890: Support `$COMPOSE_FILE` in `stack deploy`

We have a bunch of microservices with a separate compose YAML file per service, which combine to form a single docker-compose stack. Docker-compose supports an environment variable `$COMPOSE_FILE` that is a path-separated list of compose manifests to be combined to form the stack. This works just like `docker-compose -f` but greatly simplifies CLI interaction when you have a large number of files (we have 30+).

https://docs.docker.com/compose/reference/envvars/#compose_file

Add support for $COMPOSE_FILE that works like `okteto stack deploy -f`.

As mentioned, `-f` works as a workaround but is very unwieldy when there's a large number of files.

Youngestdev:
Fixes #1890

## Proposed changes

- Added support for `$COMPOSE_FILE` env variable.
- Deploy multiple stack files listed in the `COMPOSE_FILE` env variable.


cmd/stack/deploy.go
```diff
@@ -26,6 +29,13 @@ import (
 	"github.com/spf13/cobra"
 )
 
+func splitComposeFileEnv(value string) []string {
```
pchico83, line 32: could you add a unit test for this function?

cmd/stack/deploy.go
```diff
@@ -42,6 +52,12 @@ func Deploy(ctx context.Context) *cobra.Command {
 				}
 			}
 
+			composeEnv, present := os.LookupEnv(model.ComposeFile)
```
pchico83, line 55: @jLopezbarb what would be a good way to test this?

pkg/model/const.go
```diff
@@ -282,4 +282,6 @@ const (
 
 	// GithubRepositoryEnvVar defines the server to be used
 	GithubServerURLEnvVar = "GITHUB_SERVER_URL"
+
+	ComposeFile = "COMPOSE_FILE"
```
jLopezbarb, line 286:
```suggestion
	ComposeFileEnvVar = "COMPOSE_FILE"
```

cmd/stack/deploy.go
jLopezbarb: We can encapsulate this code into a function that receives stackPath as parameter and check what happen if:
1. Stackpath is empty and compose file is set
2. Stackpath is empty and compose file is set with more than one path
3. Stackpath is set and compose file is set

Youngestdev: Doesn't this block of code doesn't cover the points 1 and 2 🤔 ?

```go
if len(options.StackPath) == 0 && present {
				options.StackPath = splitComposeFileEnv(composeEnv)
			}

Youngestdev: I'll rewrite as a function and update PR.

pchico83:
@Youngestdev what's the state of this PR?

Youngestdev:
I'm going to look into writing the tests for the PR. 

Javier & I were supposed to look into it but he's on PTO atm. @pchico83

pchico83:
@jLopezbarb there are pending conflicts

pchico83 (approved):

PR #2046, opened 2021-12-03
Jlopezbarb/tty support

jLopezbarb:
Adds support for tty on deploy commands

## Proposed changes
- Adds tty support
-


cmd/deploy/deploy.go
```diff
@@ -205,13 +208,33 @@ func (dc *deployCommand) runDeploy(ctx context.Context, cwd string, opts *Option
...
+				return
+			}
+		}
+		exit <- nil
+	}()
+
+	select {
+	case <-stop:
+		log.Infof("CTRL+C received, starting shutdown sequence")
+		dc.cleanUp(ctx)
```
ifbyol, line 229: Wouldn't the defer command at line 196 would be executed before the return?

cmd/deploy/deploy.go
```diff
@@ -205,13 +208,33 @@ func (dc *deployCommand) runDeploy(ctx context.Context, cwd string, opts *Option
...
+		return errors.ErrIntSig
+	case err := <-exit:
+		dc.cleanUp(ctx)
```
ifbyol, line 232: Same about the defer

cmd/deploy/deploy.go
```diff
@@ -205,13 +208,33 @@ func (dc *deployCommand) runDeploy(ctx context.Context, cwd string, opts *Option
...
+			if err := dc.executor.Execute(command, opts.Variables); err != nil {
+				log.Infof("error executing command '%s': %s", command, err.Error())
+				exit <- fmt.Errorf("error executing command '%s': %s", command, err.Error())
+				return
+			}
+		}
+		exit <- nil
+	}()
+
+	select {
```
ifbyol, line 226: Wouldn't this be needed also for the execution of destroy commands?

ifbyol (commented):
Can you add a video on how the tty output would look like?

pchico83:
@jLopezbarb if `tty` is selected, `okteto build` commands inside a pipeline should be shown in `tty`

jLopezbarb:
> @jLopezbarb if `tty` is selected, `okteto build` commands inside a pipeline should be shown in `tty`

If we show them in tty we can't collapse when finalising the command because both commands (okteto build and okteto deploy) writes on stdout at the same time

jLopezbarb:
The code is quite outdated and the conflicts generated are more complex than redoing it, I'm going to open this PR again based on main

PR #2047, opened 2021-12-03
Read build from manifest

teresaromero:
Fixes #

## Proposed changes

- Add Build to new manifest declaration
- Parse to BuildParams struct the build options set at manifest for each image

### DIFF
--- a/pkg/model/build.go
+++ b/pkg/model/build.go
@@ -1,0 +1,46 @@
+package model
+
+type rawBuildParams struct {
+	Image     string   `json:"image,omitempty" yaml:"image,omitempty"`
+	Context   string   `json:"context,omitempty" yaml:"context,omitempty"`
+	File      string   `json:"dockerfile,omitempty" yaml:"dockerfile,omitempty"`
+	Target    string   `json:"target,omitempty" yaml:"target,omitempty"`
+	Args      []string `json:"args,omitempty" yaml:"args,omitempty"`
+	Secrets   []string `json:"secrets,omitempty" yaml:"secrets,omitempty"`
+	CacheFrom []string `json:"cacheFrom,omitempty" yaml:"cacheFrom,omitempty"`
+}
+
+type BuildParams struct {
+	Image     string   `json:"image,omitempty" yaml:"image,omitempty"`
+	Context   string   `json:"context,omitempty" yaml:"context,omitempty"`
+	File      string   `json:"dockerfile,omitempty" yaml:"dockerfile,omitempty"`
+	Target    string   `json:"target,omitempty" yaml:"target,omitempty"`
+	Args      []string `json:"args,omitempty" yaml:"args,omitempty"`
+	Secrets   []string `json:"secrets,omitempty" yaml:"secrets,omitempty"`
+	CacheFrom []string `json:"cacheFrom,omitempty" yaml:"cacheFrom,omitempty"`
+}
+
+// UnmarshalYAML Implements the Unmarshaler interface of the yaml pkg.
+func (build *BuildParams) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	var rawString string
+	err := unmarshal(&rawString)
+	if err == nil {
+		build.Context = rawString
+
+		return nil
+	} else {
+		var buildParams rawBuildParams
+		err = unmarshal(&buildParams)
+		if err != nil {
+			return err
+		}
+
+		build.Image = buildParams.Image
+		build.Context = buildParams.Context
+		build.File = buildParams.File
+		build.Target = buildParams.Target
+		build.Args = buildParams.Args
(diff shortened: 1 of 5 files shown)


pkg/model/manifest.go
```diff
@@ -13,21 +13,24 @@
...
-	Destroy []string     `json:"destroy,omitempty" yaml:"destroy,omitempty"`
-
-	Type     string `yaml:"-"`
-	Filename string `yaml:"-"`
+	Name     string        `json:"name,omitempty" yaml:"name,omitempty"`
+	Icon     string        `json:"icon,omitempty" yaml:"icon,omitempty"`
+	Deploy   *DeployInfo   `json:"deploy,omitempty" yaml:"deploy,omitempty"`
+	Dev      ManifestDevs  `json:"devs,omitempty" yaml:"devs,omitempty"`
+	Destroy  []string      `json:"destroy,omitempty" yaml:"destroy,omitempty"`
+	Build    ManifestBuild `json:"build,omitempty" yaml:"build,omitempty"`
```
pchico83, line 27: is it possible to reuse `BuildInfo`?
https://github.com/okteto/okteto/blob/fd03d2c861691e15320dd99f63c97af18674cda7/pkg/model/dev.go#L116

pkg/model/build.go
```diff
@@ -0,0 +1,46 @@
...
+	Image     string   `json:"image,omitempty" yaml:"image,omitempty"`
+	Context   string   `json:"context,omitempty" yaml:"context,omitempty"`
+	File      string   `json:"dockerfile,omitempty" yaml:"dockerfile,omitempty"`
+	Target    string   `json:"target,omitempty" yaml:"target,omitempty"`
+	Args      []string `json:"args,omitempty" yaml:"args,omitempty"`
+	Secrets   []string `json:"secrets,omitempty" yaml:"secrets,omitempty"`
+	CacheFrom []string `json:"cacheFrom,omitempty" yaml:"cacheFrom,omitempty"`
+}
+
+type BuildParams struct {
```
pchico83, line 13: is it possible to reuse `BuildInfo`?
https://github.com/okteto/okteto/blob/fd03d2c861691e15320dd99f63c97af18674cda7/pkg/model/dev.go#L116

pkg/model/manifest_test.go
```diff
@@ -0,0 +1,62 @@
...
+		name              string
+		manifestYAML      []byte
+		manifestNotExists bool
+		expectedErr       bool
+		expectedManifest  *ManifestBuild
+	}{
+		{
+			name: "exists-manifest",
+			manifestYAML: []byte(`build:
+  firstImage: ./first
```
pchico83, line 21: I would use a sample with a value for every field supported in the build section

pkg/model/serializer_test.go
```diff
@@ -1778,3 +1778,41 @@ compose:
...
+	tests := []struct {
+		name            string
+		buildManifest   []byte
+		expected        ManifestBuild
+		isErrorExpected bool
+	}{
+		{
+			name: "build params at manifest",
+			buildManifest: []byte(`firstBuild: ./first
+secondBuild:
```
pchico83, line 1792: I would use a sample with a value for every field supported in the build section

pkg/model/serializer.go
```diff
@@ -716,7 +716,7 @@ func (d *ManifestDevs) UnmarshalYAML(unmarshal func(interface{}) error) error {
 }
 
 func isManifestFieldNotFound(err error) bool {
-	manifestFields := []string{"devs", "name", "icon", "variables", "deploy", "destroy"}
+	manifestFields := []string{"devs", "name", "icon", "variables", "deploy", "destroy", "build"}
```
pchico83, line 719: @jLopezbarb based on the spec, it should be `dev`, not `devs`:
https://www.notion.so/okteto/Okteto-DEV-915b72bf1a8f4275879974e544a2d614#5ee9c33a37484727a72f5607912f6bfb

teresaromero:
re-opened with new branch name

pkg/model/build.go
teresaromero: the keys for this struct are different, for example, image is called Name, should i add Image to BuildInfo ?

pkg/model/manifest.go
teresaromero: Build will still need to be a map for the different images defined, you mean changing the BuildParams struct for each image to BuildInfo?

pkg/model/build.go
pchico83: Yeap, let's add `Image`. `Name` and where it's used will be deprecated after the build section is released

pkg/model/manifest.go
pchico83: Right

PR #2048, opened 2021-12-03
Fix annotations restore original

jLopezbarb:

Fixes: With the addition of https://github.com/okteto/okteto/pull/2037 it was not cleaning some of the okteto annotations, which caused modifying the original deployment

## Proposed changes
- Remove those annotations when performing okteto down
-

### DIFF
--- a/pkg/k8s/apps/translate.go
+++ b/pkg/k8s/apps/translate.go
@@ -111,6 +111,11 @@ func (tr *Translation) DevModeOff() error {
 	tr.App.SetReplicas(getPreviousAppReplicas(tr.App))
 	delete(tr.App.ObjectMeta().Annotations, model.AppReplicasAnnotation)
 
+	delete(tr.App.ObjectMeta().Annotations, model.OktetoStignoreAnnotation)
+	delete(tr.App.TemplateObjectMeta().Annotations, model.OktetoStignoreAnnotation)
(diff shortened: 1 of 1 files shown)


jLopezbarb:
We are adding these two annotations to Dev.Metadata.Annotations while running `okteto up`:
- `dev.okteto.com/stignore` is added here https://github.com/okteto/okteto/blob/master/cmd/up/stignore.go#L100
- "dev.okteto.com/sync" is added here https://github.com/okteto/okteto/blob/master/cmd/up/stignore.go#L100

When running okteto down  we have to remove all the annotations that the user has introduced + these two that are generated by up command

pkg/k8s/apps/translate.go
```diff
@@ -111,6 +111,11 @@ func (tr *Translation) DevModeOff() error {
 	tr.App.SetReplicas(getPreviousAppReplicas(tr.App))
 	delete(tr.App.ObjectMeta().Annotations, model.AppReplicasAnnotation)
 
+	delete(tr.App.ObjectMeta().Annotations, model.OktetoStignoreAnnotation)
```
ifbyol, line 114: I would include a comment in the code to explain why these are being removed explicitly (more or less the same comment you have put in this review)

ifbyol (approved):

pchico83 (approved):

PR #2049, opened 2021-12-03
Jlopezbarb/add deploy destroy e2e

### LINKED ISSUE #2045: E2E tests for `okteto deploy` and `okteto destroy`
We need an automatic e2e test for the deploy/destroy functionality using the new okteto manifest

jLopezbarb:
Fixes #2045

## Proposed changes
- Fixes the nil pointer exception for devs field
- Adds [built-in-environment-variables](https://okteto.com/docs/cloud/okteto-pipeline/#built-in-environment-variables) if they are not present

### DIFF
--- a/cmd/context/utils.go
+++ b/cmd/context/utils.go
@@ -196,9 +196,6 @@ func GetManifest(ctx context.Context, srcFolder string, opts ManifestOptions) (*
 			log.Infof("could not load manifest: %s", err.Error())
 		}
 
-		if manifest.Name == "" {
-			manifest.Name = opts.Name
-		}
 		manifest.Type = "pipeline"
 		manifest.Filename = pipelinePath
 		return manifest, nil
--- a/cmd/context/utils.go
+++ b/cmd/context/utils.go
@@ -196,6 +196,9 @@ func GetManifest(ctx context.Context, srcFolder string, opts ManifestOptions) (*
 			log.Infof("could not load manifest: %s", err.Error())
 		}
 
+		if manifest.Name == "" {
+			manifest.Name = opts.Name
+		}
 		manifest.Type = "pipeline"
 		manifest.Filename = pipelinePath
 		return manifest, nil
--- a/cmd/deploy/deploy.go
+++ b/cmd/deploy/deploy.go
@@ -386,7 +386,8 @@ func addEnvVars(ctx context.Context, cwd string) error {
 		if isClean {
 			os.Setenv(model.OktetoGitCommitEnvVar, sha)
 		} else {
-			os.Setenv(model.OktetoGitCommitEnvVar, fmt.Sprintf("%s-local", sha))
+			sha := utils.GetRandomSHA(ctx, cwd)
+			os.Setenv(model.OktetoGitCommitEnvVar, sha)
 		}
 	}
 	if os.Getenv(model.OktetoRegistryURLEnvVar) == "" {
--- a/cmd/deploy/deploy.go
+++ b/cmd/deploy/deploy.go
@@ -379,8 +379,15 @@ func addEnvVars(ctx context.Context, cwd string) error {
 		if err != nil {
 			log.Infof("could not retrieve sha: %s", err)
 		}
-
-		os.Setenv(model.OktetoGitCommitEnvVar, sha)
+		isClean, err := utils.IsCleanDirectory(ctx, cwd)
+		if err != nil {
+			log.Infof("could not status: %s", err)
+		}
+		if isClean {
+			os.Setenv(model.OktetoGitCommitEnvVar, sha)
+		} else {
+			os.Setenv(model.OktetoGitCommitEnvVar, fmt.Sprintf("%s-local", sha))
+		}
 	}
 	if os.Getenv(model.OktetoRegistryURLEnvVar) == "" {
 		os.Setenv(model.OktetoRegistryURLEnvVar, okteto.Context().Registry)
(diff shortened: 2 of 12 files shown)


cmd/context/utils.go
```diff
@@ -196,6 +196,9 @@ func GetManifest(ctx context.Context, srcFolder string, opts ManifestOptions) (*
 			log.Infof("could not load manifest: %s", err.Error())
 		}
 
+		if manifest.Name == "" {
```
pchico83, line 199: I think we shouldn't have `name` in the manifest yet

cmd/deploy/deploy.go
```diff
@@ -362,3 +364,29 @@ func getProxyHandler(name, token string, clusterConfig *rest.Config) (http.Handl
 	return handler, nil
 
 }
+
+func addEnvVars(ctx context.Context, cwd string) error {
```
pchico83, line 368: what's this for?

jLopezbarb: I'm adding this because we were doing it on the backend and it's documented [here](https://okteto.com/docs/cloud/okteto-pipeline/#built-in-environment-variables)

cmd/context/utils.go
jLopezbarb: We can have it if the manifest is the old one where the name was mandatory

pchico83: But that should be part of `dev`, no?
I think we should remove this one:
https://github.com/okteto/okteto/blob/master/pkg/model/manifest.go#L20

cmd/deploy/deploy.go
ifbyol: hmm, I think that populating `OKTETO_GIT_COMMIT` env var for `okteto deploy` could cause some issues with the build optimization.

Let say, I have the local repo movies cloned with some local changes I want to test. If I run it and it tries to build the image `okteto.dev/api:${OKTETO_GIT_COMMIT}` but there is already an image built with that commit which doesn't contain my local changes, I wouldn't be deploying exactly what I want to deploy because it would reuse the one already built.

pchico83: @jLopezbarb what value is returned by `utils.GetGitCommit` if you have uncommitted changes?

jLopezbarb: The value is the same, is the last commit until you create a new one. One solution could be:
If there are no new changes get the commit.
If there are local changes add to the commit sha the suffix local.
This option will fix the first time of deploying it but if you want to redeploy it will deploy the changes that you did in the first deploy commit. @ifbyol suggested me to inject a suffix or special tag that skips the build optimisation.
Another solution could be show a warning to the user telling that the `OKTETO_GIT_COMMIT` env var has not been injected due to having local changes 
What do you think could be the best solution?

pchico83: @jLopezbarb I would add a random suffix, or we can also ignore the build optimization if `OKTETO_GIT_COMMIT` ends with `local`. Both of them work.

cmd/utils/git.go
```diff
@@ -40,3 +40,38 @@ func GetBranch(ctx context.Context, path string) (string, error) {
...
+	if err != nil {
+		return false, fmt.Errorf("failed to infer the git repo's current branch: %w", err)
+	}
+
+	status, err := worktree.Status()
+	if err != nil {
+		return false, fmt.Errorf("failed to infer the git repo's status: %w", err)
+	}
+
+	return status.IsClean(), nil
```
pchico83, line 76: is it possible to get a hash of the changes here?

jLopezbarb: no without committing the changes, because timestamp is part of that hash. I have generated a random string

pchico83 (approved):

ifbyol (approved):

jLopezbarb:
I have changed the repo to deploy because the go-ci image doesn't have helm installed and could not deploy that repo

pchico83:
> I have changed the repo to deploy because the go-ci image doesn't have helm installed and could not deploy that repo

`helm` should be available, at least in the latest tag:
https://github.com/okteto/golang-ci/blob/main/Dockerfile#L20

jLopezbarb:
> > I have changed the repo to deploy because the go-ci image doesn't have helm installed and could not deploy that repo
> 
> `helm` should be available, at least in the latest tag: https://github.com/okteto/golang-ci/blob/main/Dockerfile#L20

Sorry, those tests are running with okteto/dev:go1.17 instead of okteto/golang-ci:1.17.3

pchico83:
> > > I have changed the repo to deploy because the go-ci image doesn't have helm installed and could not deploy that repo
> > 
> > 
> > `helm` should be available, at least in the latest tag: https://github.com/okteto/golang-ci/blob/main/Dockerfile#L20
> 
> Sorry, those tests are running with okteto/dev:go1.17 instead of okteto/golang-ci:1.17.3.

Not sure how that image is built... would it work if we move to `okteto/golang-ci`? I would like to test the movies sample, it's a more complicated pipeline

jLopezbarb:
> > > > I have changed the repo to deploy because the go-ci image doesn't have helm installed and could not deploy that repo
> > > 
> > > 
> > > `helm` should be available, at least in the latest tag: https://github.com/okteto/golang-ci/blob/main/Dockerfile#L20
> > 
> > 
> > Sorry, those tests are running with okteto/dev:go1.17 instead of okteto/golang-ci:1.17.3.
> 
> Not sure how that image is built... would it work if we move to `okteto/golang-ci`? I would like to test the movies sample, it's a more complicated pipeline

We don't have other external dependencies on Golang-ci such as curl. should we add helm to dev which I think is this Dockerfile https://github.com/okteto/devenv/blob/main/dev/Dockerfile or should we add curl to golang-ci?

pchico83:
> golang-ci?

I would standarize on golang-ci

PR #2052, opened 2021-12-09
Read build from manifest

### LINKED ISSUE #2038: Okteto build from manifest
I want to define the images that I want to build in my okteto manifest.

The new **okteto manifest** defines a section for building the images of your application:

```
# okteto.yml

build:
  frontend:
    context: frontend
    dockerfile: Dockerfile
  api:
    context: api
    dockerfile: Dockerfile
```

Fields defining a build at the manifest:

- `image`: optional - tag for the image when pushing it to registry. This is optional when you are using okteto platform registry.
- `context`: required - where the Dockerfile for building the image is located
- `dockerfile`: optional - the name for the Dockerfile to use for building the image
- `target: optional - the stage at the dockerfile to use
- `args`: optional - the arguments passed when making the build

The command `okteto build` will now build all the images described in the manifest.

```bash
okteto build
```

```bash
okteto build frontend
```

Also, the current build command with tag flag can be run to achieve the previous behavior.

```bash
okteto build -t your.registry.com/frontend:dev
```

teresaromero:
Resolves Partially #2038 

## Proposed changes

- detect build definitions to new okteto manifest definition
- parse builds definitions to Manifest struct, adding new fields into known BuildInfo struct
- if context is NOT defined, the default context will be relative path using the declaration name
- if dockerfile is not defined, Dockerfile is default


pkg/model/dev.go
```diff
@@ -120,6 +120,8 @@ type BuildInfo struct {
 	CacheFrom  []string    `yaml:"cache_from,omitempty"`
 	Target     string      `yaml:"target,omitempty"`
 	Args       Environment `yaml:"args,omitempty"`
+	Image      string      `yaml:"image,omitempty"`
+	Secrets    []string    `yaml:"secrets,omitempty"`
```
pchico83, line 124: why are we adding secrets?

pkg/model/dev.go
```diff
@@ -608,6 +612,20 @@ func setBuildDefaults(build *BuildInfo) {
 	}
 }
 
+func setBuildsManifestDefaults(builds ManifestBuild) {
+	for name, b := range builds {
+		if b == nil {
+			b = &BuildInfo{}
+		}
+		if b.Context == "" {
+			b.Context = fmt.Sprintf("./%s", name)
```
pchico83, line 621: Default should be `.`

pkg/model/dev.go
```diff
@@ -608,6 +612,20 @@ func setBuildDefaults(build *BuildInfo) {
...
+		}
+		if _, err := url.ParseRequestURI(b.Context); err != nil && b.Dockerfile == "" {
+			b.Dockerfile = "Dockerfile"
```
pchico83, line 624: Why are we parsing `b.Context`? I think the default should be `b.Context` + `Dockerfile`
@jLopezbarb what's the default in stacks?

pkg/model/serializer.go
```diff
@@ -286,7 +288,12 @@ func (buildInfo *BuildInfo) UnmarshalYAML(unmarshal func(interface{}) error) err
 	var rawString string
 	err := unmarshal(&rawString)
 	if err == nil {
+		if strings.HasPrefix(rawString, "./") || strings.HasPrefix(rawString, ".") {
+			buildInfo.Context = rawString
+			return nil
+		}
 		buildInfo.Name = rawString
```
pchico83, line 295: @jLopezbarb why is this? in this case, only `context` is defined, no?

pkg/model/dev.go
teresaromero: because is also an option at the cli command, we are not using this then?

pkg/model/serializer.go
teresaromero: i added this because rawString is assigned as Name when using BuildInfo, but in the build manifest, if the build is a string can be the rel path to the dockerfile...

pkg/model/dev.go
```diff
@@ -374,6 +375,14 @@ func Read(bytes []byte) (*Manifest, error) {
 		})
 	}
 
+	for _, b := range manifest.Build {
+		if b.Name != "" {
```
pchico83, line 379: I think we should return an error here. `Name` is not supported here

pchico83 (approved):

jLopezbarb (approved):

pkg/model/dev.go
teresaromero: This is for parsing the relPath as Context and not as name when using BuildInfo Object

PR #2063, opened 2021-12-12
Support for markdown syntax to show preview endpoints

pchico83:

Markdown is supported for most git providers. `markdown` output is very convenient to add a comment on a PR for example when integrating the Okteto CLI with different CI providers.

@jLopezbarb I think we can also delete some code from our github action using `markdown` format output

### DIFF
--- a/cmd/preview/endpoints.go
+++ b/cmd/preview/endpoints.go
@@ -63,16 +63,18 @@ func Endpoints(ctx context.Context) *cobra.Command {
 			return err
 		},
 	}
-	cmd.Flags().StringVarP(&output, "output", "o", "", "output format. One of: ['json']")
+	cmd.Flags().StringVarP(&output, "output", "o", "", "output format. One of: ['json', 'markdown']")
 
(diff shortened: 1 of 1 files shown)


cmd/preview/endpoints.go
```diff
@@ -63,16 +63,18 @@ func Endpoints(ctx context.Context) *cobra.Command {
...
+	cmd.Flags().StringVarP(&output, "output", "o", "", "output format. One of: ['json', 'markdown']")
 
 	return cmd
 }
 
 func validateOutput(output string) error {
-	if output != "" && output != "json" {
-		return fmt.Errorf("output format is not accepted. Value must be one of: ['json']")
+	switch output {
+	case "", "json", "md", "markdown":
```
rberrelleza, line 73: Can we only use "md" or "markdown " but not both? Just to keep things simple.

rberrelleza (approved):

jLopezbarb (approved):

PR #2072, opened 2021-12-15
Refactor build command

teresaromero:
## Proposed changes

- Keeping okteto build compatible with the build manifest
- new functionality will be only triggered if OKTETO_READ_MANIFEST is true as ENV
- when reading the build manifest, if there are definitions, okteto build will build this from the manifest, if not, fallback is current logic
- If there is no manifest,  command will build according to the Dockerfile on the path specified.
- When using flags with the command and building from manifest, the flags will override the values of the manifest, in case the build is only of one image


pchico83:
> ## Proposed changes
> or command uses the -t flag, command will build according to the Dockerfile on the path specified.

I would just do it based on the presence of the new okteto manifest

cmd/build.go
```diff
@@ -42,44 +44,69 @@ func Build(ctx context.Context) *cobra.Command {
...
-				options.File = filepath.Join(path, "Dockerfile")
-			}
+			manifestPath := contextCMD.GetOktetoManifestPath(path)
+			if manifestPath != "" {
+				buildManifest, err := model.GetBuildManifest(manifestPath)
+				if err != nil {
+					return err
+				}
+				if len(buildManifest) != 0 {
+					image := ""
```
pchico83, line 65: I would call this service. `image` is more the image name we will build and push to the registry

cmd/build.go
```diff
@@ -93,3 +120,72 @@ func Build(ctx context.Context) *cobra.Command {
...
+		b.Image = setOktetoImageTag(name)
+	}
+
+	opts := build.BuildOptions{
+		BuildArgs:  model.SerializeBuildArgs(b.Args),
+		CacheFrom:  b.CacheFrom,
+		Target:     b.Target,
+		Path:       b.Context,
+		Tag:        b.Image,
+		OutputMode: "tty",
```
pchico83, line 173: @jLopezbarb what's the envvar to set this in the pipelines? we shouldn't hardcode `tty`

cmd/build.go
```diff
@@ -93,3 +120,72 @@ func Build(ctx context.Context) *cobra.Command {
...
+	} else {
+		log.Success(fmt.Sprintf("Image '%s' successfully pushed", options.Tag))
+	}
+
+	analytics.TrackBuild(okteto.Context().Builder, true)
+	return nil
+}
+
+func buildFromManifest(name string, b *model.BuildInfo) error {
+	if b.Image == "" {
```
pchico83, line 163: if `context` is not an okteto cluster, we should fail here

cmd/context/utils.go
```diff
@@ -369,3 +369,21 @@ func fileExistsAndNotDir(filename string) bool {
 	}
 	return !info.IsDir()
 }
+
+func GetOktetoManifestPath(path string) string {
```
pchico83, line 373: we should also support to specify the path with `-f`

pchico83: and the `--namespace` flag

teresaromero: The namespace flag, value by default according to the spec is the value specified at the manifest. we currently havent got implemented this field on the manifest, should i open a pr to add this field to manifest ? if not, i can implement the flag and default namespace be the one in the context?

pchico83: yes to both. the order is flag > manifest > context. in the manifest is also optional

cmd/build.go
```diff
@@ -41,45 +42,70 @@ func Build(ctx context.Context) *cobra.Command {
...
+			if options.File != "" {
+				if err := utils.CheckIfRegularFile(options.File); err != nil {
+					return fmt.Errorf("invalid File: %s", err.Error())
+				}
+				manifestPath = options.File
+			} else {
+				manifestPath = contextCMD.GetOktetoManifestPath(manifestPath)
 			}
-			options.Path = path
+			if manifestPath != "" {
```
jmacelroy, line 59: These changes seem long and a bit nested? Is it worth pulling it out into a function and writing a test to make sure we hit all the correct conditionals created when we expect to?

cmd/build.go
```diff
@@ -32,7 +33,7 @@ func Build(ctx context.Context) *cobra.Command {
 
 	options := build.BuildOptions{}
 	cmd := &cobra.Command{
-		Use:   "build [PATH]",
+		Use:   "build [service]",
```
pchico83, line 36: while we don't document the new build command and the new okteto manifest, I would keep documenting the current behavior. Same for `-f`

cmd/build.go
```diff
@@ -91,5 +92,51 @@ func Build(ctx context.Context) *cobra.Command {
...
+	} else {
+		log.Success(fmt.Sprintf("Image '%s' successfully pushed", opts.Tag))
+	}
+
+	analytics.TrackBuild(okteto.Context().Builder, true)
+	return nil
+}
+
+func readFromManifest() bool {
+	r, err := strconv.ParseBool(os.Getenv("OKTETO_READ_MANIFEST"))
```
pchico83, line 137: Let's call it `OKTETO_ENABLE_MANIFEST_V2`

cmd/context/utils.go
```diff
@@ -369,3 +369,21 @@ func fileExistsAndNotDir(filename string) bool {
 	}
 	return !info.IsDir()
 }
+
+func GetOktetoManifestPath(file string) string {
```
teresaromero, line 373: @pchico83 not sure this func should be at this package, thought moving it to utils?

cmd/build.go
```diff
@@ -41,45 +44,21 @@ func Build(ctx context.Context) *cobra.Command {
 				return err
 			}
 
-			path := "."
-			if len(args) == 1 {
-				path = args[0]
+			m, ok := isManifestV2(options.File)
+			if m != nil && ok {
+				err := buildV2(m, options, args)
```
pchico83, line 49: you can return `buildV2(m, options, args)` directly

cmd/build.go
```diff
@@ -41,45 +44,21 @@ func Build(ctx context.Context) *cobra.Command {
...
-			if okteto.Context().Builder == "" {
-				log.Information("Building your image using your local docker daemon")
-			} else {
-				log.Information("Running your build in %s...", okteto.Context().Builder)
-			}
-
-			ctx := context.Background()
-			if err := build.Run(ctx, options); err != nil {
-				analytics.TrackBuild(okteto.Context().Builder, false)
+			err := buildV1(options, args)
```
pchico83, line 56: `return buildV1(options, args)` directly

cmd/build.go
```diff
@@ -91,5 +70,105 @@ func Build(ctx context.Context) *cobra.Command {
...
+
+		err := buildV1(opts, []string{service})
+		if err != nil {
+			return err
+		}
+		return nil
+	}
+
+	for service, b := range m {
+		opts := build.OptsFromManifest(service, b)
```
pchico83, line 124: if we are not using build overwrites in this case, we should fail if they are specified

cmd/context/utils.go
pchico83: @ifbyol do we already have a function for this?

pkg/cmd/build/build.go
```diff
@@ -194,3 +196,32 @@ func translateDockerErr(err error) error {
 	}
 	return err
 }
+
+func OptsFromManifest(name string, b *model.BuildInfo) BuildOptions {
+	if okteto.Context().IsOkteto && b.Image == "" {
```
pchico83, line 201: we should fail if `b.Image == ""` and it's not `!okteto.Context().IsOkteto`

pkg/cmd/build/build.go
```diff
@@ -194,3 +196,32 @@ func translateDockerErr(err error) error {
...
+		Target:    b.Target,
+		Path:      b.Context,
+		Tag:       b.Image,
+		File:      filepath.Join(b.Context, b.Dockerfile),
+	}
+	if len(b.Args) != 0 {
+		opts.BuildArgs = model.SerializeBuildArgs(b.Args)
+	}
+
+	opts.OutputMode = setOutputMode("tty")
```
pchico83, line 216: we should get this fron the build flag `--progress`. There is also a `BUILDKIT_PROGRESS` envvar to configure it

pkg/cmd/build/build.go
```diff
@@ -194,3 +196,32 @@ func translateDockerErr(err error) error {
...
+	return opts
+}
+
+func setOktetoImageTag(name string) string {
+	imageTag := "dev"
+	okGitCommit := os.Getenv("OKTETO_GIT_COMMIT")
```
pchico83, line 222: Let's skip all the logic around `OKTETO_GIT_COMMIT` for now

pkg/cmd/build/build.go
```diff
@@ -194,3 +196,32 @@ func translateDockerErr(err error) error {
 	}
 	return err
 }
+
+func OptsFromManifest(name string, b *model.BuildInfo) BuildOptions {
```
pchico83, line 200: could you add unit tests for this function?

cmd/build.go
```diff
@@ -91,5 +70,105 @@ func Build(ctx context.Context) *cobra.Command {
 	cmd.Flags().StringVarP(&options.OutputMode, "progress", "", "tty", "show plain/tty build output")
 	cmd.Flags().StringArrayVar(&options.BuildArgs, "build-arg", nil, "set build-time variables")
 	cmd.Flags().StringArrayVar(&options.Secrets, "secret", nil, "secret files exposed to the build. Format: id=mysecret,src=/local/secret")
+	cmd.Flags().StringVarP(&options.Namespace, "namespace", "", "", "namespace against which the image will be consumed. Default is the one defined at okteto context or okteto manifest")
 	return cmd
 }
+
+func isManifestV2(file string) (model.ManifestBuild, bool) {
```
pchico83, line 77: could you add unit tests for this function?

cmd/build.go
```diff
@@ -91,5 +70,105 @@ func Build(ctx context.Context) *cobra.Command {
 	cmd.Flags().StringVarP(&options.OutputMode, "progress", "", "tty", "show plain/tty build output")
 	cmd.Flags().StringArrayVar(&options.BuildArgs, "build-arg", nil, "set build-time variables")
 	cmd.Flags().StringArrayVar(&options.Secrets, "secret", nil, "secret files exposed to the build. Format: id=mysecret,src=/local/secret")
+	cmd.Flags().StringVarP(&options.Namespace, "namespace", "", "", "namespace against which the image will be consumed. Default is the one defined at okteto context or okteto manifest")
```
pchico83, line 73: we should consider this value when initializing the okteto context:
https://github.com/okteto/okteto/pull/2072/files#diff-5794c40aa4c9301aaf9c39eacf7d3c8511f2511cb24bdd95e0a08704ba583a5fR43

pkg/cmd/build/build.go
teresaromero: ok, this function is already using BUILDKIT_PROGRESS, and using tty as default if this is not defined, i will add the flag

cmd/context/utils.go
ifbyol: We have a similar function here: https://github.com/okteto/okteto/blob/7bc650c344105b0a76002a5eb81dc0ad8ae0b6e7/cmd/context/utils.go#L339.

It is not exactly the same but I think we should combine them to avoid to have several functions to do the same. Ideally, we should have just 1 function to load an `okteto.yaml` manifest

cmd/build.go
```diff
@@ -91,5 +60,105 @@ func Build(ctx context.Context) *cobra.Command {
...
+		return nil, false
+	}
+	mPath := contextCMD.GetOktetoManifestPath(file)
+	if mPath == "" {
+		return nil, false
+	}
+
+	mBuild, err := model.GetBuildManifest(mPath)
+	if err != nil || mBuild == nil {
+		return nil, false
```
ifbyol, line 79: Should we at least log the error if exists to our logs, so an `okteto doctor` command would include them? It could help us to troubleshoot issues

cmd/build.go
```diff
@@ -91,5 +60,105 @@ func Build(ctx context.Context) *cobra.Command {
...
+		}
+
+		opts := build.OptsFromManifest(service, b)
+		opts.Secrets = options.Secrets
+
+		err := buildV1(opts, []string{service})
+		if err != nil {
+			return err
+		}
+		return nil
```
ifbyol, line 110: you could directly return `buildV1(opts, []string{service})`

cmd/build.go
```diff
@@ -91,5 +60,105 @@ func Build(ctx context.Context) *cobra.Command {
...
+	}
+
+	for service, b := range m {
+		opts := build.OptsFromManifest(service, b)
+		err := buildV1(opts, []string{service})
```
ifbyol, line 115: Aren't we assuming that the service name would be the same than the folder name? Checking the `buildV1`, service would be used as a path. Shouldn't we use the `context` defined in the manifest for that specific service or `.` by default? the same for line 106 in this file.

I mean, If I have this manifest:

```
...
build:
  api:
    context: backend
    dockerfile: Dockerfile
    ....
...
```

and I execute `okteto build api`, aren't we passing `api` as `context` instead of `backend` to the BuildV1 function?

pkg/cmd/build/build.go
```diff
@@ -194,3 +196,28 @@ func translateDockerErr(err error) error {
...
+	}
+
+	opts := BuildOptions{
+		CacheFrom: b.CacheFrom,
+		Target:    b.Target,
+		Path:      b.Context,
+		Tag:       b.Image,
+	}
+
+	if opts.Path == "." {
```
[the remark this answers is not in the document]
teresaromero: i reverted this, it was a miss understanding from  me of the path/context. the context is not determined by the service name

cmd/build.go
teresaromero: i dont find a way to add logs without displaying them at the cli, at this point we are checking the manifest, fallback is building with v1... if i add log.Error this gets displayed into the cli and it gets built with v1 or, in the cases i've tested, the error of not finding Dockerfile for v1 is displayed before exit.... 
also, okteto doctor i dont see any logs being added into the zip file :/

pchico83 (approved):

ifbyol (approved):

PR #2076, opened 2021-12-15
Add namespace v2

### LINKED ISSUE #2043: `okteto namespace` version II
I need a better way to switch my okteto context namespace

- `okteto namespace` won't modify the kubernetes context anymore. In order to do that, we have to implement `okteto kubeconfig`, and alias of `okteto kubeconfig update`
- `okteto ns` is an alias of `okteto namespace`
- `okteto namespace`(alias for `okteto namespace use`) :  with no arguments, works similar to `okteto context` and shows a list of available namespaces or gives the option to create one. With an argument, it uses that namespace:
Output:
```
✓ Using pchico83 @ [cloud.okteto.com](http://cloud.okteto.com/) as context
 i  Run 'okteto kubeconfig update' to update your kubectl credentials
```
- `okteto namespace create`: creates a namespace. `okteto create namespace` is deprecated.
- `okteto namespace delete`: deletes a namespace. `okteto delete namespace` is deprecated.
- `okteto namespace list`: list available namespaces. `*` for current namespace. `okteto list namespace` is deprecated.

**Additional context**
This is a breaking change and it should be released on 1.15-beta

jLopezbarb:
Fixes #2043

## Proposed changes
- Refactor okteto client so it's easier to test commands
- Hide `okteto create/list/delete`
- Add `ns` and `kubeconfig` aliases
- Add selector on `okteto ns` without argument 
- Add tests for every `okteto namespace` command


cmd/build.go
```diff
@@ -41,23 +41,28 @@ func Build(ctx context.Context) *cobra.Command {
...
+				return err
+			}
 			if okteto.IsOkteto() && options.Namespace != "" {
 				create, err := utils.ShouldCreateNamespace(ctx, options.Namespace)
 				if err != nil {
 					return err
 				}
 				if create {
-					err = namespace.ExecuteCreateNamespace(ctx, options.Namespace, nil)
+					nsCmd, err := namespace.NewNamespaceCommand()
```
teresaromero, line 53: I feel this could be called namespace.NewCommand ?? i feel a bit redundant calling namespace twice :/

cmd/build.go
```diff
@@ -41,23 +41,28 @@ func Build(ctx context.Context) *cobra.Command {
...
 					if err != nil {
 						return err
 					}
+					nsCmd.CreateNamespace(ctx, &namespace.CreateOptions{Namespace: options.Namespace})
```
teresaromero, line 57: same here... as is inside the namespace command maybe nsCmd.Create ? 
regarding Options, is there any other action options? EditOptions, DeleteOptions, could it be interesting to have an Options to use in any action?

cmd/context/create.go
```diff
@@ -237,7 +245,45 @@ func (_ *ContextUse) initKubernetesContext(ctxOptions *ContextOptions) error {
...
 
-func (*ContextUse) initEnvVars() {
+func (c ContextCommand) getUserContext(ctx context.Context) (*types.UserContext, error) {
+	client, err := c.OktetoClientProvider.Provide()
+	if err != nil {
+		return nil, err
+	}
+
+	retries := 0
+	for retries <= 3 {
```
pchico83, line 255: why we need to loop here?

cmd/context/create.go
```diff
@@ -237,7 +245,45 @@ func (_ *ContextUse) initKubernetesContext(ctxOptions *ContextOptions) error {
...
+		}
+		if err != nil {
+			log.Info(err)
+			retries++
+			continue
+		}
+
+		// If userID is not on context config file we add it and save it.
+		// this prevents from relogin to actual users
+		if okteto.Context().UserID == "" && okteto.Context().IsOkteto {
```
pchico83, line 274: why wouldn't it be stored already?

cmd/namespace/namespace.go
```diff
@@ -21,16 +21,36 @@ import (
...
 // Namespace fetch credentials for a cluster namespace
 func Namespace(ctx context.Context) *cobra.Command {
 	cmd := &cobra.Command{
-		Use:    "namespace [name]",
-		Hidden: true,
-		Short:  "Download k8s credentials for a namespace",
-		Args:   utils.MaximumNArgsAccepted(1, "https://okteto.com/docs/reference/cli/#namespace"),
+		Use:     "namespace [name]",
+		Hidden:  true,
+		Short:   "Download k8s credentials for a namespace",
```
pchico83, line 51: It's just configure the current namespace of the okteto context, it doesn't download k8s credentials

cmd/namespace/use.go
```diff
@@ -0,0 +1,144 @@
...
+
+const (
+	newNamespaceOption = "Create new namespace"
+)
+
+// Namespace fetch credentials for a cluster namespace
+func Use(ctx context.Context) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "use [namespace]",
+		Hidden:  true,
```
pchico83, line 37: not hidden

cmd/namespace/use.go
```diff
@@ -0,0 +1,144 @@
...
+)
+
+const (
+	newNamespaceOption = "Create new namespace"
+)
+
+// Namespace fetch credentials for a cluster namespace
+func Use(ctx context.Context) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "use [namespace]",
```
pchico83, line 36: Can we have a flag or an envar to default to the personal namespace? I think we will need it to adapt the github namespace action

cmd/context/create.go
jLopezbarb: I just moved the code from use to create because is where this code belongs
The retry was because we saw an user that got an error here and rerunning the command worked for him. If the backend fails we retry 3 times to get the userContext

cmd/context/create.go
jLopezbarb: I just moved the code from use to create because is where this code belongs
There was a previous version where it was not saved and then we stored it for vscode plugin.

cmd/namespace/use.go
jLopezbarb: True, I will create an issue on Namespace so we don't break it after merging this PR

pchico83 (approved):

PR #2077, opened 2021-12-15
Local env var is preferred over secrets

### LINKED ISSUE #2074: Local environment should have preference over okteto secrets
**Describe the bug**
Every command of the okteto cli is adding the user okteto secrets to the environment.
Okteto secrets shouldn't overwrite variables defined in the local environment.
Also, we should red `.env` as part of the local environemnt.

jLopezbarb:
Fixes #2074

## Proposed changes
- Load `.env` at the begining of okteto
- Replace env vars by secrets if the env var is not set

### DIFF
--- a/main.go
+++ b/main.go
@@ -22,6 +22,7 @@ import (
 	"time"
 	"unicode"
 
+	"github.com/compose-spec/godotenv"
 	"github.com/okteto/okteto/cmd"
 	contextCMD "github.com/okteto/okteto/cmd/context"
 	"github.com/okteto/okteto/cmd/deploy"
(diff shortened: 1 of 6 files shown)


rberrelleza:
as part of this work, could we include creating a document (or updating an existing one) that clearly specifies the overwrite rules?  We've gotten than question asked by our users multiple times.

pchico83:
> as part of this work, could we include creating a document (or updating an existing one) that clearly specifies the overwrite rules? We've gotten this question asked by our users multiple times. since we now have:
> 
> * env vars
> * .env file (considering that in the pipeline, we also create this on the fly)
> * user-level secrets for okteto cloud users
> * cluster-level secrets for okteto cloud users
> * pipeline-level variables for okteto cloud users

In addition to that, I think that doc/section should explain that those vars can be used to expand their values anywhere in your manifests

main.go
```diff
@@ -69,6 +70,11 @@ func init() {
 		model.OktetoBinImageTag = bin
 		log.Infof("using %s as the bin image", bin)
 	}
+	if model.FileExists(".env") {
```
pchico83, line 73: I think I would do this when initializing the okteto context

pchico83 (approved):

main.go
ifbyol: yep, I think it would makes sense to do it as part of the context initialization

ifbyol (approved):

PR #2078, opened 2021-12-15
Check if ns exists and create it if not

### LINKED ISSUE #2019: `okteto` should create the namespace if needed
I have a namespace for each of my Okteto Stacks. I deploy each stack on a different namespace.
In order to share the stacks with my team, I am adding this to my manifests:
```
namespace: xxx-${OKTETO_USERNAME}
```
but people are getting this error because `xxx-${OKTETO_USERNAME}` doesn't exist yet.
```
Ingresses.networking.k8s.io is forbidden: User "system:serviceaccount:okteto:xxxxx" cannot list resource "ingresses" in API group "networking.k8s.io" in the namespace "xxx-pchico83"
```

When `namespace` is defined in a okteto manifest, and the namespace doesn't exit, okteto should prompt to the user if he wants to create the namespace.

jLopezbarb:

Fixes #2019

## Proposed changes
- Check if ns exists, and if not create it asking the user
-

### DIFF
--- a/cmd/deploy/deploy.go
+++ b/cmd/deploy/deploy.go
@@ -92,15 +92,7 @@ func Deploy(ctx context.Context) *cobra.Command {
 			// This is needed because the deploy command needs the original kubeconfig configuration even in the execution within another
 			// deploy command. If not, we could be proxying a proxy and we would be applying the incorrect deployed-by label
 			os.Setenv(model.OktetoWithinDeployCommandContextEnvVar, "false")
-			ctxOpts := &contextCMD.ContextOptions{
-				Namespace: options.Namespace,
-				Show:      true,
-			}
-			if err := contextCMD.Run(ctx, ctxOpts); err != nil {
-				return err
-			}
-
-			if okteto.IsOkteto() {
+			if okteto.IsOkteto() && options.Namespace != "" {
 				create, err := utils.ShouldCreateNamespace(ctx, options.Namespace)
 				if err != nil {
 					return err
@@ -111,6 +103,13 @@ func Deploy(ctx context.Context) *cobra.Command {
 						return err
 					}
 				}
+			}
+
+			ctxOpts := &contextCMD.ContextOptions{
+				Namespace: options.Namespace,
+			}
+			if err := contextCMD.Run(ctx, ctxOpts); err != nil {
+				return err
 			}
 
 			cwd, err := os.Getwd()
--- a/cmd/deploy/deploy.go
+++ b/cmd/deploy/deploy.go
@@ -227,7 +227,7 @@ func (dc *deployCommand) runDeploy(ctx context.Context, cwd string, opts *Option
 		// Set BUILDKIT_PROGRESS=plain env variable, so all the commands disable docker tty builds
 		fmt.Sprintf("%s=true", model.BuildkitProgressEnvVar),
 		// Set OKTETO_NAMESPACE=namespace-name env variable, so all the commandsruns on the same namespace
-		fmt.Sprintf("%s=%s", model.OktetoNamespaceEnvVar, opts.Namespace),
+		fmt.Sprintf("%s=%s", model.OktetoNamespaceEnvVar, okteto.Context().Namespace),
 	)
 
 	for _, command := range opts.Manifest.Deploy.Commands {
--- a/cmd/deploy/deploy.go
+++ b/cmd/deploy/deploy.go
@@ -30,6 +30,7 @@ import (
 
 	"github.com/google/uuid"
 	contextCMD "github.com/okteto/okteto/cmd/context"
+	"github.com/okteto/okteto/cmd/namespace"
 	"github.com/okteto/okteto/cmd/utils"
 	"github.com/okteto/okteto/pkg/config"
 	"github.com/okteto/okteto/pkg/log"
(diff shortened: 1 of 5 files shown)


cmd/stack/deploy.go
```diff
@@ -47,6 +50,30 @@ func Deploy(ctx context.Context) *cobra.Command {
...
+				if err != nil {
+					return err
+				}
+				if !hasAccess {
+					log.Warning("Namespace '%s' not found", s.Namespace)
+					create, err := utils.AskYesNo(fmt.Sprintf("Do you want to create the namespace '%s'? [y/n] ", s.Namespace))
+					if err != nil {
+						return err
+					}
+					if create {
```
jLopezbarb, line 68: is there a way to not have so much indentation. I was thinking about encapsulating alll this code in a function but there is a lot of indentation as well...

cmd/stack/deploy.go
```diff
@@ -47,6 +50,30 @@ func Deploy(ctx context.Context) *cobra.Command {
...
+				if err != nil {
+					return err
+				}
+				hasAccess, err := utils.HasAccessToNamespace(ctx, s.Namespace, c)
+				if err != nil {
+					return err
+				}
+				if !hasAccess {
+					log.Warning("Namespace '%s' not found", s.Namespace)
+					create, err := utils.AskYesNo(fmt.Sprintf("Do you want to create the namespace '%s'? [y/n] ", s.Namespace))
```
jLopezbarb, line 64: @rlamana should we change the color of this question to the purple one we are using on ctx command? This also affects when stignore/okteto manifest is not found

rlamana: Yes, all questions or input asked to the user should have the same format in the CLI.

cmd/stack/deploy.go
```diff
@@ -47,6 +50,30 @@ func Deploy(ctx context.Context) *cobra.Command {
 				return err
 			}
 
+			if okteto.IsOkteto() {
```
pchico83, line 53: we should do the same for `oktack deploy`

cmd/stack/deploy.go
pchico83, line 64: `The namespace x doesn't exist. Do you want to create it?`

cmd/stack/deploy.go
pchico83: yes, I would have a function for this. also, I would fail fast and return an error if `create` is false. it will also help with indentation

pchico83 (approved):

cmd/stack/deploy.go
```diff
@@ -47,6 +50,30 @@ func Deploy(ctx context.Context) *cobra.Command {
...
+				c, err := okteto.NewOktetoClient()
+				if err != nil {
+					return err
+				}
+				hasAccess, err := utils.HasAccessToNamespace(ctx, s.Namespace, c)
```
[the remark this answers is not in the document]
jLopezbarb: If that happen when the user tries to create the namespace the backend will return an error when calling create namespace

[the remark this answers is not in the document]
ifbyol: I know, but maybe we can handle that scenario in a different way. You are asking the user if they want to create a namespace, they say yes, and then, it fails because there is a namespace with the same name. Anyways, I think this could be good enough and we can add that case if needed in the future

cmd/deploy/deploy.go
```diff
@@ -209,6 +227,8 @@ func (dc *deployCommand) runDeploy(ctx context.Context, cwd string, opts *Option
 		fmt.Sprintf("%s=true", model.OktetoDisableSpinnerEnvVar),
 		// Set BUILDKIT_PROGRESS=plain env variable, so all the commands disable docker tty builds
 		fmt.Sprintf("%s=true", model.BuildkitProgressEnvVar),
+		// Set OKTETO_NAMESPACE=namespace-name env variable, so all the commandsruns on the same namespace
+		fmt.Sprintf("%s=%s", model.OktetoNamespaceEnvVar, opts.Namespace),
```
ifbyol, line 231: Shouldn't this be `okteto.Context().Name` to use the same namespace we are using in the kubeconfig?

jLopezbarb: yes, you're right

ifbyol: yep, I didn't copy the whole sentence xDD

cmd/deploy/deploy.go
```diff
@@ -91,7 +92,23 @@ func Deploy(ctx context.Context) *cobra.Command {
 			// This is needed because the deploy command needs the original kubeconfig configuration even in the execution within another
 			// deploy command. If not, we could be proxying a proxy and we would be applying the incorrect deployed-by label
 			os.Setenv(model.OktetoWithinDeployCommandContextEnvVar, "false")
-			if err := contextCMD.Run(ctx, &contextCMD.ContextOptions{}); err != nil {
+			if okteto.IsOkteto() && options.Namespace != "" {
+				create, err := utils.ShouldCreateNamespace(ctx, options.Namespace)
```
ifbyol, line 96: For this case doesn't matter because pipeline installer is not specifying any namespace when using `okteto deploy` command but we have to bear in mind for the future that the command is being used in a non-interactive way in the installer

ifbyol (approved):

PR #2081, opened 2021-12-16
Add okteto sample annotation

jLopezbarb:
Add label to confirm if its okteto sample

## Proposed changes
- Creates the annotations okteto `dev.okteto.com/sample` to stack and deploy to check if it's an okteto repo

### DIFF
--- a/cmd/deploy/deploy.go
+++ b/cmd/deploy/deploy.go
@@ -365,6 +365,13 @@ func getProxyHandler(name, token string, clusterConfig *rest.Config) (http.Handl
 			}
 			metadata.Labels[model.DeployedByLabel] = name
 
+			if metadata.Annotations == nil {
+				metadata.Annotations = map[string]string{}
+			}
+			if utils.IsOktetoRepo() {
+				metadata.Annotations[model.OktetoSampleAnnotation] = "true"
+			}
+
 			metadataAsByte, err := json.Marshal(metadata)
 			if err != nil {
 				log.Infof("could not process resource's metadata: %s", err)
--- a/cmd/utils/git.go
+++ b/cmd/utils/git.go
@@ -17,10 +17,17 @@ import (
 	"context"
 	"fmt"
 	"math/rand"
+	"os"
 	"strings"
 
 	"github.com/go-git/go-git/v5"
+	"github.com/go-git/go-git/v5/plumbing/transport"
+	"github.com/okteto/okteto/pkg/log"
+	"github.com/okteto/okteto/pkg/model"
+	"k8s.io/utils/pointer"
 )
+
+var isOktetoSample *bool
 
 func GetBranch(ctx context.Context, path string) (string, error) {
 	repo, err := git.PlainOpen(path)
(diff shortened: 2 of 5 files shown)


cmd/deploy/deploy.go
```diff
@@ -198,6 +200,17 @@ func (dc *deployCommand) runDeploy(ctx context.Context, cwd string, opts *Option
 
 	defer dc.cleanUp(ctx)
 
+	c, _, err := dc.k8sClientProvider.Provide(okteto.Context().Cfg)
+	if err != nil {
+		return err
+	}
+
+	cfg := deploy.TranslateConfigMap(ctx, opts.Name)
```
ifbyol, line 208: This may impact on how a pipeline is displayed in the UI when the `okteto deploy` command is executed from the pipeline installer (where another config map is created). Maybe there is nothing to do as we are doing something similar with stacks but we should test it

jLopezbarb: We will set the annotation on every resource instead of setting it on the cfgMap

cmd/deploy/deploy.go
```diff
@@ -346,6 +350,13 @@ func getProxyHandler(name, token string, clusterConfig *rest.Config) (http.Handl
 			}
 			metadata.Labels[model.DeployedByLabel] = name
 
+			if metadata.Annotations == nil {
+				metadata.Annotations = map[string]string{}
+			}
+			if isOktetoRepo {
```
jmacelroy, line 356: This is not safe to use, at least by convention. It's technically probably fine but it's bad practice to use any other return value from a function if the err is non-nil.

Best practice would be to explicitly set `isOktetoRepo` to `false` in the error handling.

cmd/deploy/deploy.go
```diff
@@ -272,6 +272,10 @@ func getProxyHandler(name, token string, clusterConfig *rest.Config) (http.Handl
 	proxy := httputil.NewSingleHostReverseProxy(destinationURL)
 	proxy.Transport = trans
 
+	isOktetoRepo, err := utils.IsOktetoRepo()
```
jmacelroy, line 275: based on comment below on usage of these returned vars you might change this.
```suggestion
        var isOktetoRepo bool
	is, err := utils.IsOktetoRepo()
	if err == nil {isOktetoRepo = is} else{isOktetoRepo=false; log.info(....)}
```

pkg/cmd/stack/translate.go
```diff
@@ -755,13 +760,31 @@ func translateLabelSelector(svcName string, s *model.Stack) map[string]string {
...
 }
 
+func getAnnotations() map[string]string {
+	annotations := map[string]string{}
+	if isOktetoRepo == nil {
+		isOkteto, err := utils.IsOktetoRepo()
+		if err != nil {
+			log.Infof("could not check if it's okteto repo: %s", err)
+		}
+		isOktetoRepo = pointer.Bool(isOkteto)
```
jmacelroy, line 778: Same case here as previously mentioned.

pkg/cmd/stack/translate.go
```diff
@@ -755,13 +760,31 @@ func translateLabelSelector(svcName string, s *model.Stack) map[string]string {
...
+	}
+
+	if *isOktetoRepo {
```
pchico83, line 781: where is it initialized?

pkg/model/const.go
```diff
@@ -73,6 +76,9 @@ const (
 	// DeployedByLabel indicates the service account that deployed an object
 	DeployedByLabel = "dev.okteto.com/deployed-by"
 
+	// DeployLabel indicates the object is a deploy
```
pchico83, line 79: what is this for?

pkg/cmd/stack/translate.go
```diff
@@ -56,6 +57,10 @@ const (
 	pvcName = "pvc"
 )
 
+var (
+	isOktetoRepo *bool
```
ifbyol, line 61: Why does it have to be a global variable?

cmd/utils/git.go
```diff
@@ -85,3 +88,20 @@ func GetRandomSHA(ctx context.Context, path string) string {
 	}
 	return string(b)
 }
+
+func IsOktetoRepo() (bool, error) {
```
ifbyol, line 92: Can you add a test for this function including git URLs in their different forms (SSH and HTTPS)?

cmd/utils/git_test.go
```diff
@@ -105,3 +106,40 @@ func Test_getBranch(t *testing.T) {
...
+		{
+			name:     "https from okteto",
+			url:      "https://github.com/okteto/test.git",
+			expected: true,
+		},
+		{
+			name:     "https from okteto",
+			url:      "https://github.com/test/test.git",
+			expected: false,
+		},
```
ifbyol, line 135: Minor thing: can you include a couple of cases for ssh://git@github.com/...?

ifbyol (approved):

pchico83 (approved):

cmd/utils/git.go
```diff
@@ -85,3 +92,32 @@ func GetRandomSHA(ctx context.Context, path string) string {
 	}
 	return string(b)
 }
+
+func IsOktetoRepo() bool {
+	if isOktetoSample == nil {
+		path, err := os.Getwd()
+		if err != nil {
+			log.Infof("failed to get the current working directory in IsOktetoRepo: %v", err)
```
maroshii, line 100: Does this logs json if `-o json`?

PR #2088, opened 2021-12-17
autocreate fail if app with the same name exists

### LINKED ISSUE #2075: `autocreate: true` should fail if the deployment exists but it wasn't created by `okteto up`
**Describe the bug**
This should fail. If not, it can lead to inconsistent states, for example:

```
- autocreate: true
- okteto up
- okteto stack deploy
- okteto up
```

```
 x  Couldn't activate your development container
    Deployment.apps "vote-okteto" is invalid: spec.selector: Invalid value: v1.LabelSelector{MatchLabels:map[string]string{"stack.okteto.com/name":"compose-getting-started", "stack.okteto.com/service":"vote"}, MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: field is immutable
```

jLopezbarb:

Fixes #2075

## Proposed changes
- Check if okteto is deployed bu okteto up and fail if not
-

### DIFF
--- a/cmd/up/activate.go
+++ b/cmd/up/activate.go
@@ -63,6 +63,10 @@ func (up *upContext) activate() error {
 		return err
 	}
 
+	if v, ok := app.ObjectMeta().Annotations[model.OktetoAutoCreateAnnotation]; !ok || v != model.OktetoUpCmd {
+		return fmt.Errorf("resource '%s' already exists. Please consider changing the name of autocreate", up.Dev.Name)
+	}
+
 	if up.isRetry && !apps.IsDevModeOn(app) {
 		log.Information("Development container has been deactivated")
 		return nil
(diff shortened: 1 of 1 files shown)


cmd/up/activate.go
```diff
@@ -63,6 +63,10 @@ func (up *upContext) activate() error {
 		return err
 	}
 
+	if v, ok := app.ObjectMeta().Annotations[model.OktetoAutoCreateAnnotation]; !ok || v != model.OktetoUpCmd {
+		return fmt.Errorf("resource '%s' already exists. Please consider changing the name of autocreate", up.Dev.Name)
```
rberrelleza, line 67: what do we mean by `changing the name of autocreate`?

rberrelleza, line 67: I thin it would be better to leave the error as resource %s doesn't exist, and then add an `info` like: "use a different name in your okteto.yaml, or remove the autocreate property"

ifbyol (approved):

PR #2090, opened 2021-12-17
Jlopezbarb/fix json output

### LINKED ISSUE #2071: Incorrect JSON output in `okteto deploy`
**Describe the bug**
`okteto deploy -o json` should send all output in json format. There are a few lines that are not in json which makes it hard to parse the output:

```
$ okteto deploy -o json
 i  Using maroshii @ cloud.okteto.com as context
{"level":"info","stage":"okteto build -t okteto.dev/api:${OKTETO_GIT_COMMIT} api","message":" i  Running your build in tcp://addr:1234...","timestamp":1639512885}
{"level":"info","stage":"okteto build -t okteto.dev/api:${OKTETO_GIT_COMMIT} api","message":"#2 [internal] load .dockerignore","timestamp":1639512886}
...
{"level":"info","stage":"helm upgrade --install catalog chart --set tag=${OKTETO_GIT_COMMIT}","message":"TEST SUITE: None","timestamp":1639513441}
 i  Endpoints available:
  - https://catalog-chart-maroshii...
  - https://catalog-chart-maroshii...
```

**Expected behavior**
Either add a `--silent` option or send the output in json as the other lines with an empty stage (so it can be easily omitted on the other side)

jLopezbarb:
Fixes #2071

## Proposed changes
- If json is the output skip showing messages that can lead to parsing error
-


cmd/context/utils.go
```diff
@@ -200,7 +203,13 @@ func GetManifest(ctx context.Context, srcFolder string, opts ManifestOptions) (*
 		manifest.Filename = pipelinePath
 		return manifest, nil
 	}
-
+	ctxOptions := &ContextOptions{
+		Show:   true,
+		Silent: opts.Silent,
+	}
+	if err := Run(ctx, ctxOptions); err != nil {
```
ifbyol, line 210: If I'm not wrong, this would execute the ctx twice for `okteto deploy` and `okteto destroy` command. Moreover, this would be executed whenever we use the GetManifest function. Wouldn't it better to pass the `Silent` option when the context is initialized on each command?

ifbyol (commented):
Endpoints output is not in json format, I think that wouldn't fix completely the issue described in the ticket. Also, if you specify the `debug` output level, it doesn't display in JSON format either


Captura de pantalla 2021-12-20 a las 13 04 30

ifbyol:
Also, this output at the end of the destroy command printed by the library could also provoke issues in the installer with the structured logs

```W1220 13:10:48.923320   27817 warnings.go:70] networking.k8s.io/v1beta1 Ingress is deprecated in v1.19+, unavailable in v1.22+; use networking.k8s.io/v1 Ingress
W1220 13:10:48.923926   27817 warnings.go:70] batch/v1beta1 CronJob is deprecated in v1.21+, unavailable in v1.25+; use batch/v1 CronJob
W1220 13:10:48.948312   27817 warnings.go:70] extensions/v1beta1 Ingress is deprecated in v1.14+, unavailable in v1.22+; use networking.k8s.io/v1 Ingress
W1220 13:10:48.958545   27817 warnings.go:70] policy/v1beta1 PodDisruptionBudget is deprecated in v1.21+, unavailable in v1.25+; use policy/v1 PodDisruptionBudget
W1220 13:10:49.032808   27817 warnings.go:70] rbac.authorization.k8s.io/v1beta1 RoleBinding is deprecated in v1.17+, unavailable in v1.22+; use rbac.authorization.k8s.io/v1 RoleBinding
W1220 13:10:49.038789   27817 warnings.go:70] rbac.authorization.k8s.io/v1beta1 Role is deprecated in v1.17+, unavailable in v1.22+; use rbac.authorization.k8s.io/v1 Role```

Would there be a way to avoid those lines? When I implemented the command I wasn't able to avoid them

File: cmd/deploy/deploy.go [hunk 2]
```diff
@@ -105,6 +107,10 @@ func Deploy(ctx context.Context) *cobra.Command {
 				}
 			}
 
+			if options.OutputMode == "json" {
```
ifbyol, line 110: If we force it here, we should force it also in the destroy command

ifbyol (commented):
I have the feeling we are applying different patches for the output flag. For example, for json output we are hiding the message which displays the context to be used by the command and I think that could be very useful log to have in the installer pipeline to detect possible issues.

This solution is error prone as we are passing a flag through the code and adding conditions to check how to print a message.

If we need it quickly to unblock structured logs in the installer, we can merge it, but I think we should consider to change the way we log/print messages in the CLI to support this flag easily.

pchico83:
> I have the feeling we are applying different patches for the output flag. For example, for json output we are hiding the message which displays the context to be used by the command and I think that could be very useful log to have in the installer pipeline to detect possible issues.
> 
> This solution is error prone as we are passing a flag through the code and adding conditions to check how to print a message.
> 
> If we need it quickly to unblock structured logs in the installer, we can merge it, but I think we should consider to change the way we log/print messages in the CLI to support this flag easily.

@ifbyol @jLopezbarb could we use the logger library instead of prints? it could simplify how to manage the format of the output

ifbyol:
> > I have the feeling we are applying different patches for the output flag. For example, for json output we are hiding the message which displays the context to be used by the command and I think that could be very useful log to have in the installer pipeline to detect possible issues.
> > This solution is error prone as we are passing a flag through the code and adding conditions to check how to print a message.
> > If we need it quickly to unblock structured logs in the installer, we can merge it, but I think we should consider to change the way we log/print messages in the CLI to support this flag easily.
> 
> @ifbyol @jLopezbarb could we use the logger library instead of prints? it could simplify how to manage the format of the output

I think it is the best option. Not sure if all the things we do for the logs are valid using the logger library, but I think all the output logic should be together and not spread through the code

jLopezbarb:
We decided to go with @ifbyol approach and refactor the logger library to support different outputs. This means that we have to change all `fmt.Print`

cmd/context/selector.go
```diff
@@ -29,6 +29,7 @@ import (
 	"github.com/manifoldco/promptui"
 	"github.com/manifoldco/promptui/list"
 	"github.com/manifoldco/promptui/screenbuf"
+	"github.com/okteto/okteto/pkg/errors"
```
jmacelroy, line 32: Can we rename this import to oktetoErrors so that we keep open the default import if/when needed?

cmd/deploy/deploy.go
```diff
@@ -52,10 +52,13 @@ type Options struct {
 	Namespace    string
 	Variables    []string
 	Timeout      time.Duration
-	OutputMode   string
 	Manifest     *model.Manifest
 }
 
+type endpoints struct {
```
jmacelroy, line 58: Is this used anywhere?

cmd/deploy/deploy.go
```diff
@@ -417,7 +421,13 @@ func (dc *deployCommand) showEndpoints(ctx context.Context, opts *Options) error
 		sort.Slice(eps, func(i, j int) bool {
 			return len(eps[i]) < len(eps[j])
 		})
-		log.Information("Endpoints available:\n  - %s\n", strings.Join(eps, "\n  - "))
+		switch log.GetOutputFormat() {
+		case "json":
+			log.Println(fmt.Sprintf("Endpoints available:\n  - %s\n", strings.Join(eps, "\n  - ")))
```
jmacelroy, line 426: I think we should be able to simplify this and just use a single log.Printf without the Sprintf and Join call.

cmd/version.go
```diff
@@ -14,10 +14,9 @@
 package cmd
 
 import (
-	"fmt"
-
 	"github.com/okteto/okteto/cmd/utils"
 	"github.com/okteto/okteto/pkg/config"
+	"github.com/okteto/okteto/pkg/log"
```
jmacelroy, line 19: Preferably this would be `oktetoLog` as there's a built-in `log` with a well-known api.

pchico83 (approved):

pkg/log/format.go
```diff
@@ -0,0 +1,69 @@
...
+const (
+	ttyFormat   string = "tty"
+	plainFormat string = "plain"
+	jsonFormat  string = "json"
+)
+
+func (l *logger) getWriter(format string) OktetoWriter {
+	switch format {
+	case ttyFormat:
+		l.outputMode = "tty"
```
teresaromero, line 55: should we use the const instead of the strings ? 
also, not sure about this, but for ttyFormat would it make sense that the switch uses the default case?

pkg/log/json.go
```diff
@@ -0,0 +1,228 @@
...
+}
+
+// Fprintf prints a line with format
+func (w *JSONWriter) Fprintf(format string, a ...interface{}) {
+	w.Printf(format, a...)
+}
+
+// Fprintln prints a line with format
+func (w *JSONWriter) Fprintln(args ...interface{}) {
+	msg := fmt.Sprint(args...)
```
jpf-okteto, line 194: I see `log.Println()` being used in many places just to insert a new line, which is not needed when output format is json. I would add a clause here to return if msg length is zero. I don't find useful a log message as the following:

```json
{"level":"info","stage":"","msg":""}
```

jpf-okteto (changes requested):
+1 on:

- using constants instead of strings
- aliasing imports that collide with well known packages (e.g. errors, log)

everything else LGTM at first glance.

pkg/log/json.go
jpf-okteto: Logrus does inject an empty logline, with timestamp and level if no args are provided.

https://go.dev/play/p/Vg-jM2-rw-N

Maybe we should not use at all these `fmt.Println()` statements.

jmacelroy (approved):

cmd/deploy/deploy.go
```diff
@@ -521,7 +521,13 @@ func (dc *deployCommand) showEndpoints(ctx context.Context, opts *Options) error
 		sort.Slice(eps, func(i, j int) bool {
 			return len(eps[i]) < len(eps[j])
 		})
-		log.Information("Endpoints available:\n  - %s\n", strings.Join(eps, "\n  - "))
+		switch oktetoLog.GetOutputFormat() {
+		case "json":
```
jpf-okteto, line 525: use constant

main.go
```diff
@@ -89,16 +90,19 @@ func main() {
...
 
 		},
 		PersistentPostRun: func(ccmd *cobra.Command, args []string) {
-			log.Infof("finished %s", strings.Join(os.Args, " "))
+			oktetoLog.Infof("finished %s", strings.Join(os.Args, " "))
 		},
 	}
 
 	root.PersistentFlags().StringVarP(&logLevel, "loglevel", "l", "warn", "amount of information outputted (debug, info, warn, error)")
+	root.PersistentFlags().StringVarP(&outputMode, "output", "o", "tty", "output format (tty, plain, json)")
```
jpf-okteto, line 104: use constant at flag's default value

pkg/log/format.go
```diff
@@ -0,0 +1,70 @@
...
+	Println(args ...interface{})
+	Print(args ...interface{})
+	Fprintf(format string, a ...interface{})
+	Printf(format string, a ...interface{})
+
+	IsInteractive() bool
+}
+
+const (
+	ttyFormat   string = "tty"
```
jpf-okteto, line 47: export these constants so it can be accessed from elsewhere

pkg/ssh/manager_test.go
```diff
@@ -24,6 +24,7 @@ import (
 
 	"github.com/gliderlabs/ssh"
 	"github.com/okteto/okteto/pkg/log"
```
jpf-okteto, line 26: pending log statements to be refactored

jpf-okteto (changes requested):

cmd/version.go
jpf-okteto: fixed at https://github.com/okteto/okteto/pull/2090/commits/d6992e7ceb1e7a4ce52b30c04186665b1bc135b7

cmd/context/selector.go
jpf-okteto: fixed at https://github.com/okteto/okteto/pull/2090/commits/20e2c39a321e3888280cf4a1228413e68a8fa58b

cmd/deploy/deploy.go
```diff
@@ -522,7 +523,7 @@ func (dc *deployCommand) showEndpoints(ctx context.Context, opts *Options) error
 			return len(eps[i]) < len(eps[j])
 		})
 		switch oktetoLog.GetOutputFormat() {
-		case "json":
+		case log.TTYFormat:
```
jpf-okteto, line 526:
```suggestion
		case oktetoLog.TTYFormat:
```

cmd/deploy/deploy.go
```diff
@@ -35,6 +35,7 @@ import (
 	"github.com/okteto/okteto/pkg/cmd/build"
 	"github.com/okteto/okteto/pkg/config"
 	oktetoErrors "github.com/okteto/okteto/pkg/errors"
+	"github.com/okteto/okteto/pkg/log"
```
jpf-okteto, line 38:
```suggestion
```

jpf-okteto (changes requested):

pkg/log/format.go
jpf-okteto: done at https://github.com/okteto/okteto/pull/2090/commits/70a154bf61361af75ffe8195ec69ad0e0a0ba951

main.go
jpf-okteto: done at https://github.com/okteto/okteto/pull/2090/commits/70a154bf61361af75ffe8195ec69ad0e0a0ba951

cmd/deploy/deploy.go
jpf-okteto: done at https://github.com/okteto/okteto/pull/2090/commits/70a154bf61361af75ffe8195ec69ad0e0a0ba951

cmd/deploy/deploy.go
jpf-okteto: done at https://github.com/okteto/okteto/pull/2090/commits/e94628c8dfce42213346c67e745b2bc3e0d235fe

pkg/ssh/manager_test.go
jpf-okteto: done at https://github.com/okteto/okteto/pull/2090/commits/e94628c8dfce42213346c67e745b2bc3e0d235fe

jpf-okteto (approved):

PR #2093, opened 2021-12-17
Add context improvement

### LINKED ISSUE #2042: `okteto context` fixes and improvement

- `okteto context use xxx` should create a context if it doesn't exit
- Hide from `help` all context commands but `show`, `use`, `list` and `delete`

**Additional context**
This feature will go in 1.15-beta

jLopezbarb:
Fixes #2042 

## Proposed changes
- Hide ctx commands 
- okteto ctx use creates the ctx if you pass an url as argument

### DIFF
--- a/cmd/context/create.go
+++ b/cmd/context/create.go
@@ -16,6 +16,7 @@ package context
 import (
 	"context"
 	"fmt"
+	"net/url"
 	"os"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


cmd/context/create.go
```diff
@@ -238,3 +244,12 @@ func (*ContextUse) initEnvVars() {
 		}
 	}
 }
+
+func isNewUrl(u string) bool {
```
ifbyol, line 248: Where is this checking that the URL is new?

cmd/context/create.go
```diff
@@ -108,15 +109,20 @@ func (c *ContextUse) UseContext(ctx context.Context, ctxOptions *ContextOptions)
...
+		isNewUrl := isNewUrl(ctxOptions.Context)
+		if !isValidCluster(ctxOptions.Context) && !isNewUrl {
 			return errors.UserError{E: fmt.Errorf("invalid okteto context '%s'", ctxOptions.Context),
 				Hint: "Please run 'okteto context' to select one context"}
 		}
-
-		transformedCtx := okteto.K8sContextToOktetoUrl(ctx, ctxOptions.Context, ctxOptions.Namespace, c.k8sClientProvider)
-		if transformedCtx != ctxOptions.Context {
-			ctxOptions.Context = transformedCtx
+		if isNewUrl {
```
jmacelroy, line 117: Using the same name as the function in the same package is a bit odd. Is this always non-nil and therefore true because it's a function here or is this the return value and the function is no longer usable?

cmd/context/create.go
teresaromero: This is more like a "is Url?" func? why the new ? should be different from some previous url?

teresaromero (approved):

PR #2110, opened 2021-12-28
ManifestV2 context/namespace fixes

pchico83:

### DIFF
--- a/cmd/context/utils.go
+++ b/cmd/context/utils.go
@@ -200,6 +200,8 @@ func LoadManifestV2WithContext(ctx context.Context, namespace, path string) erro
 
 	manifest, err := GetManifestV2(cwd, path)
 	if err != nil {
+		//GetManifestV2 should take care of all error conditions and possible paths
+		//https://github.com/okteto/okteto/issues/2111
 		if err != errors.ErrManifestNotFound {
 			return err
 		}
--- a/cmd/context/utils.go
+++ b/cmd/context/utils.go
@@ -186,12 +186,43 @@ func LoadStackWithContext(ctx context.Context, name, namespace string, stackPath
 	return s, nil
 }
 
+//LoadManifestV2WithContext initializes the okteto context taking into account command flags and manifest namespace/context fields
+func LoadManifestV2WithContext(ctx context.Context, namespace, path string) error {
+	ctxOptions := &ContextOptions{
+		Namespace: namespace,
+		Show:      true,
+	}
+
+	cwd, err := os.Getwd()
+	if err != nil {
+		return err
+	}
+
+	manifest, err := GetManifestV2(cwd, path)
+	if err != nil {
+		if err != errors.ErrManifestNotFound {
+			return err
+		}
+	} else {
+		ctxOptions.Context = manifest.Context
+		if ctxOptions.Namespace == "" {
+			ctxOptions.Namespace = manifest.Namespace
+		}
+	}
+
+	if err := Run(ctx, ctxOptions); err != nil {
+		return err
+	}
+
+	return nil
+}
+
 // GetManifest Loads a manifest
 func GetManifest(ctx context.Context, srcFolder string, opts ManifestOptions) (*model.Manifest, error) {
 	pipelinePath := getPipelinePath(srcFolder, opts.Filename)
 	if pipelinePath != "" {
 		log.Debugf("Found okteto manifest %s", pipelinePath)
-		manifest, err := LoadManifestWithContext(ctx, pipelinePath, opts.Namespace, opts.K8sContext)
+		manifest, err := utils.LoadManifest(pipelinePath)
 		if err != nil {
 			log.Infof("could not load manifest: %s", err.Error())
 		}
@@ -208,10 +239,6 @@ func GetManifest(ctx context.Context, srcFolder string, opts ManifestOptions) (*
 	}
 
 	oktetoSubPath := getOktetoSubPath(srcFolder, src)
-	devs := []string{}
-	if oktetoSubPath != "" {
-		devs = append(devs, oktetoSubPath)
-	}
 	chartSubPath := getChartsSubPath(srcFolder, src)
 	if chartSubPath != "" {
 		log.Infof("Found chart")
@@ -400,5 +427,5 @@ func GetManifestV2(basePath, file string) (*model.Manifest, error) {
 	if manifestPath != "" {
 		return model.Get(manifestPath)
 	}
-	return nil, fmt.Errorf("okteto manifest not found")
+	return nil, errors.ErrManifestNotFound
 }
--- a/cmd/deploy/deploy.go
+++ b/cmd/deploy/deploy.go
@@ -50,7 +50,6 @@ type Options struct {
 	ManifestPath string
 	Name         string
 	Namespace    string
-	K8sContext   string
 	Variables    []string
 	Timeout      time.Duration
 	OutputMode   string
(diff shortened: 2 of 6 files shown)


cmd/context/utils.go
```diff
@@ -186,12 +186,43 @@ func LoadStackWithContext(ctx context.Context, name, namespace string, stackPath
...
+	}
+
+	cwd, err := os.Getwd()
+	if err != nil {
+		return err
+	}
+
+	manifest, err := GetManifestV2(cwd, path)
+	if err != nil {
+		if err != errors.ErrManifestNotFound {
```
jmacelroy, line 203: Is this still able to run if the manifest is not found? If so, I think this is worth a comment as to why the error handling is this way.

cmd/context/utils.go
```diff
@@ -186,12 +186,43 @@ func LoadStackWithContext(ctx context.Context, name, namespace string, stackPath
 	return s, nil
 }
 
+//LoadManifestV2WithContext initializes the okteto context taking into account command flags and manifest namespace/context fields
+func LoadManifestV2WithContext(ctx context.Context, namespace, path string) error {
```
jmacelroy, line 190: A utils dir and context package are not a great choice of package or filenames as it is a sort of grab bag and lacks focus. The package is probably what I care about most as it doesn't seem like good isolation to put all sorts of things that obtain info from context in here but should focus more on something like `manifest`. I don't think we should fix that package as part of this commit but we could stop growing it. I think this function might fit better in a package that's focused on the actual manifest or possibly a slightly less focused package related to general actions of getting, parsing and checking various config files that okteto uses?

cmd/context/utils.go
```diff
@@ -186,12 +186,43 @@ func LoadStackWithContext(ctx context.Context, name, namespace string, stackPath
...
+	return nil
+}
+
 // GetManifest Loads a manifest
 func GetManifest(ctx context.Context, srcFolder string, opts ManifestOptions) (*model.Manifest, error) {
 	pipelinePath := getPipelinePath(srcFolder, opts.Filename)
 	if pipelinePath != "" {
 		log.Debugf("Found okteto manifest %s", pipelinePath)
-		manifest, err := LoadManifestWithContext(ctx, pipelinePath, opts.Namespace, opts.K8sContext)
+		manifest, err := utils.LoadManifest(pipelinePath)
```
jmacelroy, line 225: This function might belong better with a good location for the new function above? I don't know how many users of LoadManifest there are but if it's a small number it's probably worth the move. Why is loading of manifest separated between so many packages that are focused on types, interfaces or are generic? How would someone easily begin to read code associated with manifests and could they do it with a simple text editor?

cmd/context/utils.go
```diff
@@ -400,5 +427,5 @@ func GetManifestV2(basePath, file string) (*model.Manifest, error) {
 	if manifestPath != "" {
 		return model.Get(manifestPath)
 	}
-	return nil, fmt.Errorf("okteto manifest not found")
+	return nil, errors.ErrManifestNotFound
```
jmacelroy, line 430: Overloading the built-in errors package can lead to confusion. This error is fairly specific and probably belongs in a package more focused on manifest. This type of organization would be more like that used with the Go language itself and makes it easier for new code contributors to associate related things while providing isolation from other custom errors we may make that probably belong in similar packages.

cmd/deploy/deploy.go
```diff
@@ -92,26 +91,24 @@ func Deploy(ctx context.Context) *cobra.Command {
 			// This is needed because the deploy command needs the original kubeconfig configuration even in the execution within another
 			// deploy command. If not, we could be proxying a proxy and we would be applying the incorrect deployed-by label
 			os.Setenv(model.OktetoWithinDeployCommandContextEnvVar, "false")
-			if okteto.IsOkteto() && options.Namespace != "" {
-				create, err := utils.ShouldCreateNamespace(ctx, options.Namespace)
+
+			if err := contextCMD.LoadManifestV2WithContext(ctx, options.Namespace, options.ManifestPath); err != nil {
```
jmacelroy, line 95: Having to rename this import because it collides with a built-in package is a smell that indicating about some of the previous comments made on package organization.

cmd/deploy/deploy.go
```diff
@@ -92,26 +91,24 @@ func Deploy(ctx context.Context) *cobra.Command {
...
+				return err
+			}
+
+			if okteto.IsOkteto() {
+				create, err := utils.ShouldCreateNamespace(ctx, okteto.Context().Namespace)
```
jmacelroy, line 100: `SouldCreateNamespace` and `ExecuteCreateNamespace` are in different packages?

cmd/deploy/deploy.go
```diff
@@ -185,9 +182,8 @@ func Deploy(ctx context.Context) *cobra.Command {
 	}
 
 	cmd.Flags().StringVar(&options.Name, "name", "", "application name")
-	cmd.Flags().StringVarP(&options.ManifestPath, "file", "f", "", "path to the manifest file")
-	cmd.Flags().StringVar(&options.Namespace, "namespace", "", "application name")
-	cmd.Flags().StringVar(&options.K8sContext, "context", "", "k8s context")
+	cmd.Flags().StringVarP(&options.ManifestPath, "file", "f", "", "path to the okteto manifest file")
+	cmd.Flags().StringVarP(&options.Namespace, "namespace", "n", "", "overwrites the namespace where the application is deployed")
```
jmacelroy, line 186: What is being overridden here and is it obvious to a cli user what they are overriding?

jmacelroy (commented):
I don't know how much of it can change in the context of this one PR but I do believe it's worth it to start to improve some of the organization and structure with this commit.

cmd/context/utils.go
pchico83: Added a comment. I also open an issue to refactor how the manifest is read at the moment. Right now the logic is split into different functions and it should be a single function, with a cleaner interface, and in a different package:
https://github.com/okteto/okteto/issues/2111

cmd/context/utils.go
pchico83: It's not the built-in errors package, it's own errors package:
https://github.com/okteto/okteto/tree/master/pkg/errors

We are using a single errors package to simplify text errors reviews by non technical people

cmd/deploy/deploy.go
pchico83: Yes, `ShouldCreateNamespace` should be in a different package. I created an issue to track it:
https://github.com/okteto/okteto/issues/2112

cmd/deploy/deploy.go
pchico83: Yes, we have this flag in all our CLI commands. It allows to deploy the app in a different namespace than the one of the current okteto context. What isn't clear? How would you rephrase it?

pchico83:
> I don't know how much of it can change in the context of this one PR but I do believe it's worth it to start to improve some of the organization and structure with this commit.

I created a few issues to track some of the reactors, but this PR isn't the right place to refactor code, it's bug fix.
In particular, this one is important and requires prioritization:
https://github.com/okteto/okteto/issues/2111

cmd/deploy/deploy.go
jmacelroy: It's not obvious to me where the original namespace exists. I don't know if it's in the manifest file or part of my context. How would a user know what there current namespace is and if it's something they want to override. I would probably state where the namespace being overwritten is?

cmd/context/utils.go
jmacelroy: It's not the built-in errors package but we've taken that name which is bad form. It can goes against the grain of what someone familiar with just the built-in package is used to and causes extra cognitive load.

Do we have anyone reviewing the errors? I think I'd rather build the code to have better organization for the heaviest use case, tech reader and implementors. I do think we can solve the non-tech review case in other ways though if it's a real requirement?

jmacelroy:
> > I don't know how much of it can change in the context of this one PR but I do believe it's worth it to start to improve some of the organization and structure with this commit.
> 
> I created a few issues to track some of the reactors, but this PR isn't the right place to refactor code, it's bug fix. In particular, this one is important and requires prioritization: #2111

Awesome! Yep, I agree with this and I think the ideas I put forth are more of a long-term thing and require some discussion. I do *not* want to prevent any good progress in the interim.

PR #2114, opened 2022-01-04
Okteto deploy build flag

teresaromero:
Fixes #

## Proposed changes

- added build flag to okteto deploy command
- with this flag and a enable flag for V2 manifest, the manifest is searched for, and the images defined at the build section are being built
- after this the deploy is executed according to the deploy commands declared at the new manifest.


cmd/deploy/deploy.go
```diff
@@ -197,13 +201,45 @@ func (dc *deployCommand) runDeploy(ctx context.Context, cwd string, opts *Option
...
+	if opts.Build && contextCMD.IsManifestV2Enabled() {
+		opts.Manifest, err = contextCMD.GetManifestV2(cwd, opts.ManifestPath)
+		if err != nil {
+			return err
+		}
+		log.Debug("found okteto manifest")
+
+		if opts.Manifest.Deploy == nil {
+			return fmt.Errorf("found okteto manifest, but no deploy commands where defined")
+		}
```
ifbyol, line 214: Maybe this check can be moved after the if/else (line 244). I guess line 238 can also return a manifest without Deploy section

cmd/deploy/deploy.go
```diff
@@ -197,13 +201,45 @@ func (dc *deployCommand) runDeploy(ctx context.Context, cwd string, opts *Option
...
+		if opts.Manifest.Build != nil {
+			log.Debug("running build from manifest definition")
+
+			var buildErrs []string
+
+			for service, buildInfo := range opts.Manifest.Build {
+				buildOptions := build.BuildOptions{}
+				opts := build.OptsFromManifest(service, buildInfo, buildOptions)
+
+				ctx := context.Background()
```
ifbyol, line 225: just curiosity, why do we need to create a new context for each build? I think you can use the one received on `runDeploy`

cmd/deploy/deploy.go
teresaromero: I added the check because when i tested with empty deploy/not deploy declaration, a panic was triggered in line 267. There are two reasons behind this:
- one that, in case there is no deploy declaration, the command okteto deploy --build makes little sense, so why would i want to build the images if i dont even have a deploy command? so i'm returning at this point for that.
- other, I haven't tested all scenarios when "old" manifest is used, which will use the dc.gestManifest, this always returns a deploy section with commands depending of the "manifest" found (stack, pipeline, helm, etc...) so its kindof... never empty? i want to keep the changes as small as possible.

would it be ok to add the check (will be at the two lines) on line 266 to catch this possible panic? @ifbyol

cmd/deploy/deploy.go
teresaromero: i was using the same code from the loop at the build command, i will remove it.

cmd/deploy/deploy.go
ifbyol: I would add it, but before starting the proxy, line 244 should be fine, or even inside the else branch (line 242)

teresaromero:
@pchico83  While testing okteto deploy to deploy a stack, i've detected that the image tag created in case is not defined at the stack definition is different than the one created in case there is no image at the build definition of the new manifest, question is: should this images strings be the same in case a stack.yaml has no image and a okteto.yaml build section has no image? 
https://github.com/okteto/okteto/blob/a7deeba114e49c67bac9264de1c3f7cee2581635/pkg/cmd/stack/translate.go#L148 

"okteto.dev/movies-fronten:okteto"
"okteto.dev/frontend:dev"

pchico83:
> "okteto.dev/movies-fronten:okteto"

Yes, let's always create "okteto.dev/movies-frontent:okteto"

pchico83 (approved):

jLopezbarb (approved):

PR #2118, opened 2022-01-07
Okteto build variables setting

teresaromero:
Fixes #

## Proposed changes

- Set the ENV variables which the user can define at the manifest to refer to image, repository, tag and registry
- replace to this env values when using the manifest to deploy


cmd/deploy/deploy.go
```diff
@@ -226,12 +227,36 @@ func (dc *deployCommand) runDeploy(ctx context.Context, cwd string, opts *Option
 					buildErrs = append(buildErrs, err.Error())
 					continue
 				}
+
+				if okteto.IsOkteto() && registry.IsOktetoRegistry(opts.Tag) {
```
pchico83, line 231: We should expand envars for all type of registries, no?

pkg/registry/registry.go
```diff
@@ -40,6 +42,40 @@ type ConfigInfo struct {
 	ExposedPorts *map[string]*interface{} `json:"ExposedPorts"`
 }
 
+// GetDigest returns the pull digest for the image at the registry
+func GetDigest(registry, repository, tag string) (digest.Digest, error) {
```
pchico83, line 46: @jLopezbarb 
don't we have a function for this already?

cmd/deploy/deploy.go
```diff
@@ -283,6 +309,19 @@ func (dc *deployCommand) runDeploy(ctx context.Context, cwd string, opts *Option
 	return nil
 }
 
+func setManifestEnvVars(service, registry, repository, digest string) {
+	os.Setenv(fmt.Sprintf("build.%s.registry", service), okteto.Context().Registry)
```
pchico83, line 313: Not sure if `${api.tag}` would be cleaner than `${build.api.tag}. @rlamana what do you think?

pkg/registry/registry.go
teresaromero: We do, but to retrieve the digest that is "callable" the call has to include a special header to the registry. i spotted the same problem when i implemented at the backend https://github.com/okteto/app/pull/2954/files#diff-ed67182f5de738fa9e73e3a19fea061f98446043896d7588c6f2d7f4e7af01a3R36

pchico83: Why do you need it to be "callable"? In any case, is it possible to refactor both functions in a single one?

teresaromero: if i set this as tag, the sha does not correspond to the sha used as an image tag, so it can't pull the image.

cmd/deploy/deploy.go
pchico83: Also, I think we should prepend the pipeline name to the image name, to avoid collisions with same build names in a different pipeline

teresaromero: > We should expand envars for all type of registries, no?
you mean expand or set? i thought we could only access the registry when is the okteto registry, so the envs are related to the okteto registries? 

> prepend the pipeline name to the image name

you mean the name param in the manifest? i was looking into this today to change it when we create the image field when parsing the manifest, but this field is optional, so in case there is no name at the manifest, what would be?

pchico83: it's extracted from the repo name. @ifbyol could you link the function that extract the pipeline name?

ifbyol: We use this function https://github.com/okteto/okteto/blob/master/cmd/utils/manifest.go#L24 to get the name from the repo URL or the current folder (if it is not a repository)

pchico83 (approved):

cmd/deploy/deploy.go
```diff
@@ -287,6 +314,19 @@ func (dc *deployCommand) runDeploy(ctx context.Context, cwd string, opts *Option
...
+func setManifestEnvVars(service, registry, repository, image, tag string) {
+	os.Setenv(fmt.Sprintf("build.%s.registry", service), registry)
+	os.Setenv(fmt.Sprintf("build.%s.repository", service), repository)
+	os.Setenv(fmt.Sprintf("build.%s.image", service), image)
+	os.Setenv(fmt.Sprintf("build.%s.tag", service), tag)
+
+	log.Debug("manifest env vars set")
+}
+
+func expandManifestEnvVars(manifest string) string {
```
ifbyol, line 326: Minor thing: Can we change the param name? If I'm not wrong, this is not intended to receive an Okteto manifest to expand any env variable on it, right? I think it can be misleading. As it is a one-liner function maybe we can just replace its use. Anyways, no strong opinion, you can leave it as it is :)

ifbyol (approved):

jLopezbarb (approved):

Issue #1982, opened 2021-11-17
Support for `okteto logs` command


pchico83:
**Is your feature request related to a problem? Please describe.**
I have several services running in my namespace and I need an easy way to query logs across containers, pods, deployments, ...

**Describe the solution you'd like**
An `okteto logs` command that, by default, shows all the output generated by the containers in my namespace in the last 5m. It could have labels to filter the output by service, to show/hide timestamps and to follow logs in streaming. Also, each log line will show the container and the pod name generating that line.

jLopezbarb:
@maroshi Is this issue fixed by #2120?

maroshii:
This will be fixed by #2120 yeah but there's still some debate around how it should work

PR #2120, opened 2022-01-08
Add logs command

maroshii:
Fixes https://github.com/okteto/okteto/issues/1982

Adds `okteto logs` command through [stern](https://github.com/stern/stern) with a more simplified UX. `okteto logs` streams the output of the pods created by your okteto manifest. As the first argument, you can pass a string to filter pods by name using regular expressions.

All the options supported are:
```
  -c, --context string     context where the logs are fetched
  -e, --exclude string     exclude by pod name (regular expression)
  -f, --file string        path to the manifest file
      --follow             follow log output
  -h, --help               help for logs
  -n, --namespace string   overrides the namespace where logs are fetched
  -s, --since duration     return logs newer than a relative duration like 5s, 2m, or 3h (default 48h0m0s)
      --tail int           the number of lines from the end of the logs to show (default 100)
  -t, --timestamps         Print timestamps
```


pchico83:
This is awesome!

I would have a highger-level UX not based on pods. I would start with something simple like:

- With arguments, it queries all pods in the namespace. If the current folder has a pipeline, just queries the logs of the pods deployed by the pipeline. We add a label for everything deployed by a pipeline
- I would start with these flags: namespace, output, timezone, tail, since, timestamps and **follow*.
- The argument is to filter by application (deployment, statefulset, cronjob, daemonset or job), instead of by pod. For other filters, we can rely on pipes to `grep`

ifbyol:
> * With arguments, it queries all pods in the namespace. If the current folder has a pipeline, just queries the logs of the pods deployed by the pipeline. We add a label for everything deployed by a pipeline

The only exception are the pipeline jobs. Those don't have the `deployed-by` label, but in that case we include the label `dev.okteto.com/pipeline-name`

pchico83:
@ifbyol what about pods? I mean, pods created by a deployment, for example

ifbyol:
> @ifbyol what about pods? I mean, pods created by a deployment, for example

@pchico83  For the installer pods, we are including the labels `dev.okteto.com/pipeline-name`, but for the pods deployed within a pipeline, we are not including the `dev.okteto.com/deployed-by` label because we are only adding them to the first level object, but not for the template, so pods created by a deployment or statefulset deployed by the pipeline wouldn't have that label

maroshii:
Stern works directly with pods and pod labels. As Nacho says, if we could add the `dev.okteto.com/pipeline-name` to the pods via the template we could get logs for a single pipeline

pchico83:
> Stern works directly with pods and pod labels. As Nacho says, if we could add the `dev.okteto.com/pipeline-name` to the pods via the template we could get logs for a single pipeline

I can see two options here:
1) use the mutation webhook to check the label of the pod controller and add the label to the pod
2) query deployments, stataefulsets... to get the pod labels

1) sounds better, but it requires extra calls on the mutation webhook which is not a good practice in general. Also, it wouldn't work on vanilla clusters

maroshii:
We don't need a label to query. If no labels are specified it queries all pods in the namespace that match a regex: (`stern api-*`) otherwise it queries pods by labels (`stern -l app.kubernetes.io/component=api`). It requires either a label or a pod query as a regex.

The issue with number 2 (**query deployments, statefulsets... to get the pod labels**) is that stern watches pods not k8s apps so we cannot do the watch on those higher level components (to track redeployments for eg). This is actually what we have in the backend and it's quite cumbersome since we need watch deployments, jobs, cronjobs, statefulsets etc independently. If we can make it so that we can get logs directly through static pod labels we can replace all the backend logging implementation with this as well.

pchico83:
> We don't need a label to query. If no labels are specified it queries all pods in the namespace that match a regex: (`stern api-*`) otherwise it queries pods by labels (`stern -l app.kubernetes.io/component=api`). It requires either a label or a pod query as a regex.
> 
> The issue with number 2 (**query deployments, statefulsets... to get the pod labels**) is that stern watches pods not k8s apps so we cannot do the watch on those higher level components (to track redeployments for eg). This is actually what we have in the backend and it's quite cumbersome since we need watch deployments, jobs, cronjobs, statefulsets etc independently. If we can make it so that we can get logs directly through static pod labels we can replace all the backend logging implementation with this as well.

Then I would go with @rberrelleza suggestion for now. We can scope it by repo later

maroshii:
fixes #1982

pchico83:
@maroshii one option to limit the scope to the current repo is to use the proxy. Eveyrtime it gets a a request to deploy/update a deployment, statefulset, job, cronjob or daemonset, we add the label to the object itself, but also to the pod spec of the deployment, statefulset, job, cronjob or daemonset.

ifbyol:
> @maroshii one option to limit the scope to the current repo is to use the proxy. Eveyrtime it gets a a request to deploy/update a deployment, statefulset, job, cronjob or daemonset, we add the label to the object itself, but also to the pod spec of the deployment, statefulset, job, cronjob or daemonset.

In the pod spec we are not including the label, we would have to include it

cmd/logs/logs.go
```diff
@@ -0,0 +1,290 @@
...
+					E: fmt.Errorf("failed to get logs: %w", err),
+				}
+			}
+			return nil
+		},
+		Args: utils.MaximumNArgsAccepted(1, ""),
+	}
+
+	cmd.Flags().StringVarP(&o.namespace, "namespace", "n", o.namespace, "overrides the namespace where the pods are deployed")
+	cmd.Flags().StringVar(&o.selector, "selector", o.selector, "Selector (label query) to filter on. If present, default to \".*\" for the pod-query.")
```
pchico83, line 116: let's remove this for now

cmd/logs/logs.go
```diff
@@ -0,0 +1,290 @@
...
+	cmd.Flags().StringVar(&o.fieldSelector, "field-selector", o.fieldSelector, "Selector (field query) to filter on. If present, default to \".*\" for the pod-query.")
```
pchico83, line 117: let's remove this for now

cmd/logs/logs.go
```diff
@@ -0,0 +1,290 @@
...
+	cmd.Flags().StringVar(&o.color, "color", o.color, "Force set color output. 'auto':  colorize if tty attached, 'always': always colorize, 'never': never colorize.")
+	cmd.Flags().StringVarP(&o.containerQuery, "container", "c", o.containerQuery, "Container name when multiple containers in pod. (regular expression)")
+	cmd.Flags().StringVarP(&o.output, "output", "o", o.output, "Specify predefined template. Currently support: [default, raw, json]")
+	cmd.Flags().StringSliceVar(&o.containerStates, "container-state", o.containerStates, "Tail containers with state in running, waiting or terminated. To specify multiple states, repeat this or set comma-separated value.")
+	cmd.Flags().StringVar(&o.timezone, "timezone", o.timezone, "Set timestamps to specific timezone.")
+	cmd.Flags().BoolVar(&o.initContainers, "init-containers", o.initContainers, "Include or exclude init containers.")
```
pchico83, line 123: default to `false`?

cmd/logs/logs.go
```diff
@@ -0,0 +1,290 @@
...
+	cmd.Flags().BoolVar(&o.ephemeralContainers, "ephemeral-containers", o.ephemeralContainers, "Include or exclude ephemeral containers.")
```
pchico83, line 124: default to `false`?

cmd/logs/logs.go
```diff
@@ -0,0 +1,290 @@
...
+	cmd.Flags().DurationVarP(&o.since, "since", "s", o.since, "Return logs newer than a relative duration like 5s, 2m, or 3h.")
+	cmd.Flags().Int64Var(&o.tail, "tail", o.tail, "The number of lines from the end of the logs to show. Defaults to -1, showing all logs.")
+	cmd.Flags().StringSliceVarP(&o.exclude, "exclude", "e", o.exclude, "Log lines to exclude. (regular expression)")
+	cmd.Flags().StringVarP(&o.excludeContainer, "exclude-container", "E", o.excludeContainer, "Container name to exclude when multiple containers in pod. (regular expression)")
+	cmd.Flags().StringVar(&o.excludePod, "exclude-pod", o.excludePod, "Pod name to exclude. (regular expression)")
```
pchico83, line 129: let's remove this for now

cmd/logs/logs.go
```diff
@@ -0,0 +1,114 @@
...
+			ctx, cancel := context.WithCancel(ctx)
+			defer cancel()
+
+			manifest, err := contextCMD.LoadManifestWithContext(ctx, contextCMD.ManifestOptions{Filename: options.ManifestPath, Namespace: options.Namespace, K8sContext: options.Context})
+			if err != nil {
+				return err
+			}
+
+			if len(args) > 0 {
+				options.Include = args[0]
```
rberrelleza, line 64: Could this be more than one service name?

pchico83: it's a regexp, you can use regexp for more than one service

cmd/logs/logs.go
```diff
@@ -0,0 +1,114 @@
...
+				return errors.UserError{
+					E: fmt.Errorf("failed to get logs: %w", err),
+				}
+			}
+			return nil
+		},
+		Args: utils.MaximumNArgsAccepted(1, ""),
+	}
+
+	cmd.Flags().StringVarP(&options.ManifestPath, "file", "f", "", "path to the manifest file")
```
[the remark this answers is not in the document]
pchico83: I know, but in our case `-f` is reserved. Maybe we can use `--watch` for `follow`

cmd/logs/logs.go
```diff
@@ -0,0 +1,114 @@
...
+	cmd.Flags().StringVarP(&options.Namespace, "namespace", "n", "", "overrides the namespace where logs are fetched")
+	cmd.Flags().StringVarP(&options.Context, "context", "c", "", "context where the logs are fetched")
```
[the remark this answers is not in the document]
pchico83: you can always overwrite the context with `-c` and the namespace with `-n`. It's how it works for the majority of commands

cmd/logs/logs.go
```diff
@@ -0,0 +1,114 @@
...
+	cmd.Flags().StringVarP(&options.exclude, "exclude", "e", "", "exclude by pod name (regular expression)")
```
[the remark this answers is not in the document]
pchico83: `include` is the default, it is the only optional argument of `okteto logs`.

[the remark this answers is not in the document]
pchico83: yes, good point, I will update all help messages

cmd/logs/logs.go
```diff
@@ -0,0 +1,114 @@
...
+	cmd.Flags().BoolVarP(&options.Follow, "follow", "", false, "follow log output")
+	cmd.Flags().DurationVarP(&options.Since, "since", "s", 48*time.Hour, "return logs newer than a relative duration like 5s, 2m, or 3h")
+	cmd.Flags().Int64Var(&options.Tail, "tail", 100, "the number of lines from the end of the logs to show")
+	cmd.Flags().BoolVarP(&options.Timestamps, "timestamps", "t", false, "Print timestamps")
+
+	return cmd
+}
+
+// GetTempKubeConfigFile returns where the temp kubeConfigFile for deploy should be stored
+func GetTempKubeConfigFile(name string) string {
```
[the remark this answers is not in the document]
pchico83: We don't store kubeconfig in filesystem, we query it dynamiclly based on the okteto context. But `stern` requires the kubeconfig to be on the filesystem. Something similar happens with `okteto deploy`

cmd/logs/logs.go
```diff
@@ -0,0 +1,114 @@
...
+	Include      string
+	Since        time.Duration
+	Tail         int64
+	Timestamps   bool
+}
+
+func Logs(ctx context.Context) *cobra.Command {
+	options := &LogsOptions{}
+
+	cmd := &cobra.Command{
```
pchico83, line 51: remind to myself: add metrics

pkg/types/logs.go
```diff
@@ -0,0 +1,28 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package types
+
+import "time"
+
+type LogsOptions struct {
```
pchico83, line 18: remove this file

cmd/logs/logs.go
pchico83: maybe we can automatically do `okteto logs a b` -> `okteto logs a|b`

pchico83: also, I was considering adding a `--all` flag to list logs from the whole namespace, not just restricted to your git repo

cmd/logs/logs.go
```diff
@@ -0,0 +1,121 @@
...
+	cmd.Flags().StringVarP(&options.exclude, "exclude", "e", "", "exclude by service name (regular expression)")
+	cmd.Flags().BoolVarP(&options.Watch, "watch", "w", false, "watch the log output")
+	cmd.Flags().DurationVarP(&options.Since, "since", "s", 48*time.Hour, "return logs newer than a relative duration like 5s, 2m, or 3h")
+	cmd.Flags().Int64Var(&options.Tail, "tail", 100, "the number of lines from the end of the logs to show")
+	cmd.Flags().BoolVarP(&options.Timestamps, "timestamps", "t", false, "print timestamps")
+
+	return cmd
+}
+
+// GetTempKubeConfigFile returns where the temp kubeConfigFile for deploy should be stored
```
teresaromero, line 117:
```suggestion
// GetTempKubeConfigFile returns where the temp kubeConfigFile for logs should be stored
```

cmd/logs/stern.go
```diff
@@ -0,0 +1,110 @@
...
+	if o.exclude != "" {
+		excludePodQuery, err = regexp.Compile(o.exclude)
+		if err != nil {
+			return nil, fmt.Errorf("failed to compile regular expression for excluded pod query: %w", err)
+		}
+	}
+
+	labelSelector := labels.NewSelector()
+	if !o.All {
+		req, err := labels.NewRequirement(model.DeployedByLabel, selection.Equals, []string{manifest.Name})
```
teresaromero, line 53: i've been reading the conversations around pipelines, thinking out loud 🤔 now the pipelines will have a deployed by label, as the pipeline uses okteto deploy? but, the logs from the installer from running a pipeline/dependency will still be needed to be sse?

teresaromero (approved):

cmd/logs/stern.go
pchico83: @teresaromero not sure I follow your question. `okteto logs` are for user applications, running in the developer namespace. The installer runs in the okteto namespace

teresaromero: > @teresaromero not sure I follow your question. `okteto logs` are for user applications, running in the developer namespace. The installer runs in the okteto namespace

sorry i think i did not express myself:
- due to this comment https://github.com/okteto/okteto/pull/2120#issuecomment-1008680569 i was thinking, 
- if this `okteto logs` will show the logs of a user application deployed by a pipeline run at the installer? or we have some kind of limitation with the levels still? (this is from January so maybe the behaviour has changed)
- if this is the case, the `sse` logs that are now streamed to clientes from the installer, are still needed or will this be a replacement?

pchico83: @teresaromero we now have the label `dev.okteto.com/deployed-by` on every pod on every namespace

maroshii: @pchico83 I think Tere is referring to this PR: https://github.com/okteto/okteto/pull/3073 which is using sse to get the logs. Maybe we can use stern internally instead?

cmd/logs/logs.go
```diff
@@ -0,0 +1,121 @@
...
+			analytics.TrackLogs(err == nil, options.All)
+
+			if err != nil {
+				return errors.UserError{
+					E: fmt.Errorf("failed to get logs: %w", err),
+				}
+			}
+			return nil
+		},
+		Args: utils.MaximumNArgsAccepted(1, ""),
```
AgustinRamiroDiaz, line 101:
```suggestion
		Args: cobra.MaximumNArgs(1),
```

cmd/logs/stern.go
```diff
@@ -0,0 +1,110 @@
...
+	include, err := regexp.Compile(".*")
+	if err != nil {
+		return nil, fmt.Errorf("failed to compile regular expression for include: %w", err)
+	}
+	containerStates := []stern.ContainerState{"running"}
+	fieldSelector := fields.Everything()
+
+	return &stern.Config{
+		KubeConfig:          kubeconfigFile,
+		ContextName:         okteto.UrlToKubernetesContext(okteto.Context().Name),
```
AgustinRamiroDiaz, line 89: Do we have another way of getting the context? cc @jLopezbarb

AgustinRamiroDiaz (approved):

cmd/logs/logs.go
jLopezbarb: If we are not going to add a link to the docs maybe we should use cobra's function but I think we should add it a link to the docs as we have for all the other commands

pchico83: Right, I will add the link to the docs

cmd/logs/stern.go
jLopezbarb: No, We have to use the k8s ctx instead of the cluster URL and this is the best way to transform an URL to a kubernetes context.

jLopezbarb (approved):

PR #2123, opened 2022-01-11
fix: fail if ctx show does not have ctx set

### LINKED ISSUE #2117: okteto context show shouldn't prompt for context creation
I think that ‘okteto context show’ should fail there’s no context set in the machine, instead of prompting you to create a context. That could be used as a way to check if a context is set on the machine, for scripting purposes.

I'd like `okteto context show` to fail with an exit code of -1 if there is no okteto context on the machine.

jLopezbarb:

Fixes #2117

## Proposed changes
- I created a new ctxOption which is raiseCtxNotSet. I think its better to have a default value on every command and only use that new option on `context show`
-

### DIFF
--- a/cmd/context/cmd_test.go
+++ b/cmd/context/cmd_test.go
@@ -1,0 +1,81 @@
+// Copyright 2021 The Okteto Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package context
+
+import (
+	"testing"
+)
+
+func Test_NoArgsAcceptedCtx(t *testing.T) {
+	cmd := Show()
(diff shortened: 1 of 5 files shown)


pkg/errors/errors.go
```diff
@@ -48,6 +48,9 @@ var (
 	// ErrNotLogged is raised when we can't get the user token
 	ErrNotLogged = "your token is invalid. Please run 'okteto context use %s' and try again"
 
+	// ErrCtxNotSet is raised when we don't have ctx set
+	ErrCtxNotSet = fmt.Errorf("your context is not set. Please run 'okteto context use' and and select your context")
```
pchico83, line 52:
```suggestion
	ErrCtxNotSet = fmt.Errorf("your context is not set. Please run 'okteto context' and and select your context")
```

pchico83 (approved):

pkg/errors/errors.go
rberrelleza: There's a double `and and` in the suggestion.

rberrelleza, line 52:
```suggestion
	ErrCtxNotSet = fmt.Errorf("your context is not set. Please run 'okteto context use' to set the default context")
```

RinkiyaKeDad (approved):

cmd/context/cmd_test.go
```diff
@@ -0,0 +1,81 @@
+// Copyright 2021 The Okteto Authors
```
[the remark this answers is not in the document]
pchico83: @RinkiyaKeDad I created this issue to track it

pchico83 (approved):

PR #2125, opened 2022-01-11
Skip build if volume mount is already mounted

### LINKED ISSUE #2096: `okteto stack deploy` building image on every deploy
**Describe the bug**
If I have a docker-compose like this one:

```
services:
  vote:
    build: vote
    ports:
      - 8080:8080
    volumes:
      - vote:/src

  redis:
    image: redis
    ports:
      - 6379
    volumes:
      - redis:/data
```

The `vote` image is built on every `okteto stack deploy` because there is a mount volume, but it should only be built the first time.

jLopezbarb:

Fixes #2096

## Proposed changes
- Skip build if the volume mount is already mounted and built 
-

### DIFF
--- a/pkg/cmd/stack/translate.go
+++ b/pkg/cmd/stack/translate.go
@@ -190,10 +190,6 @@ func addVolumeMountsToBuiltImage(ctx context.Context, s *model.Stack, options *S
 	for name, svc := range s.Services {
 		notSkippableVolumeMounts := getAccessibleVolumeMounts(s, name)
 		if len(notSkippableVolumeMounts) != 0 {
-			if !hasBuiltSomething && !hasAddedAnyVolumeMounts {
-				hasAddedAnyVolumeMounts = true
-				log.Information("Running your build in %s...", okteto.Context().Builder)
-			}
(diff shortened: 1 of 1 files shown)


jmacelroy (approved):

pkg/cmd/stack/translate.go
```diff
@@ -208,6 +204,17 @@ func addVolumeMountsToBuiltImage(ctx context.Context, s *model.Stack, options *S
 			if okteto.IsOkteto() && !registry.IsOktetoRegistry(svc.Image) {
 				svc.Image = fmt.Sprintf("okteto.dev/%s-%s:okteto-with-volume-mounts", s.Name, name)
 			}
+			if !options.ForceBuild {
+				if _, err := registry.GetImageTagWithDigest(svc.Image); err != errors.ErrNotFound {
```
ifbyol, line 208: I see that if there is an error getting the image tag, but that error is not a NotFound, we are also skipping the build. Should we fail in that case or just build the image to avoid to break the execution for an experience improvement?

pchico83 (approved):

PR #2128, opened 2022-01-12
New registry client

teresaromero:
Fixes #

## Proposed changes

- 
-


pchico83:
@teresaromero this is great stuff!

pkg/registry/client.go
```diff
@@ -14,30 +14,86 @@
...
 	"github.com/okteto/okteto/pkg/log"
+	"github.com/okteto/okteto/pkg/okteto"
 )
 
-// NewRegistryClient creates a new Registry with the given URL and credentials, then Ping()s it
-// before returning it to verify that the registry is available.
-func NewRegistryClient(registryURL, username, password string) (*registry.Registry, error) {
-	transport := http.DefaultTransport
-	return newFromTransport(registryURL, username, password, transport)
+type OktetoRegistryAuthenticator struct {
```
ifbyol, line 25: Minor thing. If possible, It is really helpful (and a good practice) add comments to public exported elements (types, functions, variables, etc...). (Setting `golint` as the lint tool for Go in the VSCode helps you to detect this cases)

pkg/registry/client.go
```diff
@@ -14,30 +14,86 @@
...
+}
+
+func (a *OktetoRegistryAuthenticator) Authorization() (*authn.AuthConfig, error) {
+	return &authn.AuthConfig{
+		Username: a.username,
+		Password: a.password,
+	}, nil
+}
+
+func newBasicAuthRegistry(username, password string) *OktetoRegistryAuthenticator {
```
ifbyol, line 37: Don't you need to export `newBasicAuthRegistry`? If not, maybe `OktetoRegistryAuthenticator` can be local to the package

pkg/registry/client.go
ifbyol: If it is not needed to use our own type, maybe we can use directly the basic authenticator provided by the library: https://github.com/google/go-containerregistry/blob/main/pkg/authn/basic.go. In that way, `newBasicAuthRegistry` would just build a new `authn.Basic`

ifbyol (approved):
Really good refactor!

pkg/registry/client.go
teresaromero: a whole new world has been discovered for me with `golint`😍 - added some minor fixes regarding the registry package after running the linter. 

also, applied your suggestion and removed the custom type.
thnks

ifbyol (approved):

PR #2130, opened 2022-01-12
fix: volume mount translation

### LINKED ISSUE #2119: The user defined in a Dockerfile can't create directories inside of a volume created during stack creation
**Describe the bug**
The user defined in a Dockerfile can't create directories inside of a volume created during stack creation. From a high-level analysis it seems like the volumes we create during deployment are owned by 'root'. While normally this is not an issue, this causes problems if a) the user of the main process is not root AND b) the user tries to create files or folders in said volume. A typical scenario for this is when creating a file for logging.

**To Reproduce**

1. Deploy https://github.com/codeforequity-at/botium-speech-processing
2. See the status of the frontend service. 

**Expected behavior**
The user defined in the dockerfile should be able to read/write inside the volumes we create.

- OS: macOS
- Version 1.4.10

jLopezbarb:

Fixes #2119

## Proposed changes
- Add translation rules for volume mounts in deployments where we need to give permissions to the user/create the path in order to work
- In order to not use persistent volumes and be able to give permission, it's using EmptyDir as volumeSource for this kind of problem

### DIFF
--- a/pkg/cmd/stack/translate.go
+++ b/pkg/cmd/stack/translate.go
@@ -277,7 +277,6 @@ func translateDeployment(svcName string, s *model.Stack) *appsv1.Deployment {
 	svc := s.Services[svcName]
 
 	healthcheckProbe := getSvcProbe(svc)
-
 	return &appsv1.Deployment{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:        svcName,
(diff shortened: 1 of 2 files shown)


pchico83 (approved):
Could we add some unit tests covering the new translation cases?

teresaromero (approved):

PR #2151, opened 2022-01-18
Refactor model package

### LINKED ISSUE #2111: Refactor the `GetManifestV2` logic
The logic of `GetManifestV2` should handle everything about reading manifest. This way, the caller doesn't need to know if it's reading an old manifest, a pipeline or a v2 manifest. The interface should be `GetManifestV2(path string)`, with `path` being optional. 

If `path` isn't specified, the logic should do something similar to the `GetManifest` function.

Also, all these functions shouldbn¡t be in the `context` package. They should be in the `model` package close to the manifestV2 struct definition.

We should also have unit tests for type of files and manifests.

jLopezbarb:

Fixes #2111

## Proposed changes
- Refactor model pakage into small subpackages
-


cmd/context/utils.go
```diff
@@ -218,219 +220,6 @@ func LoadManifestV2WithContext(ctx context.Context, namespace, path string) erro
...
-	}
-
-	if manifestPath != "" {
-		return model.Get(manifestPath)
-	}
-	return nil, errors.ErrManifestNotFound
-}
-
 func IsManifestV2Enabled() bool {
 	r, err := strconv.ParseBool(os.Getenv("OKTETO_ENABLE_MANIFEST_V2"))
```
teresaromero, line 224: maybe we could set this also at the constants file?

cmd/utils/dev.go
```diff
@@ -39,36 +43,37 @@ const (
...
+			if files.FileExists(secondaryManifest) {
 				return LoadManifest(secondaryManifest)
 			}
 		}
 
 		return nil, fmt.Errorf("'%s' does not exist. Generate it by executing 'okteto init'", devPath)
 	}
 
-	manifest, err := model.Get(devPath)
+	manifest, err := manifest.GetManifestV2(devPath)
```
teresaromero, line 71: should this now be called path or manifestPath instead of devPath ?

teresaromero (approved):

jLopezbarb:
I will close this one and open it again splitted into two different PRs

PR #2181, opened 2022-01-21
skip showing deleting namespaces in okteto namespace command

### LINKED ISSUE #2140: okteto namespace shouldn't include deleting namespaces
**Describe the bug**
okteto namespace shouldn't display namespaces in the 'deleting' state (for okteto cloud users), since those are no longer functional.  I think we should filter those out of the list displayed.

RinkiyaKeDad:

Fixes #2140 

## Proposed changes

- filtering namespaces by status before adding them to the list

## Tests

Tested locally by deleting a namespace and running the `okteto namespace` command:

https://user-images.githubusercontent.com/56963264/150497775-e64f05fe-6492-47b7-b6e9-1a74f97d2c25.mov

### DIFF
--- a/cmd/namespace/use.go
+++ b/cmd/namespace/use.go
@@ -127,7 +127,7 @@ func getNamespacesSelection(ctx context.Context) ([]utils.SelectorItem, error) {
 
 	namespaces := []utils.SelectorItem{}
 	for _, space := range spaces {
-		if space.Status == "Active" {
+		if space.Status != "Deleting" {
 			namespaces = append(namespaces, utils.SelectorItem{
 				Name:   space.ID,
 				Label:  space.ID,
--- a/cmd/namespace/use.go
+++ b/cmd/namespace/use.go
@@ -127,12 +127,13 @@ func getNamespacesSelection(ctx context.Context) ([]utils.SelectorItem, error) {
 
 	namespaces := []utils.SelectorItem{}
 	for _, space := range spaces {
-
-		namespaces = append(namespaces, utils.SelectorItem{
-			Name:   space.ID,
-			Label:  space.ID,
-			Enable: true,
-		})
+		if space.Status == "Active" {
+			namespaces = append(namespaces, utils.SelectorItem{
+				Name:   space.ID,
+				Label:  space.ID,
+				Enable: true,
+			})
+		}
 	}
 
 	namespaces = append(namespaces, []utils.SelectorItem{


pchico83 (approved):

jLopezbarb (approved):

cmd/namespace/use.go
```diff
@@ -127,12 +127,13 @@ func getNamespacesSelection(ctx context.Context) ([]utils.SelectorItem, error) {
 
 	namespaces := []utils.SelectorItem{}
 	for _, space := range spaces {
-
-		namespaces = append(namespaces, utils.SelectorItem{
-			Name:   space.ID,
-			Label:  space.ID,
-			Enable: true,
-		})
+		if space.Status == "Active" {
```
ifbyol, line 130: Should we display also `Sleeping` spaces? In those spaces you can deploy things

jpf-okteto (changes requested):
+1 on @ifbyol suggestion

cmd/namespace/use.go
pchico83: One option could be:
```
if space.Status == "Terminating" {
  continue
}

Comment (jpf-okteto) [reply cmd/namespace/use.go:130]:
```
% okteto list namespace
Namespace       Status
foo-jpf-okteto  Deleting
jpf-okteto      Sleeping
```

Isn't it `Deleting`?

Comment (ifbyol) [reply cmd/namespace/use.go:130]: I think the state is `Deleting`: https://github.com/okteto/app/blob/master/backend/pkg/model/space.go#L31

File: cmd/namespace/use.go [hunk 2]
```diff
@@ -127,7 +127,7 @@ func getNamespacesSelection(ctx context.Context) ([]utils.SelectorItem, error) {
 
 	namespaces := []utils.SelectorItem{}
 	for _, space := range spaces {
-		if space.Status == "Active" {
+		if space.Status != "Deleting" {
```
jmacelroy, line 130: What are the possible status here? It seems like this has the potential to begin showing namespaces that weren't show before?

ifbyol: @jmacelroy Potential statuses here are:  `Active`,  `DestroyAllFailed`, `DestroyingAll`,  `Deleting`,  `Sleeping` and `Inactive` (This is deprecated but may happen that some space still returns this status. CLI converts it to `Sleeping`.

RinkiyaKeDad: Should I then change the if to `space.Status == "Active" || "Sleeping"`? Or would the current one (`space.Status != "Deleting"`) suffice?

ifbyol: It depends on what we want. With the current condition, if a new status is added to spaces, those spaces would automatically be listed. With the condition you are proposing, they won't appear. 

I think current condition is ok, because the other statuses (`DestroAllFailed` and `DestroyingAll`) are active spaces where the user can deploy new things, but not sure what status is displayed in the list. Can you try to execute a `Destroy All` operation from the UI and get the list to see how it looks like?

teresaromero: Should we get this status somehow with constants so they keep somewhere listed and refer to them as constants?

RinkiyaKeDad: > Can you try to execute a Destroy All operation from the UI and get the list to see how it looks like?

By "get the list" you mean using the `space.Status != "Deleting"` condition right?

> Should we get this status somehow with constants so they keep somewhere listed and refer to them as constants?

Do you mean defining constants for each of the possible statuses?

ifbyol: @RinkiyaKeDad yes, I meant using that condition to see how that state is reflected in the output

RinkiyaKeDad: @ifbyol I tried that:


https://user-images.githubusercontent.com/56963264/154237098-68e373b6-6f8c-408a-a48a-11917d119cc9.mov



So the namespace gets shown even if we click "Destroy All" in the UI. This is expected right? Since we're deleting only the deployed application resources and not the namespace itself.

ifbyol: Yeah, that is expected. I just wanted to make sure those ns also appear in the list, thanks for testing it!

ifbyol (approved):

cmd/namespace/use.go
pchico83: @RinkiyaKeDad  @rberrelleza @ifbyol I was thinking about a new approach here. It could be needed to point to a `Deleting` namespace, for example, to remove its finalizers.
Instead of ignoring them, why don't we show the namespace status as part of the namespace selector?

ifbyol: I think the Okteto service account doesn't have permissions to patch a namespace, at least I wasn't able

```
kubectl edit ns nacho
error: namespaces "nacho" could not be patched: namespaces "nacho" is forbidden: User "system:serviceaccount:okteto:e9492a1e-e86a-4cf2-bb63-90557e2a4a85" cannot patch resource "namespaces" in API group "" in the namespace "nacho"
```

Although they are still be able to patch any resource within it to remove their finalizers.

It could be an option, but not sure if it is the best experience. If we let a user to select a deleting namespace, we could leave them pointing to a non-valid context if the namespace is finally deleted. Finalizers issue should not be a common scenario and I think we should provide another solution for that at product level instead of leaving the responsibility to the user

pchico83: Let's merge it as it's and we can iterate later

jpf-okteto (approved):

PR #2191, opened 2022-01-24
Jlopezbarb/manifestv2 refactor

### LINKED ISSUE #2111: Refactor the `GetManifestV2` logic

jLopezbarb:
Fixes #2111

## Proposed changes
- Move the logic to read a manifest by a path from the context package to the model package
- Move the logic of inferring the manifest v2 to the model package


pchico83 (approved):

teresaromero (approved):

pkg/model/const.go
```diff
@@ -304,6 +304,9 @@ const (
 	// BuildkitProgressEnvVar defines the output of buildkit
 	BuildkitProgressEnvVar = "BUILDKIT_PROGRESS"
 
+	// OktetoAppNameEnvVar defines the name of the app
+	OktetoAppNameEnvVar = "OKTETO_APP_NAME"
```
pchico83, line 308:
```suggestion
	OktetoNameEnvVar = "OKTETO_NAME"
```

pkg/model/manifest.go
```diff
@@ -14,7 +14,88 @@
...
+//Archetype represents the type of manifest
+type Archetype string
+
+var (
+	//StackType represents a stack manifest type
+	StackType Archetype = "stack"
+	//OktetoType represents a okteto manifest type
+	OktetoType Archetype = "okteto"
+	//PipelineType represents a okteto pipeline manifest type
+	PipelineType Archetype = "pipeline"
```
pchico83, line 38: can we have one for the new manifest?

pkg/model/manifest.go
```diff
@@ -56,8 +144,271 @@ type DeployInfo struct {
...
+//NewDeployInfo creates a deploy Info
 func NewDeployInfo() *DeployInfo {
 	return &DeployInfo{
-		Commands: make([]string, 0),
+		Commands: []string{},
+	}
+}
+
+//GetManifestV2 gets a manifest from a path or search for the files to generate it
+func GetManifestV2(manifestPath string) (*Manifest, error) {
```
pchico83, line 155: if `okteto.yml` is the new manifest, does it take precedence over the rest of manifests?

pchico83 (approved):

pkg/model/manifest.go
jLopezbarb: Added

PR #2215, opened 2022-02-01
Git dependencies from manifest

### LINKED ISSUE #2041: Support for defining git dependencies in the okteto manifest
The new okteto manifest enables the definition of git dependencies to deploy your application. When running okteto deploy, these dependencies are built and deployed as well as the services defined in the build section.

**Additional context**
In the first iteration, this feature will be only available if running in a cluster with the okteto server component (Okteto Cloud or Okteto Enterprise)

### Sample

```
dependencies:
  - https://github.com/okteto/movies-frontend
build:
  API: .
deploy:
  - helm upgrade --install movies chart --set image=${build.api.image} 
dev:
  api:
    command: ["bash"]
    sync:
      - .:/src
    forward:
      - 8080:8080
      - 9229:9229
```

teresaromero:
Fixes #2041 

## Proposed changes

- When `--dependencies` flag is pass to the deploy command, the dependencies section of the manifest is read, and all services defines as dependency are deployed as pipeline. 

https://user-images.githubusercontent.com/40767058/151946774-f75cb5cd-fb87-4542-8666-2a55a784e8f5.mov

- When `okteto destroy` the manifest is checked, and all the pipelines defined at the dependencies section will be destroyed.

https://user-images.githubusercontent.com/40767058/151947542-82bedd3c-633a-4608-92b3-d94e1edd5ff3.mov

For the videos, example used is https://github.com/okteto/movies-api/tree/tere/manifest-v2


pkg/model/dependencies.go
```diff
@@ -0,0 +1,81 @@
...
+
+	var repo, branch, file, variables string
+
+	if dependency.Repository != "" {
+		repo = fmt.Sprintf("-r %s", dependency.Repository)
+		comm = append(comm, repo)
+	}
+
+	if dependency.Branch == "" {
+		dependency.Branch = "main"
```
teresaromero, line 64: Setting "main" as default branch when deploying a pipeline locally because the branch inferred for pipelines of other repositories is the one of the current repository if no branch is defined.

cmd/deploy/deploy.go
```diff
@@ -258,6 +262,16 @@ func (dc *deployCommand) runDeploy(ctx context.Context, cwd string, deployOption
 			}
 		}
 
+		if deployOptions.Dependencies && okteto.IsOkteto() {
```
pchico83, line 265: if `dependencies`is defined in the manifest, and `!okteto.IsOkteto()`, we should fail when reading the manifest

Also, we should deploy the dependencies on `okteto deploy` if they weren't not deployed even if `--dependencies is not specified in the deploy command

pkg/model/dependencies.go
```diff
@@ -0,0 +1,42 @@
+package model
+
+type dependenciesRaw struct {
+	Repository   string      `json:"repository,omitempty" yaml:"repository,omitempty"`
```
pchico83, line 4: should we support `name` and `wait` too?

teresaromero: `name` is being defined when declaring the dependency map at the manifest. Should we add a name to override this map key value?
i will add the wait

pchico83: ah, makes sense

pkg/model/dependencies.go
```diff
@@ -0,0 +1,42 @@
...
+	ManifestPath string      `json:"manifest,omitempty" yaml:"manifest,omitempty"`
+	Branch       string      `json:"branch,omitempty" yaml:"branch,omitempty"`
+	Variables    Environment `json:"variables,omitempty" yaml:"variables,omitempty"`
+}
+
+// ManifestDependencies represents the map of dependencies at a manifest
+type ManifestDependencies map[string]*Dependency
```
pchico83, line 11: @jpf-okteto do we need a pointer here?

pchico83 (approved):

pkg/model/dependencies.go
```diff
@@ -0,0 +1,45 @@
...
+// Dependency represents a dependency object at the manifest
+type Dependency struct {
+	Repository   string      `json:"repository" yaml:"repository"`
+	ManifestPath string      `json:"manifest,omitempty" yaml:"manifest,omitempty"`
+	Branch       string      `json:"branch,omitempty" yaml:"branch,omitempty"`
+	Variables    Environment `json:"variables,omitempty" yaml:"variables,omitempty"`
+	Wait         bool        `json:"wait,omitempty" yaml:"wait,omitempty"`
+}
+
+// UnmarshalYAML Implements the Unmarshaler interface of the yaml pkg.
```
jLopezbarb, line 23: Can we move this to serializer file?

cmd/deploy/deploy.go
```diff
@@ -336,6 +357,50 @@ func setManifestEnvVars(service, reference string) error {
 	return nil
 }
 
+func deployDependency(ctx context.Context, name string, dependency *model.Dependency) error {
```
jLopezbarb, line 360: Can we use `pipeline.ExecuteDeployPipeline(ctx, remoteOpts)` so we don't need to create new code?

jLopezbarb: The only thing that might not be the same is the logs such as `Dependency '%s' scheduled for deployment` but maybe we can create a new option to have those logs

jLopezbarb (approved):

jLopezbarb (approved):

PR #2216, opened 2022-02-01
Hide secrets and deploy vars in output

### LINKED ISSUE #2107: Okteto deploy and okteto destroy should redact okteto secrets/variables
If I have defined some Okteto secrets or variables and I use them with okteto deploy and okteto destroy command they are displayed on the command output

Instead of displaying them, I would like that the output redact somehow those values, so they are not displayed on the terminal. Something similar to what we do in the pipeline or what GitHub actions or circle ci do in their jobs

**Additional context**

jLopezbarb:

Fixes #2107 

## Proposed changes
- Hide secrets and deploy variables to be shown on the output


cmd/context/use.go
```diff
@@ -148,6 +148,7 @@ func setSecrets(secrets []types.Secret) {
 	for _, secret := range secrets {
 		if _, exists := os.LookupEnv(secret.Name); !exists {
 			os.Setenv(secret.Name, secret.Value)
+			oktetoLog.AddBanned(secret.Value)
```
pchico83, line 151: instead of banned, I would use the `obfuscate` verb

pchico83 (approved):

cmd/context/use.go
jpf-okteto: `mask` looks good too.

cmd/deploy/deploy.go
```diff
@@ -298,6 +304,7 @@ func (dc *deployCommand) runDeploy(ctx context.Context, cwd string, deployOption
 		// Set OKTETO_NAMESPACE=namespace-name env variable, so all the commandsruns on the same namespace
 		fmt.Sprintf("%s=%s", model.OktetoNamespaceEnvVar, okteto.Context().Namespace),
 	)
+	oktetoLog.StartRedact()
```
jpf-okteto, line 307: `EnableMask`/`DisableMask`

pkg/log/log.go
```diff
@@ -213,25 +220,68 @@ func Hint(format string, args ...interface{}) {
 
 // Fail prints a message with the error symbol first, and the text in red
 func Fail(format string, args ...interface{}) {
-	log.writer.Fail(format, args...)
+	msg := fmt.Sprintf(format, args...)
+	if log.isRedactMode {
```
jpf-okteto, line 224: `isMasked`

pkg/log/log.go
```diff
@@ -213,25 +220,68 @@ func Hint(format string, args ...interface{}) {
...
+	}
+	log.writer.Print(msg)
 }
 
 //IsInteractive checks if the writer is interactive
 func IsInteractive() bool {
 	return log.writer.IsInteractive()
 }
+
+//AddBanned adds a new
```
jpf-okteto, line 262: space between `//` and the content

pkg/log/log.go
```diff
@@ -213,25 +220,68 @@ func Hint(format string, args ...interface{}) {
...
+func AddBanned(word string) {
+	log.bannedWords = append(log.bannedWords, word)
```
jpf-okteto, line 264: ignore empty strings

jpf-okteto: maybe using `strings.TrimSpace` and checking `len()` later

pkg/log/log.go
```diff
@@ -213,25 +220,68 @@ func Hint(format string, args ...interface{}) {
...
+
+//StopRedact will show secrets and vars
+func StopRedact() {
+	log.isRedactMode = false
+}
+
+func redactMessage(message string) string {
+	s := message
+	for _, str := range log.bannedWords {
+		s = strings.ReplaceAll(s, str, "***")
```
jpf-okteto, line 283: `***` doesn't point which original value was replaced. WDYT about replacing with the variable name such as `$FOO`?

pkg/log/log.go
```diff
@@ -213,25 +220,68 @@ func Hint(format string, args ...interface{}) {
...
+}
+
+//StopRedact will show secrets and vars
+func StopRedact() {
+	log.isRedactMode = false
+}
+
+func redactMessage(message string) string {
+	s := message
+	for _, str := range log.bannedWords {
```
jpf-okteto, line 282: checkout https://pkg.go.dev/strings#Replacer

pkg/log/log_test.go
```diff
@@ -0,0 +1,62 @@
...
+			message:  "hello this is a test",
+			banned:   []string{"is"},
+			expected: "hello th*** *** a test",
+		},
+		{
+			name:     "overlapping banned words to replace",
+			message:  "hello this is a test",
+			banned:   []string{"this", "is"},
+			expected: "hello *** *** a test",
+		},
```
jpf-okteto, line 52: add a case for checking multiline values

pkg/log/log.go
ifbyol: That could be "dangerous" if somehow a expandEnv is executed on the string and the variable is converted to the real value. I would use `*` as it is commonly used in circleci or github actions for example

jLopezbarb: How can they expand the variable if we are just replacing the value with the key and printing it to Stdout without doing any variable expansion. I think that this could be awesome to make it clearer what's being executed. 
Also this change may require some changes on the installer if we finally do it

ifbyol: Not sure if there is any possible way of doing it, but it could happen for example that we include it in the future for some reason without realizing on this. One important thing from a security perspective is to avoid to give clues to the attackers. If you provide information about variable names, you are giving them some environment variables that contain sensitive information

Also, we have to bear in mind that this output will be stored in a configmap.

Anyways, I'm not an expert on security, from a user perspective I think it would help, but in my experience, security sometimes affects usability

PR #2218, opened 2022-02-02
Fix waiting for the display to finish before running other commands

jLopezbarb:

Fixes Deploy was not waiting for the display to finish before starting to run other commands

## Proposed changes
- Wait until the command AND the display has finished before starting another command

### DIFF
--- a/cmd/utils/executor.go
+++ b/cmd/utils/executor.go
@@ -17,6 +17,7 @@ import (
 	"bufio"
 	"os"
 	"os/exec"
+	"sync"
 
 	oktetoLog "github.com/okteto/okteto/pkg/log"
 )
@@ -31,7 +32,7 @@ type Executor struct {
 }
 
 type executorDisplayer interface {
-	display(command string)
+	display(stop chan bool, command string)
 	startCommand(cmd *exec.Cmd) error
 }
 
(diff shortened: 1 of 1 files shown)


cmd/utils/executor.go
```diff
@@ -70,10 +71,11 @@ func (e *Executor) Execute(command string, env []string) error {
 		return err
 	}
 
-	go e.displayer.display(command)
+	displayFinished := make(chan bool)
```
pchico83, line 74: since `scan` is blocking, I think we don't need goroutines here. We can do something like:
```
	if err := e.displayer.startCommand(cmd); err != nil {
		return err
	}
	for e.stdoutScanner.Scan() {
		line := e.stdoutScanner.Text()
		oktetoLog.Println(line)
	}
	return cmd.Wait()
```

jLopezbarb: That's true but we need to wait in jsonDisplayer for both(stdout and stderr) to finish

pchico83: Maybe we can use `select` there. If not, I would use goroutines only in `jsonDisplayer`

jpf-okteto (approved):

pchico83 (approved):

ifbyol (approved):
Tested locally with the same pipeline I was reproducing the issue and it seems fixed

PR #2224, opened 2022-02-02
Support multiple args okteto build

### LINKED ISSUE #2195: Support  `okteto build [service] [service]`
when running `okteto build` command using the manifest, enable the possibility of selecting multiple services to build, instead of only all or one at a time

teresaromero:
Fixes #2195 

## Proposed changes

- refactor the new okteto build with manifest so the args can admit multiple args and build selected services from the manifest


cmd/build.go
```diff
@@ -38,7 +38,7 @@ func Build(ctx context.Context) *cobra.Command {
 	options := build.BuildOptions{}
 	cmd := &cobra.Command{
 		Use:   "build [PATH]",
-		Args:  utils.MaximumNArgsAccepted(1, "https://okteto.com/docs/reference/cli/#build"),
+		Args:  utils.BuildMaximumNArgsAccepted(1, "https://okteto.com/docs/reference/cli/#build"),
```
pchico83, line 41: what's `1` for? it should accept an unlimited number of args no?

pchico83 (approved):

cmd/build.go
teresaromero: I created this func to make it compatible between versions. The new one checks if the env is enabled for manifest v2, so it will allow more than 1, while if is not enabled, it will remain only accepting 1 path

teresaromero: Should we remove the Args restrictions and make this validation at the RunE function?

jLopezbarb (approved):

PR #2226, opened 2022-02-03
Jlopezbarb/add metrics

### LINKED ISSUE #2186: Metrics fixes
Some minor tasks are missing in the CLI to finish the entire onboarding funnel. We can group them here in a single issue:
- [x] Send the "repo" property with URL of the repository as part of the TrackUp event [here](https://github.com/okteto/okteto/blob/master/cmd/up/up.go#L276). 
- [x] Send the "repo" property for "Deploy" event with a `deployType == 'stack' (triggered by `okteto stack deploy`)
- [x] Send the "cluster" property as part of the TrackUp event. There are already `clusterType` and `clusterContext`, but we should use the property `name` to be consistent with the rest of the events.

Closing #2144 in favor of this new issue that groups the minor changes needed to complete the metrics.

jLopezbarb:
Fixes #2186

## Proposed changes
- Track repo on Up and stack deploy events
- Send deployType stack on stack deploy event
- Send cluster name on every event


pkg/analytics/track.go
```diff
@@ -333,6 +340,7 @@ func track(event string, success bool, props map[string]interface{}) {
 	props["success"] = success
 	props["contextType"] = getContextType(okteto.Context().Name)
 	props["context"] = okteto.Context().Name
+	props["cluster"] = okteto.Context().Name
```
pchico83, line 343: isn't this the same as `context`?

cmd/stack/deploy.go
```diff
@@ -73,7 +73,16 @@ func Deploy(ctx context.Context) *cobra.Command {
...
+
+			cwd, err := os.Getwd()
+			if err != nil {
+				oktetoLog.Infof("failed to get the current working directory in up: %v", err)
+			}
+			repo, err := model.GetRepositoryURL(cwd)
+			if err != nil {
+				oktetoLog.Info("could not get repository from current dir")
+			}
+			analytics.TrackDeployStack(err == nil, s.IsCompose, repo)
```
pchico83, line 85: we should lo this for `okteto deploy`, and check there if it's a compose or not. `stack` commands will be deprecated, no?

pkg/analytics/track.go
jLopezbarb: Yes, I was keeping that for retrocompatibility, but we can remove it if we don't need it `context` anymore

cmd/stack/deploy.go
jLopezbarb: Stack will be deprecated but it will still be a hidden command to deploy. So until we remove them and refactor stack to be a command not callable I think it's good to have it. WDYT

pkg/analytics/track.go
pchico83: wouldn't it be better to use `context` prop instead?

cmd/stack/deploy.go
pchico83: agree, but we need to implement this on `okteto deploy` too

pkg/analytics/track.go
jLopezbarb: I think this is due to the fact that the backend is using `cluster` property instead of `context`. We need to have the same props on both so this is why we changed here, but I don't know if it makes more sense to change it in the backend to `context`

cmd/stack/deploy.go
jLopezbarb: In that case, do we have to send an event for every command on the deploy section? `okteto deploy` can deploy more than one type per execution

pchico83: for simplicy, I would go with each subcommand sending its own metrics

pkg/analytics/track.go
rlamana: `cluster` is what was defined in the spec, and what has been implemented in other events in both backend and frontend. These changes try to unify the CLI with the way we do it in other places.

rlamana: It's true that with the new context command it would've been better to use `context`, but for now I think it ok to use "cluster" and have the same property across all services (frontend, backend and cli).

pchico83: @rlamana let's have both then? they seem synonymous to me, and the right concept in the okteto platform is `context`. But having both props doesn't hurt if it simplifies the e2e configuration

rlamana: I'm ok with having both.

cmd/up/up.go
```diff
@@ -274,7 +274,15 @@ func (up *upContext) start() error {
...
-	analytics.TrackUp(true, up.Dev.Name, up.getInteractive(), len(up.Dev.Services) == 0, up.Dev.Divert != nil)
+	cwd, err := os.Getwd()
+	if err != nil {
+		oktetoLog.Infof("failed to get the current working directory in up: %v", err)
+	}
+	repo, err := model.GetRepositoryURL(cwd)
+	if err != nil {
+		oktetoLog.Info("could not get repository from current dir")
+	}
+	analytics.TrackUp(true, up.Dev.Name, repo, up.getInteractive(), len(up.Dev.Services) == 0, up.Dev.Divert != nil)
```
pchico83, line 285: @rlamana let's send if the repo belongs to the okteto organization or not

rlamana: Yeah, I thought we already agreed on that. How are we going to name that property?

pchico83: `isDemoApp`?

rlamana: `useDemoApp`?
`useOktetoApp`?
`useOktetoSampleApp`?

All good to me, as long as it describes clearly what it is for those who will use mixpanel in the future, and they can intuitively interpret what the property means...

jLopezbarb: +1 to `isOktetoApp`. I think it's clear that you are using an okteto app or we can even use `isOktetoRepository` so we know exactly what this property refers to. WDYT @rlamana @pchico83

pchico83: SGTM

rlamana: 👍

pchico83 (approved):

rlamana (approved):

PR #2238, opened 2022-02-07
stderr send warnings instead of errors

### LINKED ISSUE #2236: Stderr causes a command to be set as error
**Describe the bug**
If a command writes into stderr the deploy command will set the command as failed even when these messages are warning messages
Captura de pantalla 2022-02-04 a las 17 10 55

**Expected behavior**
Don't mark the command as failed if the final command error status is different to 0

jLopezbarb:
Fixes #2236 

## Proposed changes
- Differentiate on plain execution between messages that comes from stderr and stdout
- Send stderr as warning messages

### DIFF
--- a/cmd/utils/executor.go
+++ b/cmd/utils/executor.go
@@ -38,7 +38,8 @@ type executorDisplayer interface {
 }
 
 type plainExecutorDisplayer struct {
-	scanner *bufio.Scanner
+	stdoutScanner *bufio.Scanner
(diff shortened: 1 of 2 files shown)


pchico83 (approved):

pchico83:
@jLopezbarb should we make a 1.15.4 release with this commit? that way we can integrate it with the rest of the platform before all the 2.0 changes

PR #2241, opened 2022-02-08
Deploy build parallel

### LINKED ISSUE #2174: Check images in the build section in parallel
**Describe the bug**
`okteto deploy` checks if the images defined in the `build` section are already built. This check is done sequentially for each image, introducing a delay at the beginning of the command execution:
https://github.com/okteto/okteto/blob/64075c69c63336f876a4fa42d96eb4ef08efb841/cmd/deploy/deploy.go#L223

We should optimize this code and do all these checks in parallel.

In order to to that, we can use wait groups, channels and goroutines.

teresaromero:
Fixes #2174 

## Proposed changes

- When checking if an image has to be build when `okteto deploy`, this checks are done with a go-routine.
- if images has to be build, then they will be build sequentially  after.

### DIFF
--- a/cmd/deploy/deploy.go
+++ b/cmd/deploy/deploy.go
@@ -250,18 +250,15 @@ func (dc *DeployCommand) RunDeploy(ctx context.Context, deployOptions *Options) 
 			close(toBuild)
 
 			for {
-				select {
-				case svc := <-toBuild:
-					if svc == "" {
-						break
-					}
-					mBuildInfo := deployOptions.Manifest.Build[svc]
-					if err := runBuildAndSetEnvs(ctx, svc, mBuildInfo); err != nil {
-						return err
-					}
-					continue
+				svc, ok := <-toBuild
+				if svc == "" || !ok {
+					fmt.Println("vacio", ok)
+					break
 				}
-				break
+				mBuildInfo := deployOptions.Manifest.Build[svc]
+				if err := runBuildAndSetEnvs(ctx, svc, mBuildInfo); err != nil {
+					return err
+				}
 			}
 
 		}
--- a/cmd/deploy/deploy.go
+++ b/cmd/deploy/deploy.go
@@ -224,46 +224,28 @@ func (dc *DeployCommand) RunDeploy(ctx context.Context, deployOptions *Options) 
 
 		oktetoLog.Information("Checking to build services at manifest")
 		if deployOptions.Manifest.Build != nil {
-
+			exit := make(chan error, 1)
 			for service, mOptions := range deployOptions.Manifest.Build {
-				buildOptions := build.BuildOptions{OutputMode: oktetoLog.GetOutputFormat()}
-				if mOptions.Name == "" {
-					mOptions.Name = deployOptions.Name
-				}
-				opts := build.OptsFromManifest(service, mOptions, buildOptions)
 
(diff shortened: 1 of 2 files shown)


cmd/deploy/deploy.go
```diff
@@ -212,48 +213,42 @@ func (dc *DeployCommand) RunDeploy(ctx context.Context, deployOptions *Options)
...
-				} else if err != nil {
-					return fmt.Errorf("error checking image at registry %s: %v", opts.Tag, err)
-				} else {
-					oktetoLog.Success("Skipping build for image for service %s found: %s", service, opts.Tag)
-					if err := setManifestEnvVars(service, imageWithDigest); err != nil {
-						return err
-					}
+			for {
+				svc, ok := <-toBuild
+				if svc == "" || !ok {
```
ifbyol, line 243: Shouldn't we do the break only if there are no more elements (`!ok`)? Could it happen that some serviceToBuild is sent as empty `svc == ""` but there are still some more elements in the channel? In that case, maybe we should do a continue if `svc == "" && ok`

cmd/deploy/deploy.go
```diff
@@ -212,48 +213,42 @@ func (dc *DeployCommand) RunDeploy(ctx context.Context, deployOptions *Options)
...
-					continue
-				} else if err != nil {
-					return fmt.Errorf("error checking image at registry %s: %v", opts.Tag, err)
-				} else {
-					oktetoLog.Success("Skipping build for image for service %s found: %s", service, opts.Tag)
-					if err := setManifestEnvVars(service, imageWithDigest); err != nil {
-						return err
-					}
+			for {
+				svc, ok := <-toBuild
```
ifbyol, line 242: I think you can iterate directly over the channel as you don't need a `select` statement and you are closing the channel. I think you could do:

```
for svc := range toBuild {
}
```

cmd/deploy/deploy.go
```diff
@@ -212,48 +213,42 @@ func (dc *DeployCommand) RunDeploy(ctx context.Context, deployOptions *Options)
...
 
 		oktetoLog.Information("Checking to build services at manifest")
-		if deployOptions.Manifest.Build != nil {
-
-			for service, mOptions := range deployOptions.Manifest.Build {
-				buildOptions := build.BuildOptions{OutputMode: oktetoLog.GetOutputFormat()}
-				if mOptions.Name == "" {
-					mOptions.Name = deployOptions.Name
+		if deployOptions.Build {
+			// force build of all services
```
jLopezbarb, line 217: Could we encapsulate this into a function?

cmd/deploy/deploy.go
teresaromero: it should break if `!ok` so the `srv == ""` condition is no longer applicable, as this will be only when the channel is closed and is not returning anything. the toBuild channel only returns when the image has to be build so i cant think of anycase where it returns an empty string within the channel being open. thnks!

cmd/deploy/deploy.go
teresaromero: i saw the first comment first... haha i think this is a better approach, so i will change the code for this option

cmd/deploy/deploy.go
teresaromero: yes we can!

PR #2244, opened 2022-02-09
Remove manifest_v2 flag

teresaromero:
## Proposed changes

- remove references to MANIFEST_V2 flag
- running buildv1 or buildv2 depending on manifest version and user input

### DIFF
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -25,6 +25,7 @@ import (
 	"github.com/okteto/okteto/cmd/utils"
 	"github.com/okteto/okteto/pkg/analytics"
 	"github.com/okteto/okteto/pkg/cmd/build"
+	oktetoErrors "github.com/okteto/okteto/pkg/errors"
 	oktetoLog "github.com/okteto/okteto/pkg/log"
 	"github.com/okteto/okteto/pkg/model"
 	"github.com/okteto/okteto/pkg/okteto"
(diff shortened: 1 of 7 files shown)


cmd/deploy/deploy.go
```diff
@@ -214,53 +214,48 @@ func (dc *DeployCommand) RunDeploy(ctx context.Context, deployOptions *Options)
...
-		if deployOptions.Manifest.Build != nil {
-
-			for service, mOptions := range deployOptions.Manifest.Build {
-				buildOptions := build.BuildOptions{OutputMode: oktetoLog.GetOutputFormat()}
-				if mOptions.Name == "" {
-					mOptions.Name = deployOptions.Name
-				}
-				opts := build.OptsFromManifest(service, mOptions, buildOptions)
+	if deployOptions.Dependencies && !okteto.IsOkteto() {
+		return fmt.Errorf("deploy of dependencies is only available for Okteto instances")
```
pchico83, line 218:
```suggestion
		return fmt.Errorf("deploy of dependencies is only available in clusters managed by Okteto")
```

pchico83 (approved):

cmd/build.go
```diff
@@ -38,15 +39,48 @@ func Build(ctx context.Context) *cobra.Command {
 	options := build.BuildOptions{}
 	cmd := &cobra.Command{
 		Use:   "build [PATH]",
-		Args:  utils.BuildMaximumNArgsAccepted(1, "https://okteto.com/docs/reference/cli/#build"),
 		Short: "Build (and optionally push) a Docker image",
 		RunE: func(cmd *cobra.Command, args []string) error {
 
-			if err := contextCMD.NewContextCommand().Run(ctx, &contextCMD.ContextOptions{}); err != nil {
```
jLopezbarb, line 45: I think we need to run this to setup secrets and other information

jLopezbarb: Nevermid I just saw its removing duplicated code

jLopezbarb (approved):

cmd/build.go
teresaromero: secrets is not supported as build param at the manifest, the secrets are allowed as flag if single build is done.

PR #2251, opened 2022-02-11
fix err with manifest v2 build

teresaromero:

Fixes #

## Proposed changes

- when file is not a manifest v2 and err is returned, build with v1

### DIFF
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -44,12 +44,13 @@ func Build(ctx context.Context) *cobra.Command {
 
 			ctxOpts := &contextCMD.ContextOptions{}
 
-			manifest, err := model.GetManifestV2(options.File)
-			if err != nil {
-				return err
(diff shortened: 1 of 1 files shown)


jLopezbarb (approved):

cmd/build.go
```diff
@@ -44,12 +44,13 @@ func Build(ctx context.Context) *cobra.Command {
 
 			ctxOpts := &contextCMD.ContextOptions{}
 
-			manifest, err := model.GetManifestV2(options.File)
-			if err != nil {
-				return err
+			manifest, errManifest := model.GetManifestV2(options.File)
+			if errManifest != nil {
+				oktetoLog.Debug("error getting manifest v2 from file %s: %v", options.File, errManifest)
```
jLopezbarb, line 49:
```suggestion
				oktetoLog.Debug("error getting manifest v2 from file %s: %v. Fallback to build v1", options.File, errManifest)
```

PR #2262, opened 2022-02-16
Add CTA after deploy

### LINKED ISSUE #2202: CTA to `okteto up xxx` after `okteto deploy` is executed
If running in `tty` mode, `okteto deploy` should show a CTA to run `okteto up xxx` where xxx is the name of any deployed deployment/statefulset by `okteto deploy`.

`okteto up xxx` should start the `init` sequence if `xxx` is not defined as a dev container

jLopezbarb:

Fixes #2202

## Proposed changes
- If deploy has not deployed anything we show the user an error telling him to update deploy section
- If it has deployed something(any job/sfs/deployment) it will congratulate the user and suggest to run okteto up next

### DIFF
--- a/cmd/deploy/deploy.go
+++ b/cmd/deploy/deploy.go
@@ -350,10 +350,7 @@ func (dc *DeployCommand) RunDeploy(ctx context.Context, deployOptions *Options) 
 			oktetoLog.Success("Great work deploying your application.\n    Try running `okteto up` next")
 		}
 	} else {
-		return oktetoErrors.UserError{
-			E:    errors.New("It seems that you haven't deployed anything"),
-			Hint: "Please try updating your deploy section on your manifest and try again",
-		}
+		return oktetoErrors.ErrDeployHasNotDeployAnyResource
 	}
 
 	return nil
--- a/cmd/deploy/deploy.go
+++ b/cmd/deploy/deploy.go
@@ -155,7 +155,7 @@ func Deploy(ctx context.Context) *cobra.Command {
 			if options.Name == "" {
 				options.Name = utils.InferApplicationName(cwd)
 			}
-			options.ShowCTA = true
+			options.ShowCTA = oktetoLog.IsInteractive()
 
 			kubeconfig := NewKubeConfig()
 
--- a/cmd/deploy/deploy.go
+++ b/cmd/deploy/deploy.go
@@ -53,7 +53,10 @@ import (
 	"k8s.io/client-go/rest"
 )
 
-const headerUpgrade = "Upgrade"
+const (
+	headerUpgrade          = "Upgrade"
+	succesfullyDeployedmsg = "'%s' successfully deployed."
+)
 
 var tempKubeConfigTemplate = "%s/.okteto/kubeconfig-%s"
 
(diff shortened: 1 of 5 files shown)


pkg/cmd/pipeline/crud.go
```diff
@@ -29,3 +34,33 @@ func IsDeployed(ctx context.Context, name, namespace string, c kubernetes.Interf
...
+
+	sfsList, err := statefulsets.List(ctx, ns, labels, c)
+	if err != nil {
+		return false
+	}
+	if len(sfsList) > 0 {
+		return true
+	}
+
+	jobsList, err := jobs.List(ctx, ns, labels, c)
```
pchico83, line 57: since we don't support developing on jobs, I wouldn't consider them here

pkg/cmd/pipeline/crud.go
```diff
@@ -29,3 +34,33 @@ func IsDeployed(ctx context.Context, name, namespace string, c kubernetes.Interf
 	}
 	return cmap.Data[statusField] != ErrorStatus
 }
+
+// HasDeployedSomething checks if the pipeline has deployed any deployment/statefulset/job
+func HasDeployedSomething(ctx context.Context, name, ns string, c kubernetes.Interface) bool {
+	labels := fmt.Sprintf("%s=%s",model.DeployedByLabel, name)
+	dList, err := deployments.List(ctx, ns, labels, c)
+	if err != nil {
+		return false
```
pchico83, line 43: in this case, we should populate the internal server error to the user

pkg/errors/errors.go
```diff
@@ -134,6 +134,12 @@ var (
 
 	// ErrManifestFoundButNoDeployCommands raised when a manifest is found but no deploy commands are defined
 	ErrManifestFoundButNoDeployCommands = errors.New("found okteto manifest, but no deploy commands where defined")
+
+	// ErrDeployHasNotDeployAnyResource raised when a deploy command has not created any resource
+	ErrDeployHasNotDeployAnyResource = UserError{
+		E:    errors.New("It seems that you haven't deployed anything"),
```
pchico83, line 140: we should have two versions of this error, one for deployment errors, another one if you didn't deploy anything. But the hint should be the same.

pchico83 (approved):

pkg/errors/errors.go
jLopezbarb: I think that you can have a deployment in an error status and debug it by running `okteto up`. Does this makes sense?

pchico83: Note that it's not the same that the deploy command fails, than the container is failing later after deployment.
if the deploy fails I think it's fine to ask to fix the deploy commands

pkg/cmd/pipeline/crud.go
```diff
@@ -29,3 +33,25 @@ func IsDeployed(ctx context.Context, name, namespace string, c kubernetes.Interf
 	}
 	return cmap.Data[statusField] != ErrorStatus
 }
+
+// HasDeployedSomething checks if the pipeline has deployed any deployment/statefulset/job
```
ifbyol, line 37:
```suggestion
// HasDeployedSomething checks if the pipeline has deployed any deployment/statefulset
```

you are not checking jobs, not sure if we should or not

cmd/deploy/deploy.go
```diff
@@ -320,30 +326,48 @@ func (dc *DeployCommand) RunDeploy(ctx context.Context, deployOptions *Options)
...
+				}
+			}
+			data.Status = pipeline.DeployedStatus
+		} else {
+			err = oktetoErrors.UserError{
+				E:    oktetoErrors.ErrDeployHasNotDeployAnyResource,
+				Hint: "Please try updating your deploy section on your manifest and try again",
+			}
+			oktetoLog.AddToBuffer(oktetoLog.InfoLevel, err.Error())
+			data.Status = pipeline.ErrorStatus
```
ifbyol, line 363: I think It is odd that we mark as error the deploy if nothing was deployed, specially in the UI. The deploy would appear as Error but the logs wouldn't show any error (as we don't add an stage here, this log wouldn't appear in the UI with the structured logs).

I don't know if there would make sense any scenario in which you execute a deploy without deploying anything, but depending on how other dependencies are deployed, there could be some scenarios.

If each dependency is deployed as an independent pipeline and I'm using my manifest just to deploy all the dependencies but it doesn't deploy anything by itself, I think this would happen

ifbyol (approved):
Apart from the comment about setting the error state when nothing is deployed, the rest looks good and we can always revisit that part

pkg/cmd/pipeline/crud.go
jLopezbarb: We shouldn't because we don't support developing on jobs

ifbyol: then I think we are coupling the logic of showing the CTA with the pipeline status. If this function returns a false, could be that something was deployed (a cronjob, a job, a daemonset or something else) successfully, but we are returning false, so we are marking the deploy as failed (check my previous comment)

PR #2268, opened 2022-02-17
Add deprecation message to okteto push command

### LINKED ISSUE #2165: Deprecate `okteto push`
In order to simplify the okteto dev experience, we should have a single way to build and redeploy an application. This can be done now with `okteto deploy --build`. More information [here](https://github.com/okteto/okteto/issues/2157).

jLopezbarb:

Fixes #2165 

## Proposed changes
- Add deprecation message if its not inside a `okteto deploy` (we would show the warning even when the user only has an okteto yaml)
- Hide push from the list of available commands


cmd/push.go
```diff
@@ -52,11 +52,14 @@ type pushOptions struct {
 func Push(ctx context.Context) *cobra.Command {
 	pushOpts := &pushOptions{}
 	cmd := &cobra.Command{
-		Use:   "push [svc]",
-		Short: "Build, push and redeploy source code to the target app",
-		Args:  utils.MaximumNArgsAccepted(1, "https://okteto.com/docs/reference/cli/#push"),
+		Hidden: true,
+		Use:    "push [svc]",
```
pchico83, line 56: does `okteto push` without arguments still work if using an old manifest?

pchico83 (approved):

cmd/push.go
jLopezbarb: yes, It will get the old manifest and use that as the dev manifest

PR #2270, opened 2022-02-17
Check "skip-if-exists" by accessing the configmap

pchico83:
## Proposed changes
Instead of checking if a pipeline exists by calling the okteto API, we now check the existence of the configmap.
Accessing the Okteto API had issues when running on preview environments, because the `GetNamespace` request was returning 404


teresaromero (approved):
Looks good to me

ifbyol (approved):

PR #2284, opened 2022-02-24
Persist Bash History Between Iterations of a Dev Container

teresaromero:

Fixes https://github.com/okteto/app/issues/3231

## Proposed changes

Set ENVS to control the bash history inside the dev container. This solution creates the bashrc file at local repo and syncs this with the one at the dev container.

- HISTSIZE and HISTFILESIZE - set as env to make history file able to save more lines
- HISTCONTROL set to ignoreboth:erasedups - do not save the duplicated commands 
- HISTFILE - bashrc ; save history on this file which is sync with the app being developed
- BASHOPTS - histppend - the history is append to the file and not overwritten
- PROMPT_COMMAND - "history -a ; history -c ; history -r ; $PROMPT_COMMAND" - file is loaded to history at init and reloaded & saved with each command

### DIFF
--- a/pkg/model/dev.go
+++ b/pkg/model/dev.go
@@ -548,6 +548,32 @@ func (dev *Dev) SetDefaults() error {
 		}
 	}
 
+	dev.Environment = append(dev.Environment,
+		EnvVar{
+			Name:  "HISTSIZE",
+			Value: "10000000",
+		},
+		EnvVar{
+			Name:  "HISTFILESIZE",
+			Value: "10000000",
+		},
+		EnvVar{
+			Name:  "HISTCONTROL",
+			Value: "ignoreboth:erasedups",
+		},
+		EnvVar{
+			Name:  "HISTFILE",
+			Value: "bashrc",
+		},
+		EnvVar{
+			Name:  "BASHOPTS",
+			Value: "histappend",
+		},
+		EnvVar{
+			Name:  "PROMPT_COMMAND",
+			Value: "history -a ; history -c ; history -r ; $PROMPT_COMMAND",
+		})
+
 	return nil
 }
 


ifbyol (approved):
I'm not very familiar with bash history but this looks a good solution

jmacelroy (commented):
I checked this branch out and tried it out and I hit an issue. 

```
jmacelroy:api app> history
    1  history
bash: history: /var/okteto/bashrc/.bash_history: cannot create: No such file or directory
jmacelroy:api app>
```
If I create that path it work only for that up session and does not work after I run `okteto down`

rberrelleza:
What would happen if the continer doesn't have `bash` but instead it has `sh`or other shells?

pkg/k8s/apps/translate.go
```diff
@@ -337,6 +337,32 @@ func TranslateEnvVars(c *apiv1.Container, rule *model.TranslationRule) {
 			c.Env = append(c.Env, apiv1.EnvVar{Name: envvar.Name, Value: value})
 		}
 	}
+
+	c.Env = append(c.Env,
```
rberrelleza, line 341: would it make more sense to execute this on the SSH session instead? that way a) we have more information on the available shell sessions and b) we don't have to "expose" this implementation detail in the kubernetes manifest?

teresaromero:
> I checked this branch out and tried it out and I hit an issue.
> 
> ```
> jmacelroy:api app> history
>     1  history
> bash: history: /var/okteto/bashrc/.bash_history: cannot create: No such file or directory
> jmacelroy:api app>
> ```
> 
> If I create that path it work only for that up session and does not work after I run `okteto down`

@jmacelroy  this is wierd... i've tested again and i haven't seen this. maybe someone else can test this? 
might be that the shell is different than bash? 

> would it make more sense to execute this on the SSH session instead? that way a) we have more information on the available shell sessions and b) we don't have to "expose" this implementation detail in the kubernetes manifest?

@rberrelleza i haven't take that into account, wrongly assuming the shell was bash in all cases. i have little knowledge regarding the ssh session, but i am assuming then that the "saved" history for the sessions would have to be locally saved instead on a persistent volume? the point of using the manifest was to set the volume to store the commands to be restored while the volume is alive. 

i need some light here on "how" and "cases" to be sure the implementation covers the expectations

rberrelleza:
@teresaromero scratch my comments, I think that we have a dependency on bash in `okteto up` and `okteto exec` already. @pchico83 and @jLopezbarb can confirm.

I still think that we should inject that env var during the command, or directly in the SSH session instead of modifying the yaml, but that's mostly to prevent leaking an implementation detail. You should be able to set it at the SSH session [in here](https://github.com/okteto/remote/blob/main/pkg/ssh/ssh.go#L185), by using setting the value of `s.Env`  but I'm not 100% sure

jmacelroy:
Here's a video of my quick test https://www.loom.com/share/79ef1d001350412a852f5201d3e28041

ifbyol:
@teresaromero I think this is happening because movies example doesn't have the persistent volume enabled and your changes are relying on the persistent volumen: https://github.com/okteto/movies/blob/main/api/okteto.yml#L9.

Maybe we should include the file only if the persistent volume is enabled. Or we can find another way to do it

teresaromero:
> @teresaromero I think this is happening because movies example doesn't have the persistent volume enabled and your changes are relying on the persistent volume: https://github.com/okteto/movies/blob/main/api/okteto.yml#L9.
> 
> Maybe we should include the file only if the persistent volume is enabled. Or we can find another way of doing it

@ifbyol  You are right, the envs were being set despite persistent volume was enabled or not. I refactored the solution so the env and the volume is set only when persistent volume is enabled. this means the only way to save the commands and share them between devs is having this option enabled, which is by default true.

@jmacelroy let me know if this works for you now

pchico83 (approved):

teresaromero:
Opening docs issue to document this change https://github.com/okteto/docs/issues/30

PR #2299, opened 2022-03-01
fix: stop other okteto for the same app

### LINKED ISSUE #2294: Okteto CLI fails to 'okteto up' when development is active and port forwarding is enabled
**Describe the bug**
When trying to perform and `okteto up` for a service we're getting the error: 
`Couldn't connect to your development container: local port 2351 is already in-use in your local machine`

When attempting to replicate the issue it appears that when we define port forwarding and already have an `okteto up` running, issuing an `okteto up` for the same service will cause this error.  

**To Reproduce**

1. Define a service manifest with port forwarding
2. Run `okteto up` in one terminal
3. Run `okteto up` in another terminal
4. See error `Couldn't connect to your development container: local port 2351 is already in-use in your local machine`

**Expected behavior**
Normally, using `okteto up` will terminal the other session, but in this instance it usually doesn't. 

- OS: MacOS
- Version: 1.15.3

**Additional context**
None

jLopezbarb:

Fixes #2294

## Proposed changes
- Creates an uuid for every up execution 
- Inject that uuid as an annotation to dev deployment/statefulset
- If it's a retry and the uuid of the app has changed stop the application
- Forward will wait for the other okteto execution to stop to get the port
- If another program (not okteto) has taken the port. It will throw an error telling the user to free the port up

### DIFF
--- a/cmd/up/forwards.go
+++ b/cmd/up/forwards.go
@@ -16,8 +16,10 @@ package up
 import (
 	"context"
 	"fmt"
+	"time"
 
 	"github.com/okteto/okteto/cmd/utils"
+	oktetoErrors "github.com/okteto/okteto/pkg/errors"
 	"github.com/okteto/okteto/pkg/k8s/forward"
 	oktetoLog "github.com/okteto/okteto/pkg/log"
 	"github.com/okteto/okteto/pkg/model"
@@ -89,6 +91,12 @@ func (up *upContext) sshForwards(ctx context.Context) error {
 			up.Dev.Forward[idx] = forwardWithServiceName
 			f = forwardWithServiceName
 		}
+
+		err := waitUntilPortIsAvailable(up.Dev.Interface, f.Local, up.Dev.Timeout.Default)
+		if err != nil {
+			return err
+		}
+
 		if err := up.Forwarder.Add(f); err != nil {
 			return err
 		}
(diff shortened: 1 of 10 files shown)


cmd/up/forwards.go
```diff
@@ -107,3 +115,22 @@ func (up *upContext) sshForwards(ctx context.Context) error {
 
 	return up.Forwarder.Start(up.Pod.Name, up.Dev.Namespace)
 }
+
+func waitUntilPortIsAvailable(iface string, port int, timeout time.Duration) error {
+	ticker := time.NewTicker(5 * time.Second)
```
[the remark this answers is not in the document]
pchico83: check every 1s for 30s

cmd/up/up.go
```diff
@@ -364,7 +367,10 @@ func (up *upContext) activateLoop() {
 				<-t.C
 			}
 		}
-
+		if up.isRetry && up.hasDevIDChanged(ctx) {
+			up.Exit <- fmt.Errorf("could not connect to dev: another okteto process is modifying it")
```
pchico83, line 371:
```suggestion
			up.Exit <- fmt.Errorf("session disconnected: there is another `okteto up` session on this container")
```
Could we also show a hint to run `okteto exec`?

pkg/model/const.go
```diff
@@ -34,6 +34,9 @@ const (
 	// OktetoSampleAnnotation indicates that the repo is a okteto sample
 	OktetoSampleAnnotation = "dev.okteto.com/sample"
 
+	// OktetoDevIDAnnotation indicates that the dev uid
+	OktetoDevIDAnnotation = "dev.okteto.com/dev-id"
```
pchico83, line 38:
```suggestion
	OktetoSessionIDAnnotation = "dev.okteto.com/session-id"
```

pchico83 (approved):

pchico83 (approved):

PR #2303, opened 2022-03-02
fix: okteto down cause nil pointer exception

jLopezbarb:

Fixes okteto down was causing a nil pointer exception on retrying autocreate devs

## Proposed changes
- If we can not find any app we should retry activate loop
-


ifbyol (approved):

cmd/up/up.go
```diff
@@ -401,7 +401,7 @@ func (up *upContext) activateLoop() {
 
 func (up *upContext) hasDevIDChanged(ctx context.Context) bool {
```
pchico83, line 402: should it be `hasSessionIDChanged` instead of `hasDevIDChanged`?

pchico83 (approved):

PR #2305, opened 2022-03-02
fix compose build integration

jLopezbarb:

Fixes: It was rebuilding the image on the deploy if there was a stack

## Proposed changes
- If run okteto build with a compose it will run and include the host volumes on the built image
- If run okteto deploy it will build the images at the beginning and not while deploying stack

### DIFF
--- a/pkg/model/const.go
+++ b/pkg/model/const.go
@@ -317,4 +317,7 @@ const (
 
 	// OktetoDefaultImageTag default tag assigned to image to build
 	OktetoDefaultImageTag = "okteto"
+
+	// OktetoImageTagWithVolumes is the tag assigned to an image with volume mounts
+	OktetoImageTagWithVolumes = "okteto-with-volume-mounts"
 )
--- a/pkg/model/manifest.go
+++ b/pkg/model/manifest.go
@@ -545,10 +545,7 @@ func (m *Manifest) InferFromStack() (*Manifest, error) {
 		}
 		for _, v := range svcInfo.VolumeMounts {
 			if pathExistsAndDir(v.LocalPath) {
-				d.Sync.Folders = append(d.Sync.Folders, SyncFolder{
-					LocalPath:  v.LocalPath,
-					RemotePath: v.RemotePath,
-				})
+				d.Sync.Folders = append(d.Sync.Folders, SyncFolder(v))
 			}
 			toMount = append(toMount, v)
 		}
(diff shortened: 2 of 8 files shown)


pkg/model/const.go
```diff
@@ -317,4 +317,7 @@ const (
 
 	// OktetoDefaultImageTag default tag assigned to image to build
 	OktetoDefaultImageTag = "okteto"
+
+	// OktetoImageTagWithVolumes is the tag assigned to an image with volume mounts
+	OktetoImageTagWithVolumes = "okteto-with-volume-mounts"
```
pchico83, line 322: can we build the same tag in both cases? I think it would simplify the code

pkg/model/manifest.go
```diff
@@ -236,15 +236,15 @@ func GetManifestV2(manifestPath string) (*Manifest, error) {
...
-			inferredManifest.mergeWithOktetoManifest(devManifest)
-		} else if inferredManifest.Type == StackType {
+		if inferredManifest.Type == StackType {
 			inferredManifest, err = inferredManifest.InferFromStack()
 			if err != nil {
 				return nil, err
 			}
-			//TODO: uncomment
-			// inferredManifest.Deploy.Compose.Stack.Endpoints = devManifest.Deploy.Endpoints
+			inferredManifest.Deploy.Compose.Stack.Endpoints = devManifest.Deploy.Endpoints
```
pchico83, line 244: I am getting a segmentation fault error here when running `okteto build` in this repo:
https://github.com/pchico83/compose-getting-started

pchico83 (commented):
If I run `okteto build` and `okteto deploy` on this repo:
https://github.com/pchico83/compose-getting-started

The image for `vote` is registry.cloud.okteto.net/pchico83/compose-getting-started-vote:okteto-with-volume-mounts, but it should be the tag based on the `sha256:56df26d0b78b786c248d633251ca` to make sure the containers are redeployed if the image has changed.

pkg/model/const.go
jLopezbarb: I don't think it's a good idea because of this use case:
1.  User run `okteto deploy` for the first time 
2. It builds the first image with the tag `okteto`
3. It fails on including the volumes
4. User runs again `okteto deploy` and it won't be built because the tag exists (right now we are checking if the tag with the volumes exists)

pchico83: Makes sense

pkg/model/manifest.go
jLopezbarb: I can't repro it, are you using any flag on the build command?

pchico83: Nope, just `okteto build` or `okteto deploy`. Let me try again

pchico83 (approved):

PR #2310, opened 2022-03-03
fix: stack deploy is showing endpoints when fail

jLopezbarb:

Fixes Stack deploy was showing endpoints when there was a failure on  the command

## Proposed changes
- Check if the stack deploy has an error


cmd/stack/deploy.go
```diff
@@ -116,6 +116,8 @@ func (c *DeployCommand) RunDeploy(ctx context.Context, s *model.Stack, options *
 	analytics.TrackDeployStack(err == nil, s.IsCompose, utils.IsOktetoRepo())
 	if err == nil {
```
teresaromero, line 117:
```suggestion
	if err != nil {
	return err
	}
	oktetoLog.Success("Compose '%s' successfully deployed", s.Name)
```

teresaromero (approved):

PR #2314, opened 2022-03-04
fix: up selector move correctly

### LINKED ISSUE #2291: Development container selector is not working properly
**Describe the bug**
When I execute `okteto up` or `okteto down` and the selector is being displayed, there are 2 issues:
* Each time I execute the command I get a different order
* There are options I cannot select, it seems the selector progress 2 items at a time, and at some point I cannot move the cursor anymore ( It seems that I cannot go to previous options)

**To Reproduce**

1. Create a manifest with several dev containers defined
2. Execute `okteto up` or `okteto down`

**Expected behavior**
Each time I execute the command I get the same order and the cursor moves one item at a time

- OS: macOS

**Additional context**

https://user-images.githubusercontent.com/3510171/155585661-a40bca5f-0eb9-4d61-9521-f50fa084333f.mov

jLopezbarb:

Fixes #2291

## Proposed changes
- Order by length and alphabetically so the order keeps always the same
- Fix move between options

### DIFF
--- a/cmd/utils/dev.go
+++ b/cmd/utils/dev.go
@@ -20,6 +20,7 @@ import (
 	"os"
 	"path/filepath"
 	"runtime"
+	"sort"
 	"strings"
 
 	"github.com/manifoldco/promptui"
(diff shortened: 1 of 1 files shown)


ifbyol (approved):
Tested the branch with the repo I was having the error and It works perfectly!!

PR #2319, opened 2022-03-06
Metrics Fixes

maroshii:
Send the deploy origin in the deploy mutation. Adding a special source: `okteto-deploy` used for deploy commands ran from within another okteto deploy

### DIFF
--- a/cmd/deploy/deploy.go
+++ b/cmd/deploy/deploy.go
@@ -314,7 +314,6 @@ func (dc *DeployCommand) RunDeploy(ctx context.Context, deployOptions *Options) 
 		fmt.Sprintf("%s=%s", model.OktetoNamespaceEnvVar, okteto.Context().Namespace),
 	)
 	oktetoLog.EnableMasking()
-
 	err = dc.deploy(ctx, deployOptions)
 	oktetoLog.DisableMasking()
 
--- a/cmd/deploy/deploy.go
+++ b/cmd/deploy/deploy.go
@@ -179,8 +179,25 @@ func Deploy(ctx context.Context) *cobra.Command {
 			}
 			startTime := time.Now()
 			err = c.RunDeploy(ctx, options)
-			duration := time.Since(startTime)
-			analytics.TrackDeploy(err == nil, utils.IsOktetoRepo(), err, duration, c.PipelineType)
+
+			deployType := "custom"
+			if options.Manifest.IsV2 &&
+				options.Manifest.Deploy != nil &&
+				options.Manifest.Deploy.Compose != nil &&
+				options.Manifest.Deploy.Compose.Manifest != nil {
+				deployType = "compose"
+			}
+			analytics.TrackDeploy(analytics.TrackDeployMetadata{
+				Success:                err == nil,
+				IsOktetoRepo:           utils.IsOktetoRepo(),
+				Duration:               time.Since(startTime),
+				PipelineType:           c.PipelineType,
+				DeployType:             deployType,
+				IsPreview:              false, // TODO: How should we get this value?
+				HasDependenciesSection: options.Manifest.IsV2 && len(options.Manifest.Dependencies) > 0,
+				HasBuildSection:        options.Manifest.IsV2 && len(options.Manifest.Build) > 0,
+			})
+
 			return err
 		},
 	}
@@ -246,6 +263,10 @@ func (dc *DeployCommand) RunDeploy(ctx context.Context, deployOptions *Options) 
 	}
 
 	for depName, dep := range deployOptions.Manifest.Dependencies {
+		dep.Variables = append(dep.Variables, model.EnvVar{
+			Name:  "OKTETO_ORIGIN",
+			Value: "okteto-deploy",
+		})
 		pipOpts := &pipelineCMD.DeployOptions{
 			Name:         depName,
 			Repository:   dep.Repository,
--- a/cmd/deploy/deploy.go
+++ b/cmd/deploy/deploy.go
@@ -187,13 +187,14 @@ func Deploy(ctx context.Context) *cobra.Command {
 				options.Manifest.Deploy.Compose.Manifest != nil {
 				deployType = "compose"
 			}
+
 			analytics.TrackDeploy(analytics.TrackDeployMetadata{
 				Success:                err == nil,
 				IsOktetoRepo:           utils.IsOktetoRepo(),
 				Duration:               time.Since(startTime),
 				PipelineType:           c.PipelineType,
 				DeployType:             deployType,
-				IsPreview:              false, // TODO: How should we get this value?
+				IsPreview:              os.Getenv(model.OktetoCurrentDeployBelongsToPreview) == "true",
 				HasDependenciesSection: options.Manifest.IsV2 && len(options.Manifest.Dependencies) > 0,
 				HasBuildSection:        options.Manifest.IsV2 && len(options.Manifest.Build) > 0,
 			})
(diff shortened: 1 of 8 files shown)


pkg/okteto/pipeline.go
```diff
@@ -58,6 +61,7 @@ func (c *OktetoClient) DeployPipeline(ctx context.Context, name, repository, bra
 			"branch":     graphql.String(branch),
 			"variables":  variablesVariable,
 			"filename":   graphql.String(filename),
+			"source":     graphql.String(origin),
 		}
 
 		err := mutate(ctx, &mutation, queryVariables, c.client)
```
jLopezbarb, line 67: Should we check if the mutation returns the new error `"Unknown argument \"source\" on type \"GitDeploy\"")` so we can run the deprecated version of `deployPipeline`

pkg/config/config.go
```diff
@@ -261,3 +261,18 @@ func GetOktetoContextsStorePath() string {
...
+// GetDeployOrigin gets the pipeline deploy origin. This is the initiator of the
+// deploy action: web, cli, github-action, etc
+func GetDeployOrigin() (src string) {
+	src = os.Getenv(model.OktetoOriginEnvVar)
+	if src == "" {
+		src = "cli"
+	}
+	// deploys within another okteto deploy take precedence as a deploy origin.
+	// This is running okteto pipeline deploy as a step of another okteto deploy
+	if os.Getenv(model.OktetoWithinDeployCommandContextEnvVar) == "true" {
```
ifbyol, line 274: Maybe we need to revisit the places where we are checking this env var. Deploy command sets it as false almost at the beginning of the command: https://github.com/okteto/okteto/blob/master/cmd/deploy/deploy.go#L131

I see several checks that I'm not sure they are working as expected

pkg/okteto/pipeline.go
ifbyol: yes, I think we should do it because the `source` parameter was released on the OE 0.10.7 and probably there are still OE in previous versions. @pchico83 Can you confirm is there is any OE in version < 0.10.7?

pchico83: yes, this change should be backward compatible. One option is to pass this value as an envvar of the pipeline deployment, not sure if that makes things easier or not

maroshii: Got it. If we use the deprecatedDeploy tho it would send all deploys on 0.10.7 to the deprecated path of 0.10.0, right? I'll try to send it as an envvar as Pablo suggests

pkg/config/config.go
maroshii: This works for multi repo with `okteto pipeline deploy`. Eg: https://github.com/okteto/movies-api/blob/main/okteto-pipeline.yml#L3 

Not sure if `okteto deploy` inside another `okteto deploy`  is a thing? If so the metrics will be somewhat skewed but not sure how many of these occurrences there will be. I can always create another envvar

ifbyol: I don't think it will be frequent. For now I wouldn't consider that case until users start using it.

There is one case that I'm not sure which should be the right value for the metric. If you execute `okteto deploy -b master` or `okteto deploy -r https://github.com/okteto/movies`, it will execute directly a pipeline https://github.com/okteto/okteto/blob/master/cmd/deploy/deploy.go#L116 but it wouldn't be considered within an okteto deploy command. Which honestly, not sure if it is right or not.

Also, bear in mind that a pipeline is executed in the backend as an `okteto deploy`, in case it affects somehow the metrics

maroshii: @ifbyol can you take another look? I added `OKTETO_SKIP_CONFIG_CREDENTIALS_UPDATE` for the proxy case you describe and kept the origin `OKTETO_WITHIN_DEPLOY_COMMAND_CONTEXT` independent from all that so we can use it safely everywhere else knowing it will not change

ifbyol: @maroshii looks good to me, but there are other places where that variable was using before and not sure if they should use the new one or the old one. Looking in to the code, I would say they have to use `OKTETO_WITHIN_DEPLOY_COMMAND_CONTEXT`, but @jLopezbarb can you double check, please? If you have doubts about what we are talking about, let me know and we can do a quick call

cmd/deploy/deploy.go
```diff
@@ -179,8 +179,25 @@ func Deploy(ctx context.Context) *cobra.Command {
...
+				options.Manifest.Deploy.Compose.Manifest != nil {
+				deployType = "compose"
+			}
+			analytics.TrackDeploy(analytics.TrackDeployMetadata{
+				Success:                err == nil,
+				IsOktetoRepo:           utils.IsOktetoRepo(),
+				Duration:               time.Since(startTime),
+				PipelineType:           c.PipelineType,
+				DeployType:             deployType,
+				IsPreview:              false, // TODO: How should we get this value?
```
maroshii, line 196: @jLopezbarb is there a way to know this here?

maroshii:
NOTE TO SELF: we are still missing to track and send OKTETO_ORIGIN as a variable in the deployPreview mutation

cmd/deploy/deploy.go
jLopezbarb: The only thing I can think of is checking if its an okteto cluster and then call to the API to check if exists the namespace as preview or namespace

ifbyol: Possible options I see are:
* Use client-go library to get the ns information and check annotations/labels (scope). I don't like it because we would be coupling the CLI with the logic of checking the scope to know if it is a preview env or a normal namespace and it would be harder to remove the scope in the future
* Pass an environment variable to the pipeline job specifying if it is a normal pipeline or a preview environment. In that way, that environment variable would be injected in the `okteto deploy` command and it would be available here to check

pkg/analytics/track.go
```diff
@@ -253,20 +254,39 @@ func TrackDestroyStack(success bool) {
...
+	DeployType             string
+	IsPreview              bool
+	HasDependenciesSection bool
+	HasBuildSection        bool
+}
+
+// TrackDeploy sends a tracking event to mixpanel when the user deploys a pipeline
+func TrackDeploy(m TrackDeployMetadata) {
+	// skip deploys from nested okteto deploys and manifest dependencies
+	if config.GetDeployOrigin() == "okteto-deploy" {
```
ifbyol, line 272: 👍

pkg/analytics/track.go
```diff
@@ -49,7 +49,7 @@ const (
 	doctorEvent              = "Doctor"
 	buildEvent               = "Build"
 	buildTransientErrorEvent = "BuildTransientError"
-	deployEvent              = "Deploy"
+	deployEvent              = "OktetoDeploy"
```
maroshii, line 52: @rlamana this is the new event name. Between this change and the recent backend change we'll no longer get the `Deploy` event. Just confirming we are ok with this

rlamana: So, is this the event for the nested ones? Or in general? If we end up starting a new mixpanel project like we've been discussing, I want this event to still be `Deploy`. And if we end up using a prefix, I think all these events (UP, CONTEXT, etc..) should have the same prefix.

pkg/analytics/track.go
```diff
@@ -32,7 +32,7 @@ import (
 const (
 	// skipcq GSC-G101
 	// This is mixpanel's public token, is needed to send analytics to the project
-	mixpanelToken = "92fe782cdffa212d8f03861fbf1ea301"
+	mixpanelToken = "db84e1636cf6489d13582457eb63dfeb"
```
maroshii, line 35: TODO: revert this change

ifbyol (approved):
LGTM with the context I have

pchico83 (approved):

PR #2336, opened 2022-03-09
removing duplicate code for okteto ns and ctx

RinkiyaKeDad:

## Proposed changes

- Removed duplicated code since we do the same thing when we run `okteto ns use` and `okteto ns` (same for `okteto ctx`)

https://user-images.githubusercontent.com/56963264/157429353-439fa939-3ddb-42b4-9a57-85211b44d9cf.mov

### DIFF
--- a/cmd/namespace/use.go
+++ b/cmd/namespace/use.go
@@ -62,9 +62,6 @@ func Use(ctx context.Context) *cobra.Command {
 				return err
 			}
 			err = nsCmd.Use(ctx, namespace)
-			if err != nil {
-				return err
-			}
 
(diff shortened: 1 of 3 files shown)


cmd/namespace/use.go
```diff
@@ -62,6 +62,10 @@ func Use(ctx context.Context) *cobra.Command {
 				return err
 			}
 			err = nsCmd.Use(ctx, namespace)
+			if err != nil {
```
pchico83, line 65: I think you need to remove this. That way we run `analytics.TrackNamespace(err == nil, len(args) > 0)` to track command errors

pchico83 (approved):

PR #2342, opened 2022-03-10
feat: set vars as options as env vars

jLopezbarb:

Fixes Deploy variables was not taken in count to expand manifest commands

## Proposed changes
- Set variables at the beingining of the command so they can be expanded on the okteto manifest
-

### DIFF
--- a/cmd/deploy/validate.go
+++ b/cmd/deploy/validate.go
@@ -31,3 +31,13 @@ func validateOptionVars(variables []string) error {
 	}
 	return nil
 }
+
+func setOptionVarsAsEnvs(variables []string) error {
+	for _, v := range variables {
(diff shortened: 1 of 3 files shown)


pchico83 (approved):

cmd/deploy/validate.go
```diff
@@ -0,0 +1,33 @@
...
+
+package deploy
+
+import (
+	"fmt"
+	"os"
+	"strings"
+)
+
+func validateOptionVars(variables []string) error {
```
rberrelleza, line 22: I don't think we should call this validate, this is validating AND setting the environment variables for the command. Can we do that in two separate functions, to make it easy to read?

PR #2357, opened 2022-03-15
fix manifest init

jLopezbarb:
Fixes Manifest init

## Proposed changes
- Fixes all the bugs with okteto init
- Create flag v1 for retrocompatibility
- Show docker compose option if there is at least one compose file
- Show manifest path when created
- Show a hint to run okteto init to continue configuring the manifest
- Fix indentation (4-> 2)
- Save with relative paths
- Add a commented dev section
- Marhsal compose on simple notation
- Only set dev env when there is a volume mount
- Call to okteto init on okteto up if no manifest

### DIFF
--- a/cmd/init.go
+++ b/cmd/init.go
@@ -78,6 +78,6 @@ func Init() *cobra.Command {
 	cmd.Flags().StringVarP(&opts.Context, "context", "c", "", "context target for generating the okteto manifest")
 	cmd.Flags().StringVarP(&opts.DevPath, "file", "f", utils.DefaultManifest, "path to the manifest file")
 	cmd.Flags().BoolVarP(&opts.Overwrite, "replace", "r", false, "overwrite existing manifest file")
-	cmd.Flags().BoolVarP(&opts.Version1, "v1", "", false, "execute v1")
+	cmd.Flags().BoolVarP(&opts.Version1, "v1", "", false, "create a v1 okteto manifest: www.okteto.com/docs/reference/manifest-v1/")
 	return cmd
 }
--- a/cmd/init.go
+++ b/cmd/init.go
@@ -79,5 +79,7 @@ func Init() *cobra.Command {
 	cmd.Flags().StringVarP(&opts.DevPath, "file", "f", utils.DefaultManifest, "path to the manifest file")
 	cmd.Flags().BoolVarP(&opts.Overwrite, "replace", "r", false, "overwrite existing manifest file")
 	cmd.Flags().BoolVarP(&opts.Version1, "v1", "", false, "create a v1 okteto manifest: www.okteto.com/docs/reference/manifest-v1/")
+	cmd.Flags().BoolVarP(&opts.AutoDeploy, "deploy", "", false, "deploy the application after generate the okteto manifest")
+	cmd.Flags().BoolVarP(&opts.AutoConfigureDev, "configure-devs", "", false, "configure devs after deploying the application")
 	return cmd
 }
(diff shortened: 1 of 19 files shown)


cmd/init.go
```diff
@@ -0,0 +1,83 @@
...
+			}
+			return nil
+		},
+	}
+
+	cmd.Flags().StringVarP(&opts.Namespace, "namespace", "n", "", "namespace target for generating the okteto manifest")
+	cmd.Flags().StringVarP(&opts.Context, "context", "c", "", "context target for generating the okteto manifest")
+	cmd.Flags().StringVarP(&opts.DevPath, "file", "f", utils.DefaultManifest, "path to the manifest file")
+	cmd.Flags().BoolVarP(&opts.Overwrite, "replace", "r", false, "overwrite existing manifest file")
+	cmd.Flags().BoolVarP(&opts.Version1, "v1", "", false, "execute v1")
```
pchico83, line 81:
```suggestion
	cmd.Flags().BoolVarP(&opts.Version1, "v1", "", false, "create a v1 okteto manifest: www.okteto.com/docs/reference/manifest-v1/")
```

pchico83:
`correctly` -> `successfully`

pkg/model/manifest.go
```diff
@@ -51,13 +52,26 @@ const (
...
-	deployHeadComment       = "The deploy section defines how to deploy your development environment\nMore info: https://www.okteto.com/docs/reference/manifest/#deploy"
-	devHeadComment          = "The dev section defines how to activate a development container\nMore info: https://www.okteto.com/docs/reference/manifest/#dev"
+	buildSvcEnvVars   = "You can use the following env vars to refer to this image in your deploy commands:\n - OKTETO_BUILD_%s_REGISTRY: image registry\n - OKTETO_BUILD_%s_REPOSITORY: image repo\n - OKTETO_BUILD_%s_IMAGE: image name\n - OKTETO_BUILD_%s_TAG: image tag"
+	deployHeadComment = "The deploy section defines how to deploy your development environment\nMore info: https://www.okteto.com/docs/reference/manifest/#deploy"
+	devHeadComment    = "The dev section defines how to activate a development container\nMore info: https://www.okteto.com/docs/reference/manifest/#dev"
+	devExample        = `dev:
```
pchico83, line 58: just keep `dev:`

pkg/model/manifest.go
```diff
@@ -51,13 +52,26 @@ const (
...
+    - name=$USER
+  forward:
+    - 8080:80
+  reverse:
+    - 9000:9001`
 	dependenciesHeadComment = "The dependencies section defines other git repositories to be deployed as part of your development environment\nMore info: https://www.okteto.com/docs/reference/manifest/#dependencies"
 	dependenciesExample     = `dependencies:
-  - https://github.com/okteto/b
-  - frontend/okteto.yml`
+  - https://github.com/okteto/b`
```
pchico83, line 74:
```suggestion
  - https://github.com/okteto/sample`
```

pchico83 (commented):
when running `okteto init` and `build` and `deploy` are already defined, we should only ask for deployment if the dev environment is not running. After this, somehow it's inferring all my dev containers instead of asking one by one, and dev containers are not stored in the okteto manifest.

pkg/model/manifest.go
jLopezbarb: I updated it with a more generic example, but I think it's useful to have a simple example to show the user the things you can configure from the dev environment

PR #2379, opened 2022-03-17
fix: ignore image if build exists and is okteto registry

### LINKED ISSUE #2359: `image` should be ignored if `build` is define in a compose file
**Describe the bug**
When running in clusters managed by okteto, `image` should be ignored if `build` is defined.

**Expected behavior**
In these cases, we should always build/push the image to the Okteto Registry.
We can show a warning in this case.

jLopezbarb:

Fixes #2359

This was a regression because of the changes that we needed to introduce to have the build working with stack 
## Proposed changes
- Add the if statement missing

### DIFF
--- a/pkg/cmd/stack/translate.go
+++ b/pkg/cmd/stack/translate.go
@@ -124,18 +124,19 @@ func translateBuildImages(ctx context.Context, s *model.Stack, options *StackDep
 		if svcInfo.Build == nil && len(svcInfo.VolumeMounts) == 0 {
 			continue
 		}
-		if !okteto.IsOkteto() && svcInfo.Image == "" {
+		buildInfo := svcInfo.Build
+		if !okteto.IsOkteto() && buildInfo.Image == "" {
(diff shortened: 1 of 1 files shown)


pkg/cmd/stack/translate.go
```diff
@@ -124,18 +124,19 @@ func translateBuildImages(ctx context.Context, s *model.Stack, options *StackDep
...
 			continue
 		}
-		if !okteto.IsOkteto() && svcInfo.Image == "" {
+		buildInfo := svcInfo.Build
+		if !okteto.IsOkteto() && buildInfo.Image == "" {
 			return fmt.Errorf("'build' and 'image' fields of service '%s' cannot be empty", svcName)
 		}
 
+		if okteto.IsOkteto() && !registry.IsOktetoRegistry(buildInfo.Image) {
+			buildInfo.Image = fmt.Sprintf("okteto.dev/%s-%s:okteto", s.Name, svcName)
```
pchico83, line 133: should it take into account if it is `okteto-with-volumes`?

pchico83 (approved):

Issue #2343, opened 2022-03-10
Output when deleting a preview env is misleading


jmacelroy:
**Describe the bug**
okteto preview destroy <name> returns success even if no deletion occurs. It seems to happen whenever the preview environment doesn't exist or if I don't have permissions to delete it.

**To Reproduce**
Attempt to delete a preview environment that doesn't exist or you don't have access to using `okteto preview destroy` The command exists cleanly and states `Preview environment destroyed` even though

**Expected behavior**
If I attempt to delete a preview env that doesn't exist then the command should respond stating I attempted to remove a preview environment that doesn't exist.

If I attempt to delete a preview env that I don't have access for should state that I am lacking permissions.

**Desktop (please complete the following information):**

- OS: macOS
- Version: 1.15.6

PR #2387, opened 2022-03-18
Don't bypass "not-found-error" when destroying preview environment

AgustinRamiroDiaz:
Fixes #
Helps with [Issue 2343](https://github.com/okteto/okteto/issues/2343)
More explicit error to the user when destroying preview environments

## Proposed changes

- Send the error through the error chain

### DIFF
--- a/pkg/okteto/preview.go
+++ b/pkg/okteto/preview.go
@@ -227,9 +227,6 @@ func (c *OktetoClient) DestroyPreview(ctx context.Context, name string) error {
 	}
 
 	err := mutate(ctx, &mutation, variables, c.client)
-	if oktetoErrors.IsNotFound(err) {
-		return nil
-	}
 	return err
 }
 


pkg/okteto/preview.go
```diff
@@ -227,9 +227,6 @@ func (c *OktetoClient) DestroyPreview(ctx context.Context, name string) error {
 	}
 
 	err := mutate(ctx, &mutation, variables, c.client)
-	if oktetoErrors.IsNotFound(err) {
```
pchico83, line 230: @AgustinRamiroDiaz why we need this change? Destroy is usually an idempotent operation

AgustinRamiroDiaz: It's to give an explicit error to the user. I forgot to say that it it related to [this](https://github.com/okteto/okteto/issues/2343) issue

pchico83: @AgustinRamiroDiaz but is it an error? I would show an info message here, but I wouldn't fail

AgustinRamiroDiaz: I get what you are saying. What benefits do we get from not failing in the Cobra command?

jLopezbarb: When we run this command from a Github Action we shouldn't fail if the preview doesn't exists. It can give the user an error that the user shouldn't be aware of

pchico83: @jmacelroy that's a good point, on the cli side we don't know if `not-found` if you don't have access to the preview, or if the preview doesn't exist...
Yes, I think we have to return an error here

AgustinRamiroDiaz: We can log with oktetoLog.Information. In this way users know they had a problem and the cli doesn't fail, letting other workflows still functional

pchico83: @AgustinRamiroDiaz I would go with the error as it's currently in the PR

jLopezbarb (approved):

rberrelleza:
@pchico83 @AgustinRamiroDiaz  this is a breaking change, it might cause existing preview environments configuration to fail.   If we need this change, can we document it on the CLI and in the GitHub Action? I think it should also reflect on the version number, but not sure what we are doing there.

PR #2398, opened 2022-03-22
add json/yaml outputs for okteto ctx list

### LINKED ISSUE #2204: `okteto context list` to print the list of contexts as json or yaml
I'd like the option to have the output of `okteto context list`  be json or yaml instead of tabs. This is needed to integrate with other tools.

Parsing the output is possible, but not ideal.

RinkiyaKeDad:

Fixes #2204 

## Proposed changes

https://user-images.githubusercontent.com/56963264/159427525-6d9b1055-a38c-43c8-991a-a994c7ff9d88.mov

I had to change the `output` persistent flag in `main.go` to a different name because otherwise, the new `output` flag for `ctx list` wasn't showing up in the command help. 

The older `output` flag was anyways confusing since it referred to the output format for logs and not a particular command. Also, does that even need to be a persistent flag?

### DIFF
--- a/cmd/context/list.go
+++ b/cmd/context/list.go
@@ -15,6 +15,7 @@ package context
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"os"
 	"text/tabwriter"
@@ -22,7 +23,10 @@ import (
 	"github.com/okteto/okteto/cmd/utils"
 	"github.com/okteto/okteto/pkg/okteto"
 	"github.com/spf13/cobra"
+	"gopkg.in/yaml.v3"
 )
+
+var output string
 
 // Lists all contexts managed by okteto
 func List() *cobra.Command {
@@ -40,7 +44,7 @@ func List() *cobra.Command {
 			return nil
 		},
 	}
-
+	cmd.Flags().StringVarP(&output, "output", "o", "", "Output format. One of: json|yaml")
 	return cmd
 }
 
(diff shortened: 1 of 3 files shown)


main.go
```diff
@@ -100,8 +100,8 @@ func main() {
 		},
 	}
 
-	root.PersistentFlags().StringVarP(&logLevel, "loglevel", "l", "warn", "amount of information outputted (debug, info, warn, error)")
-	root.PersistentFlags().StringVarP(&outputMode, "output", "o", oktetoLog.TTYFormat, "output format (tty, plain, json)")
+	root.PersistentFlags().StringVarP(&logLevel, "log-level", "l", "warn", "amount of information outputted (debug, info, warn, error)")
+	root.PersistentFlags().StringVar(&outputMode, "log-output", oktetoLog.TTYFormat, "output format for logs (tty, plain, json)")
```
jLopezbarb, line 104: We can't change this without coordinating with the team. The pipeline installer depends on this flag to run the pipelines. I would avoid changing this flag as much as possible but we can change it if we coordinate with the team. WDYT @ifbyol @pchico?

cmd/context/list.go
```diff
@@ -52,20 +56,41 @@ func executeListContext(ctx context.Context) error {
...
+			}
+			fmt.Fprintf(w, "%s\t%s\t%s\t%s\n", ctx.Name, ctx.Namespace, ctx.Builder, ctx.Registry)
+		}
+		w.Flush()
+	} else if output == "json" {
+		ctxRaw, err := json.MarshalIndent(ctxs, "", "\t")
+		if err != nil {
+			return err
+		}
+		fmt.Print(string(ctxRaw))
```
jLopezbarb, line 84: If we use new line at the end we can get rid off the '%' at the end of the command
```suggestion
		fmt.Println(string(ctxRaw))
```

main.go
RinkiyaKeDad: If we keep this flag the same, the changes in the PR should still work - only that `okteto context help` would not mention the new `output=json/yaml` flag.

pchico83: @jLopezbarb @ifbyol let's do this change, `-o` is a very common flag for output format.
Since this is more like an internal flag, we can easily change the installer when upgrading the cli in the installer.
For example, buildkit uses `--progress` for this.

ifbyol: Ok, as soon as these changes are in a CLI release, we can update the pipeline installer to pass the new flag. It is also something that we should see easily when upgrading the CLI in the backend because logs wouldn't display properly

pchico83: @RinkiyaKeDad to reduce overhead on monday's release, let's merge this right after releasing CLI 2.0.2

pchico83: @RinkiyaKeDad green light to merge whenever you are ready :-)

pchico83 (approved):

PR #2413, opened 2022-03-23
Tag image with manifest name

teresaromero:
## Proposed changes

- when using new manifest, if name is defined for the manifest, this should be part of the image tag and not the infered repository name
-


pchico83 (approved):

jLopezbarb (approved):

teresaromero:
has to change more files because the pointer recommendation of deepsource, please review again :) thanks

pchico83 (approved):

jLopezbarb (approved):

PR #2414, opened 2022-03-23
fix: integration test

teresaromero:

## Proposed changes

- nil pointer to options tag
-

### DIFF
--- a/pkg/cmd/stack/translate.go
+++ b/pkg/cmd/stack/translate.go
@@ -158,7 +158,7 @@ func translateBuildImages(ctx context.Context, s *model.Stack, options *StackDep
 		}
 
 		volumesToInclude := svcInfo.VolumeMounts
-		var options *build.BuildOptions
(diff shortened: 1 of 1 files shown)


pkg/cmd/stack/translate.go
```diff
@@ -158,7 +158,7 @@ func translateBuildImages(ctx context.Context, s *model.Stack, options *StackDep
 		}
 
 		volumesToInclude := svcInfo.VolumeMounts
-		var options *build.BuildOptions
+		options := &build.BuildOptions{}
```
jLopezbarb, line 161: is there any other definition like `var options *build.BuildOptions` we should replace it everywhere

pchico83 (approved):

pkg/cmd/stack/translate.go
teresaromero: i havent seen more

jLopezbarb (approved):

PR #2415, opened 2022-03-23
Update info messages about global registry

pchico83:


jLopezbarb (approved):

cmd/build.go
```diff
@@ -202,7 +202,7 @@ func buildV2(manifest *model.Manifest, cmdOptions *build.BuildOptions, args []st
 			oktetoLog.Debug("found OKTETO_GIT_COMMIT, optimizing the build flow")
 			globalReference := strings.Replace(cmdOptsFromManifest.Tag, okteto.DevRegistry, okteto.GlobalRegistry, 1)
 			if _, err := registry.GetImageTagWithDigest(globalReference); err == nil {
-				oktetoLog.Information("skipping build: image %s is already built at global registry", globalReference)
+				oktetoLog.Information("Skipping '%s' build. Image exists at Okteto Global Registry", service)
```
rberrelleza, line 205: I don't think there's value in mentioning that the image is in the global registry, since that is an implementation detail.

What about something like: 
"Image already exists in the registry"

PR #2456, opened 2022-03-30
feat: execute destroy on remote if branch or repo options

### LINKED ISSUE #2449: Support `okteto destroy` `--repository` and `--branch` flags
If `--repository` or `--branch` flags are specified, the destroy is executed on remote (the equivalent to `okteto pipeline destroy`.

jLopezbarb:

Fixes #2449

## Proposed changes
- If okteto destroy command has branch or repository flags execute it on remote
-

### DIFF
--- a/cmd/destroy/destroy.go
+++ b/cmd/destroy/destroy.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"os"
 	"os/signal"
+	"time"
 
 	"strings"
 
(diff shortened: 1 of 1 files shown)


rberrelleza:
Have we considered a --remote option or something like that? 

I don't think that is going to be obvious for the user that passing the branch means that it'll run on cloud vs local

pchico83:
@rberrelleza I also agree the current experience is confusing, although I don't have a clear alternative.
I would merge this one to have feature parity with `okteto deploy` while we have a better solution

pchico83:
> @pchico83 With regards to parity, I think we should do opposite of what you suggest and instead remove the --branch and --repo from deploy instead, since it seems like everyone in this thread agrees that's not the right way to do this. We can let "okteto pipeline deploy/destroy" live a bit longer without a deprecation message for now?

@rberrelleza SGTM

jLopezbarb:
Closing in favour of  https://github.com/okteto/okteto/pull/2477

PR #2457, opened 2022-03-30
fix: inject secrets on manifest

jLopezbarb:

Fixes a bug on okteto up where it reads the manifest before injecting the 

## Proposed changes
- ReLoad manifest after injecting secrets
-

### DIFF
--- a/cmd/context/utils.go
+++ b/cmd/context/utils.go
@@ -159,6 +159,11 @@ func LoadManifestWithContext(ctx context.Context, opts ManifestOptions) (*model.
 		return nil, err
 	}
 
+	// We need to read it again to propagate secrets env vars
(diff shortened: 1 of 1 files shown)


cmd/context/utils.go
```diff
@@ -159,6 +159,11 @@ func LoadManifestWithContext(ctx context.Context, opts ManifestOptions) (*model.
 		return nil, err
 	}
 
+	// We need to read it again to propagate secrets env vars
```
pchico83, line 162: why is the manifest loaded before this point? to the `context` and `namespace` from the manifest if needed?

jLopezbarb: Yes, we need it for the context resource which is just the context and the namespace

pchico83 (approved):

PR #2460, opened 2022-03-30
fix: ca-data-override

### LINKED ISSUE #2459: `Okteto deploy` doesn't work properly with minikube
**Describe the bug**
When I have my kubeconfig pointing to a minikube cluster and I try to deploy something using `okteto deploy` or similar (e.g. executing `okteto init` and launch the dev environment), I get an error executing kubectl commands. Here you have an example of the error. It happened to one customer

image

It might happen that we are not setting properly the kubeconfig to connect to the proxy or the kubeconfig used by the proxy to connect to kubernetes API.

Also, to prevent the warning we see in the image, we should set as empty the `certificate-authority` property when we set `certificate-authority-data`.

**To Reproduce**

1. Following this tutorial using minikube: https://dchinchure.medium.com/okteto-kubernetes-for-fast-go-development-7c3408296dff
2. When `okteto init` is executed, launch also the dev environment

**Expected behavior**
The deploy command should work as If I were using directly `kubectl`

-

**Additional context**
-

jLopezbarb:

Fixes Minikube proxy was failing to deploy because it had a `ca-authority-data` and a `ca-authority` 

## Proposed changes
- If the kubeconfig had `ca-authority` we need to remove it and  set only `ca-authority-data`

fixes #2459

### DIFF
--- a/cmd/deploy/config.go
+++ b/cmd/deploy/config.go
@@ -56,6 +56,9 @@ func (k *KubeConfig) Modify(port int, sessionToken, destKubeconfigFile string) e
 	// Change server to our proxy
 	clusterInfo.Server = fmt.Sprintf("https://localhost:%d", port)
 	// Set the certificate authority to talk with the proxy
+	if clusterInfo.CertificateAuthority != "" {
+		clusterInfo.CertificateAuthority = ""
+	}
 	clusterInfo.CertificateAuthorityData = cert
 
 	// Save on disk the config changes


ifbyol (approved):

jLopezbarb:
I want to mention that this was affecting not only Minikube. It was affecting all the clusters that have a path the authority certificate path set and not the authority certificate data

pchico83 (approved):

PR #2475, opened 2022-04-01
fix: file was not taking in count directory

### LINKED ISSUE #2430: `deploy` commands should run in the context of the folder where the okteto manifest is located
**Describe the bug**
If I run `okteto deploy -f my-folder/oktet.yml`, the deploy commands should run in the context of the folder `my-folder`.

For example, if I have a folder **app** with files `app.py` and the following `okteto.yml`:
```
deploy:
  - ls
```

running `okteto deploy -f app/okteto.yml` should print:
```
app.py
okteto.yml
```

jLopezbarb:

Fixes #2430

## Proposed changes
- If a path is used as a flag, we should run from the command from the directory of the path instead of current dir
-

### DIFF
--- a/cmd/deploy/deploy.go
+++ b/cmd/deploy/deploy.go
@@ -21,6 +21,7 @@ import (
 	"net/url"
 	"os"
 	"os/signal"
+	"path/filepath"
 	"sort"
 	"strings"
 	"time"
@@ -131,7 +132,12 @@ func Deploy(ctx context.Context) *cobra.Command {
 			// This is needed because the deploy command needs the original kubeconfig configuration even in the execution within another
 			// deploy command. If not, we could be proxying a proxy and we would be applying the incorrect deployed-by label
 			os.Setenv(model.OktetoSkipConfigCredentialsUpdate, "false")
-
+			if options.ManifestPath != "" {
+				dir := filepath.Dir(options.ManifestPath)
+				if err := os.Chdir(dir); err != nil {
+					return err
+				}
+			}
 			if err := contextCMD.LoadManifestV2WithContext(ctx, options.Namespace, options.K8sContext, options.ManifestPath); err != nil {
 				if err := contextCMD.NewContextCommand().Run(ctx, &contextCMD.ContextOptions{Namespace: options.Namespace}); err != nil {
 					return err
--- a/cmd/destroy/destroy.go
+++ b/cmd/destroy/destroy.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"os"
 	"os/signal"
+	"path/filepath"
 
 	"strings"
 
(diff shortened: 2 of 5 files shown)


pchico83 (approved):

cmd/deploy/deploy.go
```diff
@@ -131,7 +132,12 @@ func Deploy(ctx context.Context) *cobra.Command {
 			// This is needed because the deploy command needs the original kubeconfig configuration even in the execution within another
 			// deploy command. If not, we could be proxying a proxy and we would be applying the incorrect deployed-by label
 			os.Setenv(model.OktetoSkipConfigCredentialsUpdate, "false")
-
+			if options.ManifestPath != "" {
```
ifbyol, line 135: Does this take into consideration the case in which the manifest is within a `.okteto` folder? I think that is checked automatically and in that case `ManifestPath` would be empty

jLopezbarb: Yes it will, but only if the user specify the main file (not the one into the `.okteto` folder). If the user run `okteto deploy -f .okteto/okteto.yaml` it will run inside the `.okteto` folder.

ifbyol: yeah, but I mean, If I have stored the manifest in `.okteto/okteto.yml`, I would probably have defined the commands and file paths taking that into account. When I run `okteto deploy` (without the - f parameter), it will detect correctly the manifest because we check that path but the commands will probably fail because it is being executed from the current directory instead of doing it from `.okteto`

jLopezbarb: After talking with @pchico83, we have agreed that the `.okteto` directory is a special case which makes your manifests be in the root directory but hidden. Before we merge we have to make that if the user specifies the path `in .okteto` we do not deploy from that folder, but from the previous one from that

cmd/utils/manifest.go
```diff
@@ -31,3 +31,12 @@ func InferName(cwd string) string {
 	oktetoLog.Info("inferring name from git repository URL")
 	return model.TranslateURLToName(repo)
 }
+
+// GetWorkdirFromManifestPath sets the path
+func GetWorkdirFromManifestPath(manifestPath string) string {
+	dir := filepath.Dir(manifestPath)
+	if filepath.Base(dir) == ".okteto" {
```
pchico83, line 38: I could do `okteto deploy -f ~/app/.okteto/okteto.yml`, should we cover that case?

jLopezbarb: it will be executed from `~/app/`. Do we want to run it from `~`?

pchico83: It should be `~/app/`, perfect!

jLopezbarb: I will add a test for that before merging

PR #2479, opened 2022-04-01
feat: add endpoints command

jLopezbarb:

## Proposed changes
- Adds endpoints command
-

### DIFF
--- a/pkg/model/manifest.go
+++ b/pkg/model/manifest.go
@@ -474,6 +474,7 @@ func inferHelmTags(path string) string {
 		}
 		if image.Image != "" {
 			result += fmt.Sprintf(" --set %s.image=${OKTETO_BUILD_%s_IMAGE}", svcName, strings.ToUpper(svcName))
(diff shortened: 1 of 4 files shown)


pkg/model/manifest.go
```diff
@@ -474,6 +474,7 @@ func inferHelmTags(path string) string {
 		}
 		if image.Image != "" {
 			result += fmt.Sprintf(" --set %s.image=${OKTETO_BUILD_%s_IMAGE}", svcName, strings.ToUpper(svcName))
+			os.Setenv(fmt.Sprintf("OKTETO_BUILD_%s_IMAGE", strings.ToUpper(svcName)), image.Image)
```
pchico83, line 477: is this change related to this PR?

pchico83 (approved):

pkg/model/manifest.go
jLopezbarb: not related, but it's a bug that needs to be fixed. When we deploy a helm chart without Okteto Manifest we were not setting that env var which was causing an error

PR #2484, opened 2022-04-02
Remove DinD sidecard support

pchico83:
This feature doesn't provide a good DX with Okteto and it's not currently in use by any user.


rberrelleza:
Could we deprecate it first, and remove it on 2.1?

rberrelleza (commented):
Could we deprecate it first, and remove it on 2.1?

jLopezbarb (approved):

PR #2485, opened 2022-04-02
Add the "deployed-by" label for every pod created by `okteto deploy`

pchico83:

This is useful to identify the pods created by `okteto deploy`, for example, to query their logs.
The logic will be useful for other on the fly transformations we will need for divert.

### DIFF
--- a/cmd/deploy/translate.go
+++ b/cmd/deploy/translate.go
@@ -1,0 +1,209 @@
+// Copyright 2022 The Okteto Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package deploy
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/okteto/okteto/cmd/utils"
+	"github.com/okteto/okteto/pkg/model"
+	appsv1 "k8s.io/api/apps/v1"
+	batchv1 "k8s.io/api/batch/v1"
+	apiv1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func translateBody(b []byte, name string) ([]byte, error) {
+	var body map[string]json.RawMessage
+	if err := json.Unmarshal(b, &body); err != nil {
+		return nil, fmt.Errorf("could not unmarshal request: %s", err)
+	}
+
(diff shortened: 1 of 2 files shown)


cmd/deploy/translate.go
```diff
@@ -0,0 +1,209 @@
...
+	return nil
+}
+
+func translateStatefulSetSpec(body map[string]json.RawMessage, name string) error {
+	var spec appsv1.StatefulSetSpec
+	if err := json.Unmarshal(body["spec"], &spec); err != nil {
+		return fmt.Errorf("could not process statefulset spec: %s", err)
+	}
+	spec.Template.Labels[model.DeployedByLabel] = name
+	for i := range spec.VolumeClaimTemplates {
```
pchico83, line 130: @ifbyol with this, we will already know which volumes where created by statefulsets. I think we had logic to detect them on `okteto destroy -v`, could you point me to where that logic is? I would like to add a comment there to remove the logic at some point.

ifbyol: @pchico83 this is the function we use for what you mention: https://github.com/okteto/okteto/blob/master/pkg/k8s/namespaces/namespace.go#L142

teresaromero (approved):

jLopezbarb (approved):

ifbyol (approved):
@maroshii I think theses changes simplify https://github.com/okteto/okteto/pull/2120 and the use of `stern` for the logs

cmd/deploy/translate.go
pchico83: @ifbyol I won't modify the volumeclaimtempalte yet, modifying an existing statefulset volumeclaimtemplate is unsupported

ifbyol: @pchico83 other option could be to add that label in the volumes only on POST operations, not sure if there could be any operation in which could provoke an error

pchico83: @ifbyol good point, but I think we will always need to handle both use cases, because people can deploy things with kubectl directly

ifbyol: @pchico83 but even using directly `kubectl`, the API operation under the hood to create the statefulset would be a POST, right? I mean, Kubernetes API is rest, so we could check which operations are POST in the proxy to include the label or not. We are currently handling only POST and PUT: https://github.com/okteto/okteto/blob/master/cmd/deploy/proxy.go#L210

pchico83: yes, but I could create the statefulset with volumeclaimtemplates with `kubectl` and then, we will always need the code to remove those volumes manually

ifbyol: oh, yes, to delete the volume, we will need the code that does it manually, but starting to include these labels on creation will help us to group those volumes with the pipelines they belong when they are sleeping for example

pchico83: makes sense, I will update the logic :+1:

pchico83: still problematic, I could create the sfs with `okteto deploy` and then a `kubectl apply` would fail...

ifbyol: of, so let's not include it then

pchico83: sounds good. auto modifying this field is risky...

PR #2492, opened 2022-04-04
Jlopezbarb/add docker extension flag

jLopezbarb:
## Proposed changes
- Adds a docker extension flag for `okteto up` command which sets output format to plain and evicts every user input


pchico83:
@jLopezbarb let's keep this in a feature branch for now

PR #2496, opened 2022-04-04
fix: remove previous running proccess

### LINKED ISSUE #2294: Okteto CLI fails to 'okteto up' when development is active and port forwarding is enabled

jLopezbarb:

Fixes #2294

## Proposed changes
- check syncthing process and remove the okteto parents
-

### DIFF
--- a/cmd/up/process.go
+++ b/cmd/up/process.go
@@ -1,0 +1,190 @@
+// Copyright 2022 The Okteto Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package up
+
+import (
+	"fmt"
+	"io"
+	"runtime"
+	"strings"
+	"time"
+
+	gops "github.com/mitchellh/go-ps"
+	"github.com/okteto/okteto/pkg/config"
+	oktetoLog "github.com/okteto/okteto/pkg/log"
+	"github.com/shirou/gopsutil/process"
+)
+
+func (up *upContext) stopRunningUpOnSameFolder() error {
+	pList, err := process.Processes()
+	if err != nil {
+		return err
+	}
+	folderName := config.GetAppHome(up.Dev.Namespace, up.Dev.Name)
+	for _, p := range pList {
+		if p.Pid == 0 {
+			continue
+		}
+
+		name, err := p.Name()
+		if err != nil {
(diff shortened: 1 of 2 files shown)


cmd/up/process.go
```diff
@@ -0,0 +1,190 @@
...
+		if err != nil {
+			// it's expected go get EOF if the process no longer exists at this point.
+			if err != io.EOF {
+				oktetoLog.Infof("error getting name for process %d: %s", p.Pid, err.Error())
+			}
+			continue
+		}
+
+		// workaround until https://github.com/shirou/gopsutil/issues/1043 is fixed
+		if name == "" && runtime.GOOS == "darwin" && runtime.GOARCH == "arm64" {
```
rberrelleza, line 50: I think this is already fixed

cmd/up/process.go
```diff
@@ -0,0 +1,190 @@
...
+func getParent(p *process.Process) (*process.Process, error) {
+	name, err := p.Name()
+	if err != nil {
+		return nil, fmt.Errorf("can not get parent name")
+	}
+	parent, err := p.Parent()
+	if err != nil {
+		return nil, fmt.Errorf("can not find parent")
+	}
+	if parent.Pid < 100 {
```
rberrelleza, line 108: Is this also valid in windows?

pchico83:
@jLopezbarb we already have logic to hard terminate synching https://github.com/okteto/okteto/blob/master/pkg/syncthing/syncthing.go#L795.

Could we update that one to also kill the running `okteto up`? Listing processes is slow. that logic uses a channel to do this in parallel with the rest of the up sequence.

pchico83 (approved):
Looks great!
Could we write a test that runs one `up` after another one to validate the first `up` gets killed?

pkg/syncthing/syncthing.go
```diff
@@ -855,6 +853,27 @@ func (s *Syncthing) HardTerminate() error {
...
+	parent, err := p.Parent()
+	if err != nil {
+		return nil, fmt.Errorf("can not find parent")
+	}
+	if runtime.GOOS == "windows" && (parent.Pid == 0 || parent.Pid == 4) {
+		return nil, fmt.Errorf("can't remove root process")
+	} else if runtime.GOOS != "windows" && parent.Pid < 100 {
+		return nil, fmt.Errorf("can't remove root process")
+	}
+	pName, err := parent.Name()
```
teresaromero, line 870:
```suggestion
	pName, _ := parent.Name()
```

teresaromero (approved):

PR #2498, opened 2022-04-05
Add env value when empty

### LINKED ISSUE #2239: Empty variables are not expanded with its value from the environment
**Describe the bug**
Variable expansion doesn't work with this syntax:

```
services:
  nginx:
    image: nginx
    environment:
      TEST_ENV:
```

but they do with this one:
```
services:
  nginx:
    image: nginx
    environment:
      TEST_ENV: ${TEST_ENV}
```

teresaromero:

Fixes #2239 

## Proposed changes

- when a service environment var is empty value, get env by name and assign the value

### DIFF
--- a/pkg/model/stack_serializer_test.go
+++ b/pkg/model/stack_serializer_test.go
@@ -1602,7 +1602,9 @@ func Test_Environment(t *testing.T) {
 		t.Run(tt.name, func(t *testing.T) {
 
 			if tt.setEnvar {
-				os.Setenv("OKTETO_ENVTEST", "myvalue")
+				if err := os.Setenv("OKTETO_ENVTEST", "myvalue"); err != nil {
+					t.Fatal(err)
+				}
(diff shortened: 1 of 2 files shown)


jLopezbarb:
Can you add a test here:
https://github.com/okteto/okteto/blob/master/pkg/model/stack_serializer_test.go#L1566

pkg/model/stack_serializer_test.go
```diff
@@ -1588,6 +1601,13 @@ func Test_Environment(t *testing.T) {
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 
+			if tt.setEnvar {
```
jLopezbarb, line 1604: We could always set this variable. That way we can remove `tt.setEnvar` and the code is cleaner

jLopezbarb (approved):

PR #2503, opened 2022-04-05
refactor: build command

### LINKED ISSUE #2410: Refactor `okteto build` logic
**Describe the bug**
There are several places when the Okteto CLI is building docker images:
- okteto build
- okteto deploy (if images don't exit)
- okteto deploy --build
- okteto stack deploy (if images don't exit)
- okteto stack deploy --build

Looks like every one of them is using a slightly different `build` logic.
We should unify this implementation

jLopezbarb:
Fixes #2410

## Proposed changes
- Create build package so we have everything that we need for building in just one place
- Add unit tests mocking all the builds


pkg/types/build.go
```diff
@@ -0,0 +1,31 @@
...
+	File          string
+	NoCache       bool
+	OutputMode    string
+	Path          string
+	Secrets       []string
+	Tag           string
+	Target        string
+	Namespace     string
+	BuildToGlobal bool
+	Args          []string
```
teresaromero, line 29: minor thing, can we add here some comments regarding the meaning of each Option, i find Args and BuildArgs very similar and maybe this can be confusing in the future

cmd/build/buildv2.go
```diff
@@ -0,0 +1,226 @@
...
+	oktetoLog "github.com/okteto/okteto/pkg/log"
+	"github.com/okteto/okteto/pkg/model"
+	"github.com/okteto/okteto/pkg/okteto"
+	"github.com/okteto/okteto/pkg/registry"
+	"github.com/okteto/okteto/pkg/types"
+)
+
+// BuildV2 builds the images defined by a manifest
+func (bc *Command) BuildV2(ctx context.Context, manifest *model.Manifest, options *types.BuildOptions) error {
+	toBuildSvcs := getToBuildSvcs(manifest, options)
```
teresaromero, line 34: what about the services passed as args ? a user can select only 1 or multiple services from the manifest, not necesary are all of them

cmd/build/buildv2.go
```diff
@@ -0,0 +1,226 @@
...
+	isStackManifest := manifest.Type == model.StackType
+	buildSvcInfo := getBuildInfoWithoutVolumeMounts(manifest.Build[svcName], isStackManifest)
+
+	buildOptions := build.OptsFromBuildInfo(manifest.Name, svcName, buildSvcInfo, &types.BuildOptions{})
+
+	// Check if the tag is already on global/dev registry and skip
+	if build.ShouldOptimizeBuild(buildOptions.Tag) && !buildOptions.BuildToGlobal {
+		oktetoLog.Debug("found OKTETO_GIT_COMMIT, optimizing the build flow")
+		globalReference := strings.Replace(buildOptions.Tag, okteto.DevRegistry, okteto.GlobalRegistry, 1)
+		if _, err := registry.NewOktetoRegistry().GetImageTagWithDigest(globalReference); err == nil {
```
teresaromero, line 104: don't we include the registry as dependency from the `bc`command? could we do then bc.registry.GetImageTagWithDigest?

cmd/build/buildv2.go
```diff
@@ -0,0 +1,226 @@
...
+			oktetoLog.Debugf("Skipping '%s' build. Image already exists at the Okteto Registry", svcName)
+			return globalReference, nil
+		}
+		if registry.IsDevRegistry(buildOptions.Tag) {
+			if _, err := registry.NewOktetoRegistry().GetImageTagWithDigest(buildOptions.Tag); err == nil {
```
teresaromero, line 109: same here

cmd/build/buildv2.go
```diff
@@ -0,0 +1,226 @@
...
+				oktetoLog.Debugf("skipping build: image %s is already built", buildOptions.Tag)
+				return buildOptions.Tag, nil
+			}
+		}
+	}
+	if err := bc.BuildV1(ctx, buildOptions); err != nil {
+		return "", err
+	}
+	imageTagWithDigest, err := bc.Registry.GetImageTagWithDigest(buildOptions.Tag)
```
teresaromero, line 118: like this ? :)

cmd/deploy/deploy.go
```diff
@@ -313,10 +325,10 @@ func (dc *DeployCommand) RunDeploy(ctx context.Context, deployOptions *Options)
 		}
 	}
 
-	deployOptions.Manifest, err = deployOptions.Manifest.ExpandEnvVars()
-	if err != nil {
-		return err
-	}
+	// deployOptions.Manifest, err = deployOptions.Manifest.ExpandEnvVars()
```
teresaromero, line 328: ???

cmd/destroy/destroy.go
```diff
@@ -174,7 +174,7 @@ func (dc *destroyCommand) runDestroy(ctx context.Context, opts *Options) error {
 			Destroy: []model.DeployCommand{},
 		}
 	}
-	manifest, err = manifest.ExpandEnvVars()
+	err = manifest.ExpandEnvVars()
```
teresaromero, line 177:
```suggestion
	return manifest.ExpandEnvVars()
```

cmd/build/buildv2.go
jLopezbarb: I added them on the options. Function `getToBuildSvcs` checks if options args is not empty and return it, so that case should be covered

cmd/deploy/deploy.go
jLopezbarb: Removing it because it was already expanded after the build step

pkg/types/build.go
ifbyol: +1, I'm always get confused by them

cmd/build/buildv2.go
teresaromero: i did not saw that! awesome!

teresaromero (approved):
LGTM

pkg/types/build.go
jLopezbarb: 👍 
I will add a comment and rename `options.Args` into `options.CommandArgs` so it's clearer what those variables refers to.

cmd/build/environment.go
```diff
@@ -0,0 +1,40 @@
...
+	"fmt"
+	"os"
+	"strings"
+
+	oktetoLog "github.com/okteto/okteto/pkg/log"
+	"github.com/okteto/okteto/pkg/registry"
+)
+
+// SetServicetEnvVars set okteto build env vars
+func (*Command) SetServicetEnvVars(service, reference string) error {
```
ifbyol, line 26: is the `t` after `service` intended? Also, I don'w see any case in which this function returns an error, I think we can delete the return type (unless it is needed for some interface). It would simplify the client as we wouldn't have to check the error

cmd/build/validation.go
```diff
@@ -0,0 +1,120 @@
...
+	oktetoLog "github.com/okteto/okteto/pkg/log"
+	"github.com/okteto/okteto/pkg/model"
+	"github.com/okteto/okteto/pkg/okteto"
+	"github.com/okteto/okteto/pkg/registry"
+	"github.com/okteto/okteto/pkg/types"
+	"golang.org/x/sync/errgroup"
+)
+
+// GetServicesToBuild returns the services it has to built because they are not already built
+func (bc *Command) GetServicesToBuild(ctx context.Context, manifest *model.Manifest) ([]string, error) {
```
ifbyol, line 32: This doesn't sound to me as a function to be within a `validation.go` file. For example, in `buildv2.go` there is a function called `validateOptions` that would fit here.

I think `services.go` would be a better name for this file

cmd/build/buildv2.go
```diff
@@ -0,0 +1,226 @@
...
+func (bc *Command) buildSvcFromDockerfile(ctx context.Context, manifest *model.Manifest, svcName string, options *types.BuildOptions) (string, error) {
+	oktetoLog.Information("Building image for service '%s'", svcName)
+	isStackManifest := manifest.Type == model.StackType
+	buildSvcInfo := getBuildInfoWithoutVolumeMounts(manifest.Build[svcName], isStackManifest)
+
+	buildOptions := build.OptsFromBuildInfo(manifest.Name, svcName, buildSvcInfo, &types.BuildOptions{})
+
+	// Check if the tag is already on global/dev registry and skip
+	if build.ShouldOptimizeBuild(buildOptions.Tag) && !buildOptions.BuildToGlobal {
+		oktetoLog.Debug("found OKTETO_GIT_COMMIT, optimizing the build flow")
```
ifbyol, line 102: nit: From a client perspective sounds a bit weird this debug. As a client, I shouldn't care how the check to be optimized the build is done. Just `optimizing the build flow` would be a better log probably.

If in the future the check is done in other way, this could be missleading

cmd/build/validation.go
```diff
@@ -0,0 +1,120 @@
...
+
+func (bc *Command) checkServicesToBuild(service string, manifest *model.Manifest, ch chan string) error {
+	buildInfo := manifest.Build[service]
+	isStack := manifest.Type == model.StackType
+	if isStack && okteto.IsOkteto() && !registry.IsOktetoRegistry(buildInfo.Image) {
+		buildInfo.Image = ""
+	}
+	opts := build.OptsFromBuildInfo(manifest.Name, service, buildInfo, &types.BuildOptions{})
+
+	if build.ShouldOptimizeBuild(opts.Tag) {
```
ifbyol, line 70: I think there is some logic similar to this in `buildv2` file. Do they have any difference? We are doing this refactor to avoid to have different build logic. I would like to avoid to have different logic to optimize the builds. Can you double check?

cmd/build/buildv2.go
```diff
@@ -0,0 +1,226 @@
...
+	oktetoErrors "github.com/okteto/okteto/pkg/errors"
+	oktetoLog "github.com/okteto/okteto/pkg/log"
+	"github.com/okteto/okteto/pkg/model"
+	"github.com/okteto/okteto/pkg/okteto"
+	"github.com/okteto/okteto/pkg/registry"
+	"github.com/okteto/okteto/pkg/types"
+)
+
+// BuildV2 builds the images defined by a manifest
+func (bc *Command) BuildV2(ctx context.Context, manifest *model.Manifest, options *types.BuildOptions) error {
```
ifbyol, line 33: Sounds a bit weird that the command have 2 different builds functions. I think that the command should have 1 single build function and has a dependency with a builder. That builder would be a V1 or V2, but not both at the same time to avoid confusions.

pkg/cmd/stack/translate.go
```diff
@@ -120,131 +117,32 @@ func translateServiceEnvFile(ctx context.Context, svc *model.Service, svcName, f
...
-			buildInfo = &model.BuildInfo{
-				VolumesToInclude: svcInfo.VolumeMounts,
-			}
-		} else {
-			buildInfo.Image = svcInfo.Image
+	manifest := model.NewManifestFromStack(s)
+	builder := buildCMD.NewBuildCommand()
+	if options.ForceBuild {
+		buildOptions := &types.BuildOptions{}
+		if err := builder.BuildV2(ctx, manifest, buildOptions); err != nil {
```
ifbyol, line 124: Would this work for a stack with a v2 manifest?

ifbyol: Following up my comment about having `BuildV1` and `BuildV2` in the same command. It would be less confusing from a client perspective to create a single builder and use `builder.Build`

cmd/build/validation.go
jLopezbarb: Yes, this is intended, one is to prepare for building and the other one to check if it's already built

cmd/build/v2/environment.go
```diff
@@ -0,0 +1,40 @@
...
+)
+
+// SetServiceEnvVars set okteto build env vars
+func (*OktetoBuilder) SetServiceEnvVars(service, reference string) {
+	reg, repo, tag, image := registry.GetReferecenceEnvs(reference)
+
+	oktetoLog.Debugf("envs registry=%s repository=%s image=%s tag=%s", reg, repo, image, tag)
+
+	// Can't add env vars with -
+	sanitizedSvc := strings.ToUpper(strings.ReplaceAll(service, "-", "_"))
```
AgustinRamiroDiaz, line 32: I think we'd be better not asuming anything from the services. We could replace everything that is not `[A-Z0-9_]` for an `_`

AgustinRamiroDiaz (commented):
Great work! It's a lot of refactor

cmd/build/v2/environment.go
jLopezbarb: We are just replacing dashes into underscores because we can not set an envvar with dashes

ifbyol (commented):
It would be nice to have this in a branch where the integration tests run to see if after the refactor they are passing.

Apart from that, I see that sometimes we are using directly the v1 or v2 builder but II don't see a pattern or something obvious to indicate why we are use one or the other. From a reviewer/new person starting with the CLI it is a bit confusing that

cmd/deploy/deploy.go
```diff
@@ -97,6 +96,7 @@ type DeployCommand struct {
 	Executor           executor.ManifestExecutor
 	TempKubeconfigFile string
 	K8sClientProvider  okteto.K8sClientProvider
+	Builder            *buildv2.OktetoBuilder
```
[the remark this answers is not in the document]
jLopezbarb: We were implementing another buildV2 on deploy. The refactor was to achieve two only ways of building images, v2 is to build from a manifest all the images and v1 to build from a docker file

cmd/build/v2/environment_test.go
```diff
@@ -0,0 +1,221 @@
...
+			if tagEnvValue != tt.expTag {
+				t.Errorf("tag - expected %s , got %s", tt.expTag, tagEnvValue)
+
+			}
+
+		})
+	}
+}
+
+func TestExpandCommandVariables(t *testing.T) {
```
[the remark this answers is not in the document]
jLopezbarb: Yes, but I added on this test on this file because `SetServiceEnvVars` is only called inside of a build and we are checking that the build flow sets those env vars on the manifest

cmd/build/v2/environment_test.go
```diff
@@ -0,0 +1,221 @@
...
+
+	// error from the build
+	assert.NoError(t, err)
+
+	// Not substituted by empty string
+	assert.NotEmpty(t, manifest.Deploy.Commands[0].Command)
+	assert.NotEqual(t, manifest.Deploy.Commands[0].Command, "${OKTETO_BUILD_TEST_IMAGE}")
+}
+
+func TestExpandStackVariables(t *testing.T) {
```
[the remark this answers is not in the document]
jLopezbarb: Same as above

cmd/push.go
```diff
@@ -289,7 +290,7 @@ func buildImage(ctx context.Context, dev *model.Dev, imageFromApp, oktetoRegistr
 		BuildArgs:  buildArgs,
 		OutputMode: pushOpts.Progress,
 	}
-	if err := build.Run(ctx, buildOptions); err != nil {
```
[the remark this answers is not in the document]
jLopezbarb: It's different because here we are applying it to a Build specification of the dev environment, on the other hand on the deploy command we were calling `build.Run` for every build specification on the full manifest

PR #2508, opened 2022-04-06
Custom login html (and redirect) for "okteto context use"

pchico83:

Fixes https://github.com/okteto/docker-desktop-extension/issues/20


pkg/cmd/login/html.go
```diff
@@ -98,7 +98,100 @@ var loginHTML = []byte(`
...
+    </div>
+  </body>
+</html>
+`)
+
+var loginHTMLDockerDesktop = []byte(`
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta http-equiv="refresh" content="1; url = docker-desktop://dashboard/open">
```
pchico83, line 112: This is what implements the redirect to docker desktop after 1s

jLopezbarb:
Should we add a metric to track how many users are using this through docker desktop?

cmd/context/use.go
```diff
@@ -61,6 +63,11 @@ Or a Kubernetes context:
 				ctxOptions.Context = strings.TrimSuffix(args[0], "/")
 			}
 
+			switch ctxOptions.Origin {
+			case login.CLIOrigin, login.DockerDesktopOrigin:
+			default:
+				return fmt.Errorf("'origin' mus be either '%s' or '%s'", login.CLIOrigin, login.DockerDesktopOrigin)
```
rberrelleza, line 69: the VSCode extension sets a different origin

pkg/cmd/login/browser.go
```diff
@@ -44,9 +50,17 @@ func (h *Handler) handle() http.Handler {
 			return
 		}
 
-		if _, err := w.Write(loginHTML); err != nil {
-			h.errChan <- fmt.Errorf("failed to write to the response: %s", err)
-			w.WriteHeader(http.StatusInternalServerError)
+		switch h.origin {
+		case CLIOrigin:
+			if _, err := w.Write(loginHTMLCLI); err != nil {
```
rberrelleza, line 55: can you set a variable instead so we don't repeat the same code twice?

pkg/cmd/login/html.go
rberrelleza: Instead of duplicating the template, you can add a templated string and pass the full " <meta http-equiv="refresh" content="1; url = docker-desktop://dashboard/open">" as an optional value.

rberrelleza:
> Should we add a metric to track how many users are using this through docker desktop?

I think we already track the origin on mixpanel right @pchico83 ?

cmd/context/use.go
```diff
@@ -61,6 +63,11 @@ Or a Kubernetes context:
 				ctxOptions.Context = strings.TrimSuffix(args[0], "/")
 			}
 
+			switch ctxOptions.Origin {
+			case login.CLIOrigin, login.DockerDesktopOrigin:
+			default:
```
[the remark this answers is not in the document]
pchico83: we do have a default value, this is only if you set manually a value that's not supported

cmd/context/use.go
pchico83: the docker desktop extension doesn't support env vars at command execution, that's why we need to use a `hidden` flag for now

pkg/cmd/login/html.go
pchico83: yes!

pchico83:
> Should we add a metric to track how many users are using this through docker desktop?

Good point! I added the `OKTETO_ORIGIN` envvar. This way, our tracking logic already handles the origin

jLopezbarb (approved):

PR #2510, opened 2022-04-07
fix: compose ports on dev manifest

### LINKED ISSUE #2495: Debugger ports, like 5005 should be exposed on localhost when using docker compose
**Describe the bug**
I have a docker compose file with this forward:
```
ports:
  - 5005:5005
```

but 5005 is not exposed in localhost.

**Additional context**
We don't want to create a public DNS in this case, but 5005 should be forwarded to localhost

jLopezbarb:

Fixes #2495

## Proposed changes
- The list of ports that should not create an endpoint should be added on up manifest
-


integration/up_test.go
```diff
@@ -744,6 +746,72 @@ func TestUpAutocreate(t *testing.T) {
...
+	svc, err := getService(ctx, namespace, "vote")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if len(svc.Spec.Ports) != 1 {
+		t.Fatalf("Expected to have only one endpoint for svc 'vote' but got %d", len(svc.Spec.Ports))
+	}
+
+	if !model.IsPortAvailable("localhost", 5005) {
```
pchico83, line 797: could we check 5005 is not added as a public endpoint?

pchico83 (approved):

teresaromero (approved):

PR #2513, opened 2022-04-07
Fix envs from envfile

### LINKED ISSUE #2381: Support for empty env vars when using `env_files`
**Describe the bug**
I am using an `env_file` in my compose file like this one:
```
AUTH0_SECRET_KEY=
```
but it's not getting the value from my `AUTH0_SECRET_KEY=` env var. I have to change my file to be:
```
AUTH0_SECRET_KEY=${AUTH0_SECRET_KEY}
```
which is quite redundant.

teresaromero:

Fixes #2381 

## Proposed changes

- check the value for env from envfiles at stack or dev and infer the value if empty
-

### DIFF
--- a/pkg/model/dev.go
+++ b/pkg/model/dev.go
@@ -614,10 +614,15 @@ func (dev *Dev) expandEnvFiles() error {
 		}
 
 		for name, value := range envMap {
-			dev.Environment = append(
-				dev.Environment,
-				EnvVar{Name: name, Value: value},
-			)
+			if value == "" {
+				value = os.Getenv(name)
+			}
+			if value != "" {
+				dev.Environment = append(
+					dev.Environment,
+					EnvVar{Name: name, Value: value},
+				)
+			}
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


jLopezbarb (approved):

pkg/model/dev.go
```diff
@@ -614,10 +614,15 @@ func (dev *Dev) expandEnvFiles() error {
 		}
 
 		for name, value := range envMap {
-			dev.Environment = append(
-				dev.Environment,
-				EnvVar{Name: name, Value: value},
-			)
+			if value == "" {
```
ifbyol, line 617: The code in this function and the one in `translateServiceEnvFile` within translate.go looks very similar. Not in this PR but I think we have to consider to extract the code to a common place to avoid slightly different implementations of the same code as we have/had with `build`.

I'm assuming that the logic is the same. Taking a quick look seems almost the same

ifbyol (approved):

pkg/model/dev.go
teresaromero: yes, also with Environment variables

PR #2515, opened 2022-04-08
added --all flag in okteto down

### LINKED ISSUE #2255: Support for `okteto down --all`
I want to disable all my dev containers in a single command.

`okteto down -all` deactivates all my active dev containers

RinkiyaKeDad:

Fixes #2255

## Proposed changes

- Had to create a new function which returns a slice of all dev containers because we were using `GetDevFromManifest` in a lot of other places

Still have to test this out, will update the PR after that.

### DIFF
--- a/cmd/down.go
+++ b/cmd/down.go
@@ -63,19 +63,10 @@ func Down() *cobra.Command {
 			}
 
 			if all {
-				devs, err := utils.GetAllDevsFromManifest(manifest)
+				err := allDown(ctx, manifest, rm)
 				if err != nil {
 					return err
 				}
-				for _, dev := range devs {
-					if err := runDown(ctx, dev, rm); err != nil {
-						analytics.TrackDown(false)
-						err = fmt.Errorf("%w\n    Find additional logs at: %s/okteto.log", err, config.GetAppHome(dev.Namespace, dev.Name))
-						return err
-					}
-				}
-				analytics.TrackDown(true)
-				return nil
 			}
 
 			devName := ""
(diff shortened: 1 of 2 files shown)


cmd/down.go
```diff
@@ -61,6 +62,22 @@ func Down() *cobra.Command {
 				return err
 			}
 
+			if all {
+				devs, err := utils.GetAllDevsFromManifest(manifest)
```
pchico83, line 66: why can't you iterate directly on `range manifest.Dev`?
Also, could you move this logic to its own function?

pchico83 (approved):

cmd/down.go
```diff
@@ -83,11 +91,27 @@ func Down() *cobra.Command {
...
 	return cmd
 }
 
+func allDown(ctx context.Context, manifest *model.Manifest, rm bool) error {
+	if len(manifest.Dev) == 0 {
+		return fmt.Errorf("okteto manifest has no 'dev' section. Configure it with 'okteto init'")
+	}
+	for _, dev := range manifest.Dev {
+		if err := runDown(ctx, dev, rm); err != nil {
+			analytics.TrackDown(false)
```
jLopezbarb, line 106: Should we track the down event on all the defined devs? 
What happens if the dev environment is not on dev mode? We don't fail but we are tracking that event

cmd/down.go
```diff
@@ -83,11 +91,27 @@ func Down() *cobra.Command {
 
 	cmd.Flags().StringVarP(&devPath, "file", "f", utils.DefaultManifest, "path to the manifest file")
 	cmd.Flags().BoolVarP(&rm, "volumes", "v", false, "remove persistent volume")
+	cmd.Flags().BoolVarP(&all, "all", "A", false, "deactivate all running dev containers")
```
teresaromero, line 94: minor doubt: is this case sensitive? is there a reason for "A" and not "a" ?

cmd/down.go
teresaromero: What happens now if the service is not on dev mode and the user runs `okteto down <service>` ?

teresaromero (approved):

cmd/down.go
RinkiyaKeDad: I'm not sure what `TrackDown` does, lmk what you decide and I can update the PR accordingly then.

cmd/down.go
RinkiyaKeDad: Yup it is case sensitive. I went with "A" instead of "a" because that is what's commonly used by other tools to represent "all". For example: `kubectl get pods -A`

PR #2523, opened 2022-04-11
add the interface for static pod

### LINKED ISSUE #2516: [question] How to connect the static pod? -- the application not found
**Describe the bug**

I updated the image of controller-manager 's pod in 「kind」，but I don't know how to connect the static pod ? When I execute the follow command, it show **Application 'kube-controller-manager-star-cluster-control-plane' not found in namespace 'kube-system'**

**To Reproduce**

1. I execute the `kubectl apply` the okteto yaml to choose the static pod `kube-controller-manager-star-cluster-control-plane`
```yaml
name: kube-controller-manager-star-cluster-control-plane
namespace: kube-system
image: k8s-dev-image:v1.1
workdir: /app
command: ["bash"]
```
2. Then `okteto up`
```shell
[root@localhost ~]# okteto up
 i  Using kube-system @ kind-star-cluster as context
 x  Application 'kube-controller-manager-star-cluster-control-plane' not found in namespace 'kube-system'
    Verify that your application is running and your okteto context is pointing to the right namespace
    Or set the 'autocreate' field in your okteto manifest if you want to create a standalone development container
    More information is available here: https://okteto.com/docs/reference/cli/#up


[root@localhost ~]# k get pods -A
NAMESPACE                           NAME                                                             READY   STATUS    RESTARTS       AGE
kube-system                         coredns-64897985d-4cgn6                                          1/1     Running   2 (2d4h ago)   7d21h
kube-system                         coredns-64897985d-lfjhs                                          1/1     Running   2 (2d4h ago)   7d21h
kube-system                         etcd-star-cluster-control-plane                                  1/1     Running   0              2d4h
kube-system                         kindnet-pvxg8                                                    1/1     Running   2 (2d4h ago)   7d21h
kube-system                         kube-apiserver-star-cluster-control-plane                        1/1     Running   0              2d4h
kube-system                         kube-controller-manager-star-cluster-control-plane               1/1     Running   2 (2d4h ago)   7d21h
kube-system                         kube-proxy-jzswq                                                 1/1     Running   2 (2d4h ago)   7d21h
kube-system                         kube-scheduler-star-cluster-control-plane                        1/1     Running   2 (2d4h ago)   7d21h
local-path-storage                  local-path-provisioner-5ddd94ff66-q5dlg                          1/1     Running   5 (22h ago)    7d21h
```

**Expected behavior**

`okteto up` successfully execute

- OS: [e.g. macOS] CentOS Linux release 7.9.2009 (Core)
- Version [e.g. 22] okteto version 2.0.2 

**Additional context**

Charlie17Li:
Fixes #2516 

## Proposed changes

- add the interface for static pod

### DIFF
--- a/pkg/k8s/apps/interface.go
+++ b/pkg/k8s/apps/interface.go
@@ -46,3 +46,20 @@ type App interface {
 
 	Divert(username string) App
 }
+
+type StaticPod interface {
+	Kind() string
+	ObjectMeta() metav1.ObjectMeta
+	PodSpec() *apiv1.PodSpec
+
+	DevClone() App
+
+	CheckConditionErrors(dev *model.Dev) error
+	GetRunningPod(ctx context.Context, c kubernetes.Interface) (*apiv1.Pod, error)
+
+	Refresh(ctx context.Context, c kubernetes.Interface) error
+	Watch(ctx context.Context, result chan error, c kubernetes.Interface)
+	Deploy(ctx context.Context, c kubernetes.Interface) error
+	PatchAnnotations(ctx context.Context, c kubernetes.Interface) error
+	Destroy(ctx context.Context, c kubernetes.Interface) error
+}
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 
 # IDEs
 .vscode
+.idea
 
 # Golang binary
 cli


pkg/k8s/apps/interface.go
```diff
@@ -46,3 +46,20 @@ type App interface {
 
 	Divert(username string) App
 }
+
+type StaticPod interface {
```
pchico83, line 50: @Charlie17Li you also need to add the code to implement the interface.
For example, this is the implementation for deployments:
https://github.com/okteto/okteto/blob/master/pkg/k8s/apps/deployments.go

Charlie17Li: OK👌🏻

Charlie17Li:
@pchico83 Maybe there is some problems for implementing the interface of static pod, I can’t use `client kubernetes.Interface` to update the info of static pods.  Because I find the following [info](https://kubernetes.io/docs/tasks/configure-pod-container/static-pod/) about static pods

> The kubelet automatically tries to create a [mirror Pod](https://kubernetes.io/docs/reference/glossary/?all=true#term-mirror-pod)
 on the Kubernetes API server for each static Pod. This means that the Pods running on a node are visible on the API server, but cannot be controlled from there. The Pod names will be suffixed with the node hostname with a leading hyphen.

And I want to interact with [Kubelelt API](https://github.com/cyberark/kubeletctl/blob/master/API_TABLE.md) directly, but I find there is no API for updating pods directly. Could you please give me some advice🙈

pchico83:
> @pchico83 Maybe there is some problems for implementing the interface of static pod, I can’t use `client kubernetes.Interface` to update the info of static pods. Because I find the following [info](https://kubernetes.io/docs/tasks/configure-pod-container/static-pod/) about static pods
> 
> > The kubelet automatically tries to create a [mirror Pod](https://kubernetes.io/docs/reference/glossary/?all=true#term-mirror-pod)
> >  on the Kubernetes API server for each static Pod. This means that the Pods running on a node are visible on the API server, but cannot be controlled from there. The Pod names will be suffixed with the node hostname with a leading hyphen.
> 
> And I want to interact with [Kubelelt API](https://github.com/cyberark/kubeletctl/blob/master/API_TABLE.md) directly, but I find there is no API for updating pods directly. Could you please give me some advice🙈

Yeah, this is tricky. You would need to remove and recreate a new pod. And keep the original pod yaml to restore it on `okteto down`.

Another option would be to modify the pod with [ephemeral containers](https://kubernetes.io/docs/tasks/debug-application-cluster/debug-running-pod/#ephemeral-container) but this would require a major refactor and a different `okteto up` implementation approach

Charlie17Li:
@pchico83 the scenario for using static pod is probably not common, so I think the first one is more appropriate.

updating the yaml in `staticPodPath`(of kubelet config) can actually update the static pod (that I know of).But I how to login the node, and update the yaml

And maybe we should consider the permissions of the 「node」and whether the 「node」 is virtual, physical or {in docker}

pchico83:
> @pchico83 the scenario for using static pod is probably not common, so I think the first one is more appropriate.
> 
> updating the yaml in `staticPodPath`(of kubelet config) can actually update the static pod (that I know of).But I how to login the node, and update the yaml
> 
> And maybe we should consider the permissions of the 「node」and whether the 「node」 is virtual, physical or {in docker}

yes, we cannot assume access to the node. I think the best approach atm is to recreate the pod via the Kubernetes API

PR #2535, opened 2022-04-13
Check okteto name env when manifest is compose

### LINKED ISSUE #2501: Support name param for dependencies
When adding dependencies to my okteto manifest, i want to specify the name of the dependency. now the name is inferred from the repo https://github.com/okteto/okteto/blob/master/pkg/model/serializer.go#L716

support `name` as field inside the dependency model

teresaromero:

Fixes #2501 

## Proposed changes

- check the Okteto name EnvVar when unmarshalling a compose manifest. 

Repo to test extended dependencies : https://github.com/teresaromero/movies-catalog/tree/tere/extended-dependencies

## Context
After investigating this, the problem was not the name at the dependencies section, but the name assigned to an okteto compose manifest, when this is part of a deploy command inside an okteto manifest.

So, for dependencies at the okteto manifest, this can be defined short or extended:

Short:
````
name: catalog

dependencies:
  - https://github.com/okteto/movies-frontend-with-divert
  - https://github.com/okteto/movies-rentals
````

Extended:
````
name: catalog

dependencies:
    frontend:
        repository: https://github.com/okteto/movies-frontend-with-divert
    rentals:
        repository: https://github.com/okteto/movies-rentals
````
When using extended, the name for the dependency (dev environment for that dependency) should be taken from the name assigned at the manifest, in this case frontend and rentals. Also, a manifest can have a name informed, so the dev environment for that manifest should be name, in this case, catalog.

When a manifest has a deployment of a compose file, like so:
````
name: catalog

dependencies:
    frontend:
        repository: https://github.com/okteto/movies-frontend-with-divert
    rentals:
        repository: https://github.com/okteto/movies-rentals

deploy:
    - okteto deploy -f mongodb-compose.yml
````

When transforming the compose into manifest v2, the name for the manifest was inferred from repo, or from the name flag, so, even though name is defined for okteto.yaml, name for the first command of deploy of this manifest, was not being transmitted to the compose. This also affected the dependencies which had a compose deployment, so 2 dev environments where created when deploying this manifests: 1 called catalog and other called movies-catalog which is the name of the repository.

Fixing this will require updating the cli version at the installer component, as the pipelines from dependencies are done by it. Also, there is an issue at the installer, which needs to be solved prior to include this fix into it.
https://github.com/okteto/app/issues/3632

### DIFF
--- a/pkg/model/stack.go
+++ b/pkg/model/stack.go
@@ -255,6 +255,10 @@ func GetStack(name, stackPath string, isCompose bool) (*Stack, error) {
 }
 
 func getStackName(name, stackPath, actualStackName string) (string, error) {
+	nameEnvVar := os.Getenv(OktetoNameEnvVar)
+	if nameEnvVar != "" {
+		return nameEnvVar, nil
+	}
 	if name != "" {
 		if err := os.Setenv(OktetoNameEnvVar, name); err != nil {
 			return "", err
--- a/pkg/model/stack.go
+++ b/pkg/model/stack.go
@@ -207,8 +207,8 @@ const (
 	DependsOnServiceCompleted DependsOnCondition = "service_completed_successfully"
 )
 
-// GetStack returns an okteto stack object from a given file
-func GetStack(name, stackPath string, isCompose bool) (*Stack, error) {
+// GetStackFromPath returns an okteto stack object from a given file
+func GetStackFromPath(name, stackPath string, isCompose bool) (*Stack, error) {
 	b, err := os.ReadFile(stackPath)
 	if err != nil {
 		return nil, err
@@ -389,11 +389,7 @@ func (s *Stack) Validate() error {
 		}
 		svc.IgnoreSyncVolumes(s)
 	}
-	if err := validateDependsOn(s); err != nil {
-		return err
-	}
-
-	return nil
+	return validateDependsOn(s)
 }
 
 func validateStackName(name string) error {
@@ -631,7 +627,7 @@ func (r *StackResources) IsDefaultValue() bool {
 	return false
 }
 
-func (svcResources ServiceResources) IsDefaultValue() bool {
+func (svcResources *ServiceResources) IsDefaultValue() bool {
 	return svcResources.CPU.Value.IsZero() && svcResources.Memory.Value.IsZero() && svcResources.Storage.Size.Value.IsZero() && svcResources.Storage.Class == ""
 }
 
@@ -699,7 +695,7 @@ func getStack(name, manifestPath string) (*Stack, error) {
 	if isPathAComposeFile(manifestPath) {
 		isCompose = true
 	}
-	stack, err := GetStack(name, manifestPath, isCompose)
+	stack, err := GetStackFromPath(name, manifestPath, isCompose)
 	if err != nil {
 		return nil, err
 	}
@@ -730,7 +726,7 @@ func getOverrideFile(stackPath string) (*Stack, error) {
 		if isPathAComposeFile(stackPath) {
 			isCompose = true
 		}
-		stack, err := GetStack("", overridePath, isCompose)
+		stack, err := GetStackFromPath("", overridePath, isCompose)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Review by pchico83 — COMMENTED** [review]
Does this mean I cannot deploy two compose files from the same dev environment?

**Comment by teresaromero** [comment]
> Does this mean I cannot deploy two compose files from the same dev environment?

This means the name of the dev environment was not being used when deploying a compose file, so the name was inferred from the repository name. With this change, the name of the dev environment will be transmited to the compose and it will be groupped to this name, if specific name is passed though a flag `name` this would be the one used for the compose.

**Comment by pchico83** [comment]
> > Does this mean I cannot deploy two compose files from the same dev environment?
> 
> This means the name of the dev environment was not being used when deploying a compose file, so the name was inferred from the repository name. With this change, the name of the dev environment will be transmited to the compose and it will be groupped to this name, if specific name is passed though a flag `name` this would be the one used for the compose.

Not sure if I am following. The name of the stack shouldn't have any impact on the grouping. Grouping is based on the `dev.okteto.com/deployed-by` label.

**Comment by teresaromero** [comment]
> > > Does this mean I cannot deploy two compose files from the same dev environment?
> > 
> > 
> > This means the name of the dev environment was not being used when deploying a compose file, so the name was inferred from the repository name. With this change, the name of the dev environment will be transmited to the compose and it will be groupped to this name, if specific name is passed though a flag `name` this would be the one used for the compose.
> 
> Not sure if I am following. The name of the stack shouldn't have any impact on the grouping. Grouping is based on the `dev.okteto.com/deployed-by` label.

this label is set:
- if name option at the deploy command existis, this is the name for deployed-by
- if no name option, and manifest has name - this is the one
- if no name option and no manifest name - the repo name is inferred
- if no name option and no compose manifest name - the repo name WAS inferred <- now we look for the env that is being set when deploying the parent manifest, and this name will be assigned. If a compose manifest has name, this would be the one use. If no env and no name, it will be inferred from the repo 

is this the correct behavour?

**Comment by pchico83** [comment]
> > > > Does this mean I cannot deploy two compose files from the same dev environment?
> > > 
> > > 
> > > This means the name of the dev environment was not being used when deploying a compose file, so the name was inferred from the repository name. With this change, the name of the dev environment will be transmited to the compose and it will be groupped to this name, if specific name is passed though a flag `name` this would be the one used for the compose.
> > 
> > 
> > Not sure if I am following. The name of the stack shouldn't have any impact on the grouping. Grouping is based on the `dev.okteto.com/deployed-by` label.
> 
> this label is set:
> 
> * if name option at the deploy command existis, this is the name for deployed-by
> * if no name option, and manifest has name - this is the one
> * if no name option and no manifest name - the repo name is inferred
> * if no name option and no compose manifest name - the repo name WAS inferred <- now we look for the env that is being set when deploying the parent manifest, and this name will be assigned. If a compose manifest has name, this would be the one use. If no env and no name, it will be inferred from the repo
> 
> is this the correct behavour?

What I mean is that `name` isn't the only way to set the `deployed-by` label, it shouldn't be coupled.
For example, with this PR, would you be able to deploy two compose files without names on those files (`name` is not a field in the compose spec):

```
deploy:
   compose:
       - mongodb-compose.yml
       - redis-compose.yml
``` 
or would the second compose collide with the first one because we are inferring the same name?
This is a common scenario, it should work

**Comment by pchico83** [comment]
@teresaromero I tested and deploying 2 compose files with `okteto deploy` commands in the `deploy` section is already broken. We can merge this then

File: pkg/model/stack.go [hunk 1]
```diff
@@ -262,6 +262,10 @@ func getStackName(name, stackPath, actualStackName string) (string, error) {
 		return name, nil
 	}
 	if actualStackName == "" {
+		nameEnvVar := os.Getenv(OktetoNameEnvVar)
```
pchico83, line 265: could we add a unit test for this?

pchico83 (approved):

jLopezbarb (approved):

teresaromero:
added couple changes regarding renaming of funcs and pointers,  due to deepsouce recomandations

jLopezbarb (approved):

PR #2539, opened 2022-04-13
fix: syncthing was killing parent process

jLopezbarb:

Fixes e2e tests were flaky because of this change

## Proposed changes
- Check if the parent found is not the one running and kill parent if it's not
-


pchico83 (approved):

integration/up_test.go
```diff
@@ -379,11 +379,19 @@ func TestUpDeployments(t *testing.T) {
...
+
+	newProccess, err := up(ctx, &newWg, namespace, name, manifestPath, oktetoPath, newUpErrorChannel)
 	if err != nil {
 		t.Fatal(err)
 	}
 
+	_, err = getContent(indexEndpoint, 150, upErrorChannel)
+	if err != nil {
+		t.Fatalf("failed to get index content: %s", err)
+	}
```
AgustinRamiroDiaz, line 393: Should we do another `getContent` before this?

PR #2546, opened 2022-04-13
feat: if its docker extension enable detach mode

jLopezbarb:

## Proposed changes
- Adds docker desktop flag to enable always the detach mode and load context with docker desktop origin
-

### DIFF
--- a/cmd/context/utils.go
+++ b/cmd/context/utils.go
@@ -17,6 +17,7 @@ import (
 	"context"
 	"fmt"
 	"net/url"
+	"os"
 	"strings"
 
 	"github.com/okteto/okteto/cmd/utils"
@@ -144,9 +145,10 @@ func LoadManifestWithContext(ctx context.Context, opts ManifestOptions) (*model.
 	}
 
 	ctxOptions := &ContextOptions{
-		Context:   ctxResource.Context,
-		Namespace: ctxResource.Namespace,
-		Show:      true,
+		Context:       ctxResource.Context,
+		Namespace:     ctxResource.Namespace,
+		Show:          true,
+		DockerDesktop: os.Getenv(model.OktetoOriginEnvVar) == model.OktetoDockerDesktopOrigin,
 	}
 
 	if err := NewContextCommand().Run(ctx, ctxOptions); err != nil {
--- a/cmd/up/up.go
+++ b/cmd/up/up.go
@@ -51,17 +51,18 @@ const ReconnectingMessage = "Trying to reconnect to your cluster. File synchroni
 
 // UpOptions represents the options available on up command
 type UpOptions struct {
-	DevPath    string
-	Namespace  string
-	K8sContext string
-	DevName    string
-	Devs       []string
-	Remote     int
-	Deploy     bool
-	Build      bool
-	ForcePull  bool
-	Reset      bool
-	Detach     bool
+	DevPath       string
+	Namespace     string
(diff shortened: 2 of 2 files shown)


cmd/up/up.go
```diff
@@ -79,7 +80,10 @@ func Up() *cobra.Command {
 			if err := upOptions.AddArgs(cmd, args); err != nil {
 				return err
 			}
-
+			if upOptions.DockerDesktop {
+				upOptions.Detach = true
```
pchico83, line 84: I wouldn't couple this one. The extension will set the `--detach` flag

cmd/context/utils.go
```diff
@@ -144,9 +145,10 @@ func LoadManifestWithContext(ctx context.Context, opts ManifestOptions) (*model.
 	}
 
 	ctxOptions := &ContextOptions{
-		Context:   ctxResource.Context,
-		Namespace: ctxResource.Namespace,
-		Show:      true,
+		Context:       ctxResource.Context,
+		Namespace:     ctxResource.Namespace,
+		Show:          true,
+		DockerDesktop: os.Getenv(model.OktetoOriginEnvVar) == model.OktetoDockerDesktopOrigin,
```
pchico83, line 151: is this needed if we are already doing?
https://github.com/okteto/okteto/pull/2546/files#diff-c1b929a650f782cf212e9b4b0ced5a7a298598bcb7473be505e7af1f06d6cf30R85

pchico83 (approved):

cmd/up/up.go
```diff
@@ -79,7 +80,10 @@ func Up() *cobra.Command {
...
+				os.Setenv(model.OktetoOriginEnvVar, model.OktetoDockerDesktopOrigin)
```
AgustinRamiroDiaz, line 85: Since if the only place where we need this information is [here](https://github.com/okteto/okteto/blob/c77c3537724bf769bba68585cf9f2ac1bae598b4/pkg/cmd/login/browser.go#L63) I think we should only have this line as a setter

cmd/context/utils.go
AgustinRamiroDiaz: Following with my previous comment and with what Pablo said, this should not be necessary. We store it as an envvar and retrieve it as an envvar

jLopezbarb: We need it because on `okteto up` there is no other way of knowing if it's coming from the docker desktop extension

cmd/up/up.go
jLopezbarb: We also need it to know the origins while tracking where the commands come from

cmd/up/up.go
```diff
@@ -305,6 +308,8 @@ func Up() *cobra.Command {
 	cmd.Flags().MarkHidden("pull")
 	cmd.Flags().BoolVarP(&upOptions.Reset, "reset", "", false, "reset the file synchronization database")
 	cmd.Flags().BoolVarP(&upOptions.Detach, "detach", "", false, "activate one more development containers in detached mode")
+	cmd.Flags().BoolVarP(&upOptions.DockerDesktop, "docker-desktop", "", false, "if the command is executed from the Docker Desktop extension")
```
pchico83, line 311: should we also disable confirmation of `stignore` file creation?

AgustinRamiroDiaz (approved):

pchico83 (approved):

PR #2558, opened 2022-04-19
Fix global registry - build and deploy

### LINKED ISSUE #2529: The Global Registry optimization should be always active unless `--no-cache` is set
**Describe the bug**
The Global Registry optimization isn't working on pipelines.
Pipelines execute `okteto deploy --build`. This should use the Global Registry optimization.

The only way to disable the Global Registry optimization is `okteto build --no-cache`.

**Additional context**
This [repo](https://github.com/okteto/movies-with-compose/blob/main/.github/workflows/build-global.yaml) build global images on each merge to `main`, but if you deploy the repo, the Global Registry optimization isn't used.

teresaromero:
Fixes #2529

### DIFF
--- a/pkg/cmd/build/build.go
+++ b/pkg/cmd/build/build.go
@@ -252,7 +252,7 @@ func getBuildOptionsFromManifest(service string, manifestBuildInfo *model.BuildI
 	return buildOptions
 }
 
-func overrideManifestBuildOptions(manifestBuildOptions *BuildOptions, cmdBuildOptions *BuildOptions) *BuildOptions {
+func overrideManifestBuildOptions(manifestBuildOptions, cmdBuildOptions *BuildOptions) *BuildOptions {
 	if cmdBuildOptions == nil {
 		cmdBuildOptions = &BuildOptions{}
 	}
--- a/pkg/cmd/build/build.go
+++ b/pkg/cmd/build/build.go
@@ -294,11 +294,72 @@ func OptsFromManifest(service string, manifestBuildInfo *model.BuildInfo, cmdBui
 	return overrideManifestBuildOptions(manifestBuildOptions, cmdBuildOptions)
 }
 
+// optimizedGlobalBuild returns if build should be optimized because the image is already at the global registry
+func (opts *BuildOptions) optimizedGlobalBuild() (string, bool, error) {
 	envGitCommit := os.Getenv(model.OktetoGitCommitEnvVar)
 	isLocalEnvGitCommit := strings.HasPrefix(envGitCommit, model.OktetoGitCommitPrefix)
-	return registry.IsOktetoRegistry(image) &&
-		envGitCommit != "" &&
-		!isLocalEnvGitCommit
+	shouldApply := registry.IsOktetoRegistry(opts.Tag) && envGitCommit != "" && !isLocalEnvGitCommit
+	if !shouldApply {
+		return opts.Tag, false, nil
+	}
+	oktetoLog.Debugf("Applying global build optimization for image %s", opts.Tag)
+
+	globalReference := opts.Tag
+	if registry.IsDevRegistry(opts.Tag) || registry.IsExtendedOktetoRegistry(opts.Tag) {
+		globalReference = registry.ReplaceTargetRepository(opts.Tag, okteto.GlobalRegistry, okteto.DefaultGlobalNamespace)
+
+	}
+	tagWithDigest, err := registry.GetImageTagWithDigest(globalReference)
+	if err != nil {
+		return opts.Tag, false, err
+	}
+	oktetoLog.Debugf("skipping build: image %s is already built", globalReference)
+	return tagWithDigest, true, nil
+}
+
+// checkImageAtRegistry returns if build should be optimized because the image is already at the dev registry
+func (opts *BuildOptions) checkImageAtRegistry() (string, bool, error) {
+	// only check at pipelines
+	if strings.HasSuffix(opts.Tag, model.OktetoDefaultImageTag) {
+		return opts.Tag, false, nil
+	}
+	oktetoLog.Debugf("Checking registry for image %s", opts.Tag)
+	tagWithDigest, err := registry.GetImageTagWithDigest(opts.Tag)
+	if err != nil {
+		return opts.Tag, false, err
+	}
+	oktetoLog.Debugf("skipping build: image %s is already built", opts.Tag)
+	return tagWithDigest, true, nil
+}
+
+// SkipBuild returns if build has to be skipped and the tag with digest if found at registry
+func (opts *BuildOptions) SkipBuild(service string) (string, bool, error) {
+	if opts.NoCache {
+		return opts.Tag, false, nil
+	}
+	if !registry.IsOktetoRegistry(opts.Tag) {
+		return opts.Tag, false, nil
+	}
+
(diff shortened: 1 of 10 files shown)


ifbyol:
There is an ongoing refactor of the build command. How does this affect to that? Would it simplify something if we wait until the refactor is done and then, do these changes from that? If this is merge first, @jLopezbarb bear it in mind to avoid to lose something in the refactor

pkg/cmd/build/build.go
```diff
@@ -212,67 +212,172 @@ func translateDockerErr(err error) error {
...
+		return opts.Tag, false, err
+	}
+	oktetoLog.Debugf("skipping build: image %s is already built", opts.Tag)
+	return tagWithDigest, true, nil
 }
 
-// ShouldOptimizeBuild returns if optimization should be applied
-func ShouldOptimizeBuild(image string) bool {
+// SkipBuild returns if build has to be skipped and the tag with digest if found at registry
+func (opts *BuildOptions) SkipBuild(service string) (string, bool, error) {
```
ifbyol, line 347: Sounds a bit odd a function called `SkipBuild` which return something besides an error because it actually doesn't skip the build, it is just checking things. `ShouldSkipBuild` sounds better but doesn't match when returning the tag.

Do we really need the tag + the boolean? Wouldn't be enough returning the tag and the error?

pkg/cmd/build/build.go
```diff
@@ -212,67 +212,172 @@ func translateDockerErr(err error) error {
...
+	return tagWithDigest, true, nil
+}
 
-	return opts
+// checkImageAtRegistry returns if build should be optimized because the image is already at the dev registry
+func (opts *BuildOptions) checkImageAtRegistry() (string, bool, error) {
+	oktetoLog.Debugf("Checking registry for image %s", opts.Tag)
+	tagWithDigest, err := registry.GetImageTagWithDigest(opts.Tag)
+	if err != nil {
+		return opts.Tag, false, err
```
ifbyol, line 340: This is used on `SkipBuild` function which says `returns if build has to be skipped and the tag with digest if found at registry` but here (and in `optimizedGlobalBuild`), we are returning the tag even when there is an error checking in the registry (so we are not sure the tag is in the registry), it is on purpose? If so, I think it is a bit misleading

pkg/cmd/build/build.go
```diff
@@ -212,67 +212,172 @@ func translateDockerErr(err error) error {
...
+	} else if !ok {
+		// check dev registry
+		if tagWithDigest, ok, err := opts.checkImageAtRegistry(); ok {
+			oktetoLog.Debugf("Skipping '%s' build. Image already exists at Okteto Registry", service)
+			return tagWithDigest, true, nil
+		} else if err != nil && err == oktetoErrors.ErrNotFound {
+			// image not found - have to build
+			oktetoLog.Debug("image not found, building image")
+			return "", false, nil
+		}
```
ifbyol, line 378: Minor comment: as each branch has its own return, I think this reads better if you just use single `ifs` instead of `else if`

pkg/cmd/build/build.go
```diff
@@ -212,67 +212,172 @@ func translateDockerErr(err error) error {
...
+			buildOptions.Tag = fmt.Sprintf("%s/%s-%s:%s", targetRegistry, manifestBuildInfo.Name, service, tag)
+		} else {
+			buildOptions.Tag = fmt.Sprintf("%s/%s:%s", targetRegistry, service, tag)
 		}
+	}
+
+	return buildOptions
+}
+
+func overrideManifestBuildOptions(manifestBuildOptions, cmdBuildOptions *BuildOptions) *BuildOptions {
```
AgustinRamiroDiaz, line 261: I love this functional approach. We should just be careful that we are working with pointers, and they point to the same underlying data that the parameters. Sadly go doesn't include any sort of deep copy to make this a `pure function`

ifbyol: you can deep copy pointers doing something like this:

``` 
copyBuildOptions := &BuildOptions{}
*copyBuildOptions = *cmdBuildOptions
```

AgustinRamiroDiaz: The problem is that we have slices inside the struct that will not be copied by that

AgustinRamiroDiaz: There are some third party libraries that do this work of deep copying, some with code generation and others with `reflect`. 

Reading our `go.sum` seems like we have an indirect dependency to github.com/mohae/deepcopy, which seems like a simple solution. This one https://github.com/barkimedes/go-deepcopy on the other hand, is more robust since it takes into account recursive structs and that sort of pointer references

pkg/cmd/build/build.go
jLopezbarb: +1, but I prefer to use switch statements on this cases instead of creating 3 different ifs

pkg/cmd/build/build.go
```diff
@@ -212,67 +212,172 @@ func translateDockerErr(err error) error {
...
 	}
-	opts.OutputMode = setOutputMode(outputMode)
+	tagWithDigest, err := registry.GetImageTagWithDigest(globalReference)
+	if err != nil {
+		return opts.Tag, false, err
+	}
+	oktetoLog.Debugf("skipping build: image %s is already built", opts.Tag)
+	opts.Tag = globalReference
+	return tagWithDigest, true, nil
+}
```
AgustinRamiroDiaz, line 333: 1. We could remove the `bool` in the return since it's the same as `err == nil`. 
2. I think this function should be named something like `GetGlobalBuildTag` since it does not optimize the build. I'm still thinking about the `opts.Tag` being set here
3. The `oktetoLog.Debugf` should be on the caller, since it does not optimize the build

pkg/cmd/build/build.go
AgustinRamiroDiaz: Maybe the name sould be `FindBuiltImageTag`
Also I would remove the `Skipping '%s' build.` since it should be on the callers of this function

pkg/cmd/build/build.go
AgustinRamiroDiaz: What do you think about checking also the dev registry in order to merge the logic from `checkImageAtRegistry`?

teresaromero:
> There is an ongoing refactor of the build command. How does this affect to that? Would it simplify something if we wait until the refactor is done and then, do these changes from that? If this is merge first, @jLopezbarb bear it in mind to avoid to lose something in the refactor

yes, i've been sync with Javi, his PR will be first and i will rebase my changes in order to make this fix over the new code

pkg/cmd/build/build.go
teresaromero: Yes, the idea for this two functions is to return if the build should be skipped (bool) and the tag with digest (string) so its already there and we dont call again the registry. In case of error i guess the best would be to return empty string, false, and the error. Also, i see that maybe i can remove the bool value and just return (string, error) where if string is not empty, the digest has been found so we can work with that.

pkg/cmd/build/build.go
teresaromero: 1. yes, i'm rethinking this part
2. yes, i would change this. i would have to assign this new value out of the func if global tag or tag is found....
3. At the initial approach the logic was merged, but i wanted to separate the global check with the dev check, i will give it a thought
thanks!

teresaromero:
closing this one to open new PR pointing to the new build refactor https://github.com/okteto/okteto/pull/2590

PR #2571, opened 2022-04-20
Added pod name check at okteto up integration

### LINKED ISSUE #2356: e2e test to validate `okteto up` reconnects to an existing dev container
`okteto up` takes some time to create a dev container. The old container needs to be terminated, the new one created, volume is attached, ... If I disconnect from my dev container and I run `okteto up` again, `okteto up` should always reconnect to the existing dev container instead of redeploying it.

We have had regressions here before. We should have an e2e test to validate that two consecutive `okteto up`s don't redeploy the dev container

teresaromero:

Fixes #2356 

## Proposed changes

- Get initial pod and after second up and check the names are the same when reconnect

### DIFF
--- a/integration/up_test.go
+++ b/integration/up_test.go
@@ -50,6 +50,7 @@ import (
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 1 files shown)


teresaromero:
> Can we test it on the PR? Changing e2e tests is a bit complex and can break the release flow

opened https://github.com/okteto/okteto/pull/2572 with new branch name

PR #2578, opened 2022-04-21
feat: add support envvars when okteto up

### LINKED ISSUE #2276: Support for one time envvars on `okteto up`
I want to launch my dev container with custom envvars without modifying my okteto manifest.

Support for `okteto up -e VAR=value`

AdrianPedriza:
Fixes #2276

## Proposed changes

- added a new flag `--env` or `-e` to `okteto up` command. With this new feature, we are able to set environment variables in development container from cmd. Ex: `okteto up -e VAR=value`


cmd/up/up_test.go
```diff
@@ -102,3 +102,86 @@ func Test_printDisplayContext(t *testing.T) {
...
+
+func TestEnvVarIsAddedProperlyToDevContainerWhenIsSetFromCmd(t *testing.T) {
+	var tests = []struct {
+		name                    string
+		dev                     *model.Dev
+		upOptions               *UpOptions
+		expectedNumManifestEnvs int
+	}{
+		{
+			name:                    "Add only env vars from cmd to dev container",
```
jLopezbarb, line 114: Can we add a test with non ascii characters? Would you mind adding a test with some strange characters (f.e. `~$#@` etc)

cmd/up/up.go
```diff
@@ -371,12 +373,41 @@ func loadManifestOverrides(dev *model.Dev, upOptions *UpOptions) error {
...
 
+func setEnvVarsFromCmd(dev *model.Dev, upOptions *UpOptions) error {
+	envVarsIndexFromManifest := make(map[string]int)
+	for i, manifestEnv := range dev.Environment {
+		envVarsIndexFromManifest[manifestEnv.Name] = i
+	}
+
+	for _, v := range upOptions.Envs {
+		kv := strings.SplitN(v, "=", 2)
+		if len(kv) != 2 {
```
jLopezbarb, line 396: Do we support `-e KEY=${NAMESPACE}` expanding env vars?
We also have the notation:
```
environment:
  - KEY
```
Which will create the env var KEY on the container with the value of the local env var
I think its a good use case to add it

AdrianPedriza: OK, I will review both proposals

AdrianPedriza:
What should be the error message when an user gives a non correct format? f.ex: `okteto up -e VAR: value`

jLopezbarb:
> What should be the error message when an user gives a non correct format? f.ex: `okteto up -e VAR: value`

I think that should be trying to get local env var `VAR: value` and if it's empty it should fail with the `invalid variable value '%s': must follow KEY=VALUE format`, which is the default behaviour

rberrelleza:
How will this combine with the deploy and build phases of up? Will the env var be available in build, deploy, and dev? 

And if it's for dev, will it be in all services? (Eg compose with dettached mode)

rberrelleza:
Can we fail if they try to override variables that the cli injects? (E.g OKTETO_NAMESPACE)

jLopezbarb:
> How will this combine with the deploy and build phases of up? Will the env var be available in build, deploy, and dev?


Right now these one time env var are only available for the 'dev' phase, I think it is the best option because if executed in other phases it can leave the services in an undesired state. For example if there are two services with an environment variable 'PASSWORD' that each has its own value when running `okteto up` on one of them and test what happens when the password is wrong may be that the other service does not connect to the one we want to test because the environment variable is not the appropriate one.

> And if it's for dev, will it be in all services? (Eg compose with dettached mode)

It will be available on the container that the user connects to, in the case of the detached mode the one that syncs with the user folder, for the same reason as above

AdrianPedriza:
> Can we fail if they try to override variables that the cli injects? (E.g OKTETO_NAMESPACE)

@rberrelleza , branch updated to not allow okteto built-in env vars from `okteto up` command. At the moment, we will not allow to set env variables starting with `OKTETO_ `or if envvar == `BUILDKIT_HOST` to fit with https://www.okteto.com/docs/cloud/okteto-cli/#built-in-environment-variables . I know there is a lot of env vars manages by okteto (pkg/model(conts.go) so I'm not sure if with the current approach is enough or is not considered good in terms of maintenance. What do you think?

cmd/up/up.go
```diff
@@ -379,12 +380,53 @@ func loadManifestOverrides(dev *model.Dev, upOptions *UpOptions) error {
...
+		}
+	}
+
 	dev.Username = okteto.Context().Username
 	dev.RegistryURL = okteto.Context().Registry
 
 	return nil
 }
 
+func setEnvVarsFromCmd(dev *model.Dev, upOptions *UpOptions) error {
```
AgustinRamiroDiaz, line 395: We could narrow down a bit the parameters, and (this is more of a personal preference) it could be more functional. The signature would look something like `func getEnvVarsFromCmd(devEnv model.Environment, commandEnvVariables []string) (model.Environment, error)`

AdrianPedriza: after discussing this approach we have decided to adopt the more functional approach proposed

cmd/up/up.go
```diff
@@ -379,12 +380,53 @@ func loadManifestOverrides(dev *model.Dev, upOptions *UpOptions) error {
...
+		expandedEnv, err := model.ExpandEnv(varValueToAdd, true)
+		if err != nil {
+			return err
+		}
+
+		if envvIndexManifest, ok := envVarsIndexFromManifest[varNameToAdd]; ok {
+			dev.Environment[envvIndexManifest].Value = expandedEnv
+		} else {
+			dev.Environment = append(dev.Environment, model.EnvVar{Name: varNameToAdd, Value: expandedEnv})
+		}
```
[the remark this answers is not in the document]
AdrianPedriza: after discussing this approach we have decided to adopt the more functional approach proposed

AgustinRamiroDiaz (approved):

AgustinRamiroDiaz (approved):

jLopezbarb (approved):

PR #2588, opened 2022-04-25
feat: start auth flow automatically if required when refresh okteto cloud token

### LINKED ISSUE #2326: `okteto context use <url>` should start authentication flow if the local token is invalid
**Describe the bug**
If I execute `okteto context use https://cloud.okteto.com` and the token I have locally is not valid, I get the following error:

```
Your token is invalid. Please run 'okteto context use https://cloud.okteto.com/' and try again
```

It is suggesting me to execute exactly the same command. If I execute it again, it starts the authentication flow and everything works.

I think it should start directly the auth flow instead of asking the user to execute again the same command. We could print a warning indicating that the local token is not valid and need authentication again

**To Reproduce**

1. Execute `okteto context use https://cloud.okteto.com`
2. Go to the Okteto UI and refresh the token through the settings section
3. Execute `okteto context use https://cloud.okteto.com` again

**Expected behavior**
I think it should start directly the auth flow instead of asking the user to execute again the same command. We could print a warning indicating that the local token is not valid and need authentication again

- OS: macOS
- Okteto CLI Version 1.15.6

**Additional context**

AdrianPedriza:
Fixes #2326

## Proposed changes

- When a user has renewed their token in **okteto cloud** they are required to re-run `okteto context use https://cloud.okteto.com` . This process can be automated to prevent the user from re-executing the same command.
- If you verify that the error is due to an invalid token, you can retry authentication with an empty token to trigger authentication via browser.


cmd/context/create.go
```diff
@@ -227,6 +240,18 @@ func (c *ContextCommand) initOktetoContext(ctx context.Context, ctxOptions *Cont
 	return nil
 }
 
+func AuthenticateToOktetoCluster(ctx context.Context, c *ContextCommand, ctxOptions *ContextOptions) error {
```
jLopezbarb, line 243: I wouldn't add this function, I think that calling a component of the context command and checking the error could be directly on the other function

cmd/context/create.go
```diff
@@ -197,17 +197,30 @@ func (c *ContextCommand) UseContext(ctx context.Context, ctxOptions *ContextOpti
...
 		return err
 	}
-	ctxOptions.Token = user.Token
-	okteto.Context().Token = user.Token
 
 	userContext, err := c.getUserContext(ctx)
 	if err != nil {
-		return err
+		if err.Error() == fmt.Errorf(oktetoErrors.ErrNotLogged, okteto.Context().Name).Error() {
+			oktetoLog.Warning("Your token is invalid")
```
jLopezbarb, line 208:
```suggestion
			oktetoLog.Warning("Your token is invalid. Generating a new one...")
```

cmd/context/create.go
```diff
@@ -197,17 +197,30 @@ func (c *ContextCommand) UseContext(ctx context.Context, ctxOptions *ContextOpti
...
 	if err != nil {
 		return err
 	}
-	ctxOptions.Token = user.Token
-	okteto.Context().Token = user.Token
 
 	userContext, err := c.getUserContext(ctx)
 	if err != nil {
-		return err
+		if err.Error() == fmt.Errorf(oktetoErrors.ErrNotLogged, okteto.Context().Name).Error() {
```
jLopezbarb, line 207: Nit: I would add this into a function called `regenerateToken()`

cmd/context/create.go
```diff
@@ -197,17 +197,30 @@ func (c *ContextCommand) UseContext(ctx context.Context, ctxOptions *ContextOpti
...
+				return err
+			}
+
+			userContext, err = c.getUserContext(ctx)
+			if err != nil {
+				return err
+			}
+		} else {
+			return err
+		}
```
AgustinRamiroDiaz, line 221:
```suggestion
	if err != nil && err.Error() != fmt.Errorf(oktetoErrors.ErrNotLogged, okteto.Context().Name).Error() {
		return err
	} else {
		oktetoLog.Warning("Your token is invalid")
		ctxOptions.Token = ""
		err := AuthenticateToOktetoCluster(ctx, c, ctxOptions)
		if err != nil {
			return err
		}

		userContext, err = c.getUserContext(ctx)
		if err != nil {
			return err
		}
	}
```

cmd/context/create.go
AgustinRamiroDiaz, line 243: I don't think there's need to export this function

ifbyol (commented):
@jLopezbarb If I'm not wrong, `context` is being executed before any other command (at lest before most of them). Would these changes also being executed in those cases?

I mean, if I execute any other okteto command with an invalid token, I get the this error `Your token is invalid. Please run 'okteto context use https://cloud.okteto.com' and try again`. Are this going to change that flow to start the authentication? (I'm just asking, not saying that it should be that way)

jLopezbarb:
> @jLopezbarb If I'm not wrong, `context` is being executed before any other command (at lest before most of them). Would these changes also being executed in those cases?
>
> I mean, if I execute any other okteto command with an invalid token, I get the this error `Your token is invalid. Please run 'okteto context use https://cloud.okteto.com' and try again`. Are this going to change that flow to start the authentication? (I'm just asking, not saying that it should be that way)

I think that's the behaviour we want to achieve. If the token is invalid we are getting a new one and continue with the command execution, isn't it? @pchico83

AdrianPedriza:
@jLopezbarb @AgustinRamiroDiaz check new approach please

ifbyol:
> > @jLopezbarb If I'm not wrong, `context` is being executed before any other command (at lest before most of them). Would these changes also being executed in those cases?
> > I mean, if I execute any other okteto command with an invalid token, I get the this error `Your token is invalid. Please run 'okteto context use https://cloud.okteto.com' and try again`. Are this going to change that flow to start the authentication? (I'm just asking, not saying that it should be that way)
> 
> I think that's the behaviour we want to achieve. If the token is invalid we are getting a new one and continue with the command execution, isn't it? @pchico83

I don't know. It depends on the experience we want. When I opened the ticket, it was mainly because if I execute a `okteto context use <...>` and my token wasn't valid, the CLI was asking me to execute exactly the same command again, which is kind of weird from a user perspective (Why the second time executing exactly the same command would work?)

But, If I'm executing any other command, and I get that message, it **might** be ok, as it was a different command.

What I mean is, we should at least fix the first case because it is the worst from a user perspective, the second one is up to us

ifbyol:
> > > > @jLopezbarb If I'm not wrong, `context` is being executed before any other command (at lest before most of them). Would these changes also being executed in those cases?
> > > > I mean, if I execute any other okteto command with an invalid token, I get the this error `Your token is invalid. Please run 'okteto context use https://cloud.okteto.com' and try again`. Are this going to change that flow to start the authentication? (I'm just asking, not saying that it should be that way)
> > > 
> > > 
> > > I think that's the behaviour we want to achieve. If the token is invalid we are getting a new one and continue with the command execution, isn't it? @pchico83
> > 
> > 
> > I don't know. It depends on the experience we want. When I opened the ticket, it was mainly because if I execute a `okteto context use <...>` and my token wasn't valid, the CLI was asking me to execute exactly the same command again, which is kind of weird from a user perspective (Why the second time executing exactly the same command would work?)
> > But, If I'm executing any other command, and I get that message, it **might** be ok, as it was a different command.
> > What I mean is, we should at least fix the first case because it is the worst from a user perspective, the second one is up to us
> 
> I think is better from a UX standpoint to refresh the token, as we are not bypassing any security barrier and can be done automatically

@AgustinRamiroDiaz  It is not done automatically. It implies a new login of the user in Okteto with the browser flow. So, let say you are working with Cloud. If you hit that error and you are not logged in into GitHub in your browser, you will have to do the whole login flow to get your command executed.

pchico83:
> > > > > @jLopezbarb If I'm not wrong, `context` is being executed before any other command (at lest before most of them). Would these changes also being executed in those cases?
> > > > > I mean, if I execute any other okteto command with an invalid token, I get the this error `Your token is invalid. Please run 'okteto context use https://cloud.okteto.com' and try again`. Are this going to change that flow to start the authentication? (I'm just asking, not saying that it should be that way)
> > > > 
> > > > 
> > > > I think that's the behaviour we want to achieve. If the token is invalid we are getting a new one and continue with the command execution, isn't it? @pchico83
> > > 
> > > 
> > > I don't know. It depends on the experience we want. When I opened the ticket, it was mainly because if I execute a `okteto context use <...>` and my token wasn't valid, the CLI was asking me to execute exactly the same command again, which is kind of weird from a user perspective (Why the second time executing exactly the same command would work?)
> > > But, If I'm executing any other command, and I get that message, it **might** be ok, as it was a different command.
> > > What I mean is, we should at least fix the first case because it is the worst from a user perspective, the second one is up to us
> > 
> > 
> > I think is better from a UX standpoint to refresh the token, as we are not bypassing any security barrier and can be done automatically
> 
> @AgustinRamiroDiaz It is not done automatically. It implies a new login of the user in Okteto with the browser flow. So, let say you are working with Cloud. If you hit that error and you are not logged in into GitHub in your browser, you will have to do the whole login flow to get your command executed.

For commands other than `okteto context` I would return a 401 error and ask the user to run `okteto context` again.
It's a safer approach if you are running in a terminal without a browser

cmd/up/up.go
```diff
@@ -114,17 +114,22 @@ func Up() *cobra.Command {
 			}
 			manifestOpts := contextCMD.ManifestOptions{Filename: upOptions.DevPath, Namespace: upOptions.Namespace, K8sContext: upOptions.K8sContext}
 			oktetoManifest, err := contextCMD.LoadManifestWithContext(ctx, manifestOpts)
-			if err != nil && errors.Is(err, oktetoErrors.ErrManifestNotFound) {
-				if !utils.AskIfOktetoInit(upOptions.DevPath) {
+			if err != nil {
+				if err.Error() == fmt.Errorf(oktetoErrors.ErrNotLogged, okteto.CloudURL).Error() {
```
teresaromero, line 118: minor: can we wrap this into a custom error? :/

teresaromero (approved):

ifbyol (approved):

cmd/up/up.go
AdrianPedriza: correct me if Im wrong but I think that if our errors have a custom format at the moment this is the only way to compare it. **errors.Is()** will return us false, even if both have same message

PR #2589, opened 2022-04-25
feat: add hybrid development

jLopezbarb:

## Proposed changes 
- Adds the ability to create a hybrid development with docker compose
-

### DIFF
--- a/cmd/context/utils.go
+++ b/cmd/context/utils.go
@@ -171,6 +171,10 @@ func LoadManifestWithContext(ctx context.Context, opts ManifestOptions) (*model.
 		dev.Namespace = okteto.Context().Namespace
 		dev.Context = okteto.Context().Name
 	}
+	if m.Deploy != nil && m.Deploy.ComposeSection != nil && m.Deploy.ComposeSection.Stack != nil {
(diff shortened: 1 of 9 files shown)


jmacelroy:
What is the hybrid deployment feature? It seems like a tricky feature and I'm curious what the intended use is. Also, it seems like we'd need to add both unit and integration test to harden the feature. Are tests planned to be added to this PR?

ifbyol:
@jmacelroy This is for the docker extension to be able to have some of the services defined in a docker compose locally an other services in remote and having all of them connected and talking each other

pchico83:
@jLopezbarb @jmacelroy @ifbyol  let's simplify this for now and just support a `x-local: true` field to ignore services to be deployed by okteto. Based on the docker extension traction, we could implement a true hybrid development experience in Q3

pchico83:
> @jLopezbarb @jmacelroy @ifbyol let's simplify this for now and just support a `x-local: true` field to ignore services to be deployed by okteto. Based on the docker extension traction, we could implement a true hybrid development experience in Q3

@jLopezbarb the annotation should be `x-docker: true` instead of `x-local: true`

PR #2590, opened 2022-04-26
Fix global optimization when okteto deploy --build

### LINKED ISSUE #2529: The Global Registry optimization should be always active unless `--no-cache` is set

teresaromero:
Fixes #2529 

## Proposed changes

- when running okteto deploy --build at a pipeline process, check images at registry


cmd/build/v2/build.go
```diff
@@ -242,7 +242,18 @@ func (bc *OktetoBuilder) addVolumeMounts(ctx context.Context, manifest *model.Ma
 	if err != nil {
 		return "", err
 	}
-	buildOptions := build.OptsFromBuildInfo(manifest.Name, svcName, svcBuild, &types.BuildOptions{})
+	buildOptions := build.OptsFromBuildInfo(manifest.Name, svcName, svcBuild, options)
+
+	if build.ShouldOptimizeBuild(buildOptions) {
```
teresaromero, line 247: Not sure we should also apply the optimization at the volume-mounts image when pipeline? this looks for a tag not the hash so i'm afraid the tag will be built but for other base image. maybe i'm having missunderstandings about the volumes here 🤷🏼‍♀️.

cc: @pchico83 , @jLopezbarb

jLopezbarb: I don't think that the built image comes from another image, with checking that it is built I think it would be enough, because if the two images are built one after the other if there is an image and the other will also come from the last built image of the first image (I think I have explained myself terribly but I hope it is understood)😅

teresaromero: understood :D

PR #2601, opened 2022-04-27
fix: manifestv1 was not passing ctx and ns to v2

### LINKED ISSUE #2582: The namespace field in okteto.yml is ignored.
**Describe the bug**
The 'namespace' field in okteto.yml is ignored in okteto 2.1.0.

**To Reproduce**
1. Run `okteto up  -f okteto.yml`.
2. Okteto just ignores the 'namespace' field in okteto.yml files, it always using the default namespace.

**Expected behavior**
Activate my development container in okteto.yml specified namespace.

- OS: macOS
- Version: 12.3.1

jLopezbarb:

Fixes #2582

## Proposed changes
- When loading a manifest v1 okteto was transforming it to a manifest v2 not taking in count the ctx and ns provided from the manifest v1

### DIFF
--- a/pkg/model/manifest.go
+++ b/pkg/model/manifest.go
@@ -293,6 +293,8 @@ func GetManifestV2(manifestPath string) (*Manifest, error) {
 		if devManifest.IsV2 {
 			return devManifest, nil
 		}
+		devManifest.SetManifestDefaultsFromDev()
+
 		oktetoLog.AddToBuffer(oktetoLog.InfoLevel, "Okteto manifest unmarshalled successfully")
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/model/manifest.go
```diff
@@ -1057,3 +1059,15 @@ func (m *Manifest) IsDeployDefault() bool {
 	}
 	return false
 }
+
+// SetManifestDefaultsFromDev sets context and namespace from the dev
+func (m *Manifest) SetManifestDefaultsFromDev() {
```
AgustinRamiroDiaz, line 1064: There's no need to export this

pkg/model/manifest.go
```diff
@@ -1057,3 +1059,15 @@ func (m *Manifest) IsDeployDefault() bool {
...
+	if len(m.Dev) == 1 {
+		for _, devInfo := range m.Dev {
+			m.Context = devInfo.Context
+			m.Namespace = devInfo.Namespace
```
AgustinRamiroDiaz, line 1068: What would happen if `m.Context` and `m.Namespace` were already set? Is it possible?

pkg/model/manifest.go
AgustinRamiroDiaz, line 1064: There's no need to export this

pkg/model/manifest.go
AgustinRamiroDiaz, line 1068: What happens if `m.Context` or `m.Namespace` are set?

AgustinRamiroDiaz: I am testing it on my machine. It looks like `okteto up -n namespace-different-than-the-one-from-manifest` panics

AgustinRamiroDiaz: It's weird because `okteto up -f okteto.yml -n namespace-different-than-the-one-from-manifest` does work 🤔

teresaromero (approved):

PR #2602, opened 2022-04-27
feat: show options to the user if failure

### LINKED ISSUE #2547: Tell me the available options when we  I run 'okteto up service' and I get it wrong
When I run 'okteto up' and I get the service name wrong I get the following error:

```
okteto up backend
 i  Using okteto @ okteto-dev as context
 x  Development container 'backend' doesn't exist
```

I'd like the error message to tell me the available options

jLopezbarb:

Fixes #2547

## Proposed changes
- If the user introduce a wrong dev okteto will show the same error as before but now CLI will display a hint with the options that the user has.

## Screenshoot
Captura de pantalla 2022-04-27 a las 18 07 39

### DIFF
--- a/cmd/utils/dev.go
+++ b/cmd/utils/dev.go
@@ -151,19 +151,27 @@ func GetDevFromManifest(manifest *model.Manifest, devName string) (*model.Dev, e
 	} else if len(manifest.Dev) == 1 {
 		for name, dev := range manifest.Dev {
 			if devName != "" && devName != name {
-				return nil, fmt.Errorf("dev '%s' doesn't exist", devName)
+				return nil, oktetoErrors.UserError{
+					E:    fmt.Errorf("development container '%s' doesn't exist", devName),
+					Hint: fmt.Sprintf("Available options are: [%s]", name),
+				}
 			}
 			return dev, nil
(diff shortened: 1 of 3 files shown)


cmd/utils/dev.go
```diff
@@ -151,19 +151,27 @@ func GetDevFromManifest(manifest *model.Manifest, devName string) (*model.Dev, e
 	} else if len(manifest.Dev) == 1 {
 		for name, dev := range manifest.Dev {
 			if devName != "" && devName != name {
-				return nil, fmt.Errorf("dev '%s' doesn't exist", devName)
+				return nil, oktetoErrors.UserError{
```
jmacelroy, line 154: We should be able to add a test that we are getting the correct error content. This seems important to test as it's part of the UI and very helpful to have.

teresaromero (approved):

cmd/utils/dev.go
jLopezbarb: I realised that we didn't have any test for this function so I added test not just for this PR but for all the cases that don't require the user to add an input

jmacelroy (approved):

Issue #2568, opened 2022-04-19
I cannot cancel `okteto up --detach` with ctrl + c


pchico83:
**Describe the bug**
I cannot cancel okteto up --detach with ctrl + c.

**To Reproduce**

1. Launch a dev environment for https://github.com/okteto/microservices-demo-compose with `okteto up --detach`
2. Wait for some time - I don't have a clear repro here, but it happens very often
3. Press `ctrl +c`
4. Execution hangs

**Expected behavior**
`ctrl + c` exists the `okteto up` session

PR #2604, opened 2022-04-28
Fix CTRL + C when using "--detach"

pchico83:
Fixes https://github.com/okteto/okteto/issues/2568

### DIFF
--- a/cmd/up/logs.go
+++ b/cmd/up/logs.go
@@ -18,7 +18,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
-	"os/signal"
 	"regexp"
 	"text/template"
 	"time"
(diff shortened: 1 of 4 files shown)


cmd/up/logs.go
```diff
@@ -59,9 +56,9 @@ func (up *upContext) showDetachedLogs(ctx context.Context) error {
 		}
 	}()
 	select {
-	case <-stop:
-		oktetoLog.Infof("CTRL+C received, starting shutdown sequence")
-		return oktetoErrors.ErrIntSig
+	case <-ctx.Done():
```
pchico83, line 59: we should only handle `os.Interrupt` at one place. `os.Interrupt` is handled at the beginning of `okteto up`. There, the context is cancelled and we notified on `ctx.Done()`.

cmd/up/up.go
```diff
@@ -598,6 +598,9 @@ func (up *upContext) waitUntilExitOrInterruptOrApply(ctx context.Context) error
 
 func (up *upContext) applyToApps(ctx context.Context) chan error {
 	result := make(chan error, 1)
+	if up.Options.DockerDesktop {
```
pchico83, line 601: This is tricky. When running `okteto up` from the Docker Extension, maybe because they run with a max of goroutines, the cli process crashes if we start the `go tr.App.Watch` goroutines.
Since this isn't needed when running a compose file, I don't creating the watchers, but we should investigate what's the source issue

jLopezbarb (approved):

PR #2610, opened 2022-04-28
fix: deploy a subset of svcs was building everything

jLopezbarb:

Fixes Deploying a subset of services was building everything

## Proposed changes
- Fix: getting dependant services was only getting the first dependant service and not going deeper
- It was not checking the services to build given a list of services
- It was showing twice some messages because okteto stack and okteto deploy was sharing some logic

### DIFF
--- a/cmd/build/v2/services.go
+++ b/cmd/build/v2/services.go
@@ -118,3 +118,24 @@ func (bc *OktetoBuilder) checkImageAtGlobalAndSetEnvs(service string, options *t
 	return true, nil
 
 }
+
+// GetServicesToBuildFromSubset returns the services it has to built because they are not already built from a subset of services
+func (bc *OktetoBuilder) GetServicesToBuildFromSubset(ctx context.Context, manifest *model.Manifest, subset []string) ([]string, error) {
+	for name := range manifest.Build {
+		needsBuild := false
+		for _, toBuildName := range subset {
+			if name == toBuildName {
+				needsBuild = true
+				break
+			}
+		}
+		if !needsBuild {
+			delete(manifest.Build, name)
+		}
+	}
+	svcsToBuild, err := bc.GetServicesToBuild(ctx, manifest)
+	if err != nil {
+		return []string{}, err
+	}
+	return svcsToBuild, err
+}
--- a/cmd/build/v2/services.go
+++ b/cmd/build/v2/services.go
@@ -29,13 +29,21 @@ import (
 )
 
 // GetServicesToBuild returns the services it has to built because they are not already built
-func (bc *OktetoBuilder) GetServicesToBuild(ctx context.Context, manifest *model.Manifest) ([]string, error) {
+func (bc *OktetoBuilder) GetServicesToBuild(ctx context.Context, manifest *model.Manifest, svcToDeploy []string) ([]string, error) {
 	buildManifest := manifest.Build
+
+	svcToDeployMap := map[string]bool{}
+	for _, svc := range svcToDeploy {
+		svcToDeployMap[svc] = true
+	}
 
 	// check if images are at registry (global or dev) and set envs or send to build
 	toBuild := make(chan string, len(buildManifest))
 	g, _ := errgroup.WithContext(ctx)
 	for service := range buildManifest {
+		if _, ok := svcToDeployMap[service]; !ok {
+			continue
+		}
 		svc := service
 		g.Go(func() error {
 			return bc.checkServicesToBuild(svc, manifest, toBuild)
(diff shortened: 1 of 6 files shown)


cmd/build/v2/services.go
```diff
@@ -118,3 +118,24 @@ func (bc *OktetoBuilder) checkImageAtGlobalAndSetEnvs(service string, options *t
...
+func (bc *OktetoBuilder) GetServicesToBuildFromSubset(ctx context.Context, manifest *model.Manifest, subset []string) ([]string, error) {
+	for name := range manifest.Build {
+		needsBuild := false
+		for _, toBuildName := range subset {
+			if name == toBuildName {
+				needsBuild = true
+				break
+			}
+		}
+		if !needsBuild {
```
ifbyol, line 132: I think you don't need this check. In the only place where the flag is set to `true` you are breaking the loop, so if the code arrives to this point, this flag is `false`

cmd/build/v2/services.go
```diff
@@ -118,3 +118,24 @@ func (bc *OktetoBuilder) checkImageAtGlobalAndSetEnvs(service string, options *t
...
+			delete(manifest.Build, name)
```
ifbyol, line 133: Is it safe to manipulate the manifest removing things?

pkg/cmd/stack/deploy.go
```diff
@@ -698,19 +701,42 @@ func validateDefinedServices(s *model.Stack, servicesToDeploy []string) error {
 	return nil
 }
 
-func addDependentServicesIfNotPresent(ctx context.Context, s *model.Stack, options *StackDeployOptions, c kubernetes.Interface) {
-	added := make([]string, 0)
-	for _, svcToDeploy := range options.ServicesToDeploy {
+// AddDependentServicesIfNotPresent adds dependands services to deploy
+func AddDependentServicesIfNotPresent(ctx context.Context, s *model.Stack, svcsToDeploy []string, c kubernetes.Interface) []string {
+	initialSvcsToDeploy := svcsToDeploy
+	svcsToDeploy = addDependentServices(ctx, s, svcsToDeploy, c)
```
ifbyol, line 707: Nit: maybe it is safer to create another variable instead of using the one passed as parameter `svcsToDeploy`

cmd/build/v2/services.go
jLopezbarb: This break statement breaks the inner loop because we need to check if the services that needs to be built are in the manifest.Build section so we need two for statements

cmd/build/v2/services.go
jLopezbarb: We should be ignoring them, I don't like to delete them so I'll check if I can do it in another way so we don't delete them

cmd/build/v2/services.go
ifbyol: oh, yes, there were two nested loops, I didn't see it well, thanks for the explanation

teresaromero (approved):

pkg/cmd/stack/deploy.go
```diff
@@ -698,19 +701,48 @@ func validateDefinedServices(s *model.Stack, servicesToDeploy []string) error {
...
 	}
-	if len(added) > 0 {
-		oktetoLog.Warning("The following services need to be deployed because the services passed as arguments depend on them: [%s]", strings.Join(added, ", "))
+	if initialLength != len(svcsToDeploy) {
+		return addDependentServices(ctx, s, svcsToDeploy, c)
+	}
+	return svcsToDeploy
+}
+
+func getAddedSvcs(initialSvcsToDeploy, svcsToDeployWithDependencies []string) []string {
```
AgustinRamiroDiaz, line 731: We could reduce the reading (and computational) complexity of this if we transform the `initialSvcsToDeploy` into a set

cmd/deploy/deploy.go
```diff
@@ -416,7 +420,7 @@ func (dc *DeployCommand) RunDeploy(ctx context.Context, deployOptions *Options)
 	return err
 }
 
-func setDeployOptionsValuesFromManifest(deployOptions *Options, cwd string) {
+func setDeployOptionsValuesFromManifest(ctx context.Context, deployOptions *Options, cwd string, c kubernetes.Interface) {
```
AgustinRamiroDiaz, line 423: Nit: I don't like the idea of this function having these new parameters. It's not important for this feature, but we could try to pass in the dependant services instead of a whole kubernetes interface

pkg/cmd/stack/deploy.go
```diff
@@ -698,19 +701,48 @@ func validateDefinedServices(s *model.Stack, servicesToDeploy []string) error {
...
-				added = append(added, dependentSvc)
+			if !isSvcToBeDeployed(svcsToDeploy, dependentSvc) && !isSvcRunning(ctx, s.Services[dependentSvc], s.Namespace, dependentSvc, c) {
+				svcsToDeploy = append(svcsToDeploy, dependentSvc)
 			}
 		}
 	}
-	if len(added) > 0 {
-		oktetoLog.Warning("The following services need to be deployed because the services passed as arguments depend on them: [%s]", strings.Join(added, ", "))
+	if initialLength != len(svcsToDeploy) {
+		return addDependentServices(ctx, s, svcsToDeploy, c)
```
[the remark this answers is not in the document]
jLopezbarb: We return an error for cyclic dependencies when validating the stack

cmd/deploy/deploy.go
```diff
@@ -460,6 +464,9 @@ func setDeployOptionsValuesFromManifest(deployOptions *Options, cwd string) {
 				deployOptions.servicesToDeploy = append(deployOptions.servicesToDeploy, service)
 			}
 		}
+		if len(deployOptions.Manifest.Deploy.ComposeSection.ComposesInfo) > 0 {
+			deployOptions.Manifest.Deploy.ComposeSection.ComposesInfo[0].ServicesToDeploy = stack.AddDependentServicesIfNotPresent(ctx, deployOptions.Manifest.Deploy.ComposeSection.Stack, deployOptions.servicesToDeploy, c)
```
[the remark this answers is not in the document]
jLopezbarb: It does, We have a merged stack and need a list of services to deploy which we get from getting all the services to deploy on each of the deploy manifest we are not checking which services comes from which stack.
I thought to set the services on the correct stack but I couldn't do it because it needs to read again all the stacks and it doesn't provide any user experience

PR #2612, opened 2022-04-28
Fix compose extended notation

AgustinRamiroDiaz:
Signed-off-by: Agustin Ramiro Diaz <agustin.ramiro.diaz@gmail.com>

Fixes:
Only deploy endpoints and volumes from servicesToDeploy 

## Proposed changes

- Add VolumesToDeploy option
- Add EndpointsToDeploy option (Pending)

### DIFF
--- a/cmd/deploy/deploy.go
+++ b/cmd/deploy/deploy.go
@@ -65,6 +65,8 @@ type Options struct {
 	Build            bool
 	Dependencies     bool
 	servicesToDeploy []string
+	volumesToDeploy  []string
+	// endpointsToDeploy []string
 
 	Repository string
 	Branch     string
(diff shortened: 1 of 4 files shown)


AgustinRamiroDiaz:
@jLopezbarb I added `VolumesToDeploy` as a deploy option and as a stack deploy option. I haven't tested it yet, but I feel confident about this change.

The thing missing is to do the same with `EndpointsToDeploy`. Is not as easy since the real segregation has to be done at the `Rules` level of the `Endpoint`. We could copy the `Endpoint` struct entirely but that could lead to inconsistencies.

cmd/deploy/deploy.go
```diff
@@ -505,25 +507,20 @@ func onlyDeployEndpointsFromServicesToDeploy(endpoints model.EndpointSpec, servi
 	}
 }
 
-func onlyDeployVolumesFromServicesToDeploy(stack *model.Stack, servicesToDeploy map[string]bool) {
-
-	volumesToDeploy := map[string]bool{}
+func getVolumesToDeployFromServicesToDeploy(stack *model.Stack, servicesToDeploy map[string]bool) []string {
```
jLopezbarb, line 510: Can we get them on the stack package? I think that if we get here it can cause some problems when running `okteto stack deploy`. That way we don't need to create a new option

pkg/cmd/stack/destroy_test.go
```diff
@@ -277,3 +278,74 @@ func Test_destroyJobs(t *testing.T) {
 		})
 	}
 }
+
+func Test_onlyDeployVolumesFromServicesToDeploy(t *testing.T) {
```
jLopezbarb, line 282: shouldn't this be on stack/deploy_test.go

AgustinRamiroDiaz: Yes, thanks. I misread

jLopezbarb:
Nice PR. Deploy code looks cleaner now!!

jLopezbarb (approved):

PR #2617, opened 2022-04-29
fix: deploying compose on different path

jLopezbarb:
## Proposed changes
- Fix deploying from compose info on manifest should load every compose file from the path it's defined
-


AgustinRamiroDiaz (approved):
Not for this pr, but is there a way to centralize all this workdir logic so we don't have it sharded across various functions?

jLopezbarb:
We should be creating a package just for using the chdir

PR #2618, opened 2022-04-29
fix: paths on win

jLopezbarb:

Fixes paths on windows unit tests

## Proposed changes
- Change paths from unix style to non os dependent
-

### DIFF
--- a/pkg/model/stack_serializer_test.go
+++ b/pkg/model/stack_serializer_test.go
@@ -860,6 +860,8 @@ func Test_validateCommandArgs(t *testing.T) {
 func Test_validateVolumesUnmarshalling(t *testing.T) {
 	wd, _ := os.Getwd()
 	relativePathExpanded := filepath.Join(wd, "test_volume_relative_path_found")
+	pandaAbsolute := filepath.Clean("/var/lib/redpanda")
+	dockerdaemonPath := filepath.Clean("/var/run/docker.sock")
 	tests := []struct {
 		name                 string
 		manifest             []byte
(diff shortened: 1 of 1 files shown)


cmd/deploy/deploy.go
```diff
@@ -459,10 +459,6 @@ func setDeployOptionsValuesFromManifest(ctx context.Context, deployOptions *Opti
 			deployOptions.Manifest.Deploy.ComposeSection.ComposesInfo[0].ServicesToDeploy = deployOptions.servicesToDeploy
 		}
 	}
-
-	if len(deployOptions.servicesToDeploy) == 0 {
```
jLopezbarb, line 463: @AgustinRamiroDiaz Do you know why we were deploying name if services to deploy were empty

AgustinRamiroDiaz: I think it was deploys that are NOT docker-compose

pkg/model/stack_serializer.go
```diff
@@ -1138,7 +1139,16 @@ func (v *StackVolume) UnmarshalYAML(unmarshal func(interface{}) error) error {
 		return err
 	}
 
-	parts := strings.SplitN(raw, ":", 2)
+	parts := strings.Split(raw, ":")
+	if runtime.GOOS == "windows" {
+		if len(parts) >= 3 {
```
AgustinRamiroDiaz, line 1144: I think what you've done covers 99.99% of cases, but to bear in mind there could also be:
- `len(parts) == 4` : this should be 2 absolute windows paths separated by a `:`. Example `C:\somewhere\local:C:\somewhere\in\the\container`
- `len(parts) == 3` : we don't know if the absolute path is in the host or in the container. We could see if Windows allows drive names other than letters, and then evaluate a solution from there

AgustinRamiroDiaz (approved):

pkg/model/stack_serializer.go
jLopezbarb: - len(parts) == 4 is not supported. Right now we only support linux machines
- len(parts) == 3 as it's not supported on the container we can always infer that is on the host

cmd/build/v2/services.go
```diff
@@ -63,8 +56,15 @@ func (bc *OktetoBuilder) GetServicesToBuild(ctx context.Context, manifest *model
...
 
+	svcToDeployMap := map[string]bool{}
+	for _, svc := range svcToDeploy {
+		svcToDeployMap[svc] = true
+	}
 	svcsToBuildList := []string{}
 	for svc := range toBuild {
+		if _, ok := svcToDeployMap[svc]; len(svcToDeploy) > 0 && !ok {
+			continue
+		}
```
AgustinRamiroDiaz, line 67: @jLopezbarb Why are we doing this? I've traced [this issue](https://github.com/okteto/okteto/issues/2795) to this line and I'm not sure why is this here

PR #2622, opened 2022-04-30
Support for divert in the deploy section of the okteto manifest

pchico83:

This PR adds support for an experimental feature to create `divert` objects as part of your dev environment.
The syntax is like this:

```
deploy:
  compose:
    file: docker-compose.yml
    services:
      - api
  divert:
    namespace: staging
    service: api
    deployment: api
```
This is an alpha feature and won't be documented for now


cmd/deploy/deploy.go
```diff
@@ -379,6 +391,9 @@ func (dc *DeployCommand) RunDeploy(ctx context.Context, deployOptions *Options)
 	oktetoLog.SetStage("")
 
 	if err != nil {
+		if err.Error() == "interrupt signal received" {
```
teresaromero, line 394: can we use if err is ??https://github.com/okteto/okteto/blob/58682ebd38516583db8fe373582a8bcd628efd24/pkg/errors/errors.go#L127

cmd/deploy/deploy.go
```diff
@@ -508,63 +523,41 @@ func (dc *DeployCommand) deploy(ctx context.Context, opts *Options) error {
...
+		exit <- nil
+	}()
+
+	select {
+	case <-stop:
+		oktetoLog.Infof("CTRL+C received, starting shutdown sequence")
+		sp := utils.NewSpinner("Shutting down...")
+		sp.Start()
+		defer sp.Stop()
+		dc.Executor.CleanUp(errors.New("interrupt signal received"))
```
teresaromero, line 551: same here, we can use the error we already have?

pkg/model/manifest.go
```diff
@@ -638,10 +647,37 @@ func Read(bytes []byte) (*Manifest, error) {
...
 	}
+
+	if err := manifest.validate(); err != nil {
+		return nil, err
+	}
 	manifest.Manifest = bytes
 	return manifest, nil
 }
 
+func (m *Manifest) validate() error {
```
teresaromero, line 658: why don't we use `validateDivert` directly? this is for adding other validations in the future?

cmd/deploy/deploy.go
```diff
@@ -589,10 +582,42 @@ func (dc *DeployCommand) deployStack(ctx context.Context, opts *Options) error {
...
+	sp := utils.NewSpinner(fmt.Sprintf("Diverting namespace %s...", opts.Manifest.Deploy.Divert.Namespace))
+	sp.Start()
+	defer sp.Stop()
+
+	c, _, err := dc.K8sClientProvider.Provide(okteto.Context().Cfg)
+	if err != nil {
+		return err
+	}
+
+	result, err := c.NetworkingV1().Ingresses(opts.Manifest.Deploy.Divert.Namespace).List(ctx, metav1.ListOptions{})
```
jLopezbarb, line 601: Are we supporting this for k8s clusters < 1.19?

jLopezbarb: Maybe you can use https://github.com/okteto/okteto/blob/master/pkg/k8s/ingresses/crud.go#L97 to make it compatible with them

jmacelroy: We don't need to support <1.19 anymore.

jmacelroy:
Is there a way to do this without it being in main? I expect there to be quite a bit of churn and doing it in main can be distracting and slow non-experimental sustaining work down. Also, an advantage could be that we work quicker on experimenting and could have less stringent requirements for getting a merge down to an experimental branch and release until we have solidified what we want. At which point we can do a focus on getting it tested, validated and into main.

pchico83:
> Is there a way to do this without it being in main? I expect there to be quite a bit of churn and doing it in main can be distracting and slow non-experimental sustaining work down. Also, an advantage could be that we work quicker on experimenting and could have less stringent requirements for getting a merge down to an experimental branch and release until we have solidified what we want. At which point we can do a focus on getting it tested, validated and into main.

I like this proposal. In general, we need to start thinking about how to reduce complexity in the cli, and how we can accelerate. Let's start with the docker extension logic https://github.com/okteto/okteto/pull/2648 whose binaries can be created in the extension image itself.

For this particular one, I have been using a feature branch for a few weeks now, but there is already a chunk of 30 devs who will start using it and this change also requires updates in the installer image and teams clusters.
The logic is quite isolated too.
I think a feature branch will imply more cons than pros at this point in time.

cmd/deploy/deploy.go
pchico83: updated!

pkg/model/manifest.go
pchico83: yes, that's the idea. we could already move logic here, but I didn't do it on this PR

ifbyol (approved):

jLopezbarb (approved):

PR #2649, opened 2022-05-05
fix: nil pointer when okteto manifest is created

### LINKED ISSUE #2628: got a nilref when creating a new manifest
**Describe the bug**
image

**To Reproduce**

1. Clone https://github.com/rberrelleza/angular-ngrx-nx-realworld-example-app/
2. Run okteto up
3. Say `y` on the stignore step

- OS: macOS
- Version 2.2.1

AdrianPedriza:
Fixes #2628

## Proposed changes

- Set default values in manifest.dev to avoid nil pointer
- update local folder path to use a rel path if an abs path is given

### DIFF
--- a/cmd/up/up.go
+++ b/cmd/up/up.go
@@ -132,8 +132,6 @@ func Up() *cobra.Command {
 				if err != nil {
 					return err
 				}
-			} else if err != nil {
-				return err
 			}
 			wd, err := os.Getwd()
 			if err != nil {
--- a/cmd/up/up.go
+++ b/cmd/up/up.go
@@ -376,6 +376,23 @@ func LoadManifestWithInit(ctx context.Context, k8sContext, namespace, devPath st
 	if err != nil {
 		return nil, err
 	}
+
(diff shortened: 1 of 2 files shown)


AgustinRamiroDiaz (approved):
Which property was raising the nilref? Can we change so it's not a pointer?

cmd/up/up.go
```diff
@@ -376,6 +374,23 @@ func LoadManifestWithInit(ctx context.Context, k8sContext, namespace, devPath st
...
+		manifest.Context = okteto.Context().Name
+	}
+	manifest.IsV2 = true
+	for devName, d := range manifest.Dev {
+		if err := d.SetDefaults(); err != nil {
+			return nil, err
+		}
+		d.Name = devName
+		d.Namespace = manifest.Namespace
+		d.Context = manifest.Context
```
ifbyol, line 391: Not sure if possible, but should we check if they have a different value than the one in the manifest?

ifbyol (approved):

PR #2652, opened 2022-05-06
Support multile envs at stack

teresaromero:

Fixes https://github.com/okteto/app/issues/3781

## Proposed changes

- when expanding multiline envs, the string resulting of that for a stack manifest was breaking the yaml stucture and so the stack was not unmarshalled
- replace the ExpandEnv method for supporting multiline ENVs
- new method will transform into yaml3 nodes, check if there is a node with env and expand the value, then marshal again into a string so it can be unmarshalled to a stack
- tested multiple line vars with okteto manifest, and these are parsed ok - no changes to apply into okteto manifest

### DIFF
--- a/pkg/model/stack.go
+++ b/pkg/model/stack.go
@@ -14,6 +14,7 @@
 package model
 
 import (
+	"bytes"
 	"errors"
 	"fmt"
 	"net/url"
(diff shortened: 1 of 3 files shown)


pkg/model/stack.go
```diff
@@ -210,14 +212,59 @@ const (
 	DependsOnServiceCompleted DependsOnCondition = "service_completed_successfully"
 )
 
+func expandEnvScalarNode(node yaml3.Node) (yaml3.Node, error) {
```
jLopezbarb, line 215: Could we use a pointer so we don't copy the node on every call?

jLopezbarb:
Maybe we could create for now a stack_expander file to have this and the test not on the same file as the other stack function

ifbyol (approved):

jLopezbarb (approved):

PR #2653, opened 2022-05-06
fix: avoid expand env if are created on a deployment pipeline

AdrianPedriza:
Fixes #3738

## Proposed changes

- avoid substitution of environment variables from `deploy` section in manifest, they will be substituted when executing the command
- adapt test to this new approach

### DIFF
--- a/cmd/build/v2/environment_test.go
+++ b/cmd/build/v2/environment_test.go
@@ -108,58 +108,6 @@ func Test_SetServiceEnvVars(t *testing.T) {
 	}
 }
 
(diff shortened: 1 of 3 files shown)


AgustinRamiroDiaz (approved):
Can we do the same with our custom created envvars like `OKTETO_BUILD_NAME_IMAGE`?

AdrianPedriza:
> OKTETO_BUILD_NAME_IMAGE

yes, its value will be substituted when the command is executed. It is not necessary to replace it beforehand.

jLopezbarb (approved):

PR #2656, opened 2022-05-06
fix: avoid use compose files if manifest is v1

### LINKED ISSUE #2575: `okteto up` fails when a `docker-compose.yaml` file is present
**Describe the bug**
Running `okteto up` with a `docker-compose.yaml` file in the same directory fails with `Error checking image at registry`

**To Reproduce**

1. Define a simple docker-compose.yaml
```yaml
x-common-args: &common-args
  HASH: "${HASH:-local}"

version: '3.8'
services:
  installer:
    image: "installer:${HASH:-local}"
    build:
      context: .
      dockerfile: build/installer.Dockerfile
  base:
    image: "base:${HASH:-local}"
    depends_on: [installer]
    build:
      context: build
      dockerfile: base.Dockerfile
  server:
    image: "server:${HASH:-local}"
    depends_on: [installer, base]
    build:
      context: components
      dockerfile: ../build/server.Dockerfile
      args:
        <<: *common-args
```
3. Define an `okteto.yaml` in the same directory
```yaml
name: server
imagePullPolicy: IfNotPresent
container: server
command: [bash]
sync:
  - .:/root
```
3. Try `okteto up`
> Error checking image at registry base:local: error getting image tag digest: GET https://index.docker.io/v2/library/base/manifests/local: UNAUTHORIZED: authentication required; [map[Action:pull Class: Name:library/base Type:repository]]

**Expected behavior**
The `docker-compose.yaml` would be ignored, the `okteto up` command works just fine if the `docker-compose.yaml` file is deleted

- OS: macOS
- Version: 2.1.0

**Additional context**
The `docker-compose.yaml` is simply being used as a "build definition" file, so that all the docker images can be easily built using a single command: `docker-compose build`. There are no other "compose" features being used in this file. Additionally, the `base` image is used as a stage in building the `server` image.

AdrianPedriza:

Fixes #2575

## Proposed changes

- create new method to manage manifest v1 that will not merge data from docker-compose files if present

### DIFF
--- a/pkg/model/manifest.go
+++ b/pkg/model/manifest.go
@@ -360,7 +360,7 @@ func GetManifestV2(manifestPath string) (*Manifest, error) {
 	}
 
 	if manifest != nil && manifest.IsV2 {
-		return nil, nil
+		return manifest, nil
 	}
 
 	inferredManifest, err := GetInferredManifest(cwd)
--- a/pkg/model/manifest.go
+++ b/pkg/model/manifest.go
@@ -1013,7 +1013,7 @@ func (m *Manifest) WriteToFile(filePath string) error {
 		}
 	}
 
-	doc = m.reorderDocFields(doc)
+	m.reorderDocFields(&doc)
 
 	buffer := bytes.NewBuffer(nil)
 	encoder := yaml3.NewEncoder(buffer)
@@ -1032,7 +1032,7 @@ func (m *Manifest) WriteToFile(filePath string) error {
 }
 
 // reorderDocFields orders the manifest to be: name -> build -> deploy -> dependencies -> dev
-func (*Manifest) reorderDocFields(doc yaml3.Node) yaml3.Node {
+func (*Manifest) reorderDocFields(doc *yaml3.Node) {
 	contentCopy := []*yaml3.Node{}
 	nodes := []int{}
 	nameDefinitionIdx := getDocIdx(doc.Content, "name")
@@ -1118,7 +1118,6 @@ func (*Manifest) reorderDocFields(doc yaml3.Node) yaml3.Node {
 		}
 	}
 	doc.Content = contentCopy
-	return doc
 }
 
 func getDocIdxWithPrior(contents []*yaml3.Node, value string) int {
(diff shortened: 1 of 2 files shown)


pkg/model/manifest.go
```diff
@@ -263,6 +263,51 @@ func NewDeployInfo() *DeployInfo {
...
+			}
+		}
+		manifest.Filename = path
+		return manifest, nil
+	} else if manifestPath != "" && pathExistsAndDir(manifestPath) {
+		cwd = manifestPath
+	}
+
+	var devManifest *Manifest
+	if oktetoPath := getFilePath(cwd, oktetoFiles); oktetoPath != "" {
```
jLopezbarb, line 294: We already have this logic in `GetManifestV2`. Can we called it `getManifestFromOktetoFile`

pkg/model/manifest.go
```diff
@@ -263,6 +263,51 @@ func NewDeployInfo() *DeployInfo {
 	}
 }
 
+//GetManifestV1 gets a manifest from a path or search for the files to generate it
+func GetManifestV1(manifestPath string) (*Manifest, error) {
+	cwd, err := os.Getwd()
```
jLopezbarb, line 268: Same logic as GetManifestV2. Can you encapsulate this into a function called `loadManifestFromPathFile` or something similar?

pchico83 (approved):

teresaromero (approved):

pkg/model/manifest.go
```diff
@@ -286,24 +299,68 @@ func GetManifestV2(manifestPath string) (*Manifest, error) {
 		}
 		manifest.Filename = path
 		return manifest, nil
-	} else if manifestPath != "" && pathExistsAndDir(manifestPath) {
+	}
+
+	return nil, nil
```
jLopezbarb, line 304: It's a good practice to return errors in Go according to  [Effective Go](https://golang.org/doc/effective_go.html#errors), I think that you should have `return nil, CustomErr` and then check if the error is the one that you are returning. [Go blog]( https://go.dev/blog/error-handling-and-go) also has a good understanding on how to handle errors

jLopezbarb (approved):

PR #2657, opened 2022-05-06
Fix: error message missing filename

### LINKED ISSUE #2627: error message is missing file name
**Describe the bug**
The error message is missing the file name is looking for
image

**To Reproduce**

1. Create a repo with only a Dockerfile
2. Run okteto up

**Expected behavior**
I was expecting the error message to include a file name

- OS: macOS
- Version 2.2.1

**Additional context**

AdrianPedriza:

Fixes #2627

## Proposed changes

- use DefaultManifest const if a dev file has not been provided

### DIFF
--- a/cmd/utils/dev.go
+++ b/cmd/utils/dev.go
@@ -367,9 +367,6 @@ func AskForOptions(options []string, label string) (string, error) {
 
 //AskIfOktetoInit asks if okteto init should be executed
 func AskIfOktetoInit(devPath string) bool {
-	if devPath == "" {
-		devPath = DefaultManifest
-	}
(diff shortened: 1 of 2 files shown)


cmd/utils/dev.go
```diff
@@ -367,6 +367,9 @@ func AskForOptions(options []string, label string) (string, error) {
 
 //AskIfOktetoInit asks if okteto init should be executed
 func AskIfOktetoInit(devPath string) bool {
+	if devPath == "" {
+		devPath = DefaultManifest
```
jmacelroy, line 371: Is that the name of the one we'll create if the user selects "y".

jLopezbarb: Yes, When we call `okteto init` without adding the file it will create it with the name of `DefaultManifest`

jLopezbarb (approved):

PR #2660, opened 2022-05-09
fix: skip questions on docker extension

jLopezbarb:

## Proposed changes
- If no manifest is found return err
- If no dev section can be inferred it should guide the user to create one

### DIFF
--- a/cmd/up/up.go
+++ b/cmd/up/up.go
@@ -121,6 +121,10 @@ func Up() *cobra.Command {
 					return err
 				}
 
+				if upOptions.DockerDesktop {
+					return err
+				}
+
 				if !utils.AskIfOktetoInit(upOptions.DevPath) {
 					return err
 				}
@@ -143,6 +147,12 @@ func Up() *cobra.Command {
 			os.Setenv(model.OktetoNameEnvVar, oktetoManifest.Name)
 
 			if len(oktetoManifest.Dev) == 0 {
+				if upOptions.DockerDesktop {
+					return oktetoErrors.UserError{
+						E:    errors.New("compose file doesn't have volume to synchronize"),
+						Hint: "Add volumes mounts to your services syncrhonizing your local",
+					}
+				}
 				oktetoLog.Warning("okteto manifest has no 'dev' section.")
 				answer, err := utils.AskYesNo("Do you want to configure okteto manifest now? [y/n]")
 				if err != nil {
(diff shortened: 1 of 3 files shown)


ifbyol (approved):

jLopezbarb:
If no dev section is defined we show the logs of the full compose
Captura de pantalla 2022-05-09 a las 12 10 52
If dev section is defined we only show logs of the development services
Captura de pantalla 2022-05-09 a las 12 12 27

pchico83:
> If no dev section is defined we show the logs of the full compose Captura de pantalla 2022-05-09 a las 12 10 52 If dev section is defined we only show logs of the development services Captura de pantalla 2022-05-09 a las 12 12 27

Let's always show logs for all services, it gives a more consistent behavior

cmd/up/up.go
```diff
@@ -142,7 +142,7 @@ func Up() *cobra.Command {
 			}
 			os.Setenv(model.OktetoNameEnvVar, oktetoManifest.Name)
 
-			if len(oktetoManifest.Dev) == 0 {
+			if len(oktetoManifest.Dev) == 0 && !upOptions.DockerDesktop {
```
rberrelleza, line 145: @jLopezbarb can we make this more general? We need the same behavior for the vscode extension

pchico83: @rberrelleza this one in particular only applies to the docker extension, and it's only done on this feature branch.
what would you need for the vscode extension?

jmacelroy: What happens when this feature is merged to main?

pchico83: @jmacelroy we will need to evaluate how to do it. There are two options:
- Move this logic to the Docker Extension itself
- Make this logic more generic and implement the detach flag

rberrelleza: I think that making a generic option for "don't ask any questions and just assume defaults" is very useful (we need that for the okteto vscode extension for the same reason than docker desktop).  But I think that's a different requirement from a "detached mode". To me, the "ask no questions" mode sounds like a requirement for detached, but they not need to be tightly coupled. You might still want the process at the forefront, just without requiring any input.

pchico83: Makes sense. I would do that on a separate PR and analyze where we are prompting for questions in the current codebase.

PR #2695, opened 2022-05-16
fix: not show two times same error

### LINKED ISSUE #2635: CLI is printing twice the same error when executing a command and the output is json
**Describe the bug**
If I have a failed command during the execution of the commands defined in my manifest, I have a duplicated error message with json format

```
{"level":"error","stage":"exit 1","message":"error executing command 'exit 1': exit status 1","timestamp":1651651721}
{"level":"info","stage":"done","message":"EOF","timestamp":1651651721}
{"level":"error","stage":"Internal server error","message":"x  Error executing command 'exit 1': exit status 1","timestamp":1651651721}
```

With other output formats like `plain` or `tty`, we are just displaying one error

In plain:

```
COLORTERM=truecolor
INFO: Running exit 1
ERROR: Error executing command 'exit 1': exit status 1
    Update the 'deploy' section of your okteto manifest and try again
```

In tty:

```
COLORTERM=truecolor
 i  Running exit 1
 x  Error executing command 'exit 1': exit status 1
    Update the 'deploy' section of your okteto manifest and try again
```

As the 2 messages in json has a different stage, this is causing a weird behavior in the UI because it is creating an unnecessary stage `Internal Server Error` to print a message that was already printed in the previous stage.

Captura de Pantalla 2022-05-04 a las 10 22 42

It seems this line is the one printing the last message: https://github.com/okteto/okteto/blob/master/main.go#L145

**To Reproduce**

1. Define a simple okteto manifest with just this command `exit 1`
2. Deploy that okteto manifest from the UI

**Expected behavior**
If the error was provoked by the command execution, the stage `Internal Server Error` should not be shown as the error appears in the command section

- Version: `2.2.1`

**Additional context**

jLopezbarb:

Fixes #2635 

## Proposed changes
- CLI will only print one failure step on json

### DIFF
--- a/pkg/log/json.go
+++ b/pkg/log/json.go
@@ -23,15 +23,10 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-const (
-	defaultErrorStage = "Internal server error"
-)
-
 //JSONWriter writes into a JSON terminal
 type JSONWriter struct {
-	out              *logrus.Logger
-	file             *logrus.Entry
-	hasAlreadyFailed bool
+	out  *logrus.Logger
+	file *logrus.Entry
 }
 
 type jsonMessage struct {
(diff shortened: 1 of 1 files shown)


pkg/log/json.go
```diff
@@ -29,8 +29,9 @@ const (
 
 //JSONWriter writes into a JSON terminal
 type JSONWriter struct {
-	out  *logrus.Logger
-	file *logrus.Entry
+	out              *logrus.Logger
+	file             *logrus.Entry
+	hasAlreadyFailed bool
```
ifbyol, line 34: This is a little bit hacky. It is weird that a logger has a logic to print or not an error based on if it has already printed another one (could be a different one) and only if the format is json. We are coupling too much this logger with the installer but it can be used outside of it.

iirc, the original issue is that an error is printed twice in the installer. The last one is in the default error stage. Another option could be to remove the line 204 and not include the default error stage if empty.

jmacelroy: +1 this seems dangerous.

ifbyol (approved):

PR #2697, opened 2022-05-17
fix: check if has access to namespace when working with vanilla cluster

### LINKED ISSUE #2634: context list showing default namespace instead of the configured one
I hit this while using latest `okteto` 2.2.1 and trying to work around #2582:

```sh
$ okteto context use admin@kubernetes --namespace hello-world-dev
 ✓  Using context hello-world-dev @ admin@kubernetes
# seems to have worked (although the message is a bit confusing,
# it's not immediately obvious it's "namespace @ user@cluster")

# but then
$ okteto context list
Name                Namespace  Builder  Registry
admin@kubernetes *  default    docker
# also not sure what that star (*) means

# although
$ cat ~/.okteto/context/config.json 
{
        "contexts": {
                "admin@kubernetes": {
                        "name": "admin@kubernetes",
                        "namespace": "hello-world-dev"
                }
        },
        "current-context": "admin@kubernetes"
}

# and "okteto up" works as expected
```

AdrianPedriza:

Fixes #2634

## Proposed changes

- Added check that the namespace exists when the user runs `okteto context use [vanilla cluster] --namespace [nsName]`.

### DIFF
--- a/cmd/context/create.go
+++ b/cmd/context/create.go
@@ -160,26 +160,45 @@ func (c *ContextCommand) UseContext(ctx context.Context, ctxOptions *ContextOpti
 			return err
 		}
 	}
-	if ctxOptions.IsOkteto && ctxOptions.Save {
-		okClient, err := c.OktetoClientProvider.Provide()
-		if err != nil {
-			return err
+
+	if ctxOptions.Save {
+		hasAccess := false
+		if ctxOptions.IsOkteto {
+			okClient, err := c.OktetoClientProvider.Provide()
+			if err != nil {
+				return err
+			}
+			hasOktetoClientAccess, err := utils.HasAccessToOktetoClusterNamespace(ctx, ctxOptions.Namespace, okClient)
+			if err != nil {
+				return err
+			}
+
+			hasAccess = hasOktetoClientAccess
+		} else {
+			k8sClient, _, err := c.K8sClientProvider.Provide(okteto.Context().Cfg)
+			if err != nil {
+				return err
+			}
+
+			hasK8sClientAccess, err := utils.HasAccessToNamespace(ctx, ctxOptions.Namespace, k8sClient)
+			if err != nil {
+				return err
+			}
+
+			hasAccess = hasK8sClientAccess
 		}
-		hasAccess, err := utils.HasAccessToNamespace(ctx, ctxOptions.Namespace, okClient)
-		if err != nil {
-			return err
-		}
+
 		if !hasAccess {
 			return oktetoErrors.UserError{E: fmt.Errorf("namespace '%s' not found on context '%s'", ctxOptions.Namespace, ctxOptions.Context),
 				Hint: "Please verify that the namespace exists and that you have access to it.",
 			}
 		}
-	}
-	if ctxOptions.Save {
+
 		if err := c.OktetoContextWriter.Write(); err != nil {
 			return err
 		}
 	}
+
 	if created && ctxOptions.IsOkteto {
 		oktetoLog.Success("Context '%s' created", okteto.RemoveSchema(ctxOptions.Context))
 	}
--- a/cmd/utils/okteto.go
+++ b/cmd/utils/okteto.go
@@ -23,10 +23,12 @@ import (
 	"github.com/okteto/okteto/pkg/model"
 	"github.com/okteto/okteto/pkg/okteto"
 	"github.com/okteto/okteto/pkg/types"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
 )
 
-// HasAccessToNamespace checks if the user has access to a namespace/preview
-func HasAccessToNamespace(ctx context.Context, namespace string, oktetoClient types.OktetoInterface) (bool, error) {
+// HasAccessToOktetoClusterNamespace checks if the user has access to a namespace/preview
+func HasAccessToOktetoClusterNamespace(ctx context.Context, namespace string, oktetoClient types.OktetoInterface) (bool, error) {
 
 	nList, err := oktetoClient.Namespaces().List(ctx)
 	if err != nil {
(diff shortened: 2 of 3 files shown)


ifbyol (approved):

cmd/utils/okteto.go
```diff
@@ -53,6 +55,21 @@ func HasAccessToNamespace(ctx context.Context, namespace string, oktetoClient ty
 	return false, nil
 }
 
+func HasAccessToNamespace(ctx context.Context, namespace string, k8sClient kubernetes.Interface) (bool, error) {
+	nList, err := k8sClient.CoreV1().Namespaces().List(ctx, metav1.ListOptions{})
```
pchico83, line 59: listing namespaces requires admin access, the user might only have access to a single namespace

cmd/context/create.go
```diff
@@ -160,26 +160,45 @@ func (c *ContextCommand) UseContext(ctx context.Context, ctxOptions *ContextOpti
...
+				return err
+			}
+
+			hasK8sClientAccess, err := utils.HasAccessToNamespace(ctx, ctxOptions.Namespace, k8sClient)
+			if err != nil {
+				return err
+			}
+
+			hasAccess = hasK8sClientAccess
 		}
```
AgustinRamiroDiaz, line 189: Could we extract this to an auxiliary function?

I also though about creating a struct for this, but it would have a lot of dependencies. WDYT?

AgustinRamiroDiaz (approved):

cmd/context/create.go
jmacelroy: It seems like it deserves some testing which may mean that it needs to be pulled out into a function. 

It looks like we should be unit testing for...

- either an okteto client or a k8s client is created depending on the conditions
- correct access check is done.

Both of these things seem to be done as side-effect and therefore may be hard to test without some sort of re-factor.

jmacelroy: Also, after looking at the context a bit more the function is already >100 lines, which is a problem. I'd like to not make it worse. I think pulling it out and writing tests on it will be a good path for improving this bit of code.

AdrianPedriza: yes, totally agree. I will refactor it.

cmd/context/create.go
```diff
@@ -196,6 +194,34 @@ func (c *ContextCommand) UseContext(ctx context.Context, ctxOptions *ContextOpti
 	return nil
 }
 
+func hasAccessToNamespace(ctx context.Context, c *ContextCommand, ctxOptions *ContextOptions) (bool, error) {
```
AgustinRamiroDiaz, line 197: I think we should split this function into 2:
1. creation of `access` function
2. running the `access` function

In this way we would not need to modify `utils.HasAccessToK8sClusterNamespace` to be a variable, we can leave it as a function and then inject the `access` function we'd like

We could also separate the testing of `creation` and `execution` to have smaller tests

cmd/utils/okteto.go
```diff
@@ -23,10 +23,22 @@ import (
...
 	"github.com/okteto/okteto/pkg/types"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
 )
 
-// HasAccessToNamespace checks if the user has access to a namespace/preview
-func HasAccessToNamespace(ctx context.Context, namespace string, oktetoClient types.OktetoInterface) (bool, error) {
+// HasAccessToK8sClusterNamespace checks if the user has access to a namespace
+var HasAccessToK8sClusterNamespace = func(ctx context.Context, namespace string, k8sClient kubernetes.Interface) (bool, error) {
+	_, err := k8sClient.CoreV1().Namespaces().Get(ctx, namespace, metav1.GetOptions{})
```
AgustinRamiroDiaz, line 32: As we only need the function to get the namespace, narrowing the `k8sClient` dependency into just a `namespaceGetter func(context.Context, string)` would be beneficial for testing and for having better code dependencies

cmd/utils/okteto.go
```diff
@@ -74,7 +86,7 @@ func ShouldCreateNamespace(ctx context.Context, ns string) (bool, error) {
 	if err != nil {
 		return false, err
 	}
-	hasAccess, err := HasAccessToNamespace(ctx, ns, c)
+	hasAccess, err := HasAccessToOktetoClusterNamespace(ctx, ns, c)
```
AgustinRamiroDiaz, line 89: Isn't this ok to be a recursive call?

cmd/utils/okteto.go
AdrianPedriza: we have already discussed this in private but just to put it down:

the ideal would be to make use of interfaces to solve this problem but because both clients do not expose the same functionality this is not the best case with the current design. On the other hand, I think the main thing and a good starting point is to have adequate test coverage on this issue, as agreed: check that we will work with the right client in each case and that the check of whether it is accessible to the namespace is correct, both of which are included in the test. As we have seen, injecting the method to check if a k8s cluster is accessible to the namespace does not allow a simple and clean test and implies perhaps adding some complexity to the logic that the issue does not require. For the moment I would stick with the option that gives us a simple code that covers all cases. Perhaps opening an issue once everything is clearer and the benefits are clearly seen may be a good option!

AgustinRamiroDiaz: As Adrian said, since the most important part is to check the client creation, we need to mock the `k8sclient` and the `oktetoclient`.

@AdrianPedriza Reading the other test in the same file I found that we can mock the k8sClient like this:
```
	fakeCtxCommand := newFakeContextCommand(fakeOktetoClient, user, []runtime.Object{
		&corev1.Namespace{
			ObjectMeta: v1.ObjectMeta{
				Name: "test",
			},
		},
	})
```

This way we don't need to have the `HasAccessToK8sClusterNamespace` function as a variable

cmd/context/create.go
pchico83, line 197: is this executed at the beginning of each cli command? or only for `okteto context`?

AdrianPedriza: only for `okteto context` command

pchico83: should we check access to the namespace on `okteto namespace use xxxx` too?

AdrianPedriza: I think so. Maybe the user wants to use a non-existent namespace and we should notify them.

AdrianPedriza: well, reviewing your question, a user who does not use a cluster managed by Okteto would never encounter the issue if he runs okteto namespace xxxx.

cmd/utils/okteto.go
AdrianPedriza: That's right. Thank you @AgustinRamiroDiaz ;)

AgustinRamiroDiaz (approved):

jLopezbarb (approved):

PR #2699, opened 2022-05-19
Only create readiness probes for docker-compose health checks

pchico83:

Liveness probes introduce friction for development because the container process and container logs are lost to debug startup issues. We should only translate compose healthchecks into readiness probes, which are pretty convenient for sequentializing services start order?


pkg/cmd/stack/translate.go
```diff
@@ -138,7 +138,6 @@ func translateDeployment(svcName string, s *model.Stack) *appsv1.Deployment {
 							Resources:       translateResources(svc),
 							WorkingDir:      svc.Workdir,
 							ReadinessProbe:  healthcheckProbe,
-							LivenessProbe:   healthcheckProbe,
```
jmacelroy, line 141: Is this only for healthchecks that are added from compose files? I see how this will help debug issues that are no longer visible due to restarts but we may still have problems. In this new scenario the pod will be deployed but kube will not route traffic to it. Do you foresee that as being a similar issue that we should handle as well?

jLopezbarb: Yes, it'll only affect compose and `okteto stack` files

jLopezbarb (approved):

pkg/cmd/stack/translate.go
pchico83: `pod will be deployed but kube will not route traffic to it` -> it will if the readiness probe is passing. That is exactly what a user wants when he defines healthchecks at the compose file. the liveness was just introducing noise and making things harder to debug

PR #2706, opened 2022-05-19
Adding test coverage for env setting and validation for deployment.

jmacelroy:
Starting to add some code coverage to gain a bit of experience testing and making changes.

### DIFF
--- a/cmd/deploy/validate_test.go
+++ b/cmd/deploy/validate_test.go
@@ -14,37 +14,80 @@
 package deploy
 
 import (
+	"os"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
 )
 
-func Test_ValidateVars(t *testing.T) {
-
+func Test_setEnvVars(t *testing.T) {
 	var tests = []struct {
-		name        string
-		variables   []string
-		expectedErr bool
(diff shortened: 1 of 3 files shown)


cmd/deploy/validate_test.go
```diff
@@ -14,37 +14,80 @@
...
+					value: "three",
+				},
+			},
 		},
 	}
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			err := validateOptionVars(tt.variables)
+			t.Parallel()
```
AgustinRamiroDiaz, line 82: How does parallel testing and the os package behave together? Could it lead to problems? i.e. changing envvars for another test

jmacelroy: It can lead to problems and not just due to the parallelism. I think I'll need to use some of the newer env var testing facilities in go and possibly refactor a bit. I'm going to try it out.

jmacelroy:
This is taking longer than expected and I have to do more of a re-factor if I want to safely test the env var getting set. I'm going to close the PR for now and have a note to come back to it, possibly with an integration test.

PR #2707, opened 2022-05-20
Mario/build context default

### LINKED ISSUE #2636: Set dockerfile and context defaults and warn if dockerfile is not in relative path to context
**Describe the bug**
If I have an image on the `build` section that sets the `context` field, the value for `dockerfile` should be relative to the `context` field. In particular, if not defined, it should default to `context/Dockerfile`.

**To Reproduce**
Remove this line:
https://github.com/okteto/movies/blob/main/okteto.yml#L6

and run `okteto build frontend`. It should work because that should be the default value of the `dockerfile` field.

mnevadom:
fixes #2636 

## Proposed changes

- add scenario when context is not empty and dockerfile is


AdrianPedriza (approved):

pkg/model/manifest_test.go
```diff
@@ -404,3 +404,148 @@ func TestSetManifestDefaultsFromDev(t *testing.T) {
...
 }
+
+func TestSetManifestBuildDefaults(t *testing.T) {
+	tests := []struct {
+		name             string
+		currentManifest  *Manifest
+		expectedManifest *Manifest
+	}{
+		{
+			name: "name filled, context empty, dockerfile empty",
```
AgustinRamiroDiaz, line 415: Could we add a test for when the 3 are filled?

mnevadom: Sure! But if they are filled, the result is what you fill!
On it!!

pkg/model/manifest_test.go
```diff
@@ -404,3 +404,148 @@ func TestSetManifestDefaultsFromDev(t *testing.T) {
...
+					"test": &BuildInfo{
+						Image:            "image",
+						VolumesToInclude: []StackVolume{{LocalPath: "lp", RemotePath: "rp"}},
+					},
+				},
+			},
+		},
+	}
+
+	for _, tt := range tests {
```
jmacelroy, line 545: Can we add a `t.Parallel` here to speed up the tests?

pkg/model/manifest.go
```diff
@@ -722,6 +722,8 @@ func (m *Manifest) setDefaults() error {
 			b.Context = b.Name
 			b.Name = ""
 		}
+		b.setDockerfileDefaults()
+
 		if !(b.Image != "" && len(b.VolumesToInclude) > 0 && b.Dockerfile == "") {
 			b.setBuildDefaults()
```
jmacelroy, line 728: I think it may be easier to combine all the b.*Defaults into a function tis is simply always called at this point and all instantiation logic is placed there. As it stands it looks like we set things up in 3 different places here. 

It may also be better to do this via a New* whenever it's created with the `m` or `manifest`. It may be that a New* isn't viable if we are relying on the zero value initialization.

jmacelroy: I think we should simplify the default setting and add test coverage for all of it.

pkg/model/manifest_test.go
mnevadom: yes, we can!

pkg/model/manifest.go
mnevadom: I will try it. Problem is that the setBuildDefaults function is beeing called from a lot of places... That's why I didn't want to touch anything

pkg/model/manifest_test.go
mnevadom: It was an interesting test @AgustinRamiroDiaz

pkg/model/manifest.go
```diff
@@ -722,6 +722,8 @@ func (m *Manifest) setDefaults() error {
 			b.Context = b.Name
 			b.Name = ""
 		}
+		b.setDockerfileDefaults()
+
 		if !(b.Image != "" && len(b.VolumesToInclude) > 0 && b.Dockerfile == "") {
```
ifbyol, line 727: As we are setting the default for `b.Dockerfile` in line 725, should we remove the `b.Dockerfile == ""` from this conditional as it will always contain a value now (and provokes that the code within the if is always executed)?

mnevadom: But if the context is empty, nothing happens in the DockerfileDefaults, and it has to continue making that condition

pkg/model/dev.go
```diff
@@ -555,6 +555,17 @@ func (b *BuildInfo) setBuildDefaults() {
 	}
 }
 
+func (b *BuildInfo) setDockerfileDefaults() {
+	if b.Context == "" {
+		return
+	}
+	if b.Dockerfile == "" {
```
ifbyol, line 562: @jLopezbarb Does dockerfile intended to support relative paths, absolute paths or can it contain  any of them?

pkg/model/manifest.go
ifbyol: oh, that's true. This is confusing and very error prone. Agree with Jacob on find a way to refactor and simplify all this logic

pchico83:
@mnevadom this should be the expected behavior:
- default value for `context`-> `.`
- default value for `dockerfile`-> `Dockerfile`

Then, given a value for `context`, let's say `$CONTEXT`, and a value for `dockerfile`, let's say $DOCKERFILE`, the path where we check the dockerfile of a build is `join($CONTEXT, $DOCKERFILE)`. If that file doesn't exist, we fallback to the path `$DOCKERFILE` for backwards compatibility.
- if the fallback logic works -> show a warning telling to use a `dockerfile` path relative to `context` and the `Dockerfile` that has been selected for the build
- otherwise, show a clear error telling `join($CONTEXT, $DOCKERFILE)` doesn't exist.

mnevadom:
If everyone agrees Im gonna go with Pablo's 'simplification'. 

Im going to refactor all the places where we are having this condition into the same 'setDefaults' when we are doing the serialization of the manifest/compose. 

And create several tests with all the possibilities to check the warnings too. This test can be parallel of course.

pkg/model/manifest_test.go
```diff
@@ -404,3 +404,173 @@ func TestSetManifestDefaultsFromDev(t *testing.T) {
...
+					"test": &BuildInfo{
+						Context: "context",
+					},
+				},
+			},
+			expectedManifest: &Manifest{
+				Build: ManifestBuild{
+					"test": &BuildInfo{
+						Context:    "context",
+						Dockerfile: "context/Dockerfile",
```
jLopezbarb, line 446: Probably with throw a windows error. Can you use `filepath` module?

pkg/model/manifest_test.go
```diff
@@ -404,3 +404,173 @@ func TestSetManifestDefaultsFromDev(t *testing.T) {
...
+						Name: "name",
+					},
+				},
+			},
+			expectedManifest: &Manifest{
+				Build: ManifestBuild{
+					"test": &BuildInfo{
+						Context:    "name",
+						Name:       "",
+						Dockerfile: "name/Dockerfile",
```
jLopezbarb, line 428: Probably with throw a windows error. Can you use filepath module?

mnevadom:
Im sorry @jmacelroy but for this particular test, I don't think its worth it to do it in parallel. 

Thing is, every test has to test the output of the manifest and the organization of the files of the repo. 

Meaning that every test creates a Dockerfile somewhere (PWD, or in a mock dir or in the context dir), and test if the setDefaults reports and error or not. 

That creates conflicts, because one test is creating in "." the Dockerfile to test that is correct to be there, and there's another ones that need NO dockerfile in "." to test we are getting an error. 

First thing that occurs to me was to name Dockerfiles with different names for every test, but one of the things tested is that we set as default the name = Dockerfile, and create the file with that name...

And the problem is that the function setDefaults is searching the context and dockerfile in PWD, and we can't change it in parallel to different places (because another option was that every test could simulate a repo in different places)


If anyone has any idea Im open to hear it!

pkg/model/manifest_test.go
```diff
@@ -404,3 +407,206 @@ func TestSetManifestDefaultsFromDev(t *testing.T) {
...
+
+			if errManifest != nil && !tt.expectedError {
+				t.Fatal(err)
+			}
+
+			if errManifest == nil && tt.expectedError {
+				t.Fatal("error not thrown")
+			}
+
+			assert.Equal(t, tt.expectedManifest, tt.currentManifest)
```
AgustinRamiroDiaz, line 598:
```suggestion
			assert.DeepEqual(t, tt.expectedManifest, tt.currentManifest)
```

AgustinRamiroDiaz:
> Im sorry @jmacelroy but for this particular test, I don't think its worth it to do it in parallel.
> 
> Thing is, every test has to test the output of the manifest and the organization of the files of the repo.
> 
> Meaning that every test creates a Dockerfile somewhere (PWD, or in a mock dir or in the context dir), and test if the setDefaults reports and error or not.
> 
> That creates conflicts, because one test is creating in "." the Dockerfile to test that is correct to be there, and there's another ones that need NO dockerfile in "." to test we are getting an error.
> 
> First thing that occurs to me was to name Dockerfiles with different names for every test, but one of the things tested is that we set as default the name = Dockerfile, and create the file with that name...
> 
> And the problem is that the function setDefaults is searching the context and dockerfile in PWD, and we can't change it in parallel to different places (because another option was that every test could simulate a repo in different places)
> 
> If anyone has any idea Im open to hear it!

@mnevadom 
It would be nice to have an isolated file system mock for each test. 
- One way would be to inject our dependencies on the file system operations we want, but that would require a big refactor since we use `os.Stat, os.Read ...` in a lot of places. It might be nice since it would allow us to decouple the code
- We could create a "folder" for each test, so they wouldn't conflict with one another. I think there was a package for go to create temporary directories for testing

pkg/model/manifest_test.go
mnevadom: image

mnevadom:
Hi @AgustinRamiroDiaz, yes, that was an option too I figured. But every time every test need to be in its own PWD, because the function Im testing (setDefaults) assumes that the context is in a relative path from PWD...

pkg/model/manifest_test.go
AgustinRamiroDiaz: Sorry, it was `reflect.DeepEqual`

pkg/model/manifest.go
```diff
@@ -717,13 +717,15 @@ func (m *Manifest) setDefaults() error {
 		}
 	}
 
-	for _, b := range m.Build {
+	for k, b := range m.Build {
 		if b.Name != "" {
 			b.Context = b.Name
 			b.Name = ""
 		}
-		if !(b.Image != "" && len(b.VolumesToInclude) > 0 && b.Dockerfile == "") {
```
ifbyol, line 725: @pchico83 @jLopezbarb  Don't we need this check anymore?

pkg/model/dev.go
```diff
@@ -546,13 +546,29 @@ func (dev *Dev) SetDefaults() error {
 	return nil
 }
 
-func (b *BuildInfo) setBuildDefaults() {
+func (b *BuildInfo) setBuildDefaults() error {
 	if b.Context == "" {
 		b.Context = "."
 	}
-	if _, err := url.ParseRequestURI(b.Context); err != nil && b.Dockerfile == "" {
```
ifbyol, line 553: @pchico83 @jLopezbarb In which cases the context could be a URI? Isn't needed anymore?

pkg/model/manifest_test.go
jLopezbarb: `assert.Equal` uses `reflect.DeepEqual` under the hood and provides a better error formatting to the dev. I think we should keep going with `assert.Equal`

pkg/model/manifest.go
jLopezbarb: Now it's done at serialisation time instead, so we don't have to check this anymore

pkg/model/dev.go
jLopezbarb: Right now we could have a GitHub repo as context and it was building the image, don't know if this change has break this scenario. @mnevadom could you check it pls?

ifbyol: If we need to recover that check, lets include a comment in the code to be aware of that please and thanks for the explanation @jLopezbarb !

pkg/model/manifest_test.go
mnevadom: Yes, I tried deep equal but the output wasn't that clear...

pkg/model/dev.go
mnevadom: Agree. Ok

mnevadom:
I have a decision to make here @pchico83  @jLopezbarb  @jmacelroy 

As it turns out, the function `setBuildDefaults` from object `BuildInfo` is beeing called a lot of times in the code. 

Problem is I just want to show the warning once. I tried to protect it by adding a `bool` to the object when the Dockerfile is modified (with context dir) or when the exception is been shown. 

But I've noticed that for some commands, for example `okteto deploy`, its not only that we call the function of the objects more than once, but we create new  `BuildInfo` objects and call it. 

That makes it harder to control the times we are showing the warning message. 
The way I see it, we only have two options: 

1. Create a global variable with the names of the services that already showed the message

- Easier and faster. Not very beautiful... And we have the problem of sending several files, if the first has a bad Dockerfile, we will show the warning, even if the second one was correcting it

3. Make a big refactor and review all the code searching for redundant creation of Manifest serializations

- At one point we should do this. The best solution but will take time and Im not sure if you want me to do this ...

mnevadom:
> I have a decision to make here @pchico83 @jLopezbarb @jmacelroy
> 
> As it turns out, the function `setBuildDefaults` from object `BuildInfo` is beeing called a lot of times in the code.
> 
> Problem is I just want to show the warning once. I tried to protect it by adding a `bool` to the object when the Dockerfile is modified (with context dir) or when the exception is been shown.
> 
> But I've noticed that for some commands, for example `okteto deploy`, its not only that we call the function of the objects more than once, but we create new `BuildInfo` objects and call it. I mean, that all the process of reading from file is been done
> 
> That makes it harder to control the times we are showing the warning message. The way I see it, we only have two options:
> 
> 1. Create a global variable with the names of the services that already showed the message
> 
> * Easier and faster. Not very beautiful... And we have the problem of sending several files, if the first has a bad Dockerfile, we will show the warning, even if the second one was correcting it
> 
> 3. Make a big refactor and review all the code searching for redundant creation of Manifest serializations
> 
> * At one point we should do this. The best solution but will take time and Im not sure if you want me to do this ...

image

rlamana:
We should definitely avoid an output like that. Does this mean the manifest is being read multiple times? 🤔 I don't know how the code is structured but it sounds like this is something to optimize, isn't there a way to show notifications just on the first pass?

jLopezbarb:
> We should definitely avoid an output like that. Does this mean the manifest is being read multiple times? 🤔 I don't know how the code is structured but it sounds like this is something to optimize, isn't there a way to show notifications just on the first pass?

Yes, We need to read the manifest to get context and namespace from the manifest to set the context of the command and re-read to expand the environment variables that might be on the manifest. We should definitely get it done before the next release (`2.4.0`) but I don't think this should be done as part of this PR.
I created https://github.com/okteto/okteto/issues/2728 this morning to track it. Should we prioritise it before doing any other issue?

mnevadom:
If you want I can fix it with the global variable for this PR. And then dedicate more time to the big refactor as another issue

jmacelroy:
> @mnevadom this should be the expected behavior:
> 
> * default value for `context`-> `.`
> * default value for `dockerfile`-> `Dockerfile`
> 
> Then, given a value for `context`, let's say `$CONTEXT`, and a value for `dockerfile`, let's say `$DOCKERFILE`, the path where we check the dockerfile of a build is `join($PWD, $CONTEXT, $DOCKERFILE)`. If that file doesn't exist, we fallback to the path `join($PWD, $DOCKERFILE)` for backwards compatibility.
> 
> * if the fallback logic works -> show a warning telling to use a `dockerfile` path relative to `context` and the `Dockerfile` that has been selected for the build
> * otherwise, show a clear error telling `join($CONTEXT, $DOCKERFILE)` doesn't exist.

@mnevadom and @pchico83 Looking back through this I have a question about possible changed behavior.

If there is currently a user that has a Docker file at both pwd/context/Dockerfile and pwd/Dockerfile AND they've specified a context then we will start to use pwd/context/Dockerfile where we would have used pwd/Dockerfile. Do I understand that correctly? If so, it could be that we *magically* break a users setup where they have both. While I do think this is unlikely should we detect this scenario and warn the user?

mnevadom:
> > > @mnevadom this should be the expected behavior:
> > > 
> > > * default value for `context`-> `.`
> > > * default value for `dockerfile`-> `Dockerfile`
> > > 
> > > Then, given a value for `context`, let's say `$CONTEXT`, and a value for `dockerfile`, let's say `$DOCKERFILE`, the path where we check the dockerfile of a build is `join($PWD, $CONTEXT, $DOCKERFILE)`. If that file doesn't exist, we fallback to the path `join($PWD, $DOCKERFILE)` for backwards compatibility.
> > > 
> > > * if the fallback logic works -> show a warning telling to use a `dockerfile` path relative to `context` and the `Dockerfile` that has been selected for the build
> > > * otherwise, show a clear error telling `join($CONTEXT, $DOCKERFILE)` doesn't exist.
> > 
> > 
> > @mnevadom and @pchico83 Looking back through this I have a question about possible changed behavior.
> > If there is currently a user that has a Docker file at both pwd/context/Dockerfile and pwd/Dockerfile AND they've specified a context then we will start to use pwd/context/Dockerfile where we would have used pwd/Dockerfile. Do I understand that correctly? If so, it could be that we _magically_ break a users setup where they have both. While I do think this is unlikely should we detect this scenario and warn the user?
> 
> @mnevadom the behavior already changed in 2.3.0, right? this PR is a refactor to have the logic in a single place, right?

Here is the thing... 
To do a first small refactor, I had to modify the value in memory of the dockerfile with the JOIN of $context/$dockerfile. 

To do it well just when reading. To refactor this I had to change to places that were reading this value, to assume now is comming the right PATH. 

That changed a possible scenario for the user comparing it with the 2.3.0: 

If we had a Dockrefile in root and another one in the context, now is taking the context one (which is correct). 

But in the 2.3.0 was taking the one in the root, because there was no check and if the value of dockerfile existed then it just continues building...


From that point of view, this issue is not just a refactor, because it changes something for the user... 

My proposal is that we can leave the behaviour just as it was. And I can undo some things of the PR, and just show the warning here. 

And then create another one to do the big refactor which will parse just once the manifest, and change the scenario for the user to do it properly...

jmacelroy:
> > > > @mnevadom this should be the expected behavior:
> > > > 
> > > > * default value for `context`-> `.`
> > > > * default value for `dockerfile`-> `Dockerfile`
> > > > 
> > > > Then, given a value for `context`, let's say `$CONTEXT`, and a value for `dockerfile`, let's say `$DOCKERFILE`, the path where we check the dockerfile of a build is `join($PWD, $CONTEXT, $DOCKERFILE)`. If that file doesn't exist, we fallback to the path `join($PWD, $DOCKERFILE)` for backwards compatibility.
> > > > 
> > > > * if the fallback logic works -> show a warning telling to use a `dockerfile` path relative to `context` and the `Dockerfile` that has been selected for the build
> > > > * otherwise, show a clear error telling `join($CONTEXT, $DOCKERFILE)` doesn't exist.
> > > 
> > > 
> > > @mnevadom and @pchico83 Looking back through this I have a question about possible changed behavior.
> > > If there is currently a user that has a Docker file at both pwd/context/Dockerfile and pwd/Dockerfile AND they've specified a context then we will start to use pwd/context/Dockerfile where we would have used pwd/Dockerfile. Do I understand that correctly? If so, it could be that we _magically_ break a users setup where they have both. While I do think this is unlikely should we detect this scenario and warn the user?
> > 
> > 
> > @mnevadom the behavior already changed in 2.3.0, right? this PR is a refactor to have the logic in a single place, right?
> 
> Here is the thing... To do a first small refactor, I had to modify the value in memory of the dockerfile with the JOIN of $context/$dockerfile.
> 
> To do it well just when reading. To refactor this I had to change to places that were reading this value, to assume now is comming the right PATH.
> 
> That changed a possible scenario for the user comparing it with the 2.3.0:
> 
> If we had a Dockrefile in root and another one in the context, now is taking the context one (which is correct).
> 
> But in the 2.3.0 was taking the one in the root, because there was no check and if the value of dockerfile existed then it just continues building...
> 
> From that point of view, this issue is not just a refactor, because it changes something for the user...
> 
> My proposal is that we can leave the behaviour just as it was. And I can undo some things of the PR, and just show the warning here.
> 
> And then create another one to do the big refactor which will parse just once the manifest, and change the scenario for the user to do it properly...

It's difficult for me to determine the impact of the behavior change. It could be nothing or it could break user behavior. I also want to move forward with fixing. @pchico83 can you evaluate the latest change of behavior and provide an opinion?

Personally, I think I'm fine moving forward with the refactor and changing the behavior for this corner case as long as we are certain it's the behavior we want long-term. Thanks for explaining in such detail @mnevadom and being patient with me.

cmd/up/up.go
```diff
@@ -820,7 +820,9 @@ func setBuildEnvVars(m *model.Manifest, devName string) error {
 		imageWithDigest, err := registry.NewOktetoRegistry().GetImageTagWithDigest(opts.Tag)
 		if err == nil {
 			builder := buildv2.NewBuilderFromScratch()
+			sp.Stop()
 			builder.SetServiceEnvVars(buildName, imageWithDigest)
+			sp.Start()
```
ifbyol, line 825: Why these changes?

pkg/cmd/build/build_test.go
```diff
@@ -236,3 +240,127 @@ func Test_OptsFromBuildInfo(t *testing.T) {
 		})
 	}
 }
+
+func TestOptsFromBuildInfo(t *testing.T) {
```
ifbyol, line 244: I think the tests name doesn't match the function we actually test. It seems this is testing `extractFromContextAndDockerfile`

pkg/cmd/build/build_test.go
```diff
@@ -236,3 +240,127 @@ func Test_OptsFromBuildInfo(t *testing.T) {
...
+			oktetoLog.SetOutput(&buf)
+
+			defer func() {
+				oktetoLog.SetOutput(os.Stderr)
+			}()
+
+			file := extractFromContextAndDockerfile(buildName, tt.dockerfile, tt.svcName)
+			warningErr := strings.TrimSuffix(buf.String(), "\n")
+
+			if warningErr != "" && tt.expectedError == "" {
```
ifbyol, line 340: Not needed to address in the PR but as a general comment. In general, I would try to avoid to include ifs depending on input parameters for tests. This makes tests more difficult to read, understand and more error prone. 

`table driven test` is really good when you want to test different scenarios  of a function with different inputs and outputs but without having to include branching code in the tests.

pkg/model/manifest.go
```diff
@@ -967,7 +968,7 @@ func (m *Manifest) WriteToFile(filePath string) error {
 }
 
 // reorderDocFields orders the manifest to be: name -> build -> deploy -> dependencies -> dev
-func (*Manifest) reorderDocFields(doc yaml3.Node) yaml3.Node {
+func (*Manifest) reorderDocFields(doc *yaml3.Node) *yaml3.Node {
```
ifbyol, line 971: Why this change?

pkg/cmd/build/build_test.go
mnevadom: Right

pkg/model/manifest.go
mnevadom: deepsource change

AgustinRamiroDiaz: I have mixed feelings with this: on one hand `go` doesn't do deep copies, so passing the `doc` by value would not copy the `Content []*Node` which is where all the logic is being done. In this sense, it would be better to pass by *reference. On the other hand, passing by reference would have another behavior (in line `1056` we fill `Content`). If we are getting a pointer we can modify, we should be explicit of *ownership* and change the signature of the function, so that we don't return. 

I think this dichotomy is everywhere, so we should be careful of who *owns* the pointers at any time

pkg/model/manifest_test.go
```diff
@@ -155,6 +155,7 @@ func Test_validateDivert(t *testing.T) {
 }
 
 func TestInferFromStack(t *testing.T) {
+	dirtest := "/stack/dir/"
```
AgustinRamiroDiaz, line 158: This might have issues with other OS, can we use something like `filepath` or `os`?

jLopezbarb: `filepath.IsAbs` behaviour is different on linux and windows (on windows it expects the disk first. for example `C:/` and I don't know if it can affect) using `filepath.Clean("/stack/dir/")` would do the trick

pkg/model/manifest.go
jmacelroy: @mnevadom What's the error? The changes in this PR with pointer de-referencing seem quite dangerous.

PR #2711, opened 2022-05-23
fix: not fallback buildv1 if error unmarshalling manifest

### LINKED ISSUE #2689: okteto build and okteto up parse the manifest differently
**Describe the bug**
I have this manifest:
``` 
build:
  todo-list
  
deploy:
  - cd manifest; kustomize edit set image okteto.dev/todo-list=okteto.dev/todo-list:${OKTETO_GIT_COMMIT}
  - kubectl apply -k manifest
dev:
  todo-list:
    image: okteto/golang:1
    command: bash
    securityContext:
      capabilities:
        add:
        - SYS_PTRACE
    volumes:
    - /go/pkg/
    - /root/.cache/go-build/
    sync:
    - .:/app
    forward:
    - 2345:2345

```

When I run `okteto build`, the command ends successfully. But when I run ` okteto up`, I get this:
```
Invalid manifest:
    - line 2: cannot unmarshal !!str `todo-list` into model.ManifestBuild
    See https://okteto.com/docs/reference/manifest/ for details
``` 

Why do I get two different behaviors for the same manifest? Isn't ` up` simply calling ` build` internally for the build phase?

- OS: macOS
- Version okteto version 2.3.0-rc.1

### PR DESCRIPTION
**Pull request by AdrianPedriza**
Fixes #2689

## Proposed changes

- instead of try to build using a Dockerfile when user runs `okteto build` and there is an invalid manifest we should return an error because the user's initial goal is to use the manifest to build the necessary images.

### DIFF
--- a/cmd/build/command.go
+++ b/cmd/build/command.go
@@ -88,8 +88,7 @@ func (bc *Command) getBuilder(options *types.BuildOptions) (Builder, error) {
 
 	manifest, err := bc.GetManifest(options.File)
 	if err != nil {
-		var invalidManifestErr *oktetoErrors.InvalidManifestError
-		if errors.As(err, &invalidManifestErr) {
+		if errors.Is(err, oktetoErrors.ErrInvalidManifest) {
 			return nil, err
 		}
 
--- a/cmd/build/command.go
+++ b/cmd/build/command.go
@@ -93,7 +93,7 @@ func (bc *Command) getBuilder(options *types.BuildOptions) (Builder, error) {
 			return nil, err
 		}
 
-		oktetoLog.Info("The manifest %s is not v2 compatible, falling back to building as a v1 manifest: %v", options.File, err)
+		oktetoLog.Infof("The manifest %s is not v2 compatible, falling back to building as a v1 manifest: %v", options.File, err)
 		builder = buildv1.NewBuilder(bc.Builder, bc.Registry)
 	} else {
 		if isBuildV2(manifest) {
--- a/cmd/build/command.go
+++ b/cmd/build/command.go
@@ -92,7 +92,7 @@ func (bc *Command) getBuilder(options *types.BuildOptions) (Builder, error) {
 			return nil, err
 		}
 
-		oktetoLog.Warning("error getting manifest v2 from file %s: %v. Fallback to build v1", options.File, err)
+		oktetoLog.Info("The manifest %s is not v2 compatible, falling back to building as a v1 manifest: %v", options.File, err)
 		builder = buildv1.NewBuilder(bc.Builder, bc.Registry)
 	} else {
 		if isBuildV2(manifest) {
--- a/cmd/build/command.go
+++ b/cmd/build/command.go
@@ -88,7 +88,8 @@ func (bc *Command) getBuilder(options *types.BuildOptions) (Builder, error) {
 
 	manifest, err := bc.GetManifest(options.File)
 	if err != nil {
-		if errors.Is(err, oktetoErrors.ErrInvalidManifest) {
+		var invalidManifestErr *oktetoErrors.InvalidManifestError
+		if errors.As(err, &invalidManifestErr) {
 			return nil, err
 		}
 
--- a/cmd/build/command.go
+++ b/cmd/build/command.go
@@ -15,6 +15,7 @@ package build
 
 import (
 	"context"
+	"strings"
 
 	buildv1 "github.com/okteto/okteto/cmd/build/v1"
 	buildv2 "github.com/okteto/okteto/cmd/build/v2"
@@ -53,7 +54,11 @@ func Build(ctx context.Context) *cobra.Command {
 		RunE: func(cmd *cobra.Command, args []string) error {
 			bc := NewBuildCommand()
 
-			manifest, isBuildV2 := bc.getManifestAndBuildVersion(options)
+			manifest, isBuildV2, err := bc.getManifestAndBuildVersion(options)
+			if err != nil {
+				return err
+			}
+
 			options.CommandArgs = args
 			options.Manifest = manifest
 
@@ -80,16 +85,21 @@ func Build(ctx context.Context) *cobra.Command {
 	return cmd
 }
 
-func (bc *Command) getManifestAndBuildVersion(options *types.BuildOptions) (*model.Manifest, bool) {
+func (bc *Command) getManifestAndBuildVersion(options *types.BuildOptions) (*model.Manifest, bool, error) {
 	manifest, errManifest := bc.GetManifest(options.File)
-	if errManifest != nil {
-		oktetoLog.Debug("error getting manifest v2 from file %s: %v. Fallback to build v1", options.File, errManifest)
-	}
 
 	isBuildV2 := errManifest == nil &&
 		manifest.IsV2 &&
 		len(manifest.Build) != 0
-	return manifest, isBuildV2
+
+	if errManifest != nil {
+		if strings.HasPrefix(errManifest.Error(), "invalid manifest") {
+			return manifest, isBuildV2, errManifest
+		}
+
+		oktetoLog.Warning("error getting manifest v2 from file %s: %v. Fallback to build v1", options.File, errManifest)
+	}
+	return manifest, isBuildV2, nil
 }
 
 func (bc *Command) getBuilder(isBuildV2 bool) Builder {
--- a/cmd/build/command.go
+++ b/cmd/build/command.go
@@ -53,18 +53,12 @@ func Build(ctx context.Context) *cobra.Command {
 		Use:   "build [service...]",
 		Short: "Build and push the images defined in the 'build' section of your okteto manifest",
 		RunE: func(cmd *cobra.Command, args []string) error {
+			options.CommandArgs = args
 			bc := NewBuildCommand()
 
-			manifest, err := bc.getManifest(options)
+			builder, err := bc.getBuilder(options)
 			if err != nil {
 				return err
-			}
-
-			var builder Builder
-			if isBuildV2(manifest) {
-				builder = buildv2.NewBuilder(bc.Builder, bc.Registry)
-			} else {
-				builder = buildv1.NewBuilder(bc.Builder, bc.Registry)
 			}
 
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: cmd/build/command.go [hunk 1]
```diff
@@ -80,16 +85,21 @@ func Build(ctx context.Context) *cobra.Command {
...
-		oktetoLog.Debug("error getting manifest v2 from file %s: %v. Fallback to build v1", options.File, errManifest)
-	}
 
 	isBuildV2 := errManifest == nil &&
 		manifest.IsV2 &&
 		len(manifest.Build) != 0
-	return manifest, isBuildV2
+
+	if errManifest != nil {
+		if strings.HasPrefix(errManifest.Error(), "invalid manifest") {
```
jmacelroy, line 96: We should not be using the error string to perform any logic. This should be changed to use error types or specific error instances. This way we can rely on the type system and get the full benefit of Go.

jmacelroy: This means changing some of the `bc.GetManifest` code.

jmacelroy:
I made some comments on the issue that I think we should address in this PR.

cmd/build/command.go
ifbyol: `Read` function within `manifest.go` file is probably the place where we should return a specific error type to do what Jacob says.

cmd/build/command_test.go
```diff
@@ -37,26 +38,120 @@ func getManifestWithError(_ string) (*model.Manifest, error) {
...
+					},
+					"test-2": &model.BuildInfo{
+						Image: "test/test-2",
+					},
+				},
+			},
+			expectedAnswer: true,
+		},
+	}
+	for _, tt := range tests {
```
AdrianPedriza, line 121: not sure if t.Parallel() is convenient in these small tests

cmd/build/command.go
```diff
@@ -80,21 +89,23 @@ func Build(ctx context.Context) *cobra.Command {
...
-		len(manifest.Build) != 0
-	return manifest, isBuildV2
+	return manifest, nil
 }
 
-func (bc *Command) getBuilder(isBuildV2 bool) Builder {
-	if isBuildV2 {
-		return buildv2.NewBuilder(bc.Builder, bc.Registry)
+func isBuildV2(m *model.Manifest) bool {
+	if m == nil {
```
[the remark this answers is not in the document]
AdrianPedriza: this has to do with whether to allow fallback or not. As I said I'm not sure what use cases we currently cover with it so for now that logic is allowed. Basically what we are considering as an error to not continue with the execution of the command is that the manifest oktet given by the user is invalid, then we believe we are in a situation not expected by the user and we return the error. On the other hand, the manifest may be null because it does not exist, but this situation does not generate an error for the user. That is why the logic that checks that the manifest == nil after the error check, because at the moment a null manifest because the user has not created it does not cause an error that terminates the process.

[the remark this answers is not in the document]
jmacelroy: Yes, I feel like nil is the wrong way to control that logic. I'd expect something like a `errors.Is(err, v2NotDectected)`

cmd/build/command.go
```diff
@@ -80,21 +83,29 @@ func Build(ctx context.Context) *cobra.Command {
...
+func (bc *Command) getBuilder(options *types.BuildOptions) (Builder, error) {
+	var builder Builder
+
+	manifest, err := bc.GetManifest(options.File)
+	if err != nil {
+		if errors.Is(err, oktetoErrors.ErrInvalidManifest) {
+			return nil, err
+		}
+
+		oktetoLog.Warning("error getting manifest v2 from file %s: %v. Fallback to build v1", options.File, err)
```
jmacelroy, line 95: Will this be seen by the user? If so, I think that `Warning` may be a bit too high and `Info` may be a better choice. Also, I'd state `The manifest %s is not v2 compatible, falling back to building as a v1 manifest: %v`

pkg/errors/errors.go
```diff
@@ -36,6 +36,11 @@ type CommandError struct {
 	Reason error
 }
 
+// Error returns the error message
+func (e UserError) Is(target error) bool {
```
jmacelroy, line 40: I'm sorry, this is not correct and I didn't catch it previously. This is still just checking strings. There should be no string logic in error comparisons. It's too brittle and error strings can not be used for logic. 

We must either use a type assertion or the `==` operator to check and exact match using the type system and not string logic.

pkg/errors/errors.go
```diff
@@ -161,6 +166,9 @@ var (
 
 	// ErrDevContainerNotExists is raised when the dev container doesn't exist on dev section
 	ErrDevContainerNotExists = "development container '%s' doesn't exist"
+
+	//ErrInvalidManifest is raised when cannot unmarshal manifest properly
+	ErrInvalidManifest = fmt.Errorf("invalid manifest")
```
jmacelroy, line 171: Where is this error being thrown? I only see it being used in a test.

cmd/build/command.go
```diff
@@ -80,21 +83,30 @@ func Build(ctx context.Context) *cobra.Command {
...
+	var builder Builder
+
+	manifest, err := bc.GetManifest(options.File)
+	if err != nil {
+		var invalidManifestErr *oktetoErrors.InvalidManifestError
+		if errors.As(err, &invalidManifestErr) {
+			return nil, err
+		}
+
+		oktetoLog.Info("The manifest %s is not v2 compatible, falling back to building as a v1 manifest: %v", options.File, err)
```
jLopezbarb, line 96:
```suggestion
		oktetoLog.Infof("The manifest %s is not v2 compatible, falling back to building as a v1 manifest: %v", options.File, err)
```

cmd/build/command_test.go
```diff
@@ -37,26 +40,194 @@ func getManifestWithError(_ string) (*model.Manifest, error) {
...
```
jLopezbarb, line 124: I think you should do something like this, it will be a race condition if it's not done
```suggestion
	for _, tt := range tests {
	  tt := tt
```

cmd/build/command_test.go
```diff
@@ -37,26 +40,194 @@ func getManifestWithError(_ string) (*model.Manifest, error) {
...
+			buildCommand: &Command{
+				GetManifest: getManifestWithError,
+			},
+			expectedError:     false,
+			isBuildV2Expected: false,
+		},
+	}
+
+	options := &types.BuildOptions{}
+	for _, tt := range tests {
```
jLopezbarb, line 205:
```suggestion
	for _, tt := range tests {
	  tt := tt
```

cmd/build/command.go
```diff
@@ -80,21 +83,29 @@ func Build(ctx context.Context) *cobra.Command {
...
-func (bc *Command) getManifestAndBuildVersion(options *types.BuildOptions) (*model.Manifest, bool) {
-	manifest, errManifest := bc.GetManifest(options.File)
-	if errManifest != nil {
-		oktetoLog.Debug("error getting manifest v2 from file %s: %v. Fallback to build v1", options.File, errManifest)
+func (bc *Command) getBuilder(options *types.BuildOptions) (Builder, error) {
+	var builder Builder
+
+	manifest, err := bc.GetManifest(options.File)
+	if err != nil {
+		if errors.Is(err, oktetoErrors.ErrInvalidManifest) {
```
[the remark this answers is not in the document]
AdrianPedriza: I have changed the error handling logic a bit. Since an invalid manifest error can not be comparable with == because the cause of the error is unknown a priori.

ifbyol (approved):

jLopezbarb (approved):

cmd/build/command.go
```diff
@@ -80,21 +83,30 @@ func Build(ctx context.Context) *cobra.Command {
...
-func (bc *Command) getManifestAndBuildVersion(options *types.BuildOptions) (*model.Manifest, bool) {
-	manifest, errManifest := bc.GetManifest(options.File)
-	if errManifest != nil {
-		oktetoLog.Debug("error getting manifest v2 from file %s: %v. Fallback to build v1", options.File, errManifest)
+func (bc *Command) getBuilder(options *types.BuildOptions) (Builder, error) {
+	var builder Builder
+
+	manifest, err := bc.GetManifest(options.File)
+	if err != nil {
+		var invalidManifestErr *oktetoErrors.InvalidManifestError
```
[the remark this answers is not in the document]
AdrianPedriza: I just uploaded a new solution, I think it is one of the cleanest we can have in our case. Since we are facing a wrapped error we cannot use the `==` operator in this case. We must use the Is() method. Also, in the Golang documentation it is recommended to compare errors using Is() instead of the `==` operator.

pkg/model/manifest.go
```diff
@@ -626,7 +626,9 @@ func getOktetoManifest(devPath string) (*Manifest, error) {
 
 	manifest, err := Read(b)
 	if err != nil {
-		return nil, err
+		return nil, &oktetoErrors.InvalidManifestError{
```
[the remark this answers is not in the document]
AdrianPedriza: I'm not sure what you mean. Inside the Read() function we cover the case of not being able to perform a unmarshal correctly due to the content, adding to the error the cause of this. Basically what we are assuming is that any error produced during the reading of the manifest is considered due to an invalid manifest.

pkg/model/manifest.go
```diff
@@ -626,7 +626,7 @@ func getOktetoManifest(devPath string) (*Manifest, error) {
```
jmacelroy, line 626: If there is an i/o error reading the file, obtaining the bytes *and* the file exists, then we must not fallback.

I'm not sure we can tell directly if the file exists or not based on the error so we may need to first do a fs or os Stat call or something that is specific to the file not existing. If the file exists *and* an error occurrs then we can't fallback to v1 behavior.

AdrianPedriza: Added implementation and test that notifies of an error due to an invalid manifest if the file exists but there is no content. Also added a check if the manifest exists (if it doesn't exist, fallback). After talking with @jLopezbarb  about this we have concluded that for the moment we cover the cases where the user should be notified by an invalid manifest or otherwise fallback to buildv1. Refactoring these methods is tricky as they are used in almost every command.

PR #2713, opened 2022-05-24
Jlopezbarb/refactor unit test folders and envs

### LINKED ISSUE #2677: Simplify tests by using t.TempDir
Currently we are using os.MkdirTemp and os.RemoveAll to create and cleanup temporary directory in tests. We can use the T.TempDir function from the testing package to create temporary directory. The directory created by T.TempDir is automatically removed when the test and all its subtests complete.

This saves us at least 2 lines (error check, and cleanup) on every instance, or in some cases adds cleanup that we forgot.

Reference: https://pkg.go.dev/testing#T.TempDir

jLopezbarb:
Fixes #2677

## Proposed changes
- Use `t.TempDir` to create temporary dirs that gets removed when a test finish
- Use `t.SetEnv` to set the env vars for a test that returns back to previous state when the test finishes
- While doing this refactor detected a test which was passing because of setting an env var on a previous test. Fixed

### DIFF
--- a/cmd/build/v2/build_test.go
+++ b/cmd/build/v2/build_test.go
@@ -119,11 +119,7 @@ func TestOnlyInjectVolumeMountsInOkteto(t *testing.T) {
 		},
 		CurrentContext: "test",
 	}
-	dir, err := os.MkdirTemp("", "test")
-	if err != nil {
-		t.Fatal(err)
-	}
-	defer os.RemoveAll(dir)
+	dir := t.TempDir()
 
 	registry := test.NewFakeOktetoRegistry(nil)
 	builder := test.NewFakeOktetoBuilder(registry)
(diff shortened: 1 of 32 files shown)


cmd/build/v2/build_test.go
```diff
@@ -119,11 +119,7 @@ func TestOnlyInjectVolumeMountsInOkteto(t *testing.T) {
 		},
 		CurrentContext: "test",
 	}
-	dir, err := os.MkdirTemp("", "test")
-	if err != nil {
-		t.Fatal(err)
-	}
-	defer os.RemoveAll(dir)
+	dir := t.TempDir()
```
ifbyol, line 122: Don't we need specific names for the folder to make the test work? Some of these tests don't rely on the folder name for the build context?

pkg/config/config_test.go
```diff
@@ -92,7 +98,7 @@ func Test_homedirWindows(t *testing.T) {
 func TestGetOktetoHome(t *testing.T) {
 	dir := t.TempDir()
 
-	t.Setenv(model.OktetoFolderEnvVar, dir)
+	os.Setenv(model.OktetoFolderEnvVar, dir)
```
AgustinRamiroDiaz, line 101: Why was this rollbacked?

cmd/build/v2/build_test.go
jLopezbarb: Not those ones, the ones that needed specific names or be relative to a path I left them as they were

pkg/config/config_test.go
jLopezbarb: I overdid it by pressing ctrl+z 😅

jmacelroy (approved):

teresaromero (approved):

ifbyol (approved):

PR #2717, opened 2022-05-24
Remove dead code handling old Okteto API < 0.10

pchico83:

I checked our metrics and all our customers are in OE >= 0.10.6, so it's safe to remove the logic to handle those old APIs


rberrelleza (approved):
👏🏽👏🏽👏🏽

PR #2718, opened 2022-05-25
Fix "okteto up" metrics for vanilla clusters

pchico83:

Since CLI 1.14.2 (04 Nov 2021) we were ignoring `okteto up` metrics

### DIFF
--- a/pkg/analytics/track.go
+++ b/pkg/analytics/track.go
@@ -359,9 +359,15 @@ func track(event string, success bool, props map[string]interface{}) {
 		return
 	}
 
-	if !okteto.IsContextInitialized() || (!okteto.Context().Analytics && !okteto.IsOktetoCloud()) {
+	if !okteto.IsContextInitialized() {
+		oktetoLog.Info("failed to send analytics: okteto context not initialized")
 		return
 	}
+
+	if disabledByOktetoAdmin() {
+		return
+	}
+
 	// skip events from nested okteto deploys and manifest dependencies
 	origin := config.GetDeployOrigin()
 	if origin == "okteto-deploy" {
@@ -406,4 +412,11 @@ func track(event string, success bool, props map[string]interface{}) {
 	if err := mixpanelClient.Track(getTrackID(), event, e); err != nil {
 		oktetoLog.Infof("Failed to send analytics: %s", err)
 	}
+}
+
+func disabledByOktetoAdmin() bool {
+	if okteto.IsOktetoCloud() {
+		return false
+	}
+	return !okteto.Context().Analytics
 }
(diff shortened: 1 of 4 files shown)


pkg/analytics/track.go
```diff
@@ -359,9 +359,15 @@ func track(event string, success bool, props map[string]interface{}) {
 		return
 	}
 
-	if !okteto.IsContextInitialized() || (!okteto.Context().Analytics && !okteto.IsOktetoCloud()) {
```
pchico83, line 362: `okteto.Context().Analytics` is always `false` for vanilla clusters

pkg/analytics/track.go
```diff
@@ -359,9 +359,15 @@ func track(event string, success bool, props map[string]interface{}) {
...
+	if !okteto.IsContextInitialized() {
+		oktetoLog.Info("failed to send analytics: okteto context not initialized")
```
rberrelleza, line 363: I don't think we should log this. There are many legit reasons foe this no?

pkg/analytics/track.go
rberrelleza: Could we fix that instead, so the behavior is more obvious and consistent?

pkg/analytics/track.go
```diff
@@ -407,3 +413,13 @@ func track(event string, success bool, props map[string]interface{}) {
 		oktetoLog.Infof("Failed to send analytics: %s", err)
 	}
 }
+
+func disabledByOktetoAdmin() bool {
+	if !okteto.IsOkteto() {
+		return false
+	}
+	if okteto.IsOktetoCloud() {
```
rberrelleza, line 421: Can we remove the dependency on cloud and instead depend on what the context is telling us?

pkg/analytics/track.go
pchico83: No, if that happens, is something weird that we should investigate. this is not exposed to the user

pkg/analytics/track.go
pchico83: No, `telemetry` is disabled in Cloud, and we don't want to lose all `okteto up`s happening in Cloud

rberrelleza: Can we enable telemetry in cloud? Every "if cloud do x" block adds complexity and the possibilities to more bugs.

pchico83: That's has deeper implications is out of the scope of this PR

teresaromero (approved):

ifbyol (approved):

jLopezbarb (approved):

Issue #2719, opened 2022-05-25
Wrong log when using `autocreate`


pchico83:
**Describe the bug**
We are logging this warning when using a dev container with `autocreate`:
```
 !  could not add dev annotations due to: the application 'xxx' referred by your okteto manifest doesn't exist
 ```
 and this is fine, we shouldn't warn the user

pchico83:
This is throwing an error in 2.3.0, but not in 2.0.2

jmacelroy:
I am unable to reproduce this and will probably need more information in able to do so. I created a v2 manifest and compose for go-getting-started at https://github.com/jmacelroy/go-getting-started/blob/autocreate-with-2.0-manifest/okteto.yml

I am always successfully deploying, using `okteto up`, with no deployment existing and `autocreate: true` set.

I tried with the manifest in the branch, the manifest in the branch with `build` and `deploy` sections commented out, and with different names in compose file `services` section and the okteto manifest `dev` target. 

In all of those cases 2.3.0 worked with no errors shown. Is it possible there's an interaction issue with certain okteto server versions? I was using cloud.okteto.com

PR #2721, opened 2022-05-25
Remove warning when failing to add dev annotations

pchico83:

Fixes https://github.com/okteto/okteto/issues/2719

### DIFF
--- a/pkg/cmd/pipeline/translate.go
+++ b/pkg/cmd/pipeline/translate.go
@@ -228,27 +228,23 @@ func updateCmap(cmap *apiv1.ConfigMap, data *CfgData) error {
 }
 
 // AddDevAnnotations add deploy labels to the deployments/sfs
-func AddDevAnnotations(ctx context.Context, manifest *model.Manifest, services []string, c kubernetes.Interface) error {
+func AddDevAnnotations(ctx context.Context, manifest *model.Manifest, c kubernetes.Interface) {
 	repo := os.Getenv(model.GithubRepositoryEnvVar)
-	servicesSet := map[string]bool{}
-	for _, s := range services {
-		servicesSet[s] = true
-	}
-
 	for devName, dev := range manifest.Dev {
-		if servicesSet[devName] {
-			app, err := apps.Get(ctx, dev, manifest.Namespace, c)
-			if err != nil {
-				return err
-			}
-			if repo != "" {
-				app.ObjectMeta().Annotations[model.OktetoRepositoryAnnotation] = repo
-			}
-			app.ObjectMeta().Annotations[model.OktetoDevNameAnnotation] = devName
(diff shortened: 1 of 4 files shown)


pkg/cmd/pipeline/translate.go
```diff
@@ -228,27 +228,23 @@ func updateCmap(cmap *apiv1.ConfigMap, data *CfgData) error {
 }
 
 // AddDevAnnotations add deploy labels to the deployments/sfs
-func AddDevAnnotations(ctx context.Context, manifest *model.Manifest, services []string, c kubernetes.Interface) error {
+func AddDevAnnotations(ctx context.Context, manifest *model.Manifest, c kubernetes.Interface) {
 	repo := os.Getenv(model.GithubRepositoryEnvVar)
-	servicesSet := map[string]bool{}
```
pchico83, line 233: @AgustinRamiroDiaz if `services` is empty, which means deploying everything, we were never adding dev annotations, and we should, right?

jLopezbarb: I think it should be already done by this stage. Am I wrong?

AgustinRamiroDiaz: Once we run `setDeployOptionsValuesFromManifest`, that field should not be empty: https://github.com/okteto/okteto/blob/334b4fa2d108a732849d161014f6660fa29e1857/cmd/deploy/deploy.go#L462-L468 . Maybe the problem is that it's only ran when we have a compose?

pchico83: yes, I was using an okteto manifest and it was empty. But I think it's fine to try all the dev containers, they could be even created by a manual command

jLopezbarb (approved):

pkg/cmd/pipeline/translate.go
pchico83: My point is why to filter the dev containers mapping them with services to deploy. We can always try to annotate all of the dev containers. That filter is broken in a few cases:
- non compose deployments
- autocreate
- dev containers using `selector`

AgustinRamiroDiaz: Oh yeah, I posted the last comment and hadn't read yours. I agree, could we add some comments/tests for this function?

Also, is there a case where we'd want to fail given that we cannot annotate the dev containers?

pchico83: I removed all errors and changed them to debug info messages because these annotations are optional. We should never return an error if there is an issue annotation a dev container.

pchico83: @AgustinRamiroDiaz added a unit test

AgustinRamiroDiaz (approved):

PR #2724, opened 2022-05-26
Limit pvcstorageClassName using cloud

### LINKED ISSUE #2599: okteto up fails with volumespec error after changing the image
This was reported in Slack. This error happened after the user changed the image to alpine, and ran okteto up again.

```
Couldn't activate your development container
    error updating kubernetes volume claim: PersistentVolumeClaim "web-okteto" is invalid: spec: Forbidden: spec is immutable after creation except resources.requests for bound claims
  core.PersistentVolumeClaimSpec{
        ... // 2 identical fields
        Resources:        {Requests: {s"storage": {i: {...}, Format: "BinarySI"}}},
        VolumeName:       "pvc-8ac52988-c057-49de-b1c6-c2a444b816cd",
-       StorageClassName: &"standard",
+       StorageClassName: &"okteto-standard",
        VolumeMode:       &"Filesystem",
        DataSource:       nil,
  }

```

teresaromero:

Fixes #2599 

## Proposed changes

- When creating a PVC for a dev environment and user is in Cloud, prevent a custom storage class name to be used within the manifest. This will throw an error when storageClass is specified at manifest. In Cloud all storage classes are set by the webhook to the default one.

### DIFF
--- a/pkg/k8s/volumes/crud.go
+++ b/pkg/k8s/volumes/crud.go
@@ -20,7 +20,6 @@ import (
 
 	"github.com/google/uuid"
 	"github.com/okteto/okteto/cmd/utils"
-	"github.com/okteto/okteto/pkg/errors"
 	oktetoErrors "github.com/okteto/okteto/pkg/errors"
 	oktetoLog "github.com/okteto/okteto/pkg/log"
 	"github.com/okteto/okteto/pkg/model"
@@ -59,7 +58,7 @@ func CreateForDev(ctx context.Context, dev *model.Dev, c *kubernetes.Clientset, 
 	vClient := c.CoreV1().PersistentVolumeClaims(dev.Namespace)
 	pvc := translate(dev)
 	k8Volume, err := vClient.Get(ctx, pvc.Name, metav1.GetOptions{})
-	if err != nil && !errors.IsNotFound(err) {
+	if err != nil && !oktetoErrors.IsNotFound(err) {
 		return fmt.Errorf("error getting kubernetes volume claim: %s", err)
 	}
 	if k8Volume.Name == "" {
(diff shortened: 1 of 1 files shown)


pkg/k8s/volumes/crud.go
```diff
@@ -48,29 +49,33 @@ func List(ctx context.Context, namespace, labels string, c kubernetes.Interface)
 
 // CreateForDev deploys the volume claim for a given development container
 func CreateForDev(ctx context.Context, dev *model.Dev, c *kubernetes.Clientset, devPath string) error {
+	if okteto.IsOktetoCloud() && dev.PersistentVolumeStorageClass() != "" {
```
ifbyol, line 51: Not sure we want to include this logic in the CLI. If at some point we want to allow it, we could not change it for some specific versions of the CLI. Also, the case we want to "fix" for Cloud could happen in other Okteto instances as it is a documented setting: https://www.okteto.com/docs/enterprise/administration/configuration/#volumes.

jLopezbarb: I was gonna say the same, if it's affecting Cloud it should be affecting other Okteto clusters, so I don't know if this should be fixed on the CLI side. Maybe we should return a proper error on the mutation web hook?

teresaromero: okey, i did not understood correctly the solution for this :D thought the error should be thrown by the cli, but should be thrown by webhook? When CREATE, the webhook has also a force to use the standard, so we are not even notifying the user that even though they used "X" at the manifest, we are creating the class "Y" :/

pchico83: @teresaromero why don't we do the same on UPDATE at the webhook level?

teresaromero: yes, im on it! :) i was confused this morning - will close this PR and open new on webhook

PR #2742, opened 2022-06-01
Override stack name if option is not empty

### LINKED ISSUE #2661: `okteto deploy --name` isn't probagated to my stack name label
**Describe the bug**
I am using the `name` flag with my compose file:
```
okteto deploy --file docker-compose.yml --name name1
```
but the `--name` flag is not propagated to the stack name label.

This is problematic, because if I run `okteto deploy --file docker-compose.yml --name name1`, the second command reuses the same stack name label and replaces everything deployed by my first `okteto deploy` command.

**Additional context**
This is a serious bug if I place my compose files for different repos in the `.okteto` folder. In this case, `okteto` is always my default stack name, and everytime I deploy a new repo, it replaces my services created by previous repos

teresaromero:

Fixes #2661 

## Proposed changes

- when `okteto deploy --name <name>` this option was overriding the stack name. For this, prior to get the manifest (stack) the OktetoNameEnv is set with the overriden value so this can be used for the stack name and also the Endpoints prefix, which uses the stack name


cmd/deploy/deploy.go
```diff
@@ -152,13 +152,18 @@ func Deploy(ctx context.Context) *cobra.Command {
 				}
 			}
 
+			// when name is an option - this overrides the manifest or inferred name
+			if options.Name != "" {
+				os.Setenv(model.OktetoNameEnvVar, options.Name)
```
jmacelroy, line 157: This seems like a very indirect way of setting this and is quite brittle. We've effectively set a global variable, in env, in order to make a decision later on. What are all of the ways this env var get set and how can we track them? Is there a more direct approach we can take that eliminates this?

It seems like a better, and less brittle way, would be to set the name as a config option and somehow passing it into the code that will eventually do the deploy directly.

teresaromero: yup, this was actually not needed to be fixed :/ i've updated the scope for the PR due to the comments at the issue #2661

ifbyol (approved):

ifbyol:
Note to bear in mind. This could be considered as a non-backward compatible change. If I already executed deployed an application with current version of the CLI, I would have a dev environment called `okteto`. If I update the CLI after releasing this change and do a redeploy using the same command, now, the name of the dev environment would be different (okteto folder's parent folder) and I would deploy the entire dev environment instead of redeploying the existing one and it might end up with some k8s object collision

teresaromero:
> Note to bear in mind. This could be considered as a non-backward compatible change. If I already executed deployed an application with current version of the CLI, I would have a dev environment called `okteto`. If I update the CLI after releasing this change and do a redeploy using the same command, now, the name of the dev environment would be different (okteto folder's parent folder) and I would deploy the entire dev environment instead of redeploying the existing one and it might end up with some k8s object collision
> 
> I don't know how frequent could be this case, but commented just to take into account
> 
> cc @jmacelroy @pchico83

- Running `okteto deploy` with current -> deploys the `okteto` dev environment
- Running again `okteto deploy` with changes -> creates an empty dev environment with the correct name (inferred by the folder) but there are some warnings and the k8s components are not "updated" to be at the new dev environment

CLI logs:
```

❯ okteto deploy
 i  Using teresa @ okteto.teresa.dev.okteto.net as context
 i  Building image for service 'vote'...
 i  To rebuild your image manually run 'okteto build vote' or 'okteto deploy --build'
 i  Building image for service 'vote'
[+] Building 22.9s (11/11) FINISHED
 ✓  Image for service 'vote' pushed to registry: registry.teresa.dev.okteto.net/teresa/okteto-vote:okteto-with-volume-mounts
 i  Deploying compose
 ✓  Volume 'redis' created
 ✓  Service 'vote' created
 ✓  Service 'redis' created
 ✓  Compose 'okteto' successfully deployed
 ✓  Development environment 'okteto' successfully deployed
```

```
❯ ok deploy
 i  Using teresa @ okteto.teresa.dev.okteto.net as context
 ✓  Endpoint 'vote' created
 !  skipping creation of volume 'redis' due to name collision with volume in stack 'okteto'
 !  skipping deploy of deployment 'vote' due to name collision with deployment in stack 'okteto'
 !  skipping deploy of statefulset 'redis' due to name collision with statefulset in stack 'okteto'
 ✓  Compose 'compose-getting-started' successfully deployed
```
Screenshot 2022-06-03 at 09 09 36

ifbyol:
@teresaromero what would happen if after that I execute `okteto destroy`?

teresaromero:
> @teresaromero what would happen if after that I execute `okteto destroy`?

when using the current one, the `okteto` dev environment is deleted, but the changes, deletes the new dev environment, `okteto` still is there

Should we add some logic so the components change their labels and dev environment name to be swaped?

pchico83:
> > @teresaromero what would happen if after that I execute `okteto destroy`?
> 
> when using the current one, the `okteto` dev environment is deleted, but the changes, deletes the new dev environment, `okteto` still is there
> 
> Should we add some logic so the components change their labels and dev environment name to be swaped?

I would add logic to, in case we are in `.okteto`, replace this warning ` !  skipping ` by actually modifying the component if its dev environment name was `okteto`.

@ifbyol great catch!

teresaromero:
> > > @teresaromero what would happen if after that I execute `okteto destroy`?
> > 
> > 
> > when using the current one, the `okteto` dev environment is deleted, but the changes, deletes the new dev environment, `okteto` still is there
> > Should we add some logic so the components change their labels and dev environment name to be swaped?
> 
> I would add logic to, in case we are in `.okteto`, replace this warning `! skipping` by actually modifying the component if its dev environment name was `okteto`.
> 
> @ifbyol great catch!

I've been going tthough the code to make this modification, we have to change all components ( deployments, sfs, jobs, volumes, etc...) ?

jLopezbarb (approved):

teresaromero:
I've tried to add some logic to swap the stack names in case the old one was `okteto` but i get an error from k8s 
````
Error updating deployment of service 'vote': Deployment.apps "vote" is invalid: spec.selector: Invalid value: v1.LabelSelector{MatchLabels:map[string]string{"stack.okteto.com/name":"compose-getting-started", "stack.okteto.com/service":"vote"}, MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: field is immutable
```

as the labels are inmutable, should then display a warning message so the user can destroy the `okteto ` dev environment and start again?

cc: @pchico83 @ifbyol

pchico83:
> I've tried to add some logic to swap the stack names in case the old one was `okteto` but i get an error from k8s
> 
> ```
> Error updating deployment of service 'vote': Deployment.apps "vote" is invalid: spec.selector: Invalid value: v1.LabelSelector{MatchLabels:map[string]string{"stack.okteto.com/name":"compose-getting-started", "stack.okteto.com/service":"vote"}, MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: field is immutable
> ```
> 
> as the labels are inmutable, should then display a warning message so the user can destroy the `okteto ` dev environment and start again?
> 
> cc: @pchico83 @ifbyol

@teresaromero I would automatically  `destroy` + `deploy` the resources, and keep only the volumes. That way, the operation is transparent to the user. We already do something similar when a statefulset cannot be edited

teresaromero:
edit description with instructions on how to reproduce the change

pkg/cmd/stack/deploy.go
```diff
@@ -500,7 +503,12 @@ func deployDeployment(ctx context.Context, svcName string, s *model.Stack, c kub
 		if isNewDeployment {
 			return false, fmt.Errorf("error creating deployment of service '%s': %s", svcName, err.Error())
 		}
-		return false, fmt.Errorf("error updating deployment of service '%s': %s", svcName, err.Error())
+		if err := deployments.Destroy(ctx, old.Name, old.Namespace, c); err != nil {
+			return false, fmt.Errorf("error updating deployment of service '%s': %s", svcName, err.Error())
+		}
+		if _, err := deployments.Deploy(ctx, d, c); err != nil {
+			return false, fmt.Errorf("error updating deployment of service '%s': %s", svcName, err.Error())
+		}
```
ifbyol, line 511: This is odd, why if it failed because a temporal error redeploying? Any error in a redeploy would imply to destroy a deploy again even if it is not necessary, right?

teresaromero: you are right, in case of statefulsets we make this only when the error is not "Forbidden", do you think we should control other types of errorS?

teresaromero: @ifbyol i've reverted the change here and added an exception prior to deploying for first time to check if is the exception, if yes, then we destroy and redeploy

pkg/cmd/stack/deploy.go
```diff
@@ -488,14 +488,26 @@ func deployDeployment(ctx context.Context, svcName string, s *model.Stack, c kub
...
 			return false, fmt.Errorf("skipping deploy of deployment '%s' due to name collision with pre-existing deployment", svcName)
 		}
-		if old.Labels[model.StackNameLabel] != s.Name {
+		if old.Labels[model.StackNameLabel] != s.Name && old.Labels[model.StackNameLabel] != "okteto" {
 			return false, fmt.Errorf("skipping deploy of deployment '%s' due to name collision with deployment in stack '%s'", svcName, old.Labels[model.StackNameLabel])
 		}
 		if v, ok := old.Labels[model.DeployedByLabel]; ok {
+			if old.Labels[model.StackNameLabel] == "okteto" {
+				d.Labels[model.DeployedByLabel] = s.Name
+			}
```
ifbyol, line 497: Don't you need an else here? If you set the `deployed-by` label to the new name, the line 498 will overwrite it with the value coming from the old deployment, right?

Or moving the line 498 before doing the `okteto` check would be enough

pkg/cmd/stack/deploy.go
ifbyol: ok, I think now looks good.

Maybe we can include a comment before the if to explain why this piece of code is needed? If not, in a few months we won't remember why we do that check

teresaromero: i've added a comment with a link to this PR and some context

teresaromero:
@maroshii i've added both 2 labels: bugfix and breaking change... its definatly a bug fix as the users which use an .okteto folder for their compose and stacks will now have the dev environment name fixed, but at the same time, if someone had a dev environment deployed with prev version of cli, now the old one will be empty and they will se a new one with the correct name, not sure if this is a breking change, or "not that breaking" change.. but is a change :/

jLopezbarb (approved):

Issue #2219, opened 2022-02-02
Refactor: use the go docker-compose deserializer


pchico83:
**Is your feature request related to a problem? Please describe.**
As part of the effort to move to a 100% compose compatibility, we should move to the official compose deserializer:
https://github.com/docker/compose/tree/v2

We could keep using (and deprecate it at some point) our own serializer for `okteto-stack.***` files, which have a slightly different syntax/behavior than compose. For example, `commands` and `args` have different behaviors today depending on the name of the file.

jLopezbarb:
Just a quick update.
The compose serialised library that is used by https://github.com/docker/compose/tree/v2 is https://github.com/compose-spec/compose-go

jmacelroy:
@jLopezbarb I put this on a future release milestone. Help me evaluate when a good time to work on this would be as we do more release planning in the future.

jLopezbarb:
Could we put this one on 2.5? The last three regressions came because of this code, so having it as soon as possible can fix all the issues that we are having

pchico83:
@jLopezbarb @jmacelroy @ifbyol one option here is to use the official compose struct as the interface for `deploy/destroy` commands, and let our serializer generate this struct from our custom stack fields. This way, `deploy/deploy` don't need to handle stack-specific logic.
Ideally, the serializer could also reuse the official compose serializer, not sure if that should be done in a different PR

ifbyol:
> @jLopezbarb @jmacelroy @ifbyol one option here is to use the official compose struct as the interface for `deploy/destroy` commands, and let our serializer generate this struct from our custom stack fields. This way, `deploy/deploy` don't need to handle stack-specific logic. Ideally, the serializer could also reuse the official compose serializer, not sure if that should be done in a different PR

@pchico83  I'm not familiar with the code, but that sounds good to me. What I'm not sure what to do is how to know which serializer we should use depending on the manifest

pchico83:
This will be created in planning when needed

PR #2748, opened 2022-06-02
fix: translate stacks to compose ports when unmarshalling

### LINKED ISSUE #2747: Flexible Port Mapping Causes Incorrect Endpoint Creation.
**Describe the bug**
We are generating incorrect endpoints. We are generating an endpoint for the port 8080 and another one for the port 80 (including the port number in the endpoint) instead of just one endpoint. It was introduce with the fix for the property public in stacks.
This is a combination of a [change](https://github.com/okteto/okteto/issues/1584) we did in the past + the translations service -ingress is done in the CLI since 2.3.0 (this provoked the public property stopped working) and before it was done in the backend.

```
 movies:
    build: reverse-proxy
    ports:
      - 8080:80
```

**Expected behavior**
There should be one endpoint with a different local and remote port.

- OS: [e.g. macOS]

jLopezbarb:

Fixes #2747

## Proposed changes
- Translates stacks ports to compose syntax when unmarshalling

### DIFF
--- a/pkg/cmd/stack/translate.go
+++ b/pkg/cmd/stack/translate.go
@@ -420,7 +420,7 @@ func translateService(svcName string, s *model.Stack) *apiv1.Service {
 func getSvcPublicPorts(svcName string, s *model.Stack) []model.Port {
 	result := []model.Port{}
 	for _, p := range s.Services[svcName].Ports {
-		if !model.IsSkippablePort(p.ContainerPort) && (p.HostPort != 0 || s.Services[svcName].Public) {
+		if !model.IsSkippablePort(p.ContainerPort) && p.HostPort != 0 {
 			result = append(result, p)
 		}
 	}
--- a/pkg/cmd/stack/translate_test.go
+++ b/pkg/cmd/stack/translate_test.go
@@ -1568,6 +1568,7 @@ func TestGetSvcPublicPorts(t *testing.T) {
 						Ports: []model.Port{
 							{
 								ContainerPort: 80,
+								HostPort:      80,
 							},
 						},
 					},
--- a/pkg/model/stack.go
+++ b/pkg/model/stack.go
@@ -90,6 +90,7 @@ type Service struct {
 	Healtcheck      *HealthCheck          `yaml:"healthcheck,omitempty"`
 	User            *StackSecurityContext `yaml:"user,omitempty"`
 
+	// Fields only for okteto stacks
 	Public    bool            `yaml:"public,omitempty"`
 	Replicas  int32           `yaml:"replicas,omitempty"`
 	Resources *StackResources `yaml:"resources,omitempty"`
--- a/pkg/model/stack_serializer.go
+++ b/pkg/model/stack_serializer.go
@@ -566,6 +566,11 @@ func translateHealtcheckCurlToHTTP(healthcheck *HealthCheck) {
 func getSvcPorts(public bool, rawPorts, rawExpose []PortRaw) (bool, []Port, error) {
 	rawPorts = expandRangePorts(rawPorts)
 
+	// If public is set at this point it means that it comes from an okteto stacks and it's true
+	if public {
+		rawPorts = translateOktetoStacksPortsIntoComposeSyntax(rawPorts)
+	}
+
 	if !public && len(getAccessiblePorts(rawPorts)) == 1 {
 		public = true
 	}
@@ -595,6 +600,25 @@ func getSvcPorts(public bool, rawPorts, rawExpose []PortRaw) (bool, []Port, erro
 		}
 	}
 	return public, ports, nil
+}
+
+// translateOktetoStacksPortsIntoComposeSyntax translates okteto stack port syntax into compose ports syntax
+// We need this translation because we are using the same serializer for okteto stacks and docker compose syntax
+// Otherwise we don't know which ports should create an endpoint and which ones not
+// Example 1:
+// Stacks syntax  -> Compose syntax
+// ports:         | ports:
+//   - 5000       |   - 5000:5000
+//   - 3000       |   - 3000:3000
+// public: true   |
+func translateOktetoStacksPortsIntoComposeSyntax(ports []PortRaw) []PortRaw {
+	for idx, p := range ports {
+		if p.HostPort == 0 {
+			p.HostPort = p.ContainerPort
+		}
+		ports[idx] = p
+	}
+	return ports
 }
 
 func expandRangePorts(ports []PortRaw) []PortRaw {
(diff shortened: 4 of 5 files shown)


pkg/cmd/stack/translate.go
```diff
@@ -420,7 +420,7 @@ func translateService(svcName string, s *model.Stack) *apiv1.Service {
 func getSvcPublicPorts(svcName string, s *model.Stack) []model.Port {
 	result := []model.Port{}
 	for _, p := range s.Services[svcName].Ports {
-		if !model.IsSkippablePort(p.ContainerPort) && (p.HostPort != 0 || s.Services[svcName].Public) {
+		if !model.IsSkippablePort(p.ContainerPort) && p.HostPort != 0 {
```
jLopezbarb, line 423: This change was made in the `2.3.2` release and was causing the bug so we are reverting it

pkg/cmd/stack/translate_test.go
```diff
@@ -1568,6 +1568,7 @@ func TestGetSvcPublicPorts(t *testing.T) {
 						Ports: []model.Port{
 							{
 								ContainerPort: 80,
+								HostPort:      80,
```
jLopezbarb, line 1571: Reached this point we always need to have both ports set, because we have translated to a compose syntax so we need to have HostPort field defined

pkg/model/stack.go
```diff
@@ -90,6 +90,7 @@ type Service struct {
 	Healtcheck      *HealthCheck          `yaml:"healthcheck,omitempty"`
 	User            *StackSecurityContext `yaml:"user,omitempty"`
 
+	// Fields only for okteto stacks
```
AgustinRamiroDiaz, line 93: What does this mean? 
If it means that these fields are only used for stacks, should we create 2 structs for this? one for the `Service`, and a wrapper `StackService` which has the `Service` and also this fields?

pkg/model/stack_serializer_test.go
```diff
@@ -1853,3 +1853,84 @@ func Test_UnmarshalStackUser(t *testing.T) {
 		})
 	}
 }
+
+func Test_TranslateOktetoStackPortsToComposePorts(t *testing.T) {
```
AgustinRamiroDiaz, line 1857: 🔥

pkg/cmd/stack/translate.go
jmacelroy: Are we also reverting a fix? Can you link to what the associated issue was. I want to evaluate if we release without that fix or not.

pkg/model/stack.go
jmacelroy: I like the idea. It may change the way in which these fields are accessed.

jmacelroy (approved):
The only major comment is based around how reverting the public service code affects the release but still think we can go forward with this PR. I just need more info to make a release decision if you could link it or explain a bit.

pkg/model/stack_serializer.go
```diff
@@ -597,6 +602,25 @@ func getSvcPorts(public bool, rawPorts, rawExpose []PortRaw) (bool, []Port, erro
 	return public, ports, nil
 }
 
+// translateOktetoStacksPortsIntoComposeSyntax translates okteto stack port syntax into compose ports syntax
+// We need this translation because we are using the same serializer for okteto stacks and docker compose syntax
+// Otherwise we don't know which ports should create an endpoint and which ones not
+// Example 1:
+// Stacks syntax  -> Compose syntax
+// ports:         | ports:
+//   - 5000       |   - 5000:5000
```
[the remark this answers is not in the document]
jLopezbarb: only if the public is set to true. If public is set to true (by the user) it means that it has to create the endpoints, so we are converting it to a compose syntax.
If public is not set we don't convert any port so we can have a service with several ports and only one endpoint. For example:
```
services:
  api:
    ports:
    - 80:80
    - 5432
```
will create an endpoint for port 80 and will create a service (ClusterIP) with the ports 5432 and 80

pkg/model/stack_serializer.go
```diff
@@ -566,6 +566,11 @@ func translateHealtcheckCurlToHTTP(healthcheck *HealthCheck) {
 func getSvcPorts(public bool, rawPorts, rawExpose []PortRaw) (bool, []Port, error) {
 	rawPorts = expandRangePorts(rawPorts)
 
+	// If public is set at this point it means that it comes from an okteto stacks and it's true
+	if public {
```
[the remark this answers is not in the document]
jLopezbarb: public is a field on our okteto stacks manifest. It means to create an endpoint for a service. For example:
```
services:
  api:
    ports:
    - 80
    public: true
```
 is the way of creating an endpoint for the api service on okteto stacks.
 Full stacks reference is here: https://www.okteto.com/docs/0.10/reference/stacks/

pkg/model/stack_serializer.go
```diff
@@ -597,6 +602,25 @@ func getSvcPorts(public bool, rawPorts, rawExpose []PortRaw) (bool, []Port, erro
...
+// translateOktetoStacksPortsIntoComposeSyntax translates okteto stack port syntax into compose ports syntax
+// We need this translation because we are using the same serializer for okteto stacks and docker compose syntax
+// Otherwise we don't know which ports should create an endpoint and which ones not
+// Example 1:
+// Stacks syntax  -> Compose syntax
+// ports:         | ports:
+//   - 5000       |   - 5000:5000
+//   - 3000       |   - 3000:3000
+// public: true   |
+func translateOktetoStacksPortsIntoComposeSyntax(ports []PortRaw) []PortRaw {
```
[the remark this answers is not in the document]
jLopezbarb: I prefer to name it as translate, because we are not filling every port, we are just filling if is an okteto stack and the service is set to true, so we need to translate the ports to a compose syntax

pkg/model/stack.go
jLopezbarb: Yes, We should read each manifest (okteto stacks and compose) in a way that it doesn't collide with the other and translate them to a common struct to work afterward. I think that should be covered on  https://github.com/okteto/okteto/issues/2219

pkg/cmd/stack/translate.go
jLopezbarb: We are reverting https://github.com/okteto/okteto/pull/2735 with this change

ifbyol (approved):

pkg/model/stack_serializer.go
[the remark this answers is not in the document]
AgustinRamiroDiaz: @jLopezbarb I didn't understand. It's the same behavior whether the compose looks like this
```
services:
  api:
    ports:
    - 80:80
    - 5432
```

or this?
```
services:
  api:
    ports:
    - 80
    - 5432
```

PR #2749, opened 2022-06-02
Agustin/golang-ci-lint

AgustinRamiroDiaz:
## Proposed changes

- Address `golangci-lint` feedback

### DIFF
--- a/pkg/cmd/build/build.go
+++ b/pkg/cmd/build/build.go
@@ -251,7 +251,7 @@ func OptsFromBuildInfo(manifestName, svcName string, b *model.BuildInfo, o *type
 	}
 
 	outputMode := oktetoLog.GetOutputFormat()
-	if o != nil && o.OutputMode != "" {
+	if o.OutputMode != "" {
 		outputMode = o.OutputMode
 	}
 	opts.OutputMode = setOutputMode(outputMode)
--- a/pkg/cmd/build/build.go
+++ b/pkg/cmd/build/build.go
@@ -226,7 +226,7 @@ func OptsFromBuildInfo(manifestName, svcName string, b *model.BuildInfo, o *type
 	if okteto.Context().IsOkteto && b.Image == "" {
 		// if flag --global, point to global registry
 		targetRegistry := okteto.DevRegistry
-		if o != nil && o.BuildToGlobal {
+		if o.BuildToGlobal {
 			targetRegistry = okteto.GlobalRegistry
 		}
 		b.Image = fmt.Sprintf("%s/%s-%s:%s", targetRegistry, manifestName, svcName, model.OktetoDefaultImageTag)
--- a/pkg/model/stack_serializer_test.go
+++ b/pkg/model/stack_serializer_test.go
@@ -867,6 +867,9 @@ func Test_validateVolumesUnmarshalling(t *testing.T) {
 	}
 	defer os.RemoveAll("test_volume_relative_path_found")
 	file, err := os.Create("test-file")
+	if err != nil {
+		t.Fatal(err)
+	}
 	defer os.Remove(file.Name())
 	tests := []struct {
 		name                 string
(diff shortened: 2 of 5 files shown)


pkg/cmd/build/build.go
```diff
@@ -251,7 +251,7 @@ func OptsFromBuildInfo(manifestName, svcName string, b *model.BuildInfo, o *type
 	}
 
 	outputMode := oktetoLog.GetOutputFormat()
-	if o != nil && o.OutputMode != "" {
```
jmacelroy, line 254: I may be in favor of keeping this. We check of `o == nil` at the top of the function but it's a large function. What if that behavior or check changes or is removed? Maybe there should be a unittest for this behavior? What's the linting error?

pkg/model/stack_serializer_test.go
```diff
@@ -867,6 +867,9 @@ func Test_validateVolumesUnmarshalling(t *testing.T) {
 	}
 	defer os.RemoveAll("test_volume_relative_path_found")
 	file, err := os.Create("test-file")
+	if err != nil {
```
jmacelroy, line 870: What's the linting error here? This seems to modify our test assertions.

pkg/cmd/build/build.go
AgustinRamiroDiaz: The linting error is `possible nil pointer dereference`. I agree we should have a unit test, but I don't think it will be null at any point

pkg/model/stack_serializer_test.go
AgustinRamiroDiaz: We were not handling the error that came back from the `ok.Create` function (I think it was `unused`)

pkg/cmd/build/build.go
AgustinRamiroDiaz: If I understand correctly, the problem arises from checking `o != nil`, which expresses that `o` can be `nil` in the other places in this function. The linter suggests 2 outcomes:
- Handle nil
- Suppose that function will not recieve a nil

ifbyol: Why it doesn't complain on line `229`? It seems a similar case

```
if o != nil && o.BuildToGlobal {
```

pkg/model/stack_serializer_test.go
ifbyol: If we don't want to change the test check, other possible solution could be replace `err` by `_` and don't check the error.

Anyways, if this is some kind of precondition in our test, I think it is ok to check the error and fail

pkg/okteto/client.go
```diff
@@ -226,36 +223,3 @@ func (c *OktetoClient) Previews() types.PreviewInterface {
 func (c *OktetoClient) User() types.UserInterface {
 	return c.user
 }
-
-type traceTransport struct {
```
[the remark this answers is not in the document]
ifbyol: yeah, it seems it was added for helping us for debugging: https://github.com/okteto/okteto/pull/2545

pkg/cmd/build/build.go
AgustinRamiroDiaz: Playing around a bit more, seems like the linting is due to 2 things happening in the same scope:
- de referencing a pointer
- checking if the pointer is null

This gives the impression that one of the 2 should be changed:
- if the pointer **cannot** be null, then we **shouldn't** check at any time
- if the pointer **can** be null, we **must** check every time

@ifbyol Seems like the other case is similar but since its in another scope, and in that scope we don't do the 2 things mentioned at the beginning, it might be the case that the pointer could be nil only in that context (inside the if statement)

pkg/model/stack_serializer_test.go
AgustinRamiroDiaz: @ifbyol Exactly, as it's a precondition I would fail the test if we cannot create the file

pkg/okteto/client.go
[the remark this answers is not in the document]
AgustinRamiroDiaz: I added a rule to prevent linting on this lines

PR #2750, opened 2022-06-02
Add github action

AgustinRamiroDiaz:
## Proposed changes

- Add `golangci-lint` github action

### DIFF
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -1,0 +1,40 @@
+# See https://github.com/golangci/golangci-lint-action for more information
+name: golangci-lint
+on: [pull_request]
+permissions:
+  contents: read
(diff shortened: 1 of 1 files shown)


.github/workflows/golangci-lint.yml
```diff
@@ -0,0 +1,40 @@
+# See https://github.com/golangci/golangci-lint-action for more information
+name: golangci-lint
```
jmacelroy, line 2: I thought we may have already be doing this somewhere else. Either as part of some other meta linter or just this. It could be that the cache just needs to be reset on the other job?

AgustinRamiroDiaz: It might be the cache, I'll keep reading

AgustinRamiroDiaz: You were right, we are running it with https://github.com/github/super-linter

PR #2751, opened 2022-06-03
Added condition to skip lines starting with exclamation

### LINKED ISSUE #2679: Weird behavior with .stignore patterns
**Describe the bug**
A user stated that they get `(?d)` as a prefix in every line in their `.stignore`, even in lines starting with `#` or `!`
[context here](https://community.okteto.com/t/stignore-prefixed-d-for-all-patterns-even-if-include-file-or-exclude-patterns/78) 

**To Reproduce**

1. Create an `.stignore` with some patterns
2. Run `okteto up`
3. Run `cat .stignore`

OS: Windows10

teresaromero:

Fixes #2679 

## Proposed changes

- Added the skip condition when parsing the .stignore file - when starts with `!` do not prepend (?d)

## Context

Using the repo from @AgustinRamiroDiaz https://github.com/AgustinRamiroDiaz/okteto-stignore i've been going around the different possibilities. The remote `.stignore` is being parsed from the local one, we add the prefix (?d) in order to be able to delete a directory. Lines regarding `#includes` or `!<file or folder>` where not copied to this remote `.stignore`, i thought this had to be copy to the remote file but when i did, syncing got in an infite loop whith the following error:

```
WARNING: Error on folder \"1\" (okteto-1): loading ignores: parse error: failed to load include file .numbers: open /okteto/.numbers: no such file or directory" process=syncthing
```

So decided to just avoid lines starting with `!` as we do with `#`, as, `synthing` is using the local rules, although the remote rules are different.

### DIFF
--- a/cmd/up/stignore.go
+++ b/cmd/up/stignore.go
@@ -80,6 +80,9 @@ func addStignoreSecrets(dev *model.Dev) error {
 			if strings.HasPrefix(line, "#") {
 				continue
 			}
+			if strings.HasPrefix(line, "!") {
+				continue
+			}
 
 			_, err = writer.WriteString(fmt.Sprintf("(?d)%s\n", line))
 			if err != nil {
--- a/cmd/up/stignore.go
+++ b/cmd/up/stignore.go
@@ -15,7 +15,7 @@ package up
 
 import (
 	"bufio"
-	"crypto/md5"
+	"crypto/sha512"
 	"encoding/json"
 	"fmt"
 	"io"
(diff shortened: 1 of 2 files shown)


cmd/up/stignore.go
```diff
@@ -80,6 +80,9 @@ func addStignoreSecrets(dev *model.Dev) error {
 			if strings.HasPrefix(line, "#") {
 				continue
 			}
+			if strings.HasPrefix(line, "!") {
```
pchico83, line 83: Could we add tests for this function?

jLopezbarb:
What about lines starting with `#include`. The user said that it was failing to include from other files. Is it working?

AdrianPedriza (approved):

llindemann:
since this change, the following local .stignore file doesn't work for me:

!/src/
!/public/

*

pre 2.4.0 /src and /public were synced and everything else ignored

teresaromero:
@llindemann a fix was released with 2.4.1

PR #2755, opened 2022-06-06
Jlopezbarb/refactor tests e2e

### LINKED ISSUE #2730: Refactor e2e tests
e2e tests time has gone to more than 1 hour, and we have detected several regressions that should have been caught by the tests.

We need to refactor the test to cover more scenarios while reducing the dependencies with our own repos (we were cloning okteto samples to run e2e tests).

- **actions**
- [ ] If detect a version of cli with actions run them with that version

- **Build**
- [ ] Create a make command to run build e2e tests
- [ ] Test covering Build V2
- [ ] Test covering Build V2 defining a service
- [ ] Test covering Build V2 defining several services
- [ ] Test covering Build from a compose with volume mounts
- [x] Test covering Build V1

- **Deploy**
- [ ] Create a make command to run deploy e2e tests
- [ ] Test covering auto wake features
- [ ] Test covering Deploying from a pipeline manifest
- [ ] Test covering Deploying from a manifestV2 manifest
- [ ] Test covering Deploying from a helm chart
- [ ] Test covering Deploying from a k8s folder
- [ ] Test covering Deploying from a compose file
- [ ] Test covering Deploying from a compose file create the correct images
- [ ] All test should also test destroy command

- **Up**
- [ ] Create a make command to run up e2e tests
- [ ] Test covering Running an okteto up manifest v1 deployment
- [ ] Test covering Running an okteto up manifest v1 sfs
- [ ] Test covering Running an okteto up manifest v1 autocreate
- [ ] Test covering Running an okteto up manifest v2 deployment
- [ ] Test covering Running an okteto up manifest v2 sfs
- [ ] Test covering Running an okteto up manifest v2 autocreate
- [ ] Test covering Running an okteto up manifest v2 with deploy from the command

- **stack** (will be removed in a future)
- [ ] Create a make command to run stack e2e tests
- [ ] Test covering Deploying an okteto stack
- [ ] Test covering Deploying a compose file
- [ ] Test covering Deploying a compose file with volume mounts
- [ ] Test covering Deploying an okteto stack with secrets

- **push** (will be removed in a future)
- [ ] Create a make command to run push e2e tests
- [ ] Test covering push

jLopezbarb:
Fixes #2730

## Proposed changes
- Removes all the dependencies that we have from git (we were cloning okteto sample repositories to test the CLI). This way the CLI has more control on what it's tested. For example we were testing the stacks-getting-started repository but it's no longer an okteto stacks file so we weren't testing the stacks manifest.
- All the repos that we were testing have been reviewed to know all the features that we were testing and created simplified versions of it, that way we test the same features we were testing but the tests runs faster
- Adds scenarios that we were not testing. For example:
  -  Build with manifest v2
  - Okteto deploy inferring all the possible manifests (helm repo, k8s file, compose,okteto pipeline)
  - okteto stacks (we weren't testing at all)
  - deploying only selected services

- Going from having all the tests into a single package to having the e2e tests grouped by their command
```
├── actions (Contains all the action tests)
├── build (Contains build v1 and build v2 tests
├── commands (Contains the functions to run the commands on the tests)
├── deploy (deploy related tests)
├── deprecated (tests for deprecated commands)
│   ├── push (e2e tests)
│   └── stack (stacks command tests)
├── git.go
├── k8s-client.go
├── okteto (tests for features only available on okteto clusters)
├── okteto.go
└── up (okteto up tests)
```
- Adds make commands to run only tests of a kind:
```
make integration (runs all the e2e tests)
make integration-actions (runs actions e2e tests)
make integration-build (runs build e2e tests)
make integration-deploy (runs deploy e2e tests)
make integration-okteto (runs okteto clusters e2e tests)
make integration-up (runs okteto up e2e tests)
make integration-deprecated (runs deprecated command e2e tests)
```


integration/actions/pipeline_test.go
```diff
@@ -0,0 +1,119 @@
...
+	if err := integration.CloneGitRepoWithBranch(actionRepo, oktetoVersion); err != nil {
+		if err := integration.CloneGitRepo(actionRepo); err != nil {
+			return err
+		}
+		log.Printf("cloned repo %s main branch\n", actionRepo)
+	}
+	log.Printf("cloned repo %s \n", actionRepo)
+	defer integration.DeleteGitRepo(actionFolder)
+
+	t.Setenv(model.GithubRepositoryEnvVar, pipelineRepo)
```
ifbyol, line 69: This test still relies on movies repo. Any change on the movies repo could make this tests fail. Mentioning just to highlight that we still have some dependency for this repo. For example, this repo would be broken if we remove the `master` branch again from the remote

integration/actions/preview_test.go
```diff
@@ -0,0 +1,109 @@
...
+			return err
+		}
+		log.Printf("cloned repo %s main branch\n", actionRepo)
+	}
+	log.Printf("cloned repo %s \n", actionRepo)
+	defer integration.DeleteGitRepo(actionFolder)
+
+	log.Printf("Deploying preview %s", namespace)
+	command := oktetoPath
+	args := []string{"preview", "deploy", namespace, "--scope", "personal", "--branch", "master", "--repository", fmt.Sprintf("%s%s", githubHTTPSURL, pipelineRepo), "--wait"}
```
ifbyol, line 64: Same about the `movies` repo.

We might consider to create a specific repo (maybe a fork) for these tests. We can include something in the readme to make clear that the repo is used for running tests and we should be careful when doing changes

integration/deploy/compose_test.go
```diff
@@ -0,0 +1,308 @@
...
+	"github.com/okteto/okteto/pkg/registry"
+	"github.com/stretchr/testify/require"
+	k8sErrors "k8s.io/apimachinery/pkg/api/errors"
+)
+
+const (
+	composeTemplate = `services:
+  app:
+    build: app
+    entrypoint: echo -n $RABBITMQ_PASS > var.html && python -m http.server 8080
```
ifbyol, line 38: Where is the `RABBITMQ_PASS` environment variable defined? In out tests or in the Okteto instance where the tests is running?

If it is defined in the Okteto instance, It is sensible to fail if anyone change or remove that (any Okteto admin can do it). I think we should document all the external dependencies of that kind we have and be strict on updating it. 

Other possibility is to have a standard for this case and define these variable as `E2E_TEST_RABBITMQ_PASS`. Including the prefix you give some context to indicate that those variable are being used for e2e tests

integration/deploy/compose_test.go
```diff
@@ -0,0 +1,308 @@
...
+      app:
+        condition: service_started
+    container_name: web-svc
+    healthcheck:
+      test: service nginx status || exit 1
+      interval: 45s
+      timeout: 5m
+      retries: 5
+      start_period: 30s`
+	stacksTemplate = `services:
```
ifbyol, line 61: I would include a blank line here to separate 2 manifests. I think it reads better

integration/deploy/deploy_test.go
```diff
@@ -0,0 +1,114 @@
+//go:build integration
```
ifbyol, line 1: Would it make sense to call this file `k8s_manifest_test.go`? the rest of files within this package makes reference to the kind/source of the app to be deployed

integration/deploy/k8s_test.go
```diff
@@ -0,0 +1,53 @@
...
+
+	"github.com/okteto/okteto/integration"
+	"github.com/okteto/okteto/integration/commands"
+	"github.com/stretchr/testify/require"
+)
+
+// TestDeployPipelineFromK8s tests the following scenario:
+// - Deploying a pipeline manifest locally from a k8s file
+// - The endpoints generated are accessible
+func TestDeployPipelineFromK8s(t *testing.T) {
```
ifbyol, line 31: Why pipeline if it is using `okteto deploy` command?

integration/actions/pipeline_test.go
jLopezbarb: Pipeline and preview commands needs a URL link to deploy the application. Those are the only two commands that relies on an okteto repo. Our most stable sample is movies, this is way it's used.

integration/actions/preview_test.go
jLopezbarb: Same as above

integration/deploy/compose_test.go
jLopezbarb: I need to create a document on how to run e2e tests locally for defining all these scenarios. `RABBITMQ_PASS` is a user level secret (I'm thinking now that testing secret injection should be run on the okteto folder, so we can run all the other tests on vanilla clusters). I haven't change the name because it's already defined in `cindylopez` user (the one that we use to run e2e tests)

jLopezbarb:
> It would be nice to have all the tested scenarios documented or accesible in a more user friendly way that reading test names or comments.

How do you recommend doing this? I think it should be addressed on this PR. I was planning to create the names in TestWhenThen format, but the scenario is so big I can't describe all the things that are tested on each test

AgustinRamiroDiaz:
> > It would be nice to have all the tested scenarios documented or accesible in a more user friendly way that reading test names or comments.
> 
> How do you recommend doing this? I think it should be addressed on this PR. I was planning to create the names in TestWhenThen format, but the scenario is so big I can't describe all the things that are tested on each test

That naming could be usefull for [subtests](https://go.dev/blog/subtests) and then having a generic name in the test.

integration/actions/pipeline_test.go
ifbyol: it is the most stable repo but this week tests were broken because we have removed a remote branch. It is even using an old version of the movies app not the one we are actively using and maintaining.

I would say to create a fork or another repository called `e2e test application` or similar and start using that one. Documenting it properly in the README

integration/deploy/compose_test.go
ifbyol: But we can create the new secret in `cindylopez` and start using it.

And I would try to avoid to rely on a specific user (or info from the user) to run these tests. That makes tests error prone. An option could be to define a local `.env` in the test with the right value. In that way, you can also test secret injection and it would be a valid scenario for vanilla clusters

integration/deploy/k8s_test.go
ifbyol: The term we use now is dev environment. Using pipeline might be confusing because the reader could think that we are deploying the app using `okteto pipeline deploy` with a repo but we are using `okteto deploy` if I'm not wrong. 

I want to avoid the confusion `okteto pipeline deploy` vs `okteto deploy`

ifbyol:
> > It would be nice to have all the tested scenarios documented or accesible in a more user friendly way that reading test names or comments.
> 
> How do you recommend doing this? I think it should be addressed on this PR. I was planning to create the names in TestWhenThen format, but the scenario is so big I can't describe all the things that are tested on each test

Ideally, it should be something automatic, so if you change the test or the "documentation" you are forced to change the other. For these kind of tests, using a BDD approach/framework is usually the best solution because you write the story/scenario you want to test and then, map it to test functions. You have an example for golang [here](https://github.com/cucumber/godog)

But that would imply bigger changes and maybe change the way we do the tests so I wouldn't do it in this PR. We can start defining the use cases manually and try to be strict on updating them when something is changed, added or removed

integration/deploy/compose_test.go
jmacelroy: +1 we should allow these to be run on any cluster with any user.

integration/actions/actions_test.go
```diff
@@ -0,0 +1,67 @@
...
+	if runtime.GOOS == "windows" {
+		kubectlBinary = "kubectl.exe"
+	}
+
+	versionOutput, err := integration.RunOktetoVersion("okteto")
+	if err != nil {
+		log.Println("okteto binary not found")
+		os.Exit(1)
+	}
+	r, _ := regexp.Compile(versionRegex)
```
AgustinRamiroDiaz, line 63: We could add it as a var at the top
```
var (
	versionRegexCompiled = regexp.MustCompile(versionRegex)
)
```

teresaromero (approved):
👏🏻 👏🏻  amazing refactor! Thanks @jLopezbarb

ifbyol (approved):
Nice work!

PR #2763, opened 2022-06-08
test: use `T.TempDir` to create temporary test directory

Juneezee:
## Proposed changes
A testing cleanup. 

This pull request replaces `os.MkdirTemp` with `t.TempDir`. We can use the `T.TempDir` function from the `testing` package to create temporary directory. The directory created by `T.TempDir` is automatically removed when the test and all its subtests complete. 

This saves us at least 2 lines (error check, and cleanup) on every instance, or in some cases adds cleanup that we forgot.

Reference: https://pkg.go.dev/testing#T.TempDir

```go
func TestFoo(t *testing.T) {
	// before
	tmpDir, err := os.MkdirTemp("", "")
	if err != nil {
		t.Fatal(err)
	}
	defer os.RemoveAll(tmpDir)

	// now
	tmpDir := t.TempDir()
}
```

### DIFF
--- a/cmd/build/v2/build_test.go
+++ b/cmd/build/v2/build_test.go
@@ -119,11 +119,7 @@ func TestOnlyInjectVolumeMountsInOkteto(t *testing.T) {
 		},
 		CurrentContext: "test",
 	}
-	dir, err := os.MkdirTemp("", "test")
-	if err != nil {
-		t.Fatal(err)
-	}
-	defer os.RemoveAll(dir)
+	dir := t.TempDir()
 
 	registry := test.NewFakeOktetoRegistry(nil)
 	builder := test.NewFakeOktetoBuilder(registry)
@@ -169,14 +165,9 @@ func TestTwoStepsBuild(t *testing.T) {
 		},
 		CurrentContext: "test",
 	}
-	dir, err := os.MkdirTemp("", "test")
-	if err != nil {
-		t.Fatal(err)
-	}
-	defer os.RemoveAll(dir)
-	dir, err = createDockerfile()
+
+	dir, err := createDockerfile(t)
 	assert.NoError(t, err)
-	defer os.RemoveAll(dir)
 
 	registry := test.NewFakeOktetoRegistry(nil)
 	builder := test.NewFakeOktetoBuilder(registry)
@@ -227,9 +218,8 @@ func TestBuildWithoutVolumeMountWithoutImage(t *testing.T) {
 		CurrentContext: "test",
 	}
 
-	dir, err := createDockerfile()
+	dir, err := createDockerfile(t)
 	assert.NoError(t, err)
-	defer os.RemoveAll(dir)
 
 	registry := test.NewFakeOktetoRegistry(nil)
 	builder := test.NewFakeOktetoBuilder(registry)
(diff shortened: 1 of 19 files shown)


cmd/build/v2/build_test.go
```diff
@@ -370,13 +358,10 @@ func Test_getAccessibleVolumeMounts(t *testing.T) {
...
-func createDockerfile() (string, error) {
-	dir, err := os.MkdirTemp("", "build")
-	if err != nil {
-		return "", err
-	}
+func createDockerfile(t *testing.T) (string, error) {
+	dir := t.TempDir()
 	dockerfilePath := filepath.Join(dir, "Dockerfile")
-	err = os.WriteFile(dockerfilePath, []byte("Hello"), 0755)
+	err := os.WriteFile(dockerfilePath, []byte("Hello"), 0755)
```
AgustinRamiroDiaz, line 364: We could make use of the `t *testing.T` that we have here to fail the test

cmd/build/v2/build_test.go
```diff
@@ -370,13 +358,10 @@ func Test_getAccessibleVolumeMounts(t *testing.T) {
 	assert.Len(t, volumes, 1)
 }
 
-func createDockerfile() (string, error) {
-	dir, err := os.MkdirTemp("", "build")
-	if err != nil {
-		return "", err
-	}
+func createDockerfile(t *testing.T) (string, error) {
+	dir := t.TempDir()
```
AgustinRamiroDiaz, line 362: If we only want to use the `TempDir()` function, we could reduce the dependencies to `func createDockerfile(tempDir func() string) (string error)` and pass `t.TempDir` as parameter

AgustinRamiroDiaz (approved):

AgustinRamiroDiaz:
Thanks for your submission! Good work 😄

jLopezbarb:
Hi @Juneezee, thanks for your submission but we already tried to do this and we faced some windows problems when cleaning the temp directories after running the tests which leads into failed tests
We are waiting go to release `1.18.3` to fix it on windows. Here is the cherrypick that fixes it https://github.com/golang/go/issues/52986 but we don't have an ETA for this to be merge and for us to be updating go to 1.18.3

Juneezee:
> Hi @Juneezee, thanks for your submission but we already tried to do this and we faced some windows problems when cleaning the temp directories after running the tests which leads into failed tests We are waiting go to release `1.18.3` to fix it on windows. Here is the cherrypick that fixes it [golang/go#52986](https://github.com/golang/go/issues/52986) but we don't have an ETA for this to be merge and for us to be updating go to 1.18.3

@jLopezbarb Looking at the logs [here](https://app.circleci.com/pipelines/github/okteto/okteto/7220/workflows/e9ed0fda-8732-492c-9925-a776975cf283/jobs/18268) and the corresponding test code, it looks like we didn't close the `os.File` properly which is why the test failed. Unix permits removing an open file, Windows does not (https://github.com/golang/go/issues/50510#issuecomment-1008535852).

I'm going to push a commit that closes the `os.File` properly. I don't have a Windows machine so I am unable to test it locally, is there a way to enable the `windows-unit` job for a PR?

jLopezbarb:
> @jLopezbarb Please take a look at [9c73eb2](https://github.com/okteto/okteto/commit/9c73eb217b21ac2040e18141ca7a26f93989097c). This commit should fix the 4 failing tests on Windows.

Yeah, that made the trick.
Awesome PR, thanks a lot @Juneezee for submitting this!!!

jLopezbarb (approved):

pchico83 (approved):
Thanks @Juneezee for taking care of this, great contribution!

PR #2764, opened 2022-06-08
creating manifest if deploy fails

### LINKED ISSUE #2555: `okteto deploy` isn't storing my manifest in the configmap is `okteto deploy` fails
**Describe the bug**
I run `okteto deploy` and it fails. The Okteto UI isn't showing the manifest.
This makes it harder to debug issues, especially when sharing my namespace with a teammate.

mnevadom:
Fixes #2555


jLopezbarb (approved):

cmd/deploy/deploy.go
```diff
@@ -303,61 +324,48 @@ func (dc *DeployCommand) RunDeploy(ctx context.Context, deployOptions *Options)
...
-			return err
-		}
+		errBuild = dc.Builder.Build(ctx, buildOptions)
 	} else {
-		svcsToBuild, err := dc.Builder.GetServicesToBuild(ctx, deployOptions.Manifest, deployOptions.servicesToDeploy)
-		if err != nil {
-			return err
-		}
-		if len(svcsToBuild) != 0 {
+		svcsToBuild, errBuild := dc.Builder.GetServicesToBuild(ctx, deployOptions.Manifest, deployOptions.servicesToDeploy)
```
jmacelroy, line 345: Using the same error in multiple checks is very prone to bugs. I'd like to see us find a pattern that doesn't use such a largely scoped error.

I believe the else here is handled right away and can be a new var.

Also, I think we can probably make the error handling a function and handle the de-duplication of error handling by calling a function rather than a hoisted error that spans multiple if/else.

cmd/deploy/deploy.go
```diff
@@ -303,61 +324,48 @@ func (dc *DeployCommand) RunDeploy(ctx context.Context, deployOptions *Options)
 			SkipIfExists: !deployOptions.Dependencies,
 		}
 		if err := pipelineCMD.ExecuteDeployPipeline(ctx, pipOpts); err != nil {
+			if errStatus := updateConfigMapStatus(ctx, cfg, c, data, err); err != nil {
```
jmacelroy, line 327: Would the original execution error for the pipeline be lost if we return `errStatus`? Should we be using `wraps` What info does the user get about these errors? Do they get any info that a manifest is going to be created?

ifbyol: Apart from Jacob's questions, I think this condition is not right, we should check `errStatus != nil` instead of `err != nil`.

I think we should not return `errStatus` that is for something "internal". We could not update the configmap which is the internal resource we use to handle deployments. The user should see the error about the pipeline execution

cmd/deploy/deploy.go
```diff
@@ -303,61 +324,48 @@ func (dc *DeployCommand) RunDeploy(ctx context.Context, deployOptions *Options)
...
+		if len(svcsToBuild) != 0 && errBuild != nil {
 			buildOptions := &types.BuildOptions{
 				CommandArgs:  svcsToBuild,
 				EnableStages: true,
 				Manifest:     deployOptions.Manifest,
 			}
-			if err := dc.Builder.Build(ctx, buildOptions); err != nil {
-				return err
-			}
+			errBuild = dc.Builder.Build(ctx, buildOptions)
```
jLopezbarb, line 352: This `errBuild` refers to the one inside the else clausure, but we are checking it from outside the if else statement. Can you check it out pls

cmd/deploy/deploy.go
```diff
@@ -432,6 +440,16 @@ func (dc *DeployCommand) RunDeploy(ctx context.Context, deployOptions *Options)
 	return err
 }
 
+func updateConfigMapStatus(ctx context.Context, cfg *v1.ConfigMap, c kubernetes.Interface, data *pipeline.CfgData, err error) error {
+	oktetoLog.AddToBuffer(oktetoLog.ErrorLevel, err.Error())
+	data.Status = pipeline.ErrorStatus
+	if err := pipeline.UpdateConfigMap(ctx, cfg, data, c); err != nil {
```
jLopezbarb, line 446:
```suggestion
	return pipeline.UpdateConfigMap(ctx, cfg, data, c)
```

jLopezbarb:
Can we add an unit tests that checks that the configmap is updated when a build fails?

ifbyol (approved):
Agree with Javi on creating the test he mentioned in his last comment

cmd/deploy/deploy.go
```diff
@@ -426,12 +445,28 @@ func (dc *DeployCommand) RunDeploy(ctx context.Context, deployOptions *Options)
...
+}
+
+func updateConfigMapStatus(ctx context.Context, cfg *corev1.ConfigMap, c kubernetes.Interface, data *pipeline.CfgData, err error) error {
+	oktetoLog.AddToBuffer(oktetoLog.ErrorLevel, err.Error())
+	data.Status = pipeline.ErrorStatus
+	if err := pipeline.UpdateConfigMap(ctx, cfg, data, c); err != nil {
+		return err
+	}
+
+	return nil
```
ifbyol, line 467:
```suggestion
	return pipeline.UpdateConfigMap(ctx, cfg, data, c)
```

cmd/deploy/deploy_test.go
```diff
@@ -149,6 +149,70 @@ func TestDeployWithErrorChangingKubeConfig(t *testing.T) {
...
+			"icon":       "https://apps.okteto.com/movies/icon.png",
+			"name":       "Name",
+			"output":     "",
+			"repository": "https://github.com/okteto/movies",
+			"status":     "progressing",
+			"yaml":       "aWNvbjogaHR0cHM6Ly9hcHBzLm9rdGV0by5jb20vbW92aWVzL2ljb24ucG5nCmRlcGxveToKICAgIC0gb2t0ZXRvIGJ1aWxkIC10IG9rdGV0by5kZXYvYXBpOiR7T0tURVRPX0dJVF9DT01NSVR9IGFwaQogICAgLSBva3RldG8gYnVpbGQgLXQgb2t0ZXRvLmRldi9mcm9udGVuZDoke09LVEVUT19HSVRfQ09NTUlUfSBmcm9udGVuZAogICAgLSBoZWxtIHVwZ3JhZGUgLS1pbnN0YWxsIG1vdmllcyBjaGFydCAtLXNldCB0YWc9JHtPS1RFVE9fR0lUX0NPTU1JVH0KZGV2czoKICAgIC0gYXBpL29rdGV0by55bWwKICAgIC0gZnJvbnRlbmQvb2t0ZXRvLnltbA==",
+		},
+	}
+
+	currentCfg, err := getConfigMapFromData(ctx, data, fakeClient)
```
ifbyol, line 211:
```suggestion
	currentCfg, _ := getConfigMapFromData(ctx, data, fakeClient)
```

PR #2769, opened 2022-06-09
use OKTETO_VERSION in get-okteto.sh if provided

### LINKED ISSUE #2765: Add versions selection to get.okteto.com
Right now https://get.okteto.com selects the latest binary from a channel and there's no way to select a specific version.

~~Possible DX:~~

1. ~~Have `https://get.okteto.com/{version}` for every versions released. Probably unmaintainable if we want to  keep doing it statically. It would require to upload a copy of the installation script along side the binaries into for eg `https://get.okteto.com/2.2.2`. This script will have the version hardcoded~~
2. ~~Have a server (nginx or the like) that we can use dynamically return the script. This can enable us to have metrics around downloads, etc but it needs to be maintained.~~

~~Other ideas?~~

 cc @pchico83

maroshii:
fixes #2765

Once approved and merged, we need to upload to get.okteto.com


pchico83 (approved):
Could we also update/remove https://github.com/okteto/okteto/blob/master/docs/how-to-install-okteto-cli.md and https://github.com/okteto/okteto/blob/master/docs/RELEASE.md?
The first doc seems outdated, the second one should mention this and how to install the beta/dev channel too. Also, we should link this doc from the root README.md

pchico83:
@maroshii we can also remove [this file](https://github.com/okteto/okteto/blob/master/scripts/get-okteto-beta.sh)

jLopezbarb (approved):

maroshii:
Address feedback and opened PR in docs with this change: https://github.com/okteto/docs/pull/117

pchico83 (approved):

Issue #2770, opened 2022-06-09
dead code related to open forward without using ssh


AdrianPedriza:
**Describe the bug**
At this point it is no longer necessary to distinguish between opening the forward using ssh or not. We will always use ssh.
https://github.com/okteto/okteto/blob/master/cmd/up/forwards.go#L34

PR #2771, opened 2022-06-09
feat: support global forwarding

### LINKED ISSUE #2709: Global port forwards in okteto manifest
I have an okteto manifest with several dev containers. In every dev container, I need to expose my database ports to access my database on localhost. This makes it impossible to run two dev containers in parallel because of port collisions.

This problem is especially relevant when using hybrid development with `reverse` because I need to be able to access all my third-party dependencies on localhost.

A section in the okteto manifest for global forwards, something like this:

```
forward:
  - 6379:redis:6379
  - 27017:mongo:27017
  - 3306:mysql:3306
  - 9092:kafka:9092
dev:
  api:
    reverse:
      - 3000:3000

  frontend:
    reverse:
      - 8080:3000
```

instead of what we have today:
```
dev:
  api:
    forward:
      - 6379:redis:6379
      - 27017:mongo:27017
      - 3306:mysql:3306
      - 9092:kafka:9092
    reverse:
      - 3000:3000

  frontend:
      forward:
      - 6379:redis:6379
      - 27017:mongo:27017
      - 3306:mysql:3306
      - 9092:kafka:9092
      reverse:
          - 8080:3000
```

Any `okteto up` command will enforce the global forwards are working. 

**Additional context**
In other to avoid running a daemon in the developer machine, `okteto up` could run a goroutine to check if the global forwards are running, and create them otherwise. It requires a goroutine in a loop because if an `okteto up` creates the global forwards, then there is a second `okteto up`, and then, the first `okteto up` exits, the second `okteto up` should regenerate the global forwards.

Also, port collisions should be ignored when another `okteto up` process is running the same okteto manifest, but if the port is in-use by another program, `okteto up` should fail.

AdrianPedriza:
Fixes #2709

## Proposed changes

- Added new field to manifest: **forward** at the root level. This field will be use to set global forwards to use for all my dev environments actived. 
- If an `okteto up` session cannot set the global forward due to port collision (another `okteto up` session already did it), the session will be retrying to open the global forward until it gets done or until finish the `okteto up` process.


cmd/up/forwards_test.go
```diff
@@ -0,0 +1,112 @@
...
+		{
+			name:           "not needed global forwarding",
+			manifestInfo:   []model.Forward{},
+			expectedAnswer: false,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			answer := isNeededGlobalForwarder(tt.manifestInfo)
+			if answer != tt.expectedAnswer {
```
jLopezbarb, line 36: I think we agreed on start using assertify library for testing. Tests looks much cleaner :)

cmd/up/forwards.go
```diff
@@ -105,5 +118,98 @@ func (up *upContext) sshForwards(ctx context.Context) error {
...
+	}
+
+	return nil
+}
+
+func (up *upContext) setGlobalForwardsIfRequiredLoop(ctx context.Context) {
+	ticker := time.NewTicker(time.Second)
+
+	for {
+		if !isNeededGlobalForwarder(up.Manifest.GlobalForward) {
```
jLopezbarb, line 138: Should we check this condition outside the for loop? I think `up.Manifest.GlobalForward` changes over the time

jLopezbarb:
LGTM but I need to tests some scenarios before approving it

cmd/up/forwards.go
AdrianPedriza: we need to check it periodically because maybe we were not able to open all forwarders (due to port collision)

pkg/model/serializer_test.go
```diff
@@ -1239,9 +1239,10 @@ dev:
 sync:
   - app:/app`),
 			expected: &Manifest{
-				Build:        map[string]*BuildInfo{},
-				Deploy:       &DeployInfo{},
-				Dependencies: map[string]*Dependency{},
+				Build:         map[string]*BuildInfo{},
```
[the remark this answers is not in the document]
AdrianPedriza: global forwards shares the serializer with forwards. Do you think it is necessary to have one of your own?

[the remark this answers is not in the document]
pchico83: I would separate them. At some point, global forwards should only support the `localPort:serviceName:remotePort` +. the extended notation, and dev container forwards should only support `localPort:remotePort`

pchico83 (approved):
Great work @AdrianPedriza !

pkg/model/global_forward.go
```diff
@@ -0,0 +1,101 @@
...
+	if err != nil {
+		return err
+	}
+	gf.Local = rawGlobalForward.Local
+	gf.Remote = rawGlobalForward.Remote
+	gf.ServiceName = rawGlobalForward.ServiceName
+	gf.Labels = rawGlobalForward.Labels
+
+	if gf.Labels != nil && gf.ServiceName != "" {
+		return fmt.Errorf("Can not use ServiceName and Labels to specify the service.\nUse either the service name or labels to get the service to expose.")
```
pchico83, line 98: Could we add a link to the docs?
@jmacelroy could you review the message?

pchico83 (approved):

cmd/up/forwards.go
```diff
@@ -105,5 +118,89 @@ func (up *upContext) sshForwards(ctx context.Context) error {
 		return fmt.Errorf("failed to add entry to your SSH config file")
 	}
 
-	return up.Forwarder.Start(up.Pod.Name, up.Dev.Namespace)
+	err := up.Forwarder.Start(up.Pod.Name, up.Dev.Namespace)
```
jmacelroy, line 121: This is duplicated code that could lead to bugs possibly if one is updated and not the other. Can we use a function?

cmd/up/forwards.go
```diff
@@ -105,5 +118,89 @@ func (up *upContext) sshForwards(ctx context.Context) error {
...
+				up.GlobalForwarderStatus <- err
+				return
+			}
+
+			err = up.Forwarder.StartGlobalForwarding()
+			if err != nil {
+				up.GlobalForwarderStatus <- err
+				return
+			}
+		case <-ctx.Done():
```
jmacelroy, line 155: When is this context signaled to be done? We need to double check that we are gracefully closing this loop.

pkg/k8s/apps/crud.go
```diff
@@ -48,6 +48,30 @@ func Get(ctx context.Context, dev *model.Dev, namespace string, c kubernetes.Int
 	return &StatefulSetApp{sfs: sfs}, nil
 }
 
+// Get returns the dev pod for an app and loops until it success
+func Get(ctx context.Context, namespace, appName string, c kubernetes.Interface) (App, error) {
```
jmacelroy, line 52: Why is this function changing? If I read the diff correctly you've renamed `Get` to `GetByDev` but then this method looks modified. Who's calling it now?

pkg/k8s/apps/crud.go
```diff
@@ -48,6 +48,30 @@ func Get(ctx context.Context, dev *model.Dev, namespace string, c kubernetes.Int
...
+	d, err := deployments.Get(ctx, appName, namespace, c)
+	if err != nil {
+		return nil, err
+	}
+
+	if err == nil {
+		return &DeploymentApp{d: d}, nil
+	}
+
+	if d == nil {
```
jmacelroy, line 62: This code is unreachable, correct?

pkg/k8s/apps/crud.go
```diff
@@ -48,6 +48,30 @@ func Get(ctx context.Context, dev *model.Dev, namespace string, c kubernetes.Int
...
 }
 
+// Get returns the dev pod for an app and loops until it success
+func Get(ctx context.Context, namespace, appName string, c kubernetes.Interface) (App, error) {
+	d, err := deployments.Get(ctx, appName, namespace, c)
+	if err != nil {
+		return nil, err
+	}
+
+	if err == nil {
```
jmacelroy, line 58: This is already known to be the case and you don't need the if statement. This looks like a bug given there is code after it.

pkg/k8s/apps/crud.go
```diff
@@ -27,7 +27,7 @@ import (
 	"k8s.io/client-go/kubernetes"
 )
 
-func Get(ctx context.Context, dev *model.Dev, namespace string, c kubernetes.Interface) (App, error) {
```
jmacelroy, line 30: Why has this name changed?

pkg/k8s/forward/manager.go
```diff
@@ -118,8 +118,13 @@ func (p *PortForwardManager) Add(f model.Forward) error {
...
 }
 
+// StartGlobalForwarding is not implemented
+func (*PortForwardManager) StartGlobalForwarding() error {
+	return fmt.Errorf("not implemented")
+}
+
 // AddReverse is not implemented
-func (p *PortForwardManager) AddReverse(_ model.Reverse) error {
+func (*PortForwardManager) AddReverse(_ model.Reverse) error {
```
jmacelroy, line 127: Why does this exist now? As far as I know we aren't using any imported code for this and therefore also control the interface so we shouldn't need to add a stub to satisfy an interface. It seems like we should create a different interface and possibly combine interfaces.

pkg/model/global_forward.go
```diff
@@ -0,0 +1,101 @@
+// Copyright 2022 The Okteto Authors
```
jmacelroy, line 1: Do not add any code to model. Let's come up with a better package for this.

pkg/k8s/apps/crud.go
AdrianPedriza: this new method is used by the global forwarder to get an app given a name instead of get it using the dev info (model.Dev), for that reason I distinguished between GetByDev and Get, maybe I should rename Get to GetByName?

AdrianPedriza: You are right @jmacelroy. **Get** was a new function to get an app by name instead of using model.Dev but in the end it is not necessary and stayed that way. I reverse the changes related to it.

pkg/k8s/forward/manager.go
AdrianPedriza: This and your suggestion regarding duplicate code I think could be addressed in this issue #2770. WDYT? If so, I can add some more information in the issue to complete the refactor properly.

cmd/up/forwards.go
AdrianPedriza: graceful shutdown is handled by _activate_ method https://github.com/okteto/okteto/blob/6c7d7d1d9b067e3b68c78f443e13387c4806cfa7/cmd/up/activate.go#L49 

this context is signaled to be done when starts _shutdown_ sequence.
https://github.com/okteto/okteto/blob/6c7d7d1d9b067e3b68c78f443e13387c4806cfa7/cmd/up/up.go#L743-L746

pkg/k8s/forward/manager.go
jmacelroy: We're already making big changes and this is a new feature. I would rather not add to technical debt as part of a new feature in cli at this point. I think using the interface and the "ifs" on global vs. non-global is a non-ideal implementation that adds to brittleness and testing difficulty. I'd rather have clear code paths for global vs non-global that are handled by struct initilization and types then conditionals.

jmacelroy (approved):

PR #2781, opened 2022-06-10
fix: could not detect manifest error

jLopezbarb:

Fixes error on build and deploy that could not detect a file

## Proposed changes
- It was introduced when we fallback to build v1
- Okteto manifest should not know about `Dockerfile` fallback
- If a file is given to an okteto build command we check if the Dockerfile is correct before doing the fallback

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -209,6 +209,7 @@ require sigs.k8s.io/json v0.0.0-20211020170558-c049b76a60c6 // indirect
 
 require (
 	github.com/Sirupsen/logrus v0.0.0-00010101000000-000000000000 // indirect
+	github.com/agext/levenshtein v1.2.3 // indirect
 	github.com/creack/pty v1.1.17 // indirect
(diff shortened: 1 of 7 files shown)


go.mod
```diff
@@ -209,6 +209,7 @@ require sigs.k8s.io/json v0.0.0-20211020170558-c049b76a60c6 // indirect
 
 require (
 	github.com/Sirupsen/logrus v0.0.0-00010101000000-000000000000 // indirect
+	github.com/agext/levenshtein v1.2.3 // indirect
```
teresaromero, line 212: 👀  what is this?

teresaromero (approved):

go.mod
jLopezbarb: Indirect dependency coming from -> `github.com/moby/buildkit/frontend/dockerfile/instructions`
It's used to suggest misspells on commands, e.g. `FRON -> FROM`

pchico83 (approved):

PR #2787, opened 2022-06-13
update release.sh

### LINKED ISSUE #2783: Fix release.sh versions mutation
We are releasing to multiple channels in `release.sh` using the same artifact directory. Since we are writing the new version to `${BIN_PATH}/versions` the last item in the loop takes precedence making it so that `preferred_channel` computed below is incorrect.

maroshii:
fixes #2783


maroshii:
NOTE: this needs to be backported to `release-2.4`

ifbyol (approved):

PR #2804, opened 2022-06-15
fix: compose flag on okteto up

### LINKED ISSUE #2789: `okteto up -f` is working different from `okteto up` on compose
**Describe the bug**
`okteto up -f docker-compose.yml` returns a volume error but running `okteto up` without the flag runs successfully. The error is:
Captura de pantalla 2022-06-14 a las 12 40 51

**To Reproduce**

1. Download [movies with compose repo](https://github.com/okteto/movies-with-compose)
2. Run `okteto up -f docker-compose`
3. See error
4. Run `okteto up`
5. See that is working

**Expected behavior**
Both command should run successfully

- OS: macOS
- Version  2.4.0-beta.1

jLopezbarb:

Fixes #2789

## Proposed changes
- We can't do it as part of serialisation because the docker compose can be in another folder
- We check if it's in a relative path and if not we create a new subpath based on the sync folder path

### DIFF
--- a/pkg/model/dev.go
+++ b/pkg/model/dev.go
@@ -983,7 +983,19 @@ func (dev *Dev) ToTranslationRule(main *Dev, reset bool) *TranslationRule {
 	}
 
 	if main.PersistentVolumeEnabled() {
+		wd, err := os.Getwd()
+		if err != nil {
+			oktetoLog.Info("could not retrieve working directory")
+		}
 		for _, v := range dev.Volumes {
+			subpath := v.RemotePath
+			if filepath.IsAbs(subpath) {
+				subpath, err = filepath.Rel(wd, v.RemotePath)
+				if err != nil {
+					oktetoLog.Info("could not retrieve subpath")
+					subpath = filepath.Base(v.RemotePath)
+				}
+			}
 			rule.Volumes = append(
 				rule.Volumes,
 				VolumeMount{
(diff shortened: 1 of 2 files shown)


pkg/model/dev.go
```diff
@@ -983,7 +983,19 @@ func (dev *Dev) ToTranslationRule(main *Dev, reset bool) *TranslationRule {
...
+		}
 		for _, v := range dev.Volumes {
+			subpath := v.RemotePath
+			if filepath.IsAbs(subpath) {
+				subpath, err = filepath.Rel(wd, v.RemotePath)
+				if err != nil || strings.HasPrefix(subpath, "..") {
+					oktetoLog.Info("could not retrieve subpath")
+					subpath = filepath.Base(v.RemotePath)
+				}
+			}
```
ifbyol, line 998: I think this logic is the same in both cases, the only thing that change is the remove vs local path. Can we extract the code to a function?

pkg/model/dev.go
```diff
@@ -994,12 +994,13 @@ func (dev *Dev) ToTranslationRule(main *Dev, reset bool) *TranslationRule {
...
 		}
 		for _, sync := range dev.Sync.Folders {
+			subpath := getSubPathFromLocalPath(sync.LocalPath)
 			rule.Volumes = append(
 				rule.Volumes,
 				VolumeMount{
 					Name:      main.GetVolumeName(),
 					MountPath: sync.RemotePath,
-					SubPath:   main.getSourceSubPath(sync.LocalPath),
+					SubPath:   main.getSourceSubPath(subpath),
```
AgustinRamiroDiaz, line 1003: Is it ok to pass an empty string here? If `getSubPathFromLocalPath` fails, we are not handling the error and instead passing "" here

jLopezbarb: Yes, it will fail when creating the new deployment. But we won't never send an empty string, if it can't detect if it's relative, we will send the latest element of the path

pkg/model/dev.go
```diff
@@ -1019,6 +1020,21 @@ func (dev *Dev) ToTranslationRule(main *Dev, reset bool) *TranslationRule {
 
 	return rule
 }
+func getSubPathFromLocalPath(localPath string) string {
+	wd, err := os.Getwd()
```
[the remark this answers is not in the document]
jLopezbarb: That requires a whole refactor, which will be made in a future, so I think it will be better to work on it when we do the big refactor

ifbyol (approved):

jmacelroy (approved):

PR #2805, opened 2022-06-15
TLS version

teresaromero:
## Proposed changes

- read `OKTETO_TLS_VERSION` and apply this config to the http client transport

### DIFF
--- a/pkg/analytics/track.go
+++ b/pkg/analytics/track.go
@@ -14,6 +14,7 @@
 package analytics
 
 import (
+	"crypto/tls"
 	"net"
 	"net/http"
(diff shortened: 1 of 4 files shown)


ifbyol (commented):
Why is this change needed for? Does it fix some issue? Who set the env variable `OKTETO_TLS_VERSION`? Can you provide more context please?

teresaromero:
> Why is this change needed for? Does it fix some issue? Who set the env variable `OKTETO_TLS_VERSION`? Can you provide more context please?

@ifbyol this a branch for making test regarding some customer issues

PR #2835, opened 2022-06-27
fix: load ctx resource only once

### LINKED ISSUE #2728: Warning unmarshall messages are shown two times
**Describe the bug**
If a okteto yaml/compose file/pipeline manifest has a warning while unmarshalling the message will appear two times (one before the context initialisation and another one after it). It's weird to see the same warning several times
Captura de pantalla 2022-05-30 a las 12 35 53

**To Reproduce**

1. Create this manifest:
```
name: test
image: okteto/golang:1
command: bash
labels:
  - as:as
sync:
- .:/usr/src/app
forward:
- 2345:2345
- 8080:8080
autocreate: true
```
2. Run okteto up 

**Expected behavior**
I wanted to see only one warning message

- OS: macOs
- Version 2.3.0

**Additional context**
This is due to how we read context. We need to load the same file two times, one to extract the namespace and context (contextresource) and another one to get the manifest struct.  We should be loading only the contextresource instead of reading the whole file and extracting the context resource from the struct.

jLopezbarb:

Fixes #2728

## Proposed changes
- Created a package that contains all stuff related to manage context resource instead of having it on files inside model package.
- When reading a context resource only read the namespace and context and skip the unmarshalling of the rest of the file

## Snapshot

Captura de pantalla 2022-06-27 a las 16 13 44


pchico83:
@jLopezbarb I think we are also reading the manifest twice to get the `build` section, and then, deserialize again replacing the `OKTETO_BUILD_XXX` envvars. How are we avoiding repeated warnings in that case?

jLopezbarb:
> @jLopezbarb I think we are also reading the manifest twice to get the `build` section, and then, deserialize again replacing the `OKTETO_BUILD_XXX` envvars. How are we avoiding repeated warnings in that case?

I wouldn't do it as part of this PR. We need to think about doing a full refactor on how to read manifest from scratch, avoiding reading the same parts twice, validating only once. But that needs a completely new way of unmarshalling and reading okteto manifests

jLopezbarb:
I'll change how we load it so it uses a `io.Reader` interface to make it easier to test the unmarshal

pkg/model/file.go
```diff
@@ -13,7 +13,71 @@
...
+	}
+	// HelmChartsSubPaths represents the possible names a helm chart can have
+	HelmChartsSubPaths = []string{
+		"chart",
+		"charts",
+		"helm/chart",
+		"helm/charts",
+	}
+	// K8sManifestSubPaths represents the possible names a k8s manifest can have
+	K8sManifestSubPaths = []string{
```
AdrianPedriza, line 69: what about : **manifest, manifest.yml, manifest.yaml** ?

cmd/context/utils_test.go
```diff
@@ -19,6 +19,7 @@ import (
 	"testing"
 
 	"github.com/okteto/okteto/pkg/model"
+	"github.com/okteto/okteto/pkg/model/contextresource"
```
jmacelroy, line 22: Does this need to be a sub-package of model? I like the idea of creating a better package name with a purpose but adding it as a sub-package doesn't really help eliminate the `model` package which is the direction I'd like to go.

pkg/model/contextresource/contextresource.go
```diff
@@ -0,0 +1,100 @@
...
+// limitations under the License.
+
+package contextresource
+
+import (
+	"os"
+
+	oktetoErrors "github.com/okteto/okteto/pkg/errors"
+	oktetoLog "github.com/okteto/okteto/pkg/log"
+	"github.com/okteto/okteto/pkg/model"
```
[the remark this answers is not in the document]
ifbyol: +1

pkg/model/file.go
ifbyol: they were not [considered before](https://github.com/okteto/okteto/pull/2835/files#diff-8d681e29798275482ee84496d80823cfe6c9a12b069672c92574ef39f82fcfd2L125), so I wouldn't include them as part of this PR.

AdrianPedriza: https://github.com/okteto/okteto/issues/2857

pkg/model/file.go
```diff
@@ -13,7 +13,71 @@
 
 package model
 
-import "path/filepath"
+import (
+	"os"
+	"path/filepath"
+)
+
+var (
```
teresaromero, line 21: opinion: can we put this at a file alone? like file_names or something like that :/

teresaromero (approved):

AdrianPedriza (approved):

jLopezbarb:
I'll reopen with the rdiscovery package refactor

PR #2839, opened 2022-06-28
[Backport release-2.4] Update .DEREK.yml (#2838)

maroshii:
(cherry picked from commit 13ff89dbeae6d112051e18e380bebbb8696127cb)

Manuallyu backport from #2838


ifbyol (approved):

.DEREK.yml
```diff
@@ -9,7 +9,5 @@ features:
   - dco_check
   - comments
   - pr_description_required
-  - release_notes
-  - hacktoberfest
```
ifbyol, line 13: @maroshii just curiosity, why `hacktoberfest` was in the release branch but not in master? was it deleted in master at some point since we create the branch?

maroshii: Yeah I didn't think it was worth backporting that change. Good thing we hit the conflict so we now understand how it works.
This was the PR: https://github.com/okteto/okteto/pull/2800

ifbyol: cool, thanks for the explanation!!

PR #2842, opened 2022-06-29
added file closure to delete it with windows

mnevadom:
Windows test were failing when creating temp files and trying to delete them after the test. 

Added file closure to avoid it

### DIFF
--- a/pkg/cmd/build/build_test.go
+++ b/pkg/cmd/build/build_test.go
@@ -98,11 +98,17 @@ func Test_OptsFromBuildInfo(t *testing.T) {
 	}
 
 	df := filepath.Join(serviceContext, serviceDockerfile)
-	defer removeFile(df)
-	_, errCreate := os.Create(df)
+	dockerfile, errCreate := os.Create(df)
 	if errCreate != nil {
 		t.Fatal(err)
 	}
+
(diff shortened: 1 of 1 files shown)


jLopezbarb (approved):

pkg/cmd/build/build_test.go
```diff
@@ -327,12 +333,16 @@ func TestExtractFromContextAndDockerfile(t *testing.T) {
 			if tt.dockerfilesCreated != nil {
 				for _, df := range tt.dockerfilesCreated {
 					defer removeFile(df)
```
ifbyol, line 335: Should we remove this?

pkg/cmd/build/build_test.go
```diff
@@ -327,12 +333,16 @@ func TestExtractFromContextAndDockerfile(t *testing.T) {
...
+					dfFile, err := os.Create(df)
 					if err != nil {
 						t.Fatal(err)
 					}
 
 					log.Printf("created docker file: %s", df)
+
+					if err := dfFile.Close(); err != nil {
+						t.Fatal(err)
+					}
```
ifbyol, line 345: Should we do this as part of the defer in line 329?

PR #2844, opened 2022-06-29
fix: remove parallel test when writing in map

AdrianPedriza:

## Proposed changes

- remove Parallel option to avoid concurrent map writes


cmd/up/forwards_test.go
```diff
@@ -100,7 +100,6 @@ func TestGlobalForwarderAddsProperlyPortsToForward(t *testing.T) {
 	for _, tt := range tests {
 		tt := tt
```
jLopezbarb, line 101:
```suggestion
		
```

jLopezbarb (approved):

PR #2848, opened 2022-06-29
Add debug logs to help troubleshoot "--wait" flag

pchico83:

Add debug logs to the `--wait` flag to help troubleshoot https://github.com/okteto/app/issues/4112

### DIFF
--- a/cmd/pipeline/deploy.go
+++ b/cmd/pipeline/deploy.go
@@ -26,7 +26,6 @@ import (
 	"github.com/okteto/okteto/pkg/cmd/pipeline"
 	oktetoErrors "github.com/okteto/okteto/pkg/errors"
 	"github.com/okteto/okteto/pkg/k8s/configmaps"
-	"github.com/okteto/okteto/pkg/log"
 	oktetoLog "github.com/okteto/okteto/pkg/log"
 	"github.com/okteto/okteto/pkg/model"
 	"github.com/okteto/okteto/pkg/okteto"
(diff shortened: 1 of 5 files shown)


cmd/pipeline/deploy.go
```diff
@@ -285,7 +286,8 @@ func waitForResourcesToBeRunning(ctx context.Context, name string, timeout time.
 				return err
 			}
 			allRunning := true
-			for _, status := range resourceStatus {
+			for resourceID, status := range resourceStatus {
```
rberrelleza, line 289: What is the value of resource ID?

rberrelleza: Can we use type/name instead? The resourceID is not human readable, and we don't expose it anywhere in the UI.

pkg/okteto/pipeline.go
```diff
@@ -404,18 +407,21 @@ func (c *OktetoClient) GetResourcesStatusFromPipeline(ctx context.Context, name
 	status := make(map[string]string)
 	for _, d := range queryStruct.Space.Deployments {
 		if string(d.DeployedBy) == name {
-			status[string(d.ID)] = string(d.Status)
+			resourceName := fmt.Sprintf("deployment/%s", d.Name)
```
jmacelroy, line 410: The logic for the key creation to access an element in the status map is duplicated and prone to bugs. Can we consolidate it into a function to eliminate the duplication and reduce the change of a bug occurring by an update to the key creation in either status write or read and not in the other location? Probably, something that takes a map and some constant indicating the resource type and either applies the change or returns what should be used for the key.

jmacelroy: A test may make sense for that function as well as it should be easy and help ensure that we have the correct debug code we want in place when the opportunity to capitalize on the debug logs occurs.

ifbyol (approved):

cmd/pipeline/deploy.go
```diff
@@ -26,6 +26,7 @@ import (
 	"github.com/okteto/okteto/pkg/cmd/pipeline"
 	oktetoErrors "github.com/okteto/okteto/pkg/errors"
 	"github.com/okteto/okteto/pkg/k8s/configmaps"
+	"github.com/okteto/okteto/pkg/log"
```
jLopezbarb, line 29: already declared with oktetoLog on the next line

jLopezbarb (approved):

PR #2864, opened 2022-06-30
fix configmap filename file path

### LINKED ISSUE #2846: We are not storing the right manifest path in the configmap when the manifest is not in the root of the repo
**Describe the bug**
If I execute `okteto deploy -f <folder/okteto.yaml>` and the specified manifest is not in the root of the repository, we are not storing the right manifest path in the internal config map. This is causing that redeploys from the UI are not working properly.

I think this was introduced when we started to execute deploy commands from the path where the manifest is defined instead of doing it from the root of the repository

**To Reproduce**

1. Clone the repo https://github.com/jmacelroy/go-getting-started and checkout `target-example` branch
2. From the root of the repo execute `okteto deploy -f functional_tests/okteto.yml`
3. When it finishes, if you check the internal config map generated, you would see that the `filename` field is storing `okteto.yml` instead of `functional_tests/okteto.yml`
4. If you go to the UI and redeploy the dev env, you would see that it is using the `okteto.yml`from the root instead of the one used before because the info in the config map is not the right now

**Expected behavior**
I would expect that the filename stored in the config map is `functional_tests/okteto.yml` instead of just `okteto.yml`

- Version: 2.4.0

**Additional context**
I think this was introduced when we started to execute deploy commands from the path where the manifest is defined instead of doing it from the root of the repository

teresaromero:
Fixes #2846 

## Proposed changes

- new Option at Deploy, Destroy and Up command to save the manifest file flag as OriginalManifestPath
- Renamed DevPath to ManifestPath to align with the -f flag at the other cmds
- As cwd is now being changed to the manifest directory, Filename at the configmaps produced by these commands is now using OriginalManifestPath (relative form) 
- Removed Filename from Manifest struct, as this was misleading when storing the manifest path on the configmap


cmd/deploy/deploy.go
```diff
@@ -125,6 +127,19 @@ func Deploy(ctx context.Context) *cobra.Command {
 			// deploy command. If not, we could be proxying a proxy and we would be applying the incorrect deployed-by label
 			os.Setenv(model.OktetoSkipConfigCredentialsUpdate, "false")
 			if options.ManifestPath != "" {
+				// we need to store the original manifest path as relative path before switching the cwd
+				options.OriginalManifestPath = options.ManifestPath
+				if filepath.IsAbs(options.ManifestPath) {
+					cwd, err := os.Getwd()
```
ifbyol, line 133: I would include a comment before this line to be careful of keeping this before doing the change for the working directory. In case of a future refactor that might move this logic

cmd/deploy/deploy.go
```diff
@@ -125,6 +127,19 @@ func Deploy(ctx context.Context) *cobra.Command {
 			// deploy command. If not, we could be proxying a proxy and we would be applying the incorrect deployed-by label
 			os.Setenv(model.OktetoSkipConfigCredentialsUpdate, "false")
 			if options.ManifestPath != "" {
+				// we need to store the original manifest path as relative path before switching the cwd
+				options.OriginalManifestPath = options.ManifestPath
+				if filepath.IsAbs(options.ManifestPath) {
```
ifbyol, line 132: @pchico83 Can you review if this is what we want when someone specify the manifest file as an absolute path?

cmd/destroy/destroy.go
```diff
@@ -95,6 +97,20 @@ func Destroy(ctx context.Context) *cobra.Command {
...
+					cwd, err := os.Getwd()
+					if err != nil {
+						return err
+					}
+					relativeOriginalManifestPath, err := filepath.Rel(cwd, options.ManifestPath)
+					if err != nil {
+						return err
+					}
+					options.OriginalManifestPath = relativeOriginalManifestPath
+				}
```
ifbyol, line 112: This logic seems similar in the 3 places where we have to use it, maybe we can extract it to a function and do a unit test?

cmd/deploy/deploy.go
```diff
@@ -60,15 +61,16 @@ var tempKubeConfigTemplate = "%s/.okteto/kubeconfig-%s-%d"
...
-	Name             string
-	Namespace        string
-	K8sContext       string
-	Variables        []string
-	Manifest         *model.Manifest
-	Build            bool
-	Dependencies     bool
-	servicesToDeploy []string
+	OriginalManifestPath string
+	ManifestPath         string
```
ifbyol, line 65: I would include a comment to specify the difference between these 2 paths. Our future selves will thank us for that (same for other places where we define these 2 properties)

cmd/deploy/deploy.go
```diff
@@ -60,15 +61,16 @@ var tempKubeConfigTemplate = "%s/.okteto/kubeconfig-%s-%d"
...
-	ManifestPath     string
-	Name             string
-	Namespace        string
-	K8sContext       string
-	Variables        []string
-	Manifest         *model.Manifest
-	Build            bool
-	Dependencies     bool
-	servicesToDeploy []string
+	OriginalManifestPath string
```
ifbyol, line 64: I'm trying to think in a better name for this. `SpecifiedManifestPath` or something similar to indicate that is the one specified by the user when the command is executed. Not sure if `Original` could represent that

pkg/model/manifest.go
```diff
@@ -295,14 +293,6 @@ func getManifestFromDevFilePath(cwd, manifestPath string) (*Manifest, error) {
...
 		}
-		path := ""
-		if filepath.IsAbs(manifestPath) {
-			path, err = filepath.Rel(cwd, manifestPath)
-			if err != nil {
-				oktetoLog.Debugf("could not detect relative path to %s: %s", manifestPath, err)
-			}
-		}
-		manifest.Filename = path
 		return manifest, nil
```
ifbyol, line 296: I think you can replace the code from line 294 to this one by just 

```
return getManifestFromFile(cwd, manifestPath)
```

ifbyol (commented):
@teresaromero can you try to see what we store if, being within `functional_tests` folder in the repo we've been working with these days I execute `okteto deploy -f ../okteto.yml`?

cmd/destroy/destroy.go
teresaromero: yes, i was thinking that. The change of cwd is the same at deploy destroy and up.... i have doubts on where to create this logic 👀  at the model package under the manifest utils?. what variates is the Options Struct... but we can "unify" a lot of logic...

cmd/deploy/deploy.go
teresaromero: ManifestPathFlag ? ¿

teresaromero:
> @teresaromero can you try to see what we store if, being within `functional_tests` folder in the repo we've been working with these days I execute `okteto deploy -f ../okteto.yml`?

When running the deploy from a directory that is not root, and file is at the parent directory, manifest is deployed but when redeploying from the UI i get the following:

```
Cloning git repository 'https://***:***@github.com/jmacelroy/go-getting-started' branch 'target-example'...Cloning into '/okteto/src'...
Running okteto pipeline at:
- Commit: 0696ced38c33c69463fa43ea8d26958a34c2f189
- Message: Adding sub-project.

filename specified '../okteto.yml' is not relative to the repositoryCreated pod: installer-b389b495-586c-42e8-ba7c-f5eeb050d5ed-rgj2mJob has reached the specified backoff limit
```

So the installer cwd is root, so the path to the manifest to re-deploy is saved relative to the folder that has executed the comand at the cli (not root)

cmd/deploy/deploy.go
jLopezbarb: Maybe splitting into two structs can do the trick. One to store flags/args information and another one to be the one used by the command. WDYT?

cmd/destroy/destroy.go
```diff
@@ -95,6 +97,20 @@ func Destroy(ctx context.Context) *cobra.Command {
 		Args:  utils.NoArgsAccepted("https://okteto.com/docs/reference/cli/#destroy"),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			if options.ManifestPath != "" {
```
jLopezbarb, line 99: Is this code the same as in deploy? Could we extract a function common to both and test it with an unit test?

cmd/deploy/deploy.go
teresaromero: 🤔  i like the idea, but... afraid of the scope of the PR

cmd/destroy/destroy.go
teresaromero: yes, that is the same to up deploy and destroy. i'm working on it.

cmd/destroy/destroy.go
ifbyol: The logic per se is not related to the manifest. Just with paths, don't know if we have some specific package, but `path.go` file sounds good to me to store this logic. At the end, this is checking that, if a path is absolute, get the relative one with another one (I wouldn't use `os.Getwd()` directly in the function, but passing its value as a parameter)

ifbyol:
> > @teresaromero can you try to see what we store if, being within `functional_tests` folder in the repo we've been working with these days I execute `okteto deploy -f ../okteto.yml`?
> 
> @ifbyol When running the deploy from a directory that is not root, and file is at the parent directory, manifest is deployed but when redeploying from the UI i get the following:
> 
> ```
> Cloning git repository 'https://***:***@github.com/jmacelroy/go-getting-started' branch 'target-example'...Cloning into '/okteto/src'...
> Running okteto pipeline at:
> - Commit: 0696ced38c33c69463fa43ea8d26958a34c2f189
> - Message: Adding sub-project.
> 
> filename specified '../okteto.yml' is not relative to the repositoryCreated pod: installer-b389b495-586c-42e8-ba7c-f5eeb050d5ed-rgj2mJob has reached the specified backoff limit
> ```
> 
> So the installer cwd is root, so the path to the manifest to re-deploy is saved relative to the folder that has executed the comand at the cli (not root)
> 
> cc: @pchico83

Ok, for now, I would assume that the path specified is relative to the root of the repo (which is the common case and the one for the installer). If not, any `okteto pipeline deploy -f <...>` would not be able to be redeployed correctly from the UI. Then, for the 2.5 (or other version), we will see the best way of covering those corner cases because it would probably imply broader changes on how to calculate the git repo for example

cc @pchico83 @jmacelroy

cmd/destroy/destroy.go
teresaromero: @ifbyol @jLopezbarb i've uploaded some changes regarding this common logic.... i'm not sure if this is the best aproach. let me know if is ok and i will add the test or refactor again :D

cmd/deploy/deploy.go
```diff
@@ -268,12 +287,19 @@ func (dc *DeployCommand) RunDeploy(ctx context.Context, deployOptions *Options)
 
 	setDeployOptionsValuesFromManifest(ctx, deployOptions, cwd, c)
 
+	// the filename at the cfgmap is used by the installer to redeploy the app from the ui
+	// this parameter is just saved if a repository is being detected
+	// when repository is empty - the filename should not be saved and redeploys should be done from cli
+	if os.Getenv(model.GithubRepositoryEnvVar) == "" {
+		deployOptions.ManifestPathFlag = ""
+	}
```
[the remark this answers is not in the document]
teresaromero: as offline discussion, i moved this logic into the updatecgmap function

jLopezbarb (approved):

ifbyol (approved):

PR #2879, opened 2022-07-01
Update possible k8s manifest names

### LINKED ISSUE #2857: update possible names a k8s manifest can have
we should add **manifest, manifest.yml, manifest.yaml** as possible names of a k8s manifest.

jeremyphua:
Signed-off-by: Jeremy Phua <jeremyphuachengtoon@gmail.com>

Fixes #2857

## Proposed changes

- Add **manifest**, **manifest.yml**, **manifest.yaml** to manifestSubPath slice in `pkg/model/manifest.go`.

### DIFF
--- a/pkg/model/manifest.go
+++ b/pkg/model/manifest.go
@@ -124,9 +124,6 @@ var (
 		"helm/charts",
 	}
 	manifestSubPath = []string{
-		"manifest",
-		"manifest.yml",
-		"manifest.yaml",
 		"manifests",
 		"manifests.yml",
 		"manifests.yaml",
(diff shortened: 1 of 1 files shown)


pkg/model/manifest.go
```diff
@@ -124,6 +124,9 @@ var (
 		"helm/charts",
 	}
 	manifestSubPath = []string{
+		"manifest",
+		"manifest.yml",
+		"manifest.yaml",
```
ifbyol, line 129: As the files are processed in the order defined here and in order to keep current preference of files, can you include them at the end of the list?

ifbyol (commented):
Thanks for your contribution!

pkg/model/manifest.go
jeremyphua: Sure, I have made the changes!

jLopezbarb (approved):

teresaromero (approved):

pkg/model/manifest.go
jeremyphua: Thanks for reviewing @ifbyol!

PR #2883, opened 2022-07-04
Warning two dockerfiles fixed

mnevadom:
This warning was showing if the Dockerfile was only in PWD

If there is no other Dockerfile we don't need to show the warning


jLopezbarb:
I think we have an unit test that test when we should throw the warning. Could you add this use case to the test?

ifbyol (commented):
picky mode: can you also include the final `'` [here](https://github.com/okteto/okteto/blob/master/pkg/cmd/build/build.go#L37). The warning message is opening `'` for the path but it is not closing it

teresaromero:
> here



> picky mode: can you also include the final `'` [here](https://github.com/okteto/okteto/blob/master/pkg/cmd/build/build.go#L37). The warning message is opening `'` for the path but it is not closing it

+1 @mnevadom

jLopezbarb (approved):

PR #2892, opened 2022-07-05
docs: add code review guidelines

### LINKED ISSUE #2851: Develop Code Review Guidelines
Establish code review guidelines for the project and evangelize with regular contributors.

jLopezbarb:

Fixes #2851

## Proposed changes
- Adds code review guideline


docs/code-review.md
```diff
@@ -0,0 +1,338 @@
...
+
+Also,
+
+#### E2E tests
+
+Each command should have its own set of end to end tests to prove that the main functionality works correctly. An e2e test should be added to a new feature if it adds a new use case or breaks an existing use case.
+
+**New tests will be launched in parallel to the existing ones in order to speed up the CI process.**
+
+See how to run e2e tests to run e2e tests locally
```
maroshii, line 193: I think there's something missing here

docs/code-review.md
```diff
@@ -0,0 +1,330 @@
...
+- [ ] Does it affect other services (actions/vscode plugin/pipeline/graphql/json logs)?
+  *It can affect other services that are not the CLI itself and break scenarios that are not contemplated on the CLI repository, so we need to bear in mind all those services when reviewing a PR*
+- [ ] Does it need to add analytics?
+  *We need to add analytics to know the adoption of new features and how the users use the product*
+- [ ] Are there any code smells?
+  we should develop clean code so that if a new developer starts working on the project, he/she can understand the code as soon as possible*
+
+### PR Description
+
+The description of the pull request should state what issue it solves (or explain if there is no associated issue) and how it is solved.
```
ifbyol, line 45: I would require an issue for the CLI (personal opinion)

docs/code-review.md
```diff
@@ -0,0 +1,330 @@
...
+
+All new features or bug fixes should have at least unit tests that cover the use case. If for any reason unit tests cannot be created, at least one e2e test should be created to prove that the new functionality works correctly.
+
+Tests should fail with helpful messages saying what was wrong, with what inputs, what was actually got, and what was expected. Assume that the person debugging your failing test is not you, and is not your team.
+
+#### Unit tests
+
+The code should be testable by unit functions, which means that each function should perform an action and this action should be testable.
+
+Only one test will be created per function in which it will be tested by table driven tests with test names descriptive enough for any user to know what it does just by reading the name of the test. For example:
```
ifbyol, line 65: I wouldn't enforce the fact of creating just one test function and use table driven tests. Some scenarios, specially the ones which verify errors or normal output (like the example you posted below) force you to have some kind of flag and include an `if` within the test to do different asserts. In general, that is a bad practice in testing

docs/code-review.md
```diff
@@ -0,0 +1,330 @@
...
+```
+
+Each test should create its own scenario, eliminating and leaving the state of the machine in the same state as it was at the start of the test.
+
+#### E2E tests
+
+Each command should have its own set of end to end tests to prove that the main functionality works correctly. An e2e test should be added to a new feature if it adds a new use case or breaks an existing use case.
+**In order to speed up the CI process, all tests are executed in parallel including any new tests**
+
+### Logs
```
ifbyol, line 189: We might want to mention how they should be used. How an information message is displayed to the user or how an info message is just stored for okteto doctor purposes

docs/code-review.md
```diff
@@ -0,0 +1,330 @@
...
+if err != nil {
+    // error handling
+    return
+}
+// use x
+```
+
+### Naming convention
+
+#### Variable Names
```
ifbyol, line 294: I would include links to golang best practices: https://go.dev/doc/effective_go

docs/code-review.md
```diff
@@ -0,0 +1,330 @@
...
+
+### Code review checklist
+
+Here is a summary of the things every reviewer should be aware of while doing a CLI code review:
+
+- [ ] Does the PR have unit tests/e2e tests that covers all the scenarios?
+  *We should look forward to have all scenarios covered.*
+- [ ] Does the PR description explain what it does/solves?
+  *In the future we may need to know why the PR was created and what issues it solved, so this should be made clear in the description of the PR.*
+- [ ] Does it affect other services (actions/vscode plugin/pipeline/graphql/json logs)?
```
ifbyol, line 36: Sometimes it is hard to know if some changes affect to other services or not. For example, how logs are going to be displayed in the UI (due to the stages, etc..) or the path stored within `filename` property of configmaps created by `okteto deploy` command. 

Maybe we can be more explicit on the things we should pay more attention because affect to other components

docs/code-review.md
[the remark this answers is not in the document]
jmacelroy: That being said I think this section should describe that table tests can be quite powerful and call out *when* they are a good candidate.

docs/code-review.md
```diff
@@ -0,0 +1,330 @@
...
+#### E2E tests
+
+Each command should have its own set of end to end tests to prove that the main functionality works correctly. An e2e test should be added to a new feature if it adds a new use case or breaks an existing use case.
+**In order to speed up the CI process, all tests are executed in parallel including any new tests**
+
+### Logs
+
+Logs are an important part of the CLI because if a user has a problem, we will have to review the file generated by okteto doctor, so these logs can be used as a tool to help us identify the problem.
+
+Through these you should be able to follow the execution of the okteto command without problems, so they should be as concise as possible. Errors not returned to the user must be logged without exception as this is the only trace we could have when looking for a problem.
```
jmacelroy, line 193:
```suggestion
Through these you should be able to follow the execution of the okteto command without problems, so they should be as concise as possible. Errors not returned to the user must be logged, without exception, as this is the only trace we may have when debugging an issue.
```

docs/code-review.md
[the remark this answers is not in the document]
ifbyol: Good point, agree on describing the benefits of table tests and when they are a good fit

docs/code-review.md
```diff
@@ -0,0 +1,363 @@
...
+      - [Search for an element](#search-for-an-element)
+
+## What to look for in a code review?
+
+### Code review checklist
+
+Here is a summary of the things every reviewer should be aware of while doing a CLI code review:
+
+- [ ] Does the PR have unit tests/e2e tests that cover all the scenarios?
+  *We should look forward to having all scenarios_ covered.*
```
ifbyol, line 33:
```suggestion
  *We should look forward to having all scenarios covered.*
```

docs/code-review.md
```diff
@@ -0,0 +1,363 @@
...
+
+    for _, tt := range tests {
+        t.Run(tt.name, func(t *testing.T) {
+            t.Setenv("ENV_REQUIRED_FOR_TEST", "testValue")
+            testDir := t.TempDir()
+            err := functionName(testDir)
+            assert.NoError(t, err)
+        })
+    }
+}
```
ifbyol, line 95: I think some part of this example is misleading. The third case expects **no** error according to the description and the other 2 expects **an** error but the test is called `Test_FunctionNameWithNoError`. Also, in the line 92, we assert that **no** error is expected.

I think there is a mix of things between this example and the next one.

I think this first function has to be called `Test_FunctionNameWithError`, remove the third scenario and replace the assert in line 92 by `assert.Error(t, err)`

Then, the second example should be called `Test_FunctionNameWithNoError` (notice the **No**) and the line 100 should be replaced by `err := functionName(CorrectResource{})`

docs/code-review.md
```diff
@@ -0,0 +1,363 @@
...
+            testDir := t.TempDir()
+            err := functionName(testDir)
+            if tt.expectedErr {
+                assert.Error(t, err)
+            } else {
+                assert.NoError(t, err)
+            }
+        })
+    }
+}
```
ifbyol, line 144: I wouldn't include an example with a non-recommended approach for unit testing. Including conditionals in the test to verify some conditions or others is error prone.

Having said that, I think this example doesn't really want to use the conditional. I think we can remove the `expectedErr` as part of the test input and remove the conditional in the line 137 to only verify `assert.Error(t, err)`. All input cases are expecting an error

docs/code-review.md
```diff
@@ -0,0 +1,363 @@
...
```
ifbyol, line 178: Same comment as previous example, I think we need to review this example. Check in line 171 is not needed and can be replaced by just `assert.NoError(t, err)`

docs/code-review.md
```diff
@@ -0,0 +1,370 @@
...
+
+### PR comments
+
+Comments are the basis of all pull requests. In comments, the reviewers will ask questions about the code, request changes, and discuss what is the best approach to improve the CLI code.
+
+A respectful tone will be used when making comments and supporting documentation can be helpful for certain comments.
+
+### Tests
+
+All new features or bug fixes should have unit tests that cover the added code and the functionality it provides. When adding new features it is highly recommended that integration tests are also created and may be deemed required by a reviewer.
```
jmacelroy, line 58:
```suggestion
All new features or bug fixes must have unit tests that cover the added code and the functionality it provides. When adding new features it is highly recommended that integration tests are also created and may be deemed required by a reviewer.
```

docs/code-review.md
```diff
@@ -0,0 +1,370 @@
...
+            t.Setenv("ENV_REQUIRED_FOR_TEST", "testValue")
+            testDir := t.TempDir()
+            err := functionName(testDir)
+            assert.NoError(t, err)
+        })
+    }
+}
+```
+
+Each test should create its scenario, eliminating and leaving the state of the machine in the same state as it was at the start of the test.
```
jmacelroy, line 170:
```suggestion
Each test should have the capability of running in isolation, eliminating and leaving the state of the machine in the same state as it was at the start of the test.
```

ifbyol (approved):
LGTM, nice work @jLopezbarb !

PR #2893, opened 2022-07-05
Jlopezbarb/fix up file flag compose

jLopezbarb:
Fixes #2804 

## Proposed changes
- Fix when a compose is passed as a flag


teresaromero (approved):
tested the branch, LGTM

PR #2894, opened 2022-07-06
Fix env parsing at stack expander

### LINKED ISSUE #2597: 'okteto up' doesn't interpret '$PWD' in the volume section of a docker-compose
**Describe the bug**
I'm trying to deploy https://github.com/rberrelleza/play-with-docker/blob/master/docker-compose.yml. The volumes section looks like this:

```
    pwd:
        # pwd daemon container always needs to be named this way
        container_name: pwd
        # use the latest golang image
        image: golang
        # go to the right place and starts the app
        command: /bin/sh -c 'ssh-keygen -N "" -t rsa -f /etc/ssh/ssh_host_rsa_key >/dev/null; cd /go/src/; if [ -e /runbin/pwd ]; then /runbin/pwd  -save /pwd/sessions -name l2; else go run api.go -save /pwd/sessions -name l2; fi'
        volumes:
            # since this app creates networks and launches containers, we need to talk to docker daemon
            - /var/run/docker.sock:/var/run/docker.sock
            # mount the box mounted shared folder to the container
            - $PWD:/go/src
            - sessions:/pwd
```

When deployed, $PWD is converted to a string, instead of evaluating it to the current folder

teresaromero:

Fixes #2597 

## Proposed changes

- When introducing the stack expander with this PR to support multiline envs https://github.com/okteto/okteto/pull/2652 the case of volumes with envs was not taken into account
- This fix refactors the Stack expander so it acts as an expander per node. No transformation is done, so the Expander will expand the envs previous to the stack unmarshalling. There is logic at the stack unmarshall where we allow environments to be declared just by the key, so we populate the value with the value if its present at the environment.

### DIFF
--- a/pkg/model/stack_expander_test.go
+++ b/pkg/model/stack_expander_test.go
@@ -1,13 +1,13 @@
 package model
 
 import (
-	"os"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
 )
 
 func Test_ExpandStackEnvs(t *testing.T) {
+	t.Setenv("ENV2", "bye")
 	tests := []struct {
 		name          string
 		file          []byte
@@ -23,18 +23,16 @@ services:
         build:
             context: vote
             args:
-                - $CUSTOM_ENV
+                - ENV=hello
                 - CUSTOM2_ENV=$CUSTOM_ENV
-                - EMPTY
-                - ${ENV:-dev}
-                - ${ENV}
+                - EMPTY=
+                - ENV2=
         ports:
             - 8080:8080
         environment:
             FLASK_ENV: development
             CUSTOM_ENV: $CUSTOM_ENV
-            $CUSTOM2_ENV:
-            EMPTY:
+            ENV:
         volumes:
             - ./vote:/src
             - $CUSTOM_ENV:/src
@@ -56,18 +54,16 @@ volumes:
     build:
       context: vote
       args:
-        - MYVALUE
+        - ENV=hello
         - CUSTOM2_ENV=MYVALUE
-        - EMPTY
-        - dev
-        - ""
+        - EMPTY=
+        - ENV2=
     ports:
       - 8080:8080
     environment:
       FLASK_ENV: development
       CUSTOM_ENV: MYVALUE
-      "":
-      EMPTY:
+      ENV:
     volumes:
       - ./vote:/src
       - MYVALUE:/src
@@ -92,14 +88,13 @@ services:
         build:
             context: vote
             args:
-                - $CUSTOM_ENV
                 - CUSTOM2_ENV=$CUSTOM_ENV
         ports:
             - 8080:8080
         environment:
             FLASK_ENV: development
             CUSTOM_ENV: $CUSTOM_ENV
-            $CUSTOM2_ENV:
+            CUSTOM2_ENV:
         volumes:
             - ./myservice:/src
 
@@ -119,9 +114,6 @@ volumes:
       context: vote
       args:
         - |-
-          my first line
-          my second line
-        - |-
           CUSTOM2_ENV=my first line
           my second line
     ports:
@@ -131,7 +123,7 @@ volumes:
       CUSTOM_ENV: |-
         my first line
         my second line
-      "":
+      CUSTOM2_ENV:
     volumes:
       - ./myservice:/src
   redis:
@@ -150,9 +142,7 @@ volumes:
 	for _, tt := range tests {
 
 		t.Run(tt.name, func(t *testing.T) {
-			if err := os.Setenv("CUSTOM_ENV", tt.envValue); err != nil {
-				t.Fatal(err)
-			}
+			t.Setenv("CUSTOM_ENV", tt.envValue)
 			result, err := ExpandStackEnvs(tt.file)
 			if err != nil && !tt.expectedError {
 				t.Fatalf("expected no error, but got error: %v", err)
@@ -162,7 +152,6 @@ volumes:
 
 			assert.Equal(t, tt.expectedStack, string(result))
 
-			os.Unsetenv("CUSTOM_ENV")
 		})
 	}
 }
--- a/pkg/model/stack_test.go
+++ b/pkg/model/stack_test.go
@@ -199,7 +199,7 @@ services:
       - OPTION_A=Cats
       - OPTION_B=Dogs
       - EMPTY_VAR=
-      - $PWD
+      - PWD
     ports:
       - 80
     replicas: 2
@@ -221,7 +221,8 @@ services:
     entrypoint: e
     command: c
     volumes:
-      - /var/lib/postgresql/data`)
+      - /var/lib/postgresql/data
+      - $PWD/src`)
 	s, err := ReadStack(manifest, true)
 	if err != nil {
 		t.Fatal(err)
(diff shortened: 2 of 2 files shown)


jLopezbarb (approved):

pkg/model/stack_expander_test.go
```diff
@@ -26,6 +26,8 @@ services:
             - $CUSTOM_ENV
             - CUSTOM2_ENV=$CUSTOM_ENV
             - EMPTY
+            - ${ENV:-dev}
+            - ${ENV}
```
jLopezbarb, line 30: Can we add `$ENV`?

teresaromero: is that the same case as `$CUSTOM_ENV` ¿? first of list?

pkg/model/stack_expander.go
```diff
@@ -8,6 +8,42 @@ import (
...
+		return true
+	}
+	return false
+}
+func isCurlyEnv(value string) bool {
+	return strings.HasPrefix(value, "{") && strings.HasSuffix(value, "}")
+}
+func hasEnvDefaultValue(value string) (int, bool) {
+	return strings.Index(value, ":-"), strings.Contains(value, ":-")
+}
```
ifbyol, line 28: Can we create simple unit tests for these functions? I see that the env expansion starts to relies in this and including tests will help us to prevent issues in the future if we start to detect o include new cases

pkg/model/stack_expander.go
```diff
@@ -21,9 +57,8 @@ func expandEnvScalarNode(node *yaml3.Node) (*yaml3.Node, error) {
 	// when is a Sequence and starts with $ can be a list of envs, so transform the list to key=value format
 	case yaml3.SequenceNode:
 		for indx, subNode := range node.Content {
-			if strings.HasPrefix(subNode.Value, "$") {
+			if key, ok := isEnvStringKey(subNode.Value); ok {
```
pchico83, line 60: @teresaromero I think we don't need logic for this, it's automatically done by the golang env substitution (cc: @jLopezbarb)

teresaromero: i added this because there is the case where subNode.Value can be $ENV or $ENV:src, this is used to append as key de variable name, after the env is expanded, with the prev logic `ENV:src` was used as key, which is not correct

teresaromero: @pchico83 added unit test for the functions so its clearer the changes and what is doing the functions

cc: @ifbyol

pchico83: does `${ENV}:src` work today?

teresaromero: > does `${ENV}:src` work today?

i'm afraid not, this is an example of the issue reported... 
this serialiser was implemented tu support multiline envs in the first place (PR description). Are we using a compose serializer? I believe that whenever we are changing to use the compose serializer this would need to be checked

teresaromero:
I've remove some code from the stack expander which after reviewing the compose spec did not make much sense.
The objectinve of using the expander was to allow multiline to expand correctly to the yaml node.

Please review if this is now the correct behavour, removing this code, fixes the volume env mapping

cc: @pchico83 @ifbyol @jLopezbarb

pkg/model/stack_test.go
```diff
@@ -199,7 +199,7 @@ services:
       - OPTION_A=Cats
       - OPTION_B=Dogs
       - EMPTY_VAR=
-      - $PWD
+      - PWD
```
ifbyol, line 202: Should the other case also work? I mean, if you specify only the `$PWD`.

I wouldn't like to break some existing case and this is a test case that is changing

pkg/model/stack_expander_test.go
```diff
@@ -81,15 +87,14 @@ services:
...
-            - CUSTOM2_ENV=$CUSTOM_ENV
+            args:
+                - CUSTOM2_ENV=$CUSTOM_ENV
         ports:
             - 8080:8080
         environment:
             FLASK_ENV: development
             CUSTOM_ENV: $CUSTOM_ENV
-            $CUSTOM2_ENV:
+            CUSTOM2_ENV:
```
ifbyol, line 97: Same about the other (previous) case. Is it something we are changing?

pkg/model/stack_expander_test.go
```diff
@@ -22,19 +22,22 @@ services:
...
+                - ENV=hello
+                - CUSTOM2_ENV=$CUSTOM_ENV
+                - EMPTY=
+                - ENV2=
         ports:
             - 8080:8080
         environment:
             FLASK_ENV: development
             CUSTOM_ENV: $CUSTOM_ENV
-            $CUSTOM2_ENV:
```
ifbyol, line 34: Same about this case, why removing it?

pkg/model/stack_expander_test.go
teresaromero: We do translate it, and the initial implementation of the stack expander was taking into account this logic, but this does not comply with the compose spec, which what it does is that it expands the value, not making the key-value trick. What should we do in this case? should i recover then the logic regarding the expansion of this case into key-value? 

cc: @jLopezbarb @pchico83

pchico83: @teresaromero on which cli version was that logic added? I think that behavior is wrong, and the sooner we delete it, the better. I don't think people are using it that way unless it was implemented for a user feature request

ifbyol: if we think it is not being used, agree with deleting it

ifbyol (approved):

pkg/model/stack_expander_test.go
```diff
@@ -22,19 +22,22 @@ services:
...
             context: vote
-        args:
-            - $CUSTOM_ENV
-            - CUSTOM2_ENV=$CUSTOM_ENV
-            - EMPTY
+            args:
+                - ENV=hello
+                - CUSTOM2_ENV=$CUSTOM_ENV
+                - EMPTY=
+                - ENV2=
```
pchico83, line 29: should this one translate to `ENV2=${ENV2}`? MAybe that only works for `environment`

pkg/model/stack_expander_test.go
```diff
@@ -22,19 +22,22 @@ services:
...
         ports:
             - 8080:8080
         environment:
             FLASK_ENV: development
             CUSTOM_ENV: $CUSTOM_ENV
-            $CUSTOM2_ENV:
-            EMPTY:
+            ENV:
```
pchico83, line 35: we should test that `ENV2` translated to `ENV2=bye`

pchico83 (approved):

pkg/model/stack_expander_test.go
teresaromero: the stack expander only is responsible for expand the envs. The logic that makes this happen is at the stack unmarshall, where we check that if an env only has key, and there is a value for this key as env, we assing this value

pchico83 (approved):

pchico83:
@teresaromero should we create an issue to follow the same yaml serialization approach for the okteto manifest? I guess it has the same problem if you have a multi-line env variable

teresaromero:
> @teresaromero should we create an issue to follow the same yaml serialization approach for the okteto manifest? I guess it has the same problem if you have a multi-line env variable

https://github.com/okteto/okteto/issues/2925 opeend this issue, as i think is not happening there but we should check

PR #2898, opened 2022-07-06
fix: panic on destroy with wait flag

### LINKED ISSUE #2788: okteto pipeline panics when adding wait and no pipeline running
**Describe the bug**
`okteto pipeline destroy --name test --namespace=test --wait` panics with the following trace:
```
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x2994472]

goroutine 1 [running]:
github.com/okteto/okteto/cmd/pipeline.ExecuteDestroyPipeline({0x362a3d8, 0xc0000560c0}, 0xc0013eccf0)
	github.com/okteto/okteto/cmd/pipeline/destroy.go:104 +0x272
github.com/okteto/okteto/cmd/pipeline.destroy.func1(0xc0013fea00, {0xc0003d6780, 0x0, 0x5})
	github.com/okteto/okteto/cmd/pipeline/destroy.go:50 +0x25
github.com/spf13/cobra.(*Command).execute(0xc0013fea00, {0xc0003d6730, 0x5, 0x5})
	github.com/spf13/cobra@v1.3.0/command.go:856 +0x60e
github.com/spf13/cobra.(*Command).ExecuteC(0xc00112a780)
	github.com/spf13/cobra@v1.3.0/command.go:974 +0x3bc
	github.com/spf13/cobra@v1.3.0/command.go:902
main.main()
	github.com/okteto/okteto/main.go:136 +0x9a9
```
**To Reproduce**

1. Run `okteto pipeline destroy --name test --namespace=test --wait`
2. See error

**Expected behavior**
Finish the command correctly

- OS: macOs
- Version  2.4.0-beta.1

jLopezbarb:

Fixes #2788

## Proposed changes
- This fix a panic which affects https://github.com/marketplace/actions/destroy-okteto-pipeline 
- Adds a condition when the response on the backend is nil (response we return when there is a pipeline not found error) and there is no error

### DIFF
--- a/cmd/pipeline/destroy.go
+++ b/cmd/pipeline/destroy.go
@@ -101,8 +101,10 @@ func ExecuteDestroyPipeline(ctx context.Context, opts *DestroyOptions) error {
 		return nil
 	}
 
-	if err := waitUntilDestroyed(ctx, opts.Name, resp.Action, opts.Timeout); err != nil {
-		return err
+	if resp != nil {
+		if err := waitUntilDestroyed(ctx, opts.Name, resp.Action, opts.Timeout); err != nil {
+			return err
+		}
 	}
 
 	oktetoLog.Success("Repository '%s' successfully destroyed", opts.Name)
--- a/cmd/pipeline/deploy.go
+++ b/cmd/pipeline/deploy.go
@@ -163,10 +163,8 @@ func ExecuteDeployPipeline(ctx context.Context, opts *DeployOptions) error {
 		return nil
 	}
 
-	if resp != nil {
-		if err := waitUntilRunning(ctx, opts.Name, resp.Action, opts.Timeout); err != nil {
-			return err
-		}
+	if err := waitUntilRunning(ctx, opts.Name, resp.Action, opts.Timeout); err != nil {
+		return err
 	}
 
 	oktetoLog.Success("Repository '%s' successfully deployed", opts.Name)
--- a/cmd/pipeline/deploy.go
+++ b/cmd/pipeline/deploy.go
@@ -163,9 +163,12 @@ func ExecuteDeployPipeline(ctx context.Context, opts *DeployOptions) error {
 		return nil
 	}
 
-	if err := waitUntilRunning(ctx, opts.Name, resp.Action, opts.Timeout); err != nil {
-		return err
+	if resp != nil {
+		if err := waitUntilRunning(ctx, opts.Name, resp.Action, opts.Timeout); err != nil {
+			return err
+		}
 	}
+
 	oktetoLog.Success("Repository '%s' successfully deployed", opts.Name)
 	return nil
 }


cmd/pipeline/destroy.go
```diff
@@ -101,8 +101,10 @@ func ExecuteDestroyPipeline(ctx context.Context, opts *DestroyOptions) error {
 		return nil
 	}
 
-	if err := waitUntilDestroyed(ctx, opts.Name, resp.Action, opts.Timeout); err != nil {
-		return err
+	if resp != nil {
```
ifbyol, line 104: I think it is odd to have a nil response without any error if the pipeline is not found. I think a better approach is to return a specific error within `destroyPipeline` (it seems it is only being used here) and let the client decide what to do with it. 

In this case, we would check the error type to wait or not until destroyed. Anyways, a minor comment

ifbyol (approved):

cmd/pipeline/destroy.go
jLopezbarb: Will do and add some test mocking the okteto API

cmd/pipeline/destroy.go
```diff
@@ -91,26 +94,27 @@ func ExecuteDestroyPipeline(ctx context.Context, opts *DestroyOptions) error {
...
 	}
 
 	if !opts.Wait {
 		oktetoLog.Success("Repository '%s' scheduled for destruction", opts.Name)
 		return nil
 	}
-
-	if err := waitUntilDestroyed(ctx, opts.Name, resp.Action, opts.Timeout); err != nil {
-		return err
+	if err == nil || !oktetoErrors.IsNotFound(err) {
```
ifbyol, line 106: I think you can omit the second part of the if. You can include a comment to indicate that if the error was a not found we don't have to wait. I think that `!oktetoErrors.IsNotFound(err)` cannot be true because the line 98 would be true.

pkg/types/interface.go
```diff
@@ -33,10 +40,21 @@ type NamespaceInterface interface {
...
 }
 
+// PreviewInterface represents the client that connects to the preview functions
 type PreviewInterface interface {
 	List(ctx context.Context) ([]Preview, error)
 }
 
+// PipelineInterface represents the client that connects to the pipeline functions
+type PipelineInterface interface {
+	DeployPipeline(ctx context.Context, name, repository, branch, filename string, variables []Variable) (*GitDeployResponse, error)
```
ifbyol, line 50: As this is now within a PipelineInterface, maybe it can be called just `Deploy`. The same for `Destroy`, `GetResourcesStatus` and `GetByName`.

If it implies too much refactor, no worries, it is ok as it is

ifbyol (approved):

teresaromero (approved):

AdrianPedriza (approved):

Issue #2885, opened 2022-07-04
Application endpoints are not displayed anymore when `okteto deploy` is executed


ifbyol:
**Describe the bug**
When `okteto deploy` was being executed locally (not within a pipeline), we were displaying the endpoints of the applications at the end of the operation.

At some point, this stopped working and checking the output when executing the command with `-l=debug`, you can see 
```
INFO[0004] could not retrieve endpoints: [ingresses.networking.k8s.io](http://ingresses.networking.k8s.io/) is forbidden: User "system:serviceaccount:okteto:xxxx" cannot list resource "ingresses" in API group "[networking.k8s.io](http://networking.k8s.io/)" at the cluster scope
```

The issue is that we are not specifying the right namespace [when creating the options to show the endpoints](https://github.com/okteto/okteto/blob/master/cmd/deploy/deploy.go#L423). This is provoking that the search of ingresses is done with the namespace `""`, and users don't have permissions to do that.

We should initialize properly the namespace when checking the endpoints

**To Reproduce**

1. Clone some repo which should generate an endpoint
2. Execute `okteto deploy` from the root
3. Output doesn't contains the  generated endpoint(s)

**Expected behavior**
The generated endpoint(s ) should be listed as we had before

<img width="436" alt="Captura de Pantalla 2022-07-04 a las 11 33 29" src="https://user-images.githubusercontent.com/3510171/177127260-1321757b-0a64-4851-9604-9e2f9322f487.png">

**Desktop (please complete the following information):**

- Version: `2.4.1`

**Additional context**

We need to do something similar to this

```
diff --git a/cmd/deploy/deploy.go b/cmd/deploy/deploy.go
index 4b8d34c2..3d15692f 100644
--- a/cmd/deploy/deploy.go
+++ b/cmd/deploy/deploy.go
@@ -420,7 +420,7 @@ func (dc *DeployCommand) RunDeploy(ctx context.Context, deployOptions *Options)
                                }
                        }
                        if !utils.LoadBoolean(model.OktetoWithinDeployCommandContextEnvVar) {
-                               if err := dc.showEndpoints(ctx, &EndpointsOptions{Name: deployOptions.Name}); err != nil {
+                               if err := dc.showEndpoints(ctx, &EndpointsOptions{Name: deployOptions.Name, Namespace: deployOptions.Manifest.Namespace}); err != nil {
                                        oktetoLog.Infof("could not retrieve endpoints: %s", err)
                                }
                        }
```

[...]

Issue #2886, opened 2022-07-04
Handle `namespace` flag in the same way for all the commands that allow it


ifbyol:
**Is your feature request related to a problem? Please describe.**
Several commands support the namespace flag and also the ability to specify the namespace through the manifest and we have different behaviors depending on the command.

For example, using the repository https://github.com/jmacelroy/go-getting-started.git, if I include a `namespace` property in the `okteto.yml` with the namespace `new-ifbyol`
* When I execute `okteto deploy -n ifbyol`, the dev environment (empty) is deployed within `new-ifbyol` but the `hello` deployment is deployed within `ifbyol` namespace
<img width="1317" alt="Captura de Pantalla 2022-07-04 a las 11 58 17" src="https://user-images.githubusercontent.com/3510171/177131483-cd7bfd1b-098e-4f55-9ca5-30bd3f2765a1.png">
<img width="1452" alt="Captura de Pantalla 2022-07-04 a las 11 58 25" src="https://user-images.githubusercontent.com/3510171/177131494-9a7dd9da-24ef-4f29-ad1d-e8bf66ca64e1.png">

* When I execute `okteto destroy -n ifbyol`, I get an error `The namespace in the okteto manifest doesn't match your namespace argument`
<img width="665" alt="Captura de Pantalla 2022-07-04 a las 11 59 24" src="https://user-images.githubusercontent.com/3510171/177131671-21a1eb38-953c-4b1b-b315-7d4b33978e91.png">

Also, from a coding perspective, sometimes we are setting the Namespace specified in the `options` struct we usually create with the flags with the context namespace if the user didn't specify one and in other commands we are not doing that and create other struct with the real values for the command

**Describe the solution you'd like**
An homogeneous behavior of the Namespace flag. If the specified flag cannot be different to the one specified in the manifest, the deploy operation should fail as the destroy does.

Ideally, we should also follow the same convention when defined flags for the commands and then, how they are passed to the real command execution. We should probably decouple both and have an struct with all the specified flags for the user and then, have another different with the real values that will be used by the command (so the one specified by the user would override the default values without modifying any original value)

We should also review the behavior of the namespace flag in other commands to make sure it behaves correctly and we don't have this issue 

**Describe alternatives you've considered**
N/A

**Additional context**
I think this would help to make things less error-prone

PR #2904, opened 2022-07-07
Fix to show endpoints after "okteto deploy"

pchico83:

Fixes https://github.com/okteto/okteto/issues/2885

Testing this requires a non trivial refactor. I would do that as part of https://github.com/okteto/okteto/issues/2886

### DIFF
--- a/cmd/deploy/deploy.go
+++ b/cmd/deploy/deploy.go
@@ -439,7 +439,7 @@ func (dc *DeployCommand) RunDeploy(ctx context.Context, deployOptions *Options) 
 				}
 			}
 			if !utils.LoadBoolean(model.OktetoWithinDeployCommandContextEnvVar) {
-				if err := dc.showEndpoints(ctx, &EndpointsOptions{Name: deployOptions.Name}); err != nil {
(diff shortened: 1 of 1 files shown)


cmd/deploy/deploy.go
```diff
@@ -439,7 +439,7 @@ func (dc *DeployCommand) RunDeploy(ctx context.Context, deployOptions *Options)
 				}
 			}
 			if !utils.LoadBoolean(model.OktetoWithinDeployCommandContextEnvVar) {
-				if err := dc.showEndpoints(ctx, &EndpointsOptions{Name: deployOptions.Name}); err != nil {
+				if err := dc.showEndpoints(ctx, &EndpointsOptions{Name: deployOptions.Name, Namespace: deployOptions.Manifest.Namespace}); err != nil {
```
ifbyol, line 442: just to double check, if the namespace is not specified in the manifest, is it being initialized with the context one or the one specified in the command, right?

pchico83: yes, the value in the manifest is the one taking into account arguments and manifest values

Issue #2855, opened 2022-06-30
Define how the current working directory should be handled when the manifest is within a sub-folder


ifbyol:
**Is your feature request related to a problem? Please describe.**
Some time ago there was change to execute the commands defined in the Okteto manifest from the directory where the manifest is stored instead of executing it from the directory where the `okteto deploy` command is being executed.

For example, imagine I'm in a folder called `root` and I execute `okteto deploy -f subfolder/okteto.yml`

This generated a couple of side effects that we would need to clarify and review:
* The filename stored in the internal config map is not the right one. We are storing only the filename instead of the relative path specified from the user when the command is executed. In the example above it means that we are storing `okteto.yml` instead of `subfolder/okteto.yml`. We have created a ticket for it. https://github.com/okteto/okteto/issues/2846
* The repo URL and branch calculation is now done from the directory where the manifest is stored instead of doing it from the directory where the `okteto deploy` command is executed. In the example above it means that we try to look for a `.git` folder within `subfolder` instead of doing it within `root` We have created an issue for it. https://github.com/okteto/okteto/issues/2849
* The name for the dev environment when it is not specified in the manifest is now being inferred from the folder where the manifest is stored instead of the folder where the `okteto deploy` command is executed. In the example above, we would name the dev environment `subfolder` instead of `root`. This probably make sense.

In other words, it seems we have changed the working directory be always the one where the manifest is stored.

There might be another secondary effects we didn't face yet

pchico83:
When running `okteto deploy` there are two options:
a) the current working directory doesn't belong to a git repo. We can check this by checking if the current working directory or any of its parents has a `.git` folder on it
b) the current working directory belongs to a git repo

For a) we shouldn't store any of the following fields: `filename`, `branch` and `repoURL`.  These fields are for the Okteto UI to be able to show this information, and enable redeploys. But if we are not in a git repo, there is no need to store this information.

For b) we should get the `.git` folder and extract the git URL and branch from it. Also, `filename` should be the relative path of the okteto manifest with respect to the git root folder. Note that this value is a value independent from the current working directory. `filename` should only be stored for okteto manifests or compose files. For automatic pipelines, @ifbyol not sure if we could store the folder

ifbyol:
@pchico83 Yes, we can store the folder. In fact, we were storing the folder but with the changes to use current working directory we don't store folders anymore.

If the user specifies a folder with the `-f` flag, I think we should store that

pchico83:
[...]

PR #2910, opened 2022-07-08
Load context resource only instead of the whole file while loading ctx

### LINKED ISSUE #2728: Warning unmarshall messages are shown two times

jLopezbarb:
Fixes #2728 

## Proposed changes
- When reading a context resource only read the namespace and context and skip the unmarshalling of the rest of the file
- Created package discovery to look for okteto manifest/compose files/okteto pipelines/k8s manifests/helm charts.
- Created package filesystem with functions related to filesystem(I needed to create it because a cyclic import).


pkg/model/stack.go
```diff
@@ -35,22 +37,6 @@ import (
 var (
 	errBadStackName = "must consist of lower case alphanumeric characters or '-', and must start and end with an alphanumeric character"
 	//DefaultStackManifest default okteto stack manifest file
```
ifbyol, line 39: I think this comment doesn't apply anymore

pkg/model/stack.go
```diff
@@ -712,15 +698,19 @@ func LoadStack(name string, stackPaths []string, validate bool) (*Stack, error)
 	var resultStack *Stack
 
 	if len(stackPaths) == 0 {
-		stack, err := inferStack(name)
+		dir, err := os.Getwd()
+		if err != nil {
+			return nil, err
+		}
```
ifbyol, line 704: This might have side effects regarding the working directory change we do when loading the manifest in okteto deploy

ifbyol (commented):
Can we run the integration tests of this branch before merging it to master?

pkg/model/contextresource.go
```diff
@@ -0,0 +1,82 @@
...
+
+	"github.com/okteto/okteto/pkg/discovery"
+	oktetoErrors "github.com/okteto/okteto/pkg/errors"
+	"github.com/okteto/okteto/pkg/filesystem"
+	yaml "gopkg.in/yaml.v3"
+)
+
+// ContextResource provides the context and namespace to operate within a manifest
+type ContextResource struct {
+	Context   string
```
jmacelroy, line 27: The setters for these have some logic that helps provide a guard. Given that they should probably not be exported. Can you check if anyone is using this. If they are setting it without the setter then it may be we have a bug and if they are getting it then we need to add a getter.

jmacelroy: For marshaling within this package
``` 
// For private/package consumption due to yaml marshalling requirements.
type contextResource struct {
    Context string
    Namespace string
}

// For public consumption due to the requirement of a logical setter.
type ContextResource struct {
    context string
    namespace string
}
```

jLopezbarb: That is something that needs to be addressed but I think this PR is big enough to add more changes. Changing that requires changing all `cmd/context` and their references so the number of touched files can go up to 30 files more. Could we do it as part of the `removing model` refactor?

jmacelroy (approved):

jLopezbarb:
> Can we run the integration tests of this branch before merging it to master?

They have already run [here](https://app.circleci.com/pipelines/github/okteto/okteto/7849/workflows/36eddc9f-dacc-45d5-9aaf-bd00c57806f0)

pkg/model/stack.go
ifbyol: This scares me a bit since `LoadStack` could have different result executing it before and after the change we do in the working directory. We rely too much on `os.Getwd` I think it doesn't affect as the previous `inferStack` function was relying also in the current working directory but we should be careful with that.

ifbyol (approved):
My comment about the side effect for relying in the working directory is something that is also happening in the current code and it is something that should be part of this ticket: https://github.com/okteto/okteto/issues/2855

PR #2911, opened 2022-07-08
fix: wrong behaviour when okteto down --all

### LINKED ISSUE #2767: Wrong behavior of `okteto down --all`
**Describe the bug**
When running `okteto down --all`, we should show the dev container name of each dev container being deactivated.
Also, after `okteto down --all` deactivates dev containers, it prompts to activate a new one

Screenshot 2022-06-08 at 23 26 16

**To Reproduce**
Run `okteto down --all` on http://github.com/okteto/movies

AdrianPedriza:

Fixes #2767

## Proposed changes

- do not execute the logic related to ask the user for the dev container to deactivate if the option is `--all`
- add dev container name to deactivate

### DIFF
--- a/cmd/down.go
+++ b/cmd/down.go
@@ -84,9 +84,9 @@ func Down() *cobra.Command {
 					return err
 				}
 
-				analytics.TrackDown(true)
 			}
 
+			analytics.TrackDown(true)
 			return nil
 		},
 	}
--- a/cmd/down.go
+++ b/cmd/down.go
@@ -113,19 +113,16 @@ func allDown(ctx context.Context, manifest *model.Manifest, rm bool) error {
 
 		app, _, err := utils.GetApp(ctx, dev, c, false)
 		if err != nil {
-			if !oktetoErrors.IsNotFound(err) {
-				return err
-			}
-		} else {
-			if apps.IsDevModeOn(app) {
-				isAnyDev = true
-			}
+			return err
 		}
 
-		if err := runDown(ctx, dev, rm); err != nil {
-			analytics.TrackDown(false)
-			err = fmt.Errorf("%w\n    Find additional logs at: %s/okteto.log", err, config.GetAppHome(dev.Namespace, dev.Name))
-			return err
+		if apps.IsDevModeOn(app) {
+			isAnyDev = true
+			if err := runDown(ctx, dev, rm); err != nil {
+				analytics.TrackDown(false)
+				err = fmt.Errorf("%w\n    Find additional logs at: %s/okteto.log", err, config.GetAppHome(dev.Namespace, dev.Name))
+				return err
+			}
 		}
 	}
 
--- a/cmd/down.go
+++ b/cmd/down.go
@@ -93,8 +93,7 @@ func Down() *cobra.Command {
 				if apps.IsDevModeOn(app) {
 					if err := runDown(ctx, dev, rm); err != nil {
 						analytics.TrackDown(false)
-						err = fmt.Errorf("%w\n    Find additional logs at: %s/okteto.log", err, config.GetAppHome(dev.Namespace, dev.Name))
-						return err
+						return fmt.Errorf("%w\n    Find additional logs at: %s/okteto.log", err, config.GetAppHome(dev.Namespace, dev.Name))
 					}
 				} else {
 					oktetoLog.Success(fmt.Sprintf("Development container '%s' deactivated", dev.Name))
@@ -138,8 +137,7 @@ func allDown(ctx context.Context, manifest *model.Manifest, rm bool) error {
 			spinner.Stop()
 			if err := runDown(ctx, dev, rm); err != nil {
 				analytics.TrackDown(false)
-				err = fmt.Errorf("%w\n    Find additional logs at: %s/okteto.log", err, config.GetAppHome(dev.Namespace, dev.Name))
-				return err
+				return fmt.Errorf("%w\n    Find additional logs at: %s/okteto.log", err, config.GetAppHome(dev.Namespace, dev.Name))
 			}
 		}
 	}
--- a/cmd/down.go
+++ b/cmd/down.go
@@ -68,6 +68,8 @@ func Down() *cobra.Command {
 				if err != nil {
 					return err
 				}
+
+				oktetoLog.Success("All development containers are deactivated")
 			} else {
 				devName := ""
 				if len(args) == 1 {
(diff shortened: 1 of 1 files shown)


cmd/down.go
```diff
@@ -68,24 +68,25 @@ func Down() *cobra.Command {
...
+					analytics.TrackDown(false)
+					err = fmt.Errorf("%w\n    Find additional logs at: %s/okteto.log", err, config.GetAppHome(dev.Namespace, dev.Name))
+					return err
+				}
 
-			if err := runDown(ctx, dev, rm); err != nil {
-				analytics.TrackDown(false)
-				err = fmt.Errorf("%w\n    Find additional logs at: %s/okteto.log", err, config.GetAppHome(dev.Namespace, dev.Name))
-				return err
+				analytics.TrackDown(true)
```
ifbyol, line 87: I think this has to be outside the else, right? I mean, we want to track the command execution even when the flag `--all` is specified, right?

AdrianPedriza:
> this repo


https://user-images.githubusercontent.com/22500940/178021984-b99048f9-c88b-4ebe-9283-3ef63c3d5858.mp4

AdrianPedriza:
> @AdrianPedriza could you post a quick video running `okteto down --all`? you can use [this repo](https://github.com/okteto/movies-with-helm) as sample


https://user-images.githubusercontent.com/22500940/178022186-14e8131f-fae9-4c95-bbc4-46d604479045.mp4

pchico83:
@AdrianPedriza could we also show the dev container name in the "deactivating dev container" spinner message?
Also, if the dev container is not active, could we skip the deactivating message?

AdrianPedriza:
> @AdrianPedriza could we also show the dev container name in the "deactivating dev container" spinner message? Also, if the dev container is not active, could we skip the deactivating message?

@pchico83 I have uploaded your proposal. For cases where there is no dev container I have added a message: "You do not have any dev environment activated". It is OK for you?

pchico83:
@AdrianPedriza let's revisit the spinner behavior when running `okteto down`, `okteto down xxx` and `okteto down -all`. Both, when dev containers are active, and when they are not.

cmd/down.go
```diff
@@ -99,22 +113,41 @@ func Down() *cobra.Command {
...
-		if err := runDown(ctx, dev, rm); err != nil {
-			analytics.TrackDown(false)
-			err = fmt.Errorf("%w\n    Find additional logs at: %s/okteto.log", err, config.GetAppHome(dev.Namespace, dev.Name))
+		app, _, err := utils.GetApp(ctx, dev, c, false)
+		if err != nil {
 			return err
 		}
+
+		if apps.IsDevModeOn(app) {
+			spinner.Stop()
```
pchico83, line 136: there is a `defer spinner.Stop()`, you don't need it here, right?

cmd/down.go
```diff
@@ -68,23 +68,37 @@ func Down() *cobra.Command {
...
+					}
+				}
 
-			if err := runDown(ctx, dev, rm); err != nil {
-				analytics.TrackDown(false)
-				err = fmt.Errorf("%w\n    Find additional logs at: %s/okteto.log", err, config.GetAppHome(dev.Namespace, dev.Name))
-				return err
 			}
 
+			oktetoLog.Success("Success deactivating development containers")
```
pchico83, line 101: I would have two different messages here:
- `okteto down --all` -> All development containers are deactivated
- `okteto down xxx` -> Development container 'xxx' deactivated

pchico83 (approved):

cmd/down.go
AdrianPedriza: if I don't stop the spinner at this point, it will repeatedly swap the spinner message of the allDown function with that of the runDown function in a rather 'crazy' way.

pchico83: maybe you should pass the spinner as an argument to `runDown`?

AdrianPedriza: yes, it could be an option. Another option I have thought of is to create a struct with all these options available in the command: (like upOptions in the `up` command) to avoid functions with too many parameters. Maybe it would be cleaner.

AdrianPedriza:
> @AdrianPedriza this is part of `else`, but it should also run if `app.IsDevModeOn` is `true`, right?

yes, the message is printed also if `app.IsDevModeOn` is `true`. Since function `runDown` is used for both option (down all or down svc) this is the way I followed to cover all scenarios. It is a little weird to have message inside and outside runDown, agree. If we want to avoid this, a deeper refactor on the command will be needed I guess.

AdrianPedriza:
@pchico83 WDYT about open a new issue related to refactor down command for everything discussed in the issue? Since we have now the expected behavior in this PR maybe we could merge it with the current implementation.

pchico83:
> @pchico83 WDYT about open a new issue related to refactor down command for everything discussed in the issue? Since we have now the expected behavior in this PR maybe we could merge it with the current implementation.

@AdrianPedriza yes, makes sense

cmd/down.go
```diff
@@ -68,21 +68,37 @@ func Down() *cobra.Command {
...
+
+				app, _, err := utils.GetApp(ctx, dev, c, false)
+				if err != nil {
+					return err
+				}
+
+				if apps.IsDevModeOn(app) {
+					if err := runDown(ctx, dev, rm); err != nil {
+						analytics.TrackDown(false)
+						err = fmt.Errorf("%w\n    Find additional logs at: %s/okteto.log", err, config.GetAppHome(dev.Namespace, dev.Name))
```
jLopezbarb, line 96:
```suggestion
						return fmt.Errorf("%w\n    Find additional logs at: %s/okteto.log", err, config.GetAppHome(dev.Namespace, dev.Name))
```

cmd/down.go
```diff
@@ -68,21 +68,37 @@ func Down() *cobra.Command {
...
 
-			devName := ""
-			if len(args) == 1 {
-				devName = args[0]
-			}
-			dev, err := utils.GetDevFromManifest(manifest, devName)
-			if err != nil {
-				return err
-			}
+				oktetoLog.Success("All development containers are deactivated")
```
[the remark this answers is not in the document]
ifbyol: No, we would lose the `analytics.TrackDown(true)`. If we do that, we should make sure we include that call also here before the `return nil`

ifbyol (approved):

cmd/down.go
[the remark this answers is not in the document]
ifbyol: oh, then we are currently tracking twice the event when the `--all` flag is passed

jLopezbarb (approved):

teresaromero (approved):

PR #2915, opened 2022-07-11
Use GetKubernetesTimeout to configure our k8s proxy timeouts

pchico83:
Several users have complained about `okteto deploy` timeouts.
We were using a 10s timeout. By default, kubectl and helm don't use timeouts, I think we should follow the same approach when configuring our k8s proxy. I also added an envvar to configure the `OKTETO_SSH_TIMEOUT`

## Proposed changes

- Configure okteto deploy k8s proxy timeouts based on `OKTETO_KUBERNETES_TIMEOUT`
- Added `OKTETO_SSH_TIMEOUT` envvar to configure the ssh timeout

### DIFF
--- a/cmd/deploy/proxy.go
+++ b/cmd/deploy/proxy.go
@@ -34,6 +34,7 @@ import (
 	"github.com/okteto/okteto/pkg/k8s/labels"
 	oktetoLog "github.com/okteto/okteto/pkg/log"
 	"github.com/okteto/okteto/pkg/model"
+	"github.com/okteto/okteto/pkg/okteto"
 	appsv1 "k8s.io/api/apps/v1"
 	batchv1 "k8s.io/api/batch/v1"
 	apiv1 "k8s.io/api/core/v1"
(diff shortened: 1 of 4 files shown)


cmd/deploy/proxy.go
```diff
@@ -94,9 +95,9 @@ func NewProxy(kubeconfig *KubeConfig) (*Proxy, error) {
 	s := &http.Server{
 		Addr:         fmt.Sprintf(":%d", port),
 		Handler:      handler,
-		ReadTimeout:  5 * time.Second,
-		WriteTimeout: 10 * time.Second,
-		IdleTimeout:  120 * time.Second,
+		ReadTimeout:  okteto.GetKubernetesTimeout() * time.Second,
```
jLopezbarb, line 98: Do we need to have `* time.Second`. I would add the `* time.Second` on `okteto.GetKubernetesTimeout()` function. I think it's cleaner
```suggestion
		ReadTimeout:  okteto.GetKubernetesTimeout(),
```

pchico83: @jLopezbarb @jpf-okteto updated!

jLopezbarb (approved):

PR #2923, opened 2022-07-13
login error page

teresaromero:

Fixes https://github.com/okteto/app/issues/4173 
Fixed https://github.com/okteto/app/issues/4174

## Proposed changes

- This PR implements a new error page when login for a custom new error described at the issues
- The actual flow for login with error has a bug and we display the success page even if the flow has been not successfull. We do not have the copy messaged to display in a case of a general error, so this PR does not cover this changes, when the error is a general login error we still display the success page. This should be fixed with this issue: https://github.com/okteto/app/issues/4189


pkg/html/html.go
```diff
@@ -0,0 +1,28 @@
+package html
```
jLopezbarb, line 1: Missing copyright

jLopezbarb: Is `HTML` package only used by `cmd/login`? Maybe it's a sub-package

pkg/cmd/login/browser.go
```diff
@@ -30,41 +33,36 @@ type Handler struct {
...
 			h.errChan <- fmt.Errorf("invalid request state")
-			w.WriteHeader(http.StatusBadRequest)
 			return
 		}
 
-		data := &OriginData{
-			Meta:    "http-equiv=\"Content-type\" content=\"text/html; charset=utf-8\"",
-			Origin:  "the Okteto CLI",
-			Message: "Close this window and go back to your terminal",
+		ctx := context.Background()
```
ifbyol, line 50:
```suggestion
		ctx := r.Context()
```

teresaromero:
```suggestion
		ctx := h.Context()
```

??

teresaromero: is the one from the handler or the requesT?

pkg/errors/errors.go
```diff
@@ -178,6 +178,9 @@ var (
 
 	//ErrNotManifestContentDetected is raised when cannot load any field accepted by okteto manifest doc
 	ErrNotManifestContentDetected = errors.New("couldn't detect okteto manifest content")
+
+	//ErrGithubMissingBusinessEmail is raised when the user does not have a business email
```
jLopezbarb, line 182: NIT: I'd add this error on `pkg/okteto/errors` since is an error that is only thrown by that package

pkg/errors/errors.go
```diff
@@ -250,3 +253,18 @@ func IsClosedNetwork(err error) bool {
...
+// IsErrGithubMissingBusinessEmail returns true if the error is ErrGithubMissingBusinessEmail
+func IsErrGithubMissingBusinessEmail(err error) bool {
+	return err.Error() == ErrGithubMissingBusinessEmail.Error()
+}
+
+// TranslateAuthError returns a legible error text depending if is custom error
+func TranslateAuthError(err error) error {
+	switch {
+	case IsErrGithubMissingBusinessEmail(err):
+		return errors.New("Please add a business/work email to your Github account at https://github.com/settings/emails or try again with a different account")
```
jLopezbarb, line 266: Should this be a constant?

teresaromero: the error translated?

jLopezbarb (approved):

pkg/okteto/errors.go
```diff
@@ -0,0 +1,23 @@
...
+import "errors"
+
+var (
+	//ErrGithubMissingBusinessEmail is raised when the user does not have a business email
+	ErrGithubMissingBusinessEmail = errors.New("github-missing-business-email")
+)
+
+// IsErrGithubMissingBusinessEmail returns true if the error is ErrGithubMissingBusinessEmail
+func IsErrGithubMissingBusinessEmail(err error) bool {
+	return err.Error() == ErrGithubMissingBusinessEmail.Error()
```
AgustinRamiroDiaz, line 12: Check out `error.Is()`, maybe it's more suitable to check if the error is wrapped

pkg/cmd/login/html/html.go
```diff
@@ -0,0 +1,45 @@
...
+
+	"github.com/okteto/okteto/pkg/okteto"
+)
+
+type templateData map[string]interface{}
+
+// ExecuteSuccess renders the login success page at the browser
+func ExecuteSuccess(w http.ResponseWriter) error {
+	data := successData()
+	successTemplate := template.Must(template.New("success-response").Parse(commonTemplate))
```
AgustinRamiroDiaz, line 28: We should move the `template.Must` to the module initialization, so we crash early if the template is malformed. Same thing in `ExecuteError`

AgustinRamiroDiaz: Adding a test should be another easy win, since the content is static

AgustinRamiroDiaz (commented):
Nice work! How did you find the auth flow? Was it hard to understand? Is there some improvements/refactors you thought that could make it better?

pkg/cmd/login/html/html.go
jLopezbarb: I think it should stay in a function, otherwise, it'll be loaded when importing the package, even if we don't use the function.

ifbyol: In this case I think it is not a big deal to fail with a panic as it is not a server that would crash and generate a restart like our API. This is just a command execution, the only thing to bear in mind the user experience. I think users should not see a `panic`.

Having said that, I don't see the need of doing a `Must`, we can simply use `successTemplate, err := template.New("success-response").Parse(commonTemplate)` and return an error if it is not nil (as `ExecuteSuccess` already may return an error. 

As the `Must` function documentation says:

```
// Must is a helper that wraps a call to a function returning (*Template, error)
// and panics if the error is non-nil. It is intended for use in variable
// initializations such as
//	var t = template.Must(template.New("name").Parse("text"))
```

ifbyol (approved):

PR #2930, opened 2022-07-14
fix: exec dev selector

### LINKED ISSUE #2768: `okteto exec xxx -- bash` is ignoring the dev container name xxx
**To Reproduce**
- Clone http://github.com/okteto/movies
- Run `okteto up api` and wait for it to finish
- Run `okteto exec api -- bash`

You should get a bash terminal in the `api` container. Instead, we are prompting for the dev container where `exec` is going to happen:
Screenshot 2022-06-08 at 23 30 53

jLopezbarb:

Fixes #2768

## Proposed changes
- If the first argument is a dev name in the manifest dev section use it as the devName

### DIFF
--- a/cmd/exec_test.go
+++ b/cmd/exec_test.go
@@ -90,7 +90,6 @@ func TestGetCommandFromArgs(t *testing.T) {
 		manifest     *model.Manifest
 		args         []string
 		expectedArgs []string
-		expectedErr  error
 	}{
 		{
 			name: "first arg is on dev section",
@@ -101,7 +100,6 @@ func TestGetCommandFromArgs(t *testing.T) {
 			},
 			args:         []string{"autocreate", "sh"},
 			expectedArgs: []string{"sh"},
-			expectedErr:  nil,
 		},
 		{
 			name: "only one argument/same name as dev",
@@ -112,7 +110,6 @@ func TestGetCommandFromArgs(t *testing.T) {
 			},
 			args:         []string{"sh"},
 			expectedArgs: []string{"sh"},
-			expectedErr:  nil,
 		},
 		{
 			name: "several argument/first argument same name as dev",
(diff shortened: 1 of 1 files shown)


pkg/model/manifest_test.go
```diff
@@ -727,3 +727,35 @@ sync:
...
+			out:     true,
+		},
+		{
+			name: "devName is not on dev section",
+			devSection: ManifestDevs{
+				"autocreate": &Dev{},
+			},
+			devName: "autocreate",
+			out:     true,
+		},
```
AdrianPedriza, line 753: what about adding a test case that returns false?

cmd/utils/selector.go
```diff
@@ -38,6 +39,9 @@ const (
 	showCursor = esc + "?25h"
 )
 
+// ErrInvalidOption is raised when the selector has an invalid option
```
AdrianPedriza, line 42: why not in errors pkg?

jLopezbarb: Let's start adding errors in the package that creates them, that way we can do `selector.ErrInvalidOption`, which provides to know which package is raising the error instead of having a general errors package

AdrianPedriza (approved):

ifbyol (approved):
We are relying on the number of args and the existence of a service with the value of the first argument to know if we should display the selector or not.

I think there might a corner case that might happen.

Imagine I have a manifest with 2 services (`sh` and `front`) and then I execute `okteto exec -- sh -c python`. We would execute the command `-c python` on the service `sh`, which might not be the case. (Names in the example are not the best, but it is just to expose the edge case)

I know this is a corner case but I'm wondering if there could be another easy way of distinguish which part refers to the service name and which one refers to the command to be executed without including a flag (in the rest of commands, the dev container is specified as argument).

If there is no other way, I think we can leave with it

PR #2937, opened 2022-07-18
added condition to avoid dockerfile with "."

mnevadom:
It was showing the warning of two dockerfiles because it was comparing `Dockerfile` with `./Dockerfile`

### DIFF
--- a/pkg/cmd/build/build.go
+++ b/pkg/cmd/build/build.go
@@ -274,7 +274,7 @@ func extractFromContextAndDockerfile(context, dockerfile, svcName string) string
 		return dockerfile
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/cmd/build/build.go
```diff
@@ -274,7 +274,7 @@ func extractFromContextAndDockerfile(context, dockerfile, svcName string) string
 		return dockerfile
 	}
 
-	if joinPath != dockerfile && filesystem.FileExistsAndNotDir(dockerfile) {
+	if joinPath != filepath.Join(".", dockerfile) && filesystem.FileExistsAndNotDir(dockerfile) {
```
pchico83, line 277: Could we sanitize the paths instead? this way to check equity seems very error prone

jLopezbarb (approved):

PR #2939, opened 2022-07-19
fix: manifest deploying a compose

jLopezbarb:

Fixes Bug calculating volume mounts image to built

## Proposed changes
- Created function to check the correct image name
- Added an e2e test to deploy a compose from a manifest v2


pkg/cmd/build/build.go
```diff
@@ -239,10 +239,12 @@ func OptsFromBuildInfo(manifestName, svcName string, b *model.BuildInfo, o *type
 		if len(b.VolumesToInclude) > 0 {
 			b.Image = fmt.Sprintf("%s/%s-%s:%s", targetRegistry, manifestName, svcName, model.OktetoImageTagWithVolumes)
 		}
-
 	}
 
-	file := extractFromContextAndDockerfile(b.Context, b.Dockerfile, svcName)
+	file := ""
```
mnevadom, line 244: It was like this before: 

`file := b.Dockerfile`

Maybe we should add it before the IF to avoid breaking some case...

mnevadom (approved):

PR #2943, opened 2022-07-19
fix: redeploy service with different name

### LINKED ISSUE #2843: Deploying an stack with same service modifies the existing one
**Describe the bug**
Deploying the same compose with different name skips the creation of endpoints and deployments/sfs/jobs but modifies the current k8s service

**To Reproduce**

1. Copy the following docker-compose:
``` yaml
services:
  test:
    image: alpine
    ports:
      - 8080:8080
```
2. Run `okteto deploy --name test` on the compose folder 
3. Run `okteto deploy --name test-modify` on the compose folder 
4. Check that there are warnings skipping the creation of deployment and endpoint
5. Run `kubectl describe services test` and check that the label `stack.okteto.com/name` has the value of the second deploy (` test-modify`)

**Expected behavior**
It should skip the creation of the k8s service

- Version 2.4.0

jLopezbarb:

Fixes #2843 

## Proposed changes
- show a warning if the service is conflicting with another service
- Added unit testing with the scenarios


pkg/cmd/stack/deploy.go
```diff
@@ -555,6 +554,43 @@ func isAnyPortAvailable(ctx context.Context, svc *model.Service, stack *model.St
...
+			return err
+		}
+		spinner.Stop()
+		oktetoLog.Success("Endpoint '%s' created", svcName)
+		spinner.Start()
+		return nil
+	}
+
+	if old.GetLabels()[model.StackNameLabel] == "" {
+		oktetoLog.Warning("skipping deploy of %s due to name collision: the ingress '%s' was running before deploying your compose", svcName)
```
AgustinRamiroDiaz, line 574: I think the deploy could be from another place than the compose, right?

pkg/cmd/stack/deploy.go
```diff
@@ -555,6 +554,43 @@ func isAnyPortAvailable(ctx context.Context, svc *model.Service, stack *model.St
...
+		if err := services.Deploy(ctx, svcK8s, c); err != nil {
+			return err
+		}
+		spinner.Stop()
+		oktetoLog.Success("Endpoint '%s' created", svcName)
+		spinner.Start()
+		return nil
+	}
+
+	if old.GetLabels()[model.StackNameLabel] == "" {
```
AgustinRamiroDiaz, line 573: What is this logic? If the `model.StackNameLabel` is empty, we guess it's the ingress?

AdrianPedriza (approved):

pkg/cmd/stack/deploy.go
jLopezbarb: No, We check that the new and the old services come from a compose service by checking their labels, so the user will only get the message when the run from a compose

jLopezbarb: Yes, probably is not the right words but what we tried to explain to the user is that there is a k8s service already running with that name and we are not going to deploy the k8s service

pkg/cmd/stack/deploy.go
jLopezbarb: We are checking if there was a service deployed by another compose to show the user a proper message

AgustinRamiroDiaz (approved):

pkg/cmd/stack/deploy.go
```diff
@@ -555,6 +554,43 @@ func isAnyPortAvailable(ctx context.Context, svc *model.Service, stack *model.St
...
+		oktetoLog.Warning("skipping deploy of kubernetes service %s due to name collision: the service '%s' was running before deploying your compose", svcName)
```
ifbyol, line 574: I think you cannot use one single argument to convert 2 placeholders `%s`. You would need to do one of these:

`oktetoLog.Warning("skipping deploy of kubernetes service %s due to name collision: the service '%s' was running before deploying your compose", svcName, svcName)`

or 

`oktetoLog.Warning("skipping deploy of kubernetes service %[1]s due to name collision: the service '%[1]s' was running before deploying your compose", svcName)`

pkg/cmd/stack/deploy.go
```diff
@@ -555,6 +554,43 @@ func isAnyPortAvailable(ctx context.Context, svc *model.Service, stack *model.St
...
+		return nil
+	}
+
+	if old.GetLabels()[model.StackNameLabel] != svcK8s.GetLabels()[model.StackNameLabel] {
+		spinner.Stop()
+		oktetoLog.Warning("skipping creation of kubernetes '%s' due to name collision with endpoint in stack '%s'", svcName, old.GetLabels()[model.StackNameLabel])
```
ifbyol, line 580: Are we still using `stack` when displaying messages to the user?

ifbyol (approved):
Minor comments, functionality looks good. I have doubts about using `stack` in a message to the user. I though we were trying to deprecate that nomenclature

pkg/cmd/stack/deploy.go
```diff
@@ -555,6 +554,43 @@ func isAnyPortAvailable(ctx context.Context, svc *model.Service, stack *model.St
 	return false
 }
 
+func deployK8sService(ctx context.Context, svcName string, s *model.Stack, c kubernetes.Interface, spinner *utils.Spinner) error {
+	svcK8s := translateService(svcName, s)
+	old, err := services.Get(ctx, svcName, s.Namespace, c)
+	if err != nil {
+		if !oktetoErrors.IsNotFound(err) {
+			return fmt.Errorf("error getting ingress '%s': %w", svcName, err)
```
teresaromero, line 562: is this for the ingress or for the service? 🤔

AgustinRamiroDiaz (approved):

teresaromero (approved):

PR #2952, opened 2022-07-20
golang-ci-lint-fixes

AgustinRamiroDiaz:
## Proposed changes

- address feedback from `golangci-lint`

### DIFF
--- a/pkg/model/volumes.go
+++ b/pkg/model/volumes.go
@@ -141,6 +141,7 @@ func (dev *Dev) getSourceSubPath(path string) string {
 			}
 			if filepath.IsAbs(path) {
 				oktetoLog.Info("could not retrieve subpath")
+				path = filepath.Base(path)
 			} else {
 				p, err := filepath.Abs(path)
 				if err != nil {
--- a/pkg/model/volumes.go
+++ b/pkg/model/volumes.go
@@ -141,7 +141,6 @@ func (dev *Dev) getSourceSubPath(path string) string {
 			}
 			if filepath.IsAbs(path) {
 				oktetoLog.Info("could not retrieve subpath")
-				path = filepath.Base(path)
 			} else {
 				p, err := filepath.Abs(path)
 				if err != nil {
--- a/pkg/okteto/client.go
+++ b/pkg/okteto/client.go
@@ -16,13 +16,10 @@ package okteto
 import (
 	"context"
 	"fmt"
-	"io"
 	"net/http"
 	"net/url"
 	"os"
 	"strings"
-
-	"net/http/httputil"
 
 	oktetoErrors "github.com/okteto/okteto/pkg/errors"
 	oktetoLog "github.com/okteto/okteto/pkg/log"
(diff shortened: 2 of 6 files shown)


pkg/okteto/client.go
```diff
@@ -233,36 +230,3 @@ func (c *OktetoClient) Pipeline() types.PipelineInterface {
 func (c *OktetoClient) User() types.UserInterface {
 	return c.user
 }
-
-type traceTransport struct {
```
ifbyol, line 237: iirc correctly this was some utility included by @maroshii to debug issues. Fran, am I right? is this needed/helpful?

AgustinRamiroDiaz: I'm in favor of keeping it if it's useful. We can add linting annotations so that we don't get the linting
https://golangci-lint.run/usage/false-positives/#nolint-directive

ifbyol: I've never used it, so not sure how useful it is, I guess we can always document it in Notion as a helper and deleted. As I've never used it, my vote would be to delete it, but I'm ok also with adding the `nolint` directive just in case

pkg/model/volumes.go
```diff
@@ -141,7 +141,6 @@ func (dev *Dev) getSourceSubPath(path string) string {
 			}
 			if filepath.IsAbs(path) {
 				oktetoLog.Info("could not retrieve subpath")
-				path = filepath.Base(path)
```
ifbyol, line 144: This is the only thing that scares me a bit. We have faced several issues in the past with paths and things like that. I know this is not being used but it can be just because of a bug. @AgustinRamiroDiaz Can you check when this started to be useless to see if it was just that we forgot to removed or a potential bug?

pkg/okteto/client.go
AgustinRamiroDiaz: That'd be good. @maroshii could you explain how was this helpful when debugging?

pkg/model/volumes.go
ifbyol: Talked offline with @AgustinRamiroDiaz . It seems that this line was introduced some weeks ago with a possible fix and the first release where it was included was the `2.5.0`, so not sure if it is a real bug (if we delete this line, the if doesn't do anything and not sure it is correct). So we'll create a ticket to take a look at that carefully.

Agustin, can you create it and post it here?

AgustinRamiroDiaz: Done! https://github.com/okteto/okteto/issues/2967

jLopezbarb: We can safely remove this line, it was added because of a bug and after that, we fixed another one and we left that line but we should have cleaned that before

jLopezbarb (approved):

pkg/model/volumes.go
ifbyol: but @jLopezbarb that line was introduced in the change just previous to this removal. Without that line, the if doesn't make sense. Are we sure it wasn't done on purpose?

maroshii: This can be removed. `path` is mutated inside the `if` but never used after that. It is used in the `else` which doesn't affect. We can probably also reverse the `if`

AgustinRamiroDiaz: I agree with @ifbyol, this could be a potential bug where we forget to use a variable. That's why I wanted one more look from @jLopezbarb, who knows which behavior we need here

jLopezbarb: Since we created a new issue for that, can we revert this change and work on it in another PR?

PR #2954, opened 2022-07-21
Fallback to Preview endpoint for waiting for dependencies

pchico83:

When we wait for resources to be running, if we are in a preview environment instead of a namespace, the `GetSpace` query returns `not-found`.
This PR fallbacks to the `Preview` endpoint

### DIFF
--- a/pkg/okteto/pipeline.go
+++ b/pkg/okteto/pipeline.go
@@ -19,6 +19,7 @@ import (
 	"strings"
 
 	"github.com/okteto/okteto/pkg/config"
+	"github.com/okteto/okteto/pkg/errors"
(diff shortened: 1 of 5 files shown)


jLopezbarb (approved):

pkg/okteto/pipeline.go
```diff
@@ -19,6 +19,7 @@ import (
 	"strings"
 
 	"github.com/okteto/okteto/pkg/config"
+	"github.com/okteto/okteto/pkg/errors"
```
jLopezbarb, line 22: duplicated import, we are already importing it as `oktetoErrors`

jLopezbarb (approved):

PR #2955, opened 2022-07-21
fix: volumes not being deleted

### LINKED ISSUE #2759: Volumes Not Always Removed with Down and Destroy Commands
**Describe the bug**
This flow was identified by a prospective Customer.  They initially did 'CLI: okteto destroy' but realized their volumes (in the UI) were still alive.  They attempted 'CLI: okteto destroy -v' and it will not go back and clean up the volumes.  However, has they originally ONLY executed 'CLI: okteto destroy -v' the desired fully clean state is achieved.  As an end user, you may not realize volumes are present so being able to go back and re-run the command with the flag(s) would be an expected behaviour.

**To Reproduce**
Steps to reproduce the behaviour.  This example only creates and destroys a namespace with no other operations in between.

1. Create a new namespace e.g. 'foo' on demo.okteto.dev
2. Launch a new development environment from the UI
3. Clone code locally & set context to demo / namespace to 'foo'
4. Reproduce error by destroying without specifying volumes then attempting to destroy with volumes
5. $ okteto destroy # notice Volumes in UI remain as Unused
6. $ okteto destroy -v # notice Volumes in UI remain as Unused
7. $ kubectl get pvc # shows that the volumes still remain

NOTE 1#: If instead you originally executed only "$ okteto destroy -v" in step #5 , everything gets cleaned up as expected.

NOTE 2#: In addition, In between steps 3# and 4#, we did an  'okteto up vote' followed by 'okteto down -v vote' and the outcome described above was the same.

**Expected behavior**
Executing "CLI: okteto destroy" followed by "CLI: okteto destroy -v" should have the same outcome as only executing "CLI: okteto destroy -v".   From a users perspective, specifying the volumes can easily be overlooked so they should be able to re-execute to clean things up.

- OS: MacOS
- Okteto CLI: 2.3.2

**Additional context**

jLopezbarb:

Fixes #2759 

## Proposed changes
- Check on proxy volume claims when we create an statefulset

### DIFF
--- a/cmd/deploy/proxy.go
+++ b/cmd/deploy/proxy.go
@@ -218,6 +218,7 @@ func (ph *proxyHandler) getProxyHandler(token string, clusterConfig *rest.Config
 
 		// Modify all resources updated or created to include the label.
 		if r.Method == "PUT" || r.Method == "POST" {
+			isCreation := r.Method == "POST"
 			b, err := io.ReadAll(r.Body)
 			if err != nil {
 				oktetoLog.Infof("could not read the request body: %s", err)
@@ -230,7 +231,7 @@ func (ph *proxyHandler) getProxyHandler(token string, clusterConfig *rest.Config
 				return
 			}
 
-			b, err = ph.translateBody(b, r.Method)
+			b, err = ph.translateBody(b, isCreation)
 			if err != nil {
 				oktetoLog.Info(err)
 				rw.WriteHeader(500)
@@ -258,7 +259,7 @@ func (ph *proxyHandler) SetDivert(divertedNamespace string) {
 	ph.DivertedNamespace = divertedNamespace
 }
 
-func (ph *proxyHandler) translateBody(b []byte, method string) ([]byte, error) {
+func (ph *proxyHandler) translateBody(b []byte, isCreation bool) ([]byte, error) {
 	var body map[string]json.RawMessage
 	if err := json.Unmarshal(b, &body); err != nil {
 		oktetoLog.Infof("error unmarshalling resource body on proxy: %s", err.Error())
@@ -281,7 +282,7 @@ func (ph *proxyHandler) translateBody(b []byte, method string) ([]byte, error) {
 			return nil, err
 		}
 	case "StatefulSet":
-		if err := ph.translateStatefulSetSpec(body, method); err != nil {
+		if err := ph.translateStatefulSetSpec(body, isCreation); err != nil {
 			return nil, err
 		}
 	case "Job":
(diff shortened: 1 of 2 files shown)


cmd/deploy/proxy.go
```diff
@@ -466,3 +469,18 @@ func (ph *proxyHandler) applyDivert(podSpec *apiv1.PodSpec) {
 	searches = append(searches, podSpec.DNSConfig.Searches...)
 	podSpec.DNSConfig.Searches = searches
 }
+
+func (ph *proxyHandler) translatePvcs(podSpec *apiv1.PodSpec) {
```
pchico83, line 473: is this function in-use?

cmd/deploy/proxy.go
```diff
@@ -363,6 +363,9 @@ func (ph *proxyHandler) translateStatefulSetSpec(body map[string]json.RawMessage
 		return nil
 	}
 	labels.SetInMetadata(&spec.Template.ObjectMeta, model.DeployedByLabel, ph.Name)
+	for _, pvc := range spec.VolumeClaimTemplates {
+		labels.SetInMetadata(&pvc.ObjectMeta, model.DeployedByLabel, ph.Name)
```
pchico83, line 367: Let's do this just on create. Otherwise, an update of an existing sfs will fail because sfs don't allow label changes

jLopezbarb: I think we can do it always. If the stateful set is already deployed on another config map, the user will receive a warning message saying that we will skip that service because it's already deployed, and if the user redeploys the application, the deployed-by label will be the same so no changes will be applied

ifbyol: @jLopezbarb but what would happen for a sfs that, being already deployed for a dev environment with a previous version of the CLI, is redeployed using this new code? It would add the new label. I mean, the redeploy would happen for the same configmap so the warning wouldn't be shown.

If I try to include the label in the VolumeClaimTemplate for an already deployed statefulset, I get this error:

```
# statefulsets.apps "postgresql" was not valid:
# * spec: Forbidden: updates to statefulset spec for fields other than 'replicas', 'template', 'updateStrategy' and 'minReadySeconds' are forbidden
```

We should only include the label in the creation

cmd/deploy/proxy_test.go
```diff
@@ -34,7 +34,7 @@ func Test_TranslateInvalidResourceBody(t *testing.T) {
 		tt := tt
 		t.Run(tt.name, func(t *testing.T) {
 			t.Parallel()
-			_, err := ph.translateBody(tt.body)
+			_, err := ph.translateBody(tt.body, true)
```
AdrianPedriza, line 37: could we cover the case of not being a POST method?

AdrianPedriza (approved):

ifbyol (approved):

PR #2969, opened 2022-07-28
Fix okteto destroy manifest not found with k8s.yml

### LINKED ISSUE #2968: `okteto destroy` stops working when the dev environment doesn't have any okteto manifest or docker-compose
**Describe the bug**
If I have a dev environment that doesn't have an okteto manifest or docker-compose (for example, containing a k8s manifest) and I execute an `okteto destroy`, it fails with the error:

```
 x  Could not detect any okteto manifest
    If you have an okteto manifest file, use the flag '--file' to point to your okteto manifest file
```

It also fails if I specify the name of the dev environment that has to be deleted. I checked and this error wasn't happening in `2.4.x` release.

The dev environment is deployed successfully if I do `okteto deploy`.

**To Reproduce**

1. Get the CLI `2.5.0` or `2.5.1`
2. Clone the repo http://github.com/okteto/go-getting-started
3. In the root folder of the repo, execute `okteto deploy`. It should deploy successfully
4. Once deployment finish, execute `okteto destroy` and you should get the error

**Expected behavior**
Destroy operation should be executed as it was working in previous versions. The manifest for destroy operations is only used for getting info like dev environment name or context (but they have default values) or the destroy commands to be executed if any but it shouldn't be required.

- CLI version `2.5.0` and `2.5.1` 

**Additional context**
N/A

teresaromero:

Fixes #2968 

## Proposed changes

- Both LoadManifestWithContext and LoadContextFromPath where returning error ErrManifestNotFound as a result of GetContextResourcePath not having k8s as manifest discovery. <del>Added k8s files to be discovered as last resource of priority, when other files are not present.</del> This function only loads context, so k8s files are not present there for discovery for that reason. If no manifest with context info is found, context is init with the default. 
- <del>This solves the issue within the `okteto destroy` as `contextCMD.LoadContextFromPath` was returning the error when not finding the k8s manifest.</del> `okteto destroy` was returning err in case LoadContextFromPath had error, so this has been change to do the same as deploy, if err is ErrNotLogged, return the error, but if not, run context with default or flag values
- Why did `okteto deploy` worked ? <del>When calling `contextCMD.LoadContextFromPath` at this cmd, and we got error, unless this error was logged in error, we were creating a new context and going forward with the deployment. When `RunDeploy -> GetManifestV2 -> GetInferredManifest -> GetK8sManifestPath` so this time, we discovered the k8s manifest and deployment could continue.</del> because when LoadContextFromPath returned err != ErrNotLogged the context is initialized from default or flag values.
- Why did `okteto up` failed also? When `okteto up -> LoadManifestWithContext` returned error, there is an exception to continue and ask for `okteto init` if no manifest is found, error was being returned because we were checking the err was `oktetoErrors.ErrManifestNotFound`, and not `discovery.ErrOktetoManifestNotFound` which is the one returned in case the discovery service has not found any at the `LoadManifestWithContext` . this will still fail but was not scoped of this PR as only min changes to fix the destroy are made

### DIFF
--- a/cmd/context/utils.go
+++ b/cmd/context/utils.go
@@ -222,7 +222,10 @@ func LoadContextFromPath(ctx context.Context, namespace, k8sContext, path string
 	}
 	ctxResource, err := model.GetContextResource(path)
 	if err != nil {
-		return err
+		// here discovery.ErrOktetoManifestNotFound will return as error
+		if !errors.Is(err, oktetoErrors.ErrManifestNotFound) {
+			return err
+		}
 	}
 
 	if err := ctxResource.UpdateNamespace(namespace); err != nil {
--- a/cmd/context/utils.go
+++ b/cmd/context/utils.go
@@ -222,11 +222,12 @@ func LoadContextFromPath(ctx context.Context, namespace, k8sContext, path string
 	}
 	ctxResource, err := model.GetContextResource(path)
 	if err != nil {
-		// ErrOktetoManifestNotFound belongs to the discovery module and is not the same as ErrManifestNotFound error set at okteto custom errors module
 		// model.GetContextResource will only return discovery.ErrOktetoManifestNotFound in the event that no manifest is found
 		if !errors.Is(err, discovery.ErrOktetoManifestNotFound) {
 			return err
 		}
+		// here ctxResource is nil -> so we get a panic when the func
+		ctxResource = &model.ContextResource{}
 	}
 
 	if err := ctxResource.UpdateNamespace(namespace); err != nil {
--- a/cmd/context/utils.go
+++ b/cmd/context/utils.go
@@ -222,12 +222,7 @@ func LoadContextFromPath(ctx context.Context, namespace, k8sContext, path string
 	}
 	ctxResource, err := model.GetContextResource(path)
 	if err != nil {
-		// model.GetContextResource will only return discovery.ErrOktetoManifestNotFound in the event that no manifest is found
-		if !errors.Is(err, discovery.ErrOktetoManifestNotFound) {
-			return err
-		}
-		// here ctxResource is nil -> so we get a panic when the func
-		ctxResource = &model.ContextResource{}
+		return err
 	}
 
 	if err := ctxResource.UpdateNamespace(namespace); err != nil {
--- a/cmd/context/utils.go
+++ b/cmd/context/utils.go
@@ -222,7 +222,9 @@ func LoadContextFromPath(ctx context.Context, namespace, k8sContext, path string
 	}
 	ctxResource, err := model.GetContextResource(path)
 	if err != nil {
-		if !errors.Is(err, oktetoErrors.ErrManifestNotFound) {
+		// ErrOktetoManifestNotFound belongs to the discovery module and is not the same as ErrManifestNotFound error set at okteto custom errors module
+		// model.GetContextResource will only return discovery.ErrOktetoManifestNotFound in the event that no manifest is found
+		if !errors.Is(err, discovery.ErrOktetoManifestNotFound) {
 			return err
 		}
 	}
--- a/cmd/destroy/destroy.go
+++ b/cmd/destroy/destroy.go
@@ -121,7 +121,12 @@ func Destroy(ctx context.Context) *cobra.Command {
 				options.ManifestPath = uptManifestPath
 			}
 			if err := contextCMD.LoadContextFromPath(ctx, options.Namespace, options.K8sContext, options.ManifestPath); err != nil {
-				return err
+				if err.Error() == fmt.Errorf(oktetoErrors.ErrNotLogged, okteto.CloudURL).Error() {
+					return err
+				}
+				if err := contextCMD.NewContextCommand().Run(ctx, &contextCMD.ContextOptions{Namespace: options.Namespace}); err != nil {
+					return err
+				}
 			}
 
 			cwd, err := os.Getwd()
--- a/cmd/up/up.go
+++ b/cmd/up/up.go
@@ -32,7 +32,6 @@ import (
 	"github.com/okteto/okteto/cmd/utils"
 	"github.com/okteto/okteto/cmd/utils/executor"
 	"github.com/okteto/okteto/pkg/analytics"
-	"github.com/okteto/okteto/pkg/discovery"
 
 	"github.com/okteto/okteto/pkg/cmd/pipeline"
 	"github.com/okteto/okteto/pkg/config"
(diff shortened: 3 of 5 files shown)


cmd/context/utils.go
```diff
@@ -222,7 +222,9 @@ func LoadContextFromPath(ctx context.Context, namespace, k8sContext, path string
 	}
 	ctxResource, err := model.GetContextResource(path)
 	if err != nil {
-		if !errors.Is(err, oktetoErrors.ErrManifestNotFound) {
+		// ErrOktetoManifestNotFound belongs to the discovery module and is not the same as ErrManifestNotFound error set at okteto custom errors module
+		// model.GetContextResource will only return discovery.ErrOktetoManifestNotFound in the event that no manifest is found
+		if !errors.Is(err, discovery.ErrOktetoManifestNotFound) {
```
teresaromero, line 227: `model.GetContextResource` is only returning discovery errors, so i think here the exception should check the discovery error instead of the oktetoErrors one 
WDYT ??

cmd/up/up.go
```diff
@@ -130,7 +131,7 @@ func Up() *cobra.Command {
 					return err
 				}
 
-				if !errors.Is(err, oktetoErrors.ErrManifestNotFound) {
+				if !errors.Is(err, oktetoErrors.ErrManifestNotFound) && !errors.Is(err, discovery.ErrOktetoManifestNotFound) {
```
teresaromero, line 134: Both errors can come from the LoadManifestWithContext func

pkg/discovery/contextresource.go
```diff
@@ -38,5 +38,11 @@ func GetContextResourcePath(wd string) (string, error) {
 		return composeManifestPath, nil
 	}
 
+	k8sManifestPath, err := GetK8sManifestPath(wd)
+	if err == nil {
+		oktetoLog.Infof("context will load from %s", k8sManifestPath)
+		return k8sManifestPath, nil
+	}
+	// ErrOktetoManifestNotFound belongs to the discovery module and is not the same as ErrManifestNotFound error set at okteto custom errors module
```
teresaromero, line 46: Added this comment because it was not obvious for a moment. I got confused thinking this was the error at oktetoErros. Eventhough it should have the oktetoErrors module if it was, it was confusing to have both errors in my mind :/

ifbyol: Not sure about this, according to the function comment, it says `returns the file that will load the context resource from`. What is the context resources and how we get it from the k8s file? 

We also have `GetHelmChartPath`, I guess it doesn't work in that case either. The purpose of having a discoverability package was to have a single place where we load everything but it seems we are loading manifests from 2 different places. I think this will give us more headaches.

I don't think we have to load the k8s manifest just for the context. I think we have to do deeper research on this. If I understood correctly your comment in the PR description, the deploy works because, although we get this same error, we are finding the file in another way

cmd/context/utils.go
```diff
@@ -222,9 +222,12 @@ func LoadContextFromPath(ctx context.Context, namespace, k8sContext, path string
 	}
 	ctxResource, err := model.GetContextResource(path)
 	if err != nil {
-		if !errors.Is(err, oktetoErrors.ErrManifestNotFound) {
+		// model.GetContextResource will only return discovery.ErrOktetoManifestNotFound in the event that no manifest is found
+		if !errors.Is(err, discovery.ErrOktetoManifestNotFound) {
 			return err
 		}
+		// here ctxResource is nil -> so we get a panic when the func
+		ctxResource = &model.ContextResource{}
```
teresaromero, line 230: @ifbyol when reverting the GetContextResource, now i get the no manifest found error, which is the exception, then ctxResource is nil, and i get a nil pointer at line 233 (inside UpdateNamespace) because ctxResource is nil :/ is this a good fix? i dont feel like  it.. i thought that maybe in the case that no manifest found, GetContextResource will return empty ctxResource?

ifbyol: I think it should behave the same way that the deploy, it is almost the same thing. I'll try to get some time to take a look to this code carefully and understand why it fails

teresaromero: I've gone around this.... updated the PR, honestly i think this should return err in case of err... is "loading the context from manifest" if no manifest, no context info, so we init with the defaults or flag values as we do in the deploy and now destroy...
Updated code

cmd/context/utils.go
```diff
@@ -222,9 +222,7 @@ func LoadContextFromPath(ctx context.Context, namespace, k8sContext, path string
 	}
 	ctxResource, err := model.GetContextResource(path)
 	if err != nil {
-		if !errors.Is(err, oktetoErrors.ErrManifestNotFound) {
-			return err
-		}
+		return err
```
ifbyol, line 225: I think the logic within this function was not right. I'll try to explain myself. According to the comment, this function `initializes the okteto context taking into account command flags and manifest namespace/context fields`.

But, if a manifest were not found, we were throwing a panic because we were trying to set the namespace and context for a nil object.

Now, with this change, we are not considering never the namespace and context specified in the flag commands (the one we receive as parameter) if the manifest does not exist, which I think it is not correct.

I think we should create a context resource with the values from the flag, then, retrieve the context from the manifest and merge them.

teresaromero: > Now, with this change, we are not considering never the namespace and context specified in the flag commands (the one we receive as parameter) if the manifest does not exist, which I think it is not correct.

i think this function is just for context from manifest files, and additionaly (which i think should not be part of this func) we override with flag values. we do consider the flag values when error, now that we added the new context command after the error. when file and flags values will be correct, when err because no file, flags will be taken in consideration also

cmd/destroy/destroy.go
```diff
@@ -121,7 +121,12 @@ func Destroy(ctx context.Context) *cobra.Command {
...
 			}
 			if err := contextCMD.LoadContextFromPath(ctx, options.Namespace, options.K8sContext, options.ManifestPath); err != nil {
-				return err
+				if err.Error() == fmt.Errorf(oktetoErrors.ErrNotLogged, okteto.CloudURL).Error() {
+					return err
+				}
+				if err := contextCMD.NewContextCommand().Run(ctx, &contextCMD.ContextOptions{Namespace: options.Namespace}); err != nil {
+					return err
+				}
 			}
```
ifbyol, line 130: Just for the record and other possible reviewers. This logic is already in the `deploy` command. That is the reason why the `deploy` is working but the `destroy` doesn't. This will probably need to be reviewed carefully and refactored, but enough to fix the destroy

ifbyol (approved):

PR #2978, opened 2022-08-02
fix: use a different random seed generator

### LINKED ISSUE #2973: Flaky integration tests
**Describe the bug**
Currently, integration tests are flaky and they fail from time to time. Usually, retrying them once make them pass. This is frustrating because when you merge something and see the result you don't know if it was due to your changes or temporal issue.

The last failed executions, failure happened in different tests. Last 3 were:

`TestUpGlobalForwarding`
`TestUpDeploymentV1`
`TestUpStatefulsetV1`

All of them failed with a message similar to this one (just change the port number): `Couldn't connect to your development container: local port 44491 is already in-use in your local machine: port is already allocated`

**To Reproduce**
N/A

**Expected behavior**
N/A

jLopezbarb:

Fixes #2973 

## Proposed changes
- Use a different random seed generator
- Fallback to the previous random seed generator in case something goes wrong with the new one

### DIFF
--- a/main.go
+++ b/main.go
@@ -15,6 +15,8 @@ package main
 
 import (
 	"context"
+	cryptoRand "crypto/rand"
+	"encoding/binary"
 	"fmt"
 	"math/rand"
 	"os"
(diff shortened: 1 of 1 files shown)


AgustinRamiroDiaz:
So the problem was that 2 tests were ran in the exact same nanosecond?

jLopezbarb:
mostly yes, since we are running `t.Parallel()` both commands were generating the same seed. Getting the time in nanoseconds is not as precise as it suggests

PR #2979, opened 2022-08-02
fix: check if resources been destroyed after destroy command

jLopezbarb:

This is a follow-up on the tests regression fixed by: https://github.com/okteto/okteto/pull/2972

## Proposed changes
- Fix tests
- Add coverage for destroy command finish correctly

### DIFF
--- a/integration/commands/deploy.go
+++ b/integration/commands/deploy.go
@@ -43,6 +43,7 @@ type DestroyOptions struct {
 	Namespace    string
 	OktetoHome   string
 	Token        string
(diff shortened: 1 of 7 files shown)


ifbyol (approved):
Have you checked that without modifying the `destroy` function the tests are failing?

jLopezbarb:
yes, I had to make another commit because I forgot that the test which tests a pipeline inside a pipeline because we were only destroying one and there were still some resources in the namespace

PR #2992, opened 2022-08-10
Handle bad equal quantity

AgustinRamiroDiaz:
## Proposed changes

- Handles bad assignment of envvars with multiple `=`

### DIFF
--- a/cmd/deploy/validate.go
+++ b/cmd/deploy/validate.go
@@ -15,35 +15,37 @@ package deploy
 
 import (
 	"fmt"
-	"os"
 	"strings"
 )
 
-func validateAndSet(variables []string, setEnv func(key, value string) error) error {
-
-	if err := validateOptionVars(variables); err != nil {
-		return err
-	}
-	return setOptionVarsAsEnvs(variables, setEnv)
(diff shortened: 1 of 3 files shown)


cmd/deploy/validate.go
```diff
@@ -21,7 +21,7 @@ import (
 
 func validateOptionVars(variables []string) error {
 	for _, v := range variables {
-		kv := strings.SplitN(v, "=", 2)
```
ifbyol, line 24: I don't think we should do this. What if the value of the variable is a base 64 encoded valued with an '=' like `"STR='dGVzdA=='"`. As it is now it will correctly match variable `STR` variable value `dGVzdA==`.

With your changes, it would return an array of 4 elements and it would fail. Base64 encoded values are typical and they contains ' ='. 

We should just consider the value everything that goes after the first `=`.

I think it might be a good idea to include a test with the example I'm given above

AgustinRamiroDiaz:
@ifbyol @jpf-okteto Thanks for pointing out that envvars can have `=`. I've refactored the code a bit and addressed your suggestions!

PR #2994, opened 2022-08-11
Fix race conditions at unit tests

### LINKED ISSUE #2703: Fix and Add Race Detector to Unittest Runs
When using the data race detector https://go.dev/doc/articles/race_detector there are several issues detected. Ideally we'd find either issues with our tests or the code under test and fix these.

If we are unable to fix these then we should go ahead and exclude particular tests. Once all issues are addressed we should turn the race detector on for all the following builds.

```
go test -race -coverprofile=coverage.txt -covermode=atomic ./...
```

teresaromero:
Fixes #2703 

## Proposed changes

- updated spinner dependency and refactored implementation to log pkg
- use mutex at spinner and IsOktetoSample to fix the race conditions
- added -race flag to test cmd at makefile so ci runs with -race also


pkg/log/spinner.go
```diff
@@ -0,0 +1,111 @@
...
+
+type spinnerLogger struct {
+	sp             *sp.Spinner
+	spinnerSupport bool
+	onHold         bool
+}
+
+// holdSpinner is used within the TTYWritter to pause the spinner to display the log
+// if the spinner is Active (running) it will stop
+func holdSpinner() {
```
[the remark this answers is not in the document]
teresaromero: > I think it'd be great to stear away from this functions referring to global """singletons"""

What do you mean by `global """singletons"""` ?

pkg/log/tty.go
```diff
@@ -100,57 +100,75 @@ func (*TTYWriter) Fatalf(format string, args ...interface{}) {
...
 func (w *TTYWriter) Success(format string, args ...interface{}) {
 	log.out.Infof(format, args...)
+	holdSpinner()
 	w.Fprintf(w.out.Out, "%s %s\n", coloredSuccessSymbol, greenString(format, args...))
+	unholdSpinner()
 }
 
 // Information prints a message with the information symbol first, and the text in blue
 func (w *TTYWriter) Information(format string, args ...interface{}) {
 	log.out.Infof(format, args...)
```
AgustinRamiroDiaz, line 126: Checking out https://github.com/okteto/okteto/blob/0950b15bc123fe1a6c5567f333c50053abb93855/pkg/log/format.go#L62 , I see that we should have `log.out` set inside `w.out`. I know that might be out of scope, so feel free to create another issue for that. And I think we might need to do something similar with the `spinner`

pkg/log/spinner.go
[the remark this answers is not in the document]
AgustinRamiroDiaz: By `global singleton` I meant that we have one instance of that global variable that we are referring around the different functions.

By `as reciever` I meant method of the struct

My point is to try to evade using the global `log` variable, and I think this is an easy win without a refactor

pkg/cmd/stack/destroy.go
```diff
@@ -148,18 +147,16 @@ func destroyDeployments(ctx context.Context, spinner *utils.Spinner, s *model.St
 		if err := services.Destroy(ctx, dList[i].Name, dList[i].Namespace, c); err != nil {
 			return fmt.Errorf("error destroying service '%s': %s", dList[i].Name, err)
 		}
-		spinner.Stop()
 		if _, ok := s.Services[dList[i].Name]; ok {
 			oktetoLog.Success("Destroyed previous service '%s'", dList[i].Name)
 		} else {
 			oktetoLog.Success("Service '%s' destroyed", dList[i].Name)
 		}
-		spinner.Start()
```
AgustinRamiroDiaz, line 157: Why are we removing the spinner in some functions and not in other?

cmd/utils/git.go
```diff
@@ -27,7 +28,10 @@ import (
 	"k8s.io/utils/pointer"
 )
 
-var isOktetoSample *bool
+var (
+	isOktetoSample *bool
+	mu             sync.Mutex
```
AgustinRamiroDiaz, line 33: Can we rename it so we know it's only used by `IsOktetoRepo`? Maybe `IsOktetoRepoMutex` or `IsOktetoRepoLock`

cmd/up/dependencies.go
```diff
@@ -43,15 +43,14 @@ func downloadSyncthing() error {
...
+		oktetoLog.StartSpinner()
+		defer oktetoLog.StopSpinner()
 
 		if err := ssh.GenerateKeys(); err != nil {
-			spinner.Stop()
 			return err
 		}
 
-		spinner.Stop()
 		oktetoLog.Success("Client certificates generated")
```
AgustinRamiroDiaz, line 54: With the change, this message could have an iteration of the spinner afterwars. Is that a problem?

cmd/deploy/wait.go
```diff
@@ -72,7 +72,9 @@ func (dc *DeployCommand) waitForResourcesToBeRunning(ctx context.Context, opts *
 				return err
 			}
 			areAllRunning := true
-			for _, d := range dList {
+			ds := make([]v1.Deployment, len(dList))
+			for i := range ds {
```
AgustinRamiroDiaz, line 76: Why is this change? We are discarding what's in `dList`, and looping over an empty slice

cmd/deploy/wait.go
```diff
@@ -84,8 +86,10 @@ func (dc *DeployCommand) waitForResourcesToBeRunning(ctx context.Context, opts *
 			if err != nil {
 				return err
 			}
-			for _, sfs := range sfsList {
-				if !statefulsets.IsRunning(ctx, opts.Manifest.Namespace, sfs.Name, c) {
+			sfss := make([]v1.StatefulSet, len(sfsList))
+			for i := range sfss {
```
AgustinRamiroDiaz, line 90: Same as above

cmd/deploy/wait.go
teresaromero: my bad, this was a fix from a deepsource detection. fixed now

cmd/deploy/wait.go
teresaromero: my bad, this was a fix from a deepsource detection. fixed now

cmd/up/dependencies.go
teresaromero: When having a spinner running and displying TTY logs (in this case a Success Log) we needed to Stop and Start the spinner after. this logic has been moved to the log writter, where we have the hold and unhold functions to perform this actions within the log. This have been removed on every Stop/Start that was made to show the logs.

pkg/cmd/stack/destroy.go
teresaromero: same here https://github.com/okteto/okteto/pull/2994#discussion_r944259505

pkg/log/spinner.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+}
+
+// StopSpinner deletes FinalMSG and stops the running of the spinner
+func StopSpinner() {
+	if log.spinner.sp.FinalMSG != "" {
+		log.spinner.sp.Lock()
+		log.spinner.sp.FinalMSG = ""
+		log.spinner.sp.Unlock()
+	}
```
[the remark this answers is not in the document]
teresaromero: i've actually removed all the spinner mutex locks and unlocks, as the upgrade of the spinner package solved this mutex issues when interacting.

cmd/utils/git.go
teresaromero: i moved to the function

cmd/utils/git.go
```diff
@@ -94,6 +98,10 @@ func GetRandomSHA(ctx context.Context, path string) string {
 }
 
 func IsOktetoRepo() bool {
+
+	muOktetoSample.Lock()
```
AgustinRamiroDiaz, line 102: Why do we need this mutex?

teresaromero: updated with comment

maroshii: I just want to point out that although setting a mutex "fixes" the race condition there is something fundamentally wrong about how this is implemented that needs to be revisited. Namely mutating global variables.

In the meantime, I think we can probably rewrite this with sync.Once and an atomic int. Playground: https://go.dev/play/p/iMeYELn_qyM

teresaromero: thnks Fran, i've updated the code to use your feedback. I think i used once couple months ago but did not remember this when fixing the race.

jLopezbarb:
LGTM but I think we'd need at least another approval before merging it

jLopezbarb (approved):

AgustinRamiroDiaz (approved):

PR #2996, opened 2022-08-11
Add okteto context to dockerfile build args

odedlaz:
## Proposed changes

When building dockerfiles they are unaware of the environment they are being built at. 
This PR added the `OKTETO_NAMESPACE` automatically to the build arguments which can later be consumed by the `Dockerfile`.

This change is needed for a usecase I have where the `Dockerfile` should be aware of the okteto namespace its going to be deployed to.

I think this change is useful given that github actions do something similar: set default env variables to be consumed by the user.

### DIFF
--- a/pkg/cmd/build/build.go
+++ b/pkg/cmd/build/build.go
@@ -248,7 +248,7 @@ func OptsFromBuildInfo(manifestName, svcName string, b *model.BuildInfo, o *type
 
 	if registry.IsOktetoRegistry(b.Image) {
 		defaultBuildArgs := map[string]string{
-			model.OktetoUserNameEnvVar:        okteto.Context().Username,
+			model.OktetoContextEnvVar:        okteto.Context().Name,
 			model.OktetoNamespaceEnvVar:       okteto.Context().Namespace,
 		}
 
--- a/pkg/cmd/build/build.go
+++ b/pkg/cmd/build/build.go
@@ -246,6 +246,31 @@ func OptsFromBuildInfo(manifestName, svcName string, b *model.BuildInfo, o *type
 		file = extractFromContextAndDockerfile(b.Context, b.Dockerfile, svcName)
 	}
 
+	if registry.IsOktetoRegistry(b.Image) {
+		defaultBuildArgs := map[string]string{
+			model.OktetoUserNameEnvVar:        okteto.Context().Username,
+			model.OktetoNamespaceEnvVar:       okteto.Context().Namespace,
+		}
+
+		for _, e := range b.Args {
+			if _, exists := defaultBuildArgs[e.Name]; !exists {
+				continue;
+			}
+			// we don't want to replace build arguments that were already set by the user
+			delete(defaultBuildArgs, e.Name)
+		}
+
+		for key, val := range defaultBuildArgs {
+			if (val == "") {
+				continue;
+			}
+
+			b.Args = append(b.Args, model.EnvVar{
+				Name: key, Value: val,
+			})
+		}
+	}
+
 	opts := &types.BuildOptions{
 		CacheFrom: b.CacheFrom,
 		Target:    b.Target,
--- a/cmd/build/v2/build.go
+++ b/cmd/build/v2/build.go
@@ -77,7 +77,6 @@ func (*OktetoBuilder) LoadContext(ctx context.Context, options *types.BuildOptio
 
 	if options.Manifest.Namespace != "" {
 		ctxOpts.Namespace = options.Manifest.Namespace
-		ctxOpts.Namespace = options.Manifest.Namespace
 	}
 	if options.Namespace != "" {
 		ctxOpts.Namespace = options.Namespace
--- a/pkg/cmd/build/build_test.go
+++ b/pkg/cmd/build/build_test.go
@@ -69,12 +69,18 @@ func Test_validateImage(t *testing.T) {
 }
 
 func Test_OptsFromBuildInfo(t *testing.T) {
+	context := okteto.OktetoContext{
+		Namespace: "test",
+		Registry:  "registry.okteto",
+	}
+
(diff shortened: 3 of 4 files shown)


jmacelroy:
@jLopezbarb any thoughts on this one? Should we follow up with some testing improvements?

jLopezbarb:
I think there are two things to improve:
- I'd check that the user isn't already specifying the `OKTETO_NAMESPACE` env var, which can cause some friction on why the CLI is using the default one instead of the one specified by the user.
- I'd also add the `OKTETO_CONTEXT` env var, that way we could have the whole context on the build step.

For testing, we should be able to test that the env vars are injected correctly on the function `Test_OptsFromBuildInfo` from the file `pkg/cmd/build/build_test.go`

odedlaz:
> This is a great addition!! Thanks @odedlaz. Despite of your use case, there are a few more use cases that can benefit from this change.
> 
>  
> 
> I think we could improve the PR if:
> 
> - I'd check that the user isn't already specifying the `OKTETO_NAMESPACE` env var, which can cause some friction on why the CLI is using the default one instead of the one specified by the user.
> 
> - I'd also add the `OKTETO_CONTEXT` env var, that way we could have the whole context on the build step.
> 
> 
> 
> For testing, we should be able to test that the env vars are injected correctly on the function `Test_OptsFromBuildInfo` from the file `pkg/cmd/build/build_test.go` 
> 
> 

I'll take care of that and post a fix soon.

odedlaz:
@jLopezbarb , I'm done. this is pending approval of this #3016. Tell me what you think!

go.mod
```diff
@@ -203,7 +203,10 @@ require (
 	sigs.k8s.io/yaml v1.2.0 // indirect
 )
 
-require github.com/google/go-containerregistry v0.8.0
+require (
+	github.com/google/go-containerregistry v0.8.0
+	github.com/samber/lo v1.27.0
```
jmacelroy, line 208: @jLopezbarb and @odedlaz  I have a strong preference for not bringing in this dependency. It is quite large and we are only using what looks like a small portion of it. Adding a dependency can add a lot of long-term overhead and in this case I believe we can have similar functionality without the dependency.

odedlaz: I'll remove it to ease the PR approval. It does remove a lot of boilerplate code though.

I'd consider adding it in the future.

odedlaz: @jmacelroy , remove the [samber/lo](https://github.com/samber/lo) dependency.

pkg/cmd/build/build.go
```diff
@@ -246,6 +246,33 @@ func OptsFromBuildInfo(manifestName, svcName string, b *model.BuildInfo, o *type
 		file = extractFromContextAndDockerfile(b.Context, b.Dockerfile, svcName)
 	}
 
+	defaultBuildArgs := map[string]string{
+		model.OktetoContextEnvVar:         okteto.Context().Name,
+		model.OktetoUserNameEnvVar:        okteto.Context().Username,
+		model.OktetoTokenEnvVar:           okteto.Context().Token,
```
ifbyol, line 252: I'm not sure if the okteto token should be added. If it is used, an image would be built with that token inside. 

I think having only `model.OktetoContextEnvVar` and `model.OktetoNamespaceEnvVar` should be ok unless you need also the others, but I understood that you only need the namespace, right? @odedlaz

pkg/cmd/build/build.go
```diff
@@ -246,6 +246,33 @@ func OptsFromBuildInfo(manifestName, svcName string, b *model.BuildInfo, o *type
 		file = extractFromContextAndDockerfile(b.Context, b.Dockerfile, svcName)
 	}
 
+	defaultBuildArgs := map[string]string{
```
jLopezbarb, line 249: Could we only introduce these default args when the image is being built in okteto?
```suggestion
	if registry.IsOktetoRegistry(b.Image)
```
I don't think we should add those arguments if the image is not being build in the okteto registry

pkg/cmd/build/build.go
odedlaz: @ifbyol, you're correct. I was trying to go the extra mile, probably went too far :) removed everything except `OktetoContextEnvVar` and `OktetoNamespaceEnvVar`. Should be easy to extend in the future.

pkg/cmd/build/build.go
odedlaz: done.

odedlaz:
@jpf-okteto , I think CircleCI is not triggering here as well.

ifbyol (approved):
Thanks for the contribution!

ifbyol:
@odedlaz it seems some tests are failing

odedlaz:
> @odedlaz it seems some tests are failing

weird, I ran the test suite. anyway, will fix later today and ping you when done. thanks!

odedlaz:
> @odedlaz it seems some tests are failing

@ifbyol fixed.

AgustinRamiroDiaz:
@odedlaz Thanks for the contribution! 

We are reviewing the introduction of functional programming libraries and patterns as you suggested with `lo`, it's not the first time we have this discussion. I think it will come slowly with the update to go 1.18 allowing generics, and probably we'll introduce some of this modules to reduce boilerplate.

odedlaz:
> @odedlaz Thanks for the contribution! 
> 
> 
> 
> We are reviewing the introduction of functional programming libraries and patterns as you suggested with `lo`, it's not the first time we have this discussion. I think it will come slowly with the update to go 1.18 allowing generics, and probably we'll introduce some of this modules to reduce boilerplate.

Do you have an issue on that? I'd love to help out and maybe do the work myself.

AgustinRamiroDiaz:
@odedlaz it has always been discussed privately. I think it has raised many times so I created [an issue for it](https://github.com/okteto/okteto/issues/3026). Feel free to comment your opinions!

PR #2997, opened 2022-08-11
fix: if image is set, we should return the image

### LINKED ISSUE #2995: Okteto rebuilds image always without specifying `--build`
**Describe the bug**
Okteto rebuilds the image always without specifying the `--build` flag if the image tag specified by the user is not using the default okteto tag. 

**To Reproduce**

1. Create this okteto stack:
```yaml
services:
  test:
    image: okteto.dev/test:dev
    build:
      context: .
```
2. Create this okteto manifest:
```yaml
deploy:
  compose: okteto-stack.yml

dev:
  test:
    command: ["bash"]
    sync:
      - .:/app
```
3. Run `okteto up` and see that is building image
4. Run `okteto up` again and see that rebuilds the image

**Expected behavior**
Second execution should be skipped due to image being built previously

- OS: macOS
- Version 2.6

**Additional context**
Introduced in https://github.com/okteto/okteto/pull/2939

jLopezbarb:

Fixes #2995 

## Proposed changes
- Return the proper image to fix the scenario
- Add new test to cover the scenario proposed

### DIFF
--- a/cmd/build/v2/services.go
+++ b/cmd/build/v2/services.go
@@ -107,7 +107,7 @@ func getToBuildTag(manifestName, svcName string, b *model.BuildInfo) string {
 		return b.Image
 	case (shouldBuildFromDockerfile(b) && shouldAddVolumeMounts(b)) || shouldAddVolumeMounts(b):
 		return fmt.Sprintf("%s/%s-%s:%s", targetRegistry, manifestName, svcName, model.OktetoImageTagWithVolumes)
-	case shouldBuildFromDockerfile(b) && b.Image != "":
+	case b.Image != "" && shouldBuildFromDockerfile(b):
 		return b.Image
 	case shouldBuildFromDockerfile(b):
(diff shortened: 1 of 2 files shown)


teresaromero (approved):

cmd/build/v2/services.go
```diff
@@ -107,6 +107,8 @@ func getToBuildTag(manifestName, svcName string, b *model.BuildInfo) string {
 		return b.Image
 	case (shouldBuildFromDockerfile(b) && shouldAddVolumeMounts(b)) || shouldAddVolumeMounts(b):
 		return fmt.Sprintf("%s/%s-%s:%s", targetRegistry, manifestName, svcName, model.OktetoImageTagWithVolumes)
+	case shouldBuildFromDockerfile(b) && b.Image != "":
```
ifbyol, line 110: Sounds weird having to check if it has to be built from a dockerfile but then, return the Image from b. Would it make sense to switch the case clauses?

PR #2998, opened 2022-08-11
Update help message for deploy command

ifbyol:

## Proposed changes

- Update help for `deploy` command to indicate that the execution happens locally
- Removed `hidden` flag from pipeline command. I think we added it when we deprecated the command but we didn't remove the flag when we decided to keep it for a while

### DIFF
--- a/cmd/pipeline/pipeline.go
+++ b/cmd/pipeline/pipeline.go
@@ -41,10 +41,9 @@ func NewCommand() (*Command, error) {
 //Pipeline pipeline management commands
 func Pipeline(ctx context.Context) *cobra.Command {
 	cmd := &cobra.Command{
-		Use:    "pipeline",
-		Short:  "Pipeline management commands",
-		Args:   utils.NoArgsAccepted("https://www.okteto.com/docs/reference/cli/#pipeline"),
-		Hidden: true,
+		Use:   "pipeline",
+		Short: "Pipeline management commands",
+		Args:  utils.NoArgsAccepted("https://www.okteto.com/docs/reference/cli/#pipeline"),
 	}
 	cmd.AddCommand(deploy(ctx))
 	cmd.AddCommand(destroy(ctx))
--- a/cmd/deploy/deploy.go
+++ b/cmd/deploy/deploy.go
@@ -97,7 +97,7 @@ func Deploy(ctx context.Context) *cobra.Command {
 
 	cmd := &cobra.Command{
 		Use:   "deploy [service...]",
-		Short: "Execute the list of commands specified in the 'deploy' section of your okteto manifest",
+		Short: "Execute locally the list of commands specified in the 'deploy' section of your okteto manifest",
 		RunE: func(cmd *cobra.Command, args []string) error {
 
 			// validate cmd options


cmd/pipeline/pipeline.go
```diff
@@ -41,10 +41,9 @@ func NewCommand() (*Command, error) {
 //Pipeline pipeline management commands
 func Pipeline(ctx context.Context) *cobra.Command {
 	cmd := &cobra.Command{
-		Use:    "pipeline",
-		Short:  "Pipeline management commands",
-		Args:   utils.NoArgsAccepted("https://www.okteto.com/docs/reference/cli/#pipeline"),
-		Hidden: true,
```
ifbyol, line 47: @rlamana @jmacelroy @rberrelleza double checking that we want to display the `okteto pipeline` command in the CLI help as it is not deprecated

rberrelleza: Yes, I believe we don't have other way to execute a pipeline remotely today right?

rlamana: It is also listed in the docs, so I guess it makes sense to show it in the CLI. Do we have a deprecation plan for it?

AgustinRamiroDiaz: I think there was the idea of extending the `okteto deploy` command to have an option of running remotely, but nothing specified

AdrianPedriza (approved):

cmd/pipeline/pipeline.go
ifbyol: @rberrelleza  No, right now there is not other way of executing a pipeline remotely. 
@rlamana No, we don't have a deprecation plan yet. As Agustin said, we marked it as deprecated at some point in favor of `okteto deploy` but we saw it wasn't behaving in the same way, so we discarded that option. We removed the deprecation warning message some time ago but we didn't recover it in the CLI help. That's why I'm recovering it

rlamana: @ifbyol What do you mean "it wasn't behaving in the same way"?

ifbyol: > @ifbyol What do you mean "it wasn't behaving in the same way"?

@rlamana  When we deprecated `okteto pipeline deploy`, we included some logic in the `okteto deploy` command to trigger a remote execution. That logic wasn't intuitive, was creating more confusion around the deploy command and it wasn't behaving as the `okteto pipeline deploy` was. So we decided to "recover" the pipeline command until we had a better idea on how to deprecate it

rlamana: Ok, that's what I meant, we do need to deprecate it at some point. We probably didn't do that smoothly last time and that's why we encountered behaviors and a different experience as a result. Who know about these details and is it documented? I imagine Pablo, Javi and you are probably the only ones with enough context. Can we create a document listing those use cases / behaviors that `pipeline deploy` supports and `deploy` doesn't and finally create a deprecation plan?

ifbyol: I don't remember all the details. I think Pablo received some feedback from customers and we faced some issue, but don't remember it. I don't know if we collect it in some place but yes, we should document it. 

Which would be the priority of this? AFAIK we are stabilizing the CLI so this kind of changes are not expected to be done in the short term

jLopezbarb: We discussed a little bit here: https://github.com/okteto/okteto/discussions/2462

PR #3002, opened 2022-08-12
fix: add error message deploying compose

jLopezbarb:

Fixes an error displaying errors on the UI on Okteto clusters

## Proposed changes
- Log the error in the configmap while we are still on the stage
- Add tests to check that the error is not happening again


cmd/deploy/deploy.go
```diff
@@ -474,27 +474,36 @@ func (dc *DeployCommand) deploy(ctx context.Context, opts *Options) error {
...
 				return
 			}
+			oktetoLog.SetStage("")
 		}
 
 		// deploy endpoits if any
 		if opts.Manifest.Deploy.Endpoints != nil {
+			oktetoLog.SetStage("Endpoints configuration")
 			if err := dc.deployEndpoints(ctx, opts); err != nil {
+				oktetoLog.AddToBuffer(oktetoLog.ErrorLevel, "error deploying divert: %s", err.Error())
```
teresaromero, line 490: is this for Divert or for endpoints?

ifbyol, line 490:
```suggestion
				oktetoLog.AddToBuffer(oktetoLog.ErrorLevel, "error deploying endpoints: %s", err.Error())
```

ifbyol (approved):

cmd/deploy/deploy.go
rberrelleza: Endpoints are not deployed right? Maybe "error creating endpoints" or "generating endpoints " could be more accurate?

PR #3008, opened 2022-08-16
fix context checking when no manifest found

### LINKED ISSUE #2977: Improve the way we get the context when checking if a manifest exists
As part of https://github.com/okteto/okteto/issues/2968, we have been taking a look to the function we currently use to load the context from an okteto manifest and we detected some behavior that should be improved.

In the `LoadContextFromPath` function: 

```
func LoadContextFromPath(ctx context.Context, namespace, k8sContext, path string) error {
	ctxOptions := &ContextOptions{
		Namespace: namespace,
		Show:      true,
	}
	ctxResource, err := model.GetContextResource(path)
	if err != nil {
		// here discovery.ErrOktetoManifestNotFound will return as error
		if !errors.Is(err, oktetoErrors.ErrManifestNotFound) {
			return err
		}
	}

	if err := ctxResource.UpdateNamespace(namespace); err != nil {
		return err
	}

	if err := ctxResource.UpdateContext(k8sContext); err != nil {
		return err
	}

	ctxOptions = &ContextOptions{
		Context:   ctxResource.Context,
		Namespace: ctxResource.Namespace,
		Show:      true,
	}

	return NewContextCommand().Run(ctx, ctxOptions)
}
```

We are checking the wrong error. Check the comment, `oktetoErrors.ErrManifestNotFound` should be replaced by `discovery.ErrOktetoManifestNotFound`. First, we should make sure that `GetContextResource` returns the discovery one in case that the manifest is not found.

So, if the manifest is not found, we should not fail the execution of this function as we should use the namespace and context received as parameter.

Also, apart of that change, we should initialize `ctxResource` within the else if the error is not returned, if not, `ctxResource.UpdateNamespace` will generate a panic as `model.GetContextResource(path)` returns it as nil if there was an error

AdrianPedriza:

Fixes #2977

## Proposed changes

- fix error to check to `discovery.ErrOktetoManifestNotFound`
- initialize context when err == `discovery.ErrOktetoManifestNotFound`


jLopezbarb:
Could we unify `oktetoErrors.ErrManifestNotFound` and `discovery.ErrManifestNotFound`? I think it can be error-prone to have both

ifbyol:
> Could we unify `oktetoErrors.ErrManifestNotFound` and `discovery.ErrManifestNotFound`? I think it can be error-prone to have both

Agree with unifying both errors if they are used to represent the same. Also, careful with the error handling if we do the change

cmd/context/utils.go
```diff
@@ -156,7 +156,7 @@ func LoadManifestWithContext(ctx context.Context, opts ManifestOptions) (*model.
 
 	manifest, err := model.GetManifestV1(opts.Filename)
 	if err != nil {
-		if !errors.Is(err, oktetoErrors.ErrManifestNotFound) && !errors.Is(err, discovery.ErrOktetoManifestNotFound) {
+		if !errors.Is(err, discovery.ErrOktetoManifestNotFound) && !errors.Is(err, discovery.ErrOktetoManifestNotFound) {
```
jLopezbarb, line 159: Same error on both comparisons

pkg/discovery/contextresource_test.go
```diff
@@ -69,8 +76,12 @@ func TestGetContextResourcePathWhenExists(t *testing.T) {
 				defer f.Close()
 			}
 			result, err := GetContextResourcePath(wd)
-			assert.NoError(t, err)
-			assert.Equal(t, filepath.Join(wd, tt.expected), result)
+			if tt.expectErr != nil {
```
jLopezbarb, line 79: I'd either create another function to test the error, or assert that the errors are the same without having an if else clause:
`assert.ErrorIs(t, tt.expectedErr, err)`

ifbyol:
Can we run the E2E before merging it?

teresaromero:
i remember going thought this with https://github.com/okteto/okteto/pull/2969 can we check this case is covered with this change?

cmd/context/utils.go
```diff
@@ -218,10 +218,11 @@ func LoadStackWithContext(ctx context.Context, name, namespace string, stackPath
 func LoadContextFromPath(ctx context.Context, namespace, k8sContext, path string) error {
 	ctxResource, err := model.GetContextResource(path)
 	if err != nil {
-		// here discovery.ErrOktetoManifestNotFound will return as error
-		if !errors.Is(err, oktetoErrors.ErrManifestNotFound) {
+		if !errors.Is(err, discovery.ErrOktetoManifestNotFound) {
 			return err
 		}
+
+		ctxResource = &model.ContextResource{}
```
jLopezbarb, line 225: Could we add a test that covers this use case?

cmd/manifest/init-v2.go
```diff
@@ -519,10 +517,20 @@ func inferDeploySection(cwd string) (*model.DeployInfo, error) {
 }
 
 func inferDevsSection(cwd string) (model.ManifestDevs, error) {
-	files, err := os.ReadDir(cwd)
+	dirEntries, err := os.ReadDir(cwd)
 	if err != nil {
 		return nil, err
 	}
+
+	files := make([]fs.FileInfo, 0, len(dirEntries))
```
jLopezbarb, line 525: Why is this needed?

AdrianPedriza: due to a deepasource complain

jLopezbarb: Can we extract it into a function then?

AdrianPedriza: reverted due to issue when pulling last updates

ifbyol (approved):

PR #3011, opened 2022-08-16
Use injected logrus when possible

AgustinRamiroDiaz:
## Proposed changes

- Use the injected logrus instead of referencing the `log` global variable when possible

### DIFF
--- a/pkg/log/json.go
+++ b/pkg/log/json.go
@@ -75,55 +75,55 @@ func newJSONWriter(out *logrus.Logger, file *logrus.Entry) *JSONWriter {
 // Debug writes a debug-level log
 func (w *JSONWriter) Debug(args ...interface{}) {
 	w.out.Debug(args...)
-	if log.file != nil {
-		log.file.Debug(args...)
+	if w.file != nil {
+		w.file.Debug(args...)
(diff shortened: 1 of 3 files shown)


AgustinRamiroDiaz:
Related to https://github.com/okteto/okteto/pull/3011

ifbyol:
> Related to #3011

The link points to the PR itself.

What is the motivation of this change? Logger is an important part that is not easy to test and may even impact on how things are being displayed in the UI. Has this generated any issue? Checking the code, I cannot say if `w.out` is always initialized. If that is the case, some of this code could generate a panic at some point and it wouldn't be easy to trace.

As the real problem is having `log` as a global variable. I think this should be handled in its own task all at once. Removing `log` from a global variable and changing the way the logger is initialized and used across all the CLI

AgustinRamiroDiaz:
@ifbyol I've fixed the url.

Yes, the motivation is about removing the `log` global variable. 
You are right that it's a big risk to do this only for a fraction of the change that wouldn't solve the issue

I'll close this in favor of doing it all together

ifbyol:
thanks Agustin!

PR #3013, opened 2022-08-16
fix: backoff limit message

### LINKED ISSUE #2810: depends on should not fail if backofflimit > maxattempts
**Describe the bug**
My okteto deploy fails before it should

**To Reproduce**

1. Deploy this compose by running `okteto deploy`:
``` docker_compose.yaml
services:
  a:
    image: alpine
    command: asa
    deploy:
      restart_policy:
        condition: on-failure
        max_attempts: 5
  b:
    image: alpine
    depends_on:
      - a
```
2. See that it fails on restart 3

**Expected behavior**
If the dependant service has a backofflimit we should wait until that premise is completed

- OS: macOS
- Version 2.4

jLopezbarb:

Fixes #2810

## Proposed changes
- Backoff limit now takes in count the bakoff limit of the dependent svc
- Add tests

### DIFF
--- a/pkg/cmd/stack/deploy.go
+++ b/pkg/cmd/stack/deploy.go
@@ -352,7 +352,7 @@ func deployServices(ctx context.Context, stack *model.Stack, k8sClient kubernete
 								return fmt.Errorf("service '%s' has failed his healthcheck probes: %s", key, value)
 							}
 						}
-						if err := getFailedStatusError(ctx, stack, svcName, k8sClient, config); err != nil {
+						if err := getFailedStatusError(ctx, stack, svcName, k8sClient); err != nil {
 							return err
 						}
 						continue
@@ -428,7 +428,7 @@ func getServicesWithFailedProbes(ctx context.Context, stack *model.Stack, svcNam
 	return failedServices
 }
 
-func getFailedStatusError(ctx context.Context, stack *model.Stack, svcName string, client kubernetes.Interface, config *rest.Config) error {
+func getFailedStatusError(ctx context.Context, stack *model.Stack, svcName string, client kubernetes.Interface) error {
 	svc := stack.Services[svcName]
 	for dependingSvc := range svc.DependsOn {
 		svcLabels := map[string]string{model.StackNameLabel: stack.Name, model.StackServiceNameLabel: dependingSvc}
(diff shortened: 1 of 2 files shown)


pkg/cmd/stack/deploy.go
```diff
@@ -441,7 +441,11 @@ func getFailedStatusError(ctx context.Context, stack *model.Stack, svcName strin
 		for _, cStatus := range p.Status.ContainerStatuses {
 			totalRestarts += int(cStatus.RestartCount)
 		}
-		if totalRestarts >= maxRestartsToConsiderFailed {
+		maxSvcRestarts := stack.Services[dependingSvc].BackOffLimit
+		if maxSvcRestarts == 0 {
+			maxSvcRestarts = maxRestartsToConsiderFailed
```
teresaromero, line 446: `maxRestartsToConsiderFailed` is the default max restarts to fail?

pkg/cmd/stack/deploy.go
```diff
@@ -441,7 +441,11 @@ func getFailedStatusError(ctx context.Context, stack *model.Stack, svcName strin
...
+		}
+		if int32(totalRestarts) >= maxSvcRestarts {
```
teresaromero, line 448: if maxSvcRestarts is not int32 ? will this be compared?

pkg/cmd/stack/deploy.go
jLopezbarb: yes, It's set to 3 by default on kubernetes resources

pkg/cmd/stack/deploy.go
jLopezbarb: We should either cast `totalRestarts` to `int32` or `maxSvcRestarts` to `int`. I casted `totalRestarts` because it's more explicit, but since both of them are integers they will get compared

pkg/cmd/stack/deploy.go
```diff
@@ -428,7 +428,7 @@ func getServicesWithFailedProbes(ctx context.Context, stack *model.Stack, svcNam
 	return failedServices
 }
 
-func getFailedStatusError(ctx context.Context, stack *model.Stack, svcName string, client kubernetes.Interface, config *rest.Config) error {
+func getFailedStatusError(ctx context.Context, stack *model.Stack, svcName string, client kubernetes.Interface) error {
```
AdrianPedriza, line 431: maybe we could change this func name to something like `getErrorDueToRestartLimit` or something like that? Just a suggestion to add more context about what kind of error will be checking

AdrianPedriza (approved):

teresaromero (approved):

PR #3018, opened 2022-08-19
Fix prompt dev selector

### LINKED ISSUE #2778: typo in prompt when running okteto down
**Describe the bug**
If you run `okteto down` the prompt lists all dev containers but it says to select the container you want to "activate".

**To Reproduce**

1. Start a container
2. Run `okteto down`

**Expected behavior**
The prompt should say select the dev container you want to "deactivate". Another UX improvement could be if it lists down **only** the active containers instead of listing all containers.

- OS: macOS
- Version: Okteto CLI 2.3.3

**Additional context**
Screenshot 2022-06-10 at 12 17 49 PM

teresaromero:
Fixes #2778 

## Proposed changes

- Development container selector was displaying "activate" action on every command, this was wrong as this selector appears at: `down`, `doctor`, `exec`, `push`, `restart`, `status`. Refactored `GetDevFromManifest` and separate the logic between get the dev from manifest and asking for the user to select it by creating new function `SelectDevFromManifest`.
- Refactored OktetoSelector and created NewOktetoSelector to create a new selector with the labels and options for each case. 
- Refactor of SelectorItem: this struct was having fields just used for the context selector in order to return if the selected context was Okteto Context. Moved this logic to the context logic and kept the SelectorItem standard for all the different selectors.
- selector logic now has its own pkg `prompt` which also handles any other input from the user (questions, yes/no questions, etc)
- some format changes have also come up and i've added them to this PR (spaces to func comments)

### DIFF
--- a/cmd/doctor.go
+++ b/cmd/doctor.go
@@ -70,7 +70,8 @@ func Doctor() *cobra.Command {
 					return err
 				}
 
-				selector := utils.NewOktetoSelector("Select the development container you want to generate zip with logs:", "Development container")
+				selectorItems := utils.GetItemsForDevSelector(manifest.Dev)
+				selector := utils.NewOktetoSelector("Select the development container you want to generate zip with logs:", selectorItems, "Development container")
 				dev, err = utils.SelectDevFromManifest(manifest, selector)
 				if err != nil {
 					return err
--- a/cmd/doctor.go
+++ b/cmd/doctor.go
@@ -15,6 +15,7 @@ package cmd
 
 import (
 	"context"
+	"errors"
 
 	contextCMD "github.com/okteto/okteto/cmd/context"
 	"github.com/okteto/okteto/cmd/utils"
@@ -65,7 +66,15 @@ func Doctor() *cobra.Command {
 			}
 			dev, err := utils.GetDevFromManifest(manifest, devName)
 			if err != nil {
-				return err
+				if !errors.Is(err, oktetoErrors.ErrNoDevSelected) {
+					return err
+				}
+
+				selector := utils.NewOktetoSelector("Select the development container you want to generate zip with logs:", "Development container")
+				dev, err = utils.SelectDevFromManifest(manifest, selector)
+				if err != nil {
+					return err
+				}
 			}
 			filename, err := doctor.Run(ctx, dev, doctorOpts.DevPath, c)
 			if err == nil {
--- a/cmd/doctor.go
+++ b/cmd/doctor.go
@@ -66,7 +66,7 @@ func Doctor() *cobra.Command {
 			}
 			dev, err := utils.GetDevFromManifest(manifest, devName)
 			if err != nil {
-				if !errors.Is(err, oktetoErrors.ErrNoDevSelected) {
+				if !errors.Is(err, utils.ErrNoDevSelected) {
 					return err
 				}
 
--- a/cmd/utils/dev.go
+++ b/cmd/utils/dev.go
@@ -190,12 +190,13 @@ func sortOptions(in []string) {
 	})
 }
 
-func SelectDevFromManifest(manifest *model.Manifest, selector *OktetoSelector) (*model.Dev, error) {
+func getItemsForDevSelector(devs model.ManifestDevs) []SelectorItem {
 	devNames := []string{}
-	for name := range manifest.Dev {
+	for name := range devs {
 		devNames = append(devNames, name)
 	}
 
+	// items get sorted alfabetically
 	sortOptions(devNames)
 
 	items := []SelectorItem{}
@@ -206,8 +207,12 @@ func SelectDevFromManifest(manifest *model.Manifest, selector *OktetoSelector) (
 			Enable: true,
 		})
 	}
-	selector.Items = items
-	selector.Size = len(items)
+	return items
+}
+
+func SelectDevFromManifest(manifest *model.Manifest, selector OktetoSelectorInterface) (*model.Dev, error) {
+	items := getItemsForDevSelector(manifest.Dev)
+	selector.SetOptions(items)
 	devName, _, err := selector.Ask(context.Background())
 	if err != nil {
 		return nil, err
--- a/cmd/utils/dev.go
+++ b/cmd/utils/dev.go
@@ -211,9 +211,9 @@ func getItemsForDevSelector(devs model.ManifestDevs) []SelectorItem {
 }
 
 func SelectDevFromManifest(manifest *model.Manifest, selector OktetoSelectorInterface) (*model.Dev, error) {
-	items := getItemsForDevSelector(manifest.Dev)
-	selector.SetOptions(items)
-	devName, _, err := selector.Ask(context.Background())
+	selector.SetOptions(getItemsForDevSelector(manifest.Dev))
+
+	devName, err := selector.Ask(context.Background())
 	if err != nil {
 		return nil, err
 	}
--- a/cmd/utils/dev.go
+++ b/cmd/utils/dev.go
@@ -165,11 +165,7 @@ func GetDevFromManifest(manifest *model.Manifest, devName string) (*model.Dev, e
 	}
 
 	if devName == "" {
-		selectedDev, err := selectDevFromManifest(manifest, "Select the development container you want to activate:")
-		if err != nil {
-			return nil, err
-		}
-		return selectedDev, nil
+		return nil, oktetoErrors.ErrNoDevSelected
 	}
 
 	// iterate though options to look fot devName and return
@@ -194,7 +190,7 @@ func sortOptions(in []string) {
 	})
 }
 
-func selectDevFromManifest(manifest *model.Manifest, selectionPrompt string) (*model.Dev, error) {
+func SelectDevFromManifest(manifest *model.Manifest, selector *OktetoSelector) (*model.Dev, error) {
 	devNames := []string{}
 	for name := range manifest.Dev {
 		devNames = append(devNames, name)
@@ -210,8 +206,9 @@ func selectDevFromManifest(manifest *model.Manifest, selectionPrompt string) (*m
 			Enable: true,
 		})
 	}
-
-	devName, _, err := AskForOptionsOkteto(context.Background(), items, selectionPrompt, "Development container")
+	selector.Items = items
+	selector.Size = len(items)
+	devName, _, err := selector.Ask(context.Background())
 	if err != nil {
 		return nil, err
 	}
--- a/cmd/utils/dev.go
+++ b/cmd/utils/dev.go
@@ -15,6 +15,7 @@ package utils
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net/url"
 	"os"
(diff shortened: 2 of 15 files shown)


pkg/errors/errors.go
```diff
@@ -181,6 +181,9 @@ var (
 
 	// ErrCouldNotInferAnyManifest is raised when we can't detect any manifest to load
 	ErrCouldNotInferAnyManifest = errors.New("couldn't detect any manifest (okteto manifest, pipeline, compose, helm chart, k8s manifest)")
+
+	// ErrNoDevSelected is raised when no development environment is selected
+	ErrNoDevSelected = errors.New("No Development Environment selected")
```
jLopezbarb, line 186: Could we move this error to the function file?

cmd/utils/selector.go
```diff
@@ -42,6 +42,10 @@ const (
 // ErrInvalidOption is raised when the selector has an invalid option
```
jLopezbarb, line 42: NIT: I think we should move this file into another package, but don't know where it should fit. Since we are storing every io under `pkg/log` I think it could potentially fit there

cmd/doctor.go
```diff
@@ -65,7 +66,16 @@ func Doctor() *cobra.Command {
 			}
 			dev, err := utils.GetDevFromManifest(manifest, devName)
 			if err != nil {
-				return err
+				if !errors.Is(err, oktetoErrors.ErrNoDevSelected) {
```
jLopezbarb, line 69: We should probably also add `oktetoErrors.ErrManifestNoDevSection` to this if

teresaromero: this if controls that if there is no dev selected, we show the options to select the dev to the user. If there is no dev section, the error will be returned and showed to the user. Do we want to prompt the init so the dev is created at the manifest?

jLopezbarb: No, it's ok as it is. I read wrong the code and thought it was displaying the selector when the dev didn't exist sorry 😅

pkg/prompt/selector.go
```diff
@@ -0,0 +1,429 @@
+package prompt
```
ifbyol, line 1: Was there any effective change in this file compared to the original one? it is hard to see as the diff considers it as a new file.

pkg/prompt/selector.go
```diff
@@ -0,0 +1,429 @@
...
+}
+
+// selectorItem represents a selectable item on a selector
+type selectorItem struct {
+	Name   string
+	Label  string
+	Enable bool
+}
+
+func NewOktetoSelector(label string, options []string, selectedTpl string, preselected int) *OktetoSelector {
```
ifbyol, line 50: Can you include a comment in the exported function please?

pkg/prompt/selector.go
```diff
@@ -0,0 +1,429 @@
...
+	oktetoLog "github.com/okteto/okteto/pkg/log"
+	"golang.org/x/term"
+)
+
+const (
+	esc        = "\033["
+	showCursor = esc + "?25h"
+)
+
+type OktetoSelectorInterface interface {
```
ifbyol, line 26: Can you include a comment in the exported interface please?

pkg/model/manifest.go
```diff
@@ -1135,3 +1135,20 @@ func (d ManifestDevs) HasDev(name string) bool {
 	_, ok := d[name]
 	return ok
 }
+
+func (d ManifestDevs) SortedNameList() []string {
```
ifbyol, line 1139: Same about comment in exported functions

cmd/utils/selector.go
```diff
@@ -1,514 +0,0 @@
...
-	Registry  string
-}
-
-//AskForOptionsOkteto given some options ask the user to select one
-func AskForOptionsOkteto(ctx context.Context, options []SelectorItem, label, selectedTpl string) (string, bool, error) {
-	selectedTemplate := getSelectedTemplate(selectedTpl)
-	activeTemplate := getActiveTemplate()
-	inactiveTemplate := getInactiveTemplate()
-
-	prompt := OktetoSelector{
```
teresaromero, line 86: This logic of creating a new selector is now under NewOktetoSelector

cmd/utils/selector.go
```diff
@@ -1,514 +0,0 @@
...
-		Templates: &promptui.SelectTemplates{
-			Label:    "{{ .Label }}",
-			Selected: selectedTemplate,
-			Active:   activeTemplate,
-			Inactive: inactiveTemplate,
-			FuncMap:  promptui.FuncMap,
-		},
-	}
-
-	prompt.Templates.FuncMap["oktetoblue"] = oktetoLog.BlueString
```
teresaromero, line 99: new funcion Ask() is now performing the question, using the pointer of the OktetoSelector with the info regarding the selector

cmd/utils/selector.go
```diff
@@ -1,514 +0,0 @@
...
-	active    *template.Template
-	inactive  *template.Template
-	selected  *template.Template
-	details   *template.Template
-	help      *template.Template
-	extraInfo *template.Template
-}
-
-//SelectorItem represents a selectable item on a selector
-type SelectorItem struct {
```
teresaromero, line 70: This stuct is now private, there where fields just used to use this struct as context selector, so i created a new struct at context to represent it in that package

cmd/utils/selector.go
```diff
@@ -1,514 +0,0 @@
...
-	esc        = "\033["
-	showCursor = esc + "?25h"
-)
-
-// ErrInvalidOption is raised when the selector has an invalid option
-var ErrInvalidOption = errors.New("invalid option")
-
-// OktetoSelector represents the selector
-type OktetoSelector struct {
-	Label string
```
teresaromero, line 47: changed to private all properties of the struct except Label witch is exposed as is used on some logic out of the package, but rest are variables used internally

cmd/utils/selector.go
```diff
@@ -1,514 +0,0 @@
...
-	for _, option := range options {
-		if optionSelected == option.Name {
-			return true
-		}
-	}
-	return false
-}
-
-//Run runs the selector prompt
-func (s *OktetoSelector) Run(ctx context.Context) (string, bool, error) {
```
teresaromero, line 119: this method is now internal, Ask is the only public funcion

cmd/utils/selector.go
```diff
@@ -1,514 +0,0 @@
...
-	}
-
-	tpls.extraInfo = tpl
-
-	s.OktetoTemplates = tpls
-
-	return nil
-}
-
-func (s OktetoSelector) getInitialPosition(ctx context.Context) (int, error) {
```
teresaromero, line 378: this is removed, we no longer need to calculate this here, we calculate it at the context scope, also, this removes dependency with the okteto pkg

cmd/utils/selector.go
```diff
@@ -1,514 +0,0 @@
...
-func render(tpl *template.Template, data interface{}) []byte {
-	var buf bytes.Buffer
-	err := tpl.Execute(&buf, data)
-	if err != nil {
-		return []byte(fmt.Sprintf("%v", data))
-	}
-	return buf.Bytes()
-}
-
-type stdout struct{}
```
teresaromero, line 469: stdout have been moved to stdout.go under prompt pkg

cmd/utils/selector.go
```diff
@@ -1,514 +0,0 @@
...
-	Size  int
-
-	Templates *promptui.SelectTemplates
-	Keys      *promptui.SelectKeys
-
-	OktetoTemplates *oktetoTemplates
-}
-
-//oktetoTemplates stores the templates to render the text
-type oktetoTemplates struct {
```
teresaromero, line 58: this and all the templating functions have been moved to templates.go under prompt pkg

cmd/utils/dev.go
```diff
@@ -212,98 +202,10 @@ func GetDevFromManifest(manifest *model.Manifest, devName string) (*model.Dev, e
 		return nil, err
 	}
 
-	return manifest.Dev[devKey], nil
-}
-
-//AskYesNo prompts for yes/no confirmation
-func AskYesNo(q string) (bool, error) {
```
teresaromero, line 219: this has been moved to prompt/questions.go as is logic to get input from the user

cmd/utils/dev.go
```diff
@@ -212,98 +202,10 @@ func GetDevFromManifest(manifest *model.Manifest, devName string) (*model.Dev, e
...
-			break
-		}
-
-		oktetoLog.Fail("input must be 'y' or 'n'")
-	}
-
-	return answer == "y", nil
-}
-
-func AskForOptions(options []string, label string) (string, error) {
```
teresaromero, line 237: this has been removed and now where was used we use NewOktetoSelector -> Ask

cmd/utils/dev.go
```diff
@@ -212,98 +202,10 @@ func GetDevFromManifest(manifest *model.Manifest, devName string) (*model.Dev, e
...
-func AskIfOktetoInit(devPath string) bool {
-	result, err := AskYesNo(fmt.Sprintf("okteto manifest (%s) doesn't exist, do you want to create it? [y/n] ", devPath))
-	if err != nil {
-		return false
-	}
-	return result
-}
-
-// AsksQuestion asks a question to the user
-func AsksQuestion(q string) (string, error) {
```
teresaromero, line 280: also, this has been moved to prompt/questions.go

cmd/utils/dev.go
```diff
@@ -212,98 +202,10 @@ func GetDevFromManifest(manifest *model.Manifest, devName string) (*model.Dev, e
...
-	if err != nil {
-		oktetoLog.Infof("invalid init option: %s", err)
-		return "", fmt.Errorf("invalid option")
-	}
-
-	return options[i], nil
-}
-
-//AskIfOktetoInit asks if okteto init should be executed
-func AskIfOktetoInit(devPath string) bool {
```
teresaromero, line 271: This was only used at up.go, i moved the logic as AskYesNo from there and removed this function

cmd/utils/dev.go
```diff
@@ -212,98 +202,10 @@ func GetDevFromManifest(manifest *model.Manifest, devName string) (*model.Dev, e
...
-	oktetoLog.Question(q)
-	if _, err := fmt.Scanln(&answer); err != nil {
-		return "", err
-	}
-
-	return answer, nil
-}
-
-//AskIfDeploy asks if a new deployment must be created
-func AskIfDeploy(name, namespace string) error {
```
teresaromero, line 292: this has been removed as it was unused code

pkg/prompt/selector.go
teresaromero: @ifbyol i've commented the code under deleted selector.go and dev.go which have the "moved" items to the new pkg `prompt`

pkg/prompt/selector.go
teresaromero: i will upload a commit with this :)

cmd/utils/selector.go
teresaromero: i finally moved to `prompt` pkg as i felt this was not really log... but input from the user. Also i moved some other question related functions (Yes/No and question)

teresaromero:
i'm closing this in favor of https://github.com/okteto/okteto/pull/3033

Issue #2820, opened 2022-06-20
okteto deploy doesn't accept absolute paths on windows


jLopezbarb:
**Describe the bug**
When I run an executable, from an absolute path on windows it's not finding the correct executable because it's removing all backslash
``` yaml
deploy:
  - C:\Users\myuser\Repos\okteto\bin\okteto.exe deploy -f requirements.txt
```

**To Reproduce**

1. Copy the yaml posted above
2. Run `okteto.exe deploy`
3. It will fail with the following error:
```
 i  Using jlopezbarb @ cloud.okteto.com as context
 i  Images were already built. To rebuild your images run 'okteto build' or 'okteto deploy --build'
 i  Running C:\Users\javif\Repos\okteto\bin\okteto.exe deploy -f requirements.txt
/bin/bash: C:UsersjavifReposoktetobinokteto.exe: command not found
```

**Expected behavior**
I expect this to work on windows the same way it works on linux

**Desktop (please complete the following information):**

- OS: Windows
- Version 2.4.0

jLopezbarb:
@AgustinRamiroDiaz Do you know what can cause the removal of the backslash when running this command:
`C:\\Windows\\system32\\bash.exe C:\Users\javif\Repos\okteto\bin\okteto.exe deploy -f requirements.txt`

jLopezbarb:
This bug is also happenning on 2.3, 2.2, 2.1 and 2.0 versions

AgustinRamiroDiaz:
@jLopezbarb maybe it's related to this issue https://github.com/okteto/okteto/issues/2592

Seems like it's trying to run from bash inside wsl, that's why the message is `/bin/bash: C:UsersjavifReposoktetobinokteto.exe: command not found` (in windows `bin/bash` should throw a different error since it shouldn't exist) and that could be the reason for the exclusion of the `\`

PR #3019, opened 2022-08-19
fix: use specific terminal

### LINKED ISSUE #2592: `deploy` on Windows uses bash
**Describe the bug**
I added a custom command to deploy a service. However instead of okteto using powershell (from which it was called and where I have all necessary commands) it tried using WSL where commands are missing.

**To Reproduce**
add
```
deploy:
  - name: test
    command: whoami
```
to your `okteto.yml`. Notice that it outputs your username from WSL, not Windows.

**Expected behavior**
For okteto to not use WSL by default. There can either be a flag to tell it what shell to use or we can use `bash` as part of the command, but I think it should use the native environment by default.

- OS: Windows
- Version 10

jLopezbarb:

Fixes #2592 #2820

## Proposed changes
- Use the same terminal as the one executing the main okteto command

## Concerns
Since this is running remote and local I want to make sure it doesn't have any other implications. 

I have been testing in Windows on different terminals and it returns the correct user and we don't have the problem of absolute paths anymore.
Captura de pantalla 2022-08-19 125148

I have also tested on Linux machines and there seems to be no problem, but as this is a change that affects one of the main okteto commands I would like to know your opinion regarding this change.

### DIFF
--- a/cmd/utils/executor/executor.go
+++ b/cmd/utils/executor/executor.go
@@ -61,7 +61,7 @@ func NewExecutor(output string) *Executor {
 // Execute executes the specified command adding `env` to the execution environment
 func (e *Executor) Execute(cmdInfo model.DeployCommand, env []string) error {
 
-	cmd := exec.Command("bash", "-c", cmdInfo.Command)
+	cmd := exec.Command(cmdInfo.Command)
 	cmd.Env = append(os.Environ(), env...)
 	if err := e.displayer.startCommand(cmd); err != nil {
 		return err


AgustinRamiroDiaz:
I guess this is how it should've been from the start, but I'm also concerned if the wrapping around `bash -c` could have been due to some case we wanted to handle. Do the e2e tests run in different OS? That's something that will add more certainty

Also, I don't know if someone was using this as a kind of a "feature". i.e. using another terminal like `fish` and then putting bash commands in their `okteto.yml` to be compatible with the rest of their team. I think this is a rare case though and I'm in favor of the consistency of running the commands in the corresponding terminal.

AgustinRamiroDiaz (approved):

jLopezbarb:
> I guess this is how it should've been from the start, but I'm also concerned if the wrapping around `bash -c` could have been due to some case we wanted to handle. Do the e2e tests run in different OS? That's something that will add more certainty

The e2e tests run on both Windows and Linux, the tests will catch some errors, but because the cases are common (deploy with helm, kubectl, or stack) I don't think we can get any errors from them.

> Also, I don't know if someone was using this as a "feature". i.e. using another terminal like `fish` and then putting bash commands in their `okteto.yml` to be compatible with the rest of their team. I think this is a rare case though and I'm in favor of the consistency of running the commands in the corresponding terminal.

I don't think people used it as a feature since it changed both the OS and the working directory from which the command was launched.

cmd/utils/executor/executor.go
```diff
@@ -61,7 +61,7 @@ func NewExecutor(output string) *Executor {
 // Execute executes the specified command adding `env` to the execution environment
 func (e *Executor) Execute(cmdInfo model.DeployCommand, env []string) error {
 
-	cmd := exec.Command("bash", "-c", cmdInfo.Command)
+	cmd := exec.Command(cmdInfo.Command)
```
rberrelleza, line 64: This is going to change how commands are executed. Could we set this as a flag, announce deprecation, then do the full switch in a version or two? (Or announce it as a breaking change).

rberrelleza (approved):

cmd/utils/executor/executor.go
jLopezbarb: @rberrelleza what do you mean by setting this as a flag? the new behavior under the flag? If we do that, we'll be throwing a warning every time a user runs `okteto deploy/destroy` to use a file flag that is going to be deprecated.

If we set the old behavior under the flag, are we going to release a flag that is already deprecated? I prefer this idea and show the users a way of getting back the old behavior(via flag). We should also track how many people are having errors vs previous versions and see if we are having more errors after this change, WDYT?

ifbyol: Setting as default the new behavior could break customers' scripts/pipelines which ideally we should try to avoid unless we announce it as a breaking change and update the versions accordingly. 

If I understand correctly one of the issues, does this prevent the `okteto deploy` command to work on Windows? If so, we might enforce the new behavior just when the OS is windows. I mean, if for those cases it's not working, we won't break anything changing the behavior. For the rest of OS we can include a flag **defaulting to the old behavior**

jLopezbarb: @ifbyol The problem is that windows bash runs on an auxiliary operating system, which does not run on the same path or with the same user. Being an auxiliary operating system it does not correctly pair absolute paths.

We can activate it only for Windows, although I would like that there is no difference between different operating systems, since it is more difficult to maintain it.

ifbyol: @jLopezbarb I know, but one of our main goals is stability in the CLI, so a breaking change I don't think is a good option right now.

My point is, if in Windows this was never working as expected, doing this breaking change only for Windows would fix the issue. Now, if this wasn't working properly on some situations, then, maybe we have to rethink it because we could be broken other scenarios

AgustinRamiroDiaz: What if we add a feature flag to remove the `bach -c`? In this case we wouldn't do a breaking change and we'd also allow for the "correct" behavior with a flag like `--execute-without-bash`. We could incentivize the use of this flag and postpone it as the default behavior in some future big release

WDYT @ifbyol @jLopezbarb ?

cmd/utils/executor/executor.go
```diff
@@ -75,6 +76,18 @@ func (e *Executor) Execute(cmdInfo model.DeployCommand, env []string) error {
 	return err
 }
 
+// getCMD returns the command to be executed based on the OS
+// We needed to diferentiate between windows and other OS because we were having problems
+// with using bash on windows: see:
+// - https://github.com/okteto/okteto/issues/2820
+// - https://github.com/okteto/okteto/issues/2592
+func getCMD(command string) *exec.Cmd {
```
AdrianPedriza, line 84: `getCMDByOS` I think it is more specific. WDYT?

jLopezbarb: outdated comment

cmd/deploy/deploy.go
```diff
@@ -227,6 +228,7 @@ func Deploy(ctx context.Context) *cobra.Command {
 	cmd.Flags().StringArrayVarP(&options.Variables, "var", "v", []string{}, "set a variable (can be set more than once)")
 	cmd.Flags().BoolVarP(&options.Build, "build", "", false, "force build of images when deploying the development environment")
 	cmd.Flags().BoolVarP(&options.Dependencies, "dependencies", "", false, "deploy the dependencies from manifest")
+	cmd.Flags().BoolVarP(&options.RunWithoutBash, "run-without-bash", "", false, "execute commands with bash")
```
AgustinRamiroDiaz, line 231:
```suggestion
	cmd.Flags().BoolVarP(&options.RunWithoutBash, "run-without-bash", "", false, "execute commands without bash")
```

cmd/destroy/destroy.go
```diff
@@ -190,6 +191,7 @@ func Destroy(ctx context.Context) *cobra.Command {
 	cmd.Flags().BoolVar(&options.ForceDestroy, "force-destroy", false, "forces the development environment to be destroyed even if there is an error executing the custom destroy commands defined in the manifest")
 	cmd.Flags().StringVarP(&options.Namespace, "namespace", "n", "", "overwrites the namespace where the development environment was deployed")
 	cmd.Flags().StringVarP(&options.K8sContext, "context", "c", "", "context where the development environment was deployed")
+	cmd.Flags().BoolVarP(&options.RunWithoutBash, "run-without-bash", "", false, "execute commands with bash")
```
AgustinRamiroDiaz, line 194:
```suggestion
	cmd.Flags().BoolVarP(&options.RunWithoutBash, "run-without-bash", "", false, "execute commands without bash")
```

AgustinRamiroDiaz (approved):
I think this is a sweet spot between usability and backwards compatibility. We have to bear in mind we are adding another flag to the cli, which is something we are keeping an eye on since we are aiming for more stability

jLopezbarb:
@RinkiyaKeDad @ralamana I don't like the fact of having a flag with a negative preposition `--run-without-bash`. Do you think that the flag can be renamed so that this negative term is not used?

rberrelleza:
> @RinkiyaKeDad @ralamana I don't like the fact of having a flag with a negative preposition `--run-without-bash`. Do you think that the flag can be renamed so that this negative term is not used?

Agree with Javi. This is a very typical problem when running commands, what do other tools or platforms do?

rberrelleza:
> I guess this is how it should've been from the start, but I'm also concerned if the wrapping around `bash -c` could have been due to some case we wanted to handle. Do the e2e tests run in different OS? That's something that will add more certainty
> 
> 
> 
> Also, I don't know if someone was using this as a kind of a "feature". i.e. using another terminal like `fish` and then putting bash commands in their `okteto.yml` to be compatible with the rest of their team. I think this is a rare case though and I'm in favor of the consistency of running the commands in the corresponding terminal.

This is a very good point. I don't recall why it was added, but we didn't have -c from day 1; it was added to fix error conditions (maybe with env var injection?). Has someone gone through the commit history to see if there are any clues?

AgustinRamiroDiaz:
> @RinkiyaKeDad @ralamana I don't like the fact of having a flag with a negative preposition `--run-without-bash`. Do you think that the flag can be renamed so that this negative term is not used?

What if we rename the flag `--run-with-bash` and set the default value to `true` with `Cobra`?

AgustinRamiroDiaz:
@rberrelleza "Has someone gone through the commit history to see if there are any clues?" I've tried a bit and didn't find any clue. I used the vscode extension GitLens which helped find the PR, if anyone is interested in looking for something

RinkiyaKeDad:
> What if we rename the flag --run-with-bash and set the default value to true with Cobra?

+1 for this suggestion. We could also then simplify the name of the flag to just `--bash` maybe and make it clear in the help section what it does.

ifbyol:
@rberrelleza @AgustinRamiroDiaz @jLopezbarb   I think that the `bash -c` came when we created the `okteto deploy` command and we took the code from the backend installer. That code was there and we moved it here. 

In the installer, we always had control on the terminal and platform where the command was being run and we didn't consider that when we moved the code to the CLI.

> What if we rename the flag --run-with-bash and set the default value to true with Cobra?

That could work, but it would be also interesting to see how other CLI tools do it

jLopezbarb:
> > What if we rename the flag --run-with-bash and set the default value to true with Cobra?
> 
> +1 for this suggestion. We could also then simplify the name of the flag to just `--bash` maybe and make it clear in the help section what it does.

I don't like too much the idea of `--bash` because the user will have to run `okteto deploy --bash=false` which is not optimal on boolean flags. I prefer to use something like `okteto deploy --use-same-shell` although I don't like the name of that flag.

> @rberrelleza "Has someone gone through the commit history to see if there are any clues?" I've tried a bit and didn't find any clue. I used the vscode extension GitLens which helped find the PR, if anyone is interested in looking for something

I can see that the `bash -c` has been there since the start of the pipeline installer. Although, in the tests that I made it did not give errors I do not think that we can remove it and we have to use a flag to change the behavior.

AgustinRamiroDiaz:
@jLopezbarb @ifbyol  regarding the naming, a common convention is to shorten the wording with expressions like `--no-bash` (like the classic `--no-cache` flag). It's really short, but I prefer the long, more explicit name since it's a behavior only a few users will need, so typing it out is not a problem

jLopezbarb:
> @jLopezbarb @ifbyol regarding the naming, a common convention is to shorten the wording with expressions like `--no-bash` (like the classic `--no-cache` flag). It's really short, but I prefer the long, more explicit name since it's a behavior only a few users will need, so typing it out is not a problem
> 
> https://github.com/moby/buildkit/blob/9528036502c589626bfb60600af6ce5ee427f479/cmd/buildctl/build.go#L67

+1 to `--no-bash`! Do we have documented that the commands are executed on a bash terminal? If we have that we can refer to the flag and the information about deploy executing from a bash terminal. If not I'll create it before releasing the new CLI version.
If the question/error keeps popping up, we could include a question on FAQ to guide the users.

teresaromero (approved):

PR #3033, opened 2022-08-24
Fix dev selector typo

teresaromero:
Fixes #2778 

## Proposed changes

- Separate selection logic from GetDevFromManifest. This function will though an error (ErrNoDevSelected) if devName is empty and manifest has multiple devs defined.
- new SelectDevFromManifest fallback function when err is ErrNoDevSelected, where label is a param that represents the question of the selector

PR #3034, opened 2022-08-24
Refactor selector

### LINKED ISSUE #2778: typo in prompt when running okteto down

teresaromero:
Fixes #2778 

This PR does a refactor at the selector AskForOptionsOkteto to make  new func SelectDevFromManifest testable. 
SelectDevFromManifest func was introduces here https://github.com/okteto/okteto/pull/3033

### DIFF
--- a/cmd/context/selector.go
+++ b/cmd/context/selector.go
@@ -117,10 +117,11 @@ func getInitialPosition(options []utils.SelectorItem) int {
 	return -1
 }
 
-func selectedIsOkteto(options []utils.SelectorItem, selected string) bool {
-	for _, item := range options {
-		if item.Label == selected {
-			return item.IsOkteto
+func selectedIsOkteto(selected string) bool {
+	ctxStore := okteto.ContextStore()
+	for ctxName, okCtx := range ctxStore.Contexts {
+		if selected == ctxName {
+			return okCtx.IsOkteto
 		}
 	}
 	return false
--- a/cmd/context/selector.go
+++ b/cmd/context/selector.go
@@ -100,9 +100,9 @@ func getK8sClusters(k8sClusters []string) []utils.SelectorItem {
 }
 
 func getInitialPosition(options []utils.SelectorItem) int {
-	currentContext := okteto.Context().Name
+	currentContext := okteto.ContextStore().CurrentContext
 	for indx, item := range options {
-		if strings.Contains(item.Name, currentContext) {
+		if item.Enable && item.Name == currentContext {
 			return indx
 		}
 	}
@@ -111,10 +111,9 @@ func getInitialPosition(options []utils.SelectorItem) int {
 
 func selectedIsOkteto(selected string) bool {
 	ctxStore := okteto.ContextStore()
-	for ctxName, okCtx := range ctxStore.Contexts {
-		if selected == ctxName {
-			return okCtx.IsOkteto
-		}
+	selectedCtx, ok := ctxStore.Contexts[selected]
+	if !ok {
+		return false
 	}
-	return false
+	return selectedCtx.IsOkteto
 }
--- a/cmd/context/selector.go
+++ b/cmd/context/selector.go
@@ -34,7 +34,7 @@ func getContextsSelection(ctxOptions *ContextOptions) []utils.SelectorItem {
 	}
 	clusters := make([]utils.SelectorItem, 0)
 
-	clusters = append(clusters, utils.SelectorItem{Name: okteto.CloudURL, Label: cloudOption, Enable: true, IsOkteto: true})
+	clusters = append(clusters, utils.SelectorItem{Name: okteto.CloudURL, Label: cloudOption, Enable: true})
 	clusters = append(clusters, getOktetoClusters(true)...)
 	if len(k8sClusters) > 0 {
 		clusters = append(clusters, getK8sClusters(k8sClusters)...)
@@ -67,13 +67,9 @@ func getOktetoClusters(skipCloud bool) []utils.SelectorItem {
 		orderedOktetoClusters = append(
 			orderedOktetoClusters,
 			utils.SelectorItem{
-				Name:      ctxName,
-				Label:     ctxName,
-				Enable:    true,
-				IsOkteto:  true,
-				Namespace: okCtx.Namespace,
-				Builder:   okCtx.Builder,
-				Registry:  okCtx.Registry,
+				Name:   ctxName,
+				Label:  ctxName,
+				Enable: true,
 			})
 	}
 	sort.Slice(orderedOktetoClusters, func(i, j int) bool {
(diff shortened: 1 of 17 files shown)


teresaromero:
Deepsource checks are failing because the pipeline uses default branch master and the PR is against other PR

cmd/utils/selector.go
```diff
@@ -78,7 +77,7 @@ type SelectorItem struct {
 }
 
 //AskForOptionsOkteto given some options ask the user to select one
-func AskForOptionsOkteto(ctx context.Context, options []SelectorItem, label, selectedTpl string) (string, bool, error) {
+func AskForOptionsOkteto(ctx context.Context, options []SelectorItem, label, selectedTpl string, initialPosition int) (string, bool, error) {
```
jLopezbarb: Are we using `ctx`? I think we could remove it since the code that was using it has been moved to another file

teresaromero: this was changed some commits after :) i think u where viewing a specific commit filter

cmd/context/selector.go
```diff
@@ -92,17 +88,32 @@ func getK8sClusters(k8sClusters []string) []utils.SelectorItem {
...
+	currentContext := okteto.ContextStore().CurrentContext
+	for indx, item := range options {
+		if item.Enable && item.Name == currentContext {
+			return indx
+		}
+	}
+	return -1
+}
+
+func selectedIsOkteto(selected string) bool {
```
ifbyol, line 112: Can it be named only `isOkteto`? `selected` gives it a context where the function can be used, but at the end, this is only checking if a context belongs to Okteto or not and it can be used in other places

cmd/context/use.go
```diff
@@ -124,18 +124,21 @@ func (c *ContextCommand) Run(ctx context.Context, ctxOptions *ContextOptions) er
...
-	ctxOptions.IsOkteto = isOkteto
 
 	if isCreateNewContextOption(oktetoContext) {
 		oktetoContext, err = askForOktetoURL()
 		if err != nil {
 			return "", err
 		}
 		ctxOptions.IsOkteto = true
+	} else {
+		ctxOptions.IsOkteto = selectedIsOkteto(oktetoContext)
```
ifbyol, line 141: I'm confused with this change within the function. `ctxOptions` is being used to get the list of contexts to display and then, we modify its value based on what the user has selected. In general, this is a smell and error prone. Function parameters should be immutable and ideally this function should return all the information needed so the caller can change the `ctxOptions` based on the response.

Anyways, it is out of the scope of this PR as this is how it is now

cmd/utils/selector.go
```diff
@@ -42,7 +40,12 @@ const (
 // ErrInvalidOption is raised when the selector has an invalid option
 var ErrInvalidOption = errors.New("invalid option")
 
-// OktetoSelector represents the selector
+// OktetoSelectorInterface represents an interface for a selector
```
ifbyol, line 43: Maybe this whole file can be moved to its own package named `selector`. it would be great to start emptying the `utils` package.

If you want to do it in this same PR, please, in another commit so we can check the difference between commits

ifbyol (approved):
Is this branch going to be merged first in your other branch and they will be merge together or this will wait until the other?

cmd/utils/selector.go
teresaromero: i had that in mind, but as it was getting too big of refactor i decided to limit this refactor to make testable the selector logic. This could be a tech-debt issue which i will open and link

cmd/context/selector.go
teresaromero: i've been thinking and moved this check to the okteto pkg, i think it makes more sense

teresaromero:
> Is this branch going to be merged first in your other branch and they will be merge together or this will wait until the other?

I had in mind merging it after the other branch is merged to master, this way i can have clean deepsource checks before the refactor.

cmd/utils/selector.go
teresaromero: new issue  https://github.com/okteto/okteto/issues/3043

jLopezbarb (approved):

cmd/context/selector.go
```diff
@@ -34,7 +34,7 @@ func getContextsSelection(ctxOptions *ContextOptions) []utils.SelectorItem {
 	}
 	clusters := make([]utils.SelectorItem, 0)
 
-	clusters = append(clusters, utils.SelectorItem{Name: okteto.CloudURL, Label: cloudOption, Enable: true, IsOkteto: true})
+	clusters = append(clusters, utils.SelectorItem{Name: okteto.CloudURL, Label: cloudOption, Enable: true})
 	clusters = append(clusters, getOktetoClusters(true)...)
 	if len(k8sClusters) > 0 {
```
AgustinRamiroDiaz, line 39: nit: we could remove the if since `getK8sClusters` should work with empty slices

cmd/namespace/use.go
```diff
@@ -99,7 +99,9 @@ func (nc *NamespaceCommand) getNamespaceFromSelector(ctx context.Context) (strin
 	if err != nil {
 		return "", err
 	}
-	ns, _, err := utils.AskForOptionsOkteto(ctx, namespaces, "Select the namespace you want to use:", "Namespace")
+	initialPosition := getInitialPosition(namespaces)
+	selector := utils.NewOktetoSelector("Select the namespace you want to use:", "Namespace")
+	ns, err := selector.AskForOptionsOkteto(namespaces, initialPosition)
```
AgustinRamiroDiaz, line 104: WDYT about creating a functioni for this? I've seen that we have this pattern a few times. I think it's a sign that most of the times we are interested in the `ns` outcome of this sequence, and unifying it in one function call would simplify the api from a consumer standpoint

AgustinRamiroDiaz (approved):

cmd/context/selector.go
teresaromero: i would like to keep the changes as tight as possible to the refactor purpose, thnks!

cmd/namespace/use.go
teresaromero: i decoupled the selector creation and the question asked to be able to fake the selector behaviour. In any case we use `AskForOptionsOkteto` we want to have as a result a string with the selection, this is a repeteated pattern in every request for user information we do.

if you refer to the initialPosition func, this is required at context and namespace selection so the selector prompt starts with the arrow at the current one.

PR #3035, opened 2022-08-24
Add a flag to show a specific field in `okteto context show`

odedlaz:
Signed-off-by: Oded Lazar <odedl@talon-sec.com>

## Proposed changes

`okteto context show` prints the entire context to `stdout`. In some scenarios this is unwanted. For example, we have an automation that uses the context to infer the `namespace`. We use `jq` to extract the field we want, but that incurs an added dependency.

This PR adds a flag to `okteto context show` which is the key of the context. example:

image

### DIFF
--- a/cmd/context/show.go
+++ b/cmd/context/show.go
@@ -17,21 +17,42 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"reflect"
+	"strings"
 
 	"github.com/okteto/okteto/cmd/utils"
+	oktetoErrors "github.com/okteto/okteto/pkg/errors"
 	oktetoLog "github.com/okteto/okteto/pkg/log"
 	"github.com/okteto/okteto/pkg/okteto"
 	"github.com/spf13/cobra"
 	"gopkg.in/yaml.v2"
 )
 
+const hintUrl = "https://okteto.com/docs/reference/cli/#show"
+
+func printContextMember(ctx *okteto.OktetoContext, key string) error {
+	val := reflect.ValueOf(ctx).Elem()
+	for i := 0; i < val.NumField(); i++ {
+		if strings.EqualFold(val.Type().Field(i).Name, key) {
+			val := fmt.Sprintf("%v", val.Field(i).Interface())
+			oktetoLog.Println(val)
+			return nil
+		}
+	}
+
+	return oktetoErrors.UserError{
+		E:    fmt.Errorf("unknown context key: '%s'", key),
+		Hint: hintUrl,
+	}
+}
+
 // Show current context
 func Show() *cobra.Command {
 	var output string
 	cmd := &cobra.Command{
-		Use:   "show",
-		Args:  utils.NoArgsAccepted("https://okteto.com/docs/reference/cli/#show"),
+		Use:   "show [key]",
 		Short: "Print the current context",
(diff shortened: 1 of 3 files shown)


odedlaz:
@jLopezbarb , wdyt?

jLopezbarb:
> @jLopezbarb , wdyt?

We are looking at the concerns that using reflection may have. We will come back to you as soon as we're clear on how we want to implement this without using reflect.

AgustinRamiroDiaz:
@odedlaz I found this feature useful for the usecase you mention, and I like the proposal of extending the functionality instead of changing it.

@jLopezbarb @odedlaz I think we can get around using `reflect` if we use a `switch case` or a `map[OktetoContext] func (OktetoContext) string`. This way we can manually handle better error printing when the value doesn't match, and explicitly handle only the fields we want to expose

cmd/context/show.go
```diff
@@ -17,21 +17,42 @@ import (
...
 )
 
+const hintUrl = "https://okteto.com/docs/reference/cli/#show"
+
+func printContextMember(ctx *okteto.OktetoContext, key string) error {
+	element := reflect.ValueOf(ctx).Elem()
+	for i := 0; i < element.NumField(); i++ {
+		if strings.EqualFold(element.Type().Field(i).Name, key) {
+			memberValue := fmt.Sprintf("%v", element.Field(i).Interface())
+			oktetoLog.Println(memberValue)
```
AgustinRamiroDiaz, line 38: I think it'd be great if the new code we add is fully testable, this includes printing. 
We have 2 options here:
- Add the printing function as a parameter, something like `func printContextMember(ctx *okteto.OktetoContext, key string, print func(...interface{})) error`
- Return `(string, error)` and let the caller do the printing and error handling

odedlaz:
@AgustinRamiroDiaz I don't mind introducing this change. I just thought (and still think) that reflect is cleaner. 

We expose the same struct using reflect (json.marshal uses reflect) without altering it, so I thought that filtering by the same values will work.

I can also use the json tag as the marker instead of the field member name.

odedlaz:
@AgustinRamiroDiaz , take a look at the new commit:
- Move the reflection code to utils + added tests
- The extraction code now takes the tag name instead of the struct name

ifbyol:
@odedlaz We need to review this internally. 

`okteto context show` has 2 type of output (there is a flag in the command), `json` or `yaml`. The default one is `json` and just printing the value of the property without double quotes is not a valid json, so we would be breaking the command contract. So, in the example you provided in the screenshot, the output should be 

```
{
    "isOkteto": true
}
```

Also, from a product perspective, it might make sense to be able to select more than one key, for example: `okteto show --keys name,builder,isOkteto` and the response should be something like:

```
{
  "name": "https://cloud.okteto.com",
  "builder": "tcp://buildkit.cloud.okteto.net:1234",
  "isOkteto": true
}
```

ifbyol:
@odedlaz Just to try to understand and see if there are other alternatives that you can you use in the meantime. Do you need this to be executed in the deploy section of a manifest? Or is it some automation out of Okteto?

For example, if it is to be used in the `deploy` section, we inject the environment variable `OKTETO_NAMESPACE` that tells you the namespace where those commands are being executed

PR #3037, opened 2022-08-25
allow okteto destroy on deploy cmd

### LINKED ISSUE #2762: Support for running `okteto destroy` in the deploy section of the okteto manifest
**Describe the bug**
Some users are running `okteto destroy` as part of the deploy section, something like this:
```
deploy:
  - okteto destroy
  - helm upgrade --install myapp ./mychart
```
this is especially convenient when you do changes in your chart that don't support in-place updates, like modifying statefulsets.
Also, for development, helm is usually faster destroying + installing, than upgrading a previous installation.

Today, we show a waning because `okteto destroy` is removing the configmap created by `okteto destroy`.

**Expected behavior**
We could handle this use case better and skip the configmap removal on `okteto destroy if it's running as part of the deploy commands

jLopezbarb:
Fixes #2762

## Proposed changes
- Fixes path on native windows
- Modify destroy temp kubeconfig name to not collide with `okteto deploy` temp kubeconfig
- If okteto destroy inside okteto deploy it will skip the destruction of the configmap


cmd/destroy/destroy.go
```diff
@@ -167,12 +169,14 @@ func Destroy(ctx context.Context) *cobra.Command {
 				k8sClientProvider: okteto.NewK8sClientProvider(),
 			}
 
-			kubeconfigPath := deploy.GetTempKubeConfigFile(name)
+			kubeconfigPath := getTempKubeConfigFile(name)
```
jLopezbarb, line 172: @ifbyol Why are we creating the `tempkubeconfig` for destroy? It's not using the proxy. Could we remove it and use directly the context kubeconfig?

jLopezbarb:
I still need to create an e2e testing this functionality

cmd/deploy/kubeconfig.go
```diff
@@ -84,7 +85,8 @@ func (*KubeConfig) GetCMDAPIConfig() (*clientcmdapi.Config, error) {
 	return okteto.Context().Cfg, nil
 }
 
-//GetTempKubeConfigFile returns a where the temp kubeConfigFile should be stored
+// GetTempKubeConfigFile returns a where the temp kubeConfigFile should be stored
 func GetTempKubeConfigFile(name string) string {
-	return fmt.Sprintf(tempKubeConfigTemplate, config.GetUserHomeDir(), name, time.Now().UnixMilli())
+	tempKubeConfigTemplate := fmt.Sprintf("kubeconfig-%s-%d", name, time.Now().UnixMilli())
```
teresaromero, line 90: the destroy version of this func is `kubeconfig-destroy....` can we add here then `deploy` to the name?

cmd/destroy/destroy.go
```diff
@@ -240,16 +244,21 @@ func (dc *destroyCommand) runDestroy(ctx context.Context, opts *Options) error {
...
 		Name:      opts.Name,
 		Namespace: namespace,
 		Status:    pipeline.DestroyingStatus,
 		Filename:  opts.ManifestPathFlag,
 	}
-	cfg, err := pipeline.TranslateConfigMapAndDeploy(ctx, data, c)
-	if err != nil {
-		return err
+	var cfg *v1.ConfigMap
+	if !utils.LoadBoolean(model.OktetoWithinDeployCommandContextEnvVar) {
```
teresaromero, line 255: what does this `OktetoWithinDeployCommandContextEnvVar`mean?

cmd/deploy/kubeconfig.go
jLopezbarb: I prefer to remove it if we are not using it, but I mentioned @ifbyol to know if we needed for any other thing

cmd/destroy/destroy.go
jLopezbarb: It means that the command is being executed by an `okteto deploy`, all commands executed by `okteto deploy` have this environment variable

teresaromero (approved):

cmd/destroy/destroy.go
ifbyol: We are creating the temporal kubeconfig to avoid to change the user's kubeconfig. Bear in mind that we are setting the okteto context in the kubeconfig. But the user's current k8s context can be a different one. 

I would not remove the temporal file to avoid the side effect of changing user's kubeconfig or use the wrong kubeconfig

cmd/destroy/destroy.go
```diff
@@ -167,12 +169,14 @@ func Destroy(ctx context.Context) *cobra.Command {
...
 			}
 
-			kubeconfigPath := deploy.GetTempKubeConfigFile(name)
+			kubeconfigPath := getTempKubeConfigFile(name)
 			if err := kubeconfig.Write(okteto.Context().Cfg, kubeconfigPath); err != nil {
 				return err
 			}
 			os.Setenv("KUBECONFIG", kubeconfigPath)
-			defer os.Remove(kubeconfigPath)
+			if !utils.LoadBoolean(model.OktetoWithinDeployCommandContextEnvVar) {
```
ifbyol, line 177: This check is spread around all the code and it is error prone. We might forget to include it some where in the future.

I think we should create an abstraction and have 2 command runners. One to be used when `model.OktetoWithinDeployCommandContextEnvVar` is true and another one to be used when it is false.

So it is clearer what is the behavior in both cases. Some functions can be common but other can be defined per runner. For example, `setErrorStatus` looks a good candidate to be defined per runner. A new `DestroyConfig` function can also be a good candidate. 

In this way, we only have this check once (when creating the runner) and it is easier to see what is executed on each type

jLopezbarb: Thanks for the tip, I'll refactor the code with those changes 😄

cmd/deploy/kubeconfig.go
jLopezbarb: I added your suggestion since we need it for both commands 😉

cmd/destroy/destroy_test.go
```diff
@@ -154,7 +156,7 @@ func TestDestroyWithErrorDeletingVolumes(t *testing.T) {
 	assert.False(t, destroyer.destroyed)
 	assert.False(t, destroyer.destroyedVolumes)
 
-	//check if configmap has been created
+	// check if configmap has been created
```
AdrianPedriza, line 159: should we also check values in it?

cmd/destroy/destroy_test.go
```diff
@@ -576,7 +581,238 @@ func TestDestroyWithoutError(t *testing.T) {
...
+			}
+
+			err := cmd.runDestroy(ctx, opts)
+
+			assert.NoError(t, err)
+			assert.ElementsMatch(t, tt.want, executor.executed)
+			assert.True(t, destroyer.destroyed)
+			assert.True(t, destroyer.destroyedVolumes)
+
+			// check if configmap has been created
```
AdrianPedriza, line 815: should we also check values in it?

AdrianPedriza (approved):

cmd/destroy/destroy.go
```diff
@@ -86,6 +86,7 @@ type destroyCommand struct {
 	nsDestroyer       destroyer
 	secrets           secretHandler
 	k8sClientProvider okteto.K8sClientProvider
+	runner            runnerI
```
ifbyol, line 89: If we rename the interface, we should rename also this property
```suggestion
	configMapHandler            runnerI
```

cmd/destroy/destroy_test.go
```diff
@@ -567,6 +571,238 @@ func TestDestroyWithoutError(t *testing.T) {
...
+			fakeClient, _, err := cmd.k8sClientProvider.Provide(api.NewConfig())
+			if err != nil {
+				t.Fatal("could not create fake k8s client")
+			}
+			cfg, _ := configmaps.Get(ctx, pipeline.TranslatePipelineName(opts.Name), okteto.Context().Namespace, fakeClient)
+			assert.Nil(t, cfg)
```
ifbyol, line 590: I think this assert would be more accurate if we check that the error is a not found. cfg could be nil for other reasons, I think any error getting the configmap would return a nil `cfg`

cmd/destroy/runners.go
```diff
@@ -0,0 +1,75 @@
...
+
+// oktetoInsideOktetoRunner is the runner used when the okteto is executed
+// inside an okteto deploy command
+type oktetoInsideOktetoRunner struct{}
+
+// oktetoDefaultRunner is the runner used when the okteto is executed
+// directly
+type oktetoDefaultRunner struct{}
+
+func newRunner() runnerI {
```
ifbyol, line 44: Same about renaming the function

cmd/destroy/runners.go
```diff
@@ -0,0 +1,75 @@
...
+	if utils.LoadBoolean(model.OktetoWithinDeployCommandContextEnvVar) {
+		return &oktetoInsideOktetoRunner{}
+	}
+	return &oktetoDefaultRunner{}
+}
+
+func (*oktetoDefaultRunner) translateConfigMapAndDeploy(ctx context.Context, data *pipeline.CfgData, c kubernetes.Interface) (*apiv1.ConfigMap, error) {
```
ifbyol, line 51: Minor thing, you could reduce the function signature if you specify `kubernetes.Interface` as a dependency of `oktetoDefaultRunner`. In that way, the `oktetoInsideOktetoRunner` doesn't need that dependency and you have 1 less parameter

cmd/destroy/runners.go
```diff
@@ -0,0 +1,75 @@
...
+	setErrorStatus(context.Context, *apiv1.ConfigMap, *pipeline.CfgData, error, kubernetes.Interface) error
+}
+
+// oktetoInsideOktetoRunner is the runner used when the okteto is executed
+// inside an okteto deploy command
+type oktetoInsideOktetoRunner struct{}
+
+// oktetoDefaultRunner is the runner used when the okteto is executed
+// directly
+type oktetoDefaultRunner struct{}
```
ifbyol, line 42: Why the `okteto` prefix? I think just `default...` is ok

cmd/destroy/runners.go
```diff
@@ -0,0 +1,75 @@
...
+// or an okteto destroy directly
+type runnerI interface {
+	translateConfigMapAndDeploy(context.Context, *pipeline.CfgData, kubernetes.Interface) (*apiv1.ConfigMap, error)
+	destroyConfigMap(context.Context, *apiv1.ConfigMap, string, kubernetes.Interface) error
+	setErrorStatus(context.Context, *apiv1.ConfigMap, *pipeline.CfgData, error, kubernetes.Interface) error
+}
+
+// oktetoInsideOktetoRunner is the runner used when the okteto is executed
+// inside an okteto deploy command
+type oktetoInsideOktetoRunner struct{}
```
ifbyol, line 38: Same about `okteto` prefix. Also, I don't think `insideOkteto` is a good name or it can be misleading. What does inside Okteto means? Maybe we can name them `standalone` and `subcommand`.

Other option could be to name them according to what they really do (modify or not the config map) instead of setting the names based on the condition to use one or the other, but nothing comes to my mind

cmd/destroy/destroy_test.go
jLopezbarb: Updated. Writing comment since the outdated label has not been added to the comment

ifbyol (approved):

AdrianPedriza (approved):

PR #3050, opened 2022-08-31
TrackReconnect new param to track dev pod recreation

### LINKED ISSUE #3046: Track is `okteto up` reconnects are due to dev pod being recreated
We are tracking `okteto up` reconnects today. `okteto up` reconnects are usually because of network issues, laptop going to sleep, or dev pod recreation. It would be nice to track the latest case, to analyze the impact, for example, of using spot instances

Track if `Reconnect` is because a dev pod is recreated

teresaromero:

Resolves #3046 

## Proposed changes

- added new param to TrackReconnect
- isDepPodRecreated as cause for retry

## Test
- run `okteto up` 
- while being in sync session with dev container, kill dev pod
- when reconnecting the tracking of this event should have the reason for reconnection

### DIFF
--- a/cmd/up/activate.go
+++ b/cmd/up/activate.go
@@ -35,6 +35,7 @@ import (
 	"github.com/okteto/okteto/pkg/registry"
 	apiv1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 )
 
 func (up *upContext) activate() error {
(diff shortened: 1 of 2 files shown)


cmd/up/activate.go
```diff
@@ -119,7 +126,9 @@ func (up *upContext) activate() error {
 	}
 
 	if up.isRetry {
-		analytics.TrackReconnect(true)
+		// isDevPodRecreated if the pod UID has changed from last retry
+		isDevPodRecreated := lastPodUID != up.Pod.UID
```
jLopezbarb, line 130: @pchico83 is a boolean sufficient or would a string indicating the cause look better? If we could not detect the cause it would be an `unrecognized` until we can detect that type.
For example:

```json
{
  "cause": "unrecognised"/"dev-pod-recreated"/"lost-connection"
}
```

pchico83: Yes, a string is more flexible for future changes

jLopezbarb (approved):

ifbyol (approved):

PR #3051, opened 2022-08-31
fix secrets replacement when okteto build

### LINKED ISSUE #3044: `okteto build` is not getting my okteto secrets
**Describe the bug**
I have an okteto secret named `TEST_SECRET` and the following okteto manifest:
```
build:
  test:
    context: .
    args:
      - TEST_SECRET=${TEST_SECRET}
```
but `TEST_SECRET` is not getting its value when running `okteto build`. It does when running `okteto deploy --build` or if `okteto up` builds the images, but not for `okteto build`.     

**To Reproduce**
Create the following files:

`okteto.yml`:
```
build:
  test:
    context: .
    args:
      - TEST_SECRET=${TEST_SECRET}

deploy:
  - echo "fake deploy"
```

`Dockerfile`:
```
FROM nginx

ARG TEST_SECRET

RUN echo "TEST_SECRET: $TEST_SECRET" && sleep 10
```

`okteto deploy --build` replaces `TEST_SECRET` but `okteto build` doesn't

**Additional context**
The Okteto CLI queries okteto secrets at the beginning of command execution. We are probably missing this for the `okteto build` command

AdrianPedriza:
# secrets are not resolved when `okteto build` runs

Fixes #3044

## Proposed changes**

- Secrets are resolved once context is loaded so solution was to move `loadContext` func before read a manifest. With this change, manifest will be read once context has been laoded


cmd/build/command.go
```diff
@@ -129,3 +146,38 @@ func validateDockerfile(file string) error {
 	_, _, err = instructions.Parse(parsedDockerfile.AST)
 	return err
 }
+
+func (bc *Command) loadContext(ctx context.Context, options *types.BuildOptions) error {
```
jLopezbarb, line 150: Can we add this function to the iface?

cmd/build/command.go
```diff
@@ -90,6 +96,17 @@ func Build(ctx context.Context) *cobra.Command {
 func (bc *Command) getBuilder(options *types.BuildOptions) (Builder, error) {
 	var builder Builder
 
+	if options.File != "" && validateDockerfile(options.File) == nil {
```
ifbyol, line 99: Why this change?

cmd/build/command.go
```diff
@@ -60,14 +65,15 @@ func Build(ctx context.Context) *cobra.Command {
...
 
-			builder, err := bc.getBuilder(options)
-			if err != nil {
+			if err := bc.loadContext(ctx, options); err != nil {
 				return err
 			}
 
-			if err := builder.LoadContext(ctx, options); err != nil {
+			builder, err := bc.getBuilder(options)
+			if err != nil {
```
ifbyol, line 73: Can we include a comment explaining why this order is important? This is assuming a hidden knowledge (`getBuilder` is reading the manifest and for that, you need to read the manifest first).

I think we need to improve a little bit this command to make it more readable and less error prone. Reading the manifest should be part of the main execution flow and then, pass it to the `getBuilder` function. I think that would make a bit clearer how the command works

cmd/build/command.go
teresaromero: i think this is to check the args in the case the file is passed as an option, we only accept 1 context 

`okteto build api -f Dockerfile` - this is allowed
`okteto build api frontend -f Dockerfile` i think this is not 

maybe i'm wrong... but i recall something regarding args validation was done inside command and i dont find now where was this 🤷🏻‍♀️

ifbyol: But is it related to the the issue we are fixing?

ifbyol:
Would it makes sense to create an e2e tests for this? (it could be in a follow up PR to unblock the fix)

cmd/build/command.go
AdrianPedriza: I agree to move `GetManifest` to that level. I we do that then I see two different options:
- send the error generated in `GetManifest` as a param to `getBuilder` 
- extract logic in `getBuilder` to main execution flow and remove that func.

WDYT? @ifbyol

cmd/build/command.go
ifbyol: but then, you are changing the behavior for V2, aren't you?

AdrianPedriza: no because the logic it is only executed when we face a `manifest v1`, for that reason this _if condition_, to only show that error if it is a `manifest v1`

AdrianPedriza:
> Would it makes sense to create an e2e tests for this? (it could be in a follow up PR to unblock the fix)

https://github.com/okteto/okteto/issues/3056

cmd/build/command.go
ifbyol: Why this condition means only v1?

cmd/build/command.go
AdrianPedriza: https://github.com/okteto/okteto/issues/3058

cmd/build/command.go
AdrianPedriza: from what I understood if -f == Dockerfile then we face a manifest v1. But that condition doesn't mean that we are going to catch all manifest v1 with it so I'm going to check if it is needed throw and error using the builder type instead of the current condition. I wanted to throw an error before call `loadContext` as previously implementation but at the moment it is not possible without a refactor of the command so it will be checked before calling `loadContext`.

ifbyol: yeah, my concern was about the implicit knowledge we need to have to see that `-f == Dockerfile then we face a manifest v1.` when, in reality, we instantiate the builder based on the manifest parsed, not in the parameters received by the command.

cmd/build/command.go
```diff
@@ -129,3 +150,38 @@ func validateDockerfile(file string) error {
 	_, _, err = instructions.Parse(parsedDockerfile.AST)
 	return err
 }
+
+func (*Command) loadContext(ctx context.Context, options *types.BuildOptions) error {
+	ctxOpts := &contextCMD.ContextOptions{}
```
ifbyol, line 155: The implementation of this function is slightly different between both versions (v1 and v2) and it seems we just got the v1 implementation, not sure this is the right thing to do, we might be breaking some cases. We need to review it carefully

AdrianPedriza: already talked about that with @ifbyol. if you're curious, take a look at how the function looked like after that.

cmd/build/command.go
AdrianPedriza: comments have been added on everything that requires a deeper knowledge that a priori is not available by reading the command.

cmd/build/command.go
```diff
@@ -129,3 +149,43 @@ func validateDockerfile(file string) error {
...
+	// before calling the context command, there is need to retrieve the context and
+	// namespace through the given manifest. If the manifest is a Dockerfile, this
+	// information cannot be extracted so call to GetContextResource is skkiped.
+	if err := validateDockerfile(options.File); err != nil {
+		ctxResource, err := model.GetContextResource(options.File)
+		if err != nil {
+			return err
+		}
+
+		if options.Namespace != "" {
```
jLopezbarb, line 165: Nit: We need to do some validation between `options.Namespace` and the ones retrieved from the manifest. If we don't use [LoadContextFromPath](https://github.com/okteto/okteto/blob/3636552ee541450e82c1904ba58a097b839e75c6/cmd/context/utils.go#L231), can we update `ctxResource` with
```golang
if err := ctxResource.UpdateNamespace(namespace); err != nil {
    return err
}

if err := ctxResource.UpdateContext(k8sContext); err != nil {
    return err
}
```
and after that set the options with the content of `ctxResource`.

Those validations prevents the user to have a different context specified on the flag and the manifest

cmd/build/command.go
```diff
@@ -60,14 +65,29 @@ func Build(ctx context.Context) *cobra.Command {
...
 			builder, err := bc.getBuilder(options)
 			if err != nil {
 				return err
 			}
 
-			if err := builder.LoadContext(ctx, options); err != nil {
-				return err
+			if _, ok := builder.(*buildv1.OktetoBuilder); ok {
+				maxV1Args := 1
+				docsURL := "https://okteto.com/docs/reference/cli/#build"
```
jLopezbarb, line 82: Move it as constant

cmd/build/command.go
```diff
@@ -60,14 +65,29 @@ func Build(ctx context.Context) *cobra.Command {
...
+
 			builder, err := bc.getBuilder(options)
 			if err != nil {
 				return err
 			}
 
-			if err := builder.LoadContext(ctx, options); err != nil {
-				return err
+			if _, ok := builder.(*buildv1.OktetoBuilder); ok {
+				maxV1Args := 1
```
jLopezbarb, line 81: Move it as constant

cmd/build/command.go
AdrianPedriza: if it is used only once I think we could add both values directly to the error string, wdyt?

cmd/build/command.go
```diff
@@ -129,3 +149,43 @@ func validateDockerfile(file string) error {
...
+		ctxResource, err := model.GetContextResource(options.File)
+		if err != nil {
+			return err
+		}
+
+		if options.Namespace != "" {
+			ctxOpts.Namespace = ctxResource.Namespace
+		}
+
+		if options.Namespace != "" {
```
AdrianPedriza, line 169: error. `options.Context`

cmd/build/command.go
jLopezbarb: I don't think it should be on the error string, changing the docs url is quite difficult now that we have them all as variables with the suffix `URL` so having them inside the string can make it harder when updating docs

AdrianPedriza: yes, agree

cmd/build/command.go
```diff
@@ -60,14 +70,27 @@ func Build(ctx context.Context) *cobra.Command {
...
+			}
+
 			builder, err := bc.getBuilder(options)
 			if err != nil {
 				return err
 			}
 
-			if err := builder.LoadContext(ctx, options); err != nil {
-				return err
+			if _, ok := builder.(*buildv1.OktetoBuilder); ok {
```
jLopezbarb, line 85: NIT: Could we encapsulate this into a interface function `isV1()`

cmd/build/command.go
```diff
@@ -60,14 +70,27 @@ func Build(ctx context.Context) *cobra.Command {
 			options.CommandArgs = args
 			bc := NewBuildCommand()
 
+			// The context must be loaded before the read manifest. Otherwise,
```
jLopezbarb, line 73: NIT:
```suggestion
			// The context must be loaded before reading manifest. Otherwise,
```

jLopezbarb (approved):

teresaromero (approved):

PR #3053, opened 2022-09-01
Okteto down and exec selector only active options

### LINKED ISSUE #2946: Okteto down selector should only show activated dev environment.
**Describe the bug**
If I have 10 dev environments and I'm currently with two of them in development mode (two services `okteto up`) I'd expect that `okteto down` selector only shows the dev environments that are now on development mode and hide the other ones

**To Reproduce**

1. Create the following `docker-compose.yml`
``` yaml
services:
  svc1:
    image: busybox
    volumes:
      - .:/usr/src
  svc2:
    image: busybox
    volumes:
      - .:/usr/src
  svc3:
    image: busybox
    volumes:
      - .:/usr/src
```
2. Run `okteto up svc1`
3. Run `okteto up svc2` on another terminal
4. Run `okteto down` on another terminal
5. See that the selector has svc1, svc2, and svc3

**Expected behavior**
I'd expect that `okteto down` selector only shows the services that are on development mode.

- Version 2.4.2

teresaromero:
Resolves #2946 

## Proposed changes

In order to make the `okteto down` command selector only display the current containers that are in dev mode, the options passed to the selector had to be customise.
- Refactor `SelectDevFromManifest` so the options are passed as a list of strings
- Create new method for the Dev Manifest to get the list of keys from dev manifest
- Create new function under apps to retrieve the Deployments and Statefulsets that are in DevMode
- get options at `okteto down` from the active ones
- Added same logic to `okteto exec` in order to be able to select only the dev containers that are active

### DIFF
--- a/cmd/down.go
+++ b/cmd/down.go
@@ -83,7 +83,7 @@ func Down() *cobra.Command {
 						return err
 					}
 					selector := utils.NewOktetoSelector("Select which development container to deactivate:", "Development container")
-					dev, err = utils.SelectDevFromManifest(manifest, selector)
+					dev, err = utils.SelectDevFromManifest(manifest, selector, manifest.Dev.GetDevs())
 					if err != nil {
 						return err
 					}
--- a/cmd/down.go
+++ b/cmd/down.go
@@ -92,6 +92,7 @@ func Down() *cobra.Command {
 
 					if len(options) < 1 {
 						oktetoLog.Success("All development containers are deactivated")
+						return nil
 					}
 					dev, err = utils.SelectDevFromManifest(manifest, selector, options)
 					if err != nil {
--- a/cmd/down.go
+++ b/cmd/down.go
@@ -90,7 +90,7 @@ func Down() *cobra.Command {
 					selector := utils.NewOktetoSelector("Select which development container to deactivate:", "Development container")
 					options := apps.ListDevModeOn(ctx, manifest, c)
 
-					if len(options) < 1 {
+					if len(options) == 0 {
 						oktetoLog.Success("All development containers are deactivated")
 						return nil
 					}
--- a/cmd/down.go
+++ b/cmd/down.go
@@ -88,7 +88,14 @@ func Down() *cobra.Command {
 						return err
 					}
 					selector := utils.NewOktetoSelector("Select which development container to deactivate:", "Development container")
-					dev, err = utils.SelectDevFromManifest(manifest, selector, manifest.Dev.GetDevs())
+					options, err := apps.ListDevModeOn(ctx, manifest.Namespace, c)
+					if err != nil {
+						return err
+					}
+					if len(options) < 1 {
+						oktetoLog.Success("All development containers are deactivated")
+					}
+					dev, err = utils.SelectDevFromManifest(manifest, selector, options)
 					if err != nil {
 						return err
 					}
--- a/cmd/down.go
+++ b/cmd/down.go
@@ -88,10 +88,8 @@ func Down() *cobra.Command {
 						return err
 					}
 					selector := utils.NewOktetoSelector("Select which development container to deactivate:", "Development container")
-					options, err := apps.ListDevModeOn(ctx, manifest.Namespace, c)
-					if err != nil {
-						return err
-					}
+					options := apps.ListDevModeOn(ctx, manifest, c)
+
 					if len(options) < 1 {
 						oktetoLog.Success("All development containers are deactivated")
 					}
--- a/cmd/down.go
+++ b/cmd/down.go
@@ -64,6 +64,11 @@ func Down() *cobra.Command {
 				return err
 			}
 
+			c, _, err := okteto.GetK8sClient()
+			if err != nil {
+				return err
+			}
+
 			if all {
 				err := allDown(ctx, manifest, rm)
 				if err != nil {
@@ -87,11 +92,6 @@ func Down() *cobra.Command {
 					if err != nil {
 						return err
 					}
-				}
-
-				c, _, err := okteto.GetK8sClient()
-				if err != nil {
-					return err
 				}
 
 				app, _, err := utils.GetApp(ctx, dev, c, false)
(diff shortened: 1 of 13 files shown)


pkg/k8s/apps/crud.go
```diff
@@ -165,3 +165,26 @@ func TranslateDevMode(trMap map[string]*Translation) error {
 	}
 	return nil
 }
+
+// ListDevModeOn returns a list of strings with the names of deployments or statefulsets in DevMode
+func ListDevModeOn(ctx context.Context, namespace string, c kubernetes.Interface) ([]string, error) {
+	devModeApps := make([]string, 0)
+
+	deploymentsDev, err := deployments.List(ctx, namespace, model.DevLabel, c)
```
ifbyol, line 173: Shouldn't the label selector be `fmt.Sprintf("%s=true", model.DevLabel)`? Same for statefulsets

pkg/model/manifest.go
```diff
@@ -1178,3 +1178,12 @@ func (d ManifestDevs) HasDev(name string) bool {
 	_, ok := d[name]
 	return ok
 }
+
+// GetDevs returns a list of strings with the keys of devs defined
+func (d ManifestDevs) GetDevs() []string {
+	devs := make([]string, 0)
```
ifbyol, line 1184: Minor thing. If you know the length of the slice (in this case would be `len(d)`), it is better to create the slice with that size

pkg/k8s/apps/crud.go
teresaromero: I tested it and when the component was in dev mode the label appeared alone without "true" or "false", but i guess is better to actively look for the true ones 👍🏻

cmd/exec.go
```diff
@@ -203,7 +203,7 @@ func getDevFromArgs(manifest *model.Manifest, args []string) (*model.Dev, error)
 			return nil, err
 		}
 		selector := utils.NewOktetoSelector("Select which development container to exec:", "Development container")
-		dev, err = utils.SelectDevFromManifest(manifest, selector)
```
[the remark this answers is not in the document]
teresaromero: This is a good point, although the issue is about the down cmd i feel is relevant to add this to the PR. I will update the Pr description. thnks

pkg/k8s/apps/crud.go
```diff
@@ -165,3 +165,26 @@ func TranslateDevMode(trMap map[string]*Translation) error {
 	}
 	return nil
 }
+
+// ListDevModeOn returns a list of strings with the names of deployments or statefulsets in DevMode
+func ListDevModeOn(ctx context.Context, namespace string, c kubernetes.Interface) ([]string, error) {
```
jLopezbarb, line 170: Could we test this function?

cmd/down.go
```diff
@@ -83,15 +88,17 @@ func Down() *cobra.Command {
...
 					selector := utils.NewOktetoSelector("Select which development container to deactivate:", "Development container")
-					dev, err = utils.SelectDevFromManifest(manifest, selector)
+					options, err := apps.ListDevModeOn(ctx, manifest.Namespace, c)
+					if err != nil {
+						return err
+					}
+					if len(options) < 1 {
+						oktetoLog.Success("All development containers are deactivated")
+					}
+					dev, err = utils.SelectDevFromManifest(manifest, selector, options)
```
jLopezbarb, line 98: Two things:
1.- We should show the name of the manifest instead of the name of the kubernetes resource because we add the suffix `-okteto` and that can cause some friction with the users
2.- I think there is an use case where this may not work. Imagine that I have two projects on the same namespace:
**okteto.yaml**
```yaml
name: a
image: alpine
command: sh
sync:
- .:/usr/src/app
forward:
- 9090:8080
autocreate: true
```
and **oktetov2.yml**
```yaml
dev:
  b:
    image: alpine
    command: sh
    sync:
    - .:/usr/src/app
    forward:
    - 7121:8080
    autocreate: true
  c: 
    image: alpine
    command: sh
    sync:
    - .:/usr/src/app
    forward:
    - 7121:8080
    autocreate: true
```
If I run `okteto up -f okteto.yml` and on another terminal okteto up -f `oktetov2.yml` when I run `okteto down -f oktetov2.yml` I should only receive the services defined on the `oktetov2.yml`, but I'm not seeing a filtering over the services that are on dev mode

pkg/k8s/apps/crud_test.go
```diff
@@ -266,3 +267,225 @@ func TestValidateMountPaths(t *testing.T) {
 	}
 
 }
+
+func TestListDevModeOn(t *testing.T) {
```
jLopezbarb, line 271: Can we test that we are not getting one dev container (a deployment in dev mode) that is not defined in the manifest?

cmd/down.go
```diff
@@ -83,17 +88,17 @@ func Down() *cobra.Command {
 						return err
 					}
 					selector := utils.NewOktetoSelector("Select which development container to deactivate:", "Development container")
-					dev, err = utils.SelectDevFromManifest(manifest, selector)
+					options := apps.ListDevModeOn(ctx, manifest, c)
+
+					if len(options) < 1 {
```
jLopezbarb, line 93: nit:
```suggestion
					if len(options) == 0 {
```

cmd/down.go
teresaromero: i've changed the function so it retrieves resources from dev manifest. The name will be always the one of the dev manifest as this resource is the one with the dev mode label. The additional resource created with `-okteto` is the dev mode resource which does not have the label.

also thnks to this example, i've added logic to be aware of autocreate resources, which will have the suffix

pkg/k8s/apps/crud_test.go
teresaromero: added `"dev-is-not-at-manifest"`

jLopezbarb (approved):

AdrianPedriza (approved):

pkg/model/manifest.go
teresaromero: i've finally transformed to `[]string{}` and append the keys

PR #3055, opened 2022-09-01
Sanitize uppercase srv names

### LINKED ISSUE #2639: I cannot use an image name in my `build` manifest section with uppercase letter
**Describe the bug**
If I have a build section like this:
```
build:
  Frontend:
    context: frontend
```
 and I run `okteto build` I get this error:

```
Error building image 'registry.demo.okteto.dev/pchico83/movies-Frontend:okteto': build failed: failed to solve: invalid reference format: repository name must be lowercase
```

**Expected behavior**
Since image names are transparent to the user using the `OKTETO_BUILD_` variables, I would sanitize and lowercase the image name

teresaromero:
# Proposed changes

Fixes #2639 

When services apper with uppercase at the okteto manifest or stack we need to sanitize this strings so they are lowercase while performing commands with the manifest.

- Added hasUppercase function to detect if the service name has uppercase
-  Refactored `shouldBeSanitized` to include this condition
- Refactored conditions on `shouldBeSanitized` for better understanding
- Refactored `sanitizeName` to apply this condition for all being lowecase
- Fix a found bug: when sanitizing the Build Manifest, the key deletion was being done under manifest.Dev instead of manifest.Build

### DIFF
--- a/pkg/model/manifest.go
+++ b/pkg/model/manifest.go
@@ -682,7 +682,7 @@ func (m *Manifest) SanitizeSvcNames() error {
 			}
 			m.Build[sanitizedSvcName] = build
 			sanitizedServicesNames[buildKey] = sanitizedSvcName
-			delete(m.Dev, buildKey)
+			delete(m.Build, buildKey)
 		}
 	}
 
--- a/pkg/model/stack_serializer.go
+++ b/pkg/model/stack_serializer.go
@@ -1199,12 +1199,13 @@ func hasUppercase(name string) bool {
 	return false
 }
 func shouldBeSanitized(name string) bool {
-	return strings.Contains(name, " ") || strings.Contains(name, "_")
+	return strings.Contains(name, " ") || strings.Contains(name, "_") || hasUppercase(name)
 }
 
 func sanitizeName(name string) string {
 	name = strings.ReplaceAll(name, " ", "-")
 	name = strings.ReplaceAll(name, "_", "-")
+	name = strings.ToLower(name)
 	return name
 }
 
--- a/pkg/model/stack_serializer.go
+++ b/pkg/model/stack_serializer.go
@@ -22,6 +22,7 @@ import (
 	"strconv"
 	"strings"
 	"time"
+	"unicode"
 
 	"github.com/kballard/go-shellquote"
 	"github.com/okteto/okteto/pkg/filesystem"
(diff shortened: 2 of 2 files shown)


pkg/model/manifest.go
```diff
@@ -682,7 +682,7 @@ func (m *Manifest) SanitizeSvcNames() error {
 			}
 			m.Build[sanitizedSvcName] = build
 			sanitizedServicesNames[buildKey] = sanitizedSvcName
-			delete(m.Dev, buildKey)
+			delete(m.Build, buildKey)
```
ifbyol, line 685: Can we include some tests to this function? Checking the code it shouldn't be hard, right? if it has already some test, can we include this scenario?

AdrianPedriza (approved):

pkg/model/manifest.go
```diff
@@ -686,11 +686,11 @@ func (m *Manifest) SanitizeSvcNames() error {
 		}
 	}
 
-	for _, gf := range m.GlobalForward {
-		if shouldBeSanitized(gf.ServiceName) {
-			sanitizedSvcName := sanitizeName(gf.ServiceName)
-			gf.ServiceName = sanitizedSvcName
-			sanitizedServicesNames[gf.ServiceName] = sanitizedSvcName
+	for indx := range m.GlobalForward {
```
jLopezbarb, line 689: Can we use `idx` as the index abbreviation, that keeps consistency between the rest of the code.

pkg/model/manifest.go
```diff
@@ -686,11 +686,11 @@ func (m *Manifest) SanitizeSvcNames() error {
...
+		if shouldBeSanitized(m.GlobalForward[indx].ServiceName) {
```
jLopezbarb, line 690: Could we extract `m.GlobalForward[indx].ServiceName` at the beginning of the for loop? I think it will look cleaner than having to read such a long variable name

jLopezbarb (approved):

pkg/model/stack_serializer.go
```diff
@@ -1208,8 +1208,12 @@ func hasUnderscore(name string) bool {
 	return strings.Contains(name, "_")
 }
 
+func hasTrailingSpaces(name string) bool {
+	return strings.HasPrefix(name, " ") || strings.HasSuffix(name, " ")
+}
+
 func shouldBeSanitized(name string) bool {
-	return hasEmptySpace(name) || hasUnderscore(name) || hasUppercase(name)
+	return hasEmptySpace(name) || hasUnderscore(name) || hasUppercase(name) || hasTrailingSpaces(name)
```
jLopezbarb, line 1216: Nice catch

AgustinRamiroDiaz: If I understood correctly, we wouldn't need to check `hasTrailingSpaces` if we are already checking `hasEmptySpace`. Can you confirm @teresaromero ?

teresaromero: that is correct! :) will upload the rever change

PR #3069, opened 2022-09-05
fix: visual bug

### LINKED ISSUE #2821: Minor output issue when selecting the current context again
**Describe the bug**
If you launch okteto context and realized that the current context is the one you wanted and you select it again, the output of the selected context shows the URL of the context but also the `*` symbol that was there to indicate the current context. The `*` shouldn't be there and is a bit confusing:

Screenshot 2022-06-21 at 00 29 09

- Version 2.4.0

jLopezbarb:

# Proposed changes

Fixes #2821

- Clean the name(remove the *) after selecting the  context

## Screenshots
Captura de pantalla 2022-09-05 a las 13 39 19

Captura de pantalla 2022-09-05 a las 13 39 27

### DIFF
--- a/cmd/utils/selector.go
+++ b/cmd/utils/selector.go
@@ -275,8 +275,7 @@ func (s *OktetoSelector) run(initialPosition int) (string, error) {
 		return "", err
 	}
 
-	items, idx := l.Items()
-	item := items[idx]
+	item := strings.TrimSpace(strings.TrimSuffix(s.Items[l.Index()].Name, " *"))
 
 	sb.Reset()
 	sb.Write(render(s.OktetoTemplates.selected, item))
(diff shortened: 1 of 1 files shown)


teresaromero (approved):

cmd/utils/selector.go
```diff
@@ -275,8 +275,7 @@ func (s *OktetoSelector) run(initialPosition int) (string, error) {
 		return "", err
 	}
 
-	items, idx := l.Items()
-	item := items[idx]
+	item := strings.TrimSpace(strings.TrimSuffix(s.Items[l.Index()].Name, " *"))
```
AgustinRamiroDiaz, line 278: Wouldn't it be better to not save it with the `*`? We could change the `run` function to not modify the selector Items

jLopezbarb: We need it to show what's the actual one, and we can't change the behavior of the function

PR #3073, opened 2022-09-07
Show pipeline deploy and destroy logs when running the command with `--wait` flag

### LINKED ISSUE #2794: `okteto pipeline deploy` and `okteto preview deploy` should stream the pipeline logs
Similar to `okteto deploy`, `okteto pipeline deploy` and `okteto preview deploy` should stream the logs of the deploy commands execution. This is specially frustrating when debugging GitHub preview actions, because the pipeline logs are lost on each deploy, and from an old GH action is impossible to troubleshoot why a github action failed.

teresaromero:
# Proposed changes

Fixes #2794 

When running `okteto pipeline deploy` or `okteto pipeline destroy` logs from the installer where only available at the okteto UI.
This change retrieves the installer logs from the `sse` service from the backend and prints them at the CLI when the user runs the command with the `--wait` flag for `pipeline deploy` and `pipeline destroy`


cmd/pipeline/deploy.go
```diff
@@ -234,8 +236,21 @@ func (pc *Command) waitUntilRunning(ctx context.Context, name string, action *ty
...
-	go func() {
+	var wg sync.WaitGroup
 
+	wg.Add(1)
+	go func(wg *sync.WaitGroup) {
+		defer wg.Done()
+		sseClient := sse.NewClient()
+		if err := sseClient.StreamPipelineLogs(name, action.Name); err != nil {
+			exit <- err
+			return
```
AgustinRamiroDiaz, line 247: nit: this `return` is not needed

cmd/pipeline/deploy.go
```diff
@@ -244,7 +259,9 @@ func (pc *Command) waitUntilRunning(ctx context.Context, name string, action *ty
 
 		oktetoLog.Spinner("Waiting for containers to be healthy...")
 		exit <- pc.waitForResourcesToBeRunning(ctx, name, timeout)
-	}()
+	}(&wg)
+
+	wg.Wait()
```
AgustinRamiroDiaz, line 264: Why are we waiting for the 2 goroutines to finish? 
This will change the behavior, since before we weren't waiting for the old goroutine

pkg/sse/client.go
```diff
@@ -0,0 +1,74 @@
...
+
+const (
+	// pipelineSSEUrl is the template for pipeline logs sse event subscription
+	pipelineSSEUrl = "%s/sse/logs/%s/gitdeploy/%s?action=%s"
+)
+
+// NewClient returns a new SSE client
+func NewClient() *Client {
+	return &Client{
+		http: http.Client{},
```
AgustinRamiroDiaz, line 27: Looking at the implementation, seems like more than an `http.Client`, this `Client` depends on a function to get the sse logs. I think it would be better to inject the function, since it will reduce the points of change in this code if some day the sse logs change

It will also make this more testable, and will simplify the function `StreamPipelineLogs`

pkg/sse/events.go
```diff
@@ -0,0 +1,24 @@
...
+	// dataHeader is the sse string header for sse-data
+	dataHeader = "data: "
+
+	// pingEvent is the type of event ping
+	pingEvent = "ping"
+)
+
+// getMessage returns the parsed message from the sse-data
+func getMessage(s string) string {
+	if strings.HasPrefix(s, dataHeader) && !strings.Contains(s, pingEvent) {
```
AgustinRamiroDiaz, line 15: Could a data event contain the word `"ping"`? Maybe it's a subword or word from the installed application

cmd/pipeline/deploy.go
jLopezbarb: I think we were waiting before on the select but right now we are waiting even if the user press `CTRL+C`

cmd/pipeline/deploy.go
```diff
@@ -234,8 +236,21 @@ func (pc *Command) waitUntilRunning(ctx context.Context, name string, action *ty
...
 	exit := make(chan error, 1)
 
-	go func() {
+	var wg sync.WaitGroup
 
+	wg.Add(1)
+	go func(wg *sync.WaitGroup) {
+		defer wg.Done()
+		sseClient := sse.NewClient()
+		if err := sseClient.StreamPipelineLogs(name, action.Name); err != nil {
```
jLopezbarb, line 245: This function should check if the context is cancelled to stop the gorotuine

cmd/pipeline/deploy.go
```diff
@@ -234,8 +236,21 @@ func (pc *Command) waitUntilRunning(ctx context.Context, name string, action *ty
 	signal.Notify(stop, os.Interrupt)
 	exit := make(chan error, 1)
 
-	go func() {
+	var wg sync.WaitGroup
```
jLopezbarb, line 239: Can we use a `context.WithCancel()` instead. We need to cancel it when pressing `Ctrl+C` in the middle of the execution

jLopezbarb:
Are you going to work on the same PR for `preview` and `destroy` commands? If so I think you'll need to update the PR title

pkg/sse/client.go
jLopezbarb: Since this is calling the `oktetoClient` which we already have mocked. Can we move this code to the oktetoClient and make a new function `oktetoClient().Sse().Pipeline()` or `oktetoClient().Pipeline().Sse()`

cmd/pipeline/deploy.go
teresaromero: we were waiting before because there where the channels in place to listen to the events of exit and stop. indeed :) is the wait implementation for this flag of the pipeline :) i will take a look and give an other thought on this. thnks!

cmd/pipeline/deploy.go
maroshii, line 239: If I'm reading this right, this leaks a go routine if an error is returned.

We have `exit := make(chan error, 1)` which has a buffer of 1. This means that it can hold 1 message without blocking. Before this change it was fine because we sent only one error from either `waitToBeDeployed` or `waitForResourcesToBeRunning` (lines 241 and 246).

With this change, we are sending two messages to the exit channel so the second one will block until we have a receiver which never happens.

Also, it looks like we are doing a `Wait()` right before selecting on `<-stop` so the CTRL+C will probably not work.

In general we should make better use of the context and use it as a cascading cancellation method. I don't think we should even call `signal.Notify(stop, os.Interrupt)` here and just use the context and cancel everything when done.

Ideally, we should only have a single top level `signal.Notify(stop, os.Interrupt)` at the root command and cascade a context cancellation. Not in scope for this PR though.

Created this gist as a reference implementation in case it helps and you wanna take a look: https://gist.github.com/maroshii/0a4e0f402d1d902e5491b3eaba108ff9. It has both a context cancellation method and the signal the way it is here.

It might require changes to the other functions to receive a context and have proper cancellation though

pkg/sse/client.go
maroshii: I think what you are referring to is injecting the `io.Reader`. If this client received the io.Reader it would simply transform the response from SSE to the `eventLog`.

The issue I see with a single `Client` is that everything related to SSE will be added to it and we'll end up with a "God class".

We could have on one side the Fetcher/Reader that reads from SSE and then the Transformer which reads from the fetcher and transforms the response (line by line).

This architecture could be later extended to other log resources, events, etc.

Just thinking out loud here, no need to do it now.

cmd/pipeline/deploy.go
```diff
@@ -234,8 +236,21 @@ func (pc *Command) waitUntilRunning(ctx context.Context, name string, action *ty
...
 
+	wg.Add(1)
+	go func(wg *sync.WaitGroup) {
+		defer wg.Done()
+		sseClient := sse.NewClient()
```
maroshii, line 244: I wouldn't instantiate the sse here. We could add an interface to the command so that we inject a `PipelineLogger` and the sse client simply satisfies that interface

rberrelleza:
Have we considered requiring a "--show-logs" (or something similar) parameter so we don't change the existing behavior of the command?

pkg/sse/client.go
teresaromero: yup, i've moved the Stream function inside the Pipeline client. here we have to parse the sse event which comes line by line, one line `id`, other `event` and other `data`. we are parsing `data` to be a log object and displaying it by log.Println()

pkg/sse/events.go
teresaromero: i've changed the parsing and checking of the message, this was not necessary after all, we nevermind ping events :D

pkg/sse/events.go
```diff
@@ -0,0 +1,24 @@
...
+		return strings.TrimPrefix(s, dataHeader)
+	}
+	return ""
+}
+
+// isDone returns bool if the event msg is EOF
+func isDone(msg string) bool {
```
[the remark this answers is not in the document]
teresaromero: I've removed this function and now once the event is unmarshalled, i check the mssage is EOF and Stage is done

teresaromero:
> Have we considered requiring a "--show-logs" (or something similar) parameter so we don't change the existing behavior of the command?

@rberrelleza  yes, i thought about introducing the logs after a flag, but we decided to add them if the flag is `--wait`
https://github.com/okteto/okteto/issues/2794#issuecomment-1235305660

cmd/pipeline/deploy.go
teresaromero: added the client the the pipelinclient

jLopezbarb:
> Have we considered requiring a "--show-logs" (or something similar) parameter so we don't change the existing behavior of the command?

I don't think so, this will only be shown with `--wait` flag. Right now we are only showing a spinner saying `waiting to be deployed...`.
If we finally go with `--show-logs` we must force the user to use the it only if `--wait` is specified, removing the ability to run `okteto pipeline deploy --show-logs` since it can cause confusion because it's not going to show any logs if the `--wait` is not added

jLopezbarb (approved):

AdrianPedriza (approved):

PR #3075, opened 2022-09-07
feat: add insecure skip tls verify

### LINKED ISSUE #3059: Support insecure connections to Okteto clusters
Support establishing connections to Okteto clusters with certificates not trusted by the default OS ssl store.

A flag and/or an environment variable to enable/disable ssl store, similar to `--insecure-skip-tls-verify` in kubectl.

Untrusted SSL certificate on login could be saved to context file for later strict ssl usage.

**Additional context**
Future Okteto chart releases will include self signed certificates bootstrapping for quicker demo deployments. This ticket ensures the CLI can connect to these clusters without modifying your OS ssl store.

jpf-okteto:
Resolve #3059.

### DIFF
--- a/cmd/context/context.go
+++ b/cmd/context/context.go
@@ -21,7 +21,6 @@ import (
 
 // Context points okteto to a cluster.
 func Context() *cobra.Command {
-	var insecureSkipTlsVerify bool
 	ctxOptions := &ContextOptions{}
 	cmd := &cobra.Command{
 		Use:     "context",
@@ -40,7 +39,7 @@ To set your default context, run the ` + "`okteto context`" + ` command:
 This will prompt you to select one of your existing contexts or to create a new one.
 `,
 		PersistentPreRun: func(cmd *cobra.Command, args []string) {
-			okteto.SetInsecureSkipTLSVerifyPolicy(insecureSkipTlsVerify)
+			okteto.SetInsecureSkipTLSVerifyPolicy(ctxOptions.InsecureSkipTlsVerify)
 		},
 		RunE: Use().RunE,
 	}
@@ -54,7 +53,7 @@ This will prompt you to select one of your existing contexts or to create a new 
 	cmd.AddCommand(UpdateKubeconfigCMD())
 	cmd.AddCommand(UseNamespace())
 
-	cmd.PersistentFlags().BoolVarP(&insecureSkipTlsVerify, "insecure-skip-tls-verify", "", false, " If enabled, the server's certificate will not be checked for validity. This will make your connections insecure")
+	cmd.PersistentFlags().BoolVarP(&ctxOptions.InsecureSkipTlsVerify, "insecure-skip-tls-verify", "", false, " If enabled, the server's certificate will not be checked for validity. This will make your connections insecure")
 	cmd.Flags().StringVarP(&ctxOptions.Token, "token", "t", "", "API token for authentication")
 	cmd.Flags().StringVarP(&ctxOptions.Namespace, "namespace", "n", "", "namespace of your okteto context")
 	cmd.Flags().StringVarP(&ctxOptions.Builder, "builder", "b", "", "url of the builder service")
--- a/pkg/cmd/login/login.go
+++ b/pkg/cmd/login/login.go
@@ -47,10 +47,10 @@ func (*LoginController) AuthenticateToOktetoCluster(ctx context.Context, oktetoU
 		oktetoLog.Infof("authenticating with browser code")
 		user, err := WithBrowser(ctx, oktetoURL)
 		// If there is a TLS error, return the raw error
-		if err != nil && oktetoErrors.IsX509(err) {
+		if oktetoErrors.IsX509(err) {
 			return nil, oktetoErrors.UserError{
 				E:    err,
-				Hint: "Add the flag '--insecure-skip-tls-verify' to connect to an instance with self-signed certificates",
+				Hint: oktetoErrors.ErrX509Hint,
 			}
 		}
 		if err != nil {
--- a/pkg/okteto/client.go
+++ b/pkg/okteto/client.go
@@ -194,10 +194,10 @@ func translateAPIErr(err error) error {
 		return fmt.Errorf("unauthorized. Please run 'okteto context url' and try again")
 
 	default:
-		if strings.Contains(err.Error(), "x509") {
+		if oktetoErrors.IsX509(err) {
 			return oktetoErrors.UserError{
 				E:    err,
-				Hint: "Add the flag '--insecure-skip-tls-verify' to connect to an instance with self-signed certificates",
+				Hint: oktetoErrors.ErrX509Hint,
 			}
 		}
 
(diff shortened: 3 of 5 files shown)


pkg/okteto/client.go
```diff
@@ -96,11 +102,26 @@ func NewOktetoClientFromUrl(url string) (*OktetoClient, error) {
...
 	}
 
-	httpClient := oauth2.NewClient(context.Background(), nil)
+	ctx := backgroundContextWithHttpClient(config.IsInsecureSkipVerify())
+	httpClient := oauth2.NewClient(ctx, nil)
 
 	return newOktetoClientFromGraphqlClient(u, httpClient)
 }
 
+func backgroundContextWithHttpClient(insecureSkipVerify bool) context.Context {
```
maroshii, line 111: I would add a comment here as to why we are setting the client in the context as per [the docs](https://cs.opensource.google/go/x/oauth2/+/0ebed06d:oauth2.go;l=332-341). It wasn't immediately obvious without drilling into the oauth2 package code. Specially since if we add a token source at some point it won't be used

pkg/config/config.go
```diff
@@ -281,3 +282,16 @@ func GetDeployOrigin() (src string) {
 func RunningInInstaller() bool {
 	return os.Getenv(model.OktetoInInstaller) == "true"
 }
+
+func IsInsecureSkipVerify() bool {
+	envVarValue := os.Getenv("OKTETO_INSECURE_SKIP_TLS_VERIFY")
```
jLopezbarb, line 287: Should we use a constant for the variable? In case we decide to change we have to change it only in one place

teresaromero: i think we have it only here, but i agree maybe adding this as we have the other envs at the model pkg, but make it constant of the config pkg ?

pkg/okteto/client.go
```diff
@@ -96,11 +102,26 @@ func NewOktetoClientFromUrl(url string) (*OktetoClient, error) {
...
+	return context.WithValue(
+		context.Background(),
+		oauth2.HTTPClient,
+		&http.Client{
+			Transport: &http.Transport{
+				TLSClientConfig: &tls.Config{
+					InsecureSkipVerify: insecureSkipVerify,
```
teresaromero, line 118:
```suggestion
					InsecureSkipVerify: insecureSkipVerify, // skipcq: GSC-G402
```

can we add this so deepsource does not take this as a flag?

pkg/okteto/client.go
```diff
@@ -96,11 +102,26 @@ func NewOktetoClientFromUrl(url string) (*OktetoClient, error) {
...
 	return newOktetoClientFromGraphqlClient(u, httpClient)
 }
 
+func backgroundContextWithHttpClient(insecureSkipVerify bool) context.Context {
+	return context.WithValue(
+		context.Background(),
+		oauth2.HTTPClient,
+		&http.Client{
+			Transport: &http.Transport{
+				TLSClientConfig: &tls.Config{
```
teresaromero, line 117:
```suggestion
				TLSClientConfig: &tls.Config{ // skipcq: 
GO-S1020
```

here also, i havent used these rules, but according to docs we should add these to avoid false positives https://deepsource.io/docs/setup-analysis#universality

pchico83:
@jpf-okteto can we handle this case in this PR too?
https://github.com/okteto/okteto/issues/3124

jLopezbarb:
> @jpf-okteto can we handle this case in this PR too?
> #3124

Part of #3124 is handled by the PR #3125. Since #3125 fixes the issue for the temporal kubeconfig, can this PR only take care of the okteto and k8sClient?

jLopezbarb (dismissed):

pkg/config/config.go
```diff
@@ -283,15 +290,25 @@ func RunningInInstaller() bool {
...
 	}
 
-	oktetoLog.Debugf("insecure mode: %v", value)
-	return value
+	if insecureSkipVerify {
+		insecureSkipVerifyOnce.Do(func() { oktetoLog.Warning("Insecure mode enabled") })
+		return true
+	}
+
+	oktetoLog.Debug("insecure mode: %v", insecureSkipVerify)
```
AgustinRamiroDiaz, line 310: I think we should move this comment above the `if` before so we always get the log

pkg/cmd/login/login.go
```diff
@@ -46,6 +46,13 @@ func (*LoginController) AuthenticateToOktetoCluster(ctx context.Context, oktetoU
 	if token == "" {
 		oktetoLog.Infof("authenticating with browser code")
 		user, err := WithBrowser(ctx, oktetoURL)
+		// If there is a TLS error, return the raw error
+		if err != nil && oktetoErrors.IsX509(err) {
+			return nil, oktetoErrors.UserError{
+				E:    err,
+				Hint: "Add the flag \"--insecure-skip-tls-verify\" to connect to an instance with self-signed certificates",
```
pchico83, line 53:
```suggestion
				Hint: "Add the flag '--insecure-skip-tls-verify' to connect to an instance with self-signed certificates",
```
I think we are using single quotes in other info messages

pkg/okteto/client.go
```diff
@@ -148,6 +194,13 @@ func translateAPIErr(err error) error {
 		return fmt.Errorf("unauthorized. Please run 'okteto context url' and try again")
 
 	default:
+		if strings.Contains(err.Error(), "x509") {
+			return oktetoErrors.UserError{
+				E:    err,
+				Hint: "Add the flag \"--insecure-skip-tls-verify\" to connect to an instance with self-signed certificates",
```
pchico83, line 200:
```suggestion
				Hint: "Add the flag '--insecure-skip-tls-verify' to connect to an instance with self-signed certificates",
```

pchico83 (approved):

cmd/context/context.go
```diff
@@ -15,11 +15,13 @@ package context
 
 import (
 	"github.com/okteto/okteto/cmd/utils"
+	"github.com/okteto/okteto/pkg/okteto"
 	"github.com/spf13/cobra"
 )
 
 // Context points okteto to a cluster.
 func Context() *cobra.Command {
+	var insecureSkipTlsVerify bool
```
jLopezbarb, line 24: Can we add `insecureSkipTlsVerify` as a `ctxOptions` field?

pkg/cmd/login/login.go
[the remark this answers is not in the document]
rberrelleza: I agree with @jpf-okteto .  Self-Signed is not the only reason why you are going to get that error. It also can be for other reasons like: expired certs, change of subdomain name, use of TLS termination services (like Netscaler in Simon), use of a bastion-like service, man in the middle attacks, etc.   

Do we have a way to verify that it is indeed due to self-signed? If so, then this hint makes sense. If not, we should send them to our docs on how to troubleshoot this. 

If using self-signed certs is problematic, I don't think a hint in an error is the right place to fix it.  We should point our users to a doc on how install the cert in a trust store, or to use valid certs.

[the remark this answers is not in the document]
pchico83: https://github.com/okteto/okteto/pull/3154

PR #3077, opened 2022-09-07
Follow-up-refactor-from-2795

### LINKED ISSUE #3076: Follow up refactor from 2795
_Originally posted by @AdrianPedriza in https://github.com/okteto/okteto/pull/3057#discussion_r965035068_

AgustinRamiroDiaz:
Fixes #3076


AgustinRamiroDiaz:
Regarding the `data` package (name yet to be discussed), @jLopezbarb has told me his concerned about it being like the `utils` package which we are trying to deprecate.

I feel like there's a difference between this two packages:
- `utils` is a mix between all the code that had been written many times, with the objective of DRY code. This evolved into a very big package without a clear path in sight, which led to it becoming hard to manage
- `data` package doesn't aim to reduce our common code duplications we get inside okteto. `data` aims to give us the tools to get more readable code that `go` doesn't have by "design". `data` helps by abstracting the common patterns we encounter everyday, like writing a dozen of times the conversions between sets and slices

teresaromero:
can we name this more descriptive? what is the goal of the refactor? what is the scope? :) thnks

PR #3085, opened 2022-09-09
deprecate manifestv1

### LINKED ISSUE #3066: Deprecate okteto manifest v1
Okteto CLI 2.0 support the [okteto manifest v2](https://www.okteto.com/docs/reference/manifest/) which supersedes the functionality of the okteto manifest v1. Most of our users have already migrated to the manifest v2, and the manifest v1 isn't documented in the [okteto docs](https://github.com/okteto/docs) since June 2002.

Show a warning when `okteto up` is running versus an okteto manifest v1. The warning show indicate that support for manifest v1 will be removed in Okteto CLI 3.0. Also show an information message with a link to the migrating from manifest [v1 to v2 tuturial](https://www.okteto.com/docs/reference/manifest-migration/).

**Additional context**
Add a metric to track, on each `okteto up` event, if the user is using manifest v1, v2 or a plain compose file.

jLopezbarb:
# Proposed changes

Fixes #3066 

- Include a warning and an event to track how many people are using manifest v1 on the up command

## Screenshot

Captura de Pantalla 2022-09-09 a las 13 42 53

### DIFF
--- a/cmd/up/up.go
+++ b/cmd/up/up.go
@@ -321,6 +321,11 @@ func Up() *cobra.Command {
 				upOptions.Deploy = true
 			}
 
+			if up.Manifest.Type == model.OktetoManifestType && !up.Manifest.IsV2 {
+				oktetoLog.Warning("okteto manifest v1 is deprecated and will be removed in okteto 3.0")
+				oktetoLog.Println(fmt.Sprintf("    %s", oktetoLog.BlueString("Consider upgrading to the new manifest schema: https://www.okteto.com/docs/reference/manifest-migration/")))
+			}
+
 			err = up.start()
 
 			if err != nil {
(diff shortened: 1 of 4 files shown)


cmd/up/up.go
```diff
@@ -321,6 +321,11 @@ func Up() *cobra.Command {
 				upOptions.Deploy = true
 			}
 
+			if up.Manifest.Type == model.OktetoManifestType && !up.Manifest.IsV2 {
+				oktetoLog.Warning("okteto manifest v1 is deprecated and will be removed in okteto 3.0")
```
jLopezbarb, line 325: @RinkiyaKeDad Could you check the warning message? Thanks in advance

teresaromero (approved):
LGTM, once the copy for the message is revised

rberrelleza:
@pchico83 @jLopezbarb is there anything in the v1 manifest that it's not supported on the v2 manifest ?

cmd/up/up.go
```diff
@@ -321,6 +321,11 @@ func Up() *cobra.Command {
...
+				oktetoLog.Println(fmt.Sprintf("    %s", oktetoLog.BlueString("Consider upgrading to the new manifest schema: https://www.okteto.com/docs/reference/manifest-migration/")))
```
rberrelleza, line 326: I think we need stronger language since people won't have an option soonish.

pchico83: What about?
```
Follow this guide to upgrade to the new okteto manifest schema:
https://www.okteto.com/docs/reference/manifest-migration/
```

jLopezbarb:
> @pchico83 @jLopezbarb is there anything in the v1 manifest that it's not supported on the v2 manifest?

Everything that could be done in manifest v1 can be done in manifest v2. For example, we have moved the build section from dev(manifest v1) to the top-level build section(manifest v2) but it can still be referenced in the dev section (manifest v2)

PR #3088, opened 2022-09-09
feat: add keep annotation

### LINKED ISSUE #3082: Support an annotation to persist resources on `okteto destroy`
I have deployed my dev environment, and I want to make sure some resources are not deleted on `okteto destroy`.

Tools like Helm use the annotation `helm.sh/resource-policy": keep` to achieve this scenario. We should do the same and support the annotation `dev.okteto.com/policy": keep`

**Additional context**
This is very useful when using a global namespace shared by many developers, to make sure things are not deleted by mistake

jLopezbarb:

# Proposed changes

Fixes #3082

- Skip the deletion of the volume if it has `dev.okteto.com/policy` set to `keep`.

## Concerns

The annotation is just working on volumes, do we need to extend the behavior to every resource(if  `dev.okteto.com/policy` set to `keep` we shouldn't delete the resource)
cc: @pchico83

### DIFF
--- a/integration/deploy/compose_test.go
+++ b/integration/deploy/compose_test.go
@@ -41,6 +41,8 @@ const (
     ports:
       - 8080
       - 8913
+	labels:
+	  dev.okteto.com/policy: keep
   nginx:
     image: nginx
     volumes:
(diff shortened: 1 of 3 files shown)


jLopezbarb:
Pending to create `DestroyWithLabel` tests

pchico83:
@jLopezbarb yes, the annotation must work for any kind of resource, not just volumes

pchico83 (approved):

jLopezbarb:
> Pending to create `DestroyWithLabel` tests

We can't unit test the function because we can't inject the fake k8s client and mocking every http call to k8s is not worth it. Created: https://github.com/okteto/okteto/issues/3091 to remove the dependency and create a proper unit test

AdrianPedriza (approved):

teresaromero (approved):

PR #3090, opened 2022-09-12
Added BuildSecrets to build manifest

### LINKED ISSUE #2566: Support for build secrets
Build secrets are now part of the compose spec:
https://github.com/compose-spec/compose-spec/pull/238/files

teresaromero:
# Proposed changes

Fixes #2566 

Secrets was supported via `okteto build` flag, now secrets can be added to the okteto manifest under the build section

### DIFF
--- a/pkg/cmd/build/build.go
+++ b/pkg/cmd/build/build.go
@@ -273,7 +273,7 @@ func OptsFromBuildInfo(manifestName, svcName string, b *model.BuildInfo, o *type
 
 	if len(b.Secrets) > 0 {
 		if o.Secrets == nil {
-			o.Secrets = make([]string, 0)
+			o.Secrets = []string{}
 		}
 		// add to the build the secrets at the manifest
 		for id, src := range b.Secrets {
--- a/pkg/cmd/build/build.go
+++ b/pkg/cmd/build/build.go
@@ -271,16 +271,6 @@ func OptsFromBuildInfo(manifestName, svcName string, b *model.BuildInfo, o *type
 		}
 	}
 
-	if len(b.Secrets) > 0 {
-		if o.Secrets == nil {
-			o.Secrets = []string{}
-		}
-		// add to the build the secrets at the manifest
-		for id, src := range b.Secrets {
-			o.Secrets = append(o.Secrets, fmt.Sprintf("id=%s,src=%s", id, src))
-		}
-	}
-
 	opts := &types.BuildOptions{
 		CacheFrom: b.CacheFrom,
 		Target:    b.Target,
@@ -290,6 +280,11 @@ func OptsFromBuildInfo(manifestName, svcName string, b *model.BuildInfo, o *type
 		BuildArgs: model.SerializeBuildArgs(b.Args),
 		NoCache:   o.NoCache,
 		Secrets:   o.Secrets,
+	}
+
+	// add to the build the secrets at the manifest
+	for id, src := range b.Secrets {
+		opts.Secrets = append(opts.Secrets, fmt.Sprintf("id=%s,src=%s", id, src))
 	}
 
 	outputMode := oktetoLog.GetOutputFormat()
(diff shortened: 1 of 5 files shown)


AgustinRamiroDiaz:
Do we currently support this secrets on the docker compose manifest?

pkg/model/serializer.go
```diff
@@ -46,6 +46,7 @@ type buildInfoRaw struct {
 	VolumesToInclude []StackVolume  `yaml:"-"`
 	ExportCache      string         `yaml:"export_cache,omitempty"`
 	DependsOn        BuildDependsOn `yaml:"depends_on,omitempty"`
+	Secrets          BuildSecrets   `yaml:"secrets,omitempty"`
```
AgustinRamiroDiaz, line 49: Can we add a comment to this struct like

"it's an auxiliary struct that should match `BuildInfo` and it's used to handle recursion in custom un/marshalling"

So we understand why this struct has to change when `BuildInfo` changes

pkg/cmd/build/build.go
```diff
@@ -271,6 +271,16 @@ func OptsFromBuildInfo(manifestName, svcName string, b *model.BuildInfo, o *type
 		}
 	}
 
+	if len(b.Secrets) > 0 {
+		if o.Secrets == nil {
```
AgustinRamiroDiaz, line 275: nit: this is not necessary as go handles `append` for nil slices (not the same for nil maps though)

pkg/cmd/build/build.go
```diff
@@ -271,6 +271,16 @@ func OptsFromBuildInfo(manifestName, svcName string, b *model.BuildInfo, o *type
...
+			o.Secrets = make([]string, 0)
```
AgustinRamiroDiaz, line 276: `o.Secrets = make([]string, 0, len(b.Secrets))`

pkg/cmd/build/build.go
```diff
@@ -271,6 +271,16 @@ func OptsFromBuildInfo(manifestName, svcName string, b *model.BuildInfo, o *type
...
+		}
+		// add to the build the secrets at the manifest
+		for id, src := range b.Secrets {
+			o.Secrets = append(o.Secrets, fmt.Sprintf("id=%s,src=%s", id, src))
```
AgustinRamiroDiaz, line 280: Not in the scope for this pr: did you find hard to add this secret with this particular formatting? Should we refactor the `Secrets` to be an array of `struct{id string, src string}`?

AgustinRamiroDiaz (approved):

pkg/cmd/build/build.go
teresaromero: `Secrets` is an option for the `okteto build` command, which is already implemented. This PR enables that this option can be added to the services declared inside the build section of the manifest as the issue describes

pkg/model/serializer.go
teresaromero: i think this is self-explanatory taking into account this is a private struct used under the Unmarshall functions

jLopezbarb:
Can we also translate the compose files that has secrets to the okteto manifest using those?

pkg/cmd/build/build.go
teresaromero: I will go forward with an empty slice with 0 length so it will increment when reading the struct key and value

teresaromero:
> Do we currently support this secrets on the docker compose manifest?

I don't see this is currently implemented under `stack` manifest, for this we should implment it as the spec of compose indicates https://github.com/compose-spec/compose-spec/pull/238/files 

should this be scope of the PR ? 
cc: @jLopezbarb @pchico83

AdrianPedriza (approved):

AdrianPedriza (approved):

pchico83:
> > Do we currently support this secrets on the docker compose manifest?
> 
> I don't see this is currently implemented under `stack` manifest, for this we should implment it as the spec of compose indicates https://github.com/compose-spec/compose-spec/pull/238/files
> 
> should this be scope of the PR ? cc: @jLopezbarb @pchico83

@teresaromero let's do it just for the okteto manifest `build` section, that unblocks the scenario

jLopezbarb (approved):

PR #3092, opened 2022-09-12
Return clear error when docker compose doesn't have dev section

### LINKED ISSUE #2876: Fail with a clearer error when docker compose doesn't have dev section
**Describe the bug**
If I run okteto up on a repo with a docker compose and the inferred okteto manifest (docker-compose is transformed into a okteto manifest) doesn't have any dev section (only services with local volumes are transformed into devs) we fail with an unclear error

**To Reproduce**

1. Copy the following docker-compose
```yaml
services:
  web:
    image: alpine
    ports:
    - 8080

```
2. Run `okteto deploy` 
3. Run `okteto up web`
4. See warning indicating that there isn't any dev configured
5. Type `n` when asked if you want to configure
6. See error indicating that no manifest has no dev section

**Expected behavior**
A clear and concise error on why I can't run `okteto up` from my docker-compose file

- Version 2.4.1

**Additional context**
Captura de pantalla 2022-07-01 a las 9 50 57

AgustinRamiroDiaz:
# Proposed changes

Fixes #2876

Checks `oktetoManifest.Type` to return an error when 
- there are not any dev sections
- type is `compose`

### DIFF
--- a/cmd/up/up.go
+++ b/cmd/up/up.go
@@ -158,6 +158,9 @@ func Up() *cobra.Command {
 			os.Setenv(model.OktetoNameEnvVar, oktetoManifest.Name)
 
 			if len(oktetoManifest.Dev) == 0 {
+				if oktetoManifest.Type == model.StackType {
+					return fmt.Errorf("couldn't find any dev environment from your docker compose %q. Only services with local volumes are transformed into devs: https://www.okteto.com/docs/reference/compose/#volumes-string-optional", oktetoManifest.Name)
+				}
 				oktetoLog.Warning("okteto manifest has no 'dev' section.")
 				answer, err := utils.AskYesNo("Do you want to configure okteto manifest now? [y/n]")
 				if err != nil {
--- a/cmd/up/up.go
+++ b/cmd/up/up.go
@@ -53,6 +53,8 @@ import (
 // ReconnectingMessage is the message shown when we are trying to reconnect
 const ReconnectingMessage = "Trying to reconnect to your cluster. File synchronization will automatically resume when the connection improves."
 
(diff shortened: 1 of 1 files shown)


cmd/up/up.go
```diff
@@ -158,6 +158,9 @@ func Up() *cobra.Command {
 			os.Setenv(model.OktetoNameEnvVar, oktetoManifest.Name)
 
 			if len(oktetoManifest.Dev) == 0 {
+				if oktetoManifest.Type == model.StackType {
+					return fmt.Errorf("couldn't find any dev environment from your docker compose %q. Only services with local volumes are transformed into devs: https://www.okteto.com/docs/reference/compose/#volumes-string-optional", oktetoManifest.Name)
```
rberrelleza, line 162: Let's make this more user centric, since this error us more likely to hit new users. Maybe something like:

"Your docker compose file is not currently supported: Okteto requires a "volume to be defined. See %url for more information."

cmd/up/up.go
```diff
@@ -53,6 +53,8 @@ import (
 // ReconnectingMessage is the message shown when we are trying to reconnect
 const ReconnectingMessage = "Trying to reconnect to your cluster. File synchronization will automatically resume when the connection improves."
 
+const composeVolumesUrl = "https://www.okteto.com/docs/reference/compose/#volumes-string-optional"
```
teresaromero, line 56: can we have here 

```

const (
ReconnectingMessage = "Trying to reconnect to your cluster. File synchronization will automatically resume when the connection improves."

composeVolumesUrl = "https://www.okteto.com/docs/reference/compose/#volumes-string-optional"

)
```

teresaromero (approved):

PR #3096, opened 2022-09-13
Different PV default sizes between cloud/staging and others

### LINKED ISSUE #3045: Increase the default size of the okteto dev volume
**Describe the bug**
Motivated by Okteto Cloud quotas, the default size of the okteto volume is 2Gi, which is usually not enough for realistic applications. This is generating many `no space errors` when people are experimenting with Okteto.

**Expected behavior**
Increase the default size to 5Gi. Keep 2Gi if the context is Staging or Okteto Cloud. As part of this work, if `okteto up` fails because the new volume size is larger than the previous one (not all storage classes support size expansion) we shouldn't recreate the volume automatically

AgustinRamiroDiaz:
# Proposed changes

## Fixes #3045

- Handles default PV sizes depending on the context
- Handle error when updating PVC

## Additional context
We had a conversation with @pchico83 about the UX for the migration from 2Gi to 5Gi and we settled for only showing a warning if the update failed due to the upscaling. We won't try to recreate the volume because we'd need to delete the pod that has the volume attached, which takes a lot more time.


AgustinRamiroDiaz:
@pchico83 As discussed, I handled the error when updating dev volumes. Currently I added only the catch for errors like `"persistentvolumeclaims \"%s\" is forbidden: only dynamically provisioned pvc can be resized and the storageclass that provisions the pvc must support resize"`

Should we catch all errors and never fail? We'd still be able to log the error in debug mode

pchico83:
> @pchico83 As discussed, I handled the error when updating dev volumes. Currently I added only the catch for errors like `"persistentvolumeclaims \"%s\" is forbidden: only dynamically provisioned pvc can be resized and the storageclass that provisions the pvc must support resize"`
> 
> Should we catch all errors and never fail? We'd still be able to log the error in debug mode

@AgustinRamiroDiaz Let's do it just for resize errors for now

pkg/k8s/volumes/crud.go
```diff
@@ -71,12 +71,21 @@ func CreateForDev(ctx context.Context, dev *model.Dev, c *kubernetes.Clientset,
...
+			}
+			oktetoLog.Debug("could not update pvc in namespace: %w", dev.Namespace, err)
+			oktetoLog.Warning("Error while updating the dev volume: try running 'okteto down -v' and 'okteto up', or talk to your administrator (the PVC's storage class must support 'allowVolumeExpansion' to be able to upscale dev volumes)")
 		}
 	}
 	return nil
 }
 
+func isDynamicallyProvisionedPVCError(err error, pvcName string) bool {
+	errorString := fmt.Sprintf("persistentvolumeclaims \"%s\" is forbidden: only dynamically provisioned pvc can be resized and the storageclass that provisions the pvc must support resize", pvcName)
```
rberrelleza, line 85: Is there a way to get this string (or the error) from the kube client sdk? It's very easy for this to change and break the error handling(we saw this with the volume snapshots recently)

pkg/k8s/volumes/crud.go
```diff
@@ -99,7 +108,7 @@ func checkPVCValues(pvc *apiv1.PersistentVolumeClaim, dev *model.Dev, devPath st
 		return fmt.Errorf("current okteto volume size is wrong. Run '%s' and try again", utils.GetDownCommand(devPath))
 	}
 	if currentSize.Cmp(resource.MustParse(dev.PersistentVolumeSize())) > 0 {
-		if currentSize.Cmp(resource.MustParse("10Gi")) != 0 || dev.PersistentVolumeSize() != model.OktetoDefaultPVSize {
+		if currentSize.Cmp(resource.MustParse("10Gi")) != 0 || dev.HasDefaultPersistentVolumeSize() {
```
rberrelleza, line 111: MustParse panics, I think is better to use Parse and handle the error ourselves.

pkg/model/volumes.go
```diff
@@ -165,14 +170,36 @@ func (dev *Dev) PersistentVolumeEnabled() bool {
...
 	if dev.PersistentVolumeInfo.Size == "" {
-		return OktetoDefaultPVSize
+		return dev.getDefaultPersistentVolumeSize()
 	}
 	return dev.PersistentVolumeInfo.Size
 }
 
+func (dev *Dev) isOktetoCloud() bool { // TODO: inject this
+	switch dev.Context {
+	case "https://cloud.okteto.com", "https://staging.okteto.dev":
```
rberrelleza, line 183: This is a code smell. Why do we need to know if it's cloud for a test? Could we validate based on a feature or something instead?

pkg/k8s/volumes/crud.go
jLopezbarb: probably we can use `k8s.io/apimachinery/pkg/api/errors`. We're already getting some k8s errors from it

pkg/model/volumes.go
jLopezbarb: I don't think we should be validating from the `dev.Context`. We can use `okteto.IsOktetoCloud `, that will check if it's okteto cloud or staging.

> This is a code smell. Why do we need to know if it's cloud for a test? Could we validate based on a feature or something instead?

We need to check if it's cloud/staging to keep the current volume sizes. If we update cloud and staging to use 5Gb the users will only be able to use one pv which will make their dev experience worse

pkg/k8s/volumes/crud.go
AgustinRamiroDiaz: The only thing I found that identifies this error is the message. The error only has these fields set:

```yaml
Status: "Failure"
Message: "persistentvolumeclaims \"web-okteto\" is forbidden: only dynamically provisioned pvc can be resized and the storageclass that provisions the pvc must support resize"
Reason: "Forbidden"
Details.Kind: "persistentvolumeclaims"
Code: 403
```

Reading from https://github.com/kubernetes/apimachinery/blob/master/pkg/api/errors/errors.go, I couldn't find anything related to this specific error

AgustinRamiroDiaz: @teresaromero From what I've seen, there are other `Forbidden` errors that we don't want to catch here. Catching all `Forbidden` errors could result to wrong printing to the user. 

For example, we could catch the error `error when creating ".": persistentvolumeclaims "wp-pv-claim" is forbidden: exceeded quota` and we'd be printing to the user the message about volume expansion

jLopezbarb: Can we use the cause error field to filter them?

AgustinRamiroDiaz: @jLopezbarb If you are talking about the `Causes` slice, it is empty.

Here's the error with all its fields expanded:
image

pkg/model/volumes.go
AgustinRamiroDiaz: @jLopezbarb why wouldn't we check with the `Dev` struct?

pkg/k8s/volumes/crud.go
jLopezbarb: Then I don't know how we can get the message without rewriting the message 😵‍💫

pkg/model/volumes.go
jLopezbarb: Dev is used to set `okteto.Context()` and there you have all the information about the context. Using `Dev` you only get the name but using `okteto.Context()` you have more information

ifbyol: To have a unified single place where we do the check based on the context

pkg/model/volumes.go
```diff
@@ -183,7 +210,7 @@ func (dev *Dev) PersistentVolumeStorageClass() string {
 
 func (dev *Dev) AreDefaultPersistentVolumeValues() bool {
```
ifbyol, line 211: @jLopezbarb What is this function used for? Might it have some impact on users that deployed a dev volume with previous default value and they start using the new defaults?

pkg/k8s/volumes/crud.go
```diff
@@ -98,8 +108,12 @@ func checkPVCValues(pvc *apiv1.PersistentVolumeClaim, dev *model.Dev, devPath st
...
 		return fmt.Errorf("current okteto volume size is wrong. Run '%s' and try again", utils.GetDownCommand(devPath))
 	}
-	if currentSize.Cmp(resource.MustParse(dev.PersistentVolumeSize())) > 0 {
-		if currentSize.Cmp(resource.MustParse("10Gi")) != 0 || dev.PersistentVolumeSize() != model.OktetoDefaultPVSize {
+	devPVSize, err := resource.ParseQuantity(dev.PersistentVolumeSize())
+	if err != nil {
+		return fmt.Errorf("error parsing dev volume size %q: %w", dev.PersistentVolumeSize(), err)
+	}
+	if currentSize.Cmp(devPVSize) > 0 {
+		if currentSize.Cmp(resource.MustParse("10Gi")) != 0 || dev.HasDefaultPersistentVolumeSize() {
```
ifbyol, line 116: Wouldn't this return an error if someone deployed a dev volume with previous version (2Gi) and then, it executes it again with the new version (5Gi)? It would return that it doesn't have the default persistent volume size

pkg/k8s/volumes/crud.go
```diff
@@ -71,12 +71,22 @@ func CreateForDev(ctx context.Context, dev *model.Dev, c *kubernetes.Clientset,
 		pvc.Spec.VolumeName = k8Volume.Spec.VolumeName
 		_, err = vClient.Update(ctx, pvc, metav1.UpdateOptions{})
 		if err != nil {
-			return fmt.Errorf("error updating kubernetes volume claim: %s", err)
+			if !isDynamicallyProvisionedPVCError(err, pvc.Name) {
+				return fmt.Errorf("error updating kubernetes volume claim: %s", err)
+			}
+			oktetoLog.Debug("could not update pvc in namespace: %w", dev.Namespace, err)
+			oktetoLog.Warning("Error while updating the dev volume: try running 'okteto down -v' and 'okteto up', or talk to your administrator (the PVC's storage class must support 'allowVolumeExpansion' to be able to upscale dev volumes)")
```
ifbyol, line 78: If I understood correctly the ticket, it says that the volume has to be recreated for deployments when the update fails. Where are we recreating the volume? I don't see it

pkg/k8s/volumes/crud.go
AgustinRamiroDiaz: I think it should enter in the other if block since the `currenSize` would be 2Gi and the `devPVSize` would be 5Gi. I don't know what the restart annotation is for though.

One thing I found while investigating this is that we are passing the wrong `pvc` to this function. In `CreateForDev` we should be passing `k8Volume`. As it is now, we are passing the pvc inferred from `dev`, so we are checking against the same value and this function is not doing anything :o

Should I create another ticket to work on that issue or fix it here? If it's just an issue with passing the other parameter, it would be a quick PR and this one could wait for that

pkg/model/volumes.go
AgustinRamiroDiaz: It's only used in https://github.com/okteto/okteto/blob/bd289c9ee57dd1a226ed9caefd58bdb07afab559/pkg/model/serializer.go#L1049 to check for marshaling of the dev struct. It's used to check if the values should be written to the marshal or skipped since they are the defaults we had set. I think it's ok to write the values while marshaling in this "migration" since the defaults have changed

pkg/k8s/volumes/crud.go
AgustinRamiroDiaz: We had a conversation with @pchico83 about the UX for the migration from 2Gi to 5Gi and we settled for only showing a warning if the update failed due to the upscaling (I should've written it down, sorry). 

Reading this again, I think a better UX would be to always handle the error and recreate the volume (as the [comment in the issue](https://github.com/okteto/okteto/issues/3045#issuecomment-1245857059) states). @pchico83 why did we change to showing only a warning when we couldn't upscale?

pkg/model/volumes.go
AgustinRamiroDiaz: I have 2 problems with using `okteto.Context()`
- it causes a cyclic import. It could be addressed by a module re factorization in another pr
- we'd be depending on global variables from other packages, which could cause bugs and makes it harder to debug

I feel like the check could be injected though, because it makes sense that the `Dev` struct shouldn't know about the literal strings of staging and okteto

pkg/k8s/volumes/crud.go
ifbyol: Ok, then, I think the message can be confusing for the user. We are printing a Warning message that start with `Error` and we are not breaking the execution returning an error, so I guess the process just keeps working but the volume size is the old one.

If that is the case, I think that the message should say something like `Could not increase the size of the dev volume: try running 'okteto down -v' and 'okteto up', or talk to your administrator (the PVC's storage class must support 'allowVolumeExpansion' to be able to upscale dev volumes). Previous size %d is being used` (or something similar). Maybe @RinkiyaKeDad can help with the text.

One concern about this is that we could start prompting this message automatically to some users that are working fine with the CLI after a CLI update. @pchico83 Is it a common case that enterprise users use default values? or this change is only to avoid friction to people starting to use Okteto?

pkg/k8s/volumes/crud.go
AgustinRamiroDiaz: I've created https://github.com/okteto/okteto/pull/3117 to solve this problem

pkg/k8s/volumes/crud.go
AgustinRamiroDiaz: :+1: It makes sense to not say `Error` when we are throwing a warning

pkg/k8s/volumes/crud.go
```diff
@@ -71,12 +71,21 @@ func CreateForDev(ctx context.Context, dev *model.Dev, c *kubernetes.Clientset,
 		pvc.Spec.VolumeName = k8Volume.Spec.VolumeName
 		_, err = vClient.Update(ctx, pvc, metav1.UpdateOptions{})
 		if err != nil {
-			return fmt.Errorf("error updating kubernetes volume claim: %s", err)
+			if !isDynamicallyProvisionedPVCError(err, pvc.Name) {
+				return fmt.Errorf("error updating kubernetes volume claim: %w", err)
+			}
+			oktetoLog.Debug("could not update pvc in namespace: %w", dev.Namespace, err)
+			oktetoLog.Warning("Could not increase the size of the dev volume from %s to %s: try running 'okteto down -v' and 'okteto up', or talk to your administrator (the PVC's storage class must support 'allowVolumeExpansion' to be able to upscale dev volumes).", k8Volume.Spec.Resources.Requests[apiv1.ResourceStorage], pvc.Spec.Resources.Requests[apiv1.ResourceStorage])
```
pchico83, line 78: Could we break this line into different lines?

pchico83 (approved):

pkg/k8s/volumes/crud.go
ifbyol: yeah, better to do it in another PR. 

> I think it should enter in the other if block since the currenSize would be 2Gi and the devPVSize would be 5Gi. I don't know what the restart annotation is for though.

You are right, it will enter in the second if. I don't know what the restart annotation does neither. I assume is something needed when the dev volume size increase

pkg/model/volumes.go
ifbyol: Injecting the check could be a good option, do you think can be done without too many changes?

pkg/k8s/volumes/crud.go
```diff
@@ -71,12 +71,21 @@ func CreateForDev(ctx context.Context, dev *model.Dev, c *kubernetes.Clientset,
 		pvc.Spec.VolumeName = k8Volume.Spec.VolumeName
 		_, err = vClient.Update(ctx, pvc, metav1.UpdateOptions{})
 		if err != nil {
-			return fmt.Errorf("error updating kubernetes volume claim: %s", err)
+			if !isDynamicallyProvisionedPVCError(err, pvc.Name) {
+				return fmt.Errorf("error updating kubernetes volume claim: %w", err)
+			}
+			oktetoLog.Debug("could not update pvc in namespace: %w", dev.Namespace, err)
```
teresaromero, line 77:
```suggestion
			oktetoLog.Debug("could not update pvc in namespace %s: %w", dev.Namespace, err)
```

pkg/model/volumes.go
```diff
@@ -22,6 +22,11 @@ import (
 	oktetoLog "github.com/okteto/okteto/pkg/log"
 )
 
+const (
+	cloudAndStagingDefaultVolumeSize = "2Gi"
```
teresaromero, line 26:
```suggestion
	cloudVolumeSize = "2Gi"
```

as staging is an internal env which is basically cloud, can we make shorter variable name which means the same?

pkg/model/volumes.go
AgustinRamiroDiaz: @ifbyol @jLopezbarb There's not a simple way to inject the function because we'd end up in the same place with a circular import. 

There's a module refactor needed to fix this circular dependency, and I think it should be addressed in a follow up pr to [follow the MVC](https://github.com/okteto/okteto/blob/master/docs/code-review.md#pr-scope)

ifbyol: ok, thanks for checking it!

ifbyol (approved):

PR #3098, opened 2022-09-13
fix: deploy with build flag

### LINKED ISSUE #3097: Service not found running `okteto deploy --build`
**Describe the bug**
While performing Test Scenario 1 to the the `release-2.7-beta3` an issue was found when trying to deploy and build 

**To Reproduce**

Test Scenario 1 at the Okteto CLI checks

**Expected behavior**
The images should be built if the flag `--build` is at the deploy command

- OS: [e.g. macOS]

**Additional context**

jLopezbarb:

# Proposed changes

Fixes #3097

### DIFF
--- a/cmd/deploy/deploy.go
+++ b/cmd/deploy/deploy.go
@@ -596,10 +596,22 @@ func (dc *DeployCommand) cleanUp(ctx context.Context) {
 
 func buildImages(ctx context.Context, build func(context.Context, *types.BuildOptions) error, getServicesToBuild func(context.Context, *model.Manifest, []string) ([]string, error), deployOptions *Options) error {
 	if deployOptions.Build {
+		var stackServices map[string]bool
+
+		if stack := deployOptions.Manifest.GetStack(); stack != nil {
+			stackServices = stack.GetServicesWithBuildSection()
+		}
+
+		allServicesWithBuildSection := deployOptions.Manifest.GetBuildServices()
+		oktetoManifestBuildServices := setDifference(allServicesWithBuildSection, stackServices)
+
+		// servicesToBuildSet = (allServicesWithBuildSection - stackServices) + deployOptions.servicesToDeploy
+		servicesToBuildSet := setUnion(oktetoManifestBuildServices, stackServices)
+
 		buildOptions := &types.BuildOptions{
 			EnableStages: true,
 			Manifest:     deployOptions.Manifest,
-			CommandArgs:  deployOptions.servicesToDeploy,
+			CommandArgs:  setToSlice(servicesToBuildSet),
 		}
 		oktetoLog.Debug("force build from manifest definition")
 		if errBuild := build(ctx, buildOptions); errBuild != nil {
--- a/cmd/deploy/deploy.go
+++ b/cmd/deploy/deploy.go
@@ -595,18 +595,18 @@ func (dc *DeployCommand) cleanUp(ctx context.Context) {
 }
 
 func buildImages(ctx context.Context, build func(context.Context, *types.BuildOptions) error, getServicesToBuild func(context.Context, *model.Manifest, []string) ([]string, error), deployOptions *Options) error {
+	var stackServices map[string]bool
+
+	if stack := deployOptions.Manifest.GetStack(); stack != nil {
+		stackServices = stack.GetServicesWithBuildSection()
+	}
+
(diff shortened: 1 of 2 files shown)


cmd/deploy/deploy.go
```diff
@@ -611,14 +623,7 @@ func buildImages(ctx context.Context, build func(context.Context, *types.BuildOp
...
-		var stackServices map[string]bool
-
-		if stack := deployOptions.Manifest.GetStack(); stack != nil {
-			stackServices = stack.GetServicesWithBuildSection()
-		}
 
-		allServicesWithBuildSection := deployOptions.Manifest.GetBuildServices()
-		oktetoManifestBuildServices := setDifference(allServicesWithBuildSection, stackServices)
 		// servicesToBuildSet = (allServicesWithBuildSection - stackServices) + deployOptions.servicesToDeploy
 		servicesToBuildSet := setUnion(oktetoManifestBuildServices, sliceToSet(deployOptions.servicesToDeploy))
```
AgustinRamiroDiaz, line 628: Based on the new logic that I'm seeing, I think that we should apply the same here: only account for services which have a build section

cmd/deploy/deploy.go
```diff
@@ -595,11 +595,23 @@ func (dc *DeployCommand) cleanUp(ctx context.Context) {
...
+
+	if stack := deployOptions.Manifest.GetStack(); stack != nil {
+		stackServices = stack.GetServicesWithBuildSection()
+	}
+
+	allServicesWithBuildSection := deployOptions.Manifest.GetBuildServices()
+	oktetoManifestBuildServices := setDifference(allServicesWithBuildSection, stackServices)
+
 	if deployOptions.Build {
+		finalStackIntersection := setIntersection(stackServices, sliceToSet(deployOptions.servicesToDeploy))
```
AgustinRamiroDiaz, line 608: I think this does the correct behavior, but it's complex to understand.

I'm thinking something like this might be simpler to understand:
```
...
allServicesWithBuildSection := deployOptions.Manifest.GetBuildServices()
oktetoManifestBuildServices := setDifference(allServicesWithBuildSection, stackServices)
servicesToDeployWithBuildSection := setIntersection(allServicesWithBuildSection, sliceToSet(deployOptions.servicesToDeploy))
servicesToBuildSet := setUnion(oktetoManifestBuildServices, servicesToDeployWithBuildSection)
```

teresaromero: i am not sure i am understanding this

cmd/deploy/deploy.go
teresaromero: I see the only thing change here is common logic for both build and not build flag has been taken out the if/else. What do you mean with `apply the same here?`

AgustinRamiroDiaz: I think that in both cases (build forced and unforced) we should be collecting the same images. The only difference would be in the unforced build that we then filter that slice of images if they have been already built

So I think we should do here the same we are doing in the other block 

```
finalStackIntersection := setIntersection(stackServices, sliceToSet(deployOptions.servicesToDeploy))
servicesToBuildSet := setUnion(oktetoManifestBuildServices, finalStackIntersection)
```

cmd/deploy/deploy.go
AgustinRamiroDiaz: If I understand this logic correctly, with Javi's change we'll build:
- `deployOptions.servicesToDeploy` that are defined in the stack and have a build section
- services with build section defined in the `okteto.yml` manifest

Since the `manifest`'s build section contains the `stack`'s build services, if there's a service from `deployOptions.servicesToDeploy` that's not in the stack, it will be in the manifest and will get built anyways. 
I think the logic is more clear if we compare everything to `allServicesWithBuildSection`. So in summary we'll build:
- `deployOptions.servicesToDeploy` that have a build section
- services with build section defined in the `okteto.yml` manifest

AgustinRamiroDiaz (approved):
I've created [this pr](https://github.com/okteto/okteto/pull/3100) to explain better the changes I proposed.
The only change in logic I'd do is to get the images to build the same way for both scenarios (build forced and unforced). I think it does not affect the behavior because in the unforced path we are filtering with `getServicesToBuild`, which should take away all the `servicesToDeploy` without build section

Issue #3006, opened 2022-08-15
Add Secrets to Okteto Using CLI


jmacelroy:
Would be nice to have the ability to add Secrets via CLI, instead of manually via the UI ([Okteto Secrets | Okteto Documentation](https://www.okteto.com/docs/cloud/secrets/#manage-okteto-secrets-from-the-okteto-cloud-ui)).

My use case is that every developer in my group needs to add it’s own private secrets under the same names, so I would like to create a small script that does it for them using the CLI.

jLopezbarb:
What should the command looks like?
- `okteto secrets create`: Creates the secret, but should we allow more than one secret being set at the same time (I.E:`okteto secrets create key1=value key2=value2`)
- `okteto secrets delete`:  Deletes the secret containing that key. This command should be idempontent.

Should we create a `okteto secrets list` and `okteto secrets show` showing only the last 3 chars?

hadaromash:
A `show` command may be useful, to verify if a key exists before creating it.

faliqadlan:
can i ask, how to access okteto secrets which define in settings ?

pchico83:
Closing since this should go through the product backlog now. Also, adding commands that don't work on clusters without the okteto server side installed is porblematic

PR #3106, opened 2022-09-15
add e2e test to check if build cmd replace secrets properly

### LINKED ISSUE #3056: Add e2e for build cmd to check secrets are resolved properly
we should add an e2e test to cover the case of checking that the secrets are correctly added when reading the manifest for the `okteto build` command.  In order to release [the fix](#3051 ) that solves #3044, the suggestion for the addition a e2e test is made separately of the original PR.

AdrianPedriza:

# Proposed changes

Fixes #3056

- add e2e test where build cmd is tested using secrets in manifest build section

### DIFF
--- a/integration/okteto/build_test.go
+++ b/integration/okteto/build_test.go
@@ -1,0 +1,74 @@
+package okteto
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/okteto/okteto/integration"
+	"github.com/okteto/okteto/integration/commands"
+	"github.com/stretchr/testify/require"
+)
+
+const (
+	manifestContentWithSecrets = `
+build:
+  test:
+    context: .
+    args:
+    - RABBITMQ_PASS=${RABBITMQ_PASS}
+
+deploy:
+- echo "fake deploy"
+`
+	dockerfileUsingSecrets = `
+FROM nginx
+ARG RABBITMQ_PASS
+RUN if [ -z "$RABBITMQ_PASS" ]; then exit 1; else echo $RABBITMQ_PASS; fi
+`
+)
+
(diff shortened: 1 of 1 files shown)


AgustinRamiroDiaz:
Does this test need `RABBITMQ_PASS` to be set from the UI? 
If it does:
- Is there a way to create the secret from the CLI?
- Can we document it?

AdrianPedriza:
> Does this test need `RABBITMQ_PASS` to be set as a secret from the UI? If it does:
> 
> * Is there a way to create the secret from the CLI? Maybe we need to create another issue for this
> * Can we document it?

It is already added as a secret. For the moment we cannot add secrets in the test itself by using the CLI but there is already an issue  to create test from CLI. https://github.com/okteto/okteto/issues/3006

AgustinRamiroDiaz:
@AdrianPedriza so we'd need to add the steps to set `RABBITMQ_PASS` [here](https://github.com/okteto/okteto/blob/master/docs/how-to-run-tests.md)

AdrianPedriza:
> @AdrianPedriza so we'd need to add the steps to set `RABBITMQ_PASS` [here](https://github.com/okteto/okteto/blob/master/docs/how-to-run-tests.md)

There is no need to set it, it is already set for that environment. Anyway, I know at the moment it is coupled to the secrets you have in that environment but once you can add secrets from the CLI it will be more dynamic. I will add a reminder in the [issue](https://github.com/okteto/okteto/issues/3006) to take into account tests like this.

jLopezbarb:
Can we run e2e tests before approving it? I want to make sure it's passing

AdrianPedriza:
> Can we run e2e tests before approving it? I want to make sure it's passing

passed! @jLopezbarb

jLopezbarb (approved):

PR #3108, opened 2022-09-15
Context store may just be non-existent rather than corrupt

### LINKED ISSUE #3107: Context store may just be non-existent rather than corrupt
**Describe the bug**
When running 'okteto ctx ls' without ever having run 'okteto ctx' (for example when following https://www.okteto.com/blog/developing-microservices-by-hot-reloading-on-kubernetes-clusters/ ), the error message: "FATA[0000] okteto context store is corrupted. Delete the folder ..." appears

**To Reproduce**

1. Install okteto cli
1. Run 'okteto ctx ls'
1. See error

**Expected behavior**
The error should suggest running 'okteto ctx' as an alternate to deleting the context folder.

- OS: macOS
- Version 12.6

**Additional context**

zaheerm:
Signed-off-by: Zaheer Abbas Merali <zaheerabbas@merali.org>

# Proposed changes

Fixes #3107 

Amends the error message saying the context store may be corrupted by suggesting it is missing and recommends also to try running 'okteto ctx'

### DIFF
--- a/cmd/context/list.go
+++ b/cmd/context/list.go
@@ -37,6 +37,9 @@ func List() *cobra.Command {
 		Short:   "List available contexts",
 		RunE: func(cmd *cobra.Command, args []string) error {
 			ctx := context.Background()
+			if err := NewContextCommand().Run(ctx, &ContextOptions{raiseNotCtxError: true}); err != nil {
+				return err
+			}
 			return executeListContext(ctx)
 		},
 	}


jLopezbarb:
Hi @zaheerm. Could we do something similar to what we are doing in `okteto context show`? Running this https://github.com/okteto/okteto/blob/master/cmd/context/show.go#L38 at the start of the list command should be enough to get the proper message

zaheerm:
Thanks @jLopezbarb . That is much more elegant, so have pushed that change here now.

jLopezbarb (approved):

teresaromero (approved):

jLopezbarb:
@zaheerm Thanks again for your contribution!! This is a really nice addition

PR #3109, opened 2022-09-16
redeploy stack with 0 downtime

### LINKED ISSUE #3083: I want to redeploy my stack without downtime
**Describe the bug**
If I redeploy my `okteto stack` I want to have 0 downtimes but since the deployments created have the recreate update strategy, I can't 

**To Reproduce**

1. Copy the following `docker-compose.yml`
```yaml
services:
  app:
    image: python:alpine
    entrypoint: python -m http.server 8080
    ports:
      - 8080:8080
```
2. Run okteto deploy
3. Change the entrypoint to `python -m http.server 8081`
4. Run okteto deploy to redeploy
5. See that the pod is recreated

**Expected behavior**
I want to have 0 downtime while redeploying a compose/stack

- Version 2.6.0

jLopezbarb:

# Proposed changes

Fixes #3083

- Change from recreate strategy to rolling update

# Concerns

This change can generate new quota issues that we cannot detect if the command is not executed with the `--wait` flag.
Captura de Pantalla 2022-09-16 a las 10 45 10

### DIFF
--- a/pkg/cmd/stack/translate_test.go
+++ b/pkg/cmd/stack/translate_test.go
@@ -22,6 +22,7 @@ import (
 
 	"github.com/okteto/okteto/pkg/model"
 	"github.com/stretchr/testify/assert"
+	appsv1 "k8s.io/api/apps/v1"
 	apiv1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -152,7 +153,7 @@ func Test_translateDeployment(t *testing.T) {
 	if !reflect.DeepEqual(c.Resources, apiv1.ResourceRequirements{}) {
 		t.Errorf("Wrong container.resources: '%v'", c.Resources)
 	}
-
+	assert.Equal(t, result.Spec.Strategy, appsv1.RollingUpdateDeploymentStrategyType)
 }
 
 func Test_translateStatefulSet(t *testing.T) {
@@ -346,7 +347,7 @@ func Test_translateStatefulSet(t *testing.T) {
 	if !reflect.DeepEqual(vct.Spec, volumeClaimTemplateSpec) {
 		t.Errorf("Wrong statefulset volume claim template: '%v'", vct.Spec)
 	}
-
+	assert.Equal(t, result.Spec.UpdateStrategy, appsv1.RollingUpdateStatefulSetStrategyType)
 }
 
 func Test_translateJobWithoutVolumes(t *testing.T) {
--- a/pkg/cmd/stack/translate_test.go
+++ b/pkg/cmd/stack/translate_test.go
@@ -153,7 +153,7 @@ func Test_translateDeployment(t *testing.T) {
 	if !reflect.DeepEqual(c.Resources, apiv1.ResourceRequirements{}) {
 		t.Errorf("Wrong container.resources: '%v'", c.Resources)
 	}
-	assert.Equal(t, result.Spec.Strategy.Type, appsv1.RollingUpdateDeploymentStrategyType)
+
 }
 
 func Test_translateStatefulSet(t *testing.T) {
@@ -347,7 +347,7 @@ func Test_translateStatefulSet(t *testing.T) {
 	if !reflect.DeepEqual(vct.Spec, volumeClaimTemplateSpec) {
 		t.Errorf("Wrong statefulset volume claim template: '%v'", vct.Spec)
 	}
-	assert.Equal(t, result.Spec.UpdateStrategy.Type, appsv1.RollingUpdateStatefulSetStrategyType)
+
 }
 
 func Test_translateJobWithoutVolumes(t *testing.T) {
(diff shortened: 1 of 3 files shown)


AgustinRamiroDiaz:
@jLopezbarb would we always hit twice the deployments at some point? Maybe if we know how much resources we got at the "spike" we could prevent quota issues

Also, I don't see any change for statefulsets and maybe it's for the best:
"When using [Rolling Updates](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#rolling-updates) with the default [Pod Management Policy](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#pod-management-policies) (OrderedReady), it's possible to get into a broken state that requires [manual intervention to repair](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#forced-rollback)"
Taken from https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#limitations

ifbyol (commented):
Would it make sense to allow a new property in the stack to specify how to replace the deployment and keep the current behavior by default?

pkg/cmd/stack/translate_test.go
```diff
@@ -346,7 +347,7 @@ func Test_translateStatefulSet(t *testing.T) {
 	if !reflect.DeepEqual(vct.Spec, volumeClaimTemplateSpec) {
 		t.Errorf("Wrong statefulset volume claim template: '%v'", vct.Spec)
 	}
-
+	assert.Equal(t, result.Spec.UpdateStrategy.Type, appsv1.RollingUpdateStatefulSetStrategyType)
```
ifbyol, line 350: Is the rolling update the default for statefulsets? I don't see a change for statefulsets but we are adding this check

jLopezbarb: Yes, we already have it as the default one for sfs

jLopezbarb:
> Would it make sense to allow a new property in the stack to specify how to replace the deployment and keep the current behavior by default?

Maybe by adding an annotation to the service, we can change the redeploy strategy. But I think we should stick to rolling update as the default one since it's the most common scenario that people want to achieve

ifbyol:
> > Would it make sense to allow a new property in the stack to specify how to replace the deployment and keep the current behavior by default?
> 
> Maybe by adding an annotation to the service, we can change the redeploy strategy. But I think we should stick to rolling update as the default one since it's the most common scenario that people want to achieve

It is the most common scenario in a **production** environment. Talking about a dev environment scenario, a small downtime should be acceptable. Also, this is a change of behavior and it might have impact on the quotas, so that's why I mention a more conservative approach of keeping current behavior

jLopezbarb:
> > > Would it make sense to allow a new property in the stack to specify how to replace the deployment and keep the current behavior by default?
> > 
> > 
> > Maybe by adding an annotation to the service, we can change the redeploy strategy. But I think we should stick to rolling update as the default one since it's the most common scenario that people want to achieve
> 
> It is the most common scenario in a **production** environment. Talking about a dev environment scenario, a small downtime should be acceptable. Also, this is a change of behavior and it might have impact on the quotas, so that's why I mention a more conservative approach of keeping current behavior

After discussing with @pchico83 we decided to go with the approach of annotation on each service or set an environment variable to get the behaviour on every service (these changes only affect deployments and statefulset from compose).

The annotation should be `dev.okteto.com/upgrade` with the available options being `rolling` for rolling update and `recreate`.

The environment is `OKTETO_UPGRADE_POLICY` with the same available options as before.

The annotation will take precedence over the environment variable, that way a user can have all its services with a default upgrade policy with the environment variable and another upgrade policy for specific services

pkg/cmd/stack/translate_test.go
jLopezbarb: I don't know where I checked but I don't see it know, so I will add it too

ifbyol: It is important to not change the current behavior for sfs

jLopezbarb: Checked that the default behavior is rolling update, so we still have the same behavior

jLopezbarb:
I changed a little bit the implementation that was suggested [in the previous comment](https://github.com/okteto/okteto/pull/3109#issuecomment-1252433982).

I had to add `onDelete` to the list of possible update strategy, so the available values are `rolling` and `onDelete` for `statefulsets` and `rolling` and `recreate` for `deployments`.

I decided to change the initial annotation from `dev.okteto.com/upgrade` to `dev.okteto.com/update` to keep consistency with the k8s naming convention and the environment variable from `OKTETO_UPGRADE_POLICY` to `OKTETO_COMPOSE_UPDATE_STRATEGY` since the variable is only going to be used on the compose/stack and used the word update strategy to keep consistency with the previous annotation. WDYT?

cc: @pchico83 @ifbyol

pchico83:
> onDelete

@jLopezbarb makes sense. But let's use `on-delete` as the value, it's the format we use in every annotation value

ifbyol:
Ok, sounds good! 👍

teresaromero (approved):

PR #3115, opened 2022-09-20
add build variables as env vars for buildArgs

### LINKED ISSUE #3089: $OKTETO_BUILD_* variables not available for use in build section of manifest
**Describe the bug**
The $OKTETO_BUILD_* variables are not available for use in the `build` section of the manifest. I would like to be able to pass build arguments to  my Dockerfile using the values of a previously-built service, but they don't seem to be available at that point.

**To Reproduce**

1. Create `okteto.yaml` with the following contents:
```
build:
  service-a:
    dockerfile: Dockerfile.a
  service-b:
    dockerfile: Dockerfile.b
    depends_on: service-a
    args:
      SOURCE_REGISTRY: $OKTETO_BUILD_SERVICE_A_REGISTRY
      SOURCE_REPOSITORY: $OKTETO_BUILD_SERVICE_A_REPOSITORY
      SOURCE_TAG: $OKTETO_BUILD_SERVICE_A_TAG
```
2. Create `Dockerfile.a` with the following contents:
```
FROM bash:latest

RUN echo "IMAGE A" > /root/output_a.txt

ENTRYPOINT ["tail", "-f", "/dev/null"]
```
3. Create `Dockerfile.b` with the following contents:
```
FROM bash:latest

ARG SOURCE_REGISTRY="(DEFAULT_REGISTRY)"
ARG SOURCE_REPOSITORY="(DEFAULT_REPOSITORY)"
ARG SOURCE_TAG="(DEFAULT_TAG)"

RUN echo "IMAGE B" > /root/output_b.txt \
 && echo "SOURCE_REGISTRY=$SOURCE_REGISTRY" >> /root/output_b.txt \
 && echo "SOURCE_REPOSITORY=$SOURCE_REPOSITORY" >> /root/output_b.txt \
 && echo "SOURCE_TAG=$SOURCE_TAG" >> /root/output_b.txt \
;

ENTRYPOINT ["tail", "-f", "/dev/null"]
```
4. Run a build using at least version 2.7.0 of Okteto CLI:
```
/usr/local/bin/okteto-2.7.0-beta.2 build --log-level debug
```
5. Bring up a pod in the Okteto cluster using the new image. For example, with the following manifest:
```
apiVersion: v1
kind: Pod
metadata:
  name: test
spec:
  containers:
    - name: web
      image: registry.<OKTETO_DOMAIN>/<NAMESPACE>/okteto-service-b:okteto
```
6. Examine the contents of the file written during the build process:
```
kubectl -n <NAMESPACE> exec -it pod/test -- cat /root/output_b.txt
```

**Expected behavior**
The file should contain the following output, indicating that the build arguments were set correctly during the build process:
```
IMAGE B
SOURCE_REGISTRY=registry.<OKTETO_DOMAIN>
SOURCE_REPOSITORY=<NAMESPACE>/okteto-service-a
SOURCE_TAG=sha256:3f0f4c903f0f61cd2a8cdcd0f82cd0e23521ab2637e495bba176e8f1c485d44e
```

- OS: macOS
- Version: Ventura 13.0 beta

**Additional context**
The `$OKTETO_BUILD_*` values **are** available as build arguments within the image itself, if I add lines like:
```
ARG OKTETO_BUILD_SERVICE_A_REGISTRY
```
to the Dockerfile. But I would like to be able to use a standalone Dockerfile within Okteto without having to rewrite the Dockerfile to incorporate the Okteto-specific variables. This requires being able to pass in the argument names that it's expecting by defining them in the Okteto manifest, as described above.

jLopezbarb:

# Proposed changes

Fixes #3089

- Created a new struct to load build args before building the image (the moment where we have the previous build environment variables set)
- Expand env vars only on the environment, annotations, and labels (not in build Args).
- Add tests to get the expected behavior


teresaromero (approved):
i think the deepsource errors where fixed, can you try to update the branch with latest master and check?

jLopezbarb:
> i think the deepsource errors where fixed, can you try to update the branch with latest master and check?

Fixed. Thanks!!!

pkg/model/dev.go
```diff
@@ -156,15 +169,50 @@ func (b *BuildInfo) GetDockerfilePath() string {
 	return joinPath
 }
 
-// AddArgs add a set of args to the build information
-func (b *BuildInfo) AddArgs(args map[string]string) {
-	for k, v := range args {
-		b.Args = append(b.Args, EnvVar{
+// ExpandBuildArgs add a set of args to the build information
+func (b *BuildInfo) ExpandBuildArgs(previousImageArgs map[string]string) error {
```
ifbyol, line 173: This function is doing more than expanding the build args, it is also (or potentially) including new build args, so maybe we can call it `AddAndExpandBuildArgs`

pkg/model/dev.go
```diff
@@ -156,15 +169,50 @@ func (b *BuildInfo) GetDockerfilePath() string {
...
+		arg.Value, err = ExpandEnv(arg.Value, true)
+		if err != nil {
+			return err
+		}
+		b.Args[idx] = arg
+	}
+	return nil
+}
+
+func (b *BuildInfo) expandPreviousImageArgs(previousImageArgs map[string]string) error {
```
ifbyol, line 196:
```suggestion
func (b *BuildInfo) addExpandedPreviousImageArgs(previousImageArgs map[string]string) error {
```

What do you think about renaming the function? This is actually adding new build args if needed

pkg/model/dev_test.go
```diff
@@ -1574,3 +1574,157 @@ func TestBuildInfo_GetDockerfilePath(t *testing.T) {
...
+			buildInfo: &BuildInfo{
+				Args: BuildArgs{
+					{
+						Name:  "KEY",
+						Value: "VALUE",
+					},
+				},
+			},
+			previousImageBuilt: map[string]string{
+				"KEY2": "VALUE",
```
ifbyol, line 1658:
```suggestion
				"KEY2": "VALUE2",
```

Nit, but it can helps to avoid confusion to understand if the `VALUE` can come for `KEY` or `KEY2`. Apply to other cases where the 2 keys are used

pkg/model/serializer.go
```diff
@@ -1185,6 +1206,45 @@ func (e *Environment) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
+	return nil
+}
+
+func getBuildArgs(unmarshal func(interface{}) error) (map[string]string, error) {
+	result := make(map[string]string)
+
+	var rawList []BuildArg
+	err := unmarshal(&rawList)
+	if err == nil {
+		for _, label := range rawList {
```
ifbyol, line 1232:
```suggestion
		for _, buildArg := range rawList {
```

I think it is more accurate than `label`, which in this context I think it doesn't mean anything relevant

cmd/build/v2/build.go
```diff
@@ -184,7 +184,9 @@ func (bc *OktetoBuilder) buildSvcFromDockerfile(ctx context.Context, manifest *m
 	isStackManifest := manifest.Type == model.StackType
 	buildSvcInfo := getBuildInfoWithoutVolumeMounts(manifest.Build[svcName], isStackManifest)
 
-	buildSvcInfo.AddArgs(bc.buildEnvironments)
+	if err := buildSvcInfo.ExpandBuildArgs(bc.buildEnvironments); err != nil {
```
ifbyol, line 187: Before, the `AddArgs` function weren't expanding the build arguments and now, we are doing it. Why? When were the build args expanded before?

jLopezbarb: We were doing half of the process before. We were creating `OKTETO_BUILD` env vars as args but we weren't expanding those args if the user used a custom key such as:
```yaml
build:
  args:
    FROM_IMAGE: ${OKTETO_BUILD_APP_IMAGE}
```

pkg/model/dev.go
jLopezbarb: Should we split it in two and use two different functions instead? I don't really like the name `AddAndExpandBuildArgs`.
We can have `AddBuildArgs` and `ExpandBuildArgs` which is more descriptive but I'm afraid that every time we do one we must do the other one so that's why I included them in a single function 🤔

ifbyol: If they have to be called together, I would create a single function containing both of them, but given our code base, if I see a function called `ExpandBuildArgs` I wouldn't expect to have more arguments, I would just expect that the existent ones would be expanded.

Other option could be to just call it `AddBuildArgs`. The fact that the build args has to be expanded is internal to `BuildInfo` struct, so if everytime I include some new argument, it will be expanded, I think it is fair to call it just `AddBuildArgs`. In the function comment you can mention that the arguments will be expanded

cmd/build/v2/build.go
ifbyol: Ah ok, so we were not expanding them before?

pkg/model/dev.go
```diff
@@ -156,15 +169,50 @@ func (b *BuildInfo) GetDockerfilePath() string {
...
+// ExpandBuildArgs add a set of args to the build information
+func (b *BuildInfo) ExpandBuildArgs(previousImageArgs map[string]string) error {
+	if err := b.expandManifestBuildArgs(previousImageArgs); err != nil {
+		return err
+	}
+	return b.expandPreviousImageArgs(previousImageArgs)
+}
+
+func (b *BuildInfo) expandManifestBuildArgs(previousImageArgs map[string]string) error {
+	var err error
```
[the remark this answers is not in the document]
jLopezbarb: Thanks it looks cleaner the way yo say :)

cmd/build/v2/build.go
jLopezbarb: Not expanding, we were just adding the previous args:
```yaml
build:
  args:
    FROM_IMAGE: ${OKTETO_BUILD_APP_IMAGE}
```
was converted into:
```yaml
build:
  args:
    FROM_IMAGE: # empty
    OKTETO_BUILD_APP_IMAGE: real_image_sha
```

ifbyol (approved):

AdrianPedriza (approved):

PR #3117, opened 2022-09-21
Fix wrong arguments passed to `checkpvcvalues`

### LINKED ISSUE #3116: Wrong arguments passed to `checkPVCValues`
One thing I found while investigating this is that we are passing the wrong `pvc` to this function. In `CreateForDev` we should be passing `k8Volume`. As it is now, we are passing the pvc inferred from `dev`, so we are checking against the same value and this function is not doing anything :o

_Originally posted by @AgustinRamiroDiaz in https://github.com/okteto/okteto/pull/3096#discussion_r976557803_

AgustinRamiroDiaz:
# Proposed changes

Fixes #3116
- passed the correct value to the function
- added test cases

Changed behavior:
Before:
- `checkPVCValues` was doing nothing since we were checking against the same struct we've created 

Now:
- We'll compare the inferred dev volume that comes from the manifest to the actual PVC that's deployed in kubernetes:
  - if the dev volume is larger than the deployed one, we'll add a restart annotation. @pchico83  @jLopezbarb do you know the consequences of this? I couldn't find anything checking this value
  - if the dev volume is smaller, we'll return an error stating that you cannot downsize your dev volume and we'll guide the user to run `okteto down -v`
  - if the dev volume's storage class doesn't match with the deployed one, it will return the same error as above
  
@ifbyol @jLopezbarb This is a change in behavior that no one has complained about. We are trying to aim for stability, so maybe the best is to just discard this at all and keep the current behavior. WDYT?


jLopezbarb (approved):

pkg/k8s/volumes/crud.go
```diff
@@ -47,29 +47,29 @@ func List(ctx context.Context, namespace, labels string, c kubernetes.Interface)
...
-		oktetoLog.Infof("creating volume claim '%s'", pvc.Name)
-		_, err = vClient.Create(ctx, pvc, metav1.CreateOptions{})
+		oktetoLog.Infof("creating volume claim '%s'", pvcFromDev.Name)
+		_, err = vClient.Create(ctx, pvcFromDev, metav1.CreateOptions{})
 		if err != nil {
 			return fmt.Errorf("error creating kubernetes volume claim: %s", err)
 		}
 	} else {
-		if err := checkPVCValues(pvc, dev, devPath); err != nil {
+		if err := checkPVCValues(k8Volume, dev, devPath); err != nil {
```
ifbyol, line 64: Since when have been like that? I know this is the expected and correct behavior but I guess this can start failing where it wasn't before and it might increase the friction.

Can we include the possible impact? I mean, with the current code, what was happening? what was wrong? Was everything working fine or this was causing an error later?

Which implies the new change? Can fail scenarios that were working before?

jLopezbarb: I can't think of any scenario that was working (without any error on up) to start failing with these change. The only thing that is done is checking if we are going to have an error before we get the k8s error

AgustinRamiroDiaz: @ifbyol the change was originated [here](https://github.com/okteto/okteto/pull/1586/files). It seems to be a bug since before it was done as the proposed fix.

AgustinRamiroDiaz: @ifbyol I've added more description to the PR to explain the impact of the changes

ifbyol: ok, thanks for the information. I think we can merge it and no working scenarios should be affected

ifbyol (approved):

pkg/k8s/volumes/crud.go
```diff
@@ -47,29 +47,29 @@ func List(ctx context.Context, namespace, labels string, c kubernetes.Interface)
 }
 
 // CreateForDev deploys the volume claim for a given development container
-func CreateForDev(ctx context.Context, dev *model.Dev, c *kubernetes.Clientset, devPath string) error {
+func CreateForDev(ctx context.Context, dev *model.Dev, c kubernetes.Interface, devPath string) error {
 	vClient := c.CoreV1().PersistentVolumeClaims(dev.Namespace)
-	pvc := translate(dev)
-	k8Volume, err := vClient.Get(ctx, pvc.Name, metav1.GetOptions{})
+	pvcFromDev := translate(dev)
```
AdrianPedriza, line 52: minor: wouldn't it be more appropriate `pvcForDev` instead `pvcFromDev` ? maybe even the function name already gives the necessary context and we can just use `pvc` what do you think?

AdrianPedriza (approved):

pkg/k8s/volumes/crud.go
AgustinRamiroDiaz: @AdrianPedriza I changed the name because I think it's the reason the bug got introduced, `pvc` is very generic and can lead to misuse. I think the best name would be `pvcInferredFromDev` since it's what we actually do, but `pvcForDev` also makes sense since we are going to use it as the PVC for the dev environment. 

I'll go with `pvcForDev` since it's what you felt was more intuitive and it's short and concise :+1:

PR #3119, opened 2022-09-22
Fix secret copy

teresaromero:
# Proposed changes

Secrets declared at the manifest, or via the cmd flag `--secret` where removed when creating a Copy of the BuildInfo struct.
- This PR adds Secrets and other fields from BuildInfo that where being removed as nil pointers.
- Added integration and unit test for Copy function.

On the other hand, this PR also implements the use of templates at secret files, this way a secret can be mounted not only having  a local file, but using a template for the secret and having envs that will be replaced when running the build and mounting the secret.
- Parses the content of the secret file into builtkit as a temp file

### DIFF
--- a/pkg/cmd/build/build.go
+++ b/pkg/cmd/build/build.go
@@ -377,9 +377,9 @@ func parseTempSecrets(buildOptions *types.BuildOptions) (string, error) {
 		}
 
 		// save expanded to temp file
-		writter := bufio.NewWriter(tmpfile)
-		_, _ = writter.Write([]byte(fmt.Sprintf("%s\n", srcContent)))
-		writter.Flush()
+		writer := bufio.NewWriter(tmpfile)
+		_, _ = writer.Write([]byte(fmt.Sprintf("%s\n", srcContent)))
+		writer.Flush()
 
 		// replace src
 		buildOptions.Secrets[indx] = fmt.Sprintf("%ssrc=%s", splitSecret[0], tmpfile.Name())
--- a/pkg/cmd/build/build.go
+++ b/pkg/cmd/build/build.go
@@ -14,8 +14,10 @@
 package build
 
 import (
+	"bufio"
 	"context"
 	"fmt"
+	"io/ioutil"
 	"os"
 	"path/filepath"
 	"strings"
@@ -23,6 +25,7 @@ import (
 	"github.com/docker/docker/api/types/versions"
 	"github.com/docker/docker/client"
 	"github.com/okteto/okteto/pkg/analytics"
+	"github.com/okteto/okteto/pkg/config"
 	oktetoErrors "github.com/okteto/okteto/pkg/errors"
 	"github.com/okteto/okteto/pkg/filesystem"
 	oktetoLog "github.com/okteto/okteto/pkg/log"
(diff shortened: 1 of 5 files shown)


pkg/model/dev.go
```diff
@@ -1299,12 +1299,15 @@ func getLocalhost() string {
 // Copy clones the buildInfo without the pointers
 func (b *BuildInfo) Copy() *BuildInfo {
 	result := &BuildInfo{
```
jLopezbarb, line 1301: I think we will always forget to add the new fields here... Maybe we should use something like https://github.com/jinzhu/copier to perform this task. The problem with these modules is that they use reflect inside which can cause runtime errors.

pkg/model/dev.go
```diff
@@ -1313,8 +1316,19 @@ func (b *BuildInfo) Copy() *BuildInfo {
 	args = append(args, b.Args...)
 	result.Args = args
 
+	secrets := BuildSecrets{}
+	for k, v := range b.Secrets {
+		secrets[k] = v
```
pchico83, line 1321: Instead of passing the original content of the file, let's create a temporal file (and remove it with a defer after the build) with the content of the file after applying env substitutiton.
For example, for `.npmrc`, you would have this file as a secret:
```
egistry=https://registry.npmjs.org/
//***.jfrog.io/artifactory/api/npm/:_authToken=${NPM_TOKEN}
@scope**:registry=https://***.jfrog.io/artifactory/api/npm/private-npm
```
you won't need to commit your token, okteto will replace the token using okteto secrets or your local env vars

pkg/model/dev.go
teresaromero: i've used marshall unmarshall method, is this an option? seems to work fine

AgustinRamiroDiaz: @teresaromero we could also have a silent error with marshalling because it needs the fields to be exported. If some day we add an unexported field, it won't be copied

teresaromero: i've reverted to loop between fields and pointers.
@jLopezbarb it is a risk to forget... but this struct is not changing that much, and eventually the bug will appear when we use differnt build options if we add some new.... I prefer this than other silent errors or adding a new library....

jLopezbarb: I'm ok with it, I just wanted to make sure that this was going to be a problem when implementing new fields. Right now it only affects buildInfo so it's a thing we can deal with it but if we ever have several structs to copy we should take a look for one of the solutions

ifbyol:
@teresaromero is there a ticket linked to this? Is this a fast track or a normal issue?

teresaromero:
> @teresaromero is there a ticket linked to this? Is this a fast track or a normal issue?

This was opened after @pchico83  reported at slack to unblock the feature for secrets

ifbyol (approved):

pkg/cmd/build/build.go
```diff
@@ -115,6 +117,14 @@ func buildWithOkteto(ctx context.Context, buildOptions *types.BuildOptions) erro
 			buildOptions.ExportCache = registry.ExpandOktetoGlobalRegistry(buildOptions.ExportCache)
 		}
 	}
+
+	// inject secrets to buildkit from temp folder
+	tempSecretsFolder, err := parseTempSecrets(buildOptions)
+	if err != nil {
+		return err
```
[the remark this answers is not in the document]
teresaromero: 🤔  i've taken a look and i've moved out the `parseTempSecrets` the folder creation, this way the defer will clear all in case of error. WDYT?

jLopezbarb (approved):

pkg/cmd/build/build.go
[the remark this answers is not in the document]
ifbyol: yeah, makes sense!

ifbyol (approved):

PR #3122, opened 2022-09-23
Deepsource bug risk

teresaromero:
# Proposed changes

As part of the deepsource issue cleaning, this PR addresses the fixes for most of the bug-fix issues.


cmd/context/delete.go
```diff
@@ -40,7 +40,7 @@ func DeleteCMD() *cobra.Command {
 	return cmd
 }
 
-func Delete(ctx context.Context, okCtx string) error {
+func Delete(_ context.Context, okCtx string) error {
```
jLopezbarb, line 43: Should we completely remove if we are not using the context on this?

`func Delete(okCtx string) error `

ifbyol: Agree, if the param is not being used and the function doesn't have to satisfy any interface, we should remove it then

pkg/ssh/config.go
```diff
@@ -87,9 +87,9 @@ func (h *host) String() string {
 
 }
 
-func newParam(keyword string, args, comments []string) *param {
+func newParam(keyword string, args, _ []string) *param {
```
jLopezbarb, line 90: Should we remove this param if it's always nil I don't see the point of having it

cmd/deploy/deploy_test.go
```diff
@@ -96,17 +96,17 @@ func (*fakeKubeConfig) Read() (*rest.Config, error) {
 func (fc *fakeKubeConfig) Modify(_ int, _, _ string) error {
 	return fc.errOnModify
 }
-func (*fakeKubeConfig) GetModifiedCMDAPIConfig() (*clientcmdapi.Config, error) {
```
jLopezbarb, line 99: Was this notation failing? I see it clearer than `func (_ *fakeKubeConfig) functionName`

jLopezbarb: Nvm I just saw that you changed it in the next commit 😅

cmd/context/delete.go
teresaromero: i was conservative in this sense, as this removing would lead to any other changes, i will add a commit with this removals so the review is clear

pkg/ssh/pool.go
```diff
@@ -125,7 +125,7 @@ func getTCPConnection(ctx context.Context, serverAddr string, keepAlive time.Dur
 	return c, nil
 }
 
-func getConn(ctx context.Context, serverAddr string, maxRetries int) (net.Conn, error) {
+func getConn(ctx context.Context, serverAddr string) (net.Conn, error) {
 	var lastErr error
 	t := time.NewTicker(100 * time.Millisecond)
 	for i := 0; i < 5; i++ {
```
AgustinRamiroDiaz, line 131: Shouldn't we use maxRetries here?

AgustinRamiroDiaz (approved):
:100:  Love it! Just a minor comment :D

pkg/ssh/pool.go
jLopezbarb: Agree with @AgustinRamiroDiaz, removing a magic number is always a good solution

jLopezbarb (approved):

pkg/ssh/pool.go
teresaromero: i've reverted the change by a param `retries` and calling the function with `const defaultRetries = 5` WDYT?

jLopezbarb (approved):

Issue #2732, opened 2022-05-30
images were built message when I don't have any build image set


jLopezbarb:
**Describe the bug**
When running an `okteto deploy` command from a helm chart or a manifest v2 without build section the following message appears:
`Images were already built. To rebuild your images run 'okteto build' or 'okteto deploy --build'`
<img width="691" alt="Captura de pantalla 2022-05-30 a las 21 31 28" src="https://user-images.githubusercontent.com/25170843/171050992-99d3f1ed-7a37-40d4-9120-e9e19af3f87f.png">

**To Reproduce**

1. Clone https://github.com/okteto/movies-with-helm
2. Run `okteto deploy`
3. See message

**Expected behavior**
If the build section doesn't have any service to build we shouldn't be showing the message

**Desktop (please complete the following information):**

- OS: macOs
- Version 2.3.0

Issue #2450, opened 2022-03-29
Review `okteto deploy`and `okteto up` logs


pchico83:
The output of `okteto deploy` and `okteto up` is very important to help developers understand what's happening when they run these commands. In CLI 2.0 we have changed many of these logs.

We should analyze these logs and validate them with developer interviews.

jLopezbarb:
Closing this since we merged the PR

PR #3133, opened 2022-09-27
Review info messages shown at the beginning of each command

pchico83:

# Proposed changes

Fixes https://github.com/okteto/okteto/issues/2732 https://github.com/okteto/okteto/issues/2450

I made a general pass to our info messages at the beginning of each command execution and the errors messages we have when commands fail. I will show the changes with screenshots for each modification, let me know what you think :-)

## Okteto Build

Before:
Screenshot 2022-09-27 at 23 21 22

After (less redundant and more accurate about the Dockerfile being built):
Screenshot 2022-09-27 at 23 22 51

## Okteto Deploy with an error

Before:
Screenshot 2022-09-28 at 00 04 07

After (help message was misleading, users might need to just retry):
Screenshot 2022-09-28 at 00 04 26

## Okteto Destroy with an error

Before:
Screenshot 2022-09-28 at 00 02 55

After (less redundant):
Screenshot 2022-09-28 at 00 03 14

## Okteto Up deploying `deploy` commands

Before:
Screenshot 2022-09-28 at 00 06 07

After (less noisy):
Screenshot 2022-09-28 at 00 06 50

## Okteto Up with deploy commands already executed

Before:
Screenshot 2022-09-28 at 00 07 26

After (less noisy):
Screenshot 2022-09-28 at 13 30 56

### DIFF
--- a/cmd/build/v2/build.go
+++ b/cmd/build/v2/build.go
@@ -180,7 +180,6 @@ func (bc *OktetoBuilder) buildService(ctx context.Context, manifest *model.Manif
 }
 
 func (bc *OktetoBuilder) buildSvcFromDockerfile(ctx context.Context, manifest *model.Manifest, svcName string, options *types.BuildOptions) (string, error) {
-	oktetoLog.Information("Building image for service '%s'", svcName)
 	isStackManifest := manifest.Type == model.StackType
 	buildSvcInfo := getBuildInfoWithoutVolumeMounts(manifest.Build[svcName], isStackManifest)
 
--- a/cmd/build/v2/build.go
+++ b/cmd/build/v2/build.go
@@ -180,7 +180,7 @@ func (bc *OktetoBuilder) buildService(ctx context.Context, manifest *model.Manif
 }
 
 func (bc *OktetoBuilder) buildSvcFromDockerfile(ctx context.Context, manifest *model.Manifest, svcName string, options *types.BuildOptions) (string, error) {
-	oktetoLog.Info("Building image for service '%s'", svcName)
+	oktetoLog.Infof("Building image for service '%s'", svcName)
 	isStackManifest := manifest.Type == model.StackType
 	buildSvcInfo := getBuildInfoWithoutVolumeMounts(manifest.Build[svcName], isStackManifest)
 
--- a/cmd/build/v2/build.go
+++ b/cmd/build/v2/build.go
@@ -180,6 +180,7 @@ func (bc *OktetoBuilder) buildService(ctx context.Context, manifest *model.Manif
 }
 
 func (bc *OktetoBuilder) buildSvcFromDockerfile(ctx context.Context, manifest *model.Manifest, svcName string, options *types.BuildOptions) (string, error) {
+	oktetoLog.Info("Building image for service '%s'", svcName)
 	isStackManifest := manifest.Type == model.StackType
 	buildSvcInfo := getBuildInfoWithoutVolumeMounts(manifest.Build[svcName], isStackManifest)
 
--- a/cmd/build/v2/services.go
+++ b/cmd/build/v2/services.go
@@ -47,7 +47,6 @@ func (bc *OktetoBuilder) GetServicesToBuild(ctx context.Context, manifest *model
 	close(toBuild)
 
 	if len(toBuild) == 0 {
-		oktetoLog.Information("Images were already built. To rebuild your images run 'okteto build' or 'okteto deploy --build'")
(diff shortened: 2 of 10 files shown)


rberrelleza:
Beautiful 🤩 

I don't get the change on the "Okteto Up with deploy commands already executed" section though.

cmd/build/v2/services.go
```diff
@@ -47,7 +47,6 @@ func (bc *OktetoBuilder) GetServicesToBuild(ctx context.Context, manifest *model
 	close(toBuild)
 
 	if len(toBuild) == 0 {
-		oktetoLog.Information("Images were already built. To rebuild your images run 'okteto build' or 'okteto deploy --build'")
```
ifbyol, line 50: This one is the only one I think we can miss. This can guide/help the users to find out why the latest changes of a service are not being deployed

rlamana:
We are basically here removing most of the hints on how to use okteto, and for some I think we should keep them:

- how to rebuild when images are already built. (I think this one is important and we should keep it)
- how to redeploy when user is doing up (I can live with this one)
- how to do `okteto up [name]` when the list is provided (I think we can keep this one and be less verbose with the rest of output we displayed)

cmd/build/v2/build.go
```diff
@@ -180,7 +180,6 @@ func (bc *OktetoBuilder) buildService(ctx context.Context, manifest *model.Manif
 }
 
 func (bc *OktetoBuilder) buildSvcFromDockerfile(ctx context.Context, manifest *model.Manifest, svcName string, options *types.BuildOptions) (string, error) {
-	oktetoLog.Information("Building image for service '%s'", svcName)
```
jLopezbarb, line 183: I think we should at least log it into `oktetoLog.Info`. That way we can have a better doctor file

cmd/build/v2/services.go
jLopezbarb: I agree with nacho about having this message. I don't think we can miss this message without affecting our users

RinkiyaKeDad:
> I don't get the change on the "Okteto Up with deploy commands already executed" section though.

I think it's just removing the "dev env has already been deployed message". If that's the case I think we should still keep that message because that's useful information.

rlamana:
It could be interesting to differentiate visually the "info" messages and the "hint" messages, maybe we can create this new type and display it in a grey color to avoid creating too much noise, wdyt?

pchico83:
> We are basically here removing most of the hints on how to use okteto, and for some I think we should keep them:
> 
> * how to rebuild when images are already built. (I think this one is important and we should keep it)
> * how to redeploy when user is doing up (I can live with this one)
> * how to do `okteto up [name]` when the list is provided (I think we can keep this one and be less verbose with the rest of output we displayed)

pchico83:
@rberrelleza @rlamana @RinkiyaKeDad @jLopezbarb @ifbyol based on all your feedback, I think the messages you would like to keep are:
- when skipping images build on `okteto deploy` a message saying how to rebuild the images
- when skipping deploy on `okteto up`, a message how to rerun the deploy commands
Is that ok? am I missing something else?

ifbyol:
> @rberrelleza @rlamana @RinkiyaKeDad @jLopezbarb @ifbyol based on all your feedback, I think the messages you would like to keep are:
> 
> * when skipping images build on `okteto deploy` a message saying how to rebuild the images
> * when skipping deploy on `okteto up`, a message how to rerun the deploy commands
>   Is that ok? am I missing something else?

That's correct!

pchico83:
I updated the description of the PR based on feedback.
1) keep the current behavior when skipping image build.
2) update the message when skipping deploy. The new one looks like this:
Screenshot 2022-09-28 at 13 30 56

cmd/build/v2/build.go
```diff
@@ -180,7 +180,7 @@ func (bc *OktetoBuilder) buildService(ctx context.Context, manifest *model.Manif
...
+	oktetoLog.Info("Building image for service '%s'", svcName)
```
jLopezbarb, line 183: Use infof instead to format the string
```suggestion
	oktetoLog.Infof("Building image for service '%s'", svcName)
```

jLopezbarb (approved):

ifbyol (approved):

AdrianPedriza:
It may be clearer to read the output for the user if we put the commands in quotation marks. For example the output proposed in the `Okteto Up deploying deploy commands` section shows: `Running helm upgrade --install.... `, maybe it improves the reading between quotes the command: `Running 'helm upgrade --install...'`.

PR #3140, opened 2022-09-30
add okteto healthchecks to compose

### LINKED ISSUE #3135: Support the configuration of liveness probes when using compose healthchecks
Okteto supports Docker Compose [healthchecks](https://www.okteto.com/docs/reference/compose/#healthcheck-object-optional) to be assigned to any service in your Docker Compose file. Healthchecks are translated to Kubernetes Readiness Probes.

In some cases, users would prefer to create Kubernetes Liveness Probes. Or both of them, Readiness and Liveness Probes.

Make it configurable to create Liveness and/or Readiness Probes when using health checks.

Support the following fields in the `healthchecks` section:

- `x-okteto-readiness`: boolean (default `true`)
- `x-okteto-liveness`: boolean (default `false`)

If both are false, we should fail validation of the Docker Compose file.
Otherwise, create Liveness/Readiness Probes based on the values of these fields.

**Additional context**
Update the docs to document these new fields

jLopezbarb:
# Proposed changes

Fixes #3135 

- Added `x-okteto-readiness` and `x-okteto-liveness` to the compose healthcheck section
- Add serialisation of those new fields, without modifying the current behavior (added tests to get that expected behavior)
- Add the translation from compose healthcheck to k8s healthcheck. Check that the current implementation is the same by doing tests

### DIFF
--- a/pkg/cmd/stack/translate.go
+++ b/pkg/cmd/stack/translate.go
@@ -116,7 +116,8 @@ func translateConfigMap(s *model.Stack) *apiv1.ConfigMap {
 func translateDeployment(svcName string, s *model.Stack) *appsv1.Deployment {
 	svc := s.Services[svcName]
 
-	healthcheckProbe := getSvcProbe(svc)
+	svcHealthchecks := getSvcProbe(svc)
+
 	return &appsv1.Deployment{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:        svcName,
@@ -148,7 +149,8 @@ func translateDeployment(svcName string, s *model.Stack) *appsv1.Deployment {
 							SecurityContext: translateSecurityContext(svc),
 							Resources:       translateResources(svc),
 							WorkingDir:      svc.Workdir,
-							ReadinessProbe:  healthcheckProbe,
+							ReadinessProbe:  svcHealthchecks.readiness,
+							LivenessProbe:   svcHealthchecks.liveness,
 						},
 					},
 				},
@@ -184,7 +186,7 @@ func translateStatefulSet(svcName string, s *model.Stack) *appsv1.StatefulSet {
 	svc := s.Services[svcName]
 
 	initContainers := getInitContainers(svcName, s)
-	healthcheckProbe := getSvcProbe(svc)
+	svcHealthchecks := getSvcProbe(svc)
 
 	return &appsv1.StatefulSet{
 		ObjectMeta: metav1.ObjectMeta{
@@ -223,7 +225,8 @@ func translateStatefulSet(svcName string, s *model.Stack) *appsv1.StatefulSet {
 							VolumeMounts:    translateVolumeMounts(svc),
 							Resources:       translateResources(svc),
 							WorkingDir:      svc.Workdir,
-							ReadinessProbe:  healthcheckProbe,
+							ReadinessProbe:  svcHealthchecks.readiness,
+							LivenessProbe:   svcHealthchecks.liveness,
 						},
 					},
 				},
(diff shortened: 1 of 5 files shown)


pkg/cmd/stack/translate.go
```diff
@@ -116,7 +116,8 @@ func translateConfigMap(s *model.Stack) *apiv1.ConfigMap {
 func translateDeployment(svcName string, s *model.Stack) *appsv1.Deployment {
 	svc := s.Services[svcName]
 
-	healthcheckProbe := getSvcProbe(svc)
+	svcHealthchecks := getSvcProbe(svc)
```
AdrianPedriza, line 119: nit: `svcProbes`? to fit with the function name that return the values.

pkg/cmd/stack/translate.go
```diff
@@ -680,7 +684,13 @@ func translateResources(svc *model.Service) apiv1.ResourceRequirements {
 	return result
 }
 
-func getSvcProbe(svc *model.Service) *apiv1.Probe {
+type healtcheckProbes struct {
```
AdrianPedriza, line 687: typo: `healthcheckProbes` intead `heatlhcheckProbes`

pkg/model/stack_serializer_test.go
```diff
@@ -359,31 +359,55 @@ func Test_HealthcheckUnmarshalling(t *testing.T) {
...
+			expectedError: false,
+		},
+		{
+			name:          "healthcheck readiness=false liveness=true",
+			manifest:      []byte("services:\n  app:\n    healthcheck:\n      interval: 10s\n      x-okteto-readiness: false\n      x-okteto-liveness: true\n      timeout: 10m\n      retries: 5\n      start_period: 30s\n      test: curl --fail 0.0.0.0:8080\n    image: okteto/vote:1"),
+			expected:      &HealthCheck{HTTP: &HTTPHealtcheck{Path: "/", Port: 8080}, Interval: 10 * time.Second, Timeout: 10 * time.Minute, Retries: 5, StartPeriod: 30 * time.Second, Test: []string{}, Readiness: false, Liveness: true},
+			expectedError: false,
+		},
+		{
+			name:          "healthcheck readiness=false liveness=false",
```
AdrianPedriza, line 396: what do you think to add a test where `x-okteto-readiness: false` and `x-okteto-liveness` is not set? we should have an error like test below

pkg/cmd/stack/translate.go
AdrianPedriza, line 119: maybe we should either change function name or change variable name? I would go to rename function name to `getSvcHealthProbes`

teresaromero (approved):

AdrianPedriza (approved):

pkg/cmd/stack/translate.go
```diff
@@ -680,7 +684,13 @@ func translateResources(svc *model.Service) apiv1.ResourceRequirements {
 	return result
 }
 
-func getSvcProbe(svc *model.Service) *apiv1.Probe {
+type healthcheckProbes struct {
+	readiness *apiv1.Probe
+	liveness  *apiv1.Probe
+}
+
+func getSvcHealthProbe(svc *model.Service) *healthcheckProbes {
```
ifbyol, line 692: If this function cannot retrieve a nil, I wouldn't return a pointer. There are a potential nil pointer in all places were the result of this function is used. Nothing prevents that in a future change we return a nil within this function and then, we fail in a non-controlled way

ifbyol (commented):
The liveness and readiness probe will be always the same right? we won't allow for now different configuration for each of them, right?

jLopezbarb:
> The liveness and readiness probe will be always the same right? we won't allow for now different configuration for each of them, right?

Yes, they will be the same

PR #3142, opened 2022-10-03
added error status when destroying to the manifest

### LINKED ISSUE #3132: If `okteto destroy` fails, the Okteto UI shows `destroying` status forever
**Describe the bug**
Run `okteto destroy` with the following manifest:
```
deploy:
  - echo "deploying...."
destroy:
  - this is an error
```
check the Okteto UI and the dev environment keeps the `Destroying` state forever.  

**Expected behavior**
The development environment is marked as `error`

mnevadom:
Fixes #3132

### DIFF
--- a/cmd/destroy/cfghandler.go
+++ b/cmd/destroy/cfghandler.go
@@ -70,6 +70,7 @@ func (ch *defaultConfigMapHandler) destroyConfigMap(ctx context.Context, cfg *ap
 
 func (ch *defaultConfigMapHandler) setErrorStatus(ctx context.Context, cfg *apiv1.ConfigMap, data *pipeline.CfgData, err error) error {
 	oktetoLog.AddToBuffer(oktetoLog.InfoLevel, "Destruction failed: %s", err.Error())
+	data.Status = pipeline.ErrorStatus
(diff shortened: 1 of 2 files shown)


ifbyol:
The error still happens in 2 cases:
* If I cut the execution with ctrl + c when commands are being executed, the status `destroying` is shown forever
* If some dependency fails to be destroy, the status `destroying` is shown forever

If we fail the execution of the command, I think we should also update the status. Can you confirm @pchico83 ?


@mnevadom BTW, can you sign your commits in the CLI and include the proper `release/...` label to the PR, please?

pchico83:
> The error still happens in 2 cases:
> 
> * If I cut the execution with ctrl + c when commands are being executed, the status `destroying` is shown forever
> * If some dependency fails to be destroy, the status `destroying` is shown forever
> 
> If we fail the execution of the command, I think we should also update the status. Can you confirm @pchico83 ?
> 
> @mnevadom BTW, can you sign your commits in the CLI and include the proper `release/...` label to the PR, please?

Yes, we should change the state in the stop signal handler

ifbyol (approved):

PR #3144, opened 2022-10-04
Add "okteto@" as a prefix in OKTETO_BUILD_XXX_SHA

pchico83:

# Proposed changes

The current value of `OKTETO_BUILD_XXX_TAG` is of the form `sha256:xxxxxx`. 
This value cannot be used directly as a tag. For example, if you have a helm chart like this:
```
image: {{ .Values.repository}}:{{ .Values.tag}}
```
`.Values.tag` must be `okteto@${OKTETO_BUILD_XXX_TAG}` instead of just `OKTETO_BUILD_XXX_TAG`, which is very hard to discover by a user.

This PR is adding a new envvar `OKTETO_BUILD_XXX_SHA` whose value is `okteto@${OKTETO_BUILD_XXX_TAG}`. `OKTETO_BUILD_XXX_SHA` will be documented in favor of `OKTETO_BUILD_XXX_TAG`, but `OKTETO_BUILD_XXX_TAG` will keep the current behavior to avoid breaking changes

### DIFF
--- a/cmd/build/v2/environment.go
+++ b/cmd/build/v2/environment.go
@@ -19,6 +19,7 @@ import (
 	"strings"
 
 	oktetoLog "github.com/okteto/okteto/pkg/log"
+	"github.com/okteto/okteto/pkg/model"
 	"github.com/okteto/okteto/pkg/registry"
 )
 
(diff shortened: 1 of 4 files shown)


cmd/build/v2/environment.go
```diff
@@ -55,5 +55,15 @@ func (bc *OktetoBuilder) SetServiceEnvVars(service, reference string) {
 	os.Setenv(tagKey, tag)
 	bc.lock.Unlock()
 
+	sha := tag
+	if strings.HasPrefix(sha, "sha256:") {
+		sha = fmt.Sprintf("okteto@%s", sha)
```
ifbyol, line 60: Should we use [this constant](https://github.com/okteto/okteto/blob/master/pkg/model/const.go#L329) instead of harcoding the `okteto` value? 

If I'm not wrong, we have to use the name of the tag that was used when building the image, right? Or can we set any value there?

pchico83: You can set any arbitrary string. what's not supported is this:
```
my-image:sha256xxx
```
You need to do either  `my-image@sha256xxx` or `my-image:anything@sha256xxx`

teresaromero (approved):

jLopezbarb (approved):

PR #3154, opened 2022-10-10
Update message for untrusted certificates

pchico83:

# Proposed changes

Update message for untrusted certificates. We shouldn't assume the error is just for self-signed certificates


ifbyol (approved):
@pchico83  Release branch was already created but it wasn't released yet. If you want to include it as part of the release, you would have to do the backport to `release-2.8`

jLopezbarb (approved):

PR #3158, opened 2022-10-11
fix build when is buildv1 but dockerfile need to be inferred

### LINKED ISSUE #3157: `okteto build` doesn't work if an okteto manifest is not provided/found
**Describe the bug**
If I have a folder with a Dockerfile but I don't have any Okteto manifest, when I try to build the image with `okteto build -t <my-tag> .`, it fails with the following error:

```
 x  Could not detect any okteto manifest
    If you have an okteto manifest file, use the flag '--file' to point to your okteto manifest file

```

It stopped working on `2.7.0`

**To Reproduce**

1. Create an empty folder with a simple Docker inside it
2. Execute `okteto build -t <my-tag> .`
3. The error commented in previous section is displayed and the image is not built

**Expected behavior**
It should build the image without asking for an Okteto manifest and should behave as it was prior `2.7.0`

- Version: `2.7.0` and master

AdrianPedriza:

# Proposed changes

Fixes #3157

- instead of propagating the `ErrOktetoManifestNotFound` error, we control it to allow the context to be loaded even if it cannot be obtained from the file given by the user (in this case, the user does not indicate any). Subsequently, after finding no manifest a fallback to builder v1 will be made.
- added test case covering this scenario

### DIFF
--- a/cmd/build/command.go
+++ b/cmd/build/command.go
@@ -28,6 +28,7 @@ import (
 	"github.com/okteto/okteto/cmd/namespace"
 	"github.com/okteto/okteto/cmd/utils"
 	"github.com/okteto/okteto/pkg/cmd/build"
+	"github.com/okteto/okteto/pkg/discovery"
 	oktetoErrors "github.com/okteto/okteto/pkg/errors"
 	oktetoLog "github.com/okteto/okteto/pkg/log"
 	"github.com/okteto/okteto/pkg/model"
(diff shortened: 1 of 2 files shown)


integration/build/build_test.go
```diff
@@ -130,6 +130,17 @@ func TestBuildCommandV1(t *testing.T) {
 	}
 	require.NoError(t, commands.RunOktetoBuild(oktetoPath, options))
 	require.True(t, isImageBuilt(expectedImage))
+
+	options = &commands.BuildOptions{
```
jLopezbarb, line 134: Can you move this to a new test function. We are parallelising by tests functions so if we add it in another one we will run the test in parallel

jLopezbarb (approved):

cmd/build/command.go
```diff
@@ -164,19 +165,21 @@ func (*Command) loadContext(ctx context.Context, options *types.BuildOptions) er
...
-		if err != nil {
+		if err != nil && !errors.Is(err, discovery.ErrOktetoManifestNotFound) {
 			return err
 		}
 
-		if err := ctxResource.UpdateNamespace(options.Namespace); err != nil {
-			return err
-		}
-		ctxOpts.Namespace = ctxResource.Namespace
+		if ctxResource != nil {
```
ifbyol, line 172: Can we include a comment to specify why if the `ctxResource` is nil we keep going? There are some hidden knowledge behind this

ifbyol (approved):
Minor comment, but it looks good!

PR #3162, opened 2022-10-16
Support both string and sequence of string in a manifest's `cache_from` property

### LINKED ISSUE #2908: Support string syntax for `cache_from`
We do support this syntax for the `build` section:

```
cache_from:
   - okteto.global/gonfalon:okteto
```

but should also support a single string syntax like this:
``` 
cache_from: okteto.global/gonfalon:okteto
```

which is more user friendly and covers 95% of the use cases

volmedo:
# Proposed changes

Fixes #2908 

To allow the `cache_from` property to support both alternatives, the `CacheFrom` field's type is changed to `interface{}` and custom conversions are implemented during marshalling and unmarshalling.

### DIFF
--- a/pkg/model/dev.go
+++ b/pkg/model/dev.go
@@ -121,7 +121,7 @@ type BuildInfo struct {
 	Name             string         `yaml:"name,omitempty"`
 	Context          string         `yaml:"context,omitempty"`
 	Dockerfile       string         `yaml:"dockerfile,omitempty"`
-	CacheFrom        []string       `yaml:"cache_from,omitempty"`
+	CacheFrom        CacheFrom      `yaml:"cache_from,omitempty"`
 	Target           string         `yaml:"target,omitempty"`
 	Args             BuildArgs      `yaml:"args,omitempty"`
 	Image            string         `yaml:"image,omitempty"`
@@ -130,6 +130,9 @@ type BuildInfo struct {
 	DependsOn        BuildDependsOn `yaml:"depends_on,omitempty"`
 	Secrets          BuildSecrets   `yaml:"secrets,omitempty"`
 }
+
+// CacheFrom is a list of images to import cache from.
+type CacheFrom []string
 
 // BuildArg is an argument used on the build step.
 type BuildArg struct {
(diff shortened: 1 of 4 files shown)


volmedo:
I assume we also want to support this syntax in [stack manifests](https://github.com/volmedo/okteto/blob/a194a8e4d60461f3d671a7de86f1a09ff3fca254/pkg/model/stack_serializer.go#L243), don't we?

AgustinRamiroDiaz:
Hi @volmedo ! Thanks a lot for your submission :heart: 

I'm not aware of all the caveats of using `interface{}` and type casting. 

I suggest you check out how we do our other serializing [like with EnvFiles](https://github.com/okteto/okteto/blob/d11be68fbad723452f9088c9c56831774e7e6b42/pkg/model/serializer.go#L1275). A quick summary: we create our custom type aliases so we are able to define our custom unmarshalling, and there we try to unmarshall it into the different structures we'd like (`string` and `[]string` for example).

You could also research if there's a generic way to implement this with generics, since now we use go 1.18

volmedo:
hey @AgustinRamiroDiaz! Thanks for the suggestion, the code looks better with the custom type 🙂.

I don't really think generics are a great fit for this use case. There isn't really some code that we could reuse by adding parametric types. And these types won't be known at build time but at run time, when marshalling and unmarshalling take place.

volmedo:
btw, I'm quite certain the answer to [my previous question](https://github.com/okteto/okteto/pull/3162#issuecomment-1280070796) is affirmative and I'm using the `CacheFrom` type in `composeBuildInfo` too so that stacks manifests also support the single string syntax for `cache_from`.

If this is not correct please let me know 🙏🏻.

jLopezbarb (approved):

teresaromero (approved):

AgustinRamiroDiaz (approved):

volmedo:
is it possible that the warning thrown by CircleCI is indeed true?

image

jLopezbarb:
> is it possible that the warning thrown by CircleCI is indeed true?
> 
> image

I don't think so, It's only appearing for external contributors, the PRs that are opened by maintainers don't have the warning

jLopezbarb:
> > is it possible that the warning thrown by CircleCI is indeed true?
> > image
> 
> I don't think so, It's only appearing for external contributors, the PRs that are opened by maintainers don't have the warning

The team has verified that it was a mistake CircleCI made and there should be no problem with the merge. Thanks again for your PR

volmedo:
> The team has verified that it was a mistake CircleCI made and there should be no problem with the merge. Thanks again for your PR

happy to help! 🎉

PR #3172, opened 2022-10-18
fix: use the correct param on the API call

### LINKED ISSUE #3171: preview is not passing the proper filename if specified by file
**Describe the bug**
If you run `okteto preview -f docker-compose.yml` or any other filename the okteto preview will call the mutation but won't add the filename parameter, so the backend won't have the filename property and will leave it empty when calling the `okteto deploy` command inside.

**To Reproduce**

1. Run `okteto preview deploy --wait -f docker-compose.yml`
1. Get the preview space
1. Check that the generated configmap doesn't have filename property set: `kubectl get configmap -n {preview-name} -oyaml`
1. See that `data.filename` is empty

**Expected behavior**
It should be propagated to the backend

- Version 2.8.0

jLopezbarb:

# Proposed changes

Fixes #3171

As part of the preview refactor, where we have the following function signature `executeDeployPreview(ctx context.Context, name, scope, repository, branch, sourceUrl, filename string, variables []types.Variable)` we changed it to -> `deployPreview(ctx context.Context, opts *DeployOptions) (*types.PreviewResponse, error)` Since opts have `filename` and `file` (`filename` was deprecated) we used `opts.filename` instead of `opts.file` which was the one that should be used in the graphql mutation


cmd/preview/deploy.go
```diff
@@ -137,7 +137,7 @@ func (pw *Command) deployPreview(ctx context.Context, opts *DeployOptions) (*typ
 		})
 	}
 
-	return pw.okClient.Previews().DeployPreview(ctx, opts.name, opts.scope, opts.repository, opts.branch, opts.sourceUrl, opts.filename, varList)
+	return pw.okClient.Previews().DeployPreview(ctx, opts.name, opts.scope, opts.repository, opts.branch, opts.sourceUrl, opts.file, varList)
```
rberrelleza, line 140: please also add a unit test to ensure we catch any regressions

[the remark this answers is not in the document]
jLopezbarb: Done in [d0e0a45](https://github.com/okteto/okteto/pull/3172/commits/d0e0a45f52285561332cc057e4888ac0d4ab4e98)

jmacelroy (approved):

cmd/preview/deploy.go
jLopezbarb: There is no easy way to test it... There are two different functions that already have all tests done correctly. The problem here was more that we were not using the correct parameter. We could create a test that tests both functions one after the other but we won't get if the cfmap is getting the correct filename since is something done in the backend side

jmacelroy: There's probably an integration test that we can look at but that can be follow-up as well.

PR #3189, opened 2022-10-21
fix: add error on init container volume full

### LINKED ISSUE #2214: My `okteto up` sometimes fails with "failed to prepare subPath for volumeMount` error
**Describe the bug**
Issue looks temporal, it isusally works when I run `okteto up` again.

**Describe the solution you would like**
 
Ignore/retry these errors for 2 minutes

jLopezbarb:

# Proposed changes

Fixes #2214

- Check if fails to mount volume because is full and return the proper error

Testing `waitUntilDevelopmentContainerIsRunning` is very complex and can hide the changes done in this PR. It's a 150 lines function  that needs to inject updateFile/kubernetes interface and a go routine in the test that updates the pod and events. Should we do it as part of this PR or create a tech debt and schedule it for a following sprint since it will take a lot of time injecting all the dependencies needed? cc @jmacelroy @ifbyol

### DIFF
--- a/cmd/up/activate.go
+++ b/cmd/up/activate.go
@@ -373,6 +373,14 @@ func (up *upContext) waitUntilDevelopmentContainerIsRunning(ctx context.Context,
 					oktetoLog.Spinner("Insufficient cpu/memory in the cluster. Waiting for new nodes to come up...")
 					continue
 				}
+				if strings.Contains(e.Message, "failed to create subPath directory") {
+					return oktetoErrors.UserError{
+						E: fmt.Errorf("There is no space left in persistent volume"),
+						Hint: fmt.Sprintf(`Okteto volume is full.
+    Increase your persistent volume size, run '%s' and try 'okteto up' again.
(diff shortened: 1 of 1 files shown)


ifbyol:
The real change can be done checking just the first commit on this PR and it is really simple. So, I would try to include it as part of this PR. If it takes too much time or it is more complex that it looked at the beginning, lets create another ticket, but that function surely needs a refactor. But, we have to be careful, we should make sure there is no any change of behavior or regression introduced as part of the refactor

ifbyol:
> > Adding them to the PR looks better to me. The parent issue is not a critical fix.
> 
> > The real change can be done checking just the first commit on this PR and it is really simple. So, I would try to include it as part of this PR. If it takes too much time or it is more complex that it looked at the beginning, lets create another ticket, but that function surely needs a refactor. But, we have to be careful, we should make sure there is no any change of behavior or regression introduced as part of the refactor
> 
> It's a big change(modifies many files), I prefer to create a new issue and introduce it at the beginning of a new CLI release, that way we can verify that a refactor that adds tests won't introduce any new bug

Ok, lets do it in a separate issue

ifbyol (approved):

PR #3190, opened 2022-10-21
feat: add specific timeout to dependencies

### LINKED ISSUE #3146: Add a timeout for specific dependencies
I have a bunch of remote dependencies, and I want to specific a custom timeout for each of them

I would like a new field on the dependencies section that adds the ability to set the timeout for the dependencies:
```yaml
dependencies:
  - api:
     repository: https://github/test/api
     timeout: 10m
  - frontend:
     repository: https://github/test/frontend
     timeout: 5m
```

**Additional context**
If the timeout is not specified in the dependency, we should be using the timeout flag of the deploy command
If the timeout is not specified in the dependency nor in the timeout flag. We should be using the `OKTETO_TIMEOUT` environment variable

jLopezbarb:

# Proposed changes

Fixes #3146

- Add `getDefaultTimeout` to check if the `OKTETO_TIMEOUT` environment variable is set. Set the timeout flag default value as follows: 1. User set flag 2. `OKTETO_TIMEOUT` environment variable 3. default value (5min)
- refactor dependency serializer: I added a new type and removed a copy paste of dependency struct called `dependencyRaw`. With this refactor If we update any dependency field we don't  have to update it in the `dependencyRaw` struct.
- Added tests for dependency unmarshaller.
- Added a new dependency field `timeout` and a new exported method `GetTimeout` that returns the dependency specific timeout if it's set or the one added as argument if not
- Called the new `dependency.GetTimeout` instead of directly the timeout flag when we are creating a dependency

### DIFF
--- a/pkg/model/manifest_test.go
+++ b/pkg/model/manifest_test.go
@@ -1014,3 +1014,38 @@ func Test_SanitizeSvcNames(t *testing.T) {
 		})
 	}
 }
+
+func Test_GetTimeout(t *testing.T) {
+	tests := []struct {
+		name           string
+		defaultTimeout time.Duration
+		dependency     *Dependency
+		expected       time.Duration
+	}{
+		{
+			name:           "default timeout set and specific not",
+			defaultTimeout: 5 * time.Minute,
+			dependency:     &Dependency{},
+			expected:       5 * time.Minute,
+		},
+		{
+			name: "default timeout unset and specific set",
+			dependency: &Dependency{
+				Timeout: 10 * time.Minute,
+			},
+			expected: 10 * time.Minute,
+		},
+		{
+			name:       "both unset",
+			dependency: &Dependency{},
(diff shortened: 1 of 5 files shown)


teresaromero (approved):

pkg/model/manifest_test.go
```diff
@@ -1014,3 +1014,38 @@ func Test_SanitizeSvcNames(t *testing.T) {
...
+			name: "default timeout unset and specific set",
+			dependency: &Dependency{
+				Timeout: 10 * time.Minute,
+			},
+			expected: 10 * time.Minute,
+		},
+		{
+			name:       "both unset",
+			dependency: &Dependency{},
+			expected:   0,
```
AdrianPedriza, line 1041: Shouldn't I fallback to 5 min?

AdrianPedriza (approved):

PR #3192, opened 2022-10-22
Make AskYesNo function support default answers

### LINKED ISSUE #2793: okteto questions should default to yes
**Describe the bug**
When I install any package and I get asked to accept. I usually just press enter and it works. It seems that okteto you always need to type `y + Enter`

**To Reproduce**

1. Run okteto up on an empty dir
2. When you get ` ?  okteto manifest (okteto.yml) doesn't exist, do you want to create it? [y/n]` press enter
3. See error

**Expected behavior**
I would like that empty line should default to yes

- OS: macOS
- Version 2.4.beta-1

volmedo:
# Proposed changes

Fixes #2793

I added a new parameter to `utils.AskYesNo` function that enables the caller to set a default answer. That allows extra flexibility, as different questions could potentially have different defaults.

However, it seems that all the questions that exist today have the same sensible default of "yes". If you don't think there will ever be a question where the default could be "no", the defaults can be removed altogether to simplify the implementation.

### DIFF
--- a/cmd/utils/dev.go
+++ b/cmd/utils/dev.go
@@ -223,13 +223,31 @@ func SelectDevFromManifest(manifest *model.Manifest, selector OktetoSelectorInte
 	return manifest.Dev[devKey], nil
 }
 
+// YesNoDefault specifies what will be assumed when the user doesn't answer explicitly
+type YesNoDefault string
+
+const (
+	YesNoDefault_Unspecified = "[y/n]"
+	YesNoDefault_Yes         = "[Y/n]"
+	YesNoDefault_No          = "[y/N]"
+)
+
 // AskYesNo prompts for yes/no confirmation
-func AskYesNo(q string) (bool, error) {
+func AskYesNo(q string, d YesNoDefault) (bool, error) {
 	var answer string
 	for {
-		oktetoLog.Question(q)
-		if _, err := fmt.Scanln(&answer); err != nil {
+		oktetoLog.Question(fmt.Sprintf("%s %s: ", q, d))
+		_, err := fmt.Scanln(&answer)
+		if err != nil && err.Error() != "unexpected newline" {
 			return false, err
+		}
+
+		if answer == "" && d != YesNoDefault_Unspecified {
+			answer = "y"
+			if d == YesNoDefault_No {
+				answer = "n"
+			}
+			break
 		}
 
 		if answer == "y" || answer == "n" {
@@ -277,7 +295,7 @@ func AskForOptions(options []string, label string) (string, error) {
 
 // AskIfOktetoInit asks if okteto init should be executed
 func AskIfOktetoInit(devPath string) bool {
-	result, err := AskYesNo(fmt.Sprintf("okteto manifest (%s) doesn't exist, do you want to create it? [y/n] ", devPath))
+	result, err := AskYesNo(fmt.Sprintf("okteto manifest (%s) doesn't exist, do you want to create it?", devPath), YesNoDefault_Yes)
 	if err != nil {
 		return false
 	}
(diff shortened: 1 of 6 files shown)


cmd/utils/dev.go
```diff
@@ -223,15 +223,33 @@ func SelectDevFromManifest(manifest *model.Manifest, selector OktetoSelectorInte
...
 // AskYesNo prompts for yes/no confirmation
-func AskYesNo(q string) (bool, error) {
+func AskYesNo(q string, d YesNoDefault) (bool, error) {
 	var answer string
 	for {
-		oktetoLog.Question(q)
-		if _, err := fmt.Scanln(&answer); err != nil {
+		oktetoLog.Question(fmt.Sprintf("%s %s: ", q, d))
+		_, err := fmt.Scanln(&answer)
+		if err != nil && err.Error() != "unexpected newline" {
```
volmedo, line 241: `fmt.Scanl` will try to store a value on every argument it gets passed, and will error out if it can't.

In this case we are passing a single argument, and an empty response from the user will cause an error as not even a single value can be scanned. Unfortunately, no sentinels or error constants are used in the `fmt` package so the only way to identify that error is to compare its string representation.

jLopezbarb:
Hi @volmedo, thanks for your contribution. 
Could you signoff the commit made in order to approve the PR?

Thanks in advance

volmedo:
hey @jLopezbarb! I fixed it as soon as I saw derek's comment. I amended the commit to add the sign-off, if you check it you will see it there. Also, derek removed the `no-dco` label, so I assume everything is ok. If it isn't, please let me know.

jLopezbarb (approved):

cmd/utils/dev_test.go
```diff
@@ -435,3 +436,60 @@ func Test_SelectDevFromManifest(t *testing.T) {
...
+			name:     "honors-default-when-no-answer",
+			def:      YesNoDefault_Yes,
+			answer:   "\n",
+			expected: true,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			// Create a temp file to mock stdin
+			tmpFile, err := ioutil.TempFile("", "yes_no_tests")
```
teresaromero, line 463: Deepsource pipeline for the PR is reporting `io/ioutil`  https://github.com/okteto/okteto/pull/3192/checks?check_run_id=9041048721 

what about creating a temp dir (`t.TempDir`), and writting here a file for this instead of using `ioutil.TempFile`. Here is an example: https://github.com/okteto/okteto/blob/master/cmd/build/v1/build_test.go#L131 that way it will be cleanup when the test is finished

cmd/utils/dev.go
```diff
@@ -223,15 +223,33 @@ func SelectDevFromManifest(manifest *model.Manifest, selector OktetoSelectorInte
...
+
 // AskYesNo prompts for yes/no confirmation
-func AskYesNo(q string) (bool, error) {
+func AskYesNo(q string, d YesNoDefault) (bool, error) {
 	var answer string
 	for {
-		oktetoLog.Question(q)
-		if _, err := fmt.Scanln(&answer); err != nil {
+		oktetoLog.Question(fmt.Sprintf("%s %s: ", q, d))
+		_, err := fmt.Scanln(&answer)
```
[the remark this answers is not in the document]
jLopezbarb: IMO I have mixed feelings about this and I think this change is a personal preference. 

I prefer to use it one line if there is only one check in the if statement but with more than one check I prefer to have it splitted in multiple lines. I found it more readable, but as I said I think this is a personal preference.

cmd/utils/dev_test.go
volmedo: thank you @teresaromero! TIL that `ioutil` has been deprecated and that `t.TempDir` exists 👍🏻

cmd/utils/dev.go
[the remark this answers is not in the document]
volmedo: yes, I agree with @jLopezbarb. I usually inline the call in the `if` when the function being called only returns the error and the check is just the usual `err != nil`. In this case we are not using the first value, but the function still returns two values, and the check has two conditions. I also feel it's more readable in two separate lines.

volmedo:
damn! sorry again derek! 😅

teresaromero (approved):

Issue #3102, opened 2022-09-14
Nil dereference in up command


AgustinRamiroDiaz:
I stumbled upon the code and I've seen that we would be dereferencing a nil pointer in the up command if we get an error from `okteto.GetK8sClient()`. The problem comes from `up.Dev` being nil at that point.

https://github.com/okteto/okteto/blob/9b88e269b657d95b04e978d963ae5d22ff4887b8/cmd/up/up.go#L221-L224

jLopezbarb:
Hi @adelsz, I assigned this issue to you since the PR that you opened fixes this issue

PR #3193, opened 2022-10-24
Fix error logging crash on context load failure

adelsz:
# Issue
For me running `okteto up` results in a the following panic:
```
[signal SIGSEGV: segmentation violation code=0x2 addr=0x58 pc=0x1063f58b0]

goroutine 1 [running]:
github.com/okteto/okteto/cmd/up.Up.func1(0x14000a6d900?, {0x1400014ae60, 0x0, 0x2})
        github.com/okteto/okteto/cmd/up/up.go:230 +0x10e0

https://github.com/okteto/okteto/blob/da425cf06ae7e0877a263574ef78fe96e0f1ab3d/cmd/up/up.go#L230
```
Looks like the error reporting code at line 230 always tries to dereference a nil pointer. `up.Dev` is always nil at that line (set to nil on line 212), so `up.Dev.Context` is never referenceble.

# Fix

A basic fix is to omit the context from the error message.

### DIFF
--- a/cmd/up/up.go
+++ b/cmd/up/up.go
@@ -227,7 +227,7 @@ func Up() *cobra.Command {
 			}
 			up.Client, up.RestConfig, err = okteto.GetK8sClient()
 			if err != nil {
-				return fmt.Errorf("failed to load okteto context '%s': %v", up.Dev.Context, err)
+				return fmt.Errorf("failed to load okteto context: %v", err)
 			}
 
 			// if manifest v1 - either set autocreate: true or pass --deploy (okteto forces autocreate: true)
--- a/cmd/up/up.go
+++ b/cmd/up/up.go
@@ -227,7 +227,7 @@ func Up() *cobra.Command {
 			}
 			up.Client, up.RestConfig, err = okteto.GetK8sClient()
 			if err != nil {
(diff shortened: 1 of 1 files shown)


jLopezbarb (approved):

jLopezbarb:
Hi @adelsz, Thanks for your contribution. Could you follow the steps specified [here](https://github.com/okteto/okteto/pull/3193#issuecomment-1289261688) and sign off your commit?

cmd/up/up.go
```diff
@@ -227,7 +227,7 @@ func Up() *cobra.Command {
 			}
 			up.Client, up.RestConfig, err = okteto.GetK8sClient()
 			if err != nil {
-				return fmt.Errorf("failed to load okteto context '%s': %v", up.Dev.Context, err)
+				return fmt.Errorf("failed to load okteto context: %v", err)
```
teresaromero, line 230:
```suggestion
				return fmt.Errorf("failed to load k8s client: %v", err)
```

the error could be the context load or getting the client, i think is better to change the string of the error too. whatever is returned by `okteto.GetK8sClient` will be added to the error.

jLopezbarb:
Fixes #3102

PR #3195, opened 2022-10-25
Add namespace for ctxOptions in UseContext() method

### LINKED ISSUE #3134: `namespace` is missconfigure when using `context` in my okteto manifest
**Describe the bug**
My current context is pointing to https://cloud.okteto.com to a different namespace than the personal namespace.
If I have a manifest with the `context` field like this:

```
context: https://cloud.okteto.com
```

and run any cli comand, the command is executed in my personal namespace. But it should use my current context namespace, because I am already working on https://cloud.okteto.com.

I didn't check when the context is a kubernetes context, but we should do the same. If the context in the okteto manifest is the same as the current okteto context, keep using the namespace in the current okteto context.

abs007:

# Proposed changes

Fixes #3134 

This pr sets the namespace for ctxOptions in the UseContext() method. This will set a value for ctxOptions.namespace and hence namespace won't be set to personal/user namespace.

### DIFF
--- a/cmd/context/create.go
+++ b/cmd/context/create.go
@@ -108,18 +108,16 @@ func (c *ContextCommand) UseContext(ctx context.Context, ctxOptions *ContextOpti
 
 	ctxStore := okteto.ContextStore()
 	if okCtx, ok := ctxStore.Contexts[ctxOptions.Context]; ok && okCtx.IsOkteto {
-		ctxOptions.Namespace = okCtx.Namespace
 		ctxOptions.IsOkteto = true
 	}
 
 	if okCtx, ok := ctxStore.Contexts[okteto.AddSchema(ctxOptions.Context)]; ok && okCtx.IsOkteto {
-		ctxOptions.Context = okteto.AddSchema(ctxOptions.Context) // Should this be changed to okCtx.Context?
+		ctxOptions.Context = okteto.AddSchema(ctxOptions.Context)
 		ctxOptions.Namespace = okCtx.Namespace
 		ctxOptions.IsOkteto = true
 	}
 
 	if ctxOptions.Context == okteto.CloudURL {
-		ctxOptions.Namespace = ctxStore.Contexts[ctxOptions.Context].Namespace
 		ctxOptions.IsOkteto = true
 	}
 
@@ -127,7 +125,6 @@ func (c *ContextCommand) UseContext(ctx context.Context, ctxOptions *ContextOpti
 
 		if isUrl(ctxOptions.Context) {
 			ctxOptions.Context = strings.TrimSuffix(ctxOptions.Context, "/")
-			ctxOptions.Namespace = ctxStore.Contexts[ctxOptions.Context].Namespace
 			ctxOptions.IsOkteto = true
 		} else {
 			if !isValidCluster(ctxOptions.Context) {
@@ -137,7 +134,6 @@ func (c *ContextCommand) UseContext(ctx context.Context, ctxOptions *ContextOpti
 			transformedCtx := okteto.K8sContextToOktetoUrl(ctx, ctxOptions.Context, ctxOptions.Namespace, c.K8sClientProvider)
 			if transformedCtx != ctxOptions.Context {
 				ctxOptions.Context = transformedCtx
-				ctxOptions.Namespace = ctxStore.Contexts[ctxOptions.Context].Namespace
 				ctxOptions.IsOkteto = true
 			}
 		}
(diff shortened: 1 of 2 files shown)


abs007:
Could I get some context into how and where from are commands like `show`, `list` etc being run in the unit tests?

cmd/context/create.go
```diff
@@ -133,13 +137,15 @@ func (c *ContextCommand) UseContext(ctx context.Context, ctxOptions *ContextOpti
...
 				ctxOptions.Context = transformedCtx
+				ctxOptions.Namespace = ctxStore.Contexts[ctxOptions.Context].Namespace
 				ctxOptions.IsOkteto = true
 			}
 		}
 	}
 
 	if okCtx, ok := ctxStore.Contexts[ctxOptions.Context]; !ok {
 		ctxStore.Contexts[ctxOptions.Context] = &okteto.OktetoContext{Name: ctxOptions.Context}
+		// What should be the namespace here?
```
jLopezbarb, line 148: I don't think any namespace should be placed here since this code is to create a new namespace, and we want to set the context if the context already exists

cmd/context/create.go
```diff
@@ -133,13 +137,15 @@ func (c *ContextCommand) UseContext(ctx context.Context, ctxOptions *ContextOpti
 			transformedCtx := okteto.K8sContextToOktetoUrl(ctx, ctxOptions.Context, ctxOptions.Namespace, c.K8sClientProvider)
 			if transformedCtx != ctxOptions.Context {
 				ctxOptions.Context = transformedCtx
+				ctxOptions.Namespace = ctxStore.Contexts[ctxOptions.Context].Namespace
 				ctxOptions.IsOkteto = true
 			}
 		}
 	}
 
 	if okCtx, ok := ctxStore.Contexts[ctxOptions.Context]; !ok {
```
jLopezbarb, line 146: I'd move all the repeated code to this if else statement. It will be easier to make a change in the future if we only need to check one place

cmd/context/create.go
```diff
@@ -133,13 +137,15 @@ func (c *ContextCommand) UseContext(ctx context.Context, ctxOptions *ContextOpti
...
 		ctxStore.Contexts[ctxOptions.Context] = &okteto.OktetoContext{Name: ctxOptions.Context}
+		// What should be the namespace here?
 		created = true
 	} else if ctxOptions.Token == "" {
 		// this is to avoid login with the browser again if we already have a valid token
```
jLopezbarb, line 151: I'd add it here checking if the namespace is not already set in `ctxOptions` (it can be set by the user in a flag and we should be using that one instead/if it's not done it will prevent running `okteto namespace`) and inside the code that you have above (`ctxOptions.Namespace = ctxStore.Contexts[ctxOptions.Context].Namespace`)

abs007: Should I place `ctxOptions.Namespace = ctxStore.Contexts[ctxOptions.Context].Namespace` outside the `else if`? Wouldn't `ctxOptions.Namespace` not be set if `ctxOptions.Token` isn't empty?

cmd/context/create.go
```diff
@@ -108,22 +108,26 @@ func (c *ContextCommand) UseContext(ctx context.Context, ctxOptions *ContextOpti
...
 	ctxStore := okteto.ContextStore()
 	if okCtx, ok := ctxStore.Contexts[ctxOptions.Context]; ok && okCtx.IsOkteto {
+		ctxOptions.Namespace = okCtx.Namespace
 		ctxOptions.IsOkteto = true
 	}
 
 	if okCtx, ok := ctxStore.Contexts[okteto.AddSchema(ctxOptions.Context)]; ok && okCtx.IsOkteto {
-		ctxOptions.Context = okteto.AddSchema(ctxOptions.Context)
+		ctxOptions.Context = okteto.AddSchema(ctxOptions.Context) // Should this be changed to okCtx.Context?
+		ctxOptions.Namespace = okCtx.Namespace
```
abs007, line 117: Maybe I should keep this line since `ctxOptions.Namespace = ctxStore.Contexts[ctxOptions.Context].Namespace` won't fit here.

jLopezbarb: We must remove this line since `ctxOptions.Namespace` is the one that we one to preserve if it's set. If it's not set it will be set by default to use `ctxStore.Contexts[ctxOptions.Context].Namespace` so it's not needed here

jLopezbarb (approved):

jLopezbarb:
I'll merge this when the new release candidate is out. I've tested it and everything seems ok but since it's a change that can affect many users, I want to test it for a whole release cycle.

abs007:
Sounds good 👍

PR #3198, opened 2022-10-26
Fix divert at scale

pchico83:

# Proposed changes

Divert is using a headless service to resolve a service `s1` in namespace `n1` to the endpoints of service `s1` in another namespace `n2`. We were using the `s1` in namespace `n2` pod endpoints to do this. When pods are recreated in `n2`, the okteto webhook is updating the headless service [here](https://github.com/okteto/app/blob/master/backend/cmd/webhook.go#L810).

In some environments, the webhook is timing out and failing to update all the headless services. 

In this PR, instead of using the pod ips in namespace `n2`, we are using the `s1` cluster ip, which doesn't change between upgrades. Using this approach, we will be able to delete the logic in our webhook to update the headless services

### DIFF
--- a/pkg/errors/errors.go
+++ b/pkg/errors/errors.go
@@ -186,6 +186,11 @@ var (
 	ErrX509Hint = "Add the flag '--insecure-skip-tls-verify' to skip certificate verification.\n    Follow this link to know more about configuring your own certificates with Okteto:\n    https://www.okteto.com/docs/self-hosted/administration/certificates/"
 )
 
+// IsAlreadyExists raised if the Kubernetes API returns AlreadyExists
+func IsAlreadyExists(err error) bool {
+	return err != nil && strings.Contains(err.Error(), "already exists")
+}
+
 // IsForbidden raised if the Okteto API returns 401
 func IsForbidden(err error) bool {
 	return err != nil && strings.Contains(err.Error(), "unauthorized")
--- a/pkg/k8s/diverts/divert.go
+++ b/pkg/k8s/diverts/divert.go
@@ -22,6 +22,7 @@ import (
 	"github.com/okteto/okteto/pkg/model"
 	apiv1 "k8s.io/api/core/v1"
 	networkingv1 "k8s.io/api/networking/v1"
+	k8sErrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 )
@@ -35,7 +36,7 @@ func DivertIngress(ctx context.Context, m *model.Manifest, fromIn *networkingv1.
 		}
 		in = translateIngress(m, fromIn)
 		if _, err := c.NetworkingV1().Ingresses(m.Namespace).Create(ctx, in, metav1.CreateOptions{}); err != nil {
-			if !oktetoErrors.IsAlreadyExists(err) {
+			if !k8sErrors.IsAlreadyExists(err) {
 				return err
 			}
 		}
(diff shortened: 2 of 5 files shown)


pkg/k8s/diverts/divert.go
```diff
@@ -32,13 +33,14 @@ func DivertIngress(ctx context.Context, m *model.Manifest, fromIn *networkingv1.
 		if !oktetoErrors.IsNotFound(err) {
 			return err
 		}
-		in = translateIngress(m, fromIn, "")
+		in = translateIngress(m, fromIn)
 		if _, err := c.NetworkingV1().Ingresses(m.Namespace).Create(ctx, in, metav1.CreateOptions{}); err != nil {
-			return err
+			if !oktetoErrors.IsAlreadyExists(err) {
```
pchico83, line 38: It could be a RC here is another `okteto deploy` command is diverting the same ingress in parallel. Since they both would create the same ingress yaml, it's safe to ignore this error

pkg/k8s/diverts/divert.go
```diff
@@ -68,46 +70,44 @@ func divertService(ctx context.Context, m *model.Manifest, name string, c kubern
...
-		return divertEndpoints(ctx, m, name, c)
+		return divertEndpoints(ctx, m, from, c)
 	}
 
 	if s.Annotations[model.OktetoAutoCreateAnnotation] != "true" {
 		return nil
 	}
 
-	s = translateService(m, from, s.ResourceVersion)
+	s = translateService(m, from)
```
pchico83, line 86: I removed the `resourceVersion` check everywhere. Otherwise,  `okteto deploy` commands running in parallel might fail. Since they create the same yaml, it's safe to ignore this error.
At some point, all this logic should be handled by an operator, but that requires a major refactor

pkg/k8s/diverts/translate.go
```diff
@@ -72,22 +70,48 @@ func translateService(m *model.Manifest, s *apiv1.Service, resourceVersion strin
...
-			ResourceVersion: resourceVersion,
+			Name: s.Name,
+			Labels: map[string]string{
+				model.DeployedByLabel: m.Name,
+			},
+			Annotations: map[string]string{
+				model.OktetoAutoCreateAnnotation: "true",
+			},
+		},
+		Subsets: []apiv1.EndpointSubset{
```
pchico83, line 84: This is the main change. We use the target service ip, instead of the target pod ips.

ifbyol (approved):
Minor comments but it looks good!

pkg/k8s/diverts/divert.go
jLopezbarb: I was going to propose the same thing. The function that we need to use is: https://github.com/kubernetes/apimachinery/blob/master/pkg/api/errors/errors.go#L159

pkg/k8s/diverts/translate.go
```diff
@@ -72,22 +70,48 @@ func translateService(m *model.Manifest, s *apiv1.Service, resourceVersion strin
...
+func translateEndpoints(m *model.Manifest, s *apiv1.Service) *apiv1.Endpoints {
 	result := &apiv1.Endpoints{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:            e.Name,
-			Labels:          e.Labels,
-			Annotations:     e.Annotations,
-			ResourceVersion: resourceVersion,
+			Name: s.Name,
+			Labels: map[string]string{
+				model.DeployedByLabel: m.Name,
```
jLopezbarb, line 78: The okteto manifest can have a different name from the stack if its combined with deploy commands
```suggestion
				model.DeployedByLabel: m.GetStack().Name,
```

pkg/errors/errors.go
```diff
@@ -186,6 +186,11 @@ var (
 	ErrX509Hint = "Add the flag '--insecure-skip-tls-verify' to skip certificate verification.\n    Follow this link to know more about configuring your own certificates with Okteto:\n    https://www.okteto.com/docs/self-hosted/administration/certificates/"
 )
 
+// IsAlreadyExists raised if the Kubernetes API returns AlreadyExists
+func IsAlreadyExists(err error) bool {
```
ifbyol, line 190: I think this is not needed anymore

ifbyol (approved):
Minor comment about an unused function, but the code looks good!

jLopezbarb (approved):

PR #3199, opened 2022-10-27
Code cleanup in pkg/okteto/context.go

abs007:

# Proposed changes

Fixes: Not a particular issue but performing code cleanup as discussed in the [slack channel](https://kubernetes.slack.com/archives/CM1QMQGS0/p1666671455165989).

cc: @rberrelleza

### DIFF
--- a/pkg/okteto/context.go
+++ b/pkg/okteto/context.go
@@ -221,15 +221,10 @@ func ContextStore() *OktetoContextStore {
 		}
 		CurrentStore = ctxStore
 
-		if _, ok := CurrentStore.Contexts[CurrentStore.CurrentContext]; !ok {
-			return CurrentStore
-		}
 		return CurrentStore
 	}
 
-	CurrentStore = &OktetoContextStore{
-		Contexts: map[string]*OktetoContext{},
-	}
+	CurrentStore = &OktetoContextStore{}
 	return CurrentStore
 }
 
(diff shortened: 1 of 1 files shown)


pkg/okteto/context.go
```diff
@@ -221,15 +221,10 @@ func ContextStore() *OktetoContextStore {
...
-		if _, ok := CurrentStore.Contexts[CurrentStore.CurrentContext]; !ok {
-			return CurrentStore
-		}
 		return CurrentStore
 	}
 
-	CurrentStore = &OktetoContextStore{
-		Contexts: map[string]*OktetoContext{},
-	}
+	CurrentStore = &OktetoContextStore{}
```
jLopezbarb, line 227: I think we need to initialise the context map, otherwise it will panic whenever we try to create a new entry to `CurrentStore`.
```suggestion
	CurrentStore = &OktetoContextStore{
	     Contexts: map[string]*OktetoContext{},
	}
```

abs007: Yep, I found that changing this broke tests. I have a question though. Doesn't omitting the `Contexts` map set it to be zero valued (default value) and I think that that's what's happening in line 228?

jLopezbarb (approved):

teresaromero (approved):

ifbyol (approved):

PR #3201, opened 2022-10-27
Make the channel setting configurable with an environment variable

pchico83:

# Proposed changes

Our current method to install the dev channel has friction. People have to create the channel file, and remember to undo this change when a new official release is available.

With this change, people can install the dev channel like this:

```
curl https://get.okteto.com -sSfL | OKTETO_CHANNEL=dev sh
```

and next install will be the last stable version again

### DIFF
--- a/scripts/get-okteto.sh
+++ b/scripts/get-okteto.sh
@@ -70,29 +70,28 @@
                 fi
         fi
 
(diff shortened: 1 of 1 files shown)


maroshii (approved):
nice 👌

PR #3202, opened 2022-10-28
Jlopezbarb/syncthing removing console instead of okteto process e2e

### LINKED ISSUE #3177: Running "okteto up" twice without proper termination causes logout in Ubuntu
**Describe the bug**
Running okteto up twice causes my session to be logged out. I have found a specific scenario to reproduce it but it has happened to me without the given scenario also.

**To Reproduce**

Run "okteto up" in a project
Terminate terminal 
Run "okteto up" in a same project again
When "Activating your development container" is displayed I'm logged out of my session

**Expected behavior**
 Should start project again

- OS: Ubuntu 
- Version: 22.04

**Additional context**
From auth.log I can see:
```
Oct 19 16:29:58 mihkel-ThinkPad-X1-Extreme-Gen-4i polkitd(authority=local): Unregistered Authentication Agent for unix-session:12 (system bus name :1.703, object path /org/freedesktop/PolicyKit1/AuthenticationAgent, locale en_US.UTF-8) (disconnected from bus)
Oct 19 16:29:59 mihkel-ThinkPad-X1-Extreme-Gen-4i gdm-password]: pam_unix(gdm-password:session): session closed for user mihkel
Oct 19 16:29:59 mihkel-ThinkPad-X1-Extreme-Gen-4i systemd-logind[1130]: Session 12 logged out. Waiting for processes to exit.
Oct 19 16:29:59 mihkel-ThinkPad-X1-Extreme-Gen-4i systemd-logind[1130]: Removed session 12.
```

Running okteto cli 2.8.1

jLopezbarb:
Fixes #3177

## Proposed changes

- It seems that ubuntu has another process PID to detect if the process is not an okteto process
- Getting the parent should finish whenever we reach the okteto command.
- We need to maintain previous code to make it work on Windows

### DIFF
--- a/cmd/up/forwards.go
+++ b/cmd/up/forwards.go
@@ -82,7 +82,6 @@ func (up *upContext) sshForwards(ctx context.Context) error {
 	}
 
 	up.Forwarder = ssh.NewForwardManager(ctx, fmt.Sprintf(":%d", up.Dev.RemotePort), up.Dev.Interface, "0.0.0.0", f, up.Dev.Namespace)
-
 	if err := up.Forwarder.Add(forward.Forward{Local: up.Sy.RemotePort, Remote: syncthing.ClusterPort}); err != nil {
 		return err
 	}
@@ -91,24 +90,8 @@ func (up *upContext) sshForwards(ctx context.Context) error {
 		return err
 	}
 
-	for idx, f := range up.Dev.Forward {
-		if f.Labels != nil {
-			forwardWithServiceName, err := up.Forwarder.TransformLabelsToServiceName(f)
-			if err != nil {
-				return err
-			}
-			up.Dev.Forward[idx] = forwardWithServiceName
-			f = forwardWithServiceName
-		}
-		if err := up.Forwarder.Add(f); err != nil {
-			return err
-		}
-	}
-
-	for _, r := range up.Dev.Reverse {
-		if err := up.Forwarder.AddReverse(r); err != nil {
-			return err
-		}
+	if err := addToForwarder(up); err != nil {
+		return err
 	}
 
 	if err := ssh.AddEntry(up.Dev.Name, up.Dev.Interface, up.Dev.RemotePort); err != nil {
@@ -127,6 +110,67 @@ func (up *upContext) sshForwards(ctx context.Context) error {
 	}
 
 	return nil
+}
+
+func addToForwarder(up *upContext) error {
+	ticker := time.NewTicker(1 * time.Second)
+	to := time.NewTicker(up.Dev.Timeout.Resources)
+	var forwardErr error
+	alreadyAdded := map[int]bool{}
+	for {
+		select {
+		case <-ticker.C:
+			forwardErr = nil
+
+			for idx, f := range up.Dev.Forward {
+				if _, ok := alreadyAdded[f.Local]; ok {
+					continue
+				}
+				if f.Labels != nil {
+					forwardWithServiceName, err := up.Forwarder.TransformLabelsToServiceName(f)
+					if err != nil {
+						oktetoLog.Infof("could not create forward port: %s", err)
+						forwardErr = err
+						continue
+					}
+					up.Dev.Forward[idx] = forwardWithServiceName
+					f = forwardWithServiceName
+					alreadyAdded[f.Local] = true
+				}
+				if err := up.Forwarder.Add(f); err != nil {
+					oktetoLog.Infof("could not create forward port: %s", err)
+					forwardErr = err
+					continue
+				}
+				alreadyAdded[f.Local] = true
+			}
+			if forwardErr != nil {
+				continue
+			}
+
+			for _, r := range up.Dev.Reverse {
+				if _, ok := alreadyAdded[r.Local]; ok {
+					continue
+				}
+				if err := up.Forwarder.AddReverse(r); err != nil {
+					oktetoLog.Infof("could not create reverse port: %s", err)
+					forwardErr = err
+					continue
+				}
+				alreadyAdded[r.Local] = true
+			}
+
+			if forwardErr != nil {
+				continue
+			}
+			return nil
+		case <-to.C:
+			if forwardErr != nil {
+				return forwardErr
+			}
+			return fmt.Errorf("could not create local ports after %s", up.Dev.Timeout.Resources.String())
+		}
+	}
 }
 
 func (up *upContext) setGlobalForwardsIfRequiredLoop(ctx context.Context) {
--- a/cmd/up/forwards.go
+++ b/cmd/up/forwards.go
@@ -114,7 +114,7 @@ func (up *upContext) sshForwards(ctx context.Context) error {
 
 func addToForwarder(up *upContext) error {
 	ticker := time.NewTicker(1 * time.Second)
-	to := time.NewTicker(up.Dev.Timeout.Resources)
+	to := time.NewTicker(10 * time.Second)
 	var forwardErr error
 	alreadyAdded := map[int]bool{}
 	for {
--- a/cmd/up/pid.go
+++ b/cmd/up/pid.go
@@ -21,7 +21,81 @@ import (
 
 	"github.com/okteto/okteto/pkg/config"
 	oktetoLog "github.com/okteto/okteto/pkg/log"
+	"github.com/spf13/afero"
 )
+
+const oktetoPIDFilename = "okteto.pid"
+
+// PIDController creates get and removes the info about the OktetoPID
+type pidController struct {
+	pidFilePath string
+	filesystem  afero.Fs
+	pidProvider pidProvider
+}
+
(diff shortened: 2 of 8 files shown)


AgustinRamiroDiaz:
Why are we using names instead of ids? is it to not store the reference?

pkg/syncthing/syncthing.go
```diff
@@ -878,6 +878,10 @@ func getParent(p *process.Process) (*process.Process, error) {
 		oktetoLog.Infof("parent name is the same: %s", name)
 		return getParent(parent)
 	}
+	if !strings.Contains(pName, "okteto") {
+		oktetoLog.Infof("okteto parent not found: %s", name)
+		return getParent(parent)
```
AgustinRamiroDiaz, line 883: Maybe the name should change to `getOktetoAncestor`

jLopezbarb: I'm changing the whole PR to make it safer deleting process

pkg/syncthing/syncthing.go
```diff
@@ -878,6 +878,10 @@ func getParent(p *process.Process) (*process.Process, error) {
 		oktetoLog.Infof("parent name is the same: %s", name)
 		return getParent(parent)
 	}
+	if !strings.Contains(pName, "okteto") {
```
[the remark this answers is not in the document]
jLopezbarb: We are getting parents until we get something different than `syncthing` which was supposed to be the `okteto` process. It seems that there are some OS that instead of syncthing being depending on okteto it depended on the okteto parent (which was the terminal) and was terminating the terminal process instead of okteto

pkg/syncthing/syncthing.go
```diff
@@ -833,54 +832,17 @@ func (s *Syncthing) HardTerminate() error {
...
 			oktetoLog.Infof("error terminating syncthing %d with wait: %s", p.Pid, err.Error())
 		}
-		if parent != nil && parent.Pid != int32(pid) {
-			oktetoLog.Infof("pid: %s / name: %s / cmd: %s", p.Pid, p.Name, p.Cmdline)
-			if err := terminate(parent, true); err != nil {
-				oktetoLog.Infof("error terminating syncthing %d with wait: %s", p.Pid, err.Error())
-				continue
-			}
-		}
+
```
jLopezbarb, line 839: This code is moved to wait until the okteto notices that there is another process running

cmd/up/forwards.go
```diff
@@ -91,24 +90,8 @@ func (up *upContext) sshForwards(ctx context.Context) error {
...
-		if err := up.Forwarder.Add(f); err != nil {
-			return err
-		}
-	}
-
-	for _, r := range up.Dev.Reverse {
-		if err := up.Forwarder.AddReverse(r); err != nil {
-			return err
-		}
+	if err := addToForwarder(up); err != nil {
```
jLopezbarb, line 93: We need to wait until the other process free the ports

cmd/up/forwards.go
```diff
@@ -129,6 +112,67 @@ func (up *upContext) sshForwards(ctx context.Context) error {
 	return nil
 }
 
+func addToForwarder(up *upContext) error {
+	ticker := time.NewTicker(1 * time.Second)
+	to := time.NewTicker(up.Dev.Timeout.Resources)
+	var forwardErr error
+	alreadyAdded := map[int]bool{}
```
jLopezbarb, line 119: needed to store if it's already added to not add them twice (that's another error)

cmd/up/forwards.go
```diff
@@ -129,6 +112,67 @@ func (up *upContext) sshForwards(ctx context.Context) error {
 	return nil
 }
 
+func addToForwarder(up *upContext) error {
+	ticker := time.NewTicker(1 * time.Second)
+	to := time.NewTicker(up.Dev.Timeout.Resources)
+	var forwardErr error
```
jLopezbarb, line 118: need to store the error if the timeout finish we need to return a proper error to the user

go.mod
```diff
@@ -218,6 +218,8 @@ require (
 	sigs.k8s.io/yaml v1.2.0 // indirect
 )
 
+require github.com/spf13/afero v1.9.2
```
jLopezbarb, line 221: Added to mock filesystem calls

jLopezbarb:
@pchico83 The only problem that I've found while testing this is that killing the okteto process is faster (previous method, not working for `ubuntu`) than letting okteto know that it has been replaced by another one. I've done some tests and the time to destroy the old dev container is around 30s vs almost instantaneous (since we were removing the process). We need to wait until the other process is finished in order to create the forwarder, it will fail with port is already taken error if we don't wait.

On the other side, even if it takes more time to start the new okteto, the error displayed on the older okteto process is more verbose:

### _**Old behavior**_
Captura de Pantalla 2022-11-03 a las 13 19 23

### _**New behavior**_

Captura de Pantalla 2022-11-03 a las 15 25 51

pchico83:
> @pchico83 The only problem that I've found while testing this is that killing the okteto process is faster (previous method, not working for `ubuntu`) than letting okteto know that it has been replaced by another one. I've done some tests and the time to destroy the old dev container is around 30s vs almost instantaneous (since we were removing the process). We need to wait until the other process is finished in order to create the forwarder, it will fail with port is already taken error if we don't wait.
> 
> On the other side, even if it takes more time to start the new okteto, the error displayed on the older okteto process is more verbose:
> 
> ### _**Old behavior**_
> Captura de Pantalla 2022-11-03 a las 13 19 23
> 
> ### _**New behavior**_
> Captura de Pantalla 2022-11-03 a las 15 25 51

Let's find a faster notification channel, we could run a goroutine watching for changes in `okteto.pid`

cmd/up/forwards.go
```diff
@@ -129,6 +112,67 @@ func (up *upContext) sshForwards(ctx context.Context) error {
 	return nil
 }
 
+func addToForwarder(up *upContext) error {
+	ticker := time.NewTicker(1 * time.Second)
+	to := time.NewTicker(up.Dev.Timeout.Resources)
```
pchico83, line 117: Let's use a 10s timeout. Otherwise, if the port is taken by other app, we would fail very late

cmd/up/up.go
```diff
@@ -570,6 +573,14 @@ func (up *upContext) activateLoop() {
 		if up.isRetry || isTransientError {
 			oktetoLog.Infof("waiting for shutdown sequence to finish")
 			<-up.ShutdownCompleted
+			pidFromFile, err := up.pidController.get()
+			if err != nil {
+				oktetoLog.Infof("error getting pid: %w")
+			}
+			if pidFromFile != strconv.Itoa(os.Getpid()) {
+				up.Exit <- fmt.Errorf("development container has been replaced")
```
pchico83, line 581: Can we add a hint to run `okteto exec`? Also, I would say something like:
```
development container has been deactivated by another "okteto up" command
```

pchico83 (approved):

jLopezbarb:
Adding a comment to show the new changes
## New changes:
- Added fsnotify as dependency to watch for file change. (Thanks @pchico83, this code is much faster than the one we had before. Great win-win feature, we will fix a very annoying bug and make UX better at the same time)
- Make the code testable
- This is how it will look like the error
Captura de Pantalla 2022-11-04 a las 14 51 44

cmd/up/pid.go
```diff
@@ -19,32 +19,210 @@ import (
...
+				pid := pc.pidProvider.provide()
+				filePID, err := pc.get()
+				if err != nil {
+					oktetoLog.Infof("unable to get PID file at %s", pc.pidFilePath)
+					continue
+				}
+				if strconv.Itoa(pid) != filePID {
+					notifyCh <- oktetoErrors.UserError{
+						E:    fmt.Errorf("development container has been deactivated by another 'okteto up' command"),
+						Hint: "If you want to run a second terminal you can use 'okteto exec' command",
```
pchico83, line 183:
```suggestion
						Hint: "Use ' okteto exec' to open another terminal to your development container",
```

pchico83 (approved):

teresaromero (approved):

ifbyol (approved):

pchico83:
@jLopezbarb are we backporting this one? or are we waiting for the next cli release?

jLopezbarb:
Do we want to create a new patch with this change? I'm ok back porting it and creating a new release

pchico83:
> Do we want to create a new patch with this change? I'm ok back porting it and creating a new release

@jLopezbarb whatever you think it's better. I would say this isn't a trivial change and it should be in `master` for a while, but you are the best one to evaluate that

jLopezbarb:
I prefer to have it in master for a while (we'll cut the next release at the end of next week), so I think it will be better to have it in master for a week and test it with all the e2e tests whenever we merge into master and with manual tests whenever we start testing the release candidate

PR #3206, opened 2022-11-02
Remove "github.com/ibuildthecloud/finalizers" as dependency

### LINKED ISSUE #3091: remove `github.com/ibuildthecloud/finalizers` as dependency
**Describe the bug**
We should remove `github.com/ibuildthecloud/finalizers` as a dependency because:
- Code has not been updated in the last 2 years
- Untestable code. We can't test the code using this dependency because it doesn't allow adding fake k8s clients.
- Only using one function. We are only using one function that could be rewritten on our own since it doesn't add a lot of complexity.

**Expected behavior**
Our code should be testable and we should not inject dependencies for a small function

**Additional context**
Function used on the destroy command: https://github.com/okteto/okteto/blob/master/pkg/k8s/namespaces/namespace.go#L78

abs007:

# Proposed changes

Fixes #3091

Removed "github.com/ibuildthecloud/finalizers" as a dependency. This will allow for testing of the code.


teresaromero (approved):

pkg/k8s/namespaces/namespace.go
```diff
@@ -199,3 +229,111 @@ func (n *Namespaces) DestroySFSVolumes(ctx context.Context, ns string, opts Dele
 
 	return nil
 }
+
+func NewTrip(restConfig *rest.Config, opts *Options) (*Trip, error) {
```
ifbyol, line 233: Minor thing, Can we include a comment in this function and `Wander` mentioning that were implemented to remove the dependency to the deleted module?

ifbyol (approved):

PR #3214, opened 2022-11-10
Sanitize name for deploy and destroy name

teresaromero:
# Proposed changes

Resolves https://github.com/okteto/app/issues/4980 

When deploying or destroying a repository and the `name` is not compatible, k8s fails to create configmap to perform the action.

### DIFF
--- a/pkg/cmd/pipeline/translate.go
+++ b/pkg/cmd/pipeline/translate.go
@@ -82,11 +82,13 @@ type CfgData struct {
 	Filename   string
 	Manifest   []byte
 	Icon       string
+	// ResourceName represents the pipeline name sanitized
+	ResourceName string
 }
 
 // TranslateConfigMapAndDeploy translates the app into a configMap
 func TranslateConfigMapAndDeploy(ctx context.Context, data *CfgData, c kubernetes.Interface) (*apiv1.ConfigMap, error) {
-	cmap, err := configmaps.Get(ctx, TranslatePipelineName(data.Name), data.Namespace, c)
+	cmap, err := configmaps.Get(ctx, TranslatePipelineName(data.ResourceName), data.Namespace, c)
 	if err != nil {
 		if !oktetoErrors.IsNotFound(err) {
 			return nil, err
@@ -131,7 +133,7 @@ func UpdateConfigMap(ctx context.Context, cmap *apiv1.ConfigMap, data *CfgData, 
 	return configmaps.Deploy(ctx, cmap, cmap.Namespace, c)
 }
 
-// TranslatePipelineName translate the name into the pipeline name
+// TranslatePipelineName translate the name into the configmap name
 func TranslatePipelineName(name string) string {
 	return fmt.Sprintf("okteto-git-%s", name)
 }
@@ -173,7 +175,7 @@ func translateConfigMapSandBox(data *CfgData) *apiv1.ConfigMap {
 	cmap := &apiv1.ConfigMap{
 		ObjectMeta: metav1.ObjectMeta{
 			Namespace: data.Namespace,
-			Name:      TranslatePipelineName(data.Name),
+			Name:      TranslatePipelineName(data.ResourceName),
 			Labels: map[string]string{
 				model.GitDeployLabel: "true",
 			},
--- a/pkg/cmd/pipeline/translate.go
+++ b/pkg/cmd/pipeline/translate.go
@@ -25,6 +25,7 @@ import (
 	"strings"
 
 	oktetoErrors "github.com/okteto/okteto/pkg/errors"
+	oktetoFormat "github.com/okteto/okteto/pkg/format"
 	"github.com/okteto/okteto/pkg/k8s/apps"
 	"github.com/okteto/okteto/pkg/k8s/configmaps"
 	oktetoLog "github.com/okteto/okteto/pkg/log"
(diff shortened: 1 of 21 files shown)


pkg/model/utils.go
```diff
@@ -202,3 +203,24 @@ func getListDiff(l1, l2 []string) []string {
...
+
+	// if multiple "-" replace all for only one
+	re, err := regexp.Compile(`-(-+)`)
+	if err != nil {
+		return "", err
+	}
+	name = re.ReplaceAllString(name, "-")
+	// trim the repository name for internal use in labels
+	if len(name) > 63 {
+		return name[:63], nil
```
pchico83, line 223: this might leave and ending `_` again

teresaromero: `_` are replaced by `-` with the kube validation, but it might get a `-` indeed, leet me review this

pkg/sanitization/sanitization.go
```diff
@@ -0,0 +1,30 @@
...
+
+var (
+	// validKubeNameRegex is the regex to validate a kubernetes resource name
+	validKubeNameRegex = regexp.MustCompile(`[^a-z0-9\-]+`)
+
+	// moreThanOneHyphen is the regex to check if there is more than one hyphen together
+	moreThanOneHyphen = regexp.MustCompile(`-(-+)`)
+)
+
+func SanitizeName(name string) string {
```
ifbyol, line 16: Can you include a comment here explaining what are we sanitizing?

pkg/sanitization/sanitization.go
```diff
@@ -0,0 +1,30 @@
...
+	name = strings.TrimSpace(name)
+	name = strings.ToLower(name)
+
+	name = validKubeNameRegex.ReplaceAllString(name, "-")
+
+	name = moreThanOneHyphen.ReplaceAllString(name, "-")
+	// trim the repository name for internal use in labels
+	if len(name) > 63 {
```
ifbyol, line 24: Can we extract 63 to a constant with meaning instead of a magic number? For example, `const maxKubernetesResourceLenght = 63` or similar? I think it is important to indicate the meaning of 63

pkg/cmd/pipeline/translate.go
```diff
@@ -82,11 +82,13 @@ type CfgData struct {
 	Filename   string
 	Manifest   []byte
 	Icon       string
+	// ResourceName represents the pipeline name sanitized
+	ResourceName string
```
ifbyol, line 86: Is there a way we don't include this variable in this struct? This struct is to keep all the information that will be stored *within* the configmap but this won't. 

Maybe we can omit this in the struct and let the handler to call the sanitization. At the end, this restriction is because we use Kubernetes as "storage" for the git deploy information. Sanitization call isn't too expensive, so I don't think there is a problem on calling it several times. The only drawback of this approach is to be aware of having to call it to build the config map name.

I would try to restrict this struct to the fields stored within the config map

teresaromero: i've reverted the change and now the function to format the string is called whenever is needed

pkg/format/format.go
```diff
@@ -0,0 +1,36 @@
+package format
```
jLopezbarb, line 1: NIT: If we are going to add a package `format` I'd rename the files to know what they are formatting, for example this one could be `k8s.go`. 

I don't know if there will be lots of formatting functions, Should we create a package then? I'm thinking out loud to get some more opinions

ifbyol: > NIT: If we are going to add a package `format` I'd rename the files to know what they are formatting, for example this one could be `k8s.go`.
>
> I don't know if there will be lots of formatting functions, Should we create a package then? I'm thinking out loud to get some more opinions

Maybe changing the file to `k8s.go` could be done, but maybe this transformations can apply to other things. As you wish, but maybe we can leave it as it is now and see how it evolves

ifbyol (commented):
We need to check at least all these functions and where they are being used: https://github.com/okteto/okteto/blob/master/pkg/cmd/pipeline/crud.go#L30-L59

All of them receive a name and check the config map or things using `dev.okteto.com/deployed-by` labels

teresaromero:
> We need to check at least all these functions and where they are being used: https://github.com/okteto/okteto/blob/master/pkg/cmd/pipeline/crud.go#L30-L59
> 
> All of them receive a name and check the config map or things using `dev.okteto.com/deployed-by` labels

I've been revising this. I've made some changes in order to get the formatted name for the labels.

ifbyol:
@teresaromero can we run the e2e before merging this?

ifbyol (commented):
@teresaromero I think we need to apply the change here: https://github.com/okteto/okteto/blob/a4897660a3e57dfb9ee16a0d9a51c3d0fc6ac44b/cmd/deploy/endpoints.go#L141.

We would need to check all the places where the `deployed-by` is used and evaluate if we need to apply the changes there or not. Probably we will need it

teresaromero:
> @teresaromero I think we need to apply the change here:
> 
> https://github.com/okteto/okteto/blob/a4897660a3e57dfb9ee16a0d9a51c3d0fc6ac44b/cmd/deploy/endpoints.go#L141
> 
> .
> We would need to check all the places where the `deployed-by` is used and evaluate if we need to apply the changes there or not. Probably we will need it

I've uploaded a commit with a pass on deployed by labels and formatting.. passing again e2e

maroshii (approved):

jLopezbarb (approved):

Issue #3211, opened 2022-11-09
Possible bug in up.go


abs007:
**Describe the bug**
Should this line in `up.go`:
https://github.com/okteto/okteto/blob/39b5b7757668a2b4bea7b3e0054c3b174f010dd9/cmd/up/up.go#L374
be changed to:
```go
 } else if len(args) == maxV1Args { 
```
**Reason**
What if maxV1Args increases beyond 1? Also, seems like cleaner code to set it this way.

jLopezbarb:
Nice catch @abs007!!

abs007:
Happy to help!

jLopezbarb:
I uploaded a PR but after doing it, I reviewed [the PR that you made](https://github.com/okteto/okteto/pull/3215) and since it will be replaced I think the PR that I made is not needed anymore. I think we can tackle this issue in your PR, WDYT @abs007 ?

abs007:
Seems good. Also, I think that the pr I made would also be taking care of this slack discussion: https://kubernetes.slack.com/archives/CM1QMQGS0/p1668059807909449

abs007:
Closing this issue wrt pr #3215

PR #3215, opened 2022-11-10
feat: add support for custom commands in `okteto up`

### LINKED ISSUE #2300: Support for custom commands on `okteto up`
I want to run `okteto up` with a custom command, for example `okteto up -- bash`to debug my service

abs007:

# Proposed changes

Fixes #2300 

This allows for the option to run a command instead of the already defined command in the dev/command section of the okteto manifest file

Use case: my command is `yarn start`, but today I just want to run tests in my dev container. I can run `okteto up -- yarn test`
The extra `<command>` field will replace the `command` in the dev section:
https://www.okteto.com/docs/reference/manifest/#command-string-optional

### DIFF
--- a/cmd/up/up.go
+++ b/cmd/up/up.go
@@ -274,7 +274,7 @@ func Up() *cobra.Command {
 					return err
 				}
 			}
-			if upOptions.commandToExecute != nil {
+			if len(upOptions.commandToExecute) > 0 {
 				dev.Command.Values = upOptions.commandToExecute
 			}
 
--- a/cmd/up/up.go
+++ b/cmd/up/up.go
@@ -84,13 +84,12 @@ func Up() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "up [svc] <command>",
 		Short: "Launch your development environment",
-		//Args:  utils.MaximumNArgsAccepted(2, "https://okteto.com/docs/reference/cli/#up"),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			if okteto.InDevContainer() {
 				return oktetoErrors.ErrNotInDevContainer
 			}
 
-			if err := upOptions.AddArgs(cmd, args); err != nil {
+			if err := upOptions.AddArgs(args); err != nil {
 				return err
 			}
 
@@ -275,6 +274,9 @@ func Up() *cobra.Command {
 					return err
 				}
 			}
+			if upOptions.commandToExecute != nil {
+				dev.Command.Values = upOptions.commandToExecute
+			}
 
 			up.Dev = dev
 			if forceAutocreate {
@@ -362,7 +364,7 @@ func Up() *cobra.Command {
 }
 
 // AddArgs sets the args as options and return err if it's not compatible
-func (o *UpOptions) AddArgs(cmd *cobra.Command, args []string) error {
+func (o *UpOptions) AddArgs(args []string) error {
 
 	if len(args) == 1 {
 		o.DevName = args[0]
@@ -370,18 +372,6 @@ func (o *UpOptions) AddArgs(cmd *cobra.Command, args []string) error {
 		o.DevName = args[0]
 		o.commandToExecute = args[1:]
 	}
-
-	// maxV1Args := 2
-	// docsURL := "https://okteto.com/docs/reference/cli/#up"
-	// if len(args) > maxV1Args {
-	// 	cmd.Help()
-	// 	return oktetoErrors.UserError{
-	// 		E:    fmt.Errorf("%q accepts at most %d arg(s), but received %d", cmd.CommandPath(), maxV1Args, len(args)),
-	// 		Hint: fmt.Sprintf("Visit %s for more information.", docsURL),
-	// 	}
-	// } else if len(args) == maxV1Args {
-	// 	o.DevName = args[0]
-	// }
 
 	return nil
 }
--- a/cmd/up/up.go
+++ b/cmd/up/up.go
@@ -65,25 +65,26 @@ type UpOptions struct {
 	ManifestPathFlag string
 	// ManifestPath is the patah to the manifest used though the command execution.
 	// This might change its value during execution
-	ManifestPath string
-	Namespace    string
-	K8sContext   string
-	DevName      string
-	Devs         []string
-	Envs         []string
-	Remote       int
-	Deploy       bool
-	ForcePull    bool
-	Reset        bool
+	ManifestPath     string
+	Namespace        string
+	K8sContext       string
+	DevName          string
+	Devs             []string
+	Envs             []string
+	Remote           int
+	Deploy           bool
+	ForcePull        bool
+	Reset            bool
+	commandToExecute []string
 }
 
 // Up starts a development container
 func Up() *cobra.Command {
 	upOptions := &UpOptions{}
 	cmd := &cobra.Command{
-		Use:   "up [svc]",
+		Use:   "up [svc] <command>",
 		Short: "Launch your development environment",
-		Args:  utils.MaximumNArgsAccepted(1, "https://okteto.com/docs/reference/cli/#up"),
+		//Args:  utils.MaximumNArgsAccepted(2, "https://okteto.com/docs/reference/cli/#up"),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			if okteto.InDevContainer() {
 				return oktetoErrors.ErrNotInDevContainer
@@ -363,17 +364,24 @@ func Up() *cobra.Command {
 // AddArgs sets the args as options and return err if it's not compatible
 func (o *UpOptions) AddArgs(cmd *cobra.Command, args []string) error {
 
-	maxV1Args := 1
-	docsURL := "https://okteto.com/docs/reference/cli/#up"
-	if len(args) > maxV1Args {
-		cmd.Help()
-		return oktetoErrors.UserError{
-			E:    fmt.Errorf("%q accepts at most %d arg(s), but received %d", cmd.CommandPath(), maxV1Args, len(args)),
-			Hint: fmt.Sprintf("Visit %s for more information.", docsURL),
-		}
-	} else if len(args) == 1 {
+	if len(args) == 1 {
 		o.DevName = args[0]
+	} else if len(args) > 1 {
+		o.DevName = args[0]
+		o.commandToExecute = args[1:]
 	}
+
+	// maxV1Args := 2
+	// docsURL := "https://okteto.com/docs/reference/cli/#up"
+	// if len(args) > maxV1Args {
+	// 	cmd.Help()
+	// 	return oktetoErrors.UserError{
+	// 		E:    fmt.Errorf("%q accepts at most %d arg(s), but received %d", cmd.CommandPath(), maxV1Args, len(args)),
+	// 		Hint: fmt.Sprintf("Visit %s for more information.", docsURL),
+	// 	}
+	// } else if len(args) == maxV1Args {
+	// 	o.DevName = args[0]
+	// }
 
 	return nil
 }
--- a/cmd/up/up.go
+++ b/cmd/up/up.go
@@ -69,7 +69,6 @@ type UpOptions struct {
 	Namespace        string
 	K8sContext       string
 	DevName          string
-	Devs             []string
 	Envs             []string
 	Remote           int
 	Deploy           bool
--- a/cmd/up/up.go
+++ b/cmd/up/up.go
@@ -63,7 +63,7 @@ type UpOptions struct {
 	// ManifestPathFlag is the option -f as introduced by the user when executing this command.
 	// This is stored at the configmap as filename to redeploy from the ui.
 	ManifestPathFlag string
-	// ManifestPath is the patah to the manifest used though the command execution.
+	// ManifestPath is the path to the manifest used though the command execution.
 	// This might change its value during execution
 	ManifestPath     string
 	Namespace        string
@@ -381,7 +381,7 @@ func (o *UpOptions) AddArgs(args []string, oktetoManifest *model.Manifest) error
 			o.commandToExecute = args
 		}
 	}
-
+	oktetoLog.Infof("dev name: %s / commandToExecute: %v", o.DevName, o.commandToExecute)
 	return nil
 }
 
--- a/cmd/up/up.go
+++ b/cmd/up/up.go
@@ -81,11 +81,16 @@ type UpOptions struct {
 func Up() *cobra.Command {
 	upOptions := &UpOptions{}
 	cmd := &cobra.Command{
-		Use:   "up [svc] <command>",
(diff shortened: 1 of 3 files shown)


abs007:
release/new-feature

cmd/up/up.go
```diff
@@ -361,18 +364,13 @@ func Up() *cobra.Command {
...
-	maxV1Args := 1
-	docsURL := "https://okteto.com/docs/reference/cli/#up"
-	if len(args) > maxV1Args {
-		cmd.Help()
-		return oktetoErrors.UserError{
-			E:    fmt.Errorf("%q accepts at most %d arg(s), but received %d", cmd.CommandPath(), maxV1Args, len(args)),
-			Hint: fmt.Sprintf("Visit %s for more information.", docsURL),
-		}
-	} else if len(args) == 1 {
+	if len(args) == 1 {
```
jLopezbarb, line 369: We need to distinct if the first arg is a command or a service. 

Maybe we need to move this function after we have loaded the manifest and check the first argument vs the services defined under the dev section

abs007: So are we taking into consideration that the command passed in `okteto up <command>` would be applicable to all dev containers in comparison to `okteto up [svc] <command>` where the command will be run in the particular dev container?

jLopezbarb: okteto up can only be run on one service.

We need to distinct the following use cases:
- `okteto up`: We will show the selector if more than one dev is in the manifest and run the command defined there
- `okteto up arg1`: There are two options here, the first one the service is defined in the manifest and the second one where the svc is not defined
  -  If it's defined we need to use that dev environment and don't show the prompt. The command would be the one defined in the manifest. For example (`okteto up api`)
  - If it's not defined we'll show the selector if more than one dev is available and use the arg as the command for our dev environment. For example (`okteto up sh`)
- `okteto up arg1 arg2`. As the previous one we need to check if the first arg is a dev environment or not.
  -  If it's defined we need to use that dev environment and don't show the prompt. The second arg will be used as the command of the dev environment. For example (`okteto up api sh`)
  - If it's not defined we'll show the selector if more than one dev is available and use the args as the command for our dev environment. For example (`okteto up sh echo hola`)

abs007: So, we'll always check to see if the first arg is a service or not. 

* If yes, we use that dev environment and then run commands (if extra args exist)
* If not, we show the prompt and pass all the args as commands to be run

If there aren't any args at all then keep doing what was done before.

cmd/up/up.go
```diff
@@ -364,13 +365,21 @@ func Up() *cobra.Command {
...
 	} else if len(args) > 1 {
-		o.DevName = args[0]
-		o.commandToExecute = args[1:]
+		if _, ok := oktetoManifest.Dev[args[0]]; ok {
+			o.DevName = args[0]
+			o.commandToExecute = args[1:]
+		} else {
+			o.commandToExecute = args
+		}
 	}
```
jLopezbarb, line 384: Can we add a log here to know what's the content of both values? This will help us to debug whenever there is a problem
```suggestion
oktetoLog.Infof("dev name: %s / commandToExecute: %v", o.DevName, o.commandToExecute)
```

cmd/up/up.go
```diff
@@ -364,13 +365,21 @@ func Up() *cobra.Command {
 }
 
 // AddArgs sets the args as options and return err if it's not compatible
-func (o *UpOptions) AddArgs(args []string) error {
+func (o *UpOptions) AddArgs(args []string, oktetoManifest *model.Manifest) error {
```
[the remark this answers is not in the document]
abs007: I was wondering what [these](https://github.com/okteto/okteto/commit/961b727ee5fae30962038cee24ef5b51b78f456f) cover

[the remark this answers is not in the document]
AdrianPedriza: It was probably my mistake in describing the commit. My test was related to covering the feature that allows us to add envs for `okteto up` command. The good one is [this one](https://github.com/okteto/okteto/pull/2578/commits/1d30f0e953337a11237b4ee755db08ce58b7d9e8) so go ahead with the test! 😄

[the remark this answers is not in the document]
abs007: Got it 👍

teresaromero:
> 



> Hi, could someone explain what the `Devs` field in UpOptions is used for? https://github.com/okteto/okteto/blob/master/cmd/up/up.go#L73

Hi @abs007! This is used when running `okteto up` from docker extension or detach mode. This feature was removed at 2.3 version and this should have been removed too as UpOption. 

@jLopezbarb you would have some more context, should we remove `Devs` options since is not being used in current versions of okteto ?

abs007:
Got it. I too saw that this wasn't being used anywhere and hence thought of asking. I can remove it if needed.

jLopezbarb:
> > Hi, could someone explain what the `Devs` field in UpOptions is used for? https://github.com/okteto/okteto/blob/master/cmd/up/up.go#L73
> 
> Hi @abs007! This is used when running `okteto up` from docker extension or detach mode. This feature was removed at 2.3 version and this should have been removed too as UpOption.
> 
> @jLopezbarb you would have some more context, should we remove `Devs` options since is not being used in current versions of okteto ?

Yes @teresaromero, we don't use it anymore so it's safe to remove. Thanks @abs007 for spotting it!! 🚀

cmd/up/up_test.go
```diff
@@ -307,3 +308,47 @@ func TestEnvVarIsNotAddedWhenHasBuiltInOktetoEnvVarsFormat(t *testing.T) {
...
+			upOptions:       &UpOptions{},
+			args:            []string{"echo", "hello"},
+			Manifest:        &model.Manifest{Dev: map[string]*model.Dev{"frontend": {}}},
+			expectedCommand: []string{"echo", "hello"},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			err := tt.upOptions.AddArgs(tt.args, tt.Manifest)
+			if err != nil {
```
jLopezbarb, line 345: @abs007 We usually use [testify](https://github.com/stretchr/testify) to test, since it's more readable
```suggestion
			assert.NoError(t, err)
```

cmd/up/up_test.go
```diff
@@ -307,3 +308,47 @@ func TestEnvVarIsNotAddedWhenHasBuiltInOktetoEnvVarsFormat(t *testing.T) {
...
+				t.Fatalf("unexpected error in  AddArgs: %s", err)
+			}
+
+			if reflect.DeepEqual(tt.expectedCommand, tt.upOptions.commandToExecute) == false {
```
jLopezbarb, line 349:
```suggestion
			assert.Equal(t, tt.expectedCommand,tt.upOptions.commandToExecute)
```

cmd/up/up_test.go
```diff
@@ -307,3 +308,47 @@ func TestEnvVarIsNotAddedWhenHasBuiltInOktetoEnvVarsFormat(t *testing.T) {
...
+				t.Fatalf("error in AddArgs; expected command %v but got %v", tt.expectedCommand, tt.upOptions.commandToExecute)
+			}
+		})
```
jLopezbarb, line 352: Can we also check that the `expectedDev` is the one we want

cmd/up/up.go
AgustinRamiroDiaz: @abs007 that sounds good

AgustinRamiroDiaz: @pchico83 @jLopezbarb wouldn't it be better to add the command as an option like `--command bash`?

I've seen that @abs007 has done a great work handling the distinction between a service and a command, but I feel like it's very error prone since we do a lot of modifications to the `manifest` struct

cmd/up/up.go
```diff
@@ -63,36 +63,31 @@ type UpOptions struct {
...
+	ManifestPath     string
+	Namespace        string
+	K8sContext       string
+	DevName          string
+	Envs             []string
+	Remote           int
+	Deploy           bool
+	ForcePull        bool
+	Reset            bool
+	commandToExecute []string
```
AgustinRamiroDiaz, line 77: where are we consuming this? I mean, where is this command being executed?

cmd/up/up.go
jLopezbarb: > @pchico83 @jLopezbarb wouldn't it be better to add the command as an option like `--command bash`?

We need the command to be an array of strings. If we add it as a flag the user will have to add several `--command`. That's why all the other CLIs like `kubectl` or `docker` use it as the final argument like `docker exec -ti my_container -- "echo a && echo b"`

cmd/up/up.go
jLopezbarb: Those are the flags that the user can use when using the `okteto up` command

cmd/up/up_test.go
abs007: Should I instead go with `if assert.Error(t, err) ` over here?

cmd/up/up.go
abs007: Not really being executed but it is being passed into `dev.Command.Values` to be used later in [here](https://github.com/okteto/okteto/pull/3215/files#diff-c1b929a650f782cf212e9b4b0ced5a7a298598bcb7473be505e7af1f06d6cf30R278). Passing it just before `up.Dev = dev` makes sure that the commands (if any) in the [okteto manifest](https://www.okteto.com/docs/reference/manifest/#dev-object-optional) would be replaced by the commands entered using the cli. WDYT?

cmd/up/up.go
```diff
@@ -361,20 +364,23 @@ func Up() *cobra.Command {
...
 
-	maxV1Args := 1
-	docsURL := "https://okteto.com/docs/reference/cli/#up"
-	if len(args) > maxV1Args {
-		cmd.Help()
-		return oktetoErrors.UserError{
-			E:    fmt.Errorf("%q accepts at most %d arg(s), but received %d", cmd.CommandPath(), maxV1Args, len(args)),
-			Hint: fmt.Sprintf("Visit %s for more information.", docsURL),
+	if len(args) == 1 {
+		if _, ok := oktetoManifest.Dev[args[0]]; ok {
```
abs007, line 370: @jLopezbarb Let's say we pass in `okteto up ["echo", "hello"]`. Would there be any sort of weird error or would the `ok` just be `false`

cmd/up/up.go
AgustinRamiroDiaz: Perfect! We should check if the dev command gets modified later, but sounds good :+1:

cmd/up/up.go
```diff
@@ -274,6 +274,9 @@ func Up() *cobra.Command {
 					return err
 				}
 			}
+			if upOptions.commandToExecute != nil {
```
AgustinRamiroDiaz, line 277: We should handle empty slices too
```suggestion
			if len(upOptions.commandToExecute) > 0 {
```

cmd/up/up.go
AgustinRamiroDiaz: There shouldn't be any errors here. Also, you can do the full test by building the binary with `make build`

cmd/up/up.go
AgustinRamiroDiaz: :point_up: Checked. Almost all places are checking if empty to write a default value (from the manifest for example).. The only one that's not checking if empty is https://github.com/abs007/okteto/blob/e25e1a665d057662498aeeb10ba807ccbb4a163b/pkg/model/devrc.go#L90, which doesn't seem like an issue. @jLopezbarb can you confirm?

cmd/up/up.go
abs007: I'm able to build the binary but I can't really run commands since I don't have access to the okteto cloud.

pchico83: @jLopezbarb Let's create an account in Okteto Cloud for @abs007

cmd/up/up.go
abs007: Won't an uninitialized `nil` slice have length 0 as well?

cmd/up/up.go
jLopezbarb: It's not an issue. We expect to overwrite the command with the one used in the args

cmd/up/up.go
jLopezbarb: > @jLopezbarb Let's say we pass in okteto up ["echo", "hello"] through the cli. Would there be any sort of weird error or would the ok just be false
 
Let's say we are running the command you proposed. Since there are no devs with the name `echo` we will set `echo hello` as the command for the dev that the user will select.

> @jLopezbarb Let's create an account in Okteto Cloud for @abs007

I'll work on it. @abs007 I will contact you in case I need anything and will notify  you whenever is ready

abs007: >Let's say we are running the command you proposed. Since there are no devs with the name echo we will set echo hello as the command for the dev that the user will select.

Ah, okay. I was just wondering if there might be any issues if the user supplies the commands in an array form.

jLopezbarb (approved):

teresaromero (approved):

cmd/up/up.go
pchico83: @abs007 instead of checking for a `dev container` name, let's use a flag `--command` to overwrite the command. I find the current approach a little ambiguous. And it might get worse if we support activating several dev containers with a single up command like this `okteto up api frontend`

abs007: I'm guessing this also means reverting `AddArgs()` to it's original functionality?

jLopezbarb:
We'll cut the new okteto CLI release next week. @abs007 are you ok if we merge it after we cut the release so we can have a full test cycle over this one?

abs007:
@jLopezbarb  Sure 👍

cmd/up/up.go
abs007: Should I reopen #3211 then?

jLopezbarb: Let's reopen it until we finish implementing this feat and will take a look when we close it

abs007: Sounds good 👍

abs007:
Can I also get confirmation to unit test the flag?

jLopezbarb:
> Can I also get confirmation if I should unit test the flag?

If you can add a unit test would be great

AgustinRamiroDiaz:
@abs007 congratulations on this work :rocket:

pchico83:
@abs007 how do you feel about updating our docs?
This is the section that covers the `okteto up` command:
https://www.okteto.com/docs/reference/cli/#up
And this is the repo:
https://github.com/okteto/docs/blob/main/src/content/reference/cli.mdx#up
No worries, we can work on it if you prefer to skip the docs changes

abs007:
@pchico83  Sure, I'll work on the docs and push the changes

cmd/up/up.go
abs007: @jLopezbarb  Should I reopen #3211 now?

PR #3216, opened 2022-11-10
Documentation grammar fixes

abs007:

# Proposed changes

Made fixes in code comments

### DIFF
--- a/pkg/discovery/contextresource.go
+++ b/pkg/discovery/contextresource.go
@@ -17,8 +17,8 @@ import (
 	oktetoLog "github.com/okteto/okteto/pkg/log"
 )
 
-// GetContextResourcePath returns the file that will load the context resource from
-// here. We only include files which have context information - no k8s or helm files
+// GetContextResourcePath returns the file that will load the context resource.
+// Here we only read files which have context information - no k8s or helm files
 func GetContextResourcePath(wd string) (string, error) {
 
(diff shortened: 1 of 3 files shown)


AgustinRamiroDiaz (approved):

pkg/discovery/contextresource.go
```diff
@@ -18,7 +18,7 @@ import (
 )
 
 // GetContextResourcePath returns the file that will load the context resource from
-// here we only include files which have context information - no k8s or helm files
+// here. We only include files which have context information - no k8s or helm files
```
abs007, line 21: Could someone double-check this?

AdrianPedriza (approved):

pkg/discovery/contextresource.go
ifbyol:
```suggestion
// GetContextResourcePath returns the file that will load the context resource.
// Here we only read files which have context information - no k8s or helm files
```

I think this would be more accurate. @jLopezbarb Can you confirm?

ifbyol (approved):
Minor comment, but when addressing, everything good!

pkg/discovery/contextresource.go
jLopezbarb: Yes, I agree. That would be better

jLopezbarb (approved):

PR #3219, opened 2022-11-11
Changes to print error message in main.go when the error is returned by the command executed and avoid duplicated messages

ifbyol:

# Proposed changes

Instead of printing the error in the logs when it happens and then, returning it, we just return the error and leave the responsibility of logging the message in main.go, which is already doing it. So I just removed duplicated messages we were logging it.

I've also moved the logic to "close" the logging (send the EOF message in the `done` stage) to the end of the main.go, so any error returned by the commands would be properly logged into the right stage and then, we just send the last message.

At some point, we need to revisit all the logging and how it is being integrated in the installer, as we are duplicating messages, setting stages in inner functions and it is really complex to follow and it is very error prone.

Fixes (issue belongs to other repository)


cmd/deploy/deploy.go
```diff
@@ -327,6 +326,9 @@ func (dc *DeployCommand) RunDeploy(ctx context.Context, deployOptions *Options)
 		return err
 	}
 
+	// We are doing some operations regarding to manifest load, so if it fails, it should belong to Load Manifest stage
+	oktetoLog.SetStage("")
```
ifbyol, line 330: This is moved a bit below just because `setDeployOptionsValuesFromManifest` function is doing some validation on the manifest, so if it fails, it should belong to `Load manifest` stage

cmd/deploy/deploy.go
```diff
@@ -398,16 +400,12 @@ func (dc *DeployCommand) RunDeploy(ctx context.Context, deployOptions *Options)
 	oktetoLog.EnableMasking()
 	err = dc.deploy(ctx, deployOptions)
 	oktetoLog.DisableMasking()
-	oktetoLog.SetStage("done")
-	oktetoLog.AddToBuffer(oktetoLog.InfoLevel, "EOF")
-	oktetoLog.SetStage("")
```
ifbyol, line 403: This is now done in main.go

cmd/deploy/deploy.go
```diff
@@ -398,16 +400,12 @@ func (dc *DeployCommand) RunDeploy(ctx context.Context, deployOptions *Options)
...
 
 	if err != nil {
 		if err == oktetoErrors.ErrIntSig {
 			return nil
 		}
 		err = oktetoErrors.UserError{E: err}
-		oktetoLog.AddToBuffer(oktetoLog.InfoLevel, err.Error())
```
ifbyol, line 410: We were adding as info an error which was already logged in most of the cases. Now, this is done in main.go just once

cmd/deploy/deploy.go
```diff
@@ -450,7 +448,6 @@ func (dc *DeployCommand) deploy(ctx context.Context, opts *Options) error {
 		oktetoLog.Information("Running '%s'", command.Name)
 		oktetoLog.SetStage(command.Name)
 		if err := dc.Executor.Execute(command, opts.Variables); err != nil {
-			oktetoLog.AddToBuffer(oktetoLog.ErrorLevel, "error executing command '%s': %s", command.Name, err.Error())
```
ifbyol, line 453: This was being logged along with the one in main.go, so in this case we just return error and the one handling this error (which is main.go) should be the one responsible of logging the message

cmd/deploy/deploy.go
```diff
@@ -460,7 +457,6 @@ func (dc *DeployCommand) deploy(ctx context.Context, opts *Options) error {
 	if opts.Manifest.Deploy.ComposeSection != nil {
 		oktetoLog.SetStage("Deploying compose")
 		if err := dc.deployStack(ctx, opts); err != nil {
-			oktetoLog.AddToBuffer(oktetoLog.ErrorLevel, "error deploying compose: %s", err.Error())
```
ifbyol, line 463: Same than previous comment

cmd/deploy/deploy.go
```diff
@@ -470,7 +466,6 @@ func (dc *DeployCommand) deploy(ctx context.Context, opts *Options) error {
 	if opts.Manifest.Deploy.Endpoints != nil {
 		oktetoLog.SetStage("Endpoints configuration")
 		if err := dc.deployEndpoints(ctx, opts); err != nil {
-			oktetoLog.AddToBuffer(oktetoLog.ErrorLevel, "error generating endpoints: %s", err.Error())
```
ifbyol, line 473: Same than previous comment

cmd/deploy/deploy.go
```diff
@@ -480,7 +475,6 @@ func (dc *DeployCommand) deploy(ctx context.Context, opts *Options) error {
 	if opts.Manifest.Deploy.Divert != nil && opts.Manifest.Deploy.Divert.Namespace != opts.Manifest.Namespace {
 		oktetoLog.SetStage("Divert configuration")
 		if err := dc.deployDivert(ctx, opts); err != nil {
-			oktetoLog.AddToBuffer(oktetoLog.ErrorLevel, "error creating divert: %s", err.Error())
```
ifbyol, line 483: Same than previous comment

cmd/pipeline/deploy.go
```diff
@@ -250,7 +250,7 @@ func (pc *Command) waitUntilRunning(ctx context.Context, name string, action *ty
 		defer wg.Done()
 		err := pc.streamPipelineLogs(waitCtx, name, action.Name)
 		if err != nil {
-			oktetoLog.Warning("there was an error streaming pipeline logs: %v", err)
```
ifbyol, line 253: @pchico83 based on your suggestion from today. This will include the log in the logs but it won't be showed to the user by default

pkg/log/json.go
```diff
@@ -193,7 +193,9 @@ func (w *JSONWriter) Fail(format string, args ...interface{}) {
 	log.out.Infof(format, args...)
 	msg := fmt.Sprintf("%s %s", errorSymbol, fmt.Sprintf(format, args...))
 	if msg != "" {
-
+		if log.stage == "" {
+			log.stage = "Internal server error"
```
ifbyol, line 197: I don't like too much to show `internal server error` as stage, maybe `unexpected error`. This is what we'll display in the UI when an error occurs outside a defined stage.

cc @pchico83 @rlamana

ifbyol:
Here you have an example on how it looks like in the UI for the same case. In this case, the pipeline is failing because a dependency is failing to be deployed:

Before the change:

Captura de Pantalla 2022-11-11 a las 18 25 14

After the change:

Captura de Pantalla 2022-11-11 a las 18 25 09


The manifest used for this test is:

```
dependencies:
  test:
    repository: http://nonexistentrepo/okteto/test.git
    wait: true

deploy:
  - kubectl apply -f k8s.yml
```

pchico83:
> Here you have an example on how it looks like in the UI for the same case. In this case, the pipeline is failing because a dependency is failing to be deployed:
> 
> Before the change:
> 
> Captura de Pantalla 2022-11-11 a las 18 25 14
> 
> After the change:
> 
> Captura de Pantalla 2022-11-11 a las 18 25 09
> 
> The manifest used for this test is:
> 
> ```
> dependencies:
>   test:
>     repository: http://nonexistentrepo/okteto/test.git
>     wait: true
> 
> deploy:
>   - kubectl apply -f k8s.yml
> ```

@ifbyol why do we show internal server error?

ifbyol:
> > Here you have an example on how it looks like in the UI for the same case. In this case, the pipeline is failing because a dependency is failing to be deployed:
> 
> > 
> 
> > Before the change:
> 
> > 
> 
> > Captura de Pantalla 2022-11-11 a las 18 25 14
> 
> > 
> 
> > After the change:
> 
> > 
> 
> > Captura de Pantalla 2022-11-11 a las 18 25 09
> 
> > 
> 
> > The manifest used for this test is:
> 
> > 
> 
> > ```
> 
> > dependencies:
> 
> >   test:
> 
> >     repository: http://nonexistentrepo/okteto/test.git
> 
> >     wait: true
> 
> > 
> 
> > deploy:
> 
> >   - kubectl apply -f k8s.yml
> 
> > ```
> 
> 
> 
> @ifbyol why do we show internal server error?

@pchico83 It is the default stage we have when the error happen in a phase where we don't have defined an stage. I added an inline comment about it

cmd/pipeline/deploy.go
teresaromero: what does it mean `include the log in the logs` ?

pkg/log/json.go
teresaromero: the absence of stage is always meaning is an error? in that case, the stage could be called `errors` ? 🤔 💭

pchico83:
> > > Here you have an example on how it looks like in the UI for the same case. In this case, the pipeline is failing because a dependency is failing to be deployed:
> > 
> > 
> > > 
> > 
> > 
> > > Before the change:
> > 
> > 
> > > 
> > 
> > 
> > > Captura de Pantalla 2022-11-11 a las 18 25 14
> > 
> > 
> > > 
> > 
> > 
> > > After the change:
> > 
> > 
> > > 
> > 
> > 
> > > Captura de Pantalla 2022-11-11 a las 18 25 09
> > 
> > 
> > > 
> > 
> > 
> > > The manifest used for this test is:
> > 
> > 
> > > 
> > 
> > 
> > > 
> > 
> > 
> > > dependencies:
> > 
> > 
> > > test:
> > 
> > 
> > > ```
> > > repository: http://nonexistentrepo/okteto/test.git
> > > ```
> > 
> > 
> > > ```
> > > wait: true
> > > ```
> > 
> > 
> > > 
> > 
> > 
> > > deploy:
> > 
> > 
> > > * kubectl apply -f k8s.yml
> > 
> > 
> > > 
> > 
> > 
> > @ifbyol why do we show internal server error?
> 
> @pchico83 It is the default stage we have when the error happen in a phase where we don't have defined an stage. I added an inline comment about it

@ifbyol does it mean we don't have a stage for the deploy `dependencies` logic?

pkg/log/json.go
ifbyol: Well, it is an error because the function is `Fail` and should be called only when there is an error.

The name we use for the stage, it will be the one that appears in the UI when an unexpected error happens, not sure `errors` would be a meaningful text to show

ifbyol:
> > > > Here you have an example on how it looks like in the UI for the same case. In this case, the pipeline is failing because a dependency is failing to be deployed:
> > > 
> > > 
> > > > 
> > > 
> > > 
> > > > Before the change:
> > > 
> > > 
> > > > 
> > > 
> > > 
> > > > Captura de Pantalla 2022-11-11 a las 18 25 14
> > > 
> > > 
> > > > 
> > > 
> > > 
> > > > After the change:
> > > 
> > > 
> > > > 
> > > 
> > > 
> > > > Captura de Pantalla 2022-11-11 a las 18 25 09
> > > 
> > > 
> > > > 
> > > 
> > > 
> > > > The manifest used for this test is:
> > > 
> > > 
> > > > 
> > > 
> > > 
> > > > 
> > > 
> > > 
> > > > dependencies:
> > > 
> > > 
> > > > test:
> > > 
> > > 
> > > > ```
> > > > repository: http://nonexistentrepo/okteto/test.git
> > > > ```
> > > 
> > > 
> > > > ```
> > > > wait: true
> > > > ```
> > > 
> > > 
> > > > 
> > > 
> > > 
> > > > deploy:
> > > 
> > > 
> > > > * kubectl apply -f k8s.yml
> > > 
> > > 
> > > > 
> > > 
> > > 
> > > @ifbyol why do we show internal server error?
> > 
> > 
> > @pchico83 It is the default stage we have when the error happen in a phase where we don't have defined an stage. I added an inline comment about it
> 
> @ifbyol does it mean we don't have a stage for the deploy `dependencies` logic?

@pchico83 Exactly, currently, there is not stage defined for that phase. We should probably revisit that and bear in mind that if the `wait` flag is not set, that stage would be almost empty. But I would do it in a separate issue

cmd/pipeline/deploy.go
ifbyol: The warning was being displayed to the user as part of the command output `!  there was an error streaming preview logs: http2: server sent GOAWAY and closed the connection; LastStreamID=119, ErrCode=NO_ERROR, debug=""` running it locally, this would display that line in yellow.

`include the log in the logs` means to include it as part of the logs included in the `okteto doctor` file but not being displayed to the final user by default (if the user specify --log-leve=debug it would appear)

jLopezbarb: I think it's that the log will be in the `log file` but won't be displayed to the user in the terminal

pchico83: @ifbyol I would also add a warning only with "pipeline logs cannot be streamed due to connectivity issues". This way, the user knows why the logs aren't streaming

ifbyol: ok, makes sense, added the warning!

jLopezbarb (approved):

teresaromero (approved):

PR #3221, opened 2022-11-14
Jlopezbarb/remove magic number max args

### LINKED ISSUE #3211: Possible bug in up.go

jLopezbarb:
# Proposed changes

Fixes #3211

Remove a magic number that was in up command and test the function containing it


teresaromero (approved):

jLopezbarb:
Closing it since it will be removed in #3215

PR #3222, opened 2022-11-14
Fix clone http

### LINKED ISSUE #3150: If my repo was cloned using `http` protocol, `okteto deploy` command fails
**Describe the bug**
If I clone a repository using `http` protocol, the clone is successfully because during the clone is redirected to `https`, but the `.git` configuration locally it specifies `http`. 

For example, If I clone http://github.com/okteto/movies, it successfully clones and you can see a log saying that it is redirected to https://github.com/okteto/movies. If you check the file `.git/config`, you can see the repo URL as http://github.com/okteto/movies.

If you execute `okteto deploy` from that folder, you get the following output

```
 i  Using nacho @ okteto.nacho.dev.okteto.net as context
 x  Could not detect repo protocol
```

The issues seems in [this function](https://github.com/okteto/okteto/blob/366c67cd31a2e85a0b67ea6f999792a7256e1ff8/cmd/deploy/command_configuration.go#L160), where the `http` is not being handled. 

**To Reproduce**

1. Clone http://github.com/okteto/movies
1. From the base repository folder execute `okteto deploy`
1. You'll see the error `Could not detect repo protocol`

**Expected behavior**
The deploy doesn't fail because the protocol is not valid. It should parse correctly the repo URL and keep progressing with next steps.

If `http` protocol is found, it should be converted to `https` as we do with `ssh`

- Version: `2.7.0`

teresaromero:
# Proposed changes

Fixes #3150 

- added unit test with supported cases
- added http as allowed schema
- refactor for switchSSHRepoToHTTPS
- updated error message

### DIFF
--- a/cmd/deploy/command_configuration.go
+++ b/cmd/deploy/command_configuration.go
@@ -19,7 +19,6 @@ import (
 	"net/url"
 	"os"
 	"reflect"
-	"strings"
 
 	"github.com/okteto/okteto/cmd/utils"
 	"github.com/okteto/okteto/pkg/cmd/pipeline"
@@ -30,6 +29,16 @@ import (
 	giturls "github.com/whilp/git-urls"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/client-go/kubernetes"
+)
+
+const (
+	sshScheme   = "ssh"
+	httpScheme  = "http"
+	httpsScheme = "https"
+)
+
+var (
+	schemeError = fmt.Errorf("could not detect repository scheme, please clone your repository using https or ssh")
 )
 
 func setDeployOptionsValuesFromManifest(ctx context.Context, deployOptions *Options, cwd string, c kubernetes.Interface) error {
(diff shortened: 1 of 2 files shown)


cmd/deploy/command_configuration.go
```diff
@@ -162,17 +171,16 @@ func switchSSHRepoToHTTPS(repo string) (*url.URL, error) {
 	if err != nil {
 		return nil, err
 	}
-	if repoURL.Scheme == "ssh" {
-		repoURL.Scheme = "https"
+	switch repoURL.Scheme {
+	case sshScheme, httpScheme:
+		repoURL.Scheme = httpsScheme
 		repoURL.User = nil
-		repoURL.Path = strings.TrimSuffix(repoURL.Path, ".git")
```
teresaromero, line 168: in the case of `ssh` there was a trim of `.git` which i removed in the refactor, i think this was not needed because the repository url is transformed to https which will require the `.git` isnt it?

ifbyol: I don't think https requires the .git suffix. Both ways (with and without) should work the same. I think in some places in the code we are removing it to check if 2 repos are the same or not

ifbyol (approved):

cmd/deploy/command_configuration.go
teresaromero: found https://github.com/okteto/okteto/blob/1a94bc9816077ee0aff8f8554e8c640380099b68/pkg/okteto/pipeline.go#L231 where a comparison is made, but is not related with the func refactored here.

Rarding the `/` thing mentioned at the linked func, in this case there is also no need to remove, as when the url is transformed to string this path will be correctly parsed.

cmd/deploy/command_configuration_test.go
```diff
@@ -168,3 +169,52 @@ func Test_mergeServicesToDeployFromOptionsAndManifest(t *testing.T) {
...
+		{
+			name:        "input-not-allowed",
+			repo:        "github.com/okteto/go-getting-started.git",
+			expectedErr: errUnsupportedScheme,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			url, err := switchSSHRepoToHTTPS(tt.repo)
```
jLopezbarb, line 214: NIT: Maybe we should modify the function name to `transformURLToHTTPS` or something like that. `switchSSHRepoToHTTPS` is not the best name because we can use more schemes than `SSH`. WDYT?

teresaromero: makes sense, i renamed to `switchRepoSchemaToHTTPS`

jLopezbarb (approved):

PR #3224, opened 2022-11-15
fix: private ingress annotations

### LINKED ISSUE #3182: The sample of private endpoints + docker compose is broken
**Describe the bug**
I am adding a label to define my private endpoint, but it's not working. My endpoint is still public.

**To Reproduce**
Deploy this compose:
```
services:
  api:
    labels:
      dev.okteto.com/private: "true"
    image: nginx
    ports:
      - "8080:80"
```
with `okteto deploy`.
The endpoint is public.    
**Expected behavior**
The endpoint should be private

**Additional context**
We should populate the labels/annotations in the compose service to every resource create by okteto, including the Ingress object created for the endpoint. This would fix the issue

jLopezbarb:

# Proposed changes

Fixes #3182

Add the annotations and labels from a compose/stack file into the kubernetes endpoints.
Create tests for deploying a kubernetes ingress

### DIFF
--- a/pkg/model/stack.go
+++ b/pkg/model/stack.go
@@ -54,6 +54,7 @@ type Stack struct {
 	Endpoints EndpointSpec           `yaml:"endpoints,omitempty"`
 }
 
+// ComposeServices represents the services declared in the compose
 type ComposeServices map[string]*Service
 
 // Service represents an okteto stack service
--- a/pkg/model/stack.go
+++ b/pkg/model/stack.go
@@ -50,11 +50,11 @@ type Stack struct {
 	Volumes   map[string]*VolumeSpec `yaml:"volumes,omitempty"`
 	Namespace string                 `yaml:"namespace,omitempty"`
 	Context   string                 `yaml:"context,omitempty"`
-	Services  composeServices        `yaml:"services,omitempty"`
+	Services  ComposeServices        `yaml:"services,omitempty"`
 	Endpoints EndpointSpec           `yaml:"endpoints,omitempty"`
 }
 
-type composeServices map[string]*Service
+type ComposeServices map[string]*Service
 
 // Service represents an okteto stack service
 type Service struct {
@@ -849,7 +849,7 @@ func setEnvironmentFromFile(svc *Service, filename string) error {
 	return nil
 }
 
-func (s composeServices) toGraph() graph {
+func (s ComposeServices) toGraph() graph {
 	g := graph{}
 	for svcName, svcInfo := range s {
 		dependsOnList := []string{}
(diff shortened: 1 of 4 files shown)


pchico83:
@jLopezbarb could you double check that the docker-compose sample here is working with this change?
https://www.okteto.com/docs/cloud/private-endpoints/

teresaromero (approved):

jLopezbarb:
> @jLopezbarb could you double check that the docker-compose sample here is working with this change?
> https://www.okteto.com/docs/cloud/private-endpoints/

It's not working because `ports` field is spelled `port` and returns a unmarshall error. Also I find it odd that the examples use an image to build when we don't specify how the Dockerfile should be defined. 

I uploaded [a PR fixing those "problems"](https://github.com/okteto/docs/pull/254) In the docs repo

Double checked that setting properly the port field and having these changes work fine

pkg/model/stack.go
```diff
@@ -50,11 +50,11 @@ type Stack struct {
 	Volumes   map[string]*VolumeSpec `yaml:"volumes,omitempty"`
 	Namespace string                 `yaml:"namespace,omitempty"`
 	Context   string                 `yaml:"context,omitempty"`
-	Services  composeServices        `yaml:"services,omitempty"`
+	Services  ComposeServices        `yaml:"services,omitempty"`
 	Endpoints EndpointSpec           `yaml:"endpoints,omitempty"`
 }
 
-type composeServices map[string]*Service
+type ComposeServices map[string]*Service
```
ifbyol, line 57: Nit: Can we include a comment as we are now exporting it?

pkg/model/stack.go
```diff
@@ -50,11 +50,11 @@ type Stack struct {
 	Volumes   map[string]*VolumeSpec `yaml:"volumes,omitempty"`
 	Namespace string                 `yaml:"namespace,omitempty"`
 	Context   string                 `yaml:"context,omitempty"`
-	Services  composeServices        `yaml:"services,omitempty"`
+	Services  ComposeServices        `yaml:"services,omitempty"`
```
ifbyol, line 53: Nit: it is a bit odd that we need to make this public just for a test. How is this created in other places? as a result of marshaling/unmarshaling?

ifbyol (approved):
Minor comments, but it looks good!

pkg/model/stack.go
jLopezbarb: All other tests referring to services are declared in `model` package. We could be using `map[string]*model.Service` instead but I think we should be using the defined struct in all other tests too...

PR #3230, opened 2022-11-17
Empty cfgmap repository if not valid repo schema

### LINKED ISSUE #3228: I cannot run `okteto deploy` in a repo clone from AzureDevops Git
**Describe the bug**
I cloned my AzureDevops repo and `okteto deploy` fails with:
```
 Could not detect repo protocol
```

**Expected behavior**
We shouldn't fail parsing the git URL. The default behavior on this case should be to get the name of the current folder and don't store either the git url or the branch in the deploy configmap.

teresaromero:
# Proposed changes

Fixes #3228 

When parsing the `.git/config` url when cloned repo, if not recognised as `http`, https or ssh, we returned an error.
This fix allows not recognised schemas to work with okteto cli with local repositories.

Tested the scenario with repo go-gettting-started, changing the url param of the git config then run `okteto deploy`
UI output was showing a message not allowing redeploy and name for dev environment was last part of local repository url

### DIFF
--- a/cmd/deploy/command_configuration.go
+++ b/cmd/deploy/command_configuration.go
@@ -109,7 +109,7 @@ func mergeServicesToDeployFromOptionsAndManifest(deployOptions *Options) {
 	}
 }
 
-func addEnvVars(ctx context.Context, cwd string) error {
+func addEnvVars(ctx context.Context, cwd string) {
 	if os.Getenv(model.OktetoGitBranchEnvVar) == "" {
 		branch, err := utils.GetBranch(cwd)
 		if err != nil {
@@ -125,17 +125,12 @@ func addEnvVars(ctx context.Context, cwd string) error {
 		}
 
 		if repo != "" {
-			repoHTTPS, err := switchRepoSchemaToHTTPS(repo)
-			if err != nil {
-				return err
-			}
-			if repoHTTPS != nil {
-				repo = repoHTTPS.String()
-			} else {
-				// ir repo was parsed but hasnt got a valid schema
+			repoHTTPS := switchRepoSchemaToHTTPS(repo)
+			if repoHTTPS == nil {
 				// fallback to empty repository
-				oktetoLog.Warning("retrieved local repository")
 				repo = ""
+			} else {
+				repo = repoHTTPS.String()
 			}
 		}
 		os.Setenv(model.GithubRepositoryEnvVar, repo)
@@ -165,23 +160,25 @@ func addEnvVars(ctx context.Context, cwd string) error {
 		os.Setenv(model.OktetoTokenEnvVar, okteto.Context().Token)
 	}
 	oktetoLog.AddMaskedWord(os.Getenv(model.OktetoTokenEnvVar))
-	return nil
 }
 
-func switchRepoSchemaToHTTPS(repo string) (*url.URL, error) {
+func switchRepoSchemaToHTTPS(repo string) *url.URL {
 	repoURL, err := giturls.Parse(repo)
 	if err != nil {
-		return nil, err
+		return nil
 	}
 	switch repoURL.Scheme {
 	case sshScheme, httpScheme:
 		repoURL.Scheme = httpsScheme
 		repoURL.User = nil
-		return repoURL, nil
+		return repoURL
 	case httpsScheme:
-		return repoURL, nil
+		return repoURL
+	default:
+		// if repo was parsed but hasnt got a valid schema
+		oktetoLog.Info("retrieved local repository")
+		return nil
 	}
-	return nil, nil
 }
 
 func updateConfigMapStatusError(ctx context.Context, cfg *corev1.ConfigMap, c kubernetes.Interface, data *pipeline.CfgData, errMain error) error {
--- a/cmd/deploy/command_configuration.go
+++ b/cmd/deploy/command_configuration.go
@@ -176,7 +176,7 @@ func switchRepoSchemaToHTTPS(repo string) *url.URL {
 		return repoURL
 	default:
 		// if repo was parsed but has not a valid schema
-		oktetoLog.Info("retrieved local repository")
+		oktetoLog.Infof("retrieved schema for %s - %s", repo, repoURL.Scheme)
 		return nil
 	}
 }
--- a/cmd/deploy/command_configuration.go
+++ b/cmd/deploy/command_configuration.go
@@ -15,7 +15,6 @@ package deploy
 
 import (
 	"context"
-	"fmt"
 	"net/url"
 	"os"
 	"reflect"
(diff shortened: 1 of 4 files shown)


cmd/deploy/command_configuration.go
```diff
@@ -134,7 +129,14 @@ func addEnvVars(ctx context.Context, cwd string) error {
 			if err != nil {
 				return err
 			}
-			repo = repoHTTPS.String()
+			if repoHTTPS != nil {
+				repo = repoHTTPS.String()
+			} else {
+				// ir repo was parsed but hasnt got a valid schema
+				// fallback to empty repository
+				oktetoLog.Warning("retrieved local repository")
```
jLopezbarb, line 137: I wouldn't show a warning to the user, What do you think about logging it into the logs files with an info level?

cmd/deploy/command_configuration.go
```diff
@@ -134,7 +129,14 @@ func addEnvVars(ctx context.Context, cwd string) error {
 			if err != nil {
 				return err
 			}
-			repo = repoHTTPS.String()
+			if repoHTTPS != nil {
+				repo = repoHTTPS.String()
+			} else {
+				// ir repo was parsed but hasnt got a valid schema
```
jLopezbarb, line 135:
```suggestion
				// if repo was parsed but hasnt got a valid schema
```

cmd/deploy/command_configuration.go
```diff
@@ -178,9 +180,8 @@ func switchRepoSchemaToHTTPS(repo string) (*url.URL, error) {
 		return repoURL, nil
 	case httpsScheme:
 		return repoURL, nil
-	default:
```
jLopezbarb, line 181: I would keep the default and log it instead of returning the error in case we need it to debug upcoming bugs

cmd/deploy/command_configuration.go
```diff
@@ -178,9 +180,8 @@ func switchRepoSchemaToHTTPS(repo string) (*url.URL, error) {
...
-		return nil, errUnsupportedScheme
 	}
+	return nil, nil
```
ifbyol, line 184: I think the goal is to don't return any error. We can return just nil in line 174 and make that the function `switchRepoSchemaToHTTPS` returns just a pointer to a URL. If there is an error parsing the URL, we just return nil.

In that way, we can simplify a bit the code because we can change the function `addEnvVars` to not return an error. This case was the only one generating that scenario

ifbyol: It would be something like this:

```
diff --git a/cmd/deploy/command_configuration.go b/cmd/deploy/command_configuration.go
index 6489ce2b..470e11dc 100644
--- a/cmd/deploy/command_configuration.go
+++ b/cmd/deploy/command_configuration.go
@@ -114,7 +114,7 @@ func mergeServicesToDeployFromOptionsAndManifest(deployOptions *Options) {
        }
 }

-func addEnvVars(ctx context.Context, cwd string) error {
+func addEnvVars(ctx context.Context, cwd string) {
        if os.Getenv(model.OktetoGitBranchEnvVar) == "" {
                branch, err := utils.GetBranch(cwd)
                if err != nil {
@@ -130,11 +130,12 @@ func addEnvVars(ctx context.Context, cwd string) error {
                }

                if repo != "" {
-                       repoHTTPS, err := switchRepoSchemaToHTTPS(repo)
-                       if err != nil {
-                               return err
+                       repoHTTPS := switchRepoSchemaToHTTPS(repo)
+                       if repoHTTPS == nil {
+                               repo = ""
+                       } else {
+                               repo = repoHTTPS.String()
                        }
-                       repo = repoHTTPS.String()
                }
                os.Setenv(model.GithubRepositoryEnvVar, repo)
        }
@@ -163,23 +164,22 @@ func addEnvVars(ctx context.Context, cwd string) error {
                os.Setenv(model.OktetoTokenEnvVar, okteto.Context().Token)
        }
        oktetoLog.AddMaskedWord(os.Getenv(model.OktetoTokenEnvVar))
-       return nil
 }

-func switchRepoSchemaToHTTPS(repo string) (*url.URL, error) {
+func switchRepoSchemaToHTTPS(repo string) *url.URL {
        repoURL, err := giturls.Parse(repo)
        if err != nil {
-               return nil, err
+               return nil
        }
        switch repoURL.Scheme {
        case sshScheme, httpScheme:
                repoURL.Scheme = httpsScheme
                repoURL.User = nil
-               return repoURL, nil
+               return repoURL
        case httpsScheme:
-               return repoURL, nil
+               return repoURL
        default:
-               return nil, errUnsupportedScheme
+               return nil
        }
 }

diff --git a/cmd/deploy/deploy.go b/cmd/deploy/deploy.go
index 7c457566..36c4c313 100644
--- a/cmd/deploy/deploy.go
+++ b/cmd/deploy/deploy.go
@@ -266,9 +266,7 @@ func (dc *DeployCommand) RunDeploy(ctx context.Context, deployOptions *Options)
                return err
        }

-       if err := addEnvVars(ctx, cwd); err != nil {
-               return err
-       }
+       addEnvVars(ctx, cwd)
        oktetoLog.Debugf("creating temporal kubeconfig file '%s'", dc.TempKubeconfigFile)
        if err := dc.Kubeconfig.Modify(dc.Proxy.GetPort(), dc.Proxy.GetToken(), dc.TempKubeconfigFile); err != nil {
                oktetoLog.Infof("could not create temporal kubeconfig %s", err)
```

cmd/deploy/command_configuration.go
```diff
@@ -163,23 +160,24 @@ func addEnvVars(ctx context.Context, cwd string) error {
...
 	oktetoLog.AddMaskedWord(os.Getenv(model.OktetoTokenEnvVar))
-	return nil
 }
 
-func switchRepoSchemaToHTTPS(repo string) (*url.URL, error) {
+func switchRepoSchemaToHTTPS(repo string) *url.URL {
 	repoURL, err := giturls.Parse(repo)
 	if err != nil {
-		return nil, err
+		return nil
```
ifbyol, line 168: Might be interesting to include a info or debug with the error to have in our logs but not display it to the user

jLopezbarb (approved):

cmd/deploy/command_configuration.go
```diff
@@ -163,23 +160,24 @@ func addEnvVars(ctx context.Context, cwd string) error {
...
 		repoURL.User = nil
-		return repoURL, nil
+		return repoURL
 	case httpsScheme:
-		return repoURL, nil
+		return repoURL
 	default:
-		return nil, errUnsupportedScheme
+		// if repo was parsed but has not a valid schema
+		oktetoLog.Info("retrieved local repository")
```
ifbyol, line 179: I would include the schema to know which one was detected

ifbyol (approved):
Minor comments about logging but it looks good!

PR #3231, opened 2022-11-17
Alternative 1: Fix missing error messages in the installer

ifbyol:

Fixes https://github.com/okteto/app/issues/4732

# Proposed changes

Alternative 1: We just include logic with the `Fail` function for the json logger to include the default error stage if there is not any previous stage set.

This present 1 main drawback. When the error happens in the deploy phase (executing commands within an okteto manifest, deploying a compose, deploying a kubectl manifest or a helm application), the error is printed twice. One within the stage related to the command and another one with the `Internal server error` stage.
Captura de Pantalla 2022-11-17 a las 12 44 01

It only happens when the error occurs in the "deploy" phase due to the way we handle the error in that case to send the `done` stage at the end.

The main advantage of this is that any unexpected error happened during the execution of deploy or destroy command would be printed

### DIFF
--- a/cmd/deploy/deploy.go
+++ b/cmd/deploy/deploy.go
@@ -401,7 +401,6 @@ func (dc *DeployCommand) RunDeploy(ctx context.Context, deployOptions *Options) 
 	oktetoLog.DisableMasking()
 	oktetoLog.SetStage("done")
 	oktetoLog.AddToBuffer(oktetoLog.InfoLevel, "EOF")
-	oktetoLog.SetStage("")
 
 	if err != nil {
 		if err == oktetoErrors.ErrIntSig {
--- a/cmd/pipeline/deploy.go
+++ b/cmd/pipeline/deploy.go
@@ -250,7 +250,8 @@ func (pc *Command) waitUntilRunning(ctx context.Context, name string, action *ty
 		defer wg.Done()
 		err := pc.streamPipelineLogs(waitCtx, name, action.Name)
 		if err != nil {
-			oktetoLog.Warning("there was an error streaming pipeline logs: %v", err)
+			oktetoLog.Warning("pipeline logs cannot be streamed due to connectivity issues")
+			oktetoLog.Infof("pipeline logs cannot be streamed due to connectivity issues: %v", err)
 		}
 	}(&wg)
 
--- a/cmd/preview/deploy.go
+++ b/cmd/preview/deploy.go
@@ -159,7 +159,8 @@ func (pw *Command) waitUntilRunning(ctx context.Context, name string, a *types.A
 		defer wg.Done()
 		err := pw.okClient.Pipeline().StreamLogs(waitCtx, name, a.Name)
 		if err != nil {
-			oktetoLog.Warning("there was an error streaming preview logs: %v", err)
+			oktetoLog.Warning("preview logs cannot be streamed due to connectivity issues")
+			oktetoLog.Infof("preview logs cannot be streamed due to connectivity issues: %v", err)
 		}
 	}(&wg)
 


cmd/deploy/deploy.go
```diff
@@ -327,6 +326,8 @@ func (dc *DeployCommand) RunDeploy(ctx context.Context, deployOptions *Options)
 		return err
 	}
 
+	oktetoLog.SetStage("")
```
ifbyol, line 329: This is just moved a bit below because `setDeployOptionsValuesFromManifest` function validates some things regarding the manifest, so if it fails, it should belong to `Load manifest` stage

jLopezbarb:
I don't like too much this solution because most of the errors that the people may face are the ones from the deploy commands, which can lead to confusion having two different stages. 

We could control in the logger if there has been a previous error stage to add it the error message to that stage instead of creating a new one. Will that fix the issue?

ifbyol:
> I don't like too much this solution because most of the errors that the people may face are the ones from the deploy commands, which can lead to confusion having two different stages.

Well, that is better that not display the error as we have now 😂 . Agree that is not the best experience

> We could control in the logger if there has been a previous error stage to add it the error message to that stage instead of creating a new one. Will that fix the issue?

error stage or error log? That would couple the logic to the order of the logs. What would happen if I get an error message with a stage, then an info message and then, an unexpected error? What should we do there?

Another solution would be to move the logic to send the `done` stage just at the end of the command execution (in the `main.go` file). That wouldn't create the `Internal server error` stage and it would print the error within the command stage, but it would have a drawback when executing the command locally. In that scenario, we wouldn't be storing the log message with the `done` stage in the config map and any deploy done via `okteto deploy` locally would display the last stage with this icon instead of the green check:

Captura de Pantalla 2022-11-17 a las 8 43 43

jLopezbarb:
> error stage or error log? That would couple the logic to the order of the logs. What would happen if I get an error message with a stage, then an info message and then, an unexpected error? What should we do there?

What I meant is that we could add a new field in the [json writer](https://github.com/okteto/okteto/blob/a4897660a3e57dfb9ee16a0d9a51c3d0fc6ac44b/pkg/log/json.go#L27)
 storing the information about the stage that has thrown the error. If one stage wants to store two different errors we should be able to do it but if another stage wants to store an error we should fail because a previous stage has already thrown the error (I explained myself very badly but I hope it can be understood 😅 ).

cmd/pipeline/deploy.go
```diff
@@ -250,7 +250,8 @@ func (pc *Command) waitUntilRunning(ctx context.Context, name string, action *ty
 		defer wg.Done()
 		err := pc.streamPipelineLogs(waitCtx, name, action.Name)
 		if err != nil {
-			oktetoLog.Warning("there was an error streaming pipeline logs: %v", err)
+			oktetoLog.Warning("pipeline logs cannot be streamed due to connectivity issues")
+			oktetoLog.Infof("pipeline logs cannot be streamed due to connectivity issues: %v", err)
```
ifbyol, line 254: This change and the one in `preview/deploy.go` are a couple of improvements in logs when there is some issue streaming logs. They were reverted with the resto of changes and I'm adding them back again

cmd/deploy/deploy.go
```diff
@@ -400,7 +401,6 @@ func (dc *DeployCommand) RunDeploy(ctx context.Context, deployOptions *Options)
 	oktetoLog.DisableMasking()
 	oktetoLog.SetStage("done")
 	oktetoLog.AddToBuffer(oktetoLog.InfoLevel, "EOF")
-	oktetoLog.SetStage("")
```
ifbyol, line 403: In case of error, we don't clear the stage. This implies that the log printed by `main.go` would be under `done` stage, but any log under that stage is ignored, which is what we want in this case. It is not an idea solution, but for now, it is the best to fix the issue until we improve all the logging

ifbyol:
> > error stage or error log? That would couple the logic to the order of the logs. What would happen if I get an error message with a stage, then an info message and then, an unexpected error? What should we do there?
> 
> What I meant is that we could add a new field in the [json writer](https://github.com/okteto/okteto/blob/a4897660a3e57dfb9ee16a0d9a51c3d0fc6ac44b/pkg/log/json.go#L27) storing the information about the stage that has thrown the error. If one stage wants to store two different errors we should be able to do it but if another stage wants to store an error we should fail because a previous stage has already thrown the error (I explained myself very badly but I hope it can be understood 😅 ).

Yeah, that might work, but it is based on the assumption that 2 stages cannot contain an error, which might be valid now, but not in the future. I don't see why 2 stages couldn't have an error stage.

Anyways, after talking with @pchico83 , removing the line that was clearing the stage when the deploy ends, would do the trick. We would only clear the stage when the deploy success, like we do in the rest of stages, we just clean the stage when that operation successes.

This would imply that the final error printed by `main.go` would be under the stage `done`, but we are ignoring those logs, so for now it is ok as the error in that specific case is already printed. With that, the only drawback in UX is gone and I think this is the best solution and the way to go for now

Example of unexpected error

Captura de Pantalla 2022-11-18 a las 13 18 34

Example of error in the command execution

Captura de Pantalla 2022-11-18 a las 13 20 55

teresaromero (approved):
LGTM

jLopezbarb (approved):

PR #3232, opened 2022-11-17
Alternative 2: Fix missing error messages in the installer

ifbyol:

Fixes https://github.com/okteto/app/issues/4732

# Proposed changes

Create a specific function in `deploy.go` to log when an error happens in a place where there is not a defined stage. That function sets the default error stage and logs the error. This is another alternative to fix the issue with the logs in the installer and tries to avoid the duplicated error messages generated in the [alternative 1](https://github.com/okteto/okteto/pull/3231).

The main drawbacks of this case:
* We would need to explicitly call to the function in all the places where a stage is not defined and potentially, we could miss it in the future when including some new functionality or refactoring the code. It is very error prone
* We would need to do this also for the destroy as it is not a generic solution
* There is a phase in which it might generate an odd situation. This phase is when it has to build any image. As the stage for `Building image ...` is done in a inner function and not in the main deploy function as the rest of stages, it is hard to know which error is thrown _within_ the stage or not. So applying this approach to that specific case, we would create a `Internal server error` stage to display an error in the build.
Captura de Pantalla 2022-11-17 a las 13 30 30

### DIFF
--- a/cmd/deploy/deploy.go
+++ b/cmd/deploy/deploy.go
@@ -258,20 +258,24 @@ func Deploy(ctx context.Context) *cobra.Command {
 func (dc *DeployCommand) RunDeploy(ctx context.Context, deployOptions *Options) error {
 	cwd, err := os.Getwd()
 	if err != nil {
+		logUnstageError("failed to get the current working directory: %s", err.Error())
 		return fmt.Errorf("failed to get the current working directory: %w", err)
 	}
 
 	c, _, err := dc.K8sClientProvider.Provide(okteto.Context().Cfg)
 	if err != nil {
+		logUnstageError("could not read kubernetes configuration: %s", err.Error())
 		return err
 	}
 
 	if err := addEnvVars(ctx, cwd); err != nil {
+		logUnstageError("there was an un expected error: %s", err.Error())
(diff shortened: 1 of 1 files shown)


jLopezbarb:
Maybe we can create custom errors for whenever we are building and when the code is at the `deploy.go`  we can cast them and assign the correct stage. Will that fix that weird error?

ifbyol:
> Maybe we can create custom errors for whenever we are building and when the code is at the `deploy.go` we can cast them and assign the correct stage. Will that fix that weird error?

That would fix it, but we would be coupling error handling with logging, which I don't think we should do, and it doesn't solve the problem of code maintainability (anytime we modify or add things in deploy or destroy, we need to remember to handle/use this new function)

jLopezbarb:
We need to revisit the whole logging in the CLI, since it has some obvious disadvantages to use it as we have it today. If we really need this fix as soon as possible I'd opt for one of the two alternatives and rethink how we couple CLI logs with the UI

ifbyol:
Closing this in favor of the [alternative 1](https://github.com/okteto/okteto/pull/3231) which seems the way to go

Issue #3233, opened 2022-11-18
Refactor buildkit related code


jLopezbarb:
**Is your feature request related to a problem? Please describe.**

The way of building images is complex, can't be tested and we have two different ways of building an image depending if it's an okteto context or if it's a vanilla cluster.

**Describe the solution you'd like**

We should try to have a unique solution for both context (okteto and vanilla cluster) like the one proposed in [buildx](https://github.com/docker/buildx). This will make the code easier to maintain/add new features.

We should be able to test what happens if certain errors are returned by the build (transient errors like connection issues) or what happens if the image has not been added to the registry. This will make the code less error prone and will prevent that some scenarios to break.

We'll still need to differentiate between both context since we need to add the certificate but will make it easier to add new features.

PR #3234, opened 2022-11-18
feat: add platform flag to build command

### LINKED ISSUE #2921: Support building on different platform
I want to set the platform that my images should be built. It's common that the architecture from local is different to the one that is on the cluster so setting it can be useful on some scenarios

I'd like to have a flag on the build command and be able to set it on the context config file so all my builds on a cluster are built on the same platform

jLopezbarb:

# Proposed changes

Fixes #2921 

## Proposed changes:

- Build with local docker: Only supports 1 platform. This is also happening with `docker build` commnand. The response from the server when we try to add more than one is `Error response from daemon: "amd64,linux" is an invalid component of "linux/amd64,linux/arm64": platform specifier component must match "^[A-Za-z0-9_-]+$": invalid argument`. which is the same thrown by docker. We could get rid off this error if we move to a `buildx` approach where we don't depend on docker cli

- build in Okteto: It accepts multiple platforms just like the `buildctl` or `buildx` commands. 

### Further improvement:

In order to allow multiple platforms in local docker, simplify the code (right now we have one for local buildkit and another one for remote buildkit in okteto) and make it testable we need to refactor the code related to calling the server. I created #3233 to improve the build with local docker and reduce the complexity of the code

### DIFF
--- a/cmd/build/command.go
+++ b/cmd/build/command.go
@@ -106,6 +106,7 @@ func Build(ctx context.Context) *cobra.Command {
 	cmd.Flags().StringVarP(&options.OutputMode, "progress", "", oktetoLog.TTYFormat, "show plain/tty build output")
 	cmd.Flags().StringArrayVar(&options.BuildArgs, "build-arg", nil, "set build-time variables")
 	cmd.Flags().StringArrayVar(&options.Secrets, "secret", nil, "secret files exposed to the build. Format: id=mysecret,src=/local/secret")
+	cmd.Flags().StringVar(&options.Platform, "platform", "", "set platform if server is multi-platform capable")
 	cmd.Flags().StringVarP(&options.Namespace, "namespace", "n", "", "namespace against which the image will be consumed. Default is the one defined at okteto context or okteto manifest")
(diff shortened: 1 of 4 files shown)


ifbyol (approved):
Maybe not for this PR, but would it make sense to include an option in the manifest to specify the platform?

cmd/build/command.go
```diff
@@ -106,6 +106,7 @@ func Build(ctx context.Context) *cobra.Command {
 	cmd.Flags().StringVarP(&options.OutputMode, "progress", "", oktetoLog.TTYFormat, "show plain/tty build output")
 	cmd.Flags().StringArrayVar(&options.BuildArgs, "build-arg", nil, "set build-time variables")
 	cmd.Flags().StringArrayVar(&options.Secrets, "secret", nil, "secret files exposed to the build. Format: id=mysecret,src=/local/secret")
+	cmd.Flags().StringVar(&options.Platform, "platform", "", "set platform if server is multi-platform capable")
```
pchico83, line 109: what's the default value?

PR #3236, opened 2022-11-21
fix: show self signed warning when insecure flag

### LINKED ISSUE #3183: Don't print self-signed warning when connecting to server endpoint with publicly (or locally) trusted certificates
**Describe the bug**
Okteto CLI prints self signed warning when connecting to a server endpoint with publicly (or locally) trusted certificates.

**To Reproduce**

1. Deploy self-signed Okteto Helm Chart
1. Put a proxy in front of Okteto with publicly (or locally) trusted certificates
1. Use the CLI
1. Observe the self signed warning error

**Expected behavior**
If server endpoint from client perspective expose publicly (or locally) trusted certificates, don't show the warning.

- OS
- Version

**Additional context**

jpf-okteto:
Resolve #3183.

Signed-off-by: Javier Provecho Fernández (Okteto) jpf@okteto.com

### DIFF
--- a/pkg/okteto/context.go
+++ b/pkg/okteto/context.go
@@ -73,6 +73,7 @@ type OktetoContext struct {
 	Analytics         bool                 `json:"-" yaml:"-"`
 	ClusterType       string               `json:"-" yaml:"-"`
 	IsOkteto          bool                 `json:"isOkteto,omitempty" yaml:"isOkteto,omitempty"`
+	IsInsecure        bool                 `json:"isInsecure,omitempty" yaml:"isInsecure,omitempty"`
 }
 
 // OktetoContextViewer contains info to show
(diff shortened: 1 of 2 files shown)


teresaromero (approved):

pkg/okteto/context.go
```diff
@@ -73,6 +73,7 @@ type OktetoContext struct {
 	Analytics         bool                 `json:"-" yaml:"-"`
 	ClusterType       string               `json:"-" yaml:"-"`
 	IsOkteto          bool                 `json:"isOkteto,omitempty" yaml:"isOkteto,omitempty"`
+	IsInsecure        bool                 `json:"isInsecure,omitempty" yaml:"isInsecure,omitempty"`
```
ifbyol, line 76: Do we need the tags for json and yaml?

ifbyol: I mean, having them could imply that those fields are printed/stored in a file while marshaling to json or yaml

ifbyol: `Analytics` doesn't have flags for yaml and it can be read [here](https://github.com/okteto/okteto/blob/56df083df7af89b6373ba812bfc4a3519732af7c/pkg/analytics/track.go#L455).

Be aware that this changes would also make that property to appear in `IsInsecure` in the output of `okteto ctx show` command. Which maybe it is what we want, but just to make sure we want that

jLopezbarb (approved):

PR #3244, opened 2022-11-23
fix: show info if context is initialized by env vars

### LINKED ISSUE #2378: If `okteto context` is using a token from an environment variable, it would be useful to show that information to the user
If I set at some point the environment variable `OKTETO_TOKEN` and I execute `okteto context use https://foo`, it will try to login using the `OKTETO_TOKEN` environment variable but I might not realize I have that env set. It would be useful if the CLI displays an info message letting me know that it will use the env var.

It can be confusing if the token set is no longer available as I'm not specifying the token explicitly. 

If the command will use the value of the environment variable, I would like to see an info message letting me know.

Other option could be to execute a normal auth flow if the auth with the token fails

**Additional context**
It happened to one of our customers. They have set a personal access token in the environment variable and at some point, that access token was removed, so executing `okteto context use` command was failing all the time.

jLopezbarb:

# Proposed changes

Fixes #2378

This PR adds a warning message whenever the user is initialising the context from env vars

### DIFF
--- a/cmd/context/use.go
+++ b/cmd/context/use.go
@@ -90,8 +90,8 @@ Or a Kubernetes context:
 
 func (c *ContextCommand) Run(ctx context.Context, ctxOptions *ContextOptions) error {
 	ctxStore := okteto.ContextStore()
-	ctxOptions.initFromContext()
 	ctxOptions.initFromEnvVars()
+	ctxOptions.initFromContext()
 
 	if ctxOptions.Token == "" && kubeconfig.InCluster() && !isValidCluster(ctxOptions.Context) {
 		if ctxOptions.IsCtxCommand {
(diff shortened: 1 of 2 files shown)


cmd/context/use.go
```diff
@@ -90,8 +90,8 @@ Or a Kubernetes context:
 
 func (c *ContextCommand) Run(ctx context.Context, ctxOptions *ContextOptions) error {
 	ctxStore := okteto.ContextStore()
-	ctxOptions.initFromContext()
 	ctxOptions.initFromEnvVars()
+	ctxOptions.initFromContext()
```
jLopezbarb, line 94: We were initialising always from the env vars so only `OKTETO_TOKEN` could be replaced

jLopezbarb:
## Screenshoots:
Captura de Pantalla 2022-11-23 a las 12 35 46

teresaromero (approved):

PR #3246, opened 2022-11-23
fix: e2e test

jLopezbarb:

# Proposed changes

Fixes e2e deprecated tests

This change was intended to be safe but it had some deeper implications on some commands

### DIFF
--- a/cmd/context/use.go
+++ b/cmd/context/use.go
@@ -90,6 +90,9 @@ Or a Kubernetes context:
 
 func (c *ContextCommand) Run(ctx context.Context, ctxOptions *ContextOptions) error {
 	ctxStore := okteto.ContextStore()
+
+	// We have to maintain this order to not break some commands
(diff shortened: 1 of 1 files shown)


jLopezbarb:
The change was introduced in #3244

ifbyol:
Don't we have to revert the other changes? What are the deeper implications? To put in the context of the ticket or similar or document somewhere. Maybe a comment there why that order is needed

jLopezbarb:
> Don't we have to revert the other changes? 

The other change is just printing the warning which is what the original issue was about.

> What are the deeper implications? To put in the context of the ticket or similar or document somewhere. Maybe a comment there why that order is needed

Deeper implications is that we can't call two times `okteto context use`. I added this change in the previous PR because I didn't think about those commands. I will investigate which commands are affected and create a new issue based on it to change this properly

ifbyol (approved):

PR #3248, opened 2022-11-25
feat: add namespace to dependencies

### LINKED ISSUE #3048: Support `namespace` field in the extended `dependencies` notation
I want to automate the deployment of a dependency always in the same namespace, because this is a shared namespace by all my team members.

Support `namespace` in the `dependencies` extended notation, something like this:

```
dependencies:
  postgresql:
       repository: https://github.com/postgress
       wait: true
       namespace: integration
```

jLopezbarb:

# Proposed changes

Fixes #3048

- Added namespace to extended dependencies syntax
- Differentiate between `Flags` and command `options`. This way we only use options through the command execution and we don't have to recalculate things like the namespace used.
- Use `options.namespace` instead of `context().Namespace`. This way we don't have to change the original namespace
- Add `namespace` argument to graphql calls instead of calculating it with the okteto context when we call the graphql function
- Modify interface to fulfill these new changes


pchico83:
> @pchico83 @jLopezbarb Should we consider the namespace option for the `destroy` command?

@ifbyol we don't have a destroy command for dependencies. The user would need to go to the given namespace, and destroy the pipeline from there (if I understand your question correctly)

ifbyol:
> > @pchico83 @jLopezbarb Should we consider the namespace option for the `destroy` command?
> 
> @ifbyol we don't have a destroy command for dependencies. The user would need to go to the given namespace, and destroy the pipeline from there (if I understand your question correctly)

@pchico83 `okteto destroy` has a flag to destroy dependencies: https://github.com/okteto/okteto/blob/master/cmd/destroy/destroy.go#L194. Shouldn't we take the namespace in those cases too?

pchico83:
> > > @pchico83 @jLopezbarb Should we consider the namespace option for the `destroy` command?
> > 
> > 
> > @ifbyol we don't have a destroy command for dependencies. The user would need to go to the given namespace, and destroy the pipeline from there (if I understand your question correctly)
> 
> @pchico83 `okteto destroy` has a flag to destroy dependencies: https://github.com/okteto/okteto/blob/master/cmd/destroy/destroy.go#L194. Shouldn't we take the namespace in those cases too?

@ifbyol oh yes, you are right. we should take it into account on destroy too

cmd/pipeline/deploy.go
```diff
@@ -34,8 +34,24 @@ import (
 	"github.com/spf13/cobra"
 )
 
-// DeployOptions options for deploy pipeline command
-type DeployOptions struct {
+// flags represents the user input
+type flags struct {
```
teresaromero, line 38: ❤️

teresaromero (approved):

teresaromero:
I like the separation between flags/options and the data transformation is much more clearer :) 👏🏻

cmd/pipeline/deploy.go
jLopezbarb: I've also loved the way the codes organized. It looks much cleaner and helps to reduce dependencies with other packages

cmd/pipeline/deploy.go
```diff
@@ -227,11 +193,11 @@ func getPipelineName(repository string) string {
 	return model.TranslateURLToName(repository)
 }
 
-func (pc *Command) streamPipelineLogs(ctx context.Context, name, actionName string) error {
-	return pc.okClient.Pipeline().StreamLogs(ctx, name, actionName)
+func (pc *Command) streamPipelineLogs(ctx context.Context, name, namespace, actionName string) error {
```
ifbyol, line 196: @teresaromero Be aware of this change regarding your refactor of the stream logs logic

cmd/pipeline/deploy.go
```diff
@@ -331,3 +297,90 @@ func CheckAllResourcesRunning(name string, resourceStatus map[string]string) (bo
...
+		o.Branch = b
+	}
+
+	if o.Namespace == "" {
+		o.Namespace = okteto.Context().Namespace
+	}
+	return nil
+}
+
+func (o *DeployOptions) toPipelineDeployOptions() (types.PipelineDeployOptions, error) {
```
ifbyol, line 366: Can we make this more specific to reference that is for the client? I mean, I think it might be confusing to have a function called `toOptions` and another called `toPipelineDeployOptions` in a struct called `DeployOptions` in a `pipeline` package. What is the difference between both?

Maybe `toPipelineDeployClientOptions`?

cmd/destroy/destroy.go
```diff
@@ -263,11 +263,18 @@ func (dc *destroyCommand) runDestroy(ctx context.Context, opts *Options) error {
...
 
 	if opts.DestroyDependencies {
-		for depName := range manifest.Dependencies {
+		for depName, depInfo := range manifest.Dependencies {
 			oktetoLog.SetStage(fmt.Sprintf("Destroying dependency '%s'", depName))
+
+			namespace := okteto.Context().Namespace
+			if depInfo.Namespace != "" {
+				namespace = depInfo.Namespace
+			}
```
ifbyol, line 272: Why are we checking this for the destroy but not for the deploy? Is there any difference? I think it is not needed with the refactor and the logic in setDefaults, right?

ifbyol (commented):
* Can we include an e2e test to cover this scenario?

* Can you also double check that the refactor of flags and options is not leaving behind any parameter?

cmd/destroy/destroy.go
jLopezbarb: Tested and it was also needed in the deploy command. I've added it also to the deploy command

jLopezbarb:
Added e2e that has passed here https://app.circleci.com/pipelines/github/okteto/okteto/9094/workflows/95fcf90a-51ac-45b4-a326-dbf9f30d5b1e/jobs/25757

ifbyol (approved):

PR #3256, opened 2022-11-29
okteto destroy --all option with log stream

teresaromero:
Resolves [#5113](https://github.com/okteto/app/issues/5113)

Depends on #3249 

Proposed changes:
- New `--all` flag for `okteto destroy` cmd
- When this option is enabled, new flow for `okteto destroy`
- New function `DestroyAll` under `Namespaces` okteto client calling `destroyAllInSpace` mutation
- New function `DestroyAllLogs` under `streamClient` and print function to show the logs of this process

### DIFF
--- a/cmd/destroy/destroy.go
+++ b/cmd/destroy/destroy.go
@@ -537,16 +537,14 @@ func (pc *destroyCommand) waitForNamespaceDestroyAllToComplete(ctx context.Conte
 			case "Active":
 				if hasBeenDestroyingAll {
 					// when status is active again check if all resources have been correctly destroyed
-					// no configmap for the given namespace should be up and running
-					// check if there are configmaps with error state
 					cfgList, err := c.CoreV1().ConfigMaps(namespace).List(ctx, metav1.ListOptions{})
 					if err != nil {
 						return err
 					}
-					for _, cg := range cfgList.Items {
-						if cg.Data["status"] == "error" || cg.Data["status"] == "destroy-error" {
-							return fmt.Errorf("namespace destroy all failed: some resources were not destroyed")
-						}
+
+					// no configmap for the given namespace should exist
+					if len(cfgList.Items) > 0 {
+						return fmt.Errorf("namespace destroy all failed: some resources where not destroyed")
 					}
 					// exit the waiting loop when status is active again
 					return nil


pkg/okteto/stream.go
```diff
@@ -73,3 +79,35 @@ func printPipelineLog(line string) bool {
...
+	}
+	return stream.GetLogsFromURL(ctx, c.client, url.String(), printDestroyAllLog)
+}
+
+// printDestroyAllLog prints a line with the Message unmarshalled from line
+func printDestroyAllLog(line string) bool {
+	destroyAllLogList := []destroyAllLogFormat{}
+	json.Unmarshal([]byte(line), &destroyAllLogList)
+	for _, dLog := range destroyAllLogList {
+		// skip when the event log is in stage done and message is EOF
```
ifbyol, line 98:
```suggestion
		// skip when the message is `Done`
```

cmd/destroy/destroy.go
```diff
@@ -162,6 +163,41 @@ func Destroy(ctx context.Context) *cobra.Command {
...
+					}
+					exit <- err
+				}()
+				select {
+				case <-stop:
+					oktetoLog.Infof("CTRL+C received, exit")
+					return nil
+				case err := <-exit:
+					return err
+				}
```
ifbyol, line 198: I don't think this logic has to be here. As I see, the structure of this (and other commands) are:
* Options/flags validation
* Running some pre configuration (mainly load the context)
* And then, command logic (usually through an struct like `destroyCommand`

In this case, we are including part of the command logic in the place where we are validating flags.

I would keep the logic of validating if the context is an Okteto one, but the rest, I would move it to:
* Within the function `runDestroy`
* Another function in the `destroyCommand` struct and in the line `217` include a conditional like

```
if options.DestroyAll {
  err = c.runDestroyAll(ctx, options)
} else { 
  err = c.runDestroy(ctx, options)
}
```

At the end, this logic belongs to `destroyCommand`.

Also, we wouldn't be tracking the destroy as we are doing in the line 218. Should we track it in this case too in an special way?

cmd/destroy/destroy.go
```diff
@@ -162,6 +163,41 @@ func Destroy(ctx context.Context) *cobra.Command {
 				options.Namespace = okteto.Context().Namespace
 			}
 
+			// when option --all the cmd will destroy everything at the namespace and return
+			if options.DestroyAll {
+				if !okteto.Context().IsOkteto {
+					return errors.New("option `--all` is not available for non-Okteto clusters. Learn more: https://www.okteto.com/docs/self-hosted/")
```
pchico83, line 169: @rberrelleza could you review this message?

pkg/okteto/stream.go
```diff
@@ -73,3 +79,35 @@ func printPipelineLog(line string) bool {
...
 }
+
+// DestroyAllLogs retrieves logs from the pipeline provided and prints them, returns error
+func (c *streamClient) DestroyAllLogs(ctx context.Context, namespace string) error {
+	streamURL := fmt.Sprintf(destroyAllUrlTempleate, Context().Name, Context().Namespace)
+	url, err := url.Parse(streamURL)
+	if err != nil {
+		return err
+	}
+	return stream.GetLogsFromURL(ctx, c.client, url.String(), printDestroyAllLog)
```
pchico83, line 90: what happens with reconnections? could we retry and keep iterating on the events?

pchico83 (commented):
@teresaromero could you attach a screenshot of the final output? or a short video?

teresaromero:
> @teresaromero could you attach a screenshot of the final output? or a short video?

i've uploaded a loom video on how is looking with the latest commit of this PR https://www.loom.com/share/0128877bc0784889afad4206d0f27b59

cmd/destroy/destroy.go
teresaromero: I've done this change, also regarding tracking, i think this should be tracked separatly, or at least detect a different event as is a different use?

An other question that i just came up, i added includeVolumes as default true, wondering if we also want to add the flag -v and based on that?

pkg/okteto/stream.go
pchico83: @teresaromero that's great!
will events be shown twice after a reconnection? or do we manage to only show events once?

cmd/destroy/destroy.go
```diff
@@ -178,6 +186,14 @@ func Destroy(ctx context.Context) *cobra.Command {
 			}
 			os.Setenv("KUBECONFIG", kubeconfigPath)
 			defer os.Remove(kubeconfigPath)
+
+			// when option --all the cmd will destroy everything at the namespace and return
+			if options.DestroyAll {
+				if !okteto.Context().IsOkteto {
+					return errors.New("option `--all` is not available for non-Okteto clusters. Learn more: https://www.okteto.com/docs/self-hosted/")
+				}
+				return c.runDestroyAll(ctx, options)
```
ifbyol, line 195: We are losing the `TrackDestroy`, @pchico83 is it needed to track this action? Should we include some data in the tracking to indicate that the new flag was being used?

pkg/okteto/namespace.go
```diff
@@ -168,3 +168,23 @@ func (c *namespaceClient) SleepNamespace(ctx context.Context, namespace string)
...
+func (c *namespaceClient) DestroyAll(ctx context.Context, namespace string) error {
+	var mutation struct {
+		Space struct {
+			Id graphql.String
+		} `graphql:"destroyAllInSpace(id: $id, includeVolumes: $includeVolumes)"`
+	}
+	// includingVolumes so everything is cleaned up by default with this cmd
+	variables := map[string]interface{}{
+		"id":             graphql.String(namespace),
+		"includeVolumes": graphql.Boolean(true),
```
ifbyol, line 182: Why always true? `Destroy` operation already receives a boolean to indicate if the volumes has to be destroyed or not. I think we should use the value of that flag. At the end, the user already has an option to delete volumes or not when deleting all from the UI

cc @pchico83 @rlamana

pkg/okteto/stream.go
ifbyol: I would apply this to avoid a misleading code comment

ifbyol:
@teresaromero I've been testing it and I think we need a more complex change for 2 reasons:
* We rely on the stream logs endpoint/functionality to consider that the command has ended or not. Conceptually, this is wrong. The command should finish when the destroy all operation is finished, we shouldn't rely on the streamed logs.
* Even if the stream finish correctly, we are not even checking the final state of the operation, so we are reporting back that the command was executed successfully, but it is not the case. These are the screenshot of 2 different scenarios where the destroy all failed but we report back a success.

One of the pipeline which had to be explicitly destroyed failed during the deletion as you can see in the logs but the command says it was successfully

Captura de Pantalla 2022-12-05 a las 13 27 31

In this case, the generic `destroy all` failed and the namespace status is `destroyAllFailed` but the command says it was successful (I force this case in my dev environment) 

Captura de Pantalla 2022-12-05 a las 15 06 50


So, I think we should include the logic to monitor the namespace status and stop the command and return the proper response based on:
* If the namespace status goes from `DestroyingAll` to `DestroyAllFailed`, we should return an error
* If the namespace status goes from `DestroyingAll` to `Active`, we should list all configmaps belonging to a dev environment and check if there is any in a `error` or `destroy-error` status. If there is any, the command should return an error. If there isn't any, the command should return a success

pkg/okteto/namespace.go
teresaromero: i thought that "all" meant all... so making this default to true will also remove the volumes 🤷🏻‍♀️  
WDYT? no problem in changing it

cmd/destroy/destroy.go
teresaromero: is we keep the same event for both destroys we wont see the difference of using `--all` flag, i think we either add more meta to this event or we create a new event for destroyall. WDYT?

pkg/okteto/namespace.go
pchico83: agree with @ifbyol. `all` means all the repos and services. we already have `-v` to destroy volumes

cmd/destroy/destroy.go
pchico83: let's add an additional field to the `destroy` event. It's what we do for other commands

pkg/okteto/namespace.go
teresaromero: added this value at the mutation

ifbyol:
> > @teresaromero I've been testing it and I think we need a more complex change for 2 reasons:
> > 
> > * We rely on the stream logs endpoint/functionality to consider that the command has ended or not. Conceptually, this is wrong. The command should finish when the destroy all operation is finished, we shouldn't rely on the streamed logs.
> > * Even if the stream finish correctly, we are not even checking the final state of the operation, so we are reporting back that the command was executed successfully, but it is not the case. These are the screenshot of 2 different scenarios where the destroy all failed but we report back a success.
> > 
> > One of the pipeline which had to be explicitly destroyed failed during the deletion as you can see in the logs but the command says it was successfully
> > Captura de Pantalla 2022-12-05 a las 13 27 31
> > In this case, the generic `destroy all` failed and the namespace status is `destroyAllFailed` but the command says it was successful (I force this case in my dev environment)
> > Captura de Pantalla 2022-12-05 a las 15 06 50
> > So, I think we should include the logic to monitor the namespace status and stop the command and return the proper response based on:
> > 
> > * If the namespace status goes from `DestroyingAll` to `DestroyAllFailed`, we should return an error
> > * If the namespace status goes from `DestroyingAll` to `Active`, we should list all configmaps belonging to a dev environment and check if there is any in a `error` or `destroy-error` status. If there is any, the command should return an error. If there isn't any, the command should return a success
> 
> I was looking into it, and as i see without making any changes at the other components i see a solution on calling namespaces list, and check the status of the deleted namespace.
> 
> Other option could be to modify the mutation so it returns the id of the job, and call to see the progress of this?

I would go with checking the namespace status using k8s API directly or even using a watcher in the namespace. Even when the namespace comes back to `Active`, we should check if something is still deployed there or not as it might have failed in a pipeline destroy, in which case, the status goes back to `Active`

cmd/destroy/destroy.go
```diff
@@ -435,3 +457,106 @@ func getTempKubeConfigFile(name string) string {
...
+
+			status, ok := ns.Labels["space.okteto.com/status"]
+			if !ok {
+				return errors.New("namespace does not have label for status")
+			}
+
+			switch status {
+			case "Active":
+				if hasBeenDestroyingAll {
+					// when status is active again check if all resources have been correctly distroyed
```
ifbyol, line 539:
```suggestion
					// when status is active again check if all resources have been correctly destroyed
```

cmd/destroy/destroy.go
```diff
@@ -435,3 +457,106 @@ func getTempKubeConfigFile(name string) string {
...
+					// check if there are configmaps with error state
+					cfgList, err := c.CoreV1().ConfigMaps(namespace).List(ctx, metav1.ListOptions{})
+					if err != nil {
+						return err
+					}
+					for _, cg := range cfgList.Items {
+						if cg.Data["status"] == "error" || cg.Data["status"] == "destroy-error" {
+							return fmt.Errorf("namespace destroy all failed: some resources where not destroyed")
+						}
+					}
```
ifbyol, line 550: Probably checking that `len(cfgList.Items) > 0` is enough (and better). As the destroy all operation should be intended to delete everything, if a configmap is still there, it means that something failed

cmd/destroy/destroy.go
```diff
@@ -435,3 +457,106 @@ func getTempKubeConfigFile(name string) string {
...
+					// when status is active again check if all resources have been correctly distroyed
+					// no configmap for the given namespace should be up and running
+					// check if there are configmaps with error state
+					cfgList, err := c.CoreV1().ConfigMaps(namespace).List(ctx, metav1.ListOptions{})
+					if err != nil {
+						return err
+					}
+					for _, cg := range cfgList.Items {
+						if cg.Data["status"] == "error" || cg.Data["status"] == "destroy-error" {
+							return fmt.Errorf("namespace destroy all failed: some resources where not destroyed")
```
ifbyol, line 548:
```suggestion
							return fmt.Errorf("namespace destroy all failed: some resources were not destroyed")
```

pkg/okteto/stream.go
teresaromero: @pchico83  regarding reconnection, my mistake, what is implemented at the request for subscription is a retry for the initial connection, as this was a leading point to solve cloud issue...

a reconnection scenario where the connection is lost would require more changes. The logs come though the body stream, so we would need to detect if the scan of the body has stoped because a connection failure. the logs are pure text so there is no id or order we can identify though the stream... i think a reconnection strategy would need more refinement, as while the process continues, the reconnection time could lead us to miss some of the logs.

cc: @ifbyol

ifbyol (approved):
It looks good minor comments to take a look but approved.

I was testing the branch and I think the streaming of logs could be a bit misleading. Check how it look like

Captura de Pantalla 2022-12-15 a las 19 33 36

I think we might print the streamed log in a different color or with some indentation. WDYT @pchico83 @rlamana ? If we want to improve something, I wouldn't block the PR for this, this is already an improvement but maybe we can create a ticket or a follow up PR

pkg/okteto/stream.go
ifbyol: Yeah, I wouldn't block this PR for that reconnection. As tere says, we don't have any kind of id, so it would require some extra work or refinement

teresaromero:
> I think we might print the streamed log in a different color or with some indentation. WDYT @pchico83 @rlamana ? If we want to improve something, I wouldn't block the PR for this, this is already an improvement but maybe we can create a ticket or a follow up PR

Formatting this text is done showing the log line with oktetoLog.Println. this was already used when streaming pipeline logs. changing this format would imply more changes, i think we should not block the PR because of this and  if format change is wanted open an issue to follow that enhacement.

cmd/destroy/destroy.go
teresaromero: > we should list all configmaps belonging to a dev environment and check if there is any in a error or destroy-error status. If there is any, the command should return an error. If there isn't any, the command should return a success

i understood that `if there is any` was regarding the configmaps errors. you meant any configmap then?

ifbyol: No, at that time I was thinking just on "any configmap on that error state". But I mean, If there is any config map (regardless the state), we didn't delete all, right? 

Anyways, we can leave it as it is and change it in the future if we detect any corner case

rlamana:
Agree with @ifbyol, actually in the output he shows I'm unable to distinguish anything, can you show rhe parts that would use indentation?

pkg/okteto/stream.go
pchico83: @teresaromero could you create an issue in the product backlog with these details? that would help us to prioritize it as needed

cmd/destroy/destroy.go
```diff
@@ -178,12 +188,23 @@ func Destroy(ctx context.Context) *cobra.Command {
...
-			if err == nil {
-				oktetoLog.Success("Development environment '%s' successfully destroyed", options.Name)
-			}
 
+			// when option --all the cmd will destroy everything at the namespace and return
+			if options.DestroyAll {
+				if !okteto.Context().IsOkteto {
+					return errors.New("option `--all` is not available for non-Okteto clusters. Learn more: https://www.okteto.com/docs/self-hosted/")
+				}
+				err = c.runDestroyAll(ctx, options)
```
jLopezbarb, line 197: Should we track the destroy here?

ifbyol: It is being tracked in line 207, right? We are adding a new option/flag `DestroyAll`

jLopezbarb: I though it was in the else branch 😅

jLopezbarb:
Tested and code is working!! Please check deepsource before merging it but looks very cool

jLopezbarb (approved):

cmd/destroy/destroy.go
teresaromero: i think is more accurate that if the cfmap are still there, the job is not success, i will change it

cmd/destroy/destroy.go
```diff
@@ -178,12 +188,23 @@ func Destroy(ctx context.Context) *cobra.Command {
...
-			err = c.runDestroy(ctx, options)
-			analytics.TrackDestroy(err == nil)
-			if err == nil {
-				oktetoLog.Success("Development environment '%s' successfully destroyed", options.Name)
-			}
 
+			// when option --all the cmd will destroy everything at the namespace and return
+			if options.DestroyAll {
+				if !okteto.Context().IsOkteto {
+					return errors.New("option `--all` is not available for non-Okteto clusters. Learn more: https://www.okteto.com/docs/self-hosted/")
```
teresaromero, line 195: could we check here the copy of this message if is the correct one? 

cc: @ifbyol @pchico83 @rberrelleza @rlamana

pchico83: @teresaromero instead of failing, could we ignore the logic to destroy the pipelines in this case?

teresaromero: @pchico83 you mean at the destroy-all job that is triggered?

teresaromero:
> 



> Agree with @ifbyol, actually in the output he shows I'm unable to distinguish anything, can you show rhe parts that would use indentation?

@rlamana https://github.com/okteto/app/issues/5113 at the last comment of the issue you can find a demo video where the output is shown. 

i would need to know if there is something needed with format right now that can be done or we block this until there is a format solution.

The video shows the 2 affected "streams" which are right now pipeline destroy, pipeline deploy and destroy --all in this PR

rlamana:
@teresaromero I think this is fine. We might need to rethink how we are showing this in general, but following the same mechanism we already have in other parts, lgtm 👌

rlamana (approved):

PR #3257, opened 2022-11-29
Fix divert target port hwen target service is diverted

pchico83:

This solves a regression introduced by https://github.com/okteto/okteto/pull/3198

The problem is that when creating the headless service for an ingress, we use the `port` of the service for the endpoint. If the services get's diverted, the port is updated by weaver to be 1024. We should use the target port for the headless service, not the `port`


jmacelroy (approved):

pchico83:
Closing for now, I still need validation from users of the feature

PR #3259, opened 2022-12-01
Jlopezbarb/fix last updated not being updated

jLopezbarb:
# Proposed changes

- Add a comment on why we use a different client for configmap
- Add the `lastUpdated` annotation for the configmap created on `okteto deploy`

### DIFF
--- a/cmd/deploy/command_configuration_test.go
+++ b/cmd/deploy/command_configuration_test.go
@@ -96,7 +96,7 @@ devs:
 	assert.Equal(t, expectedCfg.Namespace, currentCfg.Namespace)
 	assert.Equal(t, expectedCfg.Labels, currentCfg.Labels)
 	assert.Equal(t, expectedCfg.Data, currentCfg.Data)
-	assert.NotNil(t, currentCfg.Annotations[model.LastUpdatedAnnotation])
+	assert.NotEmpty(t, currentCfg.Annotations[model.LastUpdatedAnnotation])
 }
 
 func Test_mergeServicesToDeployFromOptionsAndManifest(t *testing.T) {
--- a/cmd/deploy/command_configuration_test.go
+++ b/cmd/deploy/command_configuration_test.go
@@ -92,7 +92,11 @@ devs:
 		t.Fatal("error trying to get configmap from data object")
 	}
 
-	assert.Equal(t, expectedCfg, currentCfg)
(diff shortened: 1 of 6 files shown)


teresaromero (approved):

cmd/deploy/command_configuration_test.go
```diff
@@ -92,7 +92,11 @@ devs:
 		t.Fatal("error trying to get configmap from data object")
 	}
 
-	assert.Equal(t, expectedCfg, currentCfg)
+	assert.Equal(t, expectedCfg.Name, currentCfg.Name)
+	assert.Equal(t, expectedCfg.Namespace, currentCfg.Namespace)
+	assert.Equal(t, expectedCfg.Labels, currentCfg.Labels)
+	assert.Equal(t, expectedCfg.Data, currentCfg.Data)
+	assert.NotNil(t, currentCfg.Annotations[model.LastUpdatedAnnotation])
```
ifbyol, line 99: If I'm not wrong, `currentCfg.Annotations[model.LastUpdatedAnnotation]` cannot be nil as Annotations is a `map[string]string` and the default value would be an empty string. So this assert would always be true even if the annotation is not present.

There would be 2 ways of checking this:
* `_, ok := currentCfg.Annotations[model.LastUpdatedAnnotation]` and check that `ok` is true
* `value := currentCfg.Annotations[model.LastUpdatedAnnotation]` and verify that `val != ""`

Same about the tests in `deploy_test.go` and `translate_test.go`

pkg/cmd/pipeline/translate_test.go
```diff
@@ -91,6 +93,9 @@ func Test_AddDevAnnotations(t *testing.T) {
 	manifest := &model.Manifest{
 		Namespace: "namespace",
 		Dev: model.ManifestDevs{
+			"not-found-deployment": &model.Dev{
+				Name: "not-found-deployment",
+			},
```
ifbyol, line 98: Just curiosity, Why this change?

ifbyol (approved):
Code looks good but take a look to the comment I added in the tests before merging please!

pkg/cmd/pipeline/translate_test.go
jLopezbarb: That tests an untested scenario which checks what happens if the deployment is no longer running. We should skip the adding annotation function for that resource

PR #3261, opened 2022-12-01
Fix inline export when image tag and cache tag are the same

pchico83:

# Proposed changes

Fix the `type` parameter based on the cache tag. If cache tag is the one we are pushing, `inline` must be used.
More info [here](https://github.com/moby/buildkit#inline-push-image-and-cache-together)

### DIFF
--- a/pkg/cmd/build/buildkit.go
+++ b/pkg/cmd/build/buildkit.go
@@ -131,17 +131,17 @@ func getSolveOpt(buildOptions *types.BuildOptions) (*client.SolveOpt, error) {
 	}
 
 	if buildOptions.ExportCache != "" {
-		mode := "min"
+		exportType := "inline"
 		if buildOptions.ExportCache != buildOptions.Tag {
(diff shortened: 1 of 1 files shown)


jLopezbarb (approved):

pkg/cmd/build/buildkit.go
```diff
@@ -131,17 +131,17 @@ func getSolveOpt(buildOptions *types.BuildOptions) (*client.SolveOpt, error) {
 	}
 
 	if buildOptions.ExportCache != "" {
-		mode := "min"
+		exportType := "inline"
```
jmacelroy, line 134: There should be a test added for this. It looks like we should be asserting specific build options being set as a default when export cache is an empty string?

What's the best way to follow-up on that?

pchico83: @jLopezbarb is going to send a follow up PR with an e2e test too

PR #3263, opened 2022-12-02
Fix sanitize name for stacks

teresaromero:
# Proposed changes

Fixes https://github.com/okteto/app/issues/4980

As the name option is now stored as original, couple breaking workflows were spotted.
- Stack name sanitization: fix the compose name when is inferred from repo or folder name
- Helm release name sanitization: fix helm release name when not valid
- Temp kubeconfig name sanitization: use the sanitization version of name for this temp file

Added couple e2e cases to check the sanitization is working as expected


integration/okteto/deploy_test.go
```diff
@@ -145,6 +145,73 @@ func TestDeploySuccessOutput(t *testing.T) {
...
+func TestDeployWithNameSuccessOutput(t *testing.T) {
+	integration.SkipIfNotOktetoCluster(t)
+	t.Parallel()
+	oktetoPath, err := integration.GetOktetoPath()
+	require.NoError(t, err)
+
+	dir := t.TempDir()
+	require.NoError(t, createComposeScenario(dir))
+
+	testNamespace := integration.GetTestNamespace("TestSuccessOutput", user)
```
ifbyol, line 157:
```suggestion
	testNamespace := integration.GetTestNamespace("TestDeployWithNonSanitizedNameSuccess", user)
```

integration/okteto/deploy_test.go
```diff
@@ -145,6 +145,73 @@ func TestDeploySuccessOutput(t *testing.T) {
...
+	require.NoError(t, commands.RunOktetoCreateNamespace(oktetoPath, namespaceOpts))
+	require.NoError(t, commands.RunOktetoKubeconfig(oktetoPath, dir))
+	defer commands.RunOktetoDeleteNamespace(oktetoPath, namespaceOpts)
+
+	deployOptions := &commands.DeployOptions{
+		Workdir:    dir,
+		Namespace:  testNamespace,
+		OktetoHome: dir,
+		Token:      token,
+		Name:       "my deployment",
```
ifbyol, line 172: Nit: Can we insert a '/' somewhere? This is a special character which gives us headaches from time to time

integration/okteto/deploy_test.go
```diff
@@ -145,6 +145,73 @@ func TestDeploySuccessOutput(t *testing.T) {
...
+		if _, ok := stageLines[ss]; !ok {
+			t.Fatalf("deploy didn't have the stage '%s'", ss)
+		}
+		if strings.HasPrefix(ss, "Building service") {
+			if len(stageLines[ss]) < 5 {
+				t.Fatalf("Not sending build output on stage %s. Output:%s", ss, stageLines[ss])
+			}
+		}
+
+	}
```
ifbyol, line 212: The main goal of this test should be to test that the dev environment is deployed successfully, so I would not include all these logic and checks. It might be misleading. The focus of this test should be that a dev environment with a non-k8s compliance name is able to deploy successfully

integration/okteto/deploy_test.go
```diff
@@ -145,6 +145,73 @@ func TestDeploySuccessOutput(t *testing.T) {
 	}
 }
 
+func TestDeployWithNameSuccessOutput(t *testing.T) {
```
ifbyol, line 148:
```suggestion
func TestDeployWithNonSanitizedNameSuccess(t *testing.T) {
```

As mentioned in other comment, I think we should remove the output checks. There should be other tests for it

pkg/cmd/stack/deploy.go
```diff
@@ -187,14 +187,14 @@ func deploy(ctx context.Context, s *model.Stack, c kubernetes.Interface, config
...
-				endpoint.Labels[model.StackNameLabel] = s.Name
+				endpoint.Labels[model.StackNameLabel] = format.ResourceK8sMetaString(s.Name)
 			}
 			if _, ok := endpoint.Labels[model.StackEndpointNameLabel]; !ok {
 				endpoint.Labels[model.StackEndpointNameLabel] = endpointName
 			}
 
 			translateOptions := &ingresses.TranslateOptions{
-				Name:      s.Name,
+				Name:      format.ResourceK8sMetaString(s.Name),
```
ifbyol, line 197: In line 391 of this file we are also creating the same type (`ingresses.TranslateOptions`) but we are not sanitizing the name, I guess we should do it in both places

pkg/cmd/stack/translate_test.go
```diff
@@ -33,21 +33,21 @@ import (
 func Test_translateConfigMap(t *testing.T) {
 	s := &model.Stack{
 		Manifest: []byte("manifest"),
-		Name:     "stackName",
+		Name:     "stack Name",
```
ifbyol, line 36: 👏

ifbyol (approved):
Added some comments but in general looks good. Can we run the e2e in this branch before merging? Also, @jLopezbarb can you review it too, please?

integration/okteto/deploy_test.go
ifbyol: yeah, that makes sense

teresaromero:
Found a bug while testing the fix, in the case the name gets inferred from the repository url. When deploying from cli any repo or pipeline, the repo url extracted was being sanitised, i've done the changes so the name remains "original" extracted from the url and added couple comments on the changes to bear in mind that now name is the original one, and the sanitized version is used when using this name on resources.

plse review this latest changes let me know.

cc: @ifbyol @jLopezbarb

ifbyol:
I think there is one case we need to handle. In case of autodiscovery for Helm chart, we are using the value of the env var `OKTETO_NAME` like this `command := fmt.Sprintf("helm upgrade --install ${%s} %s %s", constants.OktetoNameEnvVar, chartPath, tags)`.

As we are storing the non-sanitized name in that env var, I think that would fail. Can we test it?

teresaromero:
> I think there is one case we need to handle. In case of autodiscovery for Helm chart, we are using the value of the env var `OKTETO_NAME` like this `command := fmt.Sprintf("helm upgrade --install ${%s} %s %s", constants.OktetoNameEnvVar, chartPath, tags)`.
> 
> As we are storing the non-sanitized name in that env var, I think that would fail. Can we test it?

It fails in the case `OKTETO_NAME` is not sanitized, for example "Movies with helm". A solution for this would be adding a new env `OKTETO_SANITIZED_NAME` and use this when parsing the cmd. I dont see a way of sanitizing the value of the env while running this cmd.

ifbyol:
> > I think there is one case we need to handle. In case of autodiscovery for Helm chart, we are using the value of the env var `OKTETO_NAME` like this `command := fmt.Sprintf("helm upgrade --install ${%s} %s %s", constants.OktetoNameEnvVar, chartPath, tags)`.
> > As we are storing the non-sanitized name in that env var, I think that would fail. Can we test it?
> 
> It fails in the case `OKTETO_NAME` is not sanitized, for example "Movies with helm". A solution for this would be adding a new env `OKTETO_SANITIZED_NAME` and use this when parsing the cmd. I dont see a way of sanitizing the value of the env while running this cmd.
> 
> This was already happening when `--name` was not a "valid" string, now that we keep this value as original we have to check that the value on release name is valid

Talked offline, we are adding a new environment variable with the sanitized name instead of using `OKTETO_NAME`. We've been reviewing different usages of the environment variables and the `InferName` function and we are sanitizing the name in all the cases where is needed

pkg/model/utils.go
```diff
@@ -36,15 +37,11 @@ func GetValidNameFromFolder(folder string) (string, error) {
 	}
 
 	name := filepath.Base(dir)
-	name = strings.ToLower(name)
```
ifbyol, line 39: Info for other reviewers. We are removing this as now the name of the dev environment is the original one and we sanitize it when needed (as config map name, label values, etc...)

ifbyol (approved):
Code looks good, it would be great if another person can do a review

jLopezbarb (approved):

PR #3264, opened 2022-12-02
Fix infer name for "okteto logs"

pchico83:

# Proposed changes

`okteto logs` is not inferring the name of the manifest properly. I think this logic should be part of the manifest deserialization, but that would require a larger refactor

### DIFF
--- a/cmd/logs/logs.go
+++ b/cmd/logs/logs.go
@@ -61,6 +61,13 @@ func Logs(ctx context.Context) *cobra.Command {
 			if err != nil {
 				return err
 			}
+			wd, err := os.Getwd()
+			if err != nil {
+				return err
+			}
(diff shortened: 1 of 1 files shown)


ifbyol (approved):

teresaromero (approved):

jLopezbarb (approved):

cmd/logs/logs.go
```diff
@@ -61,6 +61,13 @@ func Logs(ctx context.Context) *cobra.Command {
 			if err != nil {
 				return err
 			}
+			wd, err := os.Getwd()
+			if err != nil {
+				return err
+			}
+			if manifest.Name == "" {
```
rberrelleza, line 68: Can we add a test on this scenario to prevent regressions?

PR #3270, opened 2022-12-02
test: add e2e test regarding export cache

jLopezbarb:

# Proposed changes

Follow up of #3261 

- Add tests for #3261

### DIFF
--- a/integration/deploy/manifest_test.go
+++ b/integration/deploy/manifest_test.go
@@ -43,12 +43,69 @@ var (
 deploy:
   - kubectl apply -f k8s.yml
 `
+	oktetoManifestContentWithCache = `build:
+  app:
+    context: app
+	export-image: okteto.dev/app:dev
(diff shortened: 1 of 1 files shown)


integration/deploy/manifest_test.go
```diff
@@ -42,13 +42,70 @@ var (
     context: app
 deploy:
   - kubectl apply -f k8s.yml
+`
+	oktetoManifestContentWithCache = `build:
+  app:
+    context: app
+	export-image: okteto.dev/app:dev
```
pchico83, line 49: this should have the same indentation than `context`, no?

pchico83 (approved):

integration/deploy/manifest_test.go
jLopezbarb: Yes, I used tabs instead of whitespaces and that's why it didn't have the same indentation but for Vscode it was

PR #3278, opened 2022-12-05
Adding error handling

### LINKED ISSUE #3220: Need better error handling
**Describe the bug**
Need error handling here:
https://github.com/okteto/okteto/blob/39b5b7757668a2b4bea7b3e0054c3b174f010dd9/internal/test/utils.go#L36

**Additional context**
Need to check if `SetEnv()` func returns an error. Possibly correct it through updating linters.
Context: https://kubernetes.slack.com/archives/CM1QMQGS0/p1667276346405729

abs007:
# Proposed changes

Fixes #3220 

This pr adds the `errcheck` tool for linting purposes and also takes care of all the error handling needed.


cmd/deploy/deploy.go
```diff
@@ -150,7 +150,9 @@ func Deploy(ctx context.Context) *cobra.Command {
 					if err != nil {
 						return err
 					}
-					nsCmd.Create(ctx, &namespace.CreateOptions{Namespace: okteto.Context().Namespace})
+					if err := nsCmd.Create(ctx, &namespace.CreateOptions{Namespace: okteto.Context().Namespace}); err != nil {
```
jLopezbarb, line 153: We should be returning an error here, we will fail later on because the ns doesn't exist

cmd/namespace/use.go
```diff
@@ -155,7 +155,9 @@ func getNamespacesSelection(ctx context.Context) ([]utils.SelectorItem, error) {
 
 func askForOktetoNamespace() string {
 	var namespace string
-	oktetoLog.Question("Enter the namespace you want to use: ")
+	if err := oktetoLog.Question("Enter the namespace you want to use: "); err != nil {
```
jLopezbarb, line 158: Same with this one. If we can't throw a question we should return the proper error

cmd/stack/deploy.go
```diff
@@ -99,7 +99,9 @@ func (c *DeployCommand) RunDeploy(ctx context.Context, s *model.Stack, options *
 			if err != nil {
 				return err
 			}
-			nsCmd.Create(ctx, &namespace.CreateOptions{Namespace: s.Namespace})
+			if err := nsCmd.Create(ctx, &namespace.CreateOptions{Namespace: s.Namespace}); err != nil {
```
jLopezbarb, line 102: We should be returning an error here, we will fail later on because the ns doesn't exist

cmd/utils/dev.go
```diff
@@ -236,9 +236,10 @@ const (
 func AskYesNo(q string, d YesNoDefault) (bool, error) {
 	var answer string
 	for {
-		oktetoLog.Question(fmt.Sprintf("%s %s: ", q, d))
-		_, err := fmt.Scanln(&answer)
-		if err != nil && err.Error() != "unexpected newline" {
+		if err := oktetoLog.Question(fmt.Sprintf("%s %s: ", q, d)); err != nil {
```
jLopezbarb, line 239: Same with this one. If we can't throw a question we should return the proper error

cmd/namespace/use.go
abs007: Should I go for `oktetoLog.Errorf()` here since the func doesn't return an error?

abs007:
Should I go ahead and add err handling here too? 
https://github.com/okteto/okteto/blob/master/pkg/cmd/build/build_docker.go#L174

cmd/namespace/use.go
jLopezbarb: Let's modify the function signature to `func askForOktetoNamespace() (string, error)`

jLopezbarb:
> Should I go ahead and add err handling here too?
> https://github.com/okteto/okteto/blob/master/pkg/cmd/build/build_docker.go#L174

Yes, please

abs007:
Using OktetoLog here causes an `import cycle` error.
https://github.com/okteto/okteto/blob/master/pkg/log/json.go#L297

Should I instead return `(0, err)`

abs007:
Not tackling this one atm since we're changing the `AddArgs()` method in #3215
https://github.com/okteto/okteto/blob/master/cmd/up/up.go#L370

jLopezbarb (approved):

cmd/build/v2/services_test.go
```diff
@@ -27,7 +28,9 @@ func TestAllServicesAlreadyBuilt(t *testing.T) {
 	fakeReg := test.NewFakeOktetoRegistry(nil)
 	bc := NewBuilder(nil, fakeReg)
 	alreadyBuilt := []string{}
-	fakeReg.AddImageByName(alreadyBuilt...)
+	if err := fakeReg.AddImageByName(alreadyBuilt...); err != nil {
+		oktetoLog.Infof("error adding image to registry: %s", err)
+	}
```
AgustinRamiroDiaz, line 33: shouldn't we fail the test here? We can user `require.NoErr`

abs007: Sounds good. Should I use the require package for err handling in all the other test files too?

AgustinRamiroDiaz: @abs007 Yes please :pray:

abs007: Will do then

jLopezbarb:
@abs007 could you update the PR resolving conflicts?

abs007:
@jLopezbarb Sure. Is it alright if I get to it day after tomorrow? My exams are ending on the 10th.

jLopezbarb:
No worries @abs007. I wish you the best of luck in your exam 💪

abs007:
Thank you :)

jLopezbarb:
Unit test are failing. Could you review it @abs007 

@AgustinRamiroDiaz, since you reviewed and asked for changes, could you verify that everything is ready

cmd/up/activate.go
```diff
@@ -103,8 +103,11 @@ func (up *upContext) activate() error {
 		}
 	}
 
-	go up.initializeSyncthing()
-
+	go func() {
+		if err := up.initializeSyncthing(); err != nil {
+			oktetoLog.Infof("could not initialize syncthing: %s", err)
```
AgustinRamiroDiaz, line 108: @jLopezbarb should we add some mechanism to stop the command if this fails? 

@abs007 don't mind about this, it's out of scope for this PR

AgustinRamiroDiaz:
> It seems like build_test.go is having issues with the require package in lines 375 and 120. Should I use okteto log there instead @AgustinRamiroDiaz

@abs007 That looks like a problem in the test :eyes: . We should try to fix the test in those cases.

For future reference, okteto log is what we use to display logs to the user when running the commands. We shouldn't be using it in the tests

cmd/up/activate.go
jLopezbarb: It's just launching the syncthing process, that we have previously downloaded to make it compatible with the user OS, so this shouldn't fail in any circunstances.

Otherwise, there is a mechanism to wait for 1min and if syncthing doesn't start we show a proper error to the user here: https://github.com/okteto/okteto/pull/3278/files/c0736885136289a5a7130639c9bb63633abdac0c#diff-c39b965c9f4376229cb1d945d1dcddaca021d99664fe2c5674aee1c278999600R155

AgustinRamiroDiaz:
@abs007 After some headscratches I've found how to fix the issue in build_test.go line 375:

The first part is very similar to what you've done, except that you need to add `df := df` at the beginning of the for loop. This is due to how golang for loops work. More on that [here](https://github.com/golang/go/discussions/56010)
```go
			if tt.dockerfilesCreated != nil {
				for _, df := range tt.dockerfilesCreated {
					df := df
					dfFile, err := os.Create(df)
					require.NoError(t, err)

					log.Printf("created docker file: %s", df)

					require.NoError(t, dfFile.Close())

					defer func() {
						require.NoError(t, removeFile(df))
					}()
				}
			}
```

And then at the end of the test you should remove
```go
			if tt.dockerfilesCreated != nil {
				for _, df := range tt.dockerfilesCreated {
					removeFile(df)
				}
			}
```

If you think that this is too much, you can fix the rest and leave this to us :D

abs007:
@AgustinRamiroDiaz 

So, in the example that they mention:

While using `range`, the value (`item` in this case) just gets its value overwritten with the next val in `items`. We keep using the same `item` variable for storing the next val and hence `&item` remains the same?

And, having `item := item` extracts that specific iteration value and stores it explicitly?

AgustinRamiroDiaz:
@abs007 exactly. The `item := item` is the creation of a new local variable scoped inside the for loop. I used the same name so it gets shadowed and we don't need to think of a new name, but it could've been any name like `scopedItem` (and then we'd use that name afterwards)

abs007:
Got it. Also, we're removing the code at the end because it's duplicated?

AgustinRamiroDiaz:
> Got it. Also, we're removing the code at the end because it's duplicated?
@abs007 Exactly!

> I'm also removing all the t.Fatal() calls in the code with require.NoError(t, err). Hoping that's alright?
@abs007 Totally! That's the way we want to go with unit testing :test_tube: 

Once you are done with your changes, remember to fix the merge conflicts (`git pull origin master`)

.golangci.yml
```diff
@@ -3,11 +3,11 @@ run:
   go: 1.18
 linters:
```
jLopezbarb, line 4: After reviewing the options added by `errcheck`, I think we should be also add the following setting.
@abs007 could you add this new section?

```suggestion
# All possible options can be found here https://github.com/golangci/golangci-lint/blob/master/.golangci.reference.yml
linters-settings:
  errcheck:
    # Report about not checking of errors in type assertions: `a := b.(MyStruct)`.
    # Such cases aren't reported by default.
    # Default: false
    check-type-assertions: true
linters:
```

abs007:
Fyi, I'm working on the merge conflicts.

cmd/context/context.go
```diff
@@ -58,6 +59,8 @@ This will prompt you to select one of your existing contexts or to create a new
 	cmd.Flags().StringVarP(&ctxOptions.Namespace, "namespace", "n", "", "namespace of your okteto context")
 	cmd.Flags().StringVarP(&ctxOptions.Builder, "builder", "b", "", "url of the builder service")
 	cmd.Flags().BoolVarP(&ctxOptions.OnlyOkteto, "okteto", "", false, "only shows okteto cluster options")
-	cmd.Flags().MarkHidden("okteto")
+	if err := cmd.Flags().MarkHidden("okteto"); err != nil {
+		oktetoLog.Infof("failed to mark 'okteto' flag as hidden: %s", err)
```
AgustinRamiroDiaz, line 63: @jLopezbarb should this be a `Debug` log?

pkg/cmd/build/build_docker.go
```diff
@@ -171,12 +171,16 @@ func buildWithDockerDaemon(ctx context.Context, buildOptions *types.BuildOptions
...
+			oktetoLog.Infof("failed to set error: %s", err)
+		}
 	}
 
 	dockerBuildOptions, err := getDockerOptions(buildOptions)
 	if err != nil {
-		errors.Wrap(err, "setting build options failed")
+		if err := errors.Wrap(err, "setting build options failed"); err != nil {
+			oktetoLog.Infof("failed to set error: %s", err)
+		}
```
AgustinRamiroDiaz, line 183: @jLopezbarb these 2 errors were not being handled before, they were just being wrapped without returning them. 

Should we return them now?

pkg/cmd/stack/deploy.go
```diff
@@ -545,7 +545,9 @@ func isAnyPortAvailable(ctx context.Context, svc *model.Service, stack *model.St
 			continue
 		}
 	}
-	forwarder.Start(podName, stack.Namespace)
+	if err := forwarder.Start(podName, stack.Namespace); err != nil {
+		oktetoLog.Infof("could not start port-forward: %s", err)
```
AgustinRamiroDiaz, line 549: @jLopezbarb should we `continue` here as we do with the other errors?

pkg/k8s/namespaces/namespace.go
```diff
@@ -327,7 +327,9 @@ func (t *Trip) Wander(ctx context.Context, traveler Traveler) error {
 }
 
 func (t *Trip) listAll(ctx context.Context, traveler Traveler, client dynamic.ResourceInterface, api *metav1.APIResourceList, api2 metav1.APIResource) error {
-	t.sem.Acquire(ctx, 1)
+	if err := t.sem.Acquire(ctx, 1); err != nil {
+		oktetoLog.Infof("failed to acquire semaphore: %v", err)
+	}
```
AgustinRamiroDiaz, line 332: I think we should return the error here

cmd/build/v2/services_test.go
AgustinRamiroDiaz: Since it's a test, in this whole file we should `require.NoError` instead of logging the message

internal/test/build.go
```diff
@@ -43,7 +44,9 @@ func (fb *FakeOktetoBuilder) Run(_ context.Context, opts *types.BuildOptions) er
 	}
 
 	if opts.Tag != "" {
-		fb.Registry.AddImageByOpts(opts)
+		if err := fb.Registry.AddImageByOpts(opts); err != nil {
+			oktetoLog.Infof("error adding image to registry: %s", err)
+		}
```
AgustinRamiroDiaz, line 49: Since this is for a test, we should return the error here

AgustinRamiroDiaz (commented):
Awesome improvements to the codebase :muscle:

pkg/cmd/down/run.go
```diff
@@ -43,7 +43,9 @@ func Run(dev *model.Dev, app apps.App, trMap map[string]*apps.Translation, wait
 				return err
 			}
 		} else {
-			tr.DevModeOff()
+			if err := tr.DevModeOff(); err != nil {
+				oktetoLog.Infof("failed to turn devmode off: %s", err)
+			}
```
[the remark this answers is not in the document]
jLopezbarb: No, We should just log the error. It's a non blocking error

cmd/context/context.go
jLopezbarb: I'd leave it as Info, debug is not stored in log file which makes it harder to detect

pkg/cmd/stack/deploy.go
jLopezbarb: Yes, this is just a healthcheck while deploying an stack. We can't fail for a healtcheck and wait until the backofflimit is reached

abs007:
Thanks @AgustinRamiroDiaz 😊. Was fun solving those merge conflicts too.

pkg/cmd/stack/deploy.go
AgustinRamiroDiaz:
```suggestion
		oktetoLog.Infof("could not start port-forward: %s", err)
		continue
```

@abs007 :eyes:

AgustinRamiroDiaz (approved):
Great work :100:

abs007:
@AgustinRamiroDiaz  But isn't that line out of the for loop?

AgustinRamiroDiaz:
> @AgustinRamiroDiaz But isn't that line out of the for loop?

Oh you are right :thinking: Maybe we should return `false` in that case, but nevermind! The logging is already a great improvement :rocket: 

Thanks a lot for your contribution :1st_place_medal:

PR #3292, opened 2022-12-12
Add parameter for pass thru of allowPrivilegeEscalation (#3291)

### LINKED ISSUE #3291: securityContext in okteto.yaml does not support allowPrivilegeEscalation
**Describe the bug**
The securityContext for a container defined in the okteto.yaml files does not support setting the allowPrivilegeEscalation parameter.  In some kubernetes environments that must run in a Restricted security posture this parameter is required to be false.  It can only be set on the container and cannot be set at the pod level.

**To Reproduce**
Create a kubernetes environment where a namespace runs in the Pod Security Standard of "Restricted" and try to run okteto on any running project.  It will fail to start the 2 init containers that need to start.  Several error will be generated.  The only one that cannot be fixed by setting a pod security policy is  allowPrivilegeEscalation.  

**Expected behavior**
okteto should be able to run in a kubernetes environment with the Pod Security Standard of "Restricted".  

N/A affects all versions and all platforms.  This is based on the kubernetes environment not the client environment.

**Additional context**
I will put in a pull request for a fix for this.  It's simple enough to pass through the allowPivilegeEscalation parameter or not set it at all.

skikd636:
# Proposed changes

Fixes #3291

Added the ability to allow the parameter allowPrivilegeEscalation in the securityContext of the okteto.yaml file.

### DIFF
--- a/pkg/k8s/apps/translate.go
+++ b/pkg/k8s/apps/translate.go
@@ -453,6 +453,10 @@ func TranslateContainerSecurityContext(c *apiv1.Container, s *model.SecurityCont
 	if s.RunAsNonRoot != nil {
 		c.SecurityContext.RunAsNonRoot = s.RunAsNonRoot
 	}
+    
+	if s.AllowPrivilegeEscalation != nil {
+		c.SecurityContext.AllowPrivilegeEscalation = s.AllowPrivilegeEscalation
+	}    
 
 	if s.Capabilities == nil {
 		return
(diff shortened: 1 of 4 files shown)


pkg/k8s/apps/translate.go
```diff
@@ -453,6 +453,10 @@ func TranslateContainerSecurityContext(c *apiv1.Container, s *model.SecurityCont
 	if s.RunAsNonRoot != nil {
 		c.SecurityContext.RunAsNonRoot = s.RunAsNonRoot
 	}
+    
+	if s.AllowPrivilegeEscalation != nil {
```
jLopezbarb, line 457: Can we add a unit test of this?

jLopezbarb (approved):

AgustinRamiroDiaz (approved):

PR #3297, opened 2022-12-14
update file permission for bin image

### LINKED ISSUE #3294: In some cases, `okteto up` command fails and the dev container shows a permission issue
**Describe the bug**
A customer has reported that when they execute `okteto up`, they get this error:

```
❯ okteto up
 i  Using xxxxx @ xxxxxxx as context
 i  'xxx-xxxxx' was already deployed. To redeploy run 'okteto deploy' or 'okteto up --deploy'
 ✓  Images successfully pulled
 x  Failed to connect to your development container
    Check your development container logs for errors: 'kubectl logs xxxxxxxx',
        Check that your container can write to the destination path of your secrets.
        Run 'okteto down -v' to reset your development container and try again
```

If they go to the dev container, they see this log

```
2-13T12:34:45 Copying secret authorized_keys to /var/okteto/remote/authorized_keys
2022-12-13 12:34:45.18 UTCxxxxxx cp: cannot open '/var/okteto/secret/authorized_keys' for reading: Permission denied
```

It was working before but it stopped working suddenly.

**To Reproduce**
After some research, we have being able to reproduce the issue:

1. Create a dockerfile similar to this one

```
FROM alpine
USER 1001
```

2. Build the image and push it as `okteto.dev/hello:latest`
3. Create a manifest similar to this one (it is important to disable the persistent volume):

```
name: failing-one
image: okteto.dev/hola:latest
command: bash
autocreate: true
persistentVolume:
  enabled: false
```

4. Execute `okteto up`
5. The command should fail and if you check the logs in the dev container, you should see the logs complaining about permissions.

**Expected behavior**

There shouldn't be any permission issue as Okteto is the one controlling those paths and volumes

- Version: `2.10.1`

**Additional context**

It seems that the culprit of this issue was this change done a couple of months ago: https://github.com/okteto/okteto/pull/3095/commits/9857a624dc2f5bb4789cba3aba26c19ef7da7bcb.

A workaround for this would be to downgrade the CLI to `2.7.2`

AdrianPedriza:

# Proposed changes
- update file permissions

Fixes #3294

Permissions were too restrictive, so they were changed to laxer ones.

### DIFF
--- a/cmd/up/stignore.go
+++ b/cmd/up/stignore.go
@@ -104,7 +104,7 @@ func addStignoreSecrets(dev *model.Dev) error {
 			model.Secret{
 				LocalPath:  transformedStignorePath,
 				RemotePath: path.Join(folder.RemotePath, ".stignore"),
-				Mode:       0600,
+				Mode:       0644,
 			},
 		)
 	}
--- a/integration/up/autocreate_test.go
+++ b/integration/up/autocreate_test.go
@@ -86,6 +86,23 @@ dev:
     - 8012:8080
     autocreate: true
 `
+
+	autocreateManifestV2WithDefinedUser = `
+name: test
+image: okteto.dev/test:latest
+command: echo done
+autocreate: true
+persistentVolume:
+  enabled: false
+sync:
+- .:/app
+`
+
+	dockerfileDefinedUser = `
+FROM alpine
+USER 1001
+COPY --chown=1001:1001 . /app
+`
 )
 
 func TestUpAutocreate(t *testing.T) {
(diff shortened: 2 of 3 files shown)


cmd/up/stignore.go
```diff
@@ -104,7 +104,7 @@ func addStignoreSecrets(dev *model.Dev) error {
 			model.Secret{
 				LocalPath:  transformedStignorePath,
 				RemotePath: path.Join(folder.RemotePath, ".stignore"),
-				Mode:       0600,
+				Mode:       0644,
```
ifbyol, line 107: Do we need to change also the permissions of these two? https://github.com/okteto/okteto/blob/326d8b417a74580e1dfe2cbbc3360a5c46e15f0b/cmd/up/stignore.go#L173 and https://github.com/okteto/okteto/blob/326d8b417a74580e1dfe2cbbc3360a5c46e15f0b/cmd/up/stignore.go#L211

jLopezbarb: no, those files are used as the secret so having them locally with the least permissions, the best

AdrianPedriza: I have tried to be as restrictive as possible resolving the issue. None of these filemodes were involved

ifbyol: 👍

jLopezbarb (approved):

jLopezbarb:
Tested and couldn't find any case where its failing!!

integration/up/autocreate_test.go
```diff
@@ -85,6 +85,23 @@ dev:
...
+persistentVolume:
+  enabled: false
+sync:
+- .:/app
+`
+
+	dockerfileDefinedUser = `
+FROM alpine
+USER 1001
+COPY --chown=1001:1001 . /app
```
ifbyol, line 104: There is some magic in the test. With build v1, if the image is not present, the image is automatically built. Would it make sense to change to use a manifest v2? In that case, the manifest would have the build section and would make this much easier to understand.

Anyways, really good test to catch it and leave it like that if you want or if you think it would take much more time. Just one thing, if we leave it with manifest v1, change the name of the variable `autocreateManifestV2WithDefinedUser` to `autocreateManifestV1WithDefinedUser`

ifbyol (approved):

integration/up/autocreate_test.go
AdrianPedriza: Yes, you are right. I have added the same test for v2 of the manifest. I keep the test for v1 for proximity to the scenario that has caused us problems before

ifbyol: I didn't want you to build the test for the 2 versions but yeah, agree it is better! Thanks!!

ifbyol (approved):

PR #3298, opened 2022-12-15
add external resource parsing

### LINKED ISSUE #3266: Add external resource definition in the CLI
As a developer I want to be able to define external resources in my okteto manifest, so I can use that information in the UI to have a full view of my application.

The idea is to have a new key at the root of okteto manifest which defines the external resource that will be shown in the UI.

The definition of this new section is:
```golang
type Manifest struct {
  ... 
  ExternalResources map[string]ExternalResourceInfo
}

type externalResourceInfoUnmarshall struct {
  notes string // path from the root of the project
  endpoints []ExternalEndpoints
}

type externalResourceInfo struct {
  notes ExternalResourceNotes
  endpoints []ExternalEndpoints
}

type ExternalResourceNotes struct {
  path string
  markdown string // base64 encoded content of the path
}

type externalResourceEndpoints struct {
  name string // endpoint name
  url string // URL to display in the UI/endpoints cmd
}
``` 

## Acceptance criteria

- [x] Define the external resource structs
- [x] It should be unmarshalled correctly
  - [x] notes must not be empty
  - [x] there must be at least one endpoint
- [ ] It should show a warning and skip it if the user is not in an okteto cluster. 
  - [ ] The warning must be shown only if the external key is present 
- [x] The environment variable `$OKTETO_EXTERNAL_($NAME)_ENDPOINTS_($NAME)_URL` must be available in the deploy section
- [x] Differentiate between the unmarshall struct and the one being used in the rest of the command
- [x] Store the content of `external.$NAME.notes` in base64

AdrianPedriza:

# Proposed changes
- add reading of an external resource declaration in a manifest
- add method for creating required environment variables
- add method to load markdown content and store it in base64
- add test covering features

Fixes #3266

# Notes: 
the warning if not in an okteto cluster will be added once the integration with the commands is started.


pkg/externalresource/serializer.go
```diff
@@ -0,0 +1,50 @@
+package externalresource
+
+import (
+	"fmt"
+)
+
+type externalResourceRaw struct {
```
jLopezbarb, line 7:
```suggestion
type externalResourceUnmarshaller struct {
```

pkg/externalresource/external_resource.go
```diff
@@ -0,0 +1,58 @@
...
+	b64 "encoding/base64"
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/spf13/afero"
+)
+
+// ExternalResources represents the map of external resources at a manifest
+type ExternalResources map[string]*ExternalResource
```
jLopezbarb, line 13: Let's use a name that can differentiate more from the actual type

```suggestion
type ExternalResourceSection map[string]*ExternalResource
```

pkg/externalresource/external_resource.go
```diff
@@ -0,0 +1,58 @@
...
+// ERFilesystemManager represents ExternalResource information with the filesystem inyected
+type ERFilesystemManager struct {
+	ExternalResource ExternalResource
+	Fs               afero.Fs
+}
+
+// SetDefaults creates the necessary environment variables given an external resource
+func (er *ExternalResource) SetDefaults(externalName string) {
+	for _, endpoint := range er.Endpoints {
+		endpointUrlEnv := fmt.Sprintf("OKTETO_EXTERNAL_%s_ENDPOINTS_%s_URL", externalName, endpoint.Name)
```
jLopezbarb, line 43: We would probably need to sanitize `endpoint.Name`. We need to uppercase the name and replace `-` with `_`

pkg/externalresource/external_resource.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+// ERFilesystemManager represents ExternalResource information with the filesystem inyected
+type ERFilesystemManager struct {
+	ExternalResource ExternalResource
+	Fs               afero.Fs
+}
+
+// SetDefaults creates the necessary environment variables given an external resource
+func (er *ExternalResource) SetDefaults(externalName string) {
```
AgustinRamiroDiaz, line 42: This is only used in the test. I don't think it should be exported, and maybe it should be in the test itself

pkg/externalresource/external_resource.go
```diff
@@ -0,0 +1,65 @@
...
+		os.Setenv(endpointUrlEnv, endpoint.Url)
+	}
+}
+
+func sanitizeForEnv(name string) string {
+	return strings.ToUpper(strings.ReplaceAll(name, "-", "_"))
+}
+
+// LoadMarkdownContent loads and store markdown content related to external resource
+func (ef *ERFilesystemManager) LoadMarkdownContent(manifestPath string) error {
```
AgustinRamiroDiaz, line 56: This is only used in the test. Why do we have these functions?

AdrianPedriza: yes at the moment these functions are not used at any place but both will be used by the commands. It is due to the way the epic work has been split up

pkg/externalresource/external_resource.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+// ExternalEndpoint represents information about an endpoint
+type ExternalEndpoint struct {
+	Name string
+	Url  string
+}
+
+// ERFilesystemManager represents ExternalResource information with the filesystem inyected
+type ERFilesystemManager struct {
```
jLopezbarb, line 36: This is the only thing that annoys me... We create an struct to modify one of their fields. I think it can be better if we return externalresource modified. WDYT?

pkg/externalresource/external_resource.go
```diff
@@ -0,0 +1,65 @@
...
+	b64 "encoding/base64"
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/spf13/afero"
+)
+
+// ExternalResources represents the map of external resources at a manifest
```
teresaromero, line 13:
```suggestion
// ExternalResourcesSection represents the map of external resources at a manifest
```

pkg/externalresource/external_resource.go
```diff
@@ -0,0 +1,65 @@
...
+func (er *ExternalResource) SetDefaults(externalName string) {
+	sanitizedExternalName := sanitizeForEnv(externalName)
+	for _, endpoint := range er.Endpoints {
+		sanitizedEndpointName := sanitizeForEnv(endpoint.Name)
+		endpointUrlEnv := fmt.Sprintf("OKTETO_EXTERNAL_%s_ENDPOINTS_%s_URL", sanitizedExternalName, sanitizedEndpointName)
+		os.Setenv(endpointUrlEnv, endpoint.Url)
+	}
+}
+
+func sanitizeForEnv(name string) string {
```
teresaromero, line 51: what about spaces and special characters? does this name need to be part of any kubernetes resource?

pkg/externalresource/serializer.go
```diff
@@ -0,0 +1,50 @@
...
+	for _, entry := range result.Endpoints {
+		if _, isAdded := uniqueEndpointsNames[entry.Name]; isAdded {
+			return fmt.Errorf("there must be no duplicate names for the endpoints of an external resource")
+		}
+
+		uniqueEndpointsNames[entry.Name] = false
+	}
+
+	er.Icon = result.Icon
+	er.Notes.Path = result.Notes
```
teresaromero, line 43: for the moment only the Notes.Path is unmarshalled here? could we add a note TODO for the Markdown prop of the Notes? or where this would be done?

pkg/externalresource/external_resource.go
ifbyol: Shouldn't/can't be part of the manifest deserialization? I mean, if we are reading all the `ExternalResource` from the manifest definition, wouldn't it make sense to read the markdown there? If not, where would we be reading it?

pkg/externalresource/external_resource.go
AdrianPedriza: this sanitization is to incorporate the name of the external and its endpoints as an environment variable. But yes, it would be necessary to avoid blank spaces

pkg/externalresource/external_resource.go
AdrianPedriza: I need to know somehow the path of the manifest to build the path to the markdown and that inside the deserialization process I can't know it I think

pkg/externalresource/serializer.go
AdrianPedriza: in principle I cannot add the markdown content because I need to know the path to the manifest

pkg/externalresource/external_resource.go
AdrianPedriza: In my opinion I think that the name of the function helps not to create confusion, it simply loads the content of the markdown. Receiving as parameter a struct and returning it modified or creating a new one I believe that leaves us in a similar situation and even you are left with a more verbose function declaration. But that's my opinion, I have no strong opinion for or against it

jLopezbarb: What if is `ExternalResource` the one that has the afero dependency? That way we don't have to create another an intermediate struct. `externalResource.LoadMarkdownContent(path string)`. 

We could even not add the afero dependency as a field and initialise it inside the `externalResource.LoadMarkdownContent`.

AdrianPedriza: in most of the use of this struct this property would not be used, I believe that it is something foreign to the information that must store the `ExternalResource` struct. If we instantiate it inside the function we can not use the injection of this variable for its testing

jLopezbarb (approved):

pkg/externalresource/external_resource.go
ifbyol: oh, ok, I see. Makes sense then to have it outside the parsing

pkg/model/dev.go
```diff
@@ -892,7 +892,7 @@ func (dev *Dev) LoadRemote(pubKeyPath string) {
 	p := Secret{
 		LocalPath:  pubKeyPath,
 		RemotePath: authorizedKeysPath,
-		Mode:       0600,
+		Mode:       0644,
```
ifbyol, line 895: Why this change? Wasn't it already on master?

ifbyol (approved):
Just check the comment about the change related to permissions. I though that was already merged. Besides that, everything looks fine!

pkg/model/dev.go
AdrianPedriza: :hushed: some failure updating with master. I will check it. Thanks Nacho!

PR #3300, opened 2022-12-16
Check deployed dev environments when inferring the name

### LINKED ISSUE #3276: `okteto up` and `okteto logs` don't get the right dev environment name to be used
**Describe the bug**

When a custom name is specified while deploying a dev environment, if you execute `okteto up` or `okteto logs`, the name is not inferred properly as those commands don't have a way to specify the dev environment name and they might not know how was the custom name. 

In the case of `okteto up`, as the dev environment is not found with the inferred name, a new dev environment is deployed with the inferred name and it might cause some failures. (some k8s resources might have collisions, in the UI appears an empty dev environment, etc...)

In the case of `okteto logs`, logs cannot be streamed as the right resources are not found because the inferred name doesn't match with the deployed one

**To Reproduce**

1. Clone https://github.com/okteto/go-getting-started
2. From the root folder execute `okteto deploy --name my-fancy-go-getting-started
3. Once the dev environment is deployed, check the UI and you'll see it. Then, from the repo root folder, execute `okteto up`
4. You will see how a new dev environment with a different name is deployed. If you go to the UI, you will see something like this

Captura de Pantalla 2022-12-03 a las 12 19 05

**Expected behavior**

If the name is not specified in the manifest, we should look for possible dev environments in which their repository and manifest path matches with the ones detected in the commands. If several are found, we should display a selector, so the user could select the right one

**Additional context**
We have to bear in mind that there some place where the manifest is autodetected by the commands. In those cases, the filename should be empty in the dev environment

ifbyol:
Fixes #3276 

# Proposed changes

Modified the logic of the function `InferName` to use the following preference for inferring the name:

1. Name specified with the `--name` flag
2. Name specified in the manifest
3. Name specified in a deployed dev environment matching the repository URL + manifest path (the new scenario we are adding)
4. Name extracted from the repo URL of the working directory
5. Name extracted from the folder name of the working directory

The first 2 cases are handled directly by any commands and if no name is found, a call to the function `InferName` is being done for the rest of scenarios.

In this PR I'm adding the scenario number 3 within the `InferName` function.

Apart from that, I've moved the `InferName` function to a new package.

Finally, I added also a flag `--name` to the logs command

Some concerns I have (cc @pchico83): 

1. It might happen that the same repository with the same manifest path is deployed with different name. What should we do in this case? For now, I'm just taking the first one
2. In the way we handle the namespace involved in each command is different and it is hard to know the one we should use. All my testing and checking the code points that using `okteto.Context().Namespace` should be ok but there are cases which is very confusing. For example, [here](https://github.com/okteto/okteto/blob/ifbyol/3276-improve-infer-name-logic/cmd/deploy/endpoints.go#L108-L114) it might happen that `options.Namespace` has the `manfest.Namespace` value, but has `okteto.Context().Namespace` the same value?

In any case, I'm always defaulting to previous behavior, so in the worts case scenario there would be still cases in which this keeps behaving in the way the ticket reports. 

I think that we should unify how the namespace and those things are initialized in the commands.

Just a note to bear in mind, the first commit contains the refactor to move the `InferName`. You can check it [here](https://github.com/okteto/okteto/commit/d95d8c19d042cd65e120171410dec5f162275fd9).


pkg/devenvironment/name.go
```diff
@@ -0,0 +1,57 @@
...
+	"path/filepath"
+
+	"github.com/okteto/okteto/pkg/k8s/configmaps"
+	oktetoLog "github.com/okteto/okteto/pkg/log"
+	"github.com/okteto/okteto/pkg/model"
+	"github.com/okteto/okteto/pkg/okteto"
+	"k8s.io/client-go/kubernetes"
+)
+
+// InferName infers the application name from the folder received as parameter
```
teresaromero, line 15:
```suggestion
// InferName infers the dev environment name from the folder received as parameter
```

should this be regarding dev environment although it is an application?

pkg/devenvironment/name.go
```diff
@@ -0,0 +1,57 @@
...
+		return model.TranslateURLToName(repoURL)
+	}
+
+	oktetoLog.Information("found %d configmaps in the namespace %s", len(cfList), namespace)
+
+	for _, cmap := range cfList {
+		oktetoLog.Information("checking configmap %s", cmap.Name)
+		repo := cmap.Data["repository"]
+		if repo == "" {
+			oktetoLog.Information("configmap %s doesn't have a repo", cmap.Name)
```
teresaromero, line 38:
```suggestion
			oktetoLog.Information("configmap %s doesn't have repository", cmap.Name)
```

teresaromero: nit: the property is repository so i think is better to put the property that was not found

cmd/deploy/endpoints.go
```diff
@@ -97,7 +98,12 @@ func Endpoints(ctx context.Context) *cobra.Command {
 				if manifest.Name != "" {
 					options.Name = manifest.Name
 				} else {
-					options.Name = utils.InferName(cwd)
+					c, _, err := okteto.NewK8sClientProvider().Provide(okteto.Context().Cfg)
+					if err != nil {
+						return err
+					}
+					// TODO: Check manifest. It changes, probably need to keep the one specified by the user
```
teresaromero, line 105: i am not sure about this comment. if namespace is checked after... should be first check manifest namespace so options.Namespace has the one at the manifest? and then use this insead of okteto.Context().Namespace?

pchico83:
@ifbyol about this one:
```
It might happen that the same repository with the same manifest path is deployed with different name. What should we do in this case? For now, I'm just taking the first one
```
Let's pick the first one and show an info message

ifbyol:
After talking offline with @pchico83 , I'm leaving our of the scope of this PR the modification of `getStackName` function. It requires a bigger refactor. It affects when the repository has only a `docker-compose.yaml` or `okteto-stack.yaml`. In those cases, we won't infer the name from the dev environments deployed in the namespace. Most of the commands have the `--name` flag to cover the case.

cmd/utils/manifest.go
```diff
@@ -1,33 +0,0 @@
...
-
-import (
-	"path/filepath"
-
-	oktetoLog "github.com/okteto/okteto/pkg/log"
-	"github.com/okteto/okteto/pkg/model"
-)
-
-// InferName infers the application name from the folder received as parameter
-func InferName(cwd string) string {
```
ifbyol, line 24: Moved to a new package in this commit https://github.com/okteto/okteto/pull/3300/commits/d95d8c19d042cd65e120171410dec5f162275fd9

pkg/devenvironment/name.go
```diff
@@ -0,0 +1,103 @@
...
+	"github.com/okteto/okteto/pkg/k8s/configmaps"
+	oktetoLog "github.com/okteto/okteto/pkg/log"
+	"github.com/okteto/okteto/pkg/model"
+	"github.com/okteto/okteto/pkg/okteto"
+	"k8s.io/client-go/kubernetes"
+)
+
+// DeprecatedInferName infers the dev environment name from the folder received as parameter.
+// It is deprecated as it doesn't take into account deployed dev environments to get the non-sanitized name.
+// This is only being effectively used in push command, which will be deleted in the mext major version
```
teresaromero, line 18:
```suggestion
// This is only being effectively used in push command, which will be deleted in the next major version
```

teresaromero (approved):
LGTM. Nice comments and refactor of infer name, this will add more clarity :)

jLopezbarb (approved):

cmd/deploy/endpoints.go
ifbyol: After checking the logic, it seems ok as it is now, so I deleted the comment

ifbyol:
@teresaromero @jLopezbarb can you double check the lat commit? I had to add some of the logic in the pipeline deploy and destroy commands as they are not using the `InferName` function

ifbyol:
> > After talking offline with @pchico83 , I'm leaving our of the scope of this PR the modification of `getStackName` function. It requires a bigger refactor. It affects when the repository has only a `docker-compose.yaml` or `okteto-stack.yaml`. In those cases, we won't infer the name from the dev environments deployed in the namespace. Most of the commands have the `--name` flag to cover the case.
> 
> Code looks good👏👏 !! Can we create an issue to tackle this in the future, or at least taking it in consideration whenever we refactor/tweak those files

@jLopezbarb Ticket created! https://github.com/okteto/okteto/issues/3311

teresaromero (approved):

jLopezbarb (approved):

PR #3301, opened 2022-12-20
Okteto namespace delete use destroy all job

teresaromero:
# Proposed changes

Fixes https://github.com/okteto/app/issues/5177

Proposed changes:
- include force param as false to the delete space mutation call
- add to the namespace delete cmd the logs streamed by the destroy-all job triggered
- check namespace status while deleting, to check is not DeleteFailed

### DIFF
--- a/cmd/namespace/delete.go
+++ b/cmd/namespace/delete.go
@@ -15,7 +15,12 @@ package namespace
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"os"
+	"os/signal"
+	"sync"
+	"time"
 
 	contextCMD "github.com/okteto/okteto/cmd/context"
 	"github.com/okteto/okteto/cmd/utils"
@@ -24,6 +29,7 @@ import (
 	oktetoLog "github.com/okteto/okteto/pkg/log"
 	"github.com/okteto/okteto/pkg/okteto"
 	"github.com/spf13/cobra"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 // Delete deletes a namespace
@@ -55,9 +61,17 @@ func Delete(ctx context.Context) *cobra.Command {
 }
 
 func (nc *NamespaceCommand) ExecuteDeleteNamespace(ctx context.Context, namespace string) error {
+	oktetoLog.Spinner(fmt.Sprintf("Deleting %s namespace", namespace))
+	oktetoLog.StartSpinner()
+	defer oktetoLog.StopSpinner()
 
+	// trigger namespace deletion
 	if err := nc.okClient.Namespaces().Delete(ctx, namespace); err != nil {
-		return fmt.Errorf("failed to delete namespace: %s", err)
+		return fmt.Errorf("failed to delete namespace: %v", err)
+	}
+
+	if err := nc.watchDelete(ctx, namespace); err != nil {
+		return err
 	}
 
 	oktetoLog.Success("Namespace '%s' deleted", namespace)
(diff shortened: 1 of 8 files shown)


cmd/namespace/delete.go
```diff
@@ -76,3 +90,71 @@ func (nc *NamespaceCommand) ExecuteDeleteNamespace(ctx context.Context, namespac
...
+	to := time.NewTicker(timeout)
+
+	for {
+		select {
+		case <-to.C:
+			return fmt.Errorf("'%s' deploy didn't finish after %s", namespace, timeout.String())
+		case <-ticker.C:
+			ns, err := nc.k8sClient.CoreV1().Namespaces().Get(ctx, namespace, metav1.GetOptions{})
+			if err != nil {
+				// when err is NotFound return without error, as the namespace is correctly deleted
```
teresaromero, line 147: here i have doubts. when the namespace is deleted i get the following error :

`namespaces "XXXXX" is forbidden: User "system:serviceaccount:okteto:XXXXXXXXX" cannot get resource "namespaces" in API group "" in the namespace "XXXX"`

i was expecting not found 🤷🏻‍♀️

ifbyol: We can handle `k8serrors.IsNotFound` and `k8serrors.IsForbidden` as a success and include a comment indicating that the error returned by kubernetes is a forbidden in case the service account don't have access to all ns

cmd/namespace/delete.go
```diff
@@ -76,3 +90,71 @@ func (nc *NamespaceCommand) ExecuteDeleteNamespace(ctx context.Context, namespac
...
+				return nil
+			}
+
+			status, ok := ns.Labels["space.okteto.com/status"]
+			if !ok {
+				return errors.New("namespace does not have label for status")
+			}
+			if status == "DeleteFailed" {
+				return errors.New("namespace destroy all failed")
```
ifbyol, line 156: Which is the error displayed to the user in these scenarios?

teresaromero: i've updated this, when error in status Delete Failed `var errFailedDeleteNamespace = errors.New("failed to delete namespace")` i use this one, i am thinking that the one on the labels should probably be same? if there is no label we can not delete...

cmd/namespace/delete.go
```diff
@@ -76,3 +90,74 @@ func (nc *NamespaceCommand) ExecuteDeleteNamespace(ctx context.Context, namespac
...
+					// when err is NotFound return without error, as the namespace is correctly deleted
+					return nil
+				}
+				return err
+			}
+
+			status, ok := ns.Labels["space.okteto.com/status"]
+			if !ok {
+				return errNoStatusLabel
+			}
```
ifbyol, line 157: In this case I wouldn't return an error. If the namespace is still present, it means that it wasn't deleted. It might be because the operation failed and it should be in `DeleteFailed` but in that case it should have the label, so I would keep polling for the namespace until timeout and I wouldn't consider the fact of not having label as something special.

cmd/namespace/delete.go
```diff
@@ -76,3 +90,74 @@ func (nc *NamespaceCommand) ExecuteDeleteNamespace(ctx context.Context, namespac
...
+	to := time.NewTicker(timeout)
+
+	for {
+		select {
+		case <-to.C:
+			return fmt.Errorf("%w: namespace %s, time %s", errDeleteNamespaceTimeout, namespace, timeout.String())
+		case <-ticker.C:
+			ns, err := nc.k8sClient.CoreV1().Namespaces().Get(ctx, namespace, metav1.GetOptions{})
+			if err != nil {
+				if k8sErrors.IsNotFound(err) || k8sErrors.IsForbidden(err) {
```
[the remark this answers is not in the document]
teresaromero: is in line 148 👀

pkg/errors/errors.go
```diff
@@ -105,7 +105,7 @@ var (
 	ErrDivertNotSupported = fmt.Errorf("the 'divert' field is only supported in clusters that have Okteto installed")
 
 	// ContextIsNotOktetoCluster raised if the cluster connected is not managed by okteto
-	ErrContextIsNotOktetoCluster = fmt.Errorf("this command is only available on Okteto Cloud or Okteto Enterprise")
+	ErrContextIsNotOktetoCluster = fmt.Errorf("this command is only available in clusters where Okteto is installed.\n Follow this link to know more about configuring Okteto at your cluster: https://www.okteto.com/docs/self-hosted/")
```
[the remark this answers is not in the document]
rlamana: It think it's ok, it should be updated for all. A less cloud-based message makes sense with our current approach. @pchico83 can you also confirm?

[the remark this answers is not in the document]
teresaromero: I saw the error was "already there", so instead of creating an error with the same "message" i completed with the URL and rephrase it. for `okteto destroy --all` i created a fmt.Errorf, but i think this should be applied also.

pkg/okteto/namespace.go
```diff
@@ -121,10 +121,12 @@ func (c *namespaceClient) Delete(ctx context.Context, namespace string) error {
 	var mutation struct {
 		Space struct {
 			Id graphql.String
-		} `graphql:"deleteSpace(id: $id)"`
+		} `graphql:"deleteSpace(id: $id, force: $force)"`
 	}
 	variables := map[string]interface{}{
 		"id": graphql.String(namespace),
+		// force is disabled to delete namespace using destroy-all-namespace job
```
teresaromero, line 128: @ifbyol i was having doubts here, should we pass the value or should i prepare a param to be pass though from the command?

ifbyol: Good question, we didn't consider to add the `--force` flag in the CLI. Leave it as it is for now to not block the PR. We can add a ticket later to include.

@rlamana Should we consider to add `--force` flag to this command? It wasn't considered orginally but maybe it make sense to create a ticket to include it at some point (without blocking this PR)

pkg/errors/errors.go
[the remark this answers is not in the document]
ifbyol: yeah, just wanted to make sure we are ok on changing the message in any place where it is being used. Makes sense for me

pkg/okteto/namespace.go
rlamana: It makes sense. There shouldn't be core operations like this that can be done from the UI, but not from the CLI. Create the ticket to tackle it.

cmd/namespace/delete.go
```diff
@@ -76,3 +90,77 @@ func (nc *NamespaceCommand) ExecuteDeleteNamespace(ctx context.Context, namespac
...
+				// one or the other we assume the namespace has been deleted
+				if k8sErrors.IsNotFound(err) || k8sErrors.IsForbidden(err) {
+					return nil
+				}
+				return err
+			}
+
+			status, ok := ns.Labels["space.okteto.com/status"]
+			if !ok {
+				// when status label is not present, continue polling the namespace until timeout
```
ifbyol, line 158: Sorry for the back and forth Tere, could we include a `oktetoLog.Debugf` to comment that the namespace doesn't have a label? In case we have reports about the command taking time, we could hve that log to see if it could be a problem

ifbyol (approved):
Minor comment about including a debug message in a case but the code and logic look good!

cmd/namespace/delete.go
ifbyol: Talked offline to modify also `waitForNamespaceDestroyAllToComplete` to behave in the same way (add a debug message and `continue`)

ifbyol (approved):

PR #3304, opened 2022-12-20
Fix endpoints name in compose

teresaromero:

The bug appears when trying to deploy a compose with endpoints and custom name. The name used for the endpoints is not being sanitised.

Captura de Pantalla 2022-12-20 a las 16 07 43

This does not appear when the endpoints are declared at the okteto manifest deploy endpoints section, this was already being sanitised.
Also, when endpoints are created from the service port mapping, the name used is the service name, which is already sanitised.

How to repro:
- clone https://github.com/okteto/node-compose-getting-started
- okteto deploy --name "Use any custom name"
- an error appears when deploying the endpoints at the compose

# Proposed changes

- added comment for tracing the name sanitisation at okteto manifest deploy endpoints section
- sanitise endpoints name when stack serialisation

### DIFF
--- a/integration/deploy/endpoints_test.go
+++ b/integration/deploy/endpoints_test.go
@@ -425,3 +425,101 @@ func Test_EndpointsFromStackAndManifest(t *testing.T) {
 	}
 	require.NoError(t, commands.RunOktetoDestroy(oktetoPath, destroyOptions))
 }
+
+// Test_EndpointsFromOktetoManifest_NameOption tests the following scenario:
+// - Deploying a okteto manifest locally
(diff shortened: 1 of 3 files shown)


ifbyol:
As we talked offline, maybe there we need the sanitization but to be completely safe, I would include it [here](https://github.com/okteto/okteto/blob/master/pkg/k8s/ingresses/translate.go#L34) and [here](https://github.com/okteto/okteto/blob/master/pkg/k8s/ingresses/translate.go#L56). In that way, we can make sure that the ingress is created with a good name in k8s

ifbyol (approved):
Code and tests looks good! Pending to see the e2e results before merging, but approving the review

PR #3308, opened 2022-12-22
Solves okteto/app#5180. Destroy preview calls Okteto API to destroy the preview and stream logs

ifbyol:

# Proposed changes

Fixes https://github.com/okteto/app/issues/5180

`okteto preview destroy` command stream now logs from the destroy-all job triggered when the `destroyPreview` is called. It also watches until the operation is done.

As part of this PR, I also remove the `force` parameter in the call to `destroySpace` mutation as it might fail for Okteto instances that don't have that code yet. As the default value is false, it is not needed to send it. We'll include it when we include the `--force` option to the CLI for these commands.

In the associated ticket I added some videos on how it looks like


pkg/okteto/namespace.go
```diff
@@ -121,12 +121,10 @@ func (c *namespaceClient) Delete(ctx context.Context, namespace string) error {
 	var mutation struct {
 		Space struct {
 			Id graphql.String
-		} `graphql:"deleteSpace(id: $id, force: $force)"`
```
ifbyol, line 124: @teresaromero I've removed this as I realize that it could fail if the Okteto instance is not updated. As the default value is false, we can omit it until we include the `--force` parameter in these commands

teresaromero: can we leave a TODO comment ?

teresaromero (approved):

pkg/okteto/namespace.go
ifbyol: I created a ticket https://github.com/okteto/app/issues/5340 to keep track of it but I wouldn't add a TODO comment. It wasn't a requirement in the feature document and it is an optional parameter, so it is something we might be leave like that forever if we don't see value in adding that option in the CLI.

teresaromero: https://github.com/okteto/okteto/pull/3312

PR #3317, opened 2022-12-26
Fixing bug and adding test cases

### LINKED ISSUE #2967: Possible bug caught by golangci-lint
In the file [pkg/model/volumes.go](https://github.com/okteto/okteto/pull/2952/files#diff-f07c4ae0ba01563473de575ae41c2a4bd47b09d98b129f456cd0a2d870cdd167), in the funciton `getSourceSubPath` there's a branch that asigns the `path` variable and that's not used. 

The function has unit tests, which could be 2 things:
- tests are ok and test the desired behavior => branch should be removed
- tests miss the behavior => branch should return the path 

Additional context: https://github.com/okteto/okteto/pull/2952/files#r930695403

abs007:
# Proposed changes

Fixes #2967 

This pr simplifies the code in the `getSourceSubPath` method in `volumes.go` and also adds test cases.

### DIFF
--- a/pkg/model/volumes.go
+++ b/pkg/model/volumes.go
@@ -139,14 +139,13 @@ func (dev *Dev) getSourceSubPath(path string) string {
 		dev.parentSyncFolder = "."
 	}
 	rel, err := filepath.Rel(dev.parentSyncFolder, filepath.ToSlash(sourceSubPath))
+	if err != nil {
+		oktetoLog.Debugf("error on getSourceSubPath of '%s': %s", path, err.Error())
+	}
 	if filepath.IsAbs(sourceSubPath) {
-		if err != nil || strings.HasPrefix(rel, "..") {
-			if err != nil {
-				oktetoLog.Debugf("error on getSourceSubPath of '%s': %s", path, err.Error())
-			}
+		if strings.HasPrefix(rel, "..") {
 			if filepath.IsAbs(path) {
 				oktetoLog.Info("could not retrieve subpath")
-				path = filepath.Base(path)
 			} else {
 				p, err := filepath.Abs(path)
 				if err != nil {


abs007:
Since the `"windows-current-folder"` test fails, should I scrap this particular test case or remove the entire set (the last 3 which have `dev.parentSyncFolder = ""` )

pkg/model/volumes.go
```diff
@@ -139,14 +139,13 @@ func (dev *Dev) getSourceSubPath(path string) string {
...
 	rel, err := filepath.Rel(dev.parentSyncFolder, filepath.ToSlash(sourceSubPath))
+	if err != nil {
+		oktetoLog.Debugf("error on getSourceSubPath of '%s': %s", path, err.Error())
+	}
 	if filepath.IsAbs(sourceSubPath) {
-		if err != nil || strings.HasPrefix(rel, "..") {
-			if err != nil {
-				oktetoLog.Debugf("error on getSourceSubPath of '%s': %s", path, err.Error())
-			}
+		if strings.HasPrefix(rel, "..") {
```
ifbyol, line 146: My understanding seeing the ticket is that we should just remove line `149` but the change in this if is changing the current behavior. 

Prior this change, we were executing the logic within this if there was an error calculating the relative path in line `141`. Now, we will never enter here if there is an error, as we are checking the result of that operation. Not sure if we want that or not cc @jLopezbarb

jLopezbarb: > My understanding seeing the ticket is that we should just remove line 149 but the change in this if is changing the current behavior.

I thought the same while reading the ticket. I think that removing just the extra line is the safest move

> Prior this change, we were executing the logic within this if there was an error calculating the relative path in line 141. Now, we will never enter here if there is an error, as we are checking the result of that operation. Not sure if we want that or not cc @jLopezbarb

 We need to ensure that this logic is only executed when there is an error, this code is called when we retry a connection, and if executed twice it will modify the path and will return an error later in the code

jLopezbarb (approved):

jLopezbarb:
@abs007 Windows tests are failing, could you review them before merging it? Thanks in advance

abs007:
>Since the "windows-current-folder" test fails, should I scrap this particular test case or remove the entire set (the last 3 which have dev.parentSyncFolder = "" )

@jLopezbarb Could you check this comment out.

pkg/model/volumes.go
ifbyol: I'm not sure we should remove the err check in the if. I mean, if `sourceSubPath` is absolute and there was an error, it was always entering in this if and it could end up executing the logic in the `else` branch.

Removing the `err != nil` in the if, that path cannot happen again

PR #3329, opened 2023-01-09
Added labelSelector to configmap lists at waitForNamespaceD…

teresaromero:
# Proposed changes

Introduced at https://github.com/okteto/okteto/pull/3256 

When checking all configmaps are deleted, there was no labelselector to filter by the configmaps created by okteto.
I spotted this bug as i had one configmap that was not from okteto deployments but was at my namespace.

### DIFF
--- a/cmd/destroy/destroy.go
+++ b/cmd/destroy/destroy.go
@@ -539,7 +539,10 @@ func (pc *destroyCommand) waitForNamespaceDestroyAllToComplete(ctx context.Conte
 			case "Active":
 				if hasBeenDestroyingAll {
 					// when status is active again check if all resources have been correctly destroyed
-					cfgList, err := c.CoreV1().ConfigMaps(namespace).List(ctx, metav1.ListOptions{})
+					// list configmaps that belong okteto deployments
+					cfgList, err := c.CoreV1().ConfigMaps(namespace).List(ctx, metav1.ListOptions{
+						LabelSelector: fmt.Sprintf("%s=%s", model.GitDeployLabel, "true"),
+					})
 					if err != nil {
 						return err
 					}


AdrianPedriza (approved):

cmd/destroy/destroy.go
```diff
@@ -539,7 +539,10 @@ func (pc *destroyCommand) waitForNamespaceDestroyAllToComplete(ctx context.Conte
 			case "Active":
 				if hasBeenDestroyingAll {
 					// when status is active again check if all resources have been correctly destroyed
-					cfgList, err := c.CoreV1().ConfigMaps(namespace).List(ctx, metav1.ListOptions{})
+					// list configmaps that belong okteto deployments
+					cfgList, err := c.CoreV1().ConfigMaps(namespace).List(ctx, metav1.ListOptions{
+						LabelSelector: fmt.Sprintf("%s=%s", model.GitDeployLabel, "true"),
```
ifbyol, line 544: This would only verify for config maps created as part of a develpment environment, we would need to check if any of these labels exists (`dev.okteto.com/git-deploy`, `stack.okteto.com` or `dev.okteto.com/app`). 

I don't remember now if we could check that directly using a label selector. If not, we can list all and iterate over them to check it

teresaromero: I went around this and created a resourceLabels map, so once we retrieve the cfgmap items if any of them has one of this labels we return the error, if not, the cfgmaps are not resources one so the error is nil

ifbyol (approved):

cmd/destroy/destroy.go
```diff
@@ -539,15 +539,27 @@ func (pc *destroyCommand) waitForNamespaceDestroyAllToComplete(ctx context.Conte
 			case "Active":
 				if hasBeenDestroyingAll {
 					// when status is active again check if all resources have been correctly destroyed
+					// list configmaps that belong okteto deployments
+					resourcesLabels := map[string]bool{
```
AgustinRamiroDiaz, line 543: Not a blocker but just to bear in mind. How will this evolve over time? If we added other label to our configmaps we'd need to come here and add it also. I feel that the connection between this `resourceLabels` and the code that produces them is too far apart and only in our heads. It's a hard problem to solve :thinking:

AgustinRamiroDiaz (approved):
Nice catch!

jLopezbarb (approved):

cmd/destroy/destroy.go
```diff
@@ -539,15 +539,27 @@ func (pc *destroyCommand) waitForNamespaceDestroyAllToComplete(ctx context.Conte
...
 					}
 
-					// no configmap for the given namespace should exist
-					if len(cfgList.Items) > 0 {
-						return fmt.Errorf("namespace destroy all failed: some resources where not destroyed")
+					// no configmap for resources of the given namespace should exist
+					for _, cfg := range cfgList.Items {
+						for l := range cfg.GetLabels() {
+							if _, ok := resourcesLabels[l]; ok {
+								return fmt.Errorf("namespace destroy all failed: some resources where not destroyed")
```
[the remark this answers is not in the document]
teresaromero: if is not destroyed it would still be present at the UI, so the user can see what resources are not destroyed

teresaromero:
i'm removing backport label so this could enter on next release and not 2.11 patch, its a fix on a new feature
cc: @jLopezbarb @ifbyol

cmd/destroy/destroy.go
teresaromero: i believe is a hard problem to solve, but whenever we introduce a new resource type so check configmaps we should bear in mind, i guess the destroy will fail as there would be configmaps found for that resource

PR #3337, opened 2023-01-16
refactor remote deploy

AdrianPedriza:
# Proposed changes

- add flags forward to remote deployer
- refactor how logs are displayed to the user
- deploy always remote deploy. If no image base is set up then use a default image

### DIFF
--- a/cmd/deploy/deploy.go
+++ b/cmd/deploy/deploy.go
@@ -42,10 +42,6 @@ const (
 	succesfullyDeployedmsg = "Development environment '%s' successfully deployed"
 )
 
-var (
-	deployFromRemote = false
-)
-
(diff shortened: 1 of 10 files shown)


cmd/deploy/deploy.go
```diff
@@ -43,7 +43,7 @@ const (
 )
 
 var (
-	deployRemote = false
```
jLopezbarb, line 46: Could we move this variable to deployCommand instead of having it global

jLopezbarb (approved):

cmd/deploy/deploy.go
AdrianPedriza: done!

PR #3343, opened 2023-01-17
default OKTETO_NAMESPACE to space param in credentials query

maroshii:
A proper solution is still required to make sure we send the actual namespace value always (regardless of the envvar) but this is a safe default that fixes some issues when deploying through the okteto backend


pkg/okteto/secrets.go
```diff
@@ -84,7 +85,7 @@ func (c *userClient) GetContext(ctx context.Context) (*types.UserContext, error)
 		} `graphql:"credentials(space: $cred)"`
 	}
 	variables := map[string]interface{}{
-		"cred": graphql.String(""),
+		"cred": graphql.String(os.Getenv("OKTETO_NAMESPACE")),
```
pchico83, line 88: @jLopezbarb @maroshii this will work for pipelines. For Okteto CLI running in the laptop, `OKTETO_NAMESPACE` won't be available and we need to infer the right manifest, or the control plane won't be able to redirect to the right worker cluster

pchico83 (approved):

pkg/okteto/secrets.go
pchico83: @jLopezbarb is it safe to get the namespace from the okteto context? or it's not initialized yet?

maroshii: @pchico83 yeah this is a stopgap solution, we discuss it this morning. This is simply to unblock the deploy from the ui and `okteto pipeline deploy`. With this change we don't recreate the default namespace in the worker. A better fix is still required

ifbyol: @pchico83 the real issue that we need to fix, but we will probably need to think about it is that the function `AddOktetoCredentialsToCfg` it is not using the namespace retrieved in the credentials https://github.com/okteto/okteto/blob/0865499c19409a7c41b9b94b03d31ce7267fbde6/pkg/okteto/context.go#L360.

So even if we set here the right namespace, we are not using it to create the Kubeconfig. This works in a single multicluster approach because all the namespaces are in the same cluster, but it won't work with multicluster.

As Fran said, this is a quick fix for the installer issue, as the credentials query will be done with the right namespace and it shouldn't affect other executions, as the namespaces should be in the same cluster for current users and the function `AddOktetoCredentialsToCfg` is still using the same value

jLopezbarb: @maroshii okteto_namespace is not fixed there, but the proper fix would be to add a namespace variable to the function `getUserContext` which is called from the `getLoggedUserContext` function that contains `ctxOptions`. `ctxOptions` have the namespace if it's a command that will need it such as okteto deploy. Probably is that what you want

maroshii: @jLopezbarb you mean we should do that as part of this PR or is that part of the bigger/better fix we should do that Nacho is describing?

jLopezbarb: I'd do it as part of this PR since that would cover most of the cases, I can't ensure that `OKTETO_NAMESPACE` env var is set

maroshii: The installer always sets it so we can know for sure that it's there in those jobs which is what we are looking for to being with. It actually feels more cumbersome to me to cover _most_ of the cases and not all if we still need a follow up PR but I'm not too familiar with the cli codebase. Happy to make the changes you are suggesting but we'll need to sync since I'm not sure how to approach a more complete change.

ifbyol: @maroshii I think Javi means to do this

```
diff --git a/cmd/context/create.go b/cmd/context/create.go
index 6e268375..1e8d63e7 100644
--- a/cmd/context/create.go
+++ b/cmd/context/create.go
@@ -1,4 +1,4 @@
-// Copyright 2023 The Okteto Authors
+// Copyright 2022 The Okteto Authors
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
@@ -273,7 +273,7 @@ func getLoggedUserContext(ctx context.Context, c *ContextCommand, ctxOptions *Co

        okteto.Context().Token = user.Token

-       userContext, err := c.getUserContext(ctx)
+       userContext, err := c.getUserContext(ctx, ctxOptions.Namespace)
        if err != nil {
                return nil, err
        }
@@ -310,7 +310,7 @@ func (*ContextCommand) initKubernetesContext(ctxOptions *ContextOptions) error {
        return nil
 }

-func (c ContextCommand) getUserContext(ctx context.Context) (*types.UserContext, error) {
+func (c ContextCommand) getUserContext(ctx context.Context, ns string) (*types.UserContext, error) {
        client, err := c.OktetoClientProvider.Provide()
        if err != nil {
                return nil, err
@@ -318,7 +318,7 @@ func (c ContextCommand) getUserContext(ctx context.Context) (*types.UserContext,

        retries := 0
        for retries <= 3 {
-               userContext, err := client.User().GetContext(ctx)
+               userContext, err := client.User().GetContext(ctx, ns)

                if err != nil && oktetoErrors.IsForbidden(err) {
                        if err := c.OktetoContextWriter.Write(); err != nil {
diff --git a/pkg/okteto/secrets.go b/pkg/okteto/secrets.go
index 25703f4c..878bb316 100644
--- a/pkg/okteto/secrets.go
+++ b/pkg/okteto/secrets.go
@@ -56,7 +56,7 @@ func (c *OktetoClient) GetSecrets(ctx context.Context) ([]types.Secret, error) {
 }

 // GetSecrets returns the secrets from Okteto API
-func (c *userClient) GetContext(ctx context.Context) (*types.UserContext, error) {
+func (c *userClient) GetContext(ctx context.Context, ns string) (*types.UserContext, error) {
        var queryStruct struct {
                User struct {
                        Id              graphql.String
@@ -84,7 +84,7 @@ func (c *userClient) GetContext(ctx context.Context) (*types.UserContext, error)
                } `graphql:"credentials(space: $cred)"`
        }
        variables := map[string]interface{}{
-               "cred": graphql.String(""),
+               "cred": graphql.String(ns),
        }
        err := query(ctx, &queryStruct, variables, c.client)
        if err != nil {
diff --git a/pkg/types/interface.go b/pkg/types/interface.go
index 0f3d989e..84dbaea6 100644
--- a/pkg/types/interface.go
+++ b/pkg/types/interface.go
@@ -29,7 +29,7 @@ type OktetoInterface interface {

 // UserInterface represents the client that connects to the user functions
 type UserInterface interface {
-       GetContext(ctx context.Context) (*UserContext, error)
+       GetContext(ctx context.Context, ns string) (*UserContext, error)
 }

 // NamespaceInterface represents the client that connects to the namespace functions
```

If the environment variable `OKTETO_NAMESPACE` is set, `ctxOptions.Namespace` will contain that. Can you confirm @jLopezbarb ?

maroshii: Great, got it. Thanks @ifbyol!. Updated

cmd/context/create_test.go
```diff
@@ -426,6 +426,7 @@ func TestCheckAccessToNamespace(t *testing.T) {
 	}
 
 	for _, tt := range tests {
+		tt := tt
```
maroshii: @jLopezbarb the tests seem to be failing because of this line. It's unrelated to the PR and added it just because I saw the `loop variable tt captured by func literal` warning. I removed it but we might wanna look into it

jLopezbarb: True, I think it's because we can't run parallel tests using environment variables. I'll take a look and find a proper fix for this

maroshii:
Integration tests keep failing. I can't really tell if it's related to this change or not. It seems to be `TestUpStatefulsetV1 ` the one that's failing. @jLopezbarb any ideas?

jLopezbarb:
> Integration tests keep failing. I can't really tell if it's related to this change or not. It seems to be `TestUpStatefulsetV1 ` the one that's failing. @jLopezbarb any ideas?

The error comes by this change but I haven't had time to take a look and discover what might been failing. I'll try to look at this this afternoon

maroshii:
@jLopezbarb can you give me some pointer on what to look at? It looks like a preview deploy is failing but I can't see anything in the staging logs. This is blocking multicluster

jLopezbarb:
The last error was caused because the cluster was not available (usually because it's running an update operation). I've retrigger the job and will tell you what I see as soon as it finishes

jLopezbarb:
I found a way to reproduce the issue on e2e tests:
Run `okteto preview deploy --branch cli-e2e --repository https://github.com/okteto/movies --wait`
 using the binary produced with this branch

maroshii:
Using `--log-level=debug` shows that the error is `namespace-not-found`. I guess this makes sense because the preview environment doesn't exist at that point so all calls prior to deployPreview will fail (including the `credentials` query). Not sure the best way to fix tbh.

maroshii:
@pchico83 this is big blocker for multicluster. We cannot merge this PR and we need to discuss how to address this issue. 

**TL;DR**: We cannot pull the credentials for the right cluster beforehand because the namespace doesn't yet exist.

---

`okteto preview deploy` pulls the credentials as part of the context. We are now using the namespace in the credentials query but since the preview doesn't yet exist it returns `not-found`. I would assume the same thing would happen for the `okteto namespace` command.

This change works in the installer for previews and spaces because the api first creates the namespace and then deploys the job in the `DeployPreview` mutation so the `okteto context` succeeds.

We'll face the same problem even after we moving to the `kubetoken` auth mechanism cc @AgustinRamiroDiaz

PR #3344, opened 2023-01-18
Add fixes for remote deploy

jLopezbarb:
# Proposed changes

- Copy kubectl and helm to every deploy
- Upgrade kubectl and deploy dependencies
- Fix windows path

### DIFF
--- a/cmd/deploy/cfghandler.go
+++ b/cmd/deploy/cfghandler.go
@@ -81,10 +81,14 @@ func (ch *defaultConfigMapHandler) updateConfigMap(ctx context.Context, cfg *api
 	return errMain
 }
 
+// translateConfigMapAndDeploy with the receiver deployInsideDeployConfigMapHandler doesn't do anything
+// because we have to  control the cfmap in the main execution
 func (*deployInsideDeployConfigMapHandler) translateConfigMapAndDeploy(_ context.Context, _ *pipeline.CfgData) (*apiv1.ConfigMap, error) {
 	return nil, nil
 }
 
+// updateConfigMap with the receiver deployInsideDeployConfigMapHandler doesn't do anything
+// because we have to  control the cfmap in the main execution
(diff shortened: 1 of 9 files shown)


AdrianPedriza (approved):

AdrianPedriza (approved):

cmd/deploy/cfghandler.go
```diff
@@ -0,0 +1,91 @@
...
+	return errMain
+}
+
+func (*deployInsideDeployConfigMapHandler) translateConfigMapAndDeploy(_ context.Context, _ *pipeline.CfgData) (*apiv1.ConfigMap, error) {
+	return nil, nil
+}
+
+func (*deployInsideDeployConfigMapHandler) updateConfigMap(_ context.Context, _ *apiv1.ConfigMap, _ *pipeline.CfgData, err error) error {
+	oktetoLog.AddToBuffer(oktetoLog.InfoLevel, "Destruction failed: %s", err.Error())
+	return nil
```
AdrianPedriza, line 90: why the methods have no implementation?

jLopezbarb: Because we don't have to update the configmap if it's executed in the remote deploy

AdrianPedriza: can we add an explanatory comment about why?

jLopezbarb: Will do!!

AdrianPedriza: Wouldn't it be better to check if we are in remote or not and do the update if we are not? instead of leaving it to the interface? I say this because in this case having an empty implementation seems strange to me

jLopezbarb: Having an if steatement around in every part of the code that calls these functions is also a smell for me. Having an interface makes the tests more unitary since the code does not check an environment variable that deprives us of launching the tests in parallel

AdrianPedriza (approved):

PR #3350, opened 2023-01-20
add possible values for icon for external resource

### LINKED ISSUE #3330: Add external resources icons
As a user, I want to customize my external resource icon.

The options for introducing icons must be one of:

 - `container`
 - `dashboard`
 - `database`
 - `default`
 - `function`
 - `graph`
 - `storage`

## Acceptance criteria

- [ ] If the user introduces an icon that is not in the list we should show a proper error.
- [ ] If the user doesn't introduce any icon we must use the default one.
- [ ] Tests containing scenarios for every resource icons
- [ ] Icon must be set in the k8s object

AdrianPedriza:

# Proposed changes
- check if external resource CRD is valid against backend
- add tests covering introduced logic

Fixes #3330

### DIFF
--- a/cmd/deploy/deploy.go
+++ b/cmd/deploy/deploy.go
@@ -98,6 +98,7 @@ type DeployCommand struct {
 type ExternalResourceInterface interface {
 	Deploy(ctx context.Context, name string, ns string, externalInfo *externalresource.ExternalResource) error
 	List(ctx context.Context, ns string, labelSelector string) ([]externalresource.ExternalResource, error)
+	Validate(ctx context.Context, name string, ns string, externalInfo *externalresource.ExternalResource) error
 }
 
 // Deploy deploys the okteto manifest
@@ -364,6 +365,10 @@ func (dc *DeployCommand) RunDeploy(ctx context.Context, deployOptions *Options) 
 
 	cfg, err := getConfigMapFromData(ctx, data, c)
 	if err != nil {
+		return err
+	}
+
+	if err := dc.validateK8sResources(ctx, deployOptions.Manifest); err != nil {
 		return err
 	}
 
@@ -649,6 +654,27 @@ func (dc *DeployCommand) cleanUp(ctx context.Context) {
 	if err := dc.Proxy.Shutdown(ctx); err != nil {
 		oktetoLog.Infof("could not stop local server: %s", err)
 	}
+}
+
+func (dc *DeployCommand) validateK8sResources(ctx context.Context, manifest *model.Manifest) error {
+	if manifest.External != nil {
+		// In a cluster not managed by Okteto it is not necessary to validate the externals
+		// because they will not be deployed.
+		if okteto.IsOkteto() {
+			control, err := dc.GetExternalControl(dc.K8sClientProvider, dc.TempKubeconfigFile)
+			if err != nil {
+				return err
+			}
+
+			for externalName, externalInfo := range manifest.External {
+				err := control.Validate(ctx, externalName, manifest.Namespace, externalInfo)
+				if err != nil {
+					return err
+				}
+			}
+		}
+	}
+	return nil
 }
 
 func buildImages(ctx context.Context, build func(context.Context, *types.BuildOptions) error, getServicesToBuild func(context.Context, *model.Manifest, []string) ([]string, error), deployOptions *Options) error {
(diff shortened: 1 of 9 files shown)


pkg/externalresource/serializer.go
```diff
@@ -26,6 +43,18 @@ func (er *ExternalResource) UnmarshalYAML(unmarshal func(interface{}) error) err
 		return fmt.Errorf("there must be at least one endpoint available for the external resource")
 	}
 
+	if result.Icon == "" {
+		result.Icon = defValueForIcon
+	} else if _, ok := possibleIconValues[result.Icon]; !ok {
```
ifbyol, line 48: I think this validation should rely on the backend/Kubernetes validation. If this is just some static information and doesn't have any other implication in the CLI (like handle the icon somehow), I think it shouldn't be validated in client. If we update Kubernetes definition in the CRD to allow new values, we are forcing our customers to update to the new CLI version just to be able to use that icon because their CLI might not have that icon defined yet.

We probably need to check the dev experience to see in which phase fails or how it fails (for example, not leaving inconsistency states), but maybe there is a way of checking/validating an object definition in Kubernetes without having to create the resource.

AdrianPedriza: You are right, I think we should avoid that user experience as much as possible. Relying on validation from the backend I see two points:
- Validate on CRD creation: if it does not contain the right values, it will not be created. Although it would be the easiest option to integrate as against I think that perhaps we would fail too late? it is likely that at the time we try to create the CRD fail but previously we have built images, deployed other services .. etc
- Find a mechanism to validate the CRD before starting the deployment process. (I will work on it)

jLopezbarb: I agree that the validation must come from the backend side, but we should fail as soon as possible, without waiting for all other resources to be deployed.

If we fail after deploying the commands, compose and divert, the user may think that he didn't deploy anything and what it wasn't deployed was the external resources... 
Having to redeploy everything because of a typo it's not the best UX so validating it before deploying should be the approach to follow

ifbyol: That's why I mentioned this: `checking/validating an object definition in Kubernetes without having to create the resource.` Have we checked if there is a way of doing that maybe with `client-go` library or similar?

AdrianPedriza: I'm looking for a way to do it with `client-go`

pkg/externalresource/control.go
```diff
@@ -105,6 +105,7 @@ func translate(name string, externalResource *ExternalResource) *k8s.External {
 			},
 		},
 		Spec: k8s.ExternalResourceSpec{
+			Icon:      format.ResourceK8sMetaString(externalResource.Icon),
```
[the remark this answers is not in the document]
AdrianPedriza: since it is previously validated I believe this formatting is not necessary

jLopezbarb:
Can you take a screenshot of the message when it fails?  I want to make sure that we are showing the user the possible values or the documentation to check

AdrianPedriza:
> Can you take a screenshot of the message when it fails? I want to make sure that we are showing the user the possible values or the documentation to check

sure! 
Captura de pantalla 2023-01-20 a las 15 51 30

ifbyol (approved):

cmd/deploy/deploy.go
```diff
@@ -651,6 +656,27 @@ func (dc *DeployCommand) cleanUp(ctx context.Context) {
 	}
 }
 
+func (dc *DeployCommand) validateK8sResources(ctx context.Context, manifest *model.Manifest) error {
```
jLopezbarb, line 659: NIT: I don't know if it's the command the one that has to validate it. I think that its external struct the one that should validate itself after setting it. WDYT

AdrianPedriza: I am using the interface that exposes the behavior of the external resources package. The same principle used to make use of the Deploy() and List() logic of an external resource. In this case, the deploy command only knows that it has to call Validate() for each external resource in the manifest

jLopezbarb (approved):

cmd/deploy/deploy.go
jLopezbarb: What I mean is that I don't think that the deploy command should be the one that calls the validations. I think that it's the `GetManifest` function the one that should check that everything is correct.

We can leave it outside of this PR scope since it requires a big refactor of the function but if we create a new command that needs to validate the external resources we will need to remember that we need to call this new function that is not under the `deployCommand` scope

PR #3351, opened 2023-01-22
Wait for cronjobs to finish the execution of their jobs

pchico83:

# Proposed changes

Our current `wait` logic doesn't consider Cronjobs. This PR waits for Cronjobs to complete

### DIFF
--- a/pkg/okteto/pipeline.go
+++ b/pkg/okteto/pipeline.go
@@ -407,6 +407,12 @@ func (c *pipelineClient) GetResourcesStatus(ctx context.Context, name, namespace
 				Status     graphql.String
 				DeployedBy graphql.String
 			}
+			Cronjobs []struct {
+				ID         graphql.String
+				Name       graphql.String
+				Status     graphql.String
+				DeployedBy graphql.String
+			}
 		} `graphql:"space(id: $id)"`
 	}
 	variables := map[string]interface{}{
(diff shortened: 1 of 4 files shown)


pkg/okteto/pipeline.go
```diff
@@ -419,7 +425,7 @@ func (c *pipelineClient) GetResourcesStatus(ctx context.Context, name, namespace
 			if err != nil {
 				return nil, fmt.Errorf("could not create okteto client")
 			}
-			return okClient.Previews().GetResourcesStatusFromPreview(ctx, namespace, name)
+			return okClient.Pipeline().GetResourcesStatus(ctx, namespace, name)
```
pchico83, line 428: @jLopezbarb while implementing this, I changed this call because we are in the context of a pipeline.
Could you double check?

ifbyol: @pchico83  This was on purpose. We included this logic to handle pipelines and previews. Line 422 calls to retrieve the pipeline. If it is not-found,  we try with previews. https://github.com/okteto/okteto/pull/2954

ifbyol (approved):
We have to revert the change in `pipeline.go` line 428, but the rest of the code loos good

pchico83:
> We have to revert the change in `pipeline.go` line 428, but the rest of the code loos good

@ifbyol great, thanks for the clarification. reverted!

ifbyol (approved):

Issue #3341, opened 2023-01-17
set external endpoint URL using OKTETO_ENV file


AdrianPedriza:
After deploying the commands we must set the endpoints env variables to the correct `external.endpoint`

_Syntax_
```yaml
deploy:
  - echo "deploying commands"
  - echo "OKTETO_ENDPOINTS_DB_URL_MONGO=https://localhost:27017" >> OKTETO_ENV
external:
  db:
      endpoints:
      - name: mongo
         # url set to `https://localhost:27017`
      - name: mysql
         url: https://localhost:3306
```

# Acceptance criteria
- [ ] If OKTETO_ENV has not desired env we must fail
- [ ] External resource is able to get env value from `OKTETO_ENV` once deploy section has been executed
- [ ] Show a warning if a external resource is already set and `OKTETO_ENV` overwrites it
- [ ] Endpoints name must be unique lower cased.

PR #3354, opened 2023-01-23
add external url to okteto env file

### LINKED ISSUE #3340: add external endpoint URL to OKTETO_ENV file to be used in deploy section
Create an `OKTETO_ENV` file that is shared across the deploy commands. Every variable that is on the file must be shared across the commands

_Syntax_
```yaml
deploy:
  - echo "KEY=VALUE" >> OKTETO_ENV
  - echo $KEY # Prints VALUE
```

# Acceptance criteria
- [ ] If any endpoint related to an external resource has not been added to the external section, deploy sections must create necessary envs pointing to that endpoint following the proper format. 
- [ ] Created envs in deploy section related to external endpoints must overwrite possible initial values for the endpoint in the manifest.
- [ ] Each step in deploy section can read/write in `OKTETO_ENV` file  
- [ ] `OKTETO_ENV` file is removed after deploy dev environment. 
- [ ] `OKTETO_ENV` file is removed if cancelled.
- [ ] If one of the lines is not `KEY=VALUE` syntax, we must skip it.

AdrianPedriza:
# Proposed changes

Fix #3340 (f75b9c952084912b73e482f1fea5be86b668befc)
- create temp file (`.env`) before exec deploy commands where to overwritte environment variables for the url of the externals
- update environ with the new dynamically set values

Fix for #3341 (8342e8d4f8347874e3cd463bae9c518d192226a6)
- set external URL value if required searching values in `$OKTETO_ENV`


pkg/externalresource/external_resource.go
```diff
@@ -10,14 +10,18 @@ import (
...
+
 // ExternalResourceSection represents the map of external resources at a manifest
 type ExternalResourceSection map[string]*ExternalResource
 
 // ExternalResource represents information on an external resource
 type ExternalResource struct {
 	Icon      string
 	Notes     *Notes
-	Endpoints []ExternalEndpoint
+	Endpoints []*ExternalEndpoint
```
jLopezbarb, line 24: Why do we need to convert it to a pointer?

cmd/deploy/deploy.go
```diff
@@ -202,6 +206,7 @@ func Deploy(ctx context.Context) *cobra.Command {
 				K8sClientProvider:  okteto.NewK8sClientProvider(),
 				Builder:            buildv2.NewBuilderFromScratch(),
 				GetExternalControl: GetExternalControl,
+				Fs:                 afero.NewOsFs(),
```
jLopezbarb, line 209: We should be adding this into all deployCommand definition.

Maybe creating a constructor is the best way to go if we don't want to have unwanted panics

jLopezbarb: I think we are adding panics into init and up commands

cmd/deploy/deploy.go
```diff
@@ -656,6 +677,37 @@ func (dc *DeployCommand) cleanUp(ctx context.Context) {
 	}
 }
 
+func (dc *DeployCommand) createTempOktetoEnvFile() (afero.File, error) {
+	oktetoEnvFileDir, err := afero.TempDir(dc.Fs, "", "")
+	if err != nil {
+		return nil, err
+	}
+
+	oktetoEnvFile, err := dc.Fs.Create(fmt.Sprintf("%s/.env", oktetoEnvFileDir))
```
jLopezbarb, line 686: Should we convert `.env` to a constant?

cmd/deploy/deploy.go
```diff
@@ -656,6 +677,37 @@ func (dc *DeployCommand) cleanUp(ctx context.Context) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	os.Setenv(constants.OktetoEnvFile, oktetoEnvFile.Name())
```
jLopezbarb, line 691: Why do we need to set this env var?

cmd/deploy/deploy.go
```diff
@@ -656,6 +677,37 @@ func (dc *DeployCommand) cleanUp(ctx context.Context) {
...
+	return oktetoEnvFile, nil
+}
+
+func (*DeployCommand) updateEnvironWithOktetoEnvFile(oktetoEnvFile afero.File) error {
+	envMap, err := godotenv.ParseWithLookup(oktetoEnvFile, os.LookupEnv)
```
jLopezbarb, line 696: What happen if the env file has one of the lines bad formatted? Do we load all others? Does it fail? 

Since it's something that is the user the one that adds lines to it I think we should just skip that line

cmd/deploy/deploy.go
AdrianPedriza: to be able to use `$OKTETO_ENV` in the commands that add env to the file, for example : `echo "KEY=value" >> $OKTETO_ENV`

cmd/deploy/deploy.go
AdrianPedriza: It would only be used here by aasique... I would say no

pkg/externalresource/control.go
```diff
@@ -148,9 +153,28 @@ func translateK8sToExternal(er k8s.External) ExternalResource {
...
-func translateK8sToEndpoint(k8sEndpoint k8s.Endpoint) ExternalEndpoint {
-	return ExternalEndpoint{
+func translateK8sToEndpoint(k8sEndpoint k8s.Endpoint) *ExternalEndpoint {
+	return &ExternalEndpoint{
 		Name: k8sEndpoint.Name,
 		Url:  k8sEndpoint.Url,
 	}
 }
+
+func setValuesFromOktetoEnvFile(name string, endpoints []*ExternalEndpoint) error {
```
jLopezbarb, line 163: I think we are missing tests for this function

cmd/deploy/deploy.go
AdrianPedriza: With the current implementation and following the acceptance criteria, if one of the lines has an unaccepted format we fail the whole deployment. I think that if one of the lines does not have the right format we should fail since that line is part of the dev environment deployment process and it could not be executed correctly so it can leave an undesired state to the development environment

cmd/deploy/deploy.go
jLopezbarb, line 686: Use `filepath.Join`. Using `/` is not windows compatible

cmd/deploy/deploy.go
AdrianPedriza: I will keep an eye on it, it is true that it is prone to panics

cmd/deploy/deploy.go
jLopezbarb, line 686: We should be logging in debug which is the file used

pkg/externalresource/external_resource.go
AdrianPedriza: I did it to update the object by passing the struct of the endpoints by reference

jLopezbarb (approved):
LGTM

PR #3356, opened 2023-01-23
fix zero value for external icon field

AdrianPedriza:

# Proposed changes
external resources with zero value related to `icon` field should not need to be sent to the backend for validation

### DIFF
--- a/integration/deploy/external_test.go
+++ b/integration/deploy/external_test.go
@@ -37,6 +37,7 @@ deploy:
     command: echo hola
 external:
   test:
+    icon: default
     notes: readme.md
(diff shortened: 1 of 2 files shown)


integration/deploy/external_test.go
```diff
@@ -37,6 +37,7 @@ deploy:
     command: echo hola
 external:
   test:
+    icon: default
```
jLopezbarb, line 40: Why do we need to set icon? If icon is not set we should be setting `default` as the default value

jLopezbarb (approved):

PR #3362, opened 2023-01-26
Pull first approach

### LINKED ISSUE #3348: Pull first approach
This issue aims to create a better experience for `okteto up/deploy` if the images are already build.

The command okteto build will always build into the okteto registry, even if the image name is set to another registry. For example if we have:
```yaml
build:
  app:
    context: .
    image: my-image/app:v1
```
we will build `okteto.dev/app`

When checking images we will have to check first if the image is in the `okteto.dev` or the `okteto.global` before retrieving the image from the image.

In okteto deploy and okteto up we have to check that the image was already built

# Acceptance criteria

- [ ] Okteto build will always build into our registry if not set
- [ ] Okteto up/deploy will check if the image is in dev or global registry

jLopezbarb:
# Proposed changes

Fixes #3348 

This PR is separated into two changes to fulfill the acceptance criteria:
- `23c895b`: Always checks for the image to be in `okteto.dev`, `okteto.global` and the original image defined. Add tests for non covered function.
- `d22def7`: Move the set build environment outside of the function that checks if an image is already built or not. cc: @teresaromero 
- `b299877`: Build always to the okteto registry if we are in an okteto cluster.

## Implications

- This PR alters the current behavior of the building process, since everything will be built into the okteto registry. 
- You can still push to another registry if you run okteto build using a Dockerfile and a tag

## Tests done

- With compose using external registry:
```yaml
services:
  vote:
    build:
      context: vote
      image: jlopezbarb/vote
    ports:
      - 8080:8080
    environment:
      - FLASK_ENV=development
    command: python app.py
    volumes:
    - vote:/src
 ```
 It builds first the `okteto-with-volumes` and then creates `okteto.dev/voting-vote:okteto`
- With compose using okteto registry:
```yaml
services:
  vote:
    build:
      context: vote
      image: okteto.dev/test
    ports:
      - 8080:8080
    environment:
      - FLASK_ENV=development
    command: python app.py
 ```
Okteto builds `okteto.dev/test`

- With build section:
```yaml
build:
  vote:
    context: vote
    image: okteto.dev/test
  frontend:
    context: vote
    image: jlopezbarb/frontend-vote
```
Okteto builds `okteto.dev/test` for vote build and `okteto.dev/voting-frontend:okteto` for the frontend build. 
Redeploying it shows that the images are already built even if there is no `jlopezbarb/frontend-vote`

- Check pull first approach:
With the same manifest as before:
1. Run `okteto build -f Dockerfile -t `jlopezbarb/frontend-vote` vote`
2. Run `okteto deploy`
3. Check that is not building it because it's already built in the docker registry


cmd/build/v2/services.go
```diff
@@ -43,7 +43,25 @@ func (bc *OktetoBuilder) GetServicesToBuild(ctx context.Context, manifest *model
...
+				return nil
+			}
+			oktetoLog.Debug("Skipping build for image for service")
+
+			bc.SetServiceEnvVars(svc, digest)
+
+			if manifest.Deploy != nil && manifest.Deploy.ComposeSection != nil && manifest.Deploy.ComposeSection.Stack != nil {
+				stack := manifest.Deploy.ComposeSection.Stack
+				if svcInfo, ok := stack.Services[svc]; ok && svcInfo.Image == "" {
+					stack.Services[svc].Image = fmt.Sprintf("${OKTETO_BUILD_%s_IMAGE}", strings.ToUpper(strings.ReplaceAll(svc, "-", "_")))
```
AdrianPedriza, line 61: when is this image value from an stack loaded as env?

cmd/build/v2/services.go
```diff
@@ -74,55 +92,56 @@ func (bc *OktetoBuilder) GetServicesToBuild(ctx context.Context, manifest *model
...
-		return fmt.Errorf("error checking image at registry %s: %v", tag, err)
+	for _, tag := range tags {
+		imageWithDigest, err := bc.Registry.GetImageTagWithDigest(tag)
+		if oktetoErrors.IsNotFound(err) {
+			oktetoLog.Debugf("image not found for tag %s, building image", tag)
+			continue
+		} else if err != nil {
+			return "", fmt.Errorf("error checking image at registry %s: %w", tag, err)
+		}
+		return imageWithDigest, nil
```
AdrianPedriza, line 111: why we are looping over tags but we only return one digest? not sure if I understand this logic

cmd/build/v2/services.go
```diff
@@ -74,55 +92,56 @@ func (bc *OktetoBuilder) GetServicesToBuild(ctx context.Context, manifest *model
...
+	targetRegistries := []string{okteto.DevRegistry, okteto.GlobalRegistry}
+	for _, targetRegistry := range targetRegistries {
+		if shouldAddVolumeMounts(b) {
+			possibleTags = append(possibleTags, fmt.Sprintf("%s/%s-%s:%s", targetRegistry, sanitizedName, svcName, model.OktetoImageTagWithVolumes))
+			continue
+		}
+		if shouldBuildFromDockerfile(b) {
+			possibleTags = append(possibleTags, fmt.Sprintf("%s/%s-%s:%s", targetRegistry, sanitizedName, svcName, model.OktetoDefaultImageTag))
+		}
+	}
```
AdrianPedriza, line 142:
```suggestion
if shouldAddVolumeMounts(b) {
		possibleTags = append(possibleTags, fmt.Sprintf("%s/%s-%s:%s", okteto.DevRegistry, sanitizedName, svcName, model.OktetoImageTagWithVolumes))
		possibleTags = append(possibleTags, fmt.Sprintf("%s/%s-%s:%s", okteto.GlobalRegistry, sanitizedName, svcName, model.OktetoImageTagWithVolumes))
}
if shouldBuildFromDockerfile(b) {
		possibleTags = append(possibleTags, fmt.Sprintf("%s/%s-%s:%s", okteto.DevRegistry, sanitizedName, svcName, model.OktetoDefaultImageTag))
		possibleTags = append(possibleTags, fmt.Sprintf("%s/%s-%s:%s", okteto.GlobalRegistry, sanitizedName, svcName, model.OktetoDefaultImageTag))
	}
```
WDYT? to not check twice `shouldAddVolumeMounts(b)` and `shouldBuildFromDockerfile(b)`. If it's true for `okteto.dev` is also true for `okteto.global`

cmd/build/v2/services_test.go
```diff
@@ -186,15 +196,111 @@ func TestGetToBuildTag(t *testing.T) {
...
+					"test": {
+						Namespace: "test",
+						IsOkteto:  tt.isOkteto,
+						Registry:  "https://registry.test",
+					},
+				},
+				CurrentContext: "test",
+			}
+
+			bc := NewBuilder(nil, fakeReg)
```
AdrianPedriza, line 291:
```suggestion
			bc := NewBuilder(nil, fakeReg)
```
we can move this line to the top to avoid declare same var `bc` with same values inside the test loop

cmd/build/v2/services.go
jLopezbarb: We have to load the compose twice. The first one we need to get the builds from it, we add the build variables and then the second one we use it to load the env vars. We'll probably need to revisit this logic when we move to the official docker-compose serialiser

cmd/build/v2/services.go
jLopezbarb: We need to check if the image is already built in one of the okteto registries. If found we return that digest, which is the one we'll be using for the image

cmd/build/v2/services.go
jLopezbarb: true, this suggestion reduces the complexity of the code. Thanks for spotting it!! 🚀

cmd/build/v2/services.go
AdrianPedriza: could we looking for that digest for the specific tag? why we need to loop over the tags? for me it is like "I will try if this exists, if not, will try next one..?" Is it?

cmd/build/v2/services.go
AdrianPedriza: I said that because I saw that for a service not declared as a stack I see https://github.com/okteto/okteto/blob/658767f05dba71c00109a322ecd81f2263c72722/cmd/build/v2/services.go#L56 which loads the necessary environment variables but for services declared as part of a stack I don't see this environment variable loading

cmd/build/v2/services.go
jLopezbarb: We wanted to point `jlopezbarb/vote`, `okteto.global/vote` and `okteto.dev/vote to the same digest. That's why we try with the three different tags

jLopezbarb:
Closed since this need to be redefined

PR #3365, opened 2023-01-27
switch to personal namespace if previous namespace is missing

### LINKED ISSUE #3334: Can't switch to previous context with non existent / deleted namespace
**Describe the bug**
When using `okteto context` or `okteto context use CONTEXT_NAME`, I can't switch back to an Okteto Cluster context that points to a non existent or deleted namespace.

That namespace may have been created from a CLI instance and deleted from somewhere else, such as a different CLI instance or Okteto UI.

I need to run `okteto context delete CONTEXT_NAME` to delete the context and `okteto context` or `okteto context use CONTEXT_NAME` to create it again.

**To Reproduce**

1. Use `okteto context` or `okteto context use CONTEXT_NAME` to switch to an Okteto Cluster context.
2. Create a namespace with `okteto namespace` or `okteto namespace create NAMESPACE_NAME`.
3. Switch to a different context (any type) with `okteto context` or `okteto context use CONTEXT_NAME`.
4. Delete the namespace created in the 2nd step from a different CLI instance or Okteto UI.
5. Try to switch back to the context used in the 1st step with `okteto context` or `okteto context use CONTEXT_NAME`
6. See the error (attached below).

**Expected behavior**
I expect a warning about the namespace not being found and switch by default to my personal namespace.

- OS: macOS arm64
- Version 2.11.0

**Additional context**

Error:

```
% okteto context
 ✓  Context 'https://cloud.okteto.com' selected
 x  Namespace 'foo-jpf-okteto' not found on context 'https://cloud.okteto.com'
    Please verify that the namespace exists and that you have access to it.
```

miniscruff:
# Proposed changes

Fixes #3334

Add a new value to context options indicating if the command originated from `ctx` or `ns`.
If switching context, and we do not have access to the previous namespace, switch to our personal one instead of crashing.

### DIFF
--- a/cmd/context/create.go
+++ b/cmd/context/create.go
@@ -172,8 +172,19 @@ func (c *ContextCommand) UseContext(ctx context.Context, ctxOptions *ContextOpti
 			return err
 		}
 
-		if !hasAccess {
-			return oktetoErrors.UserError{E: fmt.Errorf("namespace '%s' not found on context '%s'", ctxOptions.Namespace, ctxOptions.Context),
+		// if using a new context, our cached namespace may have been removed
+		// so swap over to the personal namespace instead of erroring
+		if !hasAccess && !ctxOptions.FromNamespace {
+			oktetoLog.Success(
+				"No access to namespace '%s' switching to personal namespace '%s'",
+				ctxOptions.Namespace,
+				okteto.Context().PersonalNamespace,
+			)
+			currentCtx := ctxStore.Contexts[ctxOptions.Context]
+            currentCtx.Namespace = currentCtx.PersonalNamespace
+		} else if !hasAccess {
+			return oktetoErrors.UserError{
+				E:    fmt.Errorf("namespace '%s' not found on context '%s'", ctxOptions.Namespace, ctxOptions.Context),
 				Hint: "Please verify that the namespace exists and that you have access to it.",
 			}
 		}
--- a/cmd/context/create.go
+++ b/cmd/context/create.go
@@ -172,21 +172,23 @@ func (c *ContextCommand) UseContext(ctx context.Context, ctxOptions *ContextOpti
 			return err
 		}
 
-		// if using a new context, our cached namespace may have been removed
-		// so swap over to the personal namespace instead of erroring
-		if !hasAccess && !ctxOptions.FromNamespace {
-			oktetoLog.Success(
+		if !hasAccess {
+			if ctxOptions.CheckNamespaceAccess {
+				return oktetoErrors.UserError{
+					E:    fmt.Errorf("namespace '%s' not found on context '%s'", ctxOptions.Namespace, ctxOptions.Context),
+					Hint: "Please verify that the namespace exists and that you have access to it.",
+				}
+			}
+
+			// if using a new context, our cached namespace may have been removed
+			// so swap over to the personal namespace instead of erroring
+			oktetoLog.Warning(
 				"No access to namespace '%s' switching to personal namespace '%s'",
 				ctxOptions.Namespace,
 				okteto.Context().PersonalNamespace,
 			)
 			currentCtx := ctxStore.Contexts[ctxOptions.Context]
-            currentCtx.Namespace = currentCtx.PersonalNamespace
-		} else if !hasAccess {
-			return oktetoErrors.UserError{
-				E:    fmt.Errorf("namespace '%s' not found on context '%s'", ctxOptions.Namespace, ctxOptions.Context),
-				Hint: "Please verify that the namespace exists and that you have access to it.",
-			}
+			currentCtx.Namespace = currentCtx.PersonalNamespace
 		}
 
 		if err := c.OktetoContextWriter.Write(); err != nil {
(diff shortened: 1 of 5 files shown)


cmd/context/create.go
```diff
@@ -172,8 +172,19 @@ func (c *ContextCommand) UseContext(ctx context.Context, ctxOptions *ContextOpti
 			return err
 		}
 
-		if !hasAccess {
-			return oktetoErrors.UserError{E: fmt.Errorf("namespace '%s' not found on context '%s'", ctxOptions.Namespace, ctxOptions.Context),
+		// if using a new context, our cached namespace may have been removed
+		// so swap over to the personal namespace instead of erroring
+		if !hasAccess && !ctxOptions.FromNamespace {
+			oktetoLog.Success(
```
jLopezbarb, line 178: Could you provide a screenshoot of how this looks like?

I guess we should be using `oktetoLog.Warning` here, but I want to see how it looks like

miniscruff: Sure, here is the before and after.
image

miniscruff: I can swap over logging levels if you want

AgustinRamiroDiaz: Yeah I agree with @jLopezbarb , I think we should show a `Warning` or `Info` here

cmd/context/create.go
```diff
@@ -172,8 +172,19 @@ func (c *ContextCommand) UseContext(ctx context.Context, ctxOptions *ContextOpti
 			return err
 		}
 
-		if !hasAccess {
-			return oktetoErrors.UserError{E: fmt.Errorf("namespace '%s' not found on context '%s'", ctxOptions.Namespace, ctxOptions.Context),
+		// if using a new context, our cached namespace may have been removed
+		// so swap over to the personal namespace instead of erroring
+		if !hasAccess && !ctxOptions.FromNamespace {
```
AgustinRamiroDiaz, line 177: can we simplify the boolean logic here? maybe something like:
```
if !hasAccess {
  if ctxOptions.FromNamespace {
    // Return error here
  }
  // Handle default here
}
```

cmd/context/options.go
```diff
@@ -31,6 +31,7 @@ type ContextOptions struct {
 	Show                  bool
 	Save                  bool
 	IsCtxCommand          bool
+	FromNamespace         bool
```
AgustinRamiroDiaz, line 34: I find this tied to the cobra command and very specific and hard to reuse. What do you think about using a `DefaultNamespace string` property instead? That way it has an independent meaning from the command and can be reused.

Then in the error message we'd say "Couldn't access namespace ..., defaulting to ..."

AgustinRamiroDiaz (commented):
Thanks for your contribution! Code looks good overall :rocket:

cmd/context/create.go
jLopezbarb: I think we should replace it with a warning message. I think that using success hides the meaning of the message, which is that the user should know why we are redirecting him to his personal namespace. WDYT?

cmd/context/options.go
jLopezbarb: @AgustinRamiroDiaz This option is to know if the command is executed from the namespace command instead of the context one. It's the same as `IsCtxCommand`, `show` or `save`. 

If we use `DefaultNamespace` with the name of the personal namespace only when it's a namespace command it's not obvious and will be reused in a wrong way. 

We'd need to rethink how the `namespace` and `context` command are written but right now there is no plans to do it, so the least intrusive way to do it I think is this one. WDYT?

AgustinRamiroDiaz:
@miniscruff we should also handle the `FromNamespace` context option in https://github.com/okteto/okteto/blob/259e9f674e0178cfc83a75fb59845dc40164afab/cmd/context/use.go#L84

cmd/context/options.go
AgustinRamiroDiaz: hmmm yeah that's a good point. I felt like we could improve it but you are right, it'll get misused if the name is `DefaultNamespace`. I still feel that it's not a good abstraction, but to change it we'd need to rethink the whole command to have more composable pieces instead of one big command.

I still think that we should rename it since in the comment below we'd be using it in the context command too

jLopezbarb: We could name it `checkIfHasAccessToNamespace` or something like this. WDYT?

miniscruff: Yea that makes sense, let me change that around

miniscruff:
> 



> @miniscruff I think we should also handle the `FromNamespace` context option in
> 
> https://github.com/okteto/okteto/blob/259e9f674e0178cfc83a75fb59845dc40164afab/cmd/context/use.go#L84

Right, we would also need to check it if you supplied the namespace, in which case backing off would be appropriate, gotcha.

cmd/context/options.go
miniscruff: I used `CheckNamespaceAccess`, let me know if that works

cmd/context/create.go
miniscruff: Done

cmd/context/create.go
miniscruff: Done

cmd/context/options.go
jLopezbarb: That's perfect

AgustinRamiroDiaz (approved):
Changes look good! Deferring merge to @jLopezbarb

jLopezbarb (approved):

PR #3368, opened 2023-01-31
Make okteto up wait until original resource is awaken before starting dev container

ifbyol:

# Proposed changes

Fixes https://github.com/okteto/app/issues/5520

Changes in the up command to wake the namespace up (if it is sleeping and it is an okteto context) and wait until the annotation dev.okteto.com/state-before-sleeping is not present in the original deployment before deploying the dev container. The main changes are:
* If the context is an Okteto one, we call to the `wakeSpace` mutation as one of the first steps of the command
* Before cloning the original sfs/deployment, check if it is already awaken or not. The reason is because that resource could depend on other service which is not already awaken. As the wake function will awake resources in order taking into account the dependencies, if the original resource don't have the annotation it means that all their dependencies are already up

### DIFF
--- a/cmd/up/activate.go
+++ b/cmd/up/activate.go
@@ -115,6 +115,11 @@ func (up *upContext) activate() error {
 		lastPodUID = up.Pod.UID
 	}
 
+	if err := up.waitUntilAppIsAwaken(ctx, app); err != nil {
+		oktetoLog.Infof("error waiting for the original %s to be awaken: %s", app.Kind(), err.Error())
+		return err
+	}
+
 	if err := up.devMode(ctx, app, create); err != nil {
 		if oktetoErrors.IsTransient(err) {
 			return err
@@ -456,4 +461,30 @@ func getPullingMessage(message, namespace string) string {
 	}
 	toReplace := fmt.Sprintf("%s/%s", registry, namespace)
 	return strings.Replace(message, toReplace, okteto.DevRegistry, 1)
+}
+
+// waitUntilAppIsAwaken waits until the app is awaken checking if the annotation dev.okteto.com/state-before-sleeping is present in the app resource
+func (up *upContext) waitUntilAppIsAwaken(ctx context.Context, app apps.App) error {
+	if _, ok := app.ObjectMeta().Annotations[model.StateBeforeSleepingAnnontation]; !ok {
+		return nil
+	}
+
+	timeout := 5 * time.Minute
+	to := time.NewTicker(timeout)
+	ticker := time.NewTicker(10 * time.Second)
+	for {
+		select {
+		case <-to.C:
+			return fmt.Errorf("'%s' '%s' didn't wake up after %s", app.Kind(), app.ObjectMeta().Name, timeout.String())
+		case <-ticker.C:
+			if err := app.Refresh(ctx, up.Client); err != nil {
+				return err
+			}
+
+			// If the app is not sleeping anymore, we are done
+			if _, ok := app.ObjectMeta().Annotations[model.StateBeforeSleepingAnnontation]; !ok {
+				return nil
+			}
+		}
+	}
 }
--- a/cmd/up/up.go
+++ b/cmd/up/up.go
@@ -36,6 +36,8 @@ import (
 	"github.com/okteto/okteto/pkg/constants"
 	"github.com/okteto/okteto/pkg/devenvironment"
 	"github.com/okteto/okteto/pkg/discovery"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
 
 	"github.com/okteto/okteto/pkg/cmd/pipeline"
 	"github.com/okteto/okteto/pkg/config"
@@ -293,6 +295,22 @@ func Up() *cobra.Command {
 				// update autocreate property if needed to be forced
 				oktetoLog.Info("Setting Autocreate to true because manifest v1 and flag --deploy")
 				up.Dev.Autocreate = true
+			}
+
+			// only if the context is an okteto one, we should verify if the namespace has to be woken up
+			if okteto.Context().IsOkteto {
+				k8sClient, _, err := okteto.NewK8sClientProvider().Provide(okteto.Context().Cfg)
+				if err != nil {
+					return err
+				}
+				okClient, err := okteto.NewOktetoClient()
+				if err != nil {
+					return err
+				}
+				if err := wakeNamespaceIfApplies(ctx, up.Dev.Namespace, k8sClient, okClient); err != nil {
+					// If there is an error waking up namespace, we don't want to fail the up command
+					oktetoLog.Infof("failed to wake up the namespace: %s", err.Error())
+				}
 			}
 
 			if err := setBuildEnvVars(ctx, oktetoManifest); err != nil {
@@ -902,4 +920,20 @@ func setBuildEnvVars(ctx context.Context, m *model.Manifest) error {
 		Manifest:    m,
 	}
 	return builder.Build(ctx, buildOptions)
+}
+
+// wakeNamespaceIfApplies wakes the namespace if it is sleeping
+func wakeNamespaceIfApplies(ctx context.Context, ns string, k8sClient kubernetes.Interface, okClient types.OktetoInterface) error {
+	n, err := k8sClient.CoreV1().Namespaces().Get(ctx, ns, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+
+	// If the namespace is not sleeping, do nothing
+	if n.Labels["space.okteto.com/status"] != "Sleeping" {
+		return nil
+	}
+
+	oktetoLog.Information("Namespace '%s' is sleeping, waking it up...", ns)
+	return okClient.Namespaces().Wake(ctx, ns)
 }
--- a/cmd/up/up_test.go
+++ b/cmd/up/up_test.go
@@ -19,10 +19,16 @@ import (
 	"fmt"
 	"testing"
 
+	"github.com/okteto/okteto/internal/test/client"
 	oktetoErrors "github.com/okteto/okteto/pkg/errors"
 	"github.com/okteto/okteto/pkg/model"
 	"github.com/okteto/okteto/pkg/model/forward"
+	"github.com/okteto/okteto/pkg/types"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes/fake"
 )
 
 func Test_waitUntilExitOrInterrupt(t *testing.T) {
@@ -340,6 +346,55 @@ func TestCommandAddedToUpOptionsWhenPassedAsFlag(t *testing.T) {
 			}
 
 			assert.Equal(t, tt.expectedCommand, flagValue)
+		})
+	}
+}
+
+func TestWakeNamespaceIfAppliesWithoutErrors(t *testing.T) {
+	tests := []struct {
+		name              string
+		ns                v1.Namespace
+		expectedWakeCalls int
+	}{
+		{
+			name: "wake namespace if it is not sleeping",
+			ns: v1.Namespace{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "test",
+					Labels: map[string]string{
+						"space.okteto.com/status": "Active",
+					},
+				},
+			},
+			expectedWakeCalls: 0,
+		},
+		{
+			name: "wake namespace if it is sleeping",
+			ns: v1.Namespace{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "test",
+					Labels: map[string]string{
+						"space.okteto.com/status": "Sleeping",
+					},
+				},
+			},
+			expectedWakeCalls: 1,
+		},
+	}
+	ctx := context.Background()
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			k8sClient := fake.NewSimpleClientset(&tt.ns)
+			nsClient := client.NewFakeNamespaceClient([]types.Namespace{}, nil)
+			oktetoClient := &client.FakeOktetoClient{
+				Namespace: nsClient,
+			}
+
+			err := wakeNamespaceIfApplies(ctx, tt.ns.Name, k8sClient, oktetoClient)
+
+			require.NoError(t, err)
+			require.Equal(t, tt.expectedWakeCalls, nsClient.WakeCalls)
 		})
 	}
 }
--- a/internal/test/client/namespace.go
+++ b/internal/test/client/namespace.go
@@ -22,6 +22,9 @@ import (
 type FakeNamespaceClient struct {
 	namespaces []types.Namespace
 	err        error
+
+	// WakeCalls is the number of times Wake was called
+	WakeCalls int
 }
 
 func NewFakeNamespaceClient(ns []types.Namespace, err error) *FakeNamespaceClient {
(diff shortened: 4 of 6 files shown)


cmd/up/activate.go
```diff
@@ -457,3 +462,29 @@ func getPullingMessage(message, namespace string) string {
...
+			return fmt.Errorf("'%s' '%s' didn't wake up after %s", app.Kind(), app.ObjectMeta().Name, timeout.String())
+		case <-ticker.C:
+			if err := app.Refresh(ctx, up.Client); err != nil {
+				return err
+			}
+
+			// If the app is not sleeping anymore, we are done
+			if _, ok := app.ObjectMeta().Annotations[model.StateBeforeSleepingAnnontation]; !ok {
+				return nil
+			}
```
ifbyol, line 487: Probably it is a good idea to print something here to the user in case we have to wait too much, WDYT?

cmd/up/activate.go
```diff
@@ -457,3 +462,29 @@ func getPullingMessage(message, namespace string) string {
...
 	return strings.Replace(message, toReplace, okteto.DevRegistry, 1)
 }
+
+// waitUntilAppIsAwaken waits until the app is awaken checking if the annotation dev.okteto.com/state-before-sleeping is present in the app resource
+func (up *upContext) waitUntilAppIsAwaken(ctx context.Context, app apps.App) error {
+	if _, ok := app.ObjectMeta().Annotations[model.StateBeforeSleepingAnnontation]; !ok {
+		return nil
+	}
+
+	timeout := 5 * time.Minute
```
ifbyol, line 472: @pchico83 I'm setting 5 minutes as the maximum time to wait for a resource to be awaken. Do you think we should increase it?

cmd/up/activate.go
```diff
@@ -457,3 +462,29 @@ func getPullingMessage(message, namespace string) string {
 	toReplace := fmt.Sprintf("%s/%s", registry, namespace)
 	return strings.Replace(message, toReplace, okteto.DevRegistry, 1)
 }
+
+// waitUntilAppIsAwaken waits until the app is awaken checking if the annotation dev.okteto.com/state-before-sleeping is present in the app resource
+func (up *upContext) waitUntilAppIsAwaken(ctx context.Context, app apps.App) error {
```
ifbyol, line 467: In this case I'm not checking if the context is an Okteto one or not because the resource shouldn't have that annotation in a non Okteto cluster

cmd/up/activate.go
AdrianPedriza: yes IMO. User might think that there is something wrong when launching his dev env and he just has to be a little patient.

cmd/up/up.go
```diff
@@ -903,3 +921,19 @@ func setBuildEnvVars(ctx context.Context, m *model.Manifest) error {
...
+
+// wakeNamespaceIfApplies wakes the namespace if it is sleeping
+func wakeNamespaceIfApplies(ctx context.Context, ns string, k8sClient kubernetes.Interface, okClient types.OktetoInterface) error {
+	n, err := k8sClient.CoreV1().Namespaces().Get(ctx, ns, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+
+	// If the namespace is not sleeping, do nothing
+	if n.Labels["space.okteto.com/status"] != "Sleeping" {
```
AdrianPedriza, line 933: should we use a constant for this label?

cmd/up/up.go
```diff
@@ -295,6 +297,22 @@ func Up() *cobra.Command {
...
+				k8sClient, _, err := okteto.NewK8sClientProvider().Provide(okteto.Context().Cfg)
+				if err != nil {
+					return err
+				}
+				okClient, err := okteto.NewOktetoClient()
+				if err != nil {
+					return err
+				}
+				if err := wakeNamespaceIfApplies(ctx, up.Dev.Namespace, k8sClient, okClient); err != nil {
+					// If there is an error waking up namespace, we don't want to fail the up command
```
AdrianPedriza, line 311: if the up command is not directly affected by a wake-up error, could we run `wakeNamespaceIfApplies` in a goroutine?

cmd/up/activate.go
ifbyol: Yes, I'll include a message

cmd/up/up.go
ifbyol: Yes, I'll do it in a goroutine to avoid to impact on the command performance when everything is up

cmd/up/up.go
ifbyol: yeah, I can move it to a constant

AdrianPedriza: I saw that there are several places where we have hardcoded this value so we will need to call the new constant in them as well

cmd/up/activate.go
ifbyol: Done! I included a spinner

cmd/up/up.go
ifbyol: Changed in all the places

cmd/up/activate.go
```diff
@@ -457,3 +462,42 @@ func getPullingMessage(message, namespace string) string {
...
+	appToCheck := app
+	// If the app is already in dev mode, we need to check the cloned app to see if it is awaken
+	if apps.IsDevModeOn(app) {
+		var err error
+		appToCheck, err = app.GetCloned(ctx, up.Client)
+		if err != nil {
+			return err
+		}
+	}
```
ifbyol, line 476: I had to include some logic to detect if the resource is already on dev mode. In that case, we need to check the cloned resource as it is the one with the annotation to check.

This scenario might happen when a namespace is slept but some of the resources were in dev mode but without being attached to the pod

cmd/up/up.go
```diff
@@ -295,6 +297,27 @@ func Up() *cobra.Command {
 				up.Dev.Autocreate = true
 			}
 
+			// only if the context is an okteto one, we should verify if the namespace has to be woken up
+			if okteto.Context().IsOkteto {
```
ifbyol, line 301: Not sure if this is the best place to execute this or it could be before, can you check @jLopezbarb ?

pkg/k8s/apps/interface.go
```diff
@@ -38,6 +39,8 @@ type App interface {
 	// TODO: remove after people move to CLI >= 1.14
 	RestoreOriginal() error
 
+	// GetCloned returns the cloned app from Kubernetes
+	GetCloned(ctx context.Context, c kubernetes.Interface) (App, error)
```
ifbyol, line 43: The idea of this function is to get from Kubernetes the cloned resource. This is needed because of this https://github.com/okteto/okteto/pull/3368/files#r1093193476

ifbyol:
I recorded a video on how the experience looks like to the user, you can watch it [here](https://www.loom.com/share/c1bf9e49dba54bbeba72910ef9d1b8ef).

@rlamana @pchico83 can you review it? Specially the 2 messages I added and the wording. Specifically:
* if the namespace is slept, there is a new information message ` i  Namespace 'nacho' is sleeping, waking it up...` 
* If the up command has to wait until the resource is up, there is a new spinner with the message `Dev environment '%s' is sleeping. Waiting for it to wake up...`
* Wording of the messages, specially, terms wake up (not sure if we prefer to use other term) and Dev environment (we are not waiting for the dev container because at that point it is not created yet)

cmd/up/up.go
```diff
@@ -903,3 +926,19 @@ func setBuildEnvVars(ctx context.Context, m *model.Manifest) error {
...
+
+// wakeNamespaceIfApplies wakes the namespace if it is sleeping
+func wakeNamespaceIfApplies(ctx context.Context, ns string, k8sClient kubernetes.Interface, okClient types.OktetoInterface) error {
+	n, err := k8sClient.CoreV1().Namespaces().Get(ctx, ns, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+
+	// If the namespace is not sleeping, do nothing
+	if n.Labels[constants.NamespaceStatusLabel] != "Sleeping" {
```
teresaromero, line 938: nit: should `Sleeping` also be a constant?

teresaromero (approved):
LGTM

pkg/k8s/apps/deployments.go
```diff
@@ -207,3 +207,13 @@ func (i *DeploymentApp) Destroy(ctx context.Context, c kubernetes.Interface) err
 func (i *DeploymentApp) PatchAnnotations(ctx context.Context, c kubernetes.Interface) error {
 	return deployments.PatchAnnotations(ctx, i.d, c)
 }
+
+// GetCloned Returns from Kubernetes the cloned deployment
+func (i *DeploymentApp) GetCloned(ctx context.Context, c kubernetes.Interface) (App, error) {
```
jLopezbarb, line 212: Could we add a simple test for this function on deployment and statefulset?

cmd/up/up.go
```diff
@@ -295,6 +297,27 @@ func Up() *cobra.Command {
 				up.Dev.Autocreate = true
 			}
 
+			// only if the context is an okteto one, we should verify if the namespace has to be woken up
+			if okteto.Context().IsOkteto {
+				// We execute it in a goroutine to not impact the command performance
+				go func() {
+					k8sClient, _, err := okteto.NewK8sClientProvider().Provide(okteto.Context().Cfg)
```
jLopezbarb, line 304: We already have it initialised at line 241. Could we reuse that one?

cmd/up/activate.go
```diff
@@ -457,3 +462,42 @@ func getPullingMessage(message, namespace string) string {
...
+
+	if _, ok := appToCheck.ObjectMeta().Annotations[model.StateBeforeSleepingAnnontation]; !ok {
+		return nil
+	}
+
+	timeout := 5 * time.Minute
+	to := time.NewTicker(timeout)
+	ticker := time.NewTicker(10 * time.Second)
```
jLopezbarb, line 484: I'd reduce the time between calls since this is a blocking function, WDYT?

cmd/up/activate.go
jLopezbarb: I think we should be using `dev.timeout.resources`. Default value is 120 seconds. I don't know if that works for you but if we use it it's already configurable in the dev section

pkg/k8s/apps/interface.go
AgustinRamiroDiaz: What does "clone" mean in this context? Is it the `DevClone` what we are getting? 

If that's the case, WDYT about renaming to `GetDevClone()`?

pkg/k8s/apps/statefulsets.go
```diff
@@ -192,3 +192,13 @@ func (i *StatefulSetApp) PatchAnnotations(ctx context.Context, c kubernetes.Inte
 func (i *StatefulSetApp) Destroy(ctx context.Context, c kubernetes.Interface) error {
 	return statefulsets.Destroy(ctx, i.sfs.Name, i.sfs.Namespace, c)
 }
+
+// GetCloned Returns from Kubernetes the cloned statefulset
+func (i *StatefulSetApp) GetCloned(ctx context.Context, c kubernetes.Interface) (App, error) {
+	clonedName := fmt.Sprintf("%s-%s", i.sfs.Name, "okteto")
```
AgustinRamiroDiaz, line 198: If I understand correctly, I think we should use `model.DevCloneName` here to not duplicate the logic

AgustinRamiroDiaz (commented):
Changes look good! I left some minor details and some rename stuff.

I have a question regarding the architecture: Is it possible to handle the sleep+wake logic through the cli? I feel that in general this things that are in between are hard to mantain long-term since both the CLI and the backend communicate with the k8s objects

cmd/up/activate.go
```diff
@@ -115,6 +115,11 @@ func (up *upContext) activate() error {
 		lastPodUID = up.Pod.UID
 	}
 
+	if err := up.waitUntilAppIsAwaken(ctx, app); err != nil {
```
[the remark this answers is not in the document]
ifbyol: That code doesn't start any actual synchronization as the dev container is not still created. The idea is to "block" waiting for it until the last moment before the dev mode starts, so hopefully, it is not needed to wait too much because the resource is already awaken when we hit this code

pkg/k8s/apps/interface.go
ifbyol: Clone means the dev container. The one we clone from the original one.

`DevClone` creates the resource in memory, but not in k8s. `GetCloned` would retrieve that once it is in k8s.

If you think is clearer, I can change it

cmd/up/up.go
ifbyol: but the code within this `if` is calling to Okteto API (`wakeSpace` mutation), so if the context is not an Okteto one, it will fail. We need this check

cmd/up/activate.go
ifbyol: yeah makes sense, I would set 2 seconds for now

pkg/k8s/apps/statefulsets.go
ifbyol: yes, we can reuse that

pkg/k8s/apps/deployments.go
ifbyol: Yes, I can, but at the end is a simple wrapper of `statefulsets.Get`. Anyways, I'll add them

cmd/up/up.go
ifbyol: Yes, changed to use that one

ifbyol:
> Changes look good! I left some minor details and some rename stuff.
> 
> I have a question regarding the architecture: Is it possible to handle the sleep+wake logic through the cli? I feel that in general this things that are in between are hard to mantain long-term since both the CLI and the backend communicate with the k8s objects

The problem with that is that the backend (not only the API, also the webhook) would be completely coupled to the CLI as they would need the CLI to execute all the logic for wake + sleep.

At the end is like the `up` command and the cloned devs. In order to display correctly in the UI that some deployments/statefulsets are in dev mode, we use "k8s objects" to represent that.

cmd/up/activate.go
ifbyol: Not sure if 120 seconds would be enough. When things starts to be waken up in order it might take a bit more (taking also into account the time for the job to be scheduled)

cmd/up/activate.go
```diff
@@ -457,3 +462,51 @@ func getPullingMessage(message, namespace string) string {
...
+	defer to.Stop()
+	defer ticker.Stop()
+	oktetoLog.Spinner(fmt.Sprintf("Dev environment '%s' is sleeping. Waiting for it to wake up...", appToCheck.ObjectMeta().Name))
+	oktetoLog.StartSpinner()
+	defer oktetoLog.StopSpinner()
+	for {
+		select {
+		case <-to.C:
+			// In case of timeout, we just print a warning to avoid the command to fail
+			oktetoLog.Warning("Dev environment '%s' didn't wake up after %s", appToCheck.ObjectMeta().Name, timeout.String())
```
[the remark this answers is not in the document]
pchico83: The message looks good to me. @rlamana ?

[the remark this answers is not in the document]
rlamana: It looks good to me too 👌

jLopezbarb (approved):

AgustinRamiroDiaz (approved):

cmd/up/up.go
AgustinRamiroDiaz: @ifbyol what `if` are you referring to? What I mean is that `wakeNamespaceIfApplies` won't call `wakeSpace` since it will return early in the label check

ifbyol: Yeah, but it would slow down things in vanilla clusters as we would need to retrieve the ns to check that. The idea is to execute only that logic if it is needed (in an Okteto context) and execute it within a goroutine to avoid to impact on the `okteto up` command.

I think it is clearer to throw a goroutine only if the context is an Okteto one (having available that information), that executing a goroutine with the logic to execute the wait within it or not

PR #3374, opened 2023-02-03
fix: omit auth error on build

### LINKED ISSUE #3353: Error building image - `docker-credential-desktop resolves to executable in current directory`
**Describe the bug**
Whenever I try to build an image, okteto throws the following error
```
 x  Error building service 'frontend': error building image 'xxxxx/xxxxx:xxxx': build failed: failed to solve: error getting credentials - err: docker-credential-desktop resolves to executable in current directory (./docker-credential-desktop), out: ``
```

**To Reproduce**

1. Change your $PATH to `/usr/bin/:/bin/bash:/bin`
1. Check that your `~/.docker/config.json` has the field `credsStore`
1. Try to build an image
1. See error

**Expected behavior**
It shouldn't fail. If we detect `credsStore` we should be renaming it to `credStore` since it's the correct way of naming it.

Loaded here: https://github.com/okteto/okteto/blob/master/pkg/cmd/build/authprovider.go#L35

- Version 2.11.1

jLopezbarb:

# Proposed changes

Fixes #3353

If the user has an corrupted credentials store, we should not login the user into the registry but we should not fail either.
If the user needs authentication permission to create the image it will fail later.

### DIFF
--- a/pkg/cmd/build/authprovider.go
+++ b/pkg/cmd/build/authprovider.go
@@ -23,6 +23,7 @@ import (
 	"github.com/docker/cli/cli/config/types"
 	"github.com/moby/buildkit/session"
 	"github.com/moby/buildkit/session/auth"
+	oktetoLog "github.com/okteto/okteto/pkg/log"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
@@ -84,7 +85,8 @@ func (ap *authProvider) Credentials(_ context.Context, req *auth.CredentialsRequ
 	}
 	ac, err := ap.config.GetAuthConfig(req.Host)
 	if err != nil {
-		return nil, err
+		oktetoLog.Infof("could not access %s defined in %s", ap.config.CredentialsStore, ap.config.Filename)
+		return res, nil
 	}
 	if ac.IdentityToken != "" {
 		res.Secret = ac.IdentityToken
--- a/pkg/cmd/build/authprovider.go
+++ b/pkg/cmd/build/authprovider.go
@@ -16,6 +16,7 @@ package build
 import (
 	"context"
 	"io"
+	"strings"
 	"sync"
 
 	"github.com/docker/cli/cli/config"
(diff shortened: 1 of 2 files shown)


pkg/cmd/build/authprovider.go
```diff
@@ -84,7 +85,8 @@ func (ap *authProvider) Credentials(_ context.Context, req *auth.CredentialsRequ
 	}
 	ac, err := ap.config.GetAuthConfig(req.Host)
 	if err != nil {
```
ifbyol, line 87: Should we filter for a specific error or we just ignore any error?

AdrianPedriza: I imagine that any error that may occur when obtaining the credential store would be reflected later, as well as the proposed solution for the described error

AdrianPedriza (approved):

pkg/cmd/build/authprovider.go
```diff
@@ -84,6 +86,11 @@ func (ap *authProvider) Credentials(_ context.Context, req *auth.CredentialsRequ
...
+		if strings.HasPrefix(err.Error(), "error getting credentials") {
```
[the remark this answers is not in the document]
AdrianPedriza: I don't know if that will work in every case. Following the issue description I don't have same output so we wouldn't catch the error with it. This is the error I got:
Captura de pantalla 2023-02-10 a las 13 44 59
Anything I'm missing?

pkg/cmd/build/authprovider.go
```diff
@@ -84,6 +86,11 @@ func (ap *authProvider) Credentials(_ context.Context, req *auth.CredentialsRequ
 	}
 	ac, err := ap.config.GetAuthConfig(req.Host)
 	if err != nil {
+		if isErrCredentialsHelperNotAccessible(err) {
```
rberrelleza, line 89: In which circumstances will there be a reference to the helper but the helper is missing?

pkg/cmd/build/authprovider.go
```diff
@@ -94,3 +101,20 @@ func (ap *authProvider) Credentials(_ context.Context, req *auth.CredentialsRequ
 	}
 	return res, nil
 }
+
+func isErrCredentialsHelperNotAccessible(err error) bool {
```
rberrelleza, line 105: This is taking a dependency on the error message that might change. Can we add a unit test where we trigger the auth error to guard us against string changes?

AdrianPedriza: done [here](https://github.com/okteto/okteto/blob/jlopezbarb/omit_auth_error/pkg/cmd/build/authprovider_test.go#L10)

AdrianPedriza: added -> https://github.com/okteto/okteto/pull/3374/files#diff-1fb806e1dce09810a22590fe2d06868d1c0ac91f098506d101ebf73fb5f5d357R51

PR #3375, opened 2023-02-06
Don't delete volume snapshot on DestroyWithLabel if -v flag is not set

ifbyol:

# Proposed changes

Don't include volume snapshots when the flag `-v` is not set. Volume snapshots don't include `deployed-by` label, so this function won't delete it but it could happen that volume snapshots start having that label and we should make sure they are not deleted if `-v` flag is not set

### DIFF
--- a/pkg/k8s/namespaces/namespace.go
+++ b/pkg/k8s/namespaces/namespace.go
@@ -43,6 +43,7 @@ import (
 const (
 	parallelism              int64 = 10
 	volumeKind                     = "PersistentVolumeClaim"
+	volumeSnapshotKind             = "VolumeSnapshot"
 	jobKind                        = "Job"
 	resourcePolicyAnnotation       = "dev.okteto.com/policy"
 	keepPolicy                     = "keep"
@@ -146,7 +147,7 @@ func (n *Namespaces) DestroyWithLabel(ctx context.Context, ns string, opts Delet
 			return err
 		}
 		gvk := obj.GetObjectKind().GroupVersionKind()
-		if gvk.Kind == volumeKind && !opts.IncludeVolumes {
+		if isStorage(gvk.Kind) && !opts.IncludeVolumes {
 			oktetoLog.Debugf("skipping deletion of pvc '%s' because of volume flag", m.GetName())
 			return nil
 		}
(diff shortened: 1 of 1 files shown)


pchico83 (approved):

pkg/k8s/namespaces/namespace.go
```diff
@@ -146,7 +147,7 @@ func (n *Namespaces) DestroyWithLabel(ctx context.Context, ns string, opts Delet
 			return err
 		}
 		gvk := obj.GetObjectKind().GroupVersionKind()
-		if gvk.Kind == volumeKind && !opts.IncludeVolumes {
+		if isStorage(gvk.Kind) && !opts.IncludeVolumes {
 			oktetoLog.Debugf("skipping deletion of pvc '%s' because of volume flag", m.GetName())
```
AdrianPedriza, line 151: I think we should update the log message to include `ValumeSnapshot` as a possible kind to skip. Something like: `skipping deletion of %s '%s' because of volume flag, gvk.Kind, m.GetName()`

teresaromero: +1 to updating the log

teresaromero (approved):

teresaromero:
also, could we fix Deepsource? i think is a small change, although you did not change that function on the PR is at the same file 🤷🏻‍♀️ 

```
pvcNames := []string{}
```

jLopezbarb (approved):

pkg/k8s/namespaces/namespace.go
```diff
@@ -349,3 +350,14 @@ func (t *Trip) listAll(ctx context.Context, traveler Traveler, client dynamic.Re
 		return nil
 	})
 }
+
+// isStorage returns if the kind is some of storage kind
+func isStorage(kind string) bool {
+	switch kind {
```
jLopezbarb, line 356: NIT: Have you thought about using the switch case with multiple values?
```suggestion
	switch kind {
	case volumeKind, volumeSnapshotKind:
	    return true
	}
```

ifbyol: I though about it, in cases where there aren't too many branches I prefer to have them explicitly split them individually, but not strong opinion, I'll change it

pkg/k8s/namespaces/namespace.go
ifbyol: Done!

AdrianPedriza (approved):

PR #3379, opened 2023-02-10
add destroy in remote

AdrianPedriza:
# Proposed changes

Fixes # (issue)

- added `destroy` command working in remote if `--remote` flag is added or `destroy.Image` is set
- only run in remote if `--remote` flag or `deploy/destroy.image` are set in manifest

### DIFF
--- a/cmd/deploy/command_configuration.go
+++ b/cmd/deploy/command_configuration.go
@@ -21,6 +21,7 @@ import (
 
 	"github.com/okteto/okteto/cmd/utils"
 	"github.com/okteto/okteto/pkg/cmd/stack"
+	"github.com/okteto/okteto/pkg/constants"
 	oktetoLog "github.com/okteto/okteto/pkg/log"
 	"github.com/okteto/okteto/pkg/model"
 	"github.com/okteto/okteto/pkg/okteto"
@@ -35,6 +36,14 @@ const (
 )
 
 func setDeployOptionsValuesFromManifest(ctx context.Context, deployOptions *Options, cwd string, c kubernetes.Interface) error {
+	if deployOptions.RunInRemote {
+		if deployOptions.Manifest.Deploy != nil {
+			if deployOptions.Manifest.Deploy.Image == "" {
+				deployOptions.Manifest.Deploy.Image = constants.OktetoPipelineRunnerImage
+			}
+		}
+	}
+
 	if deployOptions.Manifest.Context == "" {
 		deployOptions.Manifest.Context = okteto.Context().Name
 	}
--- a/cmd/destroy/all.go
+++ b/cmd/destroy/all.go
@@ -1,0 +1,148 @@
+package destroy
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"os"
+	"os/signal"
+	"sync"
+	"time"
+
+	"github.com/okteto/okteto/cmd/utils/executor"
+	oktetoErrors "github.com/okteto/okteto/pkg/errors"
+	"github.com/okteto/okteto/pkg/k8s/secrets"
+	oktetoLog "github.com/okteto/okteto/pkg/log"
+	"github.com/okteto/okteto/pkg/okteto"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+type localDestroyAllCommand struct {
+	configMapHandler  configMapHandler
+	nsDestroyer       destroyer
+	executor          executor.ManifestExecutor
+	oktetoClient      *okteto.OktetoClient
+	secrets           secretHandler
+	k8sClientProvider okteto.K8sClientProvider
+}
+
+func newLocalDestroyerAll(
+	k8sClient *kubernetes.Clientset,
+	k8sClientProvider okteto.K8sClientProvider,
+	executor executor.ManifestExecutor,
+	nsDestroyer destroyer,
+	oktetoClient *okteto.OktetoClient,
+) *localDestroyAllCommand {
+	return &localDestroyAllCommand{
+		configMapHandler:  newConfigmapHandler(k8sClient),
+		secrets:           secrets.NewSecrets(k8sClient),
+		k8sClientProvider: k8sClientProvider,
+		oktetoClient:      oktetoClient,
+		nsDestroyer:       nsDestroyer,
+		executor:          executor,
+	}
+}
+
+func (dc *localDestroyAllCommand) destroy(ctx context.Context, opts *Options) error {
+	oktetoLog.Spinner(fmt.Sprintf("Deleting all in %s namespace", opts.Namespace))
+	oktetoLog.StartSpinner()
+	defer oktetoLog.StopSpinner()
+
+	if err := dc.oktetoClient.Namespaces().DestroyAll(ctx, opts.Namespace, opts.DestroyVolumes); err != nil {
+		return err
+	}
+
+	waitCtx, ctxCancel := context.WithCancel(ctx)
+	defer ctxCancel()
+
+	stop := make(chan os.Signal, 1)
+	defer close(stop)
+
+	signal.Notify(stop, os.Interrupt)
+	exit := make(chan error, 1)
+	defer close(exit)
+
+	var wg sync.WaitGroup
+
+	wg.Add(1)
+	go func(wg *sync.WaitGroup) {
+		defer wg.Done()
+		exit <- dc.waitForNamespaceDestroyAllToComplete(waitCtx, opts.Namespace)
+	}(&wg)
+
+	wg.Add(1)
+	go func(wg *sync.WaitGroup) {
+		defer wg.Done()
+		err := dc.oktetoClient.Stream().DestroyAllLogs(waitCtx, opts.Namespace)
+		if err != nil {
+			oktetoLog.Warning("destroy all logs cannot be streamed due to connectivity issues")
+			oktetoLog.Infof("destroy all logs cannot be streamed due to connectivity issues: %v", err)
+		}
+	}(&wg)
+
+	select {
+	case <-stop:
+		ctxCancel()
+		oktetoLog.Infof("CTRL+C received, exit")
+		return oktetoErrors.ErrIntSig
+	case err := <-exit:
+		// wait until streaming logs have finished
+		wg.Wait()
+		return err
+	}
+}
+
+func (ld *localDestroyAllCommand) waitForNamespaceDestroyAllToComplete(ctx context.Context, namespace string) error {
+	timeout := 5 * time.Minute
+	ticker := time.NewTicker(1 * time.Second)
+	to := time.NewTicker(timeout)
+
+	c, _, err := ld.k8sClientProvider.Provide(okteto.Context().Cfg)
+	if err != nil {
+		return err
+	}
+	hasBeenDestroyingAll := false
+
+	for {
+		select {
+		case <-to.C:
+			return fmt.Errorf("'%s' deploy didn't finish after %s", namespace, timeout.String())
+		case <-ticker.C:
+			ns, err := c.CoreV1().Namespaces().Get(ctx, namespace, metav1.GetOptions{})
+			if err != nil {
+				return err
+			}
+
+			status, ok := ns.Labels["space.okteto.com/status"]
+			if !ok {
+				// when status label is not present, continue polling the namespace until timeout
+				oktetoLog.Debugf("namespace %q does not have label for status", namespace)
+				continue
+			}
+
+			switch status {
+			case "Active":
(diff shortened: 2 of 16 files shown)


AdrianPedriza:
converted to draft due to I need to forward destroy commands to remote machine

AdrianPedriza:
`okteto destroy --all` does not work because that command was fixed in master and we need to update with it. Because much of the command for the remote deploy/destroy feature has been refactored I didn't want to include the changes from the update with master in this PR. It will be necessary to update the `feature/remote-deploy` branch with the latest changes that fix the destroy --all fix.

cmd/destroy/destroy.go
```diff
@@ -189,23 +182,21 @@ func Destroy(ctx context.Context) *cobra.Command {
...
-				}
-				err = c.runDestroyAll(ctx, options)
-				if err == nil {
-					oktetoLog.Success("All resources at namespace '%s' where successfully destroyed", options.Namespace)
-				}
-			} else {
-				err = c.runDestroy(ctx, options)
-				if err == nil {
-					oktetoLog.Success("Development environment '%s' successfully destroyed", options.Name)
+			manifest, err := model.GetManifestV2(options.ManifestPath)
```
ifbyol, line 185: This affects how the destroy all works. When the flag `--all` is set, it doesn't affect to any specific manifest, so doing this before the `destroy-all` might generate errors.


In the description you mention something about fixing the destroy all, you mean something related to this?

Maybe we should check the flag `--all` first? That is always executed remotely

cmd/destroy/remote.go
```diff
@@ -0,0 +1,214 @@
...
+
+	tmpl, err := template.New("dockerfile").Parse(dockerfileTemplate)
+	if err != nil {
+		return err
+	}
+
+	dockerfileSyntax := dockerfileTemplateProperties{
+		OktetoCLIImage:   constants.OktetoCLIImageForRemote,
+		UserDestroyImage: rd.manifest.Destroy.Image,
+		//OktetoBuildEnvVars: rd.builder.GetBuildEnvVars(),
```
ifbyol, line 103: Should it be deleted?

cmd/destroy/destroy.go
AdrianPedriza: To know if destroy command needs to be executed remotely or not we need to know previously some values in the manifest (check if destroy image is set) . Condition -> https://github.com/okteto/okteto/blob/adrian/deploy-destroy-same-way/cmd/destroy/destroy.go#L229. The fix I am referring to is to update with master because there is a fix that was uploaded for destroy all. I didn't want to include it in this PR, I think fits better to update later `feature/remote-deploy` branch with master's fix once this PR is merged.

ifbyol: If there is a `--all` flag, it follows other flow. With that flag, the destroy already works remotely as it runs on Okteto backend.

Including the logic to `GetManifest` can be confusing when following the code or even checking the logs and it could be "error prone" in the future if we start to handle the error at some point in the future

AdrianPedriza: yeah, you are right. It can be confusing debugging `destroy --all` if we read the manifest. I will add an exception for this flag 👍

cmd/destroy/destroy.go
```diff
@@ -94,6 +87,7 @@ type destroyCommand struct {
 	k8sClientProvider okteto.K8sClientProvider
 	configMapHandler  configMapHandler
 	oktetoClient      *okteto.OktetoClient
+	getDestroyer      func(*kubernetes.Clientset, okteto.K8sClientProvider, destroyer, executor.ManifestExecutor, *okteto.OktetoClient, *Options) (destroyInterface, error)
```
jLopezbarb, line 90: I think we can infer `kubernetes.Clientset` from `okteto.K8sClientProvider`. This reduces the number of params which is quite large for a function

cmd/destroy/destroy.go
```diff
@@ -218,345 +200,49 @@ func Destroy(ctx context.Context) *cobra.Command {
...
-
-	var wg sync.WaitGroup
-
-	wg.Add(1)
-	go func(wg *sync.WaitGroup) {
-		defer wg.Done()
-		exit <- dc.waitForNamespaceDestroyAllToComplete(waitCtx, opts.Namespace)
-	}(&wg)
+func getDestroyer(k8sClient *kubernetes.Clientset, okK8sClient okteto.K8sClientProvider, nsDestroyer destroyer, executor executor.ManifestExecutor, okClient *okteto.OktetoClient, opts *Options) (destroyInterface, error) {
+	var (
```
jLopezbarb, line 214: I think it's easier to read if we write it as a one-line var
```suggestion
	var deployer destroyInterface
```

cmd/destroy/destroy.go
```diff
@@ -218,345 +200,49 @@ func Destroy(ctx context.Context) *cobra.Command {
...
-		if err != nil {
-			oktetoLog.Warning("destroy all logs cannot be streamed due to connectivity issues")
-			oktetoLog.Infof("destroy all logs cannot be streamed due to connectivity issues: %v", err)
+	if opts.DestroyAll {
+		if !okteto.Context().IsOkteto {
+			return nil, oktetoErrors.ErrContextIsNotOktetoCluster
 		}
-	}(&wg)
+		deployer = newLocalDestroyerAll(k8sClient, okK8sClient, executor, nsDestroyer, okClient)
+		oktetoLog.Info("Destroying locally...")
```
jLopezbarb, line 223:
```suggestion
		oktetoLog.Info("Destroying all...")
```

cmd/destroy/destroy_test.go
```diff
@@ -905,19 +955,22 @@ func TestDestroyWithForceOptionAndFailedCommands(t *testing.T) {
...
 	}
 
-	err = cmd.runDestroy(ctx, opts)
+	err = ld.runDestroy(ctx, opts)
 
 	assert.Error(t, err)
-	assert.ElementsMatch(t, fakeManifest.Destroy, executor.executed)
+	assert.ElementsMatch(t, fakeManifest.Destroy.Commands, executor.executed)
 	assert.True(t, destroyer.destroyed)
 	assert.True(t, destroyer.destroyedVolumes)
```
jLopezbarb, line 976: Can we add a test for `destroy all` and another one for `destroy remote`?

cmd/destroy/local.go
```diff
@@ -0,0 +1,415 @@
...
+		k8sClientProvider: k8sClientProvider,
+		oktetoClient:      oktetoClient,
+		nsDestroyer:       nsDestroyer,
+		executor:          executor,
+	}
+}
+
+func (ld *localDestroyCommand) destroy(ctx context.Context, opts *Options) error {
+
+	if opts.RunInRemote {
```
jLopezbarb, line 75:
```suggestion
	if opts.RunInRemote && ld.manifest.Deploy != nil && ld.manifest.Deploy.Image == "" {
```

jLopezbarb: If manifest.Deploy is not a pointer we could `ld.manifest.Deploy.GetImage` without the need of checking if it's nil

cmd/destroy/local.go
```diff
@@ -0,0 +1,415 @@
...
+	err := ld.runDestroy(ctx, opts)
+	if err == nil {
+		oktetoLog.Success("Development environment '%s' successfully destroyed", opts.Name)
+	}
+	analytics.TrackDestroy(err == nil, opts.DestroyAll)
+
+	return err
+}
+
+func (dc *localDestroyAllCommand) destroy(ctx context.Context, opts *Options) error {
```
jLopezbarb, line 92: NIT: Should we move this to another file?

cmd/destroy/remote.go
```diff
@@ -0,0 +1,213 @@
...
+ENV {{ .TokenEnvVar }} {{ .TokenValue }}
+ENV {{ .RemoteDeployEnvVar }} true
+
+COPY . /okteto/src
+WORKDIR /okteto/src
+
+ENV OKTETO_INVALIDATE_CACHE {{ .RandomInt }}
+RUN okteto destroy {{ .DestroyFlags }}
+`
+	dockerignoreName = "deploy.dockerignore"
```
jLopezbarb, line 52: Can we update to use the same `dockerignore` as deploy?

cmd/destroy/remote.go
```diff
@@ -0,0 +1,213 @@
...
+
+	buildOptions := build.OptsFromBuildInfo("", "", buildInfo, &types.BuildOptions{Path: cwd, OutputMode: buildOutput})
+	buildOptions.Tag = ""
+
+	// we need to call Build() method using a remote builder. This Builder will have
+	// the same behavior as the V1 builder but with a different output taking into
+	// account that we must not confuse the user with build messages since this logic is
+	// executed in the deploy command.
+	if err := remoteBuild.NewBuilderFromScratch().Build(ctx, buildOptions); err != nil {
+		fmt.Println(err)
```
jLopezbarb, line 153: Is this a debug log?

pkg/cmd/build/logger.go
```diff
@@ -130,6 +130,11 @@ func (t *trace) display() {
 			}
 		}
 		if t.hasCommandLogs(v) {
+			oktetoLog.StopSpinner()
+			logs := strings.Join(v.logs, "")
+			if !strings.HasPrefix(logs, "{") {
```
jLopezbarb, line 135: Maybe we should add `--log-level` in the destroy command. That way it'd be already contemplated in lines 140-160

pkg/model/serializer.go
```diff
@@ -768,7 +767,7 @@ type manifestRaw struct {
 	Icon          string                                   `json:"icon,omitempty" yaml:"icon,omitempty"`
 	Deploy        *DeployInfo                              `json:"deploy,omitempty" yaml:"deploy,omitempty"`
 	Dev           ManifestDevs                             `json:"dev,omitempty" yaml:"dev,omitempty"`
-	Destroy       []DeployCommand                          `json:"destroy,omitempty" yaml:"destroy,omitempty"`
+	Destroy       *DestroyInfo                             `json:"destroy,omitempty" yaml:"destroy,omitempty"`
```
jLopezbarb, line 770: Do we need it to be a pointer? I think we are only getting information and not modifying it.

Also, Can we add a unit test for this?

cmd/destroy/remote.go
```diff
@@ -0,0 +1,213 @@
...
+
+	if opts.DestroyVolumes {
+		deployFlags = append(deployFlags, "--volumes")
+	}
+
+	if opts.ForceDestroy {
+		deployFlags = append(deployFlags, "--force-destroy")
+	}
+
+	if opts.DestroyAll {
```
AdrianPedriza, line 208: to remove: this opt cannot be added to remote

cmd/destroy/remote.go
AdrianPedriza: yes, it has been uptaded after merge https://github.com/okteto/okteto/pull/3393 into this branch

cmd/destroy/local.go
AdrianPedriza: I will!

cmd/destroy/remote.go
AdrianPedriza: it is! I will remove it

pkg/model/serializer.go
AdrianPedriza: We are using it as a pointer following `DeployIngo` approach. Anyway, we are calling `manifest.ExpandEnvVars()` wich modifies `deploy.image`

cmd/destroy/destroy_test.go
AdrianPedriza: We need to fix `destroy all` with the fix in master so I will work on test `destroy remote`

cmd/destroy/all.go
```diff
@@ -28,7 +28,7 @@ type localDestroyAllCommand struct {
 }
 
 func newLocalDestroyerAll(
-	k8sClient *kubernetes.Clientset,
+	k8sClient kubernetes.Interface,
```
jLopezbarb, line 31: Why do we need this we can create it with the provider. Injecting it can cause throttling problems, that's why we should always use the provider

cmd/deploy/command_configuration.go
```diff
@@ -35,6 +36,14 @@ const (
 )
 
 func setDeployOptionsValuesFromManifest(ctx context.Context, deployOptions *Options, cwd string, c kubernetes.Interface) error {
+	if deployOptions.RunInRemote {
```
jLopezbarb, line 39: Can we cover these new lines in the unit test?

AdrianPedriza: discussed that for now we do not introduce test on it, testing that function is out of the scope of the PR.

jLopezbarb (approved):

PR #3380, opened 2023-02-11
remove word 'context' from success log

### LINKED ISSUE #3327: Wrong info message when selecting okteto namespace
**Describe the bug**
When running: `okteto namespace use <MY-NAMESPACE>` to select a namespace, the namespace is selected successfully, but the info message says: 'Using context <MY-NAMESPACE> @ cloud.okteto.com'. Instead, the message should say: 'Using namespace <MY-NAMESPACE> @ cloud.okteto.com'

**To Reproduce**

1. Install Okteto
2. Run: 'okteto context'
3. Run: 'okteto namespace use ANY-AVAILABLE-NAMESPACE'
4. or Run: 'okteto namespace ANY-AVAILABLE-NAMESPACE'

**Expected behavior**
When selecting namespace, the info message should say 'Using namespace' instead of 'Using context'

- OS: Debian (devcontainer) 
- Version 11

**Additional context**
vscode ➜ /workspaces/myproject(development ✗) $ okteto namespace use development-MY-NAMESPACE
 !  Initializing context with the value of OKTETO_TOKEN environment variable
 ✓  Using context development-MY-NAMESPACE @ cloud.okteto.com

miniscruff:
# Proposed changes
Fixes #3327

Because the success log already outputs both the namespace and context with the format of `<ns> @ <ctx>`, the word context here does not really add anything as it sort of duplicates itself later. Plus, it can be misleading if you ran a namespace command.


cmd/context/create.go
```diff
@@ -201,7 +201,7 @@ func (c *ContextCommand) UseContext(ctx context.Context, ctxOptions *ContextOpti
 	}
 
 	if ctxOptions.IsCtxCommand {
-		oktetoLog.Success("Using context %s @ %s", okteto.Context().Namespace, okteto.RemoveSchema(ctxStore.CurrentContext))
+		oktetoLog.Success("Using %s @ %s", okteto.Context().Namespace, okteto.RemoveSchema(ctxStore.CurrentContext))
```
AdrianPedriza, line 204: I think this string is used in more places. Can you create a constant and use it in every place we use this string message?

miniscruff: I know there is this usage here: https://github.com/okteto/okteto/blob/ee65cf263db00e84bc347f8095742b614641577a/cmd/context/use.go#L137, which looks quite similar but not identical. Maybe we could merge the two into one `if isCtxCommand || show` check? Would have to find other uses of `UseContext` though. Not sure.

jLopezbarb: Let's keep them with different strings for now since one is used for changing the namespace and the other one when changing context.

jLopezbarb (approved):

AdrianPedriza (approved):

jLopezbarb:
@miniscruff Thanks for your contribution. Really nice addition!! 🚀 😄

PR #3384, opened 2023-02-16
Support for divert with Istio

pchico83:
This PR adds support for "divert" using Istio as a backend for redirecting requests based on headers.

The logic when `divert` is enabled in the okteto manifest is like this:

- Query all the virtual services in the namespace we are diverting
- For the virtual services redirecting to `divert.service`, add rules to redirect to the developer namespace if the `x-okteto-divert` header exists
- For the virtual services matching `divert.virtualServices`, replicate the virtual service in the developer namespace using their own hosts and injecting the `x-okteto-divert` header

It also fixes https://github.com/okteto/app/issues/4509 because I needed the `OKTETO_DOMAIN` envvar the [sample](https://github.com/okteto/divert-with-istio-sample) I am using for divert. This sample can be used to test the istio integration with divert.


cmd/destroy/destroy.go
```diff
@@ -290,8 +291,8 @@ func (dc *destroyCommand) runDestroy(ctx context.Context, opts *Options) error {
 	if manifest.Context == "" {
 		manifest.Context = okteto.Context().Name
 	}
-	if manifest.Namespace == okteto.Context().Namespace {
-		manifest.Namespace = okteto.Context().Namespace
+	if manifest.Namespace == "" {
```
pchico83, line 294: this was a bug affecting the divert destroy logic

pkg/cmd/stack/deploy.go
```diff
@@ -586,8 +589,7 @@ func deployK8sService(ctx context.Context, svcName string, s *model.Stack, c kub
 	}
 
 	svcK8s.ObjectMeta.ResourceVersion = old.ObjectMeta.ResourceVersion
-	driver := divert.New(nil, nil, nil)
-	driver.ApplyToService(svcK8s, old)
+	keepDivertInService(svcK8s, old)
```
pchico83, line 592: I moved this logic to the weaver driver in [this PR](https://github.com/okteto/okteto/pull/3288) but I moved it back to the stack module because it was complicating the divert driver interface. This is an optimization we do for stacks, and we will probably remove the weaver driver soon, so I think it's better to move it back here and keep the divert driver interface simpler

pkg/k8s/virtualservices/client.go
```diff
@@ -0,0 +1,32 @@
...
+import (
+	"github.com/okteto/okteto/pkg/okteto"
+	istioclientset "istio.io/client-go/pkg/clientset/versioned"
+	"k8s.io/client-go/tools/clientcmd"
+)
+
+// GetIstioClient returns a client for istio
+func GetIstioClient() (*istioclientset.Clientset, error) {
+	octx := okteto.Context()
+	clientConfig := clientcmd.NewDefaultClientConfig(*octx.Cfg, nil)
```
jLopezbarb, line 25: NIT: You can get the `restConfig` directly using the `k8sProvider`. This will make easier to test later all these functions

```suggestion
	client, config, err := okteto.NewK8sClientProvider().Provide(okteto.Context().Cfg)
```

pkg/divert/istio/divert.go
```diff
@@ -0,0 +1,222 @@
...
+		}
+		if name == vs {
+			return true
+		}
+	}
+	return false
+}
+
+func isEqualVirtualService(vs1 *istioV1beta1.VirtualService, vs2 *istioV1beta1.VirtualService) bool {
+	return reflect.DeepEqual(vs1.Spec, vs2.Spec)
```
jLopezbarb, line 221: NIT: It's not recommended to use reflect because an empty object and a nil object are the same. If we are ok with that I wouldn't change it

pkg/divert/istio/divert.go
```diff
@@ -0,0 +1,222 @@
...
+			return err
+		}
+		retries++
+	}
+	return err
+}
+
+func isInDivertVirtualServices(name string, virtualServices []string) bool {
+	for _, vs := range virtualServices {
+		if vs == "*" {
```
jLopezbarb, line 210: NIT: I think it's easier to read it with a switch case 
```suggestion
switch vs {
case "*", name:
    return true
 }
```

pkg/divert/istio/divert.go
```diff
@@ -0,0 +1,222 @@
...
+	for i := range vsList {
+		cache.developerVirtualServices[vsList[i].Name] = vsList[i]
+	}
+	return cache, nil
+}
+
+func (d *Driver) retryTranslateDeveloperVirtualService(ctx context.Context, name string) error {
+	retries := 0
+	var err error
+	for retries < 10 {
```
jLopezbarb, line 128: Can we move the magic number to constants at the top of the file. We are using it also [in line 167](https://github.com/okteto/okteto/pull/3384/files#diff-0e4a458af0d49f8538599e45205648929cbd76be5470d3fabf89cb47756c3581R167)

pkg/divert/istio/divert.go
```diff
@@ -0,0 +1,222 @@
...
+}
+
+func (d *Driver) ApplyNamespaceToProxy() string {
+	if d.Manifest.Deploy.Divert.Namespace == d.Manifest.Namespace {
+		return ""
+	}
+	return d.Manifest.Deploy.Divert.Namespace
+}
+
+func initDivertIstioCache(ctx context.Context, m *model.Manifest, c istioclientset.Interface) (*divertIstioCache, error) {
```
jLopezbarb, line 102: NIT: Should this function be part of the `istioDriver`? I think it's better to have every function related to the struct with a receiver. WDYT?

pkg/divert/driver.go
```diff
@@ -15,26 +15,48 @@ package divert
...
+		return &weaver.Driver{
+			Client:       c,
+			DivertClient: dc,
+			Manifest:     m,
+		}, nil
+	}
+
+	ic, err := virtualservices.GetIstioClient()
+	if err != nil {
+		return nil, fmt.Errorf("error creating istio client: %s", err.Error())
```
jLopezbarb, line 55: Wrapping the error let us to use `Error.Is` interface to find in the stack an specific error. 
```suggestion
		return nil, fmt.Errorf("error creating istio client: %w", err)
```

pkg/divert/driver.go
```diff
@@ -15,26 +15,48 @@ package divert
...
+	"github.com/okteto/okteto/pkg/okteto"
 	"k8s.io/client-go/kubernetes"
 )
 
 type Driver interface {
 	Deploy(ctx context.Context) error
 	Destroy(ctx context.Context) error
-	ApplyToDeployment(d1 *appsv1.Deployment, d2 *appsv1.Deployment)
-	ApplyToService(s1 *apiv1.Service, s2 *apiv1.Service)
+	ApplyNamespaceToProxy() string
```
jLopezbarb, line 33: Probably this name doesn't correspond anymore and should be renamed to `GetDivertNamespace`. Am I wrong?

cmd/destroy/destroy.go
```diff
@@ -331,6 +332,24 @@ func (dc *destroyCommand) runDestroy(ctx context.Context, opts *Options) error {
 	signal.Notify(stop, os.Interrupt)
 	exit := make(chan error, 1)
 
+	if manifest.Deploy != nil && manifest.Deploy.Divert != nil && manifest.Deploy.Divert.Namespace != manifest.Namespace {
```
jLopezbarb, line 335: Should we encapsulate this into a `destroyDivert` function?

ifbyol (approved):
One comment regarding the whole manifest within the dirver struct, but in general looks good! Approving the PR but I think we should try to avoid to pass the whole Manifest nad just the divert information needed

pkg/divert/istio/divert.go
```diff
@@ -0,0 +1,212 @@
...
+		divertVirtualServices:    map[string]*istioV1beta1.VirtualService{},
+		developerVirtualServices: map[string]*istioV1beta1.VirtualService{},
+	}
+	vsList, err := virtualservices.List(ctx, d.Manifest.Deploy.Divert.Namespace, d.IstioClient)
+	if err != nil {
+		return err
+	}
+	for i := range vsList {
+		d.cache.divertVirtualServices[vsList[i].Name] = vsList[i]
+	}
```
AgustinRamiroDiaz, line 105:
```suggestion
	for _, vs := range vsList {
		d.cache.divertVirtualServices[vs.Name] = vs
	}
```

pkg/divert/istio/divert.go
```diff
@@ -0,0 +1,212 @@
...
+
+	vsList, err = virtualservices.List(ctx, d.Manifest.Namespace, d.IstioClient)
```
AgustinRamiroDiaz, line 107: could we use a different variable here in order to not mix them up? something like `devVsList`

pkg/divert/istio/divert.go
```diff
@@ -0,0 +1,212 @@
...
+func (d *Driver) retryTranslateDeveloperVirtualService(ctx context.Context, name string) error {
+	retries := 0
+	var err error
+	for retries < UPDATE_CONFLICT_RETRIES {
+		if _, ok := d.cache.developerVirtualServices[name]; !ok {
+			vs := createIntoDeveloperVirtualService(d.Manifest, d.cache.divertVirtualServices[name])
+			err = virtualservices.Create(ctx, vs, d.IstioClient)
+			if err == nil {
+				d.cache.developerVirtualServices[name] = vs
+				return nil
```
AgustinRamiroDiaz, line 126: As it is now, if the VS already exists we are not adding it to the cache. From what I understand we should do it and we could merge it with the next `return nil`

pkg/divert/istio/divert.go
```diff
@@ -0,0 +1,212 @@
...
+	Client      kubernetes.Interface
+	IstioClient istioclientset.Interface
+	Manifest    *model.Manifest
+	cache       *divertIstioCache
+}
+
+type divertIstioCache struct {
+	divertVirtualServices    map[string]*istioV1beta1.VirtualService
+	developerVirtualServices map[string]*istioV1beta1.VirtualService
+}
```
AgustinRamiroDiaz, line 45: Can we rename this to
```suggestion
type virtualServicesCache struct {
	divert    map[string]*istioV1beta1.VirtualService
	developer map[string]*istioV1beta1.VirtualService
}
```
? The most important distinction along the code is if it's a `divert` or a `developer` VS and it's hard to differiantiate if they have both the `VirtualSevices` suffix

pkg/divert/istio/divert.go
```diff
@@ -0,0 +1,212 @@
...
+		if !k8sErrors.IsConflict(err) {
+			return err
+		}
+		retries++
+	}
+	return err
+}
+
+func (d *Driver) retryTranslateDivertVirtualService(ctx context.Context, name string) error {
+	retries := 0
```
AgustinRamiroDiaz, line 157: Nit: we could add the `retries` variable to the for loop so we can remove the `retries++`

pkg/divert/istio/virtualservices.go
```diff
@@ -0,0 +1,127 @@
...
+			httpRoute.Match[j].Headers[model.OktetoDivertHeader] = &istioNetworkingV1beta1.StringMatch{
+				MatchType: &istioNetworkingV1beta1.StringMatch_Exact{Exact: m.Namespace},
+			}
+		}
+		matchService := false
+		for j := range httpRoute.Route {
+			parts := strings.Split(httpRoute.Route[j].Destination.Host, ".")
+			if parts[0] == m.Deploy.Divert.Service {
+				httpRoute.Route[j].Destination.Host = fmt.Sprintf("%s.%s.svc.cluster.local", parts[0], m.Namespace)
+				matchService = true
```
AgustinRamiroDiaz, line 104: Nit: we could add a `break` here to stop looking

AgustinRamiroDiaz (commented):
None of my comments are blockers :+1:

cmd/deploy/deploy.go
```diff
@@ -443,6 +442,8 @@ func (dc *DeployCommand) RunDeploy(ctx context.Context, deployOptions *Options)
 		fmt.Sprintf("%s=true", oktetoLog.OktetoDisableSpinnerEnvVar),
 		// Set OKTETO_NAMESPACE=namespace-name env variable, so all the commandsruns on the same namespace
 		fmt.Sprintf("%s=%s", model.OktetoNamespaceEnvVar, okteto.Context().Namespace),
+		// Set OKTETO_DOMAIN=okteto-subdomain env variable
```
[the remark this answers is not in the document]
pchico83: Good catch! Updated to only add the envvar in okteto clusters

pkg/cmd/stack/deploy.go
pchico83: yes, this is weaver specific. but this logic will be removed soon and it was complicating the divert driver interface

pkg/cmd/stack/deploy.go
```diff
@@ -912,3 +913,39 @@ func getAddedSvcs(initialSvcsToDeploy, svcsToDeployWithDependencies []string) []
...
+	if d2.Spec.Template.Labels[model.OktetoDivertInjectSidecarLabel] == "" {
+		return
+	}
+	if d1.Spec.Template.Labels == nil {
+		d1.Spec.Template.Labels = map[string]string{}
+	}
+	d1.Spec.Template.Labels[model.OktetoDivertInjectSidecarLabel] = d2.Spec.Template.Labels[model.OktetoDivertInjectSidecarLabel]
+}
+
+func keepDivertInService(s1 *apiv1.Service, s2 *apiv1.Service) {
```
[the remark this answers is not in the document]
pchico83: yes, makes sense

pkg/divert/istio/divert.go
```diff
@@ -0,0 +1,222 @@
...
+	istioclientset "istio.io/client-go/pkg/clientset/versioned"
+	k8sErrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/client-go/kubernetes"
+)
+
+// Driver weaver struct for the divert driver
+type Driver struct {
+	Client      kubernetes.Interface
+	IstioClient istioclientset.Interface
+	Manifest    *model.Manifest
```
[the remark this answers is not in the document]
pchico83: @jLopezbarb @ifbyol @AgustinRamiroDiaz updated to pass name, namespace, and the divert struct

pkg/divert/istio/divert.go
```diff
@@ -0,0 +1,212 @@
...
+	UPDATE_CONFLICT_RETRIES = 10
+)
+
+// Driver weaver struct for the divert driver
+type Driver struct {
+	Client      kubernetes.Interface
+	IstioClient istioclientset.Interface
+	Manifest    *model.Manifest
+	cache       *divertIstioCache
+}
```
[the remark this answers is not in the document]
pchico83: You are right, updated!

pkg/divert/istio/divert.go
pchico83: updated!

pkg/divert/istio/divert.go
pchico83: I removed the whole cache logic. It was complicating the logic and not bringing better performance

pkg/divert/istio/divert.go
pchico83: I remove the comparison logic as part of removing the cache logic

pkg/divert/istio/virtualservices.go
pchico83: we need to keep iterating to check/modify all routes

jLopezbarb (approved):

ifbyol (approved):

PR #3390, opened 2023-02-16
trigger a new context store write

### LINKED ISSUE #3387: a new context cannot be initialize exclusively with envs
**Describe the bug**
if I don't have an okteto context store file but have the necessary envs to use a context, I get an error `okteto context store is corrupted. Delete the folder '%s' and try again` because I don't have the okteto context store file. 

**To Reproduce**

1. Remove okteto context file (or rename it)
2. Set `OKTETO_TOKEN` and `OKTETO_URL` envs
3. Execute `okteto deploy` using okteto/movies 

**Expected behavior**
If I have the necessary envs set to use a new context any okteto command that make use of the okteto context should be able to be exec

AdrianPedriza:
# Proposed changes

Fixes #3387

### DIFF
--- a/cmd/context/use.go
+++ b/cmd/context/use.go
@@ -93,6 +93,11 @@ Or a Kubernetes context:
 
 func (c *ContextCommand) Run(ctx context.Context, ctxOptions *ContextOptions) error {
 	ctxStore := okteto.ContextStore()
+	if len(ctxStore.Contexts) == 0 {
+		// if the context store has no context stored, set flag to save
+		// the new one generated.
+		ctxOptions.Save = true
+	}
 
 	// We have to maintain this order to not break some commands
 	// See https://github.com/okteto/okteto/issues/3247 for more information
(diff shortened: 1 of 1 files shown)


cmd/context/use.go
```diff
@@ -93,6 +93,11 @@ Or a Kubernetes context:
 
 func (c *ContextCommand) Run(ctx context.Context, ctxOptions *ContextOptions) error {
 	ctxStore := okteto.ContextStore()
+	if len(ctxStore.Contexts) == 0 {
+		// if the context store has no context stored, set flag to save
```
jLopezbarb, line 97: Can we specify that this is just when the command is not a context one, for example (build deploy up)

jLopezbarb (approved):

cmd/context/use.go
AdrianPedriza: sure!

jLopezbarb (approved):

PR #3393, opened 2023-02-17
remote deploy/destroy improvements

### LINKED ISSUE #3358: Remote deployment - build improvements
I can't run okteto build inside a remote deployment.
I want that my `deploy.dockerignore` to be call `.oktetodeployignore`
I want to set my `deploy.image` from one of the images built in the build step

## Acceptance criteria

- [x] okteto build should be available inside the deploy section
- [x] rename dockerignore for deploy to `.oktetodeployignore`
- [x] `deploy.image` can be an okteto build env var

AdrianPedriza:
# Proposed changes

Fixes #3358

### DIFF
--- a/cmd/deploy/remote.go
+++ b/cmd/deploy/remote.go
@@ -62,7 +62,7 @@ WORKDIR /okteto/src
 ENV OKTETO_INVALIDATE_CACHE {{ .RandomInt }}
 RUN okteto deploy --log-output=json {{ .DeployFlags }}
 `
-	dockerignoreName = "deploy.dockerignore"
+	dockerignoreName = ".oktetodeployignore"
 	buildOutput      = "deploy"
 )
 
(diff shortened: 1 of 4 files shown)


cmd/deploy/remote.go
```diff
@@ -187,7 +187,7 @@ func (rd *remoteDeployCommand) createDockerignoreIfNeeded(cwd, tmpDir string) er
 			return err
 		}
 
-		err = afero.WriteFile(rd.fs, fmt.Sprintf("%s/%s", tmpDir, dockerignoreName), dockerignoreContent, 0600)
+		err = afero.WriteFile(rd.fs, fmt.Sprintf("%s/%s", tmpDir, ".dockerignore"), dockerignoreContent, 0600)
```
ifbyol, line 190: I think this was the only place where we being using `dockerignoreName` constant. Should we delete that cons?

jLopezbarb (approved):

cmd/deploy/remote.go
```diff
@@ -62,7 +62,7 @@ WORKDIR /okteto/src
 ENV OKTETO_INVALIDATE_CACHE {{ .RandomInt }}
 RUN okteto deploy --log-output=json {{ .DeployFlags }}
 `
-	dockerignoreName = "deploy.dockerignore"
+	dockerignoreName = ".oktetodeployignore"
```
ifbyol, line 65: Should we change also the value here? https://github.com/okteto/okteto/blob/adrian/deploy-destroy-same-way/cmd/destroy/remote.go#L52

AdrianPedriza: oh yes, I forgot the change for destroy in remote! Thanks @ifbyol

cmd/deploy/remote.go
AdrianPedriza: It is used just before this line but I think it is not necessary for now to make it constant if it is only used once

jLopezbarb (approved):

PR #3405, opened 2023-02-27
Run okteto init v1 for vanilla clusters

pchico83:
# Proposed changes

This PR does the following things:

- Run `okteto init --v1` automatically for clusters without the server-side component of okteto. This is because `build` and `deploy` don't work well in vanilla clusters and they are going to be deprecated in Okteto CLI 3.0
- Remove unused code from the `cmd/manifest` package
- Set some options as `hidden`

### DIFF
--- a/cmd/init.go
+++ b/cmd/init.go
@@ -18,8 +18,8 @@ import (
 	"os"
 
 	contextCMD "github.com/okteto/okteto/cmd/context"
-	"github.com/okteto/okteto/cmd/manifest"
 	"github.com/okteto/okteto/cmd/utils"
+	initCMD "github.com/okteto/okteto/pkg/cmd/init"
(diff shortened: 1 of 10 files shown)


rberrelleza:
Will this create a v1 manifest? If so, we should look for a different solution, since that format is fully deprecated now right?

jLopezbarb (approved):

jLopezbarb:
> Will this create a v1 manifest? If so, we should look for a different solution, since that format is fully deprecated now right?
> 
> Could we create a v2 manifest with only the dev section?

Yes, this will create a v1 manifest, but I think that should be addressed when we finally remove the support for v1. 

If we end up doing the change in this PR we should encapsulate the dev in a `model.Manifest` struct [here](https://github.com/okteto/okteto/blob/feature/remote-deploy/cmd/manifest/init-v1.go#L159)

PR #3406, opened 2023-02-27
Add samples for running Okteto CLI in clusters without Okteto server-side component

pchico83:
Our current samples in https://www.okteto.com/docs/samples/golang/ only works with the full Okteto platform.
This PR adds a `samples` folder where we can have a set of samples that works in vanilla clusters


jLopezbarb (approved):

README.md
```diff
@@ -33,33 +33,45 @@ The end result is that the remote cluster is seen by your IDE and tools as a loc
...
+- [Ruby](samples/ruby-getting-started/README.md)
 
 ### Releases
 
-Okteto is released into three channels: stable, beta, and dev. By default when okteto is installed the stable channel is used. If you need to access features not yet widely available you can install from the beta or dev channel. More info in the [release documentation](docs/RELEASE.md).
+The Okteto CLI is released into three channels: stable, beta, and dev. By default when okteto is installed the stable channel is used. If you need to access features not yet widely available you can install from the beta or dev channel. More info in the [release documentation](docs/RELEASE.md).
 
 ## Useful links
 
 - [Getting started](https://www.okteto.com/docs/getting-started/)
```
rberrelleza, line 71: The samples we have are for the CLI standalone mode, but this is sending readers to a getting started that is written for the CLI + Cluster mode. Is that what we want?

README.md
```diff
@@ -33,33 +33,45 @@ The end result is that the remote cluster is seen by your IDE and tools as a loc
...
+- Deploy your application on Kubernetes.
+- [Install the Okteto CLI](https://www.okteto.com/docs/getting-started/#installing-okteto-cli)
+- Configure the Okteto CLI to access to your Kubernetes cluster with `okteto context`
+- Run `okteto init` from the root of your application git repository to inspect your code and generate your [Okteto manifest](https://okteto.com/docs/reference/manifest/). The Okteto manifest defines your development container
+- Run `okteto up` to activate your development container
 
-We created a [few guides to help you get started](https://github.com/okteto/samples) with `okteto` and your favorite programming language.
+We created a few guides to help you get started with the Okteto CLI and your favorite programming languages:
```
rberrelleza, line 55: this requires more explanation on cli standalone vs cli + cluster modes, and why we have samples embedded here vs samples in okteto/docs and on okteto/samples.

I find that just having a list of samples here confusing.

rberrelleza (commented):
I'm not sure this is the right place and approach for the samples.

With this, you require people to download the entire source code + understand the folder structure for a user to try a hello-world scenario. Is this easier than "donwload binary + clone small repo"?

Second, I think this makes more confusing the difference between these samples and the ones that need the cluster component. Maybe adding a clear readme entry explaining this difference helps?

README.md
pchico83: I would always send readers to the embedded samples. They will get a CTA to the docs and self-hosted when they try to use a feature not available in the CLI only version. WDYT?

PR #3409, opened 2023-02-28
Okteto kubeconfig will use kubetoken client auth if available on okteto server

AgustinRamiroDiaz:
# Proposed changes

Fixes https://github.com/okteto/app/issues/5794

Added behavior to use exec client auth for connecting to k8s when available in the okteto server

Code organization is not great, I'd love some suggestions


pkg/okteto/context.go
```diff
@@ -343,7 +343,27 @@ func AddOktetoCredentialsToCfg(cfg *clientcmdapi.Config, cred *types.Credential,
...
+			InstallHint:        "Okteto needs to be installed and in your PATH to use this context. Please visit https://www.okteto.com/docs/getting-started/ for more information.",
+			ProvideClusterInfo: true,
+			InteractiveMode:    "IfAvailable",
+		}
+	} else {
+		// fallback for okteto API before client authentication support
+		// TODO: remove once we stop supporting token based authentication
+		user.Token = cred.Token
+		user.Exec = nil
+	}
```
AgustinRamiroDiaz, line 366: All of this logic will simplify a lot once we deprecate token based auth.

There's also the option of filling both `Token` and `Exec` and kubectl will default to `Token` that would also simplify all this logic. I didn't opt for this approach because it has the downside of only using `Exec` when `Token` is not set, hence we'll use this behavior only once we deprecate the `Token`. To get early feedback on this feature and manage possible errors with time (before k8s 1.26) I think this is the best approach

ifbyol: Makes sense to do it in this way

ifbyol, line 366: Maybe this could be extracted to its own function, not sure if it worths it or not but it will probably help to unit test this, which is the most important part of this PR

ifbyol (commented):
> Code organization is not great, I'd love some suggestions

@AgustinRamiroDiaz is there anything special you don't like?

cmd/context/create.go
```diff
@@ -52,6 +54,19 @@ func NewContextCommand() *ContextCommand {
 		LoginController:      login.NewLoginController(),
 		OktetoClientProvider: okteto.NewOktetoClientProvider(),
 		OktetoContextWriter:  okteto.NewContextConfigWriter(),
+		OktetoKubeTokenPresent: func(oktetoURL string) (bool, error) {
```
AdrianPedriza, line 57:
```suggestion
		IsOktetoKubeTokenPresent: func(oktetoURL string) (bool, error) {
```

pkg/okteto/secrets.go
```diff
@@ -79,7 +79,7 @@ func (c *userClient) GetContext(ctx context.Context) (*types.UserContext, error)
 		Cred struct {
 			Server      graphql.String
 			Certificate graphql.String
-			Token       graphql.String
+			Token       graphql.String // TODO: Deprecate once we move to kubetoken auth
```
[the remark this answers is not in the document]
AdrianPedriza: can you link that issue in this TODO?

pkg/types/credential.go
```diff
@@ -17,6 +17,6 @@ package types
 type Credential struct {
 	Server      string `json:"server" yaml:"server"`
 	Certificate string `json:"certificate" yaml:"certificate"`
-	Token       string `json:"token" yaml:"token"`
+	Token       string `json:"token" yaml:"token"` // TODO: Deprecate once we move to kubetoken auth
```
AdrianPedriza, line 20: same here

cmd/context/create.go
```diff
@@ -52,6 +54,19 @@ func NewContextCommand() *ContextCommand {
...
+			if err != nil {
+				return false, fmt.Errorf("failed to parse kubetoken url: %w", err)
+			}
+
+			resp, err := http.Get(kubeTokenURL)
+			if err != nil {
+				return false, fmt.Errorf("failed to get kubetoken url: %w", err)
+			}
+
+			return resp.StatusCode == http.StatusUnauthorized, nil
```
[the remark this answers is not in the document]
AgustinRamiroDiaz: Yeah, we'd get a 404. Maybe it's better from the server's point of view since it wouldn't receive unauthorized requests, but it surely is a balance

pkg/okteto/secrets.go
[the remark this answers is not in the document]
AgustinRamiroDiaz: :+1: https://github.com/okteto/okteto/issues/3410

pkg/types/credential.go
AgustinRamiroDiaz: Sorry, what do you mean here?

AgustinRamiroDiaz: Oh I think you mean adding this https://github.com/okteto/okteto/issues/3410

pkg/okteto/context.go
AgustinRamiroDiaz: Everything was already in place for unit testing. It's a bit verbose but it's very well defined as a table test.

Did manage to find an "odd" behavior with the test: the error was handled at the middle of the function, which could lead to a partial modification of the `cfg` object

AgustinRamiroDiaz:
> > Code organization is not great, I'd love some suggestions
> 
> @AgustinRamiroDiaz is there anything special you don't like?

@ifbyol I feel like `IsOktetoKubeTokenPresent` is a bit weird there and it might be getting to specific while the `ContextCommand` handles multiple things related

ifbyol:
> > > Code organization is not great, I'd love some suggestions
> > 
> > 
> > @AgustinRamiroDiaz is there anything special you don't like?
> 
> @ifbyol I feel like `IsOktetoKubeTokenPresent` is a bit weird there and it might be getting to specific while the `ContextCommand` handles multiple things related

I would have probably injected that dependency through an interface instead of using a function. Adding a function is like increasing the responsibility of `ContextCommand` to check if the capability is enabled or not. 

Injecting an interface removes the responsibility from the command and delegates it to whatever implements the interface the command really need. In that way, the command doesn't care about the implementation, just know what it needs.

AgustinRamiroDiaz:
> I would have probably injected that dependency through an interface instead of using a function. Adding a function is like increasing the responsibility of ContextCommand to check if the capability is enabled or not.

@ifbyol you mean doing something like this?

```
type ContextCommand struct {
	K8sClientProvider    okteto.K8sClientProvider
	LoginController      login.LoginInterface
	OktetoClientProvider types.OktetoClientProvider

	OktetoContextWriter okteto.ContextConfigWriterInterface
	kubetokenChecker    kubetokenChecker
}

type kubetokenChecker interface {
	IsPresent(oktetoUrl string) (bool, error)
}

type httpKubetokenChecker struct{}

func (h *httpKubetokenChecker) IsPresent(oktetoURL string) (bool, error) {
	kubeTokenURL, err := okteto.ParseOktetoURLWithPath(oktetoURL, "auth/kubetoken")
	if err != nil {
		return false, fmt.Errorf("failed to parse kubetoken url: %w", err)
	}

	resp, err := http.Get(kubeTokenURL)
	if err != nil {
		return false, fmt.Errorf("failed to get kubetoken url: %w", err)
	}

	return resp.StatusCode == http.StatusUnauthorized, nil
}

// NewContextCommand creates a new ContextCommand
func NewContextCommand() *ContextCommand {
	return &ContextCommand{
		K8sClientProvider:    okteto.NewK8sClientProvider(),
		LoginController:      login.NewLoginController(),
		OktetoClientProvider: okteto.NewOktetoClientProvider(),
		OktetoContextWriter:  okteto.NewContextConfigWriter(),
		kubetokenChecker:     &httpKubetokenChecker{},
	}
}

```


> Injecting an interface removes the responsibility from the command and delegates it to whatever implements the interface the command really need. In that way, the command doesn't care about the implementation, just know what it needs.

I feel that the responsibility is still there since the object is still getting created here. Now it's grouped so that's a bit tidier
I think that for it to be fully injected we might need to pass it as an argument, but the refactor would be huge since `NewContextCommand` is everywhere in the code

ifbyol:
> > I would have probably injected that dependency through an interface instead of using a function. Adding a function is like increasing the responsibility of ContextCommand to check if the capability is enabled or not.
> 
> @ifbyol you mean doing something like this?
> 
> ```
> type ContextCommand struct {
> 	K8sClientProvider    okteto.K8sClientProvider
> 	LoginController      login.LoginInterface
> 	OktetoClientProvider types.OktetoClientProvider
> 
> 	OktetoContextWriter okteto.ContextConfigWriterInterface
> 	kubetokenChecker    kubetokenChecker
> }
> 
> type kubetokenChecker interface {
> 	IsPresent(oktetoUrl string) (bool, error)
> }
> 
> type httpKubetokenChecker struct{}
> 
> func (h *httpKubetokenChecker) IsPresent(oktetoURL string) (bool, error) {
> 	kubeTokenURL, err := okteto.ParseOktetoURLWithPath(oktetoURL, "auth/kubetoken")
> 	if err != nil {
> 		return false, fmt.Errorf("failed to parse kubetoken url: %w", err)
> 	}
> 
> 	resp, err := http.Get(kubeTokenURL)
> 	if err != nil {
> 		return false, fmt.Errorf("failed to get kubetoken url: %w", err)
> 	}
> 
> 	return resp.StatusCode == http.StatusUnauthorized, nil
> }
> 
> // NewContextCommand creates a new ContextCommand
> func NewContextCommand() *ContextCommand {
> 	return &ContextCommand{
> 		K8sClientProvider:    okteto.NewK8sClientProvider(),
> 		LoginController:      login.NewLoginController(),
> 		OktetoClientProvider: okteto.NewOktetoClientProvider(),
> 		OktetoContextWriter:  okteto.NewContextConfigWriter(),
> 		kubetokenChecker:     &httpKubetokenChecker{},
> 	}
> }
> ```
> 
> > Injecting an interface removes the responsibility from the command and delegates it to whatever implements the interface the command really need. In that way, the command doesn't care about the implementation, just know what it needs.
> 
> I feel that the responsibility is still there since the object is still getting created here. Now it's grouped so that's a bit tidier I think that for it to be fully injected we might need to pass it as an argument, but the refactor would be huge since `NewContextCommand` is everywhere in the code

Yeah, I meant something like that.

Even if the creation of the dependency is there, you can initialize `ContextCommand` on your own passing a dependency that satisfies the interface without knowing what is inside of it. Even If I have to create the object for the checker, I don't know (and I don't care) what is done under the hood.

The refactor you mention shouldn't be such a big deal as you have to extract only the creation to be done outside and pass a parameter to the function. It might be a lot in terms of amount, but it is not a complex one.

Anyways, I didn't mention anything in the first review, because I don't think it's too bad to have it as a function within the command

AgustinRamiroDiaz:
@jLopezbarb @ifbyol I've found an issue around the Exec credentials for other okteto commands. Particularly I've been testing `okteto deploy` and I'm getting a
`
 x  Get "https://146.148.56.200:443/api/v1/namespaces/agustinramirodiaz/configmaps/okteto-git-go-getting-started": getting credentials: exec: executable okteto failed with exit code 1
 `
 
when trying to list the configmaps
 
But the weird thing is that this doesn't happen if I use an `echo` command [here](https://github.com/okteto/okteto/blob/cf1e767781fa76c436c869827a5a54f7dd6fb99f/pkg/okteto/context.go#L354-L361) (obviously with the token as argument) the deploy works fine. Do you know what could be causing the issue? This doesn't seem to be an issue with `okteto` not being in path since there's another exec error for that. 

In the meantime, I'll try some manipulations to see if I can get some output from the `kubetoken` command being executed internally
 
I've created the branch `agustin/feature-kubetoken` in order to also have [the other pr](https://github.com/okteto/okteto/pull/3404)'s code and test it.

ifbyol:
> @jLopezbarb @ifbyol I've found an issue around the Exec credentials for other okteto commands. Particularly I've been testing `okteto deploy` and I'm getting a `x Get "https://146.148.56.200:443/api/v1/namespaces/agustinramirodiaz/configmaps/okteto-git-go-getting-started": getting credentials: exec: executable okteto failed with exit code 1`
> 
> when trying to list the configmaps
> 
> But the weird thing is that this doesn't happen if I use an `echo` command [here](https://github.com/okteto/okteto/blob/cf1e767781fa76c436c869827a5a54f7dd6fb99f/pkg/okteto/context.go#L354-L361) (obviously with the token as argument) the deploy works fine. Do you know what could be causing the issue? This doesn't seem to be an issue with `okteto` not being in path since there's another exec error for that.
> 
> In the meantime, I'll try some manipulations to see if I can get some output from the `kubetoken` command being executed internally
> 
> I've created the branch `agustin/feature-kubetoken` in order to also have [the other pr](https://github.com/okteto/okteto/pull/3404)'s code and test it.
> 
> Edit: Turns out the error is here
> 
> ```go
> 		err = contextCMD.NewContextCommand().Run(ctx, &contextCMD.ContextOptions{
> 			Context: context,
> 		})
> ```
> 
> It's returning a `internal server error, please try again` 🤔 I'm still lost here
> 
> Edit: Going deeper I've found that the problem is in [here](https://github.com/okteto/okteto/blob/cf1e767781fa76c436c869827a5a54f7dd6fb99f/cmd/context/create.go#L353)
> 
> The error is
> 
> ```
> lookup cloud_okteto_com on 127.0.0.53:53: server misbehaving
> error getting user context: Post "https://cloud_okteto_com/graphql": dial tcp: lookup cloud_okteto_com on 127.0.0.53:53: server misbehaving
> ```
> 
> Does it ring a bell? The `127.0.0...` part may indicate that it has something to do with the proxy 🤔

@AgustinRamiroDiaz I don't think it is related to the proxy, according to the log, it is trying to do the request against `https://cloud_okteto_com/graphql`. `cloud_okteto_com` doesn't seem the valid host for the `graphql` API. I think at some point, it is mixing the context name with the okteto endpoint

AgustinRamiroDiaz:
> it is trying to do the request against https://cloud_okteto_com/graphql. cloud_okteto_com doesn't seem the valid host for the graphql API. I think at some point, it is mixing the context name with the okteto endpoint

@ifbyol you are right, that's the issue. 

I'm not sure where that behavior might be coming from since it's running the exact same command :thinking: I'm thinking it could be a concurrency or environment problem. Will keep looking into it

AgustinRamiroDiaz:
With a lot of debugging and the help of @jLopezbarb , we've found that the issue was due to a weird behavior of the context command when the `OKTETO_TOKEN` envvar is set (which gets set automatically by the CLI).

The solution was really simple: using the okteto URL instead of the context name (http url vs sanitized name). This fixes the issue and I could run `okteto deploy + okteto up` for https://github.com/okteto/go-getting-started

I think we should improve the `Context` package and its usage in general, but it will be hard. For this PR, I think the best is to go directly with the oktetoURL for the command 


--- 

This will be easier to test after https://github.com/okteto/okteto/pull/3404 get merged. Until then, if you want to test this flow you can use this branch https://github.com/okteto/okteto/tree/agustin/feature-kubetoken that has the merge of this branch and the kubetoken one

---

:heart: Thanks a lot for the help to both of you @jLopezbarb and @ifbyol !

jLopezbarb (approved):

AgustinRamiroDiaz:
@jLopezbarb @ifbyol the cli e2e tests were failing and I'm addressing the problems. I've created https://github.com/okteto/okteto/pull/3481 to update this branch with the fixes

PR #3412, opened 2023-02-28
fixes remote feature

AdrianPedriza:
## Proposed changes
- e4913dfb08ea07272a6e8f7be5f062db37402134: check if there is any external using `len(external) > 0` instead `external != nil`, this will allow us to now display `external resources cannot be deployed on a cluster not managed by okteto` when is not needed 
- b3b70666c6e3f1b4ba4c2391b23c141a3c8b611d: automate cli version used in remote deploy/destroy

### DIFF
--- a/cmd/deploy/remote.go
+++ b/cmd/deploy/remote.go
@@ -108,7 +108,7 @@ func (rd *remoteDeployCommand) deploy(ctx context.Context, deployOptions *Option
 	}
 
 	dockerfileSyntax := dockerfileTemplateProperties{
-		OktetoCLIImage:     getOktetoCLIVersion(),
+		OktetoCLIImage:     getOktetoCLIVersion(config.VersionString),
 		UserDeployImage:    deployOptions.Manifest.Deploy.Image,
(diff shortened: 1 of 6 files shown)


cmd/deploy/remote.go
```diff
@@ -224,3 +226,19 @@ func getOriginalCWD(cwd, manifestPath string) string {
 	manifestPathDir := filepath.Dir(fmt.Sprintf("/%s", manifestPath))
 	return strings.TrimSuffix(cwd, manifestPathDir)
 }
+
+func getOktetoCLIVersion() string {
```
AdrianPedriza, line 230: at some point we should unify shared code between deploy/destroy remote commands

jLopezbarb, line 230: Can we add a test for this one?

jLopezbarb (approved):

cmd/deploy/remote.go
AdrianPedriza: done 21650ac162c7da8df3bb040f6e988ee8a5c17691

PR #3421, opened 2023-03-02
Jlopezbarb/add remote deploy unit test

### LINKED ISSUE #3407: Remote deploy: add unit test for the feature

jLopezbarb:
# Proposed changes

Fixes #3407 

This PR is a follow up adding unit tests for remote deploy and destroy commands.

Changes the structs to inject dependencies and be able to test properly mocking the fully the operative system calls

### DIFF
--- a/cmd/deploy/deploy.go
+++ b/cmd/deploy/deploy.go
@@ -85,6 +85,7 @@ type DeployCommand struct {
 	CfgMapHandler      configMapHandler
 	Fs                 afero.Fs
 	DivertDriver       divert.Driver
+	PipelineCMD        pipelineCMD.PipelineDeployerInterface
 
 	PipelineType model.Archetype
 	isRemote     bool
@@ -170,6 +171,10 @@ func Deploy(ctx context.Context) *cobra.Command {
 			options.servicesToDeploy = args
 
 			k8sClientProvider := okteto.NewK8sClientProvider()
+			pc, err := pipelineCMD.NewCommand()
+			if err != nil {
+				return err
+			}
 			c := &DeployCommand{
 				GetManifest: model.GetManifestV2,
 
@@ -181,6 +186,7 @@ func Deploy(ctx context.Context) *cobra.Command {
 				isRemote:           utils.LoadBoolean(constants.OKtetoDeployRemote),
 				CfgMapHandler:      NewConfigmapHandler(k8sClientProvider),
 				Fs:                 afero.NewOsFs(),
+				PipelineCMD:        pc,
 			}
 			startTime := time.Now()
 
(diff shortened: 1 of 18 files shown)


cmd/deploy/deploy.go
```diff
@@ -174,6 +175,10 @@ func Deploy(ctx context.Context) *cobra.Command {
 			options.servicesToDeploy = args
 
 			k8sClientProvider := okteto.NewK8sClientProvider()
+			pc, err := pipelineCMD.NewCommand()
+			if err != nil {
+				return err
```
AdrianPedriza, line 180: should we add more details in the error like the previous approach?

cmd/deploy/remote.go
```diff
@@ -171,12 +156,54 @@ func (rd *remoteDeployCommand) deploy(ctx context.Context, deployOptions *Option
...
+
+	if err := tmpl.Execute(dockerfile, dockerfileSyntax); err != nil {
+		return "", err
+	}
+	return dockerfile.Name(), nil
 }
 
 func (rd *remoteDeployCommand) createDockerignoreIfNeeded(cwd, tmpDir string) error {
-	dockerignoreFilePath := fmt.Sprintf("%s/%s", cwd, ".oktetodeployignore")
+	dockerignoreFilePath := fmt.Sprintf("%s/%s", cwd, oktetoDockerignoreName)
```
AdrianPedriza, line 206: I think we should use `filepath.Join()`

jLopezbarb: Nice catch 😉

cmd/deploy/remote.go
```diff
@@ -222,9 +249,14 @@ func getDeployFlags(opts *Options) []string {
...
-func getOriginalCWD(cwd, manifestPath string) string {
-	manifestPathDir := filepath.Dir(fmt.Sprintf("/%s", manifestPath))
-	return strings.TrimSuffix(cwd, manifestPathDir)
+// getOriginalCWD returns the original cwd
+func (rd *remoteDeployCommand) getOriginalCWD(manifestPath string) (string, error) {
+	cwd, err := rd.workingDirectoryCtrl.Get()
+	if err != nil {
+		return "", err
+	}
+	manifestPathDir := filepath.Dir(filepath.Clean(fmt.Sprintf("/%s", manifestPath)))
```
AdrianPedriza, line 258: same, use `filepath.Join()`

jLopezbarb: This is using `filepath.Clean` which translates every separator into OS separator. We can't join a "" with a path because it will result in "./$manifestPath", and substraction wouldn't work

AdrianPedriza (approved):

cmd/deploy/remote.go
AdrianPedriza: oh yes, I read it wrong, apologies!

PR #3427, opened 2023-03-03
e2e: fix naming convention

jLopezbarb:
# Proposed changes

Fixes e2e tests failing due to having the same name

Actions were using a different name from the rest of e2e tests. Since we run the tests in parallel it may happen that two different tests have the same name and fail because of this

### DIFF
--- a/integration/actions/namespace_test.go
+++ b/integration/actions/namespace_test.go
@@ -21,10 +21,8 @@ import (
 	"log"
 	"os"
 	"os/exec"
-	"runtime"
 	"strings"
 	"testing"
-	"time"
 
 	"github.com/okteto/okteto/integration"
 	"github.com/okteto/okteto/pkg/okteto"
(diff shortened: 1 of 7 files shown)


integration/actions/namespace_test.go
```diff
@@ -139,10 +137,3 @@ func executeDeleteNamespaceAction(namespace string) error {
 	log.Printf("deleting namespace output: \n%s\n", string(o))
 	return nil
 }
-
-func getTestNamespace() string {
-	tName := fmt.Sprintf("TestAction-%s", runtime.GOOS)
-	name := strings.ToLower(fmt.Sprintf("%s-%d", tName, time.Now().Unix()))
```
AdrianPedriza, line 145: I am curious how we have been able to generate an identical name if we are composing it with `time.Now().Unix()`

AdrianPedriza (approved):

integration/actions/namespace_test.go
AgustinRamiroDiaz: maybe it was being trimmed somewhere?

PR #3432, opened 2023-03-06
fix destroy image envsub

AdrianPedriza:
Fixes env var substitution for destroy image.

## TEST
Having an env OK_IMG_DESTROY_TEST=alpine and using this manifest:
````yaml
destroy:
  image: ${OK_IMG_DESTROY_TEST}
  commands:
  - name: First step
    command: echo hello-world

````
you should be able to execute destroy remotely.

### DIFF
--- a/cmd/destroy/destroy.go
+++ b/cmd/destroy/destroy.go
@@ -240,6 +240,10 @@ func (dc *destroyCommand) getDestroyer(opts *Options) (destroyInterface, error) 
 			}
 		}
 
+		if err := manifest.ExpandEnvVars(); err != nil {
+			oktetoLog.Infof("could not expand manifest envs: %v", err)
+		}
+
 		isRemote := utils.LoadBoolean(constants.OKtetoDeployRemote)
 
 		destroyImage := ""
--- a/cmd/destroy/destroy.go
+++ b/cmd/destroy/destroy.go
@@ -240,20 +240,19 @@ func (dc *destroyCommand) getDestroyer(opts *Options) (destroyInterface, error) 
 			}
 		}
 
-		if err := manifest.ExpandEnvVars(); err != nil {
-			oktetoLog.Infof("could not expand manifest envs: %v", err)
-		}
-
 		isRemote := utils.LoadBoolean(constants.OKtetoDeployRemote)
 
 		destroyImage := ""
 		if manifest.Destroy != nil {
-			destroyImage = manifest.Destroy.Image
+			destroyImage, err = model.ExpandEnv(manifest.Destroy.Image, false)
+			if err != nil {
+				return nil, err
+			}
 		}
 		runInRemote := !isRemote && (destroyImage != "" || opts.RunInRemote)
 
 		if runInRemote {
-			deployer = newRemoteDestroyer(manifest)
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: pkg/model/manifest.go [hunk 1]
```diff
@@ -944,6 +944,12 @@ func (manifest *Manifest) ExpandEnvVars() error {
 		}
 	}
 	if manifest.Destroy != nil {
+		if manifest.Destroy.Image != "" {
+			manifest.Destroy.Image, err = envsubst.String(manifest.Destroy.Image)
+			if err != nil {
+				return errors.New("could not parse env vars for an image used for remote destroy")
```
AgustinRamiroDiaz, line 950: I think this code is shared for remote and local destroy, right?

pkg/model/manifest.go
```diff
@@ -944,6 +944,12 @@ func (manifest *Manifest) ExpandEnvVars() error {
 		}
 	}
 	if manifest.Destroy != nil {
+		if manifest.Destroy.Image != "" {
+			manifest.Destroy.Image, err = envsubst.String(manifest.Destroy.Image)
```
AgustinRamiroDiaz, line 948: Why not use `ExpandEnv`?

cmd/destroy/destroy.go
```diff
@@ -240,6 +240,10 @@ func (dc *destroyCommand) getDestroyer(opts *Options) (destroyInterface, error)
 			}
 		}
 
+		if err := manifest.ExpandEnvVars(); err != nil {
```
AgustinRamiroDiaz, line 243: I have some concerns about expanding envvars here. The `ExpandEnvVars` method is not fully idempotent, we've had issues from some users regarding the escaping of the envvars and this would add to that pile. 

Are we sure that this is getting called only once?

pkg/model/manifest.go
ifbyol: No, when an `image` filed is specified, the operation is only remote. So the check in line `947` ensures that this would affect only to remote operation

cmd/destroy/destroy.go
AdrianPedriza: yes, you are right. We might have problems if we call twice that func. I didn't take into account that we could expand twice the manifest if we deploy directly locally. I refactored it to avoid it in 1e75def29eaa4099a418c61821084d9ff50f2a13. Thanks!!

jLopezbarb (approved):

cmd/destroy/remote.go
```diff
@@ -77,17 +77,17 @@ type dockerfileTemplateProperties struct {
...
 	builder              builder.Builder
-	manifest             *model.Manifest
+	destroyImage         string
 	fs                   afero.Fs
 	workingDirectoryCtrl filesystem.WorkingDirectoryInterface
 	temporalCtrl         filesystem.TemporalDirectoryInterface
 }
 
-func newRemoteDestroyer(manifest *model.Manifest) *remoteDestroyCommand {
+func newRemoteDestroyer(destroyImage string) *remoteDestroyCommand {
```
jLopezbarb, line 86: Nice change!!

AdrianPedriza: 🤘

PR #3435, opened 2023-03-08
refactor registry package

jLopezbarb:
# Proposed changes

- Move buildkit errors to build package: It didn't make sense to have buildkit errors in the registry package if we were just going to use them when we build images
- Move `Dockerfile` related functions to build: Same as above, It didn't make sense to have those functions in the registry package if we were not going to use them outside the build package.
- Created `Config{}` in the okteto package: This allow us to **NOT** import  the `pkg/okteto` in the registry pkg evading a cycle import. `pkg/registry`.  To avoid importing it `pkg/registry` accepts an interface that the new `Config{}` which is outside the package implements.
- Created `OktetoRegistry` struct: This structs has all the actions that the registry can perform.
- Created `OktetoRegistryClient` struct: This new struct has all the operations that requires access to internet
- Created `imageCtrl` struct: This new struct has all the operations that needs the image without accessing internet such as getting the info from the image name
- Created `replacer` struct: This struct is the one that replaces the okteto registries with the expanded form of them


jLopezbarb:
e2e test has passed [here](https://app.circleci.com/pipelines/github/okteto/okteto/9612/workflows/431bcfdf-7089-42a2-b76c-355a53030650/jobs/28200)

pkg/cmd/build/build.go
```diff
@@ -210,7 +209,7 @@ func validateImage(imageTag string) error {
 	}
 	if (reg.IsOktetoRegistry(imageTag)) && strings.Count(imageTag, "/") != 1 {
 		prefix := constants.DevRegistry
-		if registry.IsGlobalRegistry(imageTag) {
+		if strings.HasPrefix(imageTag, constants.GlobalRegistry) {
```
AdrianPedriza, line 212: what happend if user use expanded notation to point global reg?

pkg/cmd/build/build.go
```diff
@@ -33,6 +33,7 @@ import (
 	"github.com/okteto/okteto/pkg/model"
 	"github.com/okteto/okteto/pkg/okteto"
 	"github.com/okteto/okteto/pkg/registry"
+	reg2 "github.com/okteto/okteto/pkg/registry/registry"
```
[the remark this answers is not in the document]
jLopezbarb: As commented above all the `reg2`/`reg` references has been removed in the final commit. It was done to have the old code and the new one living together

pkg/cmd/build/build.go
jLopezbarb: I moved it from [here](https://github.com/okteto/okteto/blob/21bda07cdc4a8dcfe612a31c5ad90e9af6780650/pkg/registry/registry.go#L214) so it was an issue that we were having, but you are right. We should check the expanded form of it. Done in https://github.com/okteto/okteto/pull/3435/commits/07249fbe66556fc9812c21a0ce367208bd7d7609

PR #3436, opened 2023-03-08
fix: avoid panic when removing a deployment that does not exist…

### LINKED ISSUE #3388: Panic when running `okteto down` without original deployment
**Describe the bug**
When trying to run `okteto down` from an app that 
- has been deleted
- dev environment still exists
- dev environment uses a selector
the command panics

**To Reproduce**

1. `git clone https://github.com/okteto/go-getting-started` 
1. `cd go-getting-started`
1. `ok deploy`
1. `ok up`
2. `exit` from the dev container
3. `kubectl delete -f k8s.yml`
4. Modify the okteto.yml so that it starts with
    ```
    dev:
      hello-world:
        selector:
          dev.okteto.com/deployed-by: go-getting-started
    ```
1. `ok down`
5. See this error:
      ```
      panic: runtime error: index out of range [0] with length 0
      
      goroutine 1 [running]:
      github.com/okteto/okteto/pkg/k8s/deployments.GetByDev({0x2bc1e50, 0xc000128000}, 0xc001bebb00, {0xc000419608, 0x13}, {0x2bdaa50?, 0xc000e3db00?})
              github.com/okteto/okteto/pkg/k8s/deployments/crud.go:138 +0x42b
      github.com/okteto/okteto/pkg/k8s/apps.Get({0x2bc1e50, 0xc000128000}, 0xc001bebb00, {0xc000419608, 0x13}, {0x2bdaa50, 0xc000e3db00})
              github.com/okteto/okteto/pkg/k8s/apps/crud.go:32 +0x52
      github.com/okteto/okteto/cmd/utils.GetApp({0x2bc1e50, 0xc000128000}, 0xc001bebb00, {0x2bdaa50, 0xc000e3db00}, 0x0)
              github.com/okteto/okteto/cmd/utils/dev.go:384 +0x72
      github.com/okteto/okteto/cmd.Down.func1(0xc0017a8280?, {0xc00071f9c0, 0x1, 0x1?})
              github.com/okteto/okteto/cmd/down.go:103 +0x2fa
      github.com/spf13/cobra.(*Command).execute(0xc0017a8280, {0xc00071f9a0, 0x1, 0x1})
              github.com/spf13/cobra@v1.5.0/command.go:872 +0x694
      github.com/spf13/cobra.(*Command).ExecuteC(0xc000737b80)
              github.com/spf13/cobra@v1.5.0/command.go:990 +0x3b4
              github.com/spf13/cobra@v1.5.0/command.go:918
      main.main()
              github.com/okteto/okteto/main.go:146 +0x9e9
    ```

**Expected behavior**
The behavior should be the same as when we don't use a selector, and it should return a clear error to the user like 
```
 x  Application 'hello-world' not found in namespace 't-agustinramirodiaz'
    Verify that your application is running and your okteto context is pointing to the right namespace
    Or set the 'autocreate' field in your okteto manifest if you want to create a standalone development container
    More information is available here: https://okteto.com/docs/reference/cli/#up

```

- OS: linux mint
- Version 21

andreafalzetti:
# Proposed changes

Fixes #3388

Fix a `panic` occurring in `okteto down` when a deployment cannot be found by label. I've added a bunch of test cases which hopefully cover most scenarios.

Here is an example of before/after when the scrnario described in #3388 would occur:

| Before | After |
|--------|-------|
|        |        |

This is instead the message that it outputs without using a label selector:

### DIFF
--- a/cmd/utils/dev.go
+++ b/cmd/utils/dev.go
@@ -175,7 +175,7 @@ func GetDevFromManifest(manifest *model.Manifest, devName string) (*model.Dev, e
 		return nil, ErrNoDevSelected
 	}
 
-	options := []string{}
+	var options []string
 	for name, dev := range manifest.Dev {
 		if name == devName {
 			return dev, nil
@@ -197,7 +197,7 @@ func SelectDevFromManifest(manifest *model.Manifest, selector OktetoSelectorInte
 		}
 		return devs[i] < devs[j]
 	})
-	items := []SelectorItem{}
+	var items []SelectorItem
 	for _, dev := range devs {
 		items = append(items, SelectorItem{
 			Name:   dev,
@@ -396,7 +396,7 @@ func GetApp(ctx context.Context, dev *model.Dev, c kubernetes.Interface, isRetry
 			return apps.NewDeploymentApp(deployments.Sandbox(dev)), true, nil
 		}
 		if len(dev.Selector) > 0 {
-			if err == oktetoErrors.ErrNotFound {
+			if oktetoErrors.IsNotFound(err) {
 				err = oktetoErrors.UserError{
 					E:    fmt.Errorf("didn't find an application in namespace %s that matches the labels in your Okteto manifest", dev.Namespace),
 					Hint: "Update the labels or point your context to a different namespace and try again"}
--- a/pkg/k8s/deployments/crud.go
+++ b/pkg/k8s/deployments/crud.go
@@ -132,6 +132,9 @@ func GetByDev(ctx context.Context, dev *model.Dev, namespace string, c kubernete
 			validDeployments = append(validDeployments, &dList.Items[i])
 		}
 	}
+	if len(validDeployments) == 0 {
+		return nil, oktetoErrors.ErrNotFound
+	}
 	if len(validDeployments) > 1 {
 		return nil, fmt.Errorf("found '%d' deployments for labels '%s' instead of 1", len(validDeployments), dev.LabelsSelector())
 	}
--- a/pkg/k8s/deployments/crud_test.go
+++ b/pkg/k8s/deployments/crud_test.go
@@ -28,28 +28,190 @@ import (
 )
 
 func TestGet(t *testing.T) {
-	ctx := context.Background()
-	deployment := &appsv1.Deployment{
-		ObjectMeta: metav1.ObjectMeta{
-			Name:      "fake",
-			Namespace: "test",
+	tests := []struct {
+		name               string
+		deployments        *appsv1.DeploymentList
+		dev                *model.Dev
+		namespace          string
+		expectedErr        error
+		expectedFoundCount int
+	}{
+		{
+			name: "Get by Name: no deployments",
+			dev: &model.Dev{
+				Name: "fake",
+			},
+			namespace: "test",
+			deployments: &appsv1.DeploymentList{
+				Items: []appsv1.Deployment{},
+			},
+			expectedErr:        fmt.Errorf("deployments.apps \"%s\" not found", "fake"),
+			expectedFoundCount: 0,
+		},
+		{
+			name: "Get by Name: found 1 deployment",
+			dev: &model.Dev{
+				Name: "fake",
+			},
+			namespace: "test",
+			deployments: &appsv1.DeploymentList{
+				Items: []appsv1.Deployment{
+					{
+						ObjectMeta: metav1.ObjectMeta{
+							Name:      "fake",
+							Namespace: "test",
+						},
+					},
+				},
+			},
+			expectedErr:        nil,
+			expectedFoundCount: 1,
+		},
+		{
+			name: "Search by Label: no deployments found",
+			dev: &model.Dev{
+				Selector: map[string]string{
+					"deployed-by": "fake",
+				},
+			},
+			namespace: "test",
+			deployments: &appsv1.DeploymentList{
+				Items: []appsv1.Deployment{},
+			},
+			expectedErr:        oktetoErrors.ErrNotFound,
+			expectedFoundCount: 0,
+		},
+		{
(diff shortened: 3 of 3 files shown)


cmd/utils/dev.go
```diff
@@ -175,7 +175,7 @@ func GetDevFromManifest(manifest *model.Manifest, devName string) (*model.Dev, e
 		return nil, ErrNoDevSelected
 	}
 
-	options := []string{}
+	var options []string
```
andreafalzetti, line 178: ✨ fix suggested by the linter [[1]](https://deepsource.io/gh/okteto/okteto/run/851a6b0a-999a-4ffb-89ed-cd9c008f703b/go/GO-W1027?listindex=0)

cmd/utils/dev.go
```diff
@@ -197,7 +197,7 @@ func SelectDevFromManifest(manifest *model.Manifest, selector OktetoSelectorInte
 		}
 		return devs[i] < devs[j]
 	})
-	items := []SelectorItem{}
+	var items []SelectorItem
```
andreafalzetti, line 200: ✨ fix suggested by the linter [[1]](https://deepsource.io/gh/okteto/okteto/run/851a6b0a-999a-4ffb-89ed-cd9c008f703b/go/GO-W1027?listindex=0)

cmd/utils/dev.go
```diff
@@ -387,6 +387,11 @@ func GetApp(ctx context.Context, dev *model.Dev, c kubernetes.Interface, isRetry
 	app, err := apps.Get(ctx, dev, dev.Namespace, c)
 	if err != nil {
 		if !oktetoErrors.IsNotFound(err) {
+			if len(dev.Selector) > 0 {
```
andreafalzetti, line 390: the check below would never be true because it would always end up here, I figure we can merge them and also avoid the double check on "not found"

pkg/k8s/deployments/crud_test.go
```diff
@@ -28,28 +28,162 @@ import (
 )
 
 func TestGet(t *testing.T) {
-	ctx := context.Background()
-	deployment := &appsv1.Deployment{
-		ObjectMeta: metav1.ObjectMeta{
-			Name:      "fake",
-			Namespace: "test",
+	tests := []struct {
```
andreafalzetti, line 31: I've added test cases that I thought were meaningful. Because I might not have the full context yet, I appreciate if there is feedback on this, and I am happy to adjust 🙏

cmd/utils/dev.go
ifbyol: Why we need to include this check here? This if is whenever we get an error which is a not found. I mean, if the error is a transient error connecting with K8s, it might exist an application that matches with the labels but we weren't able to retrieve it for other reasons.

The other cases only happen when we got a not-found error

pkg/k8s/deployments/crud_test.go
ifbyol: I think those scenarios are perfect, @jLopezbarb can you check them too?

cmd/utils/dev.go
andreafalzetti: What I've observed is that because there is this initial check for NotFound errors, [my change](https://github.com/okteto/okteto/pull/3436/files#diff-70f6fa6cd3171b4938b73394bb9432e12df9ae8da9203c639799aad23a926161R136) would always end up in this conditional branch instead of the one below with the Selector.

My understandig is that these 2 check against the same thing:

```go
if !oktetoErrors.IsNotFound(err) {
```

and 

```go
if err == oktetoErrors.ErrNotFound {
```

so they could be merged together. Maybe I'm misisng something I can't see yet. 🤔

ifbyol: But in line `389` we are checking that the error is **not** a `NotFound`. So it might be that there is a dev with the specified label selector, but it could not be retrieved for whatever reason.

andreafalzetti: good catch @ifbyol, I have missed the `!` 😄 Fixing now.

andreafalzetti: I've reverted this change. I was too excited to open the PR 😅

andreafalzetti: I'm moving the PR to draft to take some time to review properly because I think it's not printing the right error message.

andreafalzetti: All good, I was slightly confused because in the issue the expected error message was different from what I think should be displayed.

ifbyol (approved):
Code looks good but I would wait until @jLopezbarb 's review before merging. Nice work!

pkg/k8s/deployments/crud_test.go
jLopezbarb: I think they are the ones to have. We could add also another scenario (very weird scenario): `Search by Label: Already in dev`. For this use case the deployment that is returned by the client should have `dev.okteto.com/clone` set.
This use case is a very weird scenario where the user creates the environment with the `autocreate` and then tries to connect to that dev which is not supported and we should return a `not found` error

pkg/k8s/deployments/crud.go
```diff
@@ -132,6 +132,9 @@ func GetByDev(ctx context.Context, dev *model.Dev, namespace string, c kubernete
 			validDeployments = append(validDeployments, &dList.Items[i])
 		}
 	}
+	if len(validDeployments) == 0 {
```
jLopezbarb, line 135: Can we test this new scenario? I think it's not covered in the tests that you added

pkg/k8s/deployments/crud_test.go
jLopezbarb: I think the tests are perfect!!! Review my other comment in case that test is needed but other than that LGTM

jLopezbarb (approved):

PR #3439, opened 2023-03-13
Jlopezbarb/add global push

### LINKED ISSUE #3372: feat: push into global registry for non-admin users
## Description

This approach focuses on trying not to build the same tag twice. For it we will modify the behaviors of `checkServicesToBuild` and `buildService` so that it checks if we need to build the tag again. We will check if the project directoy is clean (no changes over latest commit in the repository) to know if other people have already pushed that commit previously.

### Configuration

- [x] Check if the okteto context has push access to the global registry.

### Build

- [x] If the context doesn't have push access to global registry, the behavior doesn't change. If the image field has not been set we push into dev registry (`okteto.dev`) or into the image that the user has set.
- [x] If the image has not been set (image field doesn't exist), we should push into the global registry (`okteto.global`) only if the project directory is clean using the commit sha as tag.
- [x] If the image has not been set (image field doesn't exist), we should push into the dev registry (`okteto.dev`) if the project directory is dirty. 
- [x] Skip the build of the image if the commit sha has been already built into global registry if the project directory is clean.
- [x] If the user has the `--no-cache` flag we shouldn't skip if the image is already on the global registry

### Checking if an image exists

- [x] If the image has not been set (image field doesn't exist) and the project directory is clean, we should check in the global registry if the commit sha has already been created.
- [x] If the image has not been set (image field doesn't exist) and the project directory is dirty, we should check in the dev registry if the okteto tag has been created.

jLopezbarb:
# Proposed changes

Fixes #3372 

- e4323ac7290c2c0736f9e4addb2c64f57c4e9c71: Add check to know if the user has access to push an image to the global registry. 
  We could use this new function to check if the user has push permissions before building the image, so the user get the feedback before building the image (which can take more than 10 min). cc: @pchico83 
- 1bf7323647ab1afe1fadc60b6fc6b530d9697675: Check if the repository has any change in top of the latest commit
- fbd25dc819876dba65430fe2c706688f227f7cc5: Check in global and dev registry if the commit has been built before building it
- 56583b0db29e59d1faaf0557f967d38d70c2765c and 1a66343461af4211b113f5e8c00843dc92b541c7: Adds the behavior of building only if the conditions are met.


cmd/build/command.go
```diff
@@ -50,6 +50,7 @@ type registryInterface interface {
 	GetImageTagWithDigest(imageTag string) (string, error)
 	IsOktetoRegistry(image string) bool
 	GetImageReference(image string) (registry.OktetoImageReference, error)
+	HasGlobalPushAcces() (bool, error)
```
AdrianPedriza, line 53:
```suggestion
	HasGlobalPushAccess() (bool, error)
```

cmd/build/v1/build.go
```diff
@@ -35,32 +35,53 @@ type OktetoBuilderInterface interface {
 
 type oktetoRegistryInterface interface {
 	GetImageTagWithDigest(imageTag string) (string, error)
+	HasGlobalPushAcces() (bool, error)
```
AdrianPedriza, line 38:
```suggestion
	HasGlobalPushAccess() (bool, error)
```

pkg/registry/registry.go
```diff
@@ -22,6 +22,8 @@ import (
 	oktetoLog "github.com/okteto/okteto/pkg/log"
 )
 
+const globalTestImage = "okteto.global/test"
```
AdrianPedriza, line 25: since it is only used once, I don't think it is helpful to have it as a constant, IMO it is more descriptive to use its value in this case when call `or.imageCtrl.ExpandOktetoGlobalRegistry("okteto.global/test")`

pkg/repository/repository.go
```diff
@@ -0,0 +1,126 @@
...
+package repository
+
+import (
+	"fmt"
+
+	"github.com/go-git/go-git/v5"
+	"github.com/go-git/go-git/v5/plumbing"
+)
+
+type Repository struct {
```
AdrianPedriza, line 23: Can we add a description about the purpose of this struct?

cmd/build/v2/services.go
```diff
@@ -106,23 +108,93 @@ func (bc *OktetoBuilder) checkServicesToBuild(service string, manifest *model.Ma
...
+func (bc *OktetoBuilder) checkIfCommitIsAlreadyBuilt(ctx context.Context, manifestName, svcName string, options *types.BuildOptions) (string, bool) {
+	if !bc.Config.IsCleanProject() {
+		return "", false
+	}
+	if options.NoCache {
+		return "", false
+	}
+	sha := bc.Config.GetHash()
+
+	targetRegistries := []string{constants.GlobalRegistry, constants.DevRegistry}
```
AdrianPedriza, line 188: at this point, could we known the target registry to only check it there if image is already built?

cmd/build/v2/build.go
```diff
@@ -132,6 +149,11 @@ func (bc *OktetoBuilder) Build(ctx context.Context, options *types.BuildOptions)
 			if options.EnableStages {
```
AdrianPedriza, line 149: should we move this spinner update once we will build the image? just before `bc. buildService`

cmd/build/v2/build.go
```diff
@@ -199,6 +221,9 @@ func (bc *OktetoBuilder) buildSvcFromDockerfile(ctx context.Context, manifest *m
 		return "", fmt.Errorf("error expanding build args from service '%s': %w", svcName, err)
 	}
 
+	tagToBuild := bc.tagsToCheck(manifest.Name, svcName, buildSvcInfo)[0]
```
AdrianPedriza, line 224: indexing to the first element of the output of the `tagsToCheck` function makes us too dependent on that function I think. Could we refactor that function and make it more explanatory? When it is used it assumes that certain values will occupy certain positions and it is a bit confusing to read, here and in a few lines ahead

cmd/build/v2/services.go
```diff
@@ -106,23 +108,93 @@ func (bc *OktetoBuilder) checkServicesToBuild(service string, manifest *model.Ma
...
+func (bc *OktetoBuilder) isImageBuilt(tags []string) (string, error) {
+	for _, tag := range tags {
+		imageWithDigest, err := bc.Registry.GetImageTagWithDigest(tag)
+		if oktetoErrors.IsNotFound(err) {
+			continue
+		} else if err != nil {
+			// return error if the registry doesn't send a not found error
+			return "", fmt.Errorf("error checking image at registry %s: %v", tag, err)
+		}
+		return imageWithDigest, nil
```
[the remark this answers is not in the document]
AdrianPedriza: as you prefer. With this suggestion we will not check if `isNotFound(err)` when `err == nil` . I also think makes reading simpler to get rid of that _else_.

pkg/repository/repository_test.go
```diff
@@ -0,0 +1,242 @@
...
+					},
+				},
+			},
+			expected: expected{
+				isClean: false,
+				err:     assert.AnError,
+			},
+		},
+		{
+			name: "repository could not access worktree",
```
AdrianPedriza, line 106:
```suggestion
			name: "worktree could not access status",
```

cmd/build/v2/services_test.go
```diff
@@ -175,24 +191,222 @@ func TestGetToBuildTag(t *testing.T) {
...
+					sha:       "",
+				},
+			},
+			expected: expected{
+				image:     "",
+				hasAccess: false,
+			},
+		},
+		{
+			name: "no access no clean commit",
```
AdrianPedriza, line 335: duplicated test with the above one? I think behavior is the same

pkg/registry/registry.go
jLopezbarb: The problem with that is that you don't know what is `okteto.global/test`. Having it as a constant makes you read that it's the test image for global registry, otherwise people could ask why that image and not any other

cmd/build/v2/build.go
```diff
@@ -49,6 +58,7 @@ type OktetoBuilder struct {
 	Registry  oktetoRegistryInterface
 	V1Builder *buildv1.OktetoBuilder
 
+	Config OktetoBuilderConfigInterface
```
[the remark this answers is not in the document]
jLopezbarb: It's config about the builder since it has the global access

cmd/build/v2/services_test.go
jLopezbarb: No, in this test the user doesn't have access to global registry while in the above the user has it.

I considered removing it because of the reason that you said but decided to have it in case we change the behavior still have the same output

cmd/build/v2/build.go
ifbyol: agree, Why the first position? If we only use one, why we return a list?

cmd/build/v2/services.go
```diff
@@ -106,23 +108,95 @@ func (bc *OktetoBuilder) checkServicesToBuild(service string, manifest *model.Ma
...
 	}
-	return ""
+	return tagsToCheck
+}
+
+func getImageFromTmpl(targetRegistry, repoName, svcName, tag string) string {
+	return fmt.Sprintf("%s/%s-%s:%s", targetRegistry, repoName, svcName, tag)
+}
+
+func (bc *OktetoBuilder) checkIfCommitIsAlreadyBuilt(ctx context.Context, manifestName, svcName string, options *types.BuildOptions) (string, bool) {
```
ifbyol, line 181: I think it's a bit odd that the function is named  `checkIfCommitIsAlreadyBuilt` but the commit is not a parameter of the function

cmd/build/v2/services_test.go
ifbyol: Even if the output is the same, the input is different. You don't know if `hasAccess` might change or not the output. From the point of view of this test, the implementation is a black box and it should check that given some inputs what are the expected output, so it should check different output combinations

pkg/registry/client.go
```diff
@@ -99,6 +119,15 @@ func (c client) GetImageConfig(image string) (*v1.ConfigFile, error) {
 	return cfg, nil
 }
 
+func (c client) HasPushAccess(image string) (bool, error) {
+	ref, err := name.ParseReference(image)
+	if err != nil {
+		return false, fmt.Errorf("error checking push access: %w", err)
+	}
+	err = remote.CheckPushPermission(ref, c.getAuthHelper(ref), oktetoHttp.DefaultTransport())
```
ifbyol, line 127: `oktetoHttp.DefaultTransport()` would this work with self-signed certificate instances? I think you need to use `c.getTransport()`

cmd/build/command_test.go
```diff
@@ -18,14 +18,77 @@ import (
...
 	"github.com/stretchr/testify/assert"
 )
 
+type fakeRegistry struct {
+	err      error
+	registry map[string]fakeImage
+}
+
+// fakeImage represents the data from an image
+type fakeImage struct {
```
ifbyol, line 38: Are we really using this?

AdrianPedriza (approved):

Issue #3433, opened 2023-03-07
Show spinner while checking if the image needs to be built


jLopezbarb:
**Is your feature request related to a problem? Please describe.**
We don't show a spinner while we are checking if images needs to be built. This is a bad UX since the user can think that okteto has hanged up if the user has several images to check or if his internet connection is slow.

**Describe the solution you'd like**
I'd like to have a spinner showing: "Checking images that needs to be built" or a similar message that shows what okteto is doing in that moment

**Additional context**
Function that checks if images needs to be built: https://github.com/okteto/okteto/blob/master/cmd/build/v2/services.go#L31

NitishKumar06:
I'm working on this

NitishKumar06:
Hi @jLopezbarb this is my first time coming up with a coding issue and needs some guidance. I'm unable to reproduce this issue. Here's what I've done till now:

1. Read about `okteto build` from okteto docs
2. Understood how important is Dockerfile to build image using okteto. However I'm not able to find when do we come across this log _Build section is not defined in your okteto manifest_? I've pushed the image to docker registry and tried to play with Dockerfile and `okteto.yml` but okteto do search for `Dockerfile` instead of `okteto.yml` which makes me wonder _what is the use case of okteto.yml when we have Dockerfile defined already?_
3. Please guide me further on how to reproduce this issue. I'm eager to make a contribution here.

Thank you!

jLopezbarb:
Hi @NitishKumar06, I'm sorry I forgot to put some steps to reproduce the issue. Here is how to reproduce the issue:

1. Create the following Dockerfile (I usually create this one since it's 
```Dockerfile
FROM alpine
RUN echo hola
``` 

2. Create the following okteto manifest (create more services if you want to have more time between the messages):
```yaml
build:
  svc1:
    dockerfile: Dockerfile
  svc2:
    dockerfile: Dockerfile
  svc3:
    dockerfile: Dockerfile
  svc4:
    dockerfile: Dockerfile
  svc5:
    dockerfile: Dockerfile
```
3. Run `okteto deploy`: This will check if the services are built and start building the services. Notice that the time between the message `i  Using jlopezbarb @ cloud.okteto.com as context` and the next one is quick large without any input
[...]

PR #3441, opened 2023-03-13
Update dependencies and add spinner to check whether images needs to be built

NitishKumar06:
Fixes: #3433

### DIFF
--- a/cmd/build/v2/services.go
+++ b/cmd/build/v2/services.go
@@ -16,8 +16,6 @@ package v2
 import (
 	"context"
 	"fmt"
-	"strings"
-
 	"github.com/okteto/okteto/pkg/constants"
 	oktetoErrors "github.com/okteto/okteto/pkg/errors"
 	"github.com/okteto/okteto/pkg/format"
@@ -25,6 +23,7 @@ import (
 	"github.com/okteto/okteto/pkg/model"
 	"github.com/okteto/okteto/pkg/okteto"
 	"golang.org/x/sync/errgroup"
+	"strings"
 )
 
 // GetServicesToBuild returns the services it has to build if they are not already built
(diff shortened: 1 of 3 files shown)


cmd/build/v2/services.go
```diff
@@ -52,7 +54,15 @@ func (bc *OktetoBuilder) GetServicesToBuild(ctx context.Context, manifest *model
 	}
 	close(toBuild)
 
+	// create a spinner to be loaded while checking if images needs to be built
+	s := spinner.New(spinner.CharSets[0], 100*time.Millisecond)
```
jLopezbarb, line 58: We have a wrapper for this library. You can use `oktetoLog.Spinner("Building images...")`. 

This will create the same spinner as the other ones across the application
To stop it we call `defer oktetoLog.StopSpinner()` to stop it at the end of the function.

cmd/build/v2/services.go
```diff
@@ -52,7 +54,15 @@ func (bc *OktetoBuilder) GetServicesToBuild(ctx context.Context, manifest *model
...
+	s.Color("blue")
+
 	if len(toBuild) == 0 {
+		s.Prefix = "Building images..."
+		s.Start()
+		time.Sleep(5 * time.Second)
```
jLopezbarb, line 64: This stops the application for `5 Seconds`. Could you move the creation of the spinner to line 40 just before `// check if images are at registry (global or dev) and set envs or send to build` comment?

NitishKumar06: The spinner is taking time to load. How do we make sure the spinner runs as soon as the `if` block is hit?

NitishKumar06: I see so this was the catch! I'd be more cautious from next time.

cmd/build/v2/services.go
```diff
@@ -14,115 +14,121 @@
...
+        })
+    }
+
+    if err := g.Wait(); err != nil {
+        return nil, err
+    }
+    close(toBuild)
+
+    if len(toBuild) == 0 {
+        defer oktetoLog.StopSpinner()
```
AgustinRamiroDiaz, line 61: Shouldn't we `defer` right after `StartSpinner` in order to be sure to stop the spinner in any case?

cmd/build/v2/services.go
NitishKumar06: Resolved!

jLopezbarb (approved):

AgustinRamiroDiaz (approved):

AgustinRamiroDiaz:
@jLopezbarb LGTM

PR #3438, opened 2023-03-13
Ignore OKTETO_TOKEN warning if OKTETO_TOKEN is the one in the current…

dhanusaputra:
# Proposed changes

Fixes #3434 

This PR will ignore logging warning when env token `OKTETO_TOKEN` is the same with token inside storage

PR #3454, opened 2023-03-16
Jlopezbar/fix e2e tests

jLopezbarb:
After merging https://github.com/okteto/okteto/pull/3438 the init from env vars was broken because okteto.Context() could be nil and return a corrupted status

## Proposed changes:
- Check if the context is empty before trying to access to it
- If the user doesn't have the context saved asumes that it's empty


AdrianPedriza:
@jLopezbarb can you add more context in description for a proper review? 🙏

jLopezbarb:
> @jLopezbarb can you add more context in description for a proper review? 🙏

Done

AdrianPedriza (approved):

PR #3455, opened 2023-03-16
fix(deploy): wait if dependency is being deployed

### LINKED ISSUE #3281: `wait` doesn't wait if the dependency is being deployed
**Describe the bug**
`wait` only waits the first time a dependency is deployed, but it doesn't work if the dependency is in "deploying" state.

**To Reproduce**

1. Deploy two repos with the same dependency in parallel
1. One of the repos doesn't wait 

**Expected behavior**
Both repos should wait for the dependency.

**Additional context**
We should also redeploy the dependency if the dependency is in error state

andreafalzetti:
# Proposed changes

Fixes #3281 

Loom video demo: https://www.loom.com/share/8d7c2482a1c54f318b3f712b31c3f0a2

Follow up demo (covers scenario 4): https://www.loom.com/share/1dca9392bce049268ff97b8e44499588

Context: Currently, using `okteto deploy` on a manifest that has dependencies already being deployed, despite having `wait: true` on the dependency, does not wait for the deployment to be completed.

Proposed change: Extending the `deploy` cmd to wait when a repository is found in a "deploying" state. To achieve that, we can rely on an existing Configmap being found and its `Status` value.

Additionally, if we found that a previous job is running, we can also stream the logs. Unfortunately this is only possible for "pipeline deployments". If the deploy occurs locally, the logs won't be available for retrvial.

## How to test

There are 2 scenarios to test.

Scenario 1: **Deploy of the movies-multi-repo twice
Steps:
- Checkout this branch and build the CLI
- Clone movies-multi-repo: `git clone git@github.com:okteto/movies-multi-repo.git`
- Open two terminal sessions and run `okteto deploy` on each one
- Notice how the second one will stream the logs of the first one

Scenario 2: **Deploy `movies-frontend` and `movies-multi-repo`
Steps:
- Checkout this branch and build the CLI
- Clone movies-frontend: `git clone git@github.com:okteto/movies-frontend.git`
- Change the `okteto.yml` of `movies-frontend` removing the `depencencies` block
- Clone movies-multi-repo: `git clone git@github.com:okteto/movies-multi-repo.git`
- Change the `okteto.yml` of `movies-multi-repo` with:
```yaml
dependencies:
  movies-frontend:
    repository: https://github.com/okteto/movies-frontend
    wait: true
deploy:
  - name: Deploy MongoDB
    command: helm upgrade --install mongodb mongodb-12.1.30.tgz -f values.yml
 ```
- Open a terminal sessions for each repo and run `okteto deploy` on each one
- Notice how the second one will say that movies-frontend is already being deployed and waits until it ends.

Scenario 3: **Don't wait for the dependency**
Steps:
- Checkout this branch and build the CLI
- Clone movies-frontend: `git clone git@github.com:okteto/movies-frontend.git`
- Clone movies-multi-repo: `git clone git@github.com:okteto/movies-multi-repo.git`
- Modify the `okteto.yml` manifest of `movies-multi-repo`, set "wait: false`
- Open a terminal sessions for each repo and run `okteto deploy` on each one
- Notice how the second one will say that movies-frontend is already being scheduled for deploy and continue without waiting

Scenario 4: **Recover from `error` state**
Steps:
- Checkout this branch and build the CLI
- Clone movies-frontend: `git clone git@github.com:okteto/movies-frontend.git`
- Clone movies-multi-repo: `git clone git@github.com:okteto/movies-multi-repo.git`
- Modify the `okteto.yml` manifest of `movies-frontend`, adding a non-existing command in the deploy (so that it will break)
- Open a terminal sessions for each repo and run `okteto deploy` on each one
- Notice how the first will go into error state, but when you deploy the second one will recover to progressing and eventually to deployed

### DIFF
--- a/cmd/pipeline/deploy.go
+++ b/cmd/pipeline/deploy.go
@@ -15,6 +15,7 @@ package pipeline
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"os"
 	"os/signal"
@@ -126,14 +127,60 @@ func (pc *Command) ExecuteDeployPipeline(ctx context.Context, opts *DeployOption
 			return fmt.Errorf("failed to load okteto context '%s': %v", okteto.Context().Name, err)
 		}
 
-		_, err = configmaps.Get(ctx, pipeline.TranslatePipelineName(opts.Name), opts.Namespace, c)
-		if err == nil {
-			oktetoLog.Success("Skipping repository '%s' because it's already deployed", opts.Name)
-			return nil
-		}
+		cfgName := pipeline.TranslatePipelineName(opts.Name)
 
-		if !oktetoErrors.IsNotFound(err) {
+		cfg, err := configmaps.Get(ctx, cfgName, opts.Namespace, c)
+		if err != nil && !oktetoErrors.IsNotFound(err) {
 			return err
+		}
+		if err == nil {
+			if cfg != nil && cfg.Data != nil {
+				if cfg.Data["status"] == pipeline.DeployedStatus {
+					oktetoLog.Success("Skipping repository '%s' because it's already deployed", opts.Name)
+					return nil
+				}
+
+				if !opts.Wait && cfg.Data["status"] == pipeline.ProgressingStatus {
+					oktetoLog.Success("Repository '%s' already scheduled for deployment", opts.Name)
+					return nil
+				}
+
+				canStreamPrevLogs := cfg.Data["actionLock"] != "" && cfg.Data["actionName"] != "cli"
+
+				if opts.Wait && canStreamPrevLogs {
+					oktetoLog.Spinner(fmt.Sprintf("Repository '%s' is already being deployed, waiting for it to finish...", opts.Name))
+					oktetoLog.StartSpinner()
+					defer oktetoLog.StopSpinner()
+
+					existingAction := &types.Action{
+						ID:   cfg.Data["actionLock"],
+						Name: cfg.Data["actionName"],
+					}
+					if err := pc.waitUntilRunning(ctx, opts.Name, opts.Namespace, existingAction, opts.Timeout); err != nil {
+						return err
+					}
+					oktetoLog.Success("Repository '%s' successfully deployed", opts.Name)
+					return nil
+				}
+
+				if opts.Wait && !canStreamPrevLogs && cfg.Data["status"] == pipeline.ProgressingStatus {
+					oktetoLog.Spinner(fmt.Sprintf("Repository '%s' is already being deployed, waiting for it to finish...", opts.Name))
+					oktetoLog.StartSpinner()
+					defer oktetoLog.StopSpinner()
+
+					ticker := time.NewTicker(1 * time.Second)
+					err := configmaps.WaitForStatus(ctx, cfgName, opts.Namespace, pipeline.DeployedStatus, ticker, opts.Timeout, c)
+					if err != nil {
+						if errors.Is(err, oktetoErrors.ErrTimeout) {
+							return fmt.Errorf("timed out waiting for repository '%s' to be deployed", opts.Name)
+						}
+						return fmt.Errorf("failed to wait for repository '%s' to be deployed: %w", opts.Name, err)
+					}
+
+					oktetoLog.Success("Repository '%s' successfully deployed", opts.Name)
+					return nil
+				}
+			}
 		}
 	}
 
@@ -146,6 +193,10 @@ func (pc *Command) ExecuteDeployPipeline(ctx context.Context, opts *DeployOption
 		oktetoLog.Success("Repository '%s' scheduled for deployment", opts.Name)
 		return nil
 	}
+
+	oktetoLog.Spinner(fmt.Sprintf("Waiting for repository '%s' to be deployed...", opts.Name))
+	oktetoLog.StartSpinner()
+	defer oktetoLog.StopSpinner()
 
 	if err := pc.waitUntilRunning(ctx, opts.Name, opts.Namespace, resp.Action, opts.Timeout); err != nil {
 		return err
@@ -204,10 +255,6 @@ func (pc *Command) streamPipelineLogs(ctx context.Context, name, namespace, acti
 func (pc *Command) waitUntilRunning(ctx context.Context, name, namespace string, action *types.Action, timeout time.Duration) error {
 	waitCtx, ctxCancel := context.WithCancel(ctx)
 	defer ctxCancel()
-
-	oktetoLog.Spinner(fmt.Sprintf("Waiting for repository '%s' to be deployed...", name))
-	oktetoLog.StartSpinner()
-	defer oktetoLog.StopSpinner()
 
 	stop := make(chan os.Signal, 1)
 	signal.Notify(stop, os.Interrupt)
@@ -375,7 +422,7 @@ func (o *DeployOptions) setDefaults() error {
 }
 
 func (o *DeployOptions) toPipelineDeployClientOptions() (types.PipelineDeployOptions, error) {
-	varList := []types.Variable{}
+	var varList []types.Variable
 	for _, v := range o.Variables {
 		kv := strings.SplitN(v, "=", 2)
 		if len(kv) != 2 {
--- a/cmd/pipeline/destroy_test.go
+++ b/cmd/pipeline/destroy_test.go
@@ -25,7 +25,7 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-func TestDestroyPipelineSuccesful(t *testing.T) {
+func TestDestroyPipelineSuccessful(t *testing.T) {
 	ctx := context.Background()
 	response := &client.FakePipelineResponses{
 		DestroyResponse: &types.GitDeployResponse{
@@ -47,7 +47,7 @@ func TestDestroyPipelineSuccesful(t *testing.T) {
 	assert.NoError(t, err)
 }
 
-func TestDestroyPipelineSuccesfulWithWait(t *testing.T) {
+func TestDestroyPipelineSuccessfulWithWait(t *testing.T) {
 	ctx := context.Background()
 	response := &client.FakePipelineResponses{
 		DestroyResponse: &types.GitDeployResponse{
@@ -71,7 +71,7 @@ func TestDestroyPipelineSuccesfulWithWait(t *testing.T) {
 	assert.NoError(t, err)
 }
 
-func TestDestroyPipelineSuccesfulWithWaitStreamErr(t *testing.T) {
+func TestDestroyPipelineSuccessfulWithWaitStreamErr(t *testing.T) {
 	ctx := context.Background()
 	response := &client.FakePipelineResponses{
 		DestroyResponse: &types.GitDeployResponse{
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

File: cmd/pipeline/deploy.go [hunk 1]
```diff
@@ -375,7 +421,7 @@ func (o *DeployOptions) setDefaults() error {
 }
 
 func (o *DeployOptions) toPipelineDeployClientOptions() (types.PipelineDeployOptions, error) {
-	varList := []types.Variable{}
+	var varList []types.Variable
```
andreafalzetti, line 424: Change suggested by the linter ✨

cmd/pipeline/destroy_test.go
```diff
@@ -25,7 +25,7 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-func TestDestroyPipelineSuccesful(t *testing.T) {
+func TestDestroyPipelineSuccessful(t *testing.T) {
```
andreafalzetti, line 28: Some typos in this file 🧹

cmd/pipeline/deploy.go
```diff
@@ -126,10 +126,56 @@ func (pc *Command) ExecuteDeployPipeline(ctx context.Context, opts *DeployOption
...
 		}
 
-		_, err = configmaps.Get(ctx, pipeline.TranslatePipelineName(opts.Name), opts.Namespace, c)
+		cfgName := pipeline.TranslatePipelineName(opts.Name)
+
+		cfg, err := configmaps.Get(ctx, cfgName, opts.Namespace, c)
 		if err == nil {
-			oktetoLog.Success("Skipping repository '%s' because it's already deployed", opts.Name)
-			return nil
+			if cfg != nil && cfg.Data != nil {
```
andreafalzetti, line 133: I've haven't added tests of the function `ExecuteDeployPipeline` because it does have  dependencies which make it hard to add them, and it'd require a refactor.

I suggest having a follow-up issue for this PR to schedule for next sprint for example, to refactor this function and make it easier to test

pkg/k8s/configmaps/crud_test.go
```diff
@@ -0,0 +1,146 @@
...
+	"github.com/okteto/okteto/pkg/k8s/configmaps"
+	apiv1 "k8s.io/api/core/v1"
+	k8sErrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/kubernetes/fake"
+	k8sTesting "k8s.io/client-go/testing"
+)
+
+func TestGet(t *testing.T) {
```
andreafalzetti, line 32: This includes tests for the new code and also for the `Get` function. It felt overwelming to add testing for everything but it's something we can follow-up on with a dedicated issue.

cmd/pipeline/deploy.go
jLopezbarb: Which linter are you using? I think we should all start using the same linter to have a common agreement on these things

cmd/pipeline/deploy.go
jLopezbarb: Maybe we can encapsulate this new code into a function. Probably that makes this new code testable without having to test `ExecuteDeployPipeline` function

pkg/k8s/configmaps/crud_test.go
```diff
@@ -0,0 +1,146 @@
...
+			Namespace: ns,
+		},
+		Data: map[string]string{
+			"status": mockCfgStatus,
+		},
+	}
+	clientset := fake.NewSimpleClientset(mockCfg)
+
+	cfg, err := configmaps.Get(ctx, mockCfgName, ns, clientset)
+	if err != nil {
```
jLopezbarb, line 49: NIT: We usually use https://github.com/stretchr/testify to tests the code. There is no need to have lots of if statements and it also provides what's the difference between both objects
```suggestion
	require.NoError(t, err)
```

pkg/k8s/configmaps/crud_test.go
```diff
@@ -0,0 +1,146 @@
...
+		t.Fatalf("unexpected error returned by Get(): %v", err)
+	}
+
+	if mockCfgName != cfg.Name {
```
jLopezbarb, line 53:
```suggestion
	assert.Equal(t, mockCfgName, cfg.Name)
```

pkg/k8s/configmaps/crud_test.go
```diff
@@ -0,0 +1,146 @@
...
+		t.Errorf("Expected cfg name %s, got %s", mockCfgName, cfg.Name)
+	}
+
+	if mockCfgStatus != cfg.Data["status"] {
```
jLopezbarb, line 57:
```suggestion
	assert.Equal(t,mockCfgStatus, cfg.Data["status"] )
```

pkg/k8s/configmaps/crud_test.go
```diff
@@ -0,0 +1,146 @@
...
+
+	mockErr := errors.New("unexpected error")
+
+	clientset := fake.NewSimpleClientset()
+	clientset.Fake.PrependReactor("get", "configmaps", func(action k8sTesting.Action) (handled bool, ret runtime.Object, err error) {
+		return true, nil, mockErr
+	})
+
+	cfg, err := configmaps.Get(ctx, "fake-cfg-name", "fake-ns", clientset)
+	if cfg != nil {
```
jLopezbarb, line 73:
```suggestion
assert.Nil(t, cfg)
```

pkg/k8s/configmaps/crud_test.go
```diff
@@ -0,0 +1,146 @@
...
+		t.Errorf("Expected cfg to be nil, but got: %v", cfg)
+	}
+
+	if err == nil || err.Error() != mockErr.Error() {
```
jLopezbarb, line 77:
```suggestion
	assert.ErrorIs(t, mockErr, err)
```

pkg/k8s/configmaps/crud_test.go
```diff
@@ -0,0 +1,146 @@
...
+			Namespace: ns,
+		},
+		Data: map[string]string{
+			"status": "fake-initial-status",
+		},
+	}
+
+	clientset := fake.NewSimpleClientset(mockCfg)
+	err := configmaps.WaitForStatus(ctx, mockCfgName, ns, "fake-final-status", 1*time.Microsecond, clientset)
+	if err != oktetoErrors.ErrTimeout {
```
jLopezbarb, line 129:
```suggestion
	assert.ErrorIs(t, err, oktetoErrors.ErrTimeout)
```

cmd/pipeline/deploy.go
andreafalzetti: @jLopezbarb it's the one configured in the PR actions, DeepSource: https://github.com/okteto/okteto/pull/3455/checks?check_run_id=12068422538 

Does it block only my PRs? 😂

andreafalzetti: I guess the confusion might be the word "linter" - this seems more like a generic code health solution, but this sort of suggestions generally come from linters - sorry for the confusion :)

cmd/pipeline/deploy.go
andreafalzetti: ok I've tried the idea of extracting the new code into a function, however it became very complicated too, especially testing all the stuff that is a dependency like `configmaps.WaitForStatus` and `pc.waitUntilRunning` without having to refactor a huge amout.

It would probably be quicker and safer to takle this as an ad-hoc refactor, rewriting some code from scratch..

pkg/k8s/configmaps/crud_test.go
```diff
@@ -0,0 +1,146 @@
...
+	}
+
+	clientset := fake.NewSimpleClientset(mockCfg)
+
+	go func() {
+		mockCfg.Data["status"] = "fake-final-status"
+		clientset.CoreV1().ConfigMaps(ns).Update(ctx, mockCfg, metav1.UpdateOptions{})
+	}()
+
+	err := configmaps.WaitForStatus(ctx, mockCfgName, ns, "fake-final-status", 5*time.Second, clientset)
```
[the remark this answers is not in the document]
andreafalzetti: I will have a look at this PR today, for now if you agree we can merge as is and come back to it later once we agreed upon an approach?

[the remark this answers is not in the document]
andreafalzetti: @jLopezbarb I made a small change but now the tests run super fast, because I am injecting the ticker, which I think it's a nice thing to do anyway, so from the outside the function, it can be controlled how frequently we check of a status change. Have a look at the diff: https://github.com/okteto/okteto/compare/b6a16882a4bbb4064252df190bfb3502c3739983..ceafa44661010c21ecb57b94142aeb5ca6eab6d2

This also removes the issue of running the tests "slowly"

pkg/k8s/configmaps/crud_test.go
```diff
@@ -0,0 +1,137 @@
...
+		Data: map[string]string{
+			"status": mockCfgStatus,
+		},
+	}
+	clientset := fake.NewSimpleClientset(mockCfg)
+
+	cfg, err := configmaps.Get(ctx, mockCfgName, ns, clientset)
+
+	require.NoError(t, err)
+	assert.Equal(t, mockCfgName, cfg.Name)
```
AdrianPedriza, line 53: I think we can compare the whole object here instead compare specifics labels
```suggestion
	assert.Equal(t, mockCfg, cfg)
```

pkg/errors/errors.go
```diff
@@ -271,3 +274,8 @@ func IsClosedNetwork(err error) bool {
 func IsErrGitHubNotVerifiedEmail(err error) bool {
 	return err.Error() == ErrGitHubNotVerifiedEmail.Error()
 }
+
+// IsTimeout returns true if err is of the type timeout
+func IsTimeout(err error) bool {
```
AdrianPedriza, line 279: do we really need this func? Could we use errors.Is() directly?

AdrianPedriza:
I can't reproduce scenario 2. When I deploy `movies-frontend` and then `movies-multi-repo` I got the following for `movies-multi-repo`: 
Captura de pantalla 2023-03-21 a las 11 19 04
Do I miss something?

pkg/errors/errors.go
andreafalzetti: fair point, it had more logic before, but since I've changed to `errors.Is` I agree with you and it can be removed

andreafalzetti:
> I can't reproduce scenario 2 and 3. When I deploy `movies-frontend` and then `movies-multi-repo` I got the following for `movies-multi-repo`: Captura de pantalla 2023-03-21 a las 11 19 04 Do I miss something?

@AdrianPedriza I've realised that `movies-multi-repo` by default does not wait for dependencies to be deployed and I've changed the manifest to wait. I've updated the steps in Scenario 2 to reflect this, sorry!

pkg/k8s/configmaps/crud_test.go
andreafalzetti: good shout, done!

jLopezbarb (approved):

AdrianPedriza (approved):

AdrianPedriza:
> > I can't reproduce scenario 2 and 3. When I deploy `movies-frontend` and then `movies-multi-repo` I got the following for `movies-multi-repo`:  Do I miss something?
> 
> @AdrianPedriza I've realised that `movies-multi-repo` by default does not wait for dependencies to be deployed and I've changed the manifest to wait. I've updated the steps in Scenario 2 to reflect this, sorry!
> 
> Update: also, another change that Ive added to the instructions is to update `movies-frontend` Okteto's manifest to get rid of the dependencies (because it depends on `movies-multi-repo` itself, it was giving you that error).
> 
> Maybe it would have been easier next time to setup ad-hoc repos in my personal GitHub account for testing the PR. I wanted to keep it simple and reused the ones we have but with all the changes required I am not sure what was more convinent :)

I already got to test the scenarios! 👍

Issue #3442, opened 2023-03-14
Add graphql unit tests


jLopezbarb:
## Parent epic/feature

* Increase CLI test coverage

## Description

The package `pkg/okteto/okteto` has a low unit test coverage, making it very difficult to check if any additional test breaks previous scenarios.

Most of the files/functions in the package are easy to test since it just need to mock API graphql calls.

### Tips:

- Check that the function are used.
- Check if the function have a deprecated method. Check what's different with the original one and check in which backend version we introduced the new field. If it's not supported we should remove the deprecated fallback.
- Create the correct interface so it can be used from other parts of the project (e.g: `okClient.Previews().Create()`)

## Acceptance Criteria

- [x] `actions.go` functions are tested
- [x] `auth.go`functions are tested
- [x] `credential.go` functions is tested
- [x] `namespace.go`functions are tested
- [x] `pipeline.go`functions are tested
- [x] `preview.go` functions are tested
- [x] `secrets.go` functions are tested

jLopezbarb:
Closed since everything is merged

PR #3458, opened 2023-03-17
Jlopezbarb/add actions graphql tests

jLopezbarb:
# Proposed changes

Fixes  partly #3442

- Add graphql interface to mock it in the tests
- Add a fake graphql  client to support single calls
- Add a fake graphql multicall clients to support multiple calls in the same function
- Creates a new ticker with timeout which wraps `time.Ticker` and `time.Timer` to be able to not wait in the tests for time to pass
- Add tests for the different functions on the `actions.go` (graphql calls to get the actions)


pkg/okteto/client.go
```diff
@@ -245,7 +251,7 @@ func InDevContainer() bool {
 	return false
 }
 
-func query(ctx context.Context, query interface{}, variables map[string]interface{}, client *graphql.Client) error {
+func query(ctx context.Context, query interface{}, variables map[string]interface{}, client graphqlClientInterface) error {
 	err := client.Query(ctx, query, variables)
 	if err != nil {
 		if isAPITransientErr(err) {
```
AdrianPedriza, line 257: can we test this case? when err is transient

jLopezbarb: I plan to test that scenario when we test the `client.go` file, but I prefer to go file by file testing to avoid confussion while reviewing

pkg/okteto/pipeline.go
```diff
@@ -17,24 +17,28 @@ import (
...
 	"github.com/shurcooL/graphql"
 	giturls "github.com/whilp/git-urls"
 )
 
 type pipelineClient struct {
-	client *graphql.Client
-	url    string
+	client                    graphqlClientInterface
+	url                       string
+	tickerWithTimeoutProvider func(ticker, timeout time.Duration) oktetoTime.TickerWithTimeoutInterface
```
AgustinRamiroDiaz, line 34: I feel that the `time` package is not something we'll like long term. It's a very thin layer of wrapping that could exparse to the whole app without much gains and limiting ourselves to use golang's native concurrency model

I feel that this function itself it's a great way of handling the depnedency injection. I'd just use 2 functions here that would provide the 2 tickers

You can easily mock tickers by creating the struct with the channel, here you have an example https://github.com/okteto/app/blob/700b10e0e264beb2f3357ee724fcd027a12b5aa6/backend/pkg/controlplane/federatednamespace/crud/create_test.go#L33. This means that maybe there's no need for a provider at all, and you could directly have the tickers

AgustinRamiroDiaz (commented):
Love the initiative!

pkg/okteto/errors.go
```diff
@@ -1,12 +1,33 @@
...
+	"fmt"
+	"time"
+)
 
 var (
 	// ErrGithubMissingBusinessEmail is raised when the user does not have a business email
 	ErrGithubMissingBusinessEmail = errors.New("github-missing-business-email")
 )
 
+type pipelineTimeoutError struct {
```
AgustinRamiroDiaz, line 14: Love the typed errors approach :100:

pkg/okteto/client_test.go
```diff
@@ -25,6 +26,81 @@ import (
 	"golang.org/x/oauth2"
 )
 
+type fakeGraphQLClient struct {
```
AgustinRamiroDiaz, line 29: Shouldn't these structs go in `action_test.go`?

pkg/okteto/action.go
```diff
@@ -68,19 +75,21 @@ func (c *pipelineClient) WaitForActionToFinish(ctx context.Context, pipelineName
...
 func (c *pipelineClient) WaitForActionProgressing(ctx context.Context, pipelineName, namespace, actionName string, timeout time.Duration) error {
-	t := time.NewTicker(1 * time.Second)
-	to := time.NewTicker(timeout)
+	t := c.tickerWithTimeoutProvider(1*time.Second, timeout)
 
 	for {
 		select {
-		case <-to.C:
+		case <-t.TimeoutTick():
 			oktetoLog.Infof("action '%s' didn't progress after %s", actionName, timeout.String())
```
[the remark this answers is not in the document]
AgustinRamiroDiaz: Adding to this, we could also inject the logger to the `pipelineClient`

pkg/time/tickerwithtimeout.go
```diff
@@ -0,0 +1,39 @@
+// Copyright 2023 The Okteto Authors
```
andreafalzetti, line 1: should we name this `ticker_timeout.go` to make it easier to read?

AgustinRamiroDiaz: There are some mixed opinions on file naming, some say it must be only lower case according to some golang conventions. Particularly, I find underscores useful

pkg/okteto/action_test.go
```diff
@@ -0,0 +1,373 @@
...
+					},
+					errs: []error{},
+				},
+			},
+			expected: expected{
+				err: nil,
+			},
+		},
+		{
+			name: " successful",
```
andreafalzetti, line 337:
```suggestion
			name: "successful",
```

pkg/okteto/pipeline.go
andreafalzetti: Interesting discussion, I agree that it would be nicer to remove the abstraction/wrapper around the native `time` and use dependency injection as much as possible. I haven't tried refactoring this PR using that approach so I am not sure it's worth it or it's better to keep the provider approach.

In a much simpler scenario, I've written a function that allows injecting both `ticker` and `timeout` so it's not hard to mock in testing: https://github.com/okteto/okteto/pull/3455#discussion_r1142796151

Having said this, I am not against merging this which is already an improvement. We can always iterate on this later, if we want.

pkg/okteto/client.go
```diff
@@ -245,7 +251,7 @@ func InDevContainer() bool {
 	return false
 }
 
-func query(ctx context.Context, query interface{}, variables map[string]interface{}, client *graphql.Client) error {
+func query(ctx context.Context, query interface{}, variables map[string]interface{}, client graphqlClientInterface) error {
 	err := client.Query(ctx, query, variables)
 	if err != nil {
```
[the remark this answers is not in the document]
jLopezbarb: We have two different errors. The first one is the original error, the second one is the retry error. If the error is transient and is not nil we should not return error. That's why we check errors twice

pkg/okteto/action.go
```diff
@@ -38,19 +42,20 @@ func (c *pipelineClient) GetAction(ctx context.Context, name, namespace string)
...
-			return fmt.Errorf("'%s' didn't finish after %s", pipelineName, timeout.String())
-		case <-t.C:
+			return pipelineTimeoutError{
+				pipelineName: actionName,
+				timeout:      timeout,
+			}
+		case <-t.TickerTick():
 			a, err := c.GetAction(ctx, actionName, namespace)
 			if err != nil {
 				oktetoLog.Infof("action '%s' failed", actionName)
```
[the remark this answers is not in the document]
jLopezbarb: `Infof` is the one to add it to the log file. We should probably need to review all the logs and have two different loggers: the one to write into the log file, and the one to write in the stdout

pkg/okteto/action.go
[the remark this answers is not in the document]
jLopezbarb: Right now  we can't add the logger until we refactor that part of the code to create it in the `main.go` and pass it down. We can't ensure that the logger is initialised if we inject it directly. We need to think how to address that part of the code before doing such a big change

pkg/okteto/pipeline.go
jLopezbarb: I get what you say but I need to have the timer for the timeout and the ticker sync between them to know how many times should I call the ticker before calling the timeout function. I think all this logic  can be refactored into a smaller approach. Let me think about it and will come back with a new approach!!

pkg/okteto/client_test.go
jLopezbarb: The plan is that `fakeGraphQLClient` is reused across all the files and not only actions, i.e: `secrets_test.go`, `auth_test.go`...

pkg/time/tickerwithtimeout.go
jLopezbarb: I removed the whole pkg so we don't have this problem anymore 😜

pkg/okteto/pipeline.go
jLopezbarb: Done at [9cbd78b](https://github.com/okteto/okteto/pull/3458/commits/9cbd78bdf749d871e97d6afc1b5f8f0251633981)

AdrianPedriza (approved):

pkg/okteto/action_test.go
```diff
@@ -212,16 +190,34 @@ func TestWaitForActionToFinish(t *testing.T) {
...
+
+			go func() {
+				for i := 0; i < len(tc.cfg.client.queryResults); i++ {
+					tickerChannel <- time.Now()
+					// Wait until the ticker has been read
+					for {
+						if len(tickerChannel) == 0 {
+							break
+						}
+					}
```
AgustinRamiroDiaz, line 218: I think this is not needed since _"By default, sends and receives block until the other side is ready. This allows goroutines to synchronize without explicit locks or condition variables."_

AgustinRamiroDiaz (approved):
Awesome refactor and tests :100:

pkg/time/tickerwithtimeout.go
andreafalzetti: Thanks for suggesting that @AgustinRamiroDiaz, I've added an item (cc'd you on Notion). Happy to see feedback!

Issue #3410, opened 2023-02-28
Follow up issue for removing the credentials token here and server side


AgustinRamiroDiaz:
Create follow up issue for removing the credentials token here and server side

_Originally posted by @AgustinRamiroDiaz in https://github.com/okteto/okteto/pull/3409#discussion_r1120122687_

PR #3461, opened 2023-03-20
Remove Token from credentials query

AgustinRamiroDiaz:
# Proposed changes

This PR should get merged in some time, I'm leaving it here to not lose context and have the work in place

Fixes https://github.com/okteto/okteto/issues/3410

Removes the usage of Token from the backend Credentials (used for authentication to the k8s cluster). Builds on top of https://github.com/okteto/okteto/pull/3409

Should be merged after the clusters have been upgraded to have the kubetoken capability

Open discussion: Should we not handle the case when the okteto cluster version is old and doesn't handle the kubetoken (i.e. retrocompatibility)? Maybe it's something to be done later, so we can remove the "legacy" code around checking for the capability

### DIFF
--- a/cmd/context/create.go
+++ b/cmd/context/create.go
@@ -17,6 +17,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"net/http"
 	"net/url"
 	"os"
 	"strings"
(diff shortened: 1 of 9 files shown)


jLopezbarb:
>  Should we not handle the case when the okteto cluster version is old and doesn't handle the kubetoken (i.e. retrocompatibility)?

I think so. This change is a breaking change and will enforce everyone to upgrade to the latest okteto installation (something that maybe it's not planned by the teams using okteto). I think we'll need to figure out how can this two token retrievals live together.

teresaromero:
@AgustinRamiroDiaz can you update on the status/need for this PR?

AgustinRamiroDiaz:
@teresaromero yes. Closing this since the feature is being re-thought

PR #3462, opened 2023-03-21
chore(deploy): better err on dev deployment fail

### LINKED ISSUE #3419: okteto deploy doesn't log an error if the image is not accesible
**Describe the bug**
image

This is how it looks with debug logs.

image

We should show the reason of the error even when -l=debug is not enabled. 

**To Reproduce**

1. Set up an `image` that doesn't exist on the `deploy` command
1. `okteto deploy`

- OS: macOS
- Version 2.13.0

andreafalzetti:
# Proposed changes

Fixes #3419

When the user overrides the deploy image, but the image is not accessible (eg. does not exist or the user doesn't have enough permissions to push) the CLI fails with an error that does not give away much. To figure out the issue, one has to run again enabling debug logs.

Proposed change:
1. propagate the manifest from the deployOptions to the buildOptions so that we can know the image name which failed
2. always wrap the current error with its nested cause.

### How to test

1. Clone `movies-frontend`
```bash
git clone git@github.com:okteto/movies-frontend.git
```
2. Change the `deploy` block of the Okteto manifest to 
```yaml
deploy:
  image: ciaociaociaociaociaociao/ciao:1
  commands:
    - name: Deploy Frontend
```
3. Build the CLI on this branch
4. Run `okteto deploy` 
5. Observe the error being: `Error during development environment deployment: error building image 'ciaociaociaociaociaociao/ciao:1': Access Denied (the repository does not exist or you are not authorized to push a new image).`

| Before | After |
|--------|-------|
|        |        |


cmd/deploy/remote.go
```diff
@@ -140,14 +141,15 @@ func (rd *remoteDeployCommand) deploy(ctx context.Context, deployOptions *Option
...
+	buildOptions.Manifest = deployOptions.Manifest
 
 	// we need to call Build() method using a remote builder. This Builder will have
 	// the same behavior as the V1 builder but with a different output taking into
 	// account that we must not confuse the user with build messages since this logic is
 	// executed in the deploy command.
 	if err := rd.builderV1.Build(ctx, buildOptions); err != nil {
 		return oktetoErrors.UserError{
-			E: fmt.Errorf("Error during development environment deployment."),
+			E: errors.Wrap(err, "Error during development environment deployment"),
```
jLopezbarb, line 152: [github.com/pkg/errors](https://github.com/pkg/errors) is not part of the standard library and was replaced with the extended notation in [Go 1.13.](https://go.dev/blog/go1.13-errors)
```suggestion
			E: fmt.Errorf("Error during development environment deployment: %w", err),
```

andreafalzetti: thanks! Updated ✅

pkg/cmd/build/errors.go
```diff
@@ -32,7 +32,7 @@ func getErrorMessage(err error, tag string) error {
 	switch {
 	case isLoggedIntoRegistryButDontHavePermissions(err):
 		err = oktetoErrors.UserError{
-			E:    fmt.Errorf("error building image '%s': You are not authorized to push image '%s'", tag, imageTag),
+			E:    fmt.Errorf("error building image '%s': Access Denied (the repository does not exist or you are not authorized to push a new image).", imageTag),
```
jLopezbarb, line 35: NIT: Maybe we should create an error for `Access Denied (the repository does not exist or you are not authorized to push a new image).` and check with `errors.Is()` in `remote.go` to return the proper message.
```go
accessDeniedErr = errors.New("Access Denied (the repository does not exist or you are not authorized to push a new image).")
err = oktetoErrors.UserError{
			E:    fmt.Errorf("error building image '%s': %w", imageTag, accessDeniedErr),
```

This code is shared with image builds and since the remote is not pushing any image can result in misleading error text. WDYT?

pkg/cmd/build/authprovider_test.go
```diff
@@ -9,7 +9,7 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func Test_isErrCredentialsHelperNotAccessiblee(t *testing.T) {
+func Test_isErrCredentialsHelperNotAccessible(t *testing.T) {
```
andreafalzetti, line 12: typo double "e"

pkg/cmd/build/errors.go
```diff
@@ -86,7 +91,7 @@ func isTransientError(err error) bool {
 
 // IsLoggedIntoRegistryButDontHavePermissions returns true when the error is because the user is logged into the registry but doesn't have permissions to push the image
 func isLoggedIntoRegistryButDontHavePermissions(err error) bool {
-	return strings.Contains(err.Error(), "insufficient_scope: authorization failed")
+	return strings.Contains(err.Error(), "insufficient_scope: authorization failed") && !isPullAccessDenied(err)
```
andreafalzetti, line 94: the full errror for Pull Access Denied includes "insufficient_scope: authorization failed" so I made explicit that `isLoggedIntoRegistryButDontHavePermissions` should ignore that part.

We could move above the check for `isPullAccessDenied` but it seems better excluding IMHO.

pkg/cmd/build/errors_test.go
```diff
@@ -0,0 +1,181 @@
+package build
```
andreafalzetti, line 1: I've added some boring tests here, not that they add huge value but it feels the right thing to do.

pkg/cmd/build/errors.go
andreafalzetti: @jLopezbarb I've updated the PR according to our chat :) Please have another look 🙏

jLopezbarb (approved):

PR #3463, opened 2023-03-22
Jlopezbarb/add preview graphql tests

jLopezbarb:
# Proposed changes

- Add tests for the different functions for preview client
- Move `validateNamespace` to preview to preview file since it's the only one using it
- Remove `GetPreviewEnvByName` (unused)
- Create custom errors for preview


pkg/okteto/preview.go
```diff
@@ -337,13 +295,35 @@ func (c *previewClient) GetResourcesStatusFromPreview(ctx context.Context, previ
...
 	if strings.Contains(err.Error(), "operation-not-permitted") {
-		return oktetoErrors.UserError{E: fmt.Errorf("you are not authorized to create a global preview env"),
+		return oktetoErrors.UserError{E: ErrUnauthorizedGlobalCreation,
 			Hint: "Please log in with an administrator account or use a personal preview environment"}
 	}
 	return err
 }
+
+// validateNamespace checks if the new ns exceeds the max chars allowed by k8s and if t
+func validateNamespace(namespace, object string) error {
```
andreafalzetti, line 310: To clean things even further we could remove the global var `MAX_ALLOWED_CHARS` and define it internally to this function. The error message now has hardcoded `63` which could also be interpolated with the variable.

Alternatively, we could pass the validation parameters as input

Pseudo-code:

```go
type namespaceValidation struct {
  max_allowed_chars int
  nameValidationRegex string
}

&namespaceValidation{
  max_allowed_chars: 63,
  nameValidationRegex: "^[a-z0-9]([-a-z0-9]*[a-z0-9])?$",
}

func validateNamespace(validation namespaceValidation, namespace, object string) error {
```

This would also make testing more reliable

andreafalzetti (approved):
Great improvements, great job!

AgustinRamiroDiaz (commented):
Awesome refactor! Love the typed errors

PR #3468, opened 2023-03-22
tests: remove unused file credential.go

jLopezbarb:
# Proposed changes

- Remove `credential.go` since it's no longer used


AgustinRamiroDiaz (approved):

andreafalzetti (approved):
@jLopezbarb How was `credential.go` used before and what replaced it?

jLopezbarb:
> @jLopezbarb How was `credential.go` used before and what replaced it?

It was used long time ago but we couldn't detect the unused function because it was a struct method and that is not check by the linter

AgustinRamiroDiaz:
> > @jLopezbarb How was `credential.go` used before and what replaced it?
> 
> It was used long time ago but we couldn't detect the unused function because it was a struct method and that is not check by the linter

I couldn't find linters for unused exported methods/functions, it would be great to have some

PR #3471, opened 2023-03-23
add pipeline gql tests

jLopezbarb:
# Proposed changes

- Add tests for `pipeline.go` graphql
- Add `repository.IsEqual` to the right package (with its test)


pkg/repository/repository.go
```diff
@@ -89,8 +94,13 @@ type gitStatusInterface interface {
 }
 
 func NewRepository(path string) Repository {
+	url, err := giturls.Parse(path)
+	if err != nil {
+		oktetoLog.Infof("could not parse url: %w", err)
```
andreafalzetti, line 99: Why do we use Info level for errors?

pkg/okteto/pipeline_test.go
```diff
@@ -14,9 +14,760 @@
...
+			expected: expected{
+				response: map[string]string{
+					"deployment/error":  "error",
+					"statefulset/error": "error",
+					"job/error":         "error",
+					"cronjob/error":     "error",
+				},
+				err: nil,
+			},
+		},
```
andreafalzetti, line 757: Should we add a check for the not found?

```go
{
    name: "not found",
    input: input{
        client: &fakeGraphQLClient{
            queryResult: nil,
            err:         oktetoErrors.ErrNotFound,
        },
        namespace:  "test",
        devenvName: "non-existing",
    },
    expected: expected{
        response: nil,
        err:      fmt.Errorf("could not create okteto client"),
    },
},
```

Altough I think this will fail to compare the error (I am curious if you know how to fix it?)

Also, we could add a test for the retry mechanism (for any non-notfound errors but I guess we would need to extend the client mock to return once a failure and next a good result, so maybe it's not worth it now)

andreafalzetti (approved):
Great work!

pkg/repository/repository.go
jLopezbarb: `oktetoLog.Infof` logs the error into the log file. If we use another level for the errors, we'll show the log to the user.

pkg/okteto/pipeline_test.go
jLopezbarb: Done. The way to do it is extract the error to a variable containing the message. That way whenever we compare the error with `errors.Is` we'll be pointing to the same error and we won't need to use the string comparison

pkg/repository/repository.go
andreafalzetti: Understood! Thanks @jLopezbarb

pkg/okteto/pipeline.go
```diff
@@ -43,28 +42,73 @@ func newPipelineClient(client *graphql.Client, url string) *pipelineClient {
...
-				}
-				GitDeploy struct {
-					Id         graphql.String
-					Name       graphql.String
-					Status     graphql.String
-					Repository graphql.String
-				}
-			} `graphql:"deployGitRepository(name: $name, repository: $repository, space: $space, branch: $branch, variables: $variables, filename: $filename)"`
-		}
+		var mutation deployPipelineMutation
```
AdrianPedriza, line 111:
```suggestion
		var deployMutation deployPipelineMutation
```
I think it improves the reading to add more info about the mutation.

pkg/okteto/pipeline.go
```diff
@@ -101,34 +145,19 @@ func (c *pipelineClient) Deploy(ctx context.Context, opts types.PipelineDeployOp
...
-				GitDeploy struct {
-					Id         graphql.String
-					Name       graphql.String
-					Status     graphql.String
-					Repository graphql.String
-				}
-			} `graphql:"deployGitRepository(name: $name, repository: $repository, space: $space, branch: $branch, filename: $filename, variables: $variables)"`
-		}
-
+		var mutation deployPipelineMutation
```
AdrianPedriza, line 160:
```suggestion
		var deployMutation deployPipelineMutation
```

pkg/okteto/pipeline.go
```diff
@@ -321,32 +257,28 @@ func (c *pipelineClient) Destroy(ctx context.Context, name, namespace string, de
...
 	oktetoLog.Infof("destroy pipeline: %s/%s", namespace, name)
 	gitDeployResponse := &types.GitDeployResponse{}
 	if destroyVolumes {
-		var mutation struct {
-			GitDeployResponse struct {
-				Id     graphql.String
-				Status graphql.String
-			} `graphql:"destroyGitRepository(name: $name, space: $space, destroyVolumes: $destroyVolumes)"`
-		}
+		var mutation deprecatedDestroyPipelineWithVolumesMutation
```
AdrianPedriza, line 281:
```suggestion
		var destroyPipelineWithVolumesMutation deprecatedDestroyPipelineWithVolumesMutation
```

pkg/okteto/pipeline.go
```diff
@@ -282,33 +233,18 @@ func (c *pipelineClient) Destroy(ctx context.Context, name, namespace string, de
...
-				GitDeploy struct {
-					Id         graphql.String
-					Name       graphql.String
-					Status     graphql.String
-					Repository graphql.String
-				}
-			} `graphql:"destroyGitRepository(name: $name, space: $space)"`
-		}
-
+		var mutation destroyPipelineWithoutVolumesMutation
```
AdrianPedriza, line 247:
```suggestion
		var destroyPipelineWithoutVolumesMutation destroyPipelineWithoutVolumesMutation
```

pkg/okteto/pipeline.go
```diff
@@ -192,82 +213,12 @@ func (c *pipelineClient) GetByName(ctx context.Context, name, namespace string)
...
-				}
-				GitDeploy struct {
-					Id         graphql.String
-					Name       graphql.String
-					Status     graphql.String
-					Repository graphql.String
-				}
-			} `graphql:"destroyGitRepository(name: $name, space: $space, destroyVolumes: $destroyVolumes)"`
-		}
+		var mutation destroyPipelineWithVolumesMutation
```
AdrianPedriza, line 221:
```suggestion
		var destroyPipelineWithVolumesMutation destroyPipelineWithVolumesMutation
```

PR #3473, opened 2023-03-23
tests: add auth unit test

jLopezbarb:
# Proposed changes

- Add tests for the auth graphql flow

### DIFF
--- a/pkg/cmd/login/browser.go
+++ b/pkg/cmd/login/browser.go
@@ -47,7 +47,12 @@ func (h *Handler) handle() http.Handler {
 		}
 
 		ctx := r.Context()
-		u, err := okteto.Auth(ctx, code, h.baseURL)
+		oktetoClient, err := okteto.NewOktetoClientFromUrl(h.baseURL)
+		if err != nil {
+			h.errChan <- err
+			return
+		}
+		u, err := oktetoClient.Auth(ctx, code)
 		if err != nil {
 			if err := html.ExecuteError(w, err); err != nil {
 				h.errChan <- err
--- a/pkg/okteto/auth.go
+++ b/pkg/okteto/auth.go
@@ -15,13 +15,9 @@ package okteto
 
 import (
 	"context"
-	"encoding/json"
-	"fmt"
-	"os"
-	"regexp"
+	"errors"
 	"strings"
 
-	"github.com/okteto/okteto/pkg/config"
 	"github.com/okteto/okteto/pkg/constants"
 	oktetoErrors "github.com/okteto/okteto/pkg/errors"
 	oktetoLog "github.com/okteto/okteto/pkg/log"
@@ -29,7 +25,7 @@ import (
 	"github.com/shurcooL/graphql"
 )
 
-var reg = regexp.MustCompile("[^A-Za-z0-9]+")
+const cliSource = "cli"
 
 // Token contains the auth token and the URL it belongs to
 type Token struct {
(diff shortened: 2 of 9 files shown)


pkg/cmd/login/browser.go
```diff
@@ -47,7 +47,12 @@ func (h *Handler) handle() http.Handler {
 		}
 
 		ctx := r.Context()
-		u, err := okteto.Auth(ctx, code, h.baseURL)
+		oktetoClient, err := okteto.NewOktetoClientFromUrl(h.baseURL)
```
AdrianPedriza, line 50: I think func `NewOktetoClientFromUrl` has enough logic to add some tests for it? :)

pkg/okteto/auth.go
```diff
@@ -66,34 +112,19 @@ func Auth(ctx context.Context, code, url string) (*types.User, error) {
...
-			Token           graphql.String
-			New             graphql.Boolean
-			Registry        graphql.String
-			Buildkit        graphql.String
-			Certificate     graphql.String
-			GlobalNamespace graphql.String  `graphql:"globalNamespace"`
-			Analytics       graphql.Boolean `graphql:"telemetryEnabled"`
-		} `graphql:"auth(code: $code, source: $source)"`
-	}
+	var mutation authMutationStruct
```
AdrianPedriza, line 123:
```suggestion
	var authMutation authMutationStruct
```
I think thgis improves reading it adding more context

pkg/okteto/auth.go
```diff
@@ -107,46 +138,33 @@ func (c *OktetoClient) authUser(ctx context.Context, code string) (*types.User,
...
-			ExternalID  graphql.String `graphql:"externalID"`
-			Token       graphql.String
-			New         graphql.Boolean
-			Registry    graphql.String
-			Buildkit    graphql.String
-			Certificate graphql.String
-		} `graphql:"auth(code: $code, source: $source)"`
-	}
-
+	var mutation deprecatedAuthMutationStruct
```
AdrianPedriza, line 164:
```suggestion
	var deprecatedAuthMutation deprecatedAuthMutationStruct
```

pkg/okteto/auth.go
```diff
@@ -42,18 +38,68 @@ type Token struct {
...
+}
+
+func (authenticationErr) Error() string {
+	return "authentication error, please try again"
+}
+func (e authenticationErr) Unwrap() error {
+	return e.E
+}
+
+func newAuthenticationErr(err error) authenticationErr {
```
AdrianPedriza, line 52: I don't think we need this function, do we? We could use `authenticationErr` directly ?

pkg/okteto/context.go
```diff
@@ -55,6 +56,7 @@ const (
 
 var (
 	CurrentStore *OktetoContextStore
+	reg          = regexp.MustCompile("[^A-Za-z0-9]+")
```
AdrianPedriza, line 59: do we really need this as global var?

pkg/cmd/login/browser.go
AdrianPedriza: NIT: I see this func was added before your changes

AdrianPedriza:
minor comments, LGTM! 👍

AdrianPedriza (approved):

pkg/okteto/context.go
ifbyol: In that way the regex is compiled one once instead of each time it has to be used

AdrianPedriza: makes sense!

ifbyol (approved):

pkg/cmd/login/browser.go
jLopezbarb: That should be addressed when we do the refactor for the `client.go`

PR #3475, opened 2023-03-24
don't initialize okteto client if not using okteto cloud

etrombly:
# Proposed changes

Fixes # 3401

Tested to work on a local k8s cluster, I don't have an okteto cloud account setup, but shouldn't impact it.

### DIFF
--- a/cmd/destroy/destroy.go
+++ b/cmd/destroy/destroy.go
@@ -135,18 +135,18 @@ func Destroy(ctx context.Context) *cobra.Command {
 				return fmt.Errorf("failed to get the current working directory: %w", err)
 			}
 
+			name := options.Name
 			if options.Name == "" {
 				c, _, err := okteto.NewK8sClientProvider().Provide(okteto.Context().Cfg)
 				if err != nil {
 					return err
 				}
 				inferer := devenvironment.NewNameInferer(c)
-				options.Name = inferer.InferName(ctx, cwd, okteto.Context().Namespace, options.ManifestPathFlag)
+				name = inferer.InferName(ctx, cwd, okteto.Context().Namespace, options.ManifestPathFlag)
 				if err != nil {
 					return fmt.Errorf("could not infer environment name")
 				}
 			}
-			name := options.Name
 
 			dynClient, _, err := okteto.GetDynamicClient()
 			if err != nil {
--- a/cmd/destroy/destroy.go
+++ b/cmd/destroy/destroy.go
@@ -134,18 +134,19 @@ func Destroy(ctx context.Context) *cobra.Command {
 			if err != nil {
 				return fmt.Errorf("failed to get the current working directory: %w", err)
 			}
-			name := options.Name
+
 			if options.Name == "" {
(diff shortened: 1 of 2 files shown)


jLopezbarb (approved):

jLopezbarb:
Hi @etrombly, Changes looks good. I've send you an invitation to Okteto cloud to allow you to test the PR. Let me know if you can't access or have any problem

etrombly:
It wouldn't let me respond on the review:

I'm still setting name from options.Name, I just do it under the block that checks if options.Name is empty and initializes it. I was running into it being empty.

Honestly we probably don't even need the name variable and could just use options.Name directly.

jLopezbarb:
> I'm still setting name from options.Name, I just do it under the block that checks if options.Name is empty and initializes it. I was running into it being empty in the notification that the dev environment was deleted.

Yes, but it's modifying the `options.Name` which is needed later on and need to check if the `options.Name` is set or not. With the change you are proposing it'll always be set.

> If options.Name shouldn't be changed, then I need to see where the deletion notification is being set from, and have that use the name variable instead of options.Name.

Which delete notification? Maybe we can include another struct in the command to achieve that, but I want to understand the reason on changing the `options.Name` to be always set

etrombly:
> Which delete notification? Maybe we can include another struct in the command to achieve that, but I want to understand the reason on changing the `options.Name` to be always set

The output of `okteto destroy` looks like this without the change:

```
 i  Using eric-trombly @ dev/eric.trombly@XXXX as context
 ✓  Development environment '' successfully destroyed
```

jLopezbarb:
Let's do it in a separate issue. We'll need to check which implications it has before making the change. WDYT?

etrombly:
Sounds good, I reverted the change to options.Name and will open a new issue

> Hi @etrombly, Changes looks good. I've send you an invitation to Okteto cloud to allow you to test the PR. Let me know if you can't access or have any problem

I haven't received the invitation. It may be my work spam filter, can you try sending it to etrombly@gmail.com?

AgustinRamiroDiaz (approved):

jLopezbarb:
> Sounds good, I reverted the change to options.Name and will open a new issue
> 
> > Hi @etrombly, Changes looks good. I've send you an invitation to Okteto cloud to allow you to test the PR. Let me know if you can't access or have any problem
> 
> I haven't received the invitation. It may be my work spam filter, can you try sending it to [etrombly@gmail.com](mailto:etrombly@gmail.com)?

Can you try to login into cloud with this GH account. You should be able to login with it

etrombly:
Looks like it's working properly in okteto cloud as well:
```
 eric > ~ > git > rust-getting-started > master > $ > /home/eric/git/okteto/bin/okteto destroy
 i  Using etrombly @ cloud.okteto.com as context
 ✓  Development environment '' successfully destroyed
 ```

jLopezbarb (approved):

jLopezbarb:
Thanks @etrombly. This is a very nice addition!! It'll be ready for the next release

PR #3477, opened 2023-03-24
fix: destroy command should expand env vars

jLopezbarb:
# Proposed changes

Fixes https://github.com/okteto/app/issues/6100

- Remove command variable expansion (is already done in the executor)
- Couldn't add unit test since the command executor opens a new connection

## Tested scenarios
1. Create the following manifest:
```yaml
name: test-pipeline

#build:
#  context: .

#dependencies:
#  test:
#    repository: http://nonexistentrepo/okteto/test.git
#    wait: true

deploy:
  commands:
    - name: test variable scripting
      command: |
        HELLO_WORLD="hi!"
        echo $HELLO_WORLD

destroy:
  commands:
    - name: test variable scripting
      command: |
        ANOTHER_VAR="this is another variable"
        echo $ANOTHER_VAR
```
2. Run `okteto destroy`
3. See that variable is actually expanded

### DIFF
--- a/pkg/model/manifest.go
+++ b/pkg/model/manifest.go
@@ -948,13 +948,6 @@ func (manifest *Manifest) ExpandEnvVars() error {
 				return err
 			}
 		}
-		for idx, cmd := range manifest.Destroy.Commands {
-			cmd.Command, err = envsubst.String(cmd.Command)
-			if err != nil {
-				return errors.New("could not parse env vars")
-			}
-			manifest.Destroy.Commands[idx] = cmd
-		}
 	}
 
 	for devName, devInfo := range manifest.Dev {
--- a/pkg/model/manifest_test.go
+++ b/pkg/model/manifest_test.go
@@ -152,6 +152,36 @@ func TestManifestExpandDevEnvs(t *testing.T) {
 				},
 			},
 		},
+		{
+			name: "does not expand vars in destroy command",
+			envs: map[string]string{
+				"TEST_VAR": "test",
+			},
+			manifest: &Manifest{
+				Destroy: &DestroyInfo{
+					Image: "",
+					Commands: []DeployCommand{
+						{
+							Name: "test",
+							Command: `TEST_VAR="do-not-expand-me"
+echo $TEST_VAR`,
+						},
+					},
+				},
+			},
+			expectedManifest: &Manifest{
+				Destroy: &DestroyInfo{
+					Image: "",
+					Commands: []DeployCommand{
+						{
+							Name: "test",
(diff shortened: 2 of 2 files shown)


pkg/model/manifest.go
```diff
@@ -948,13 +948,6 @@ func (manifest *Manifest) ExpandEnvVars() error {
 				return err
 			}
 		}
-		for idx, cmd := range manifest.Destroy.Commands {
```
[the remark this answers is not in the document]
andreafalzetti: I think it's a good idea, maybe it might sound being extra careful but I agree that's what we need to make the product super stable and avoid regressions. This is a test that should cover the removal:

```go
{
    name: "does not expand vars in destroy command",
    envs: map[string]string{
        "TEST_VAR": "test",
    },
    manifest: &Manifest{
        Destroy: &DestroyInfo{
            Image: "",
            Commands: []DeployCommand{
                {
                    Name: "test",
                    Command: `TEST_VAR="do-not-expand-me"
echo $TEST_VAR`,
                },
            },
        },
    },
    expectedManifest: &Manifest{
        Destroy: &DestroyInfo{
            Image: "",
            Commands: []DeployCommand{
                {
                    Name: "test",
                    Command: `TEST_VAR="do-not-expand-me"
echo $TEST_VAR`,
                },
            },
        },
    },
```

pkg/model/manifest.go
```diff
@@ -948,13 +948,6 @@ func (manifest *Manifest) ExpandEnvVars() error {
...
-			cmd.Command, err = envsubst.String(cmd.Command)
-			if err != nil {
-				return errors.New("could not parse env vars")
-			}
-			manifest.Destroy.Commands[idx] = cmd
```
pchico83, line 956: @jLopezbarb I am curious if there was a reason to have this code in place

ifbyol: @pchico83 @jLopezbarb This PR was the first one where we included the env var expansasion for deploy and destroy commands: https://github.com/okteto/okteto/pull/2234

Then, in this PR we changed the approach to not expand the env vars in deploy commands, but we left the destroy commands: https://github.com/okteto/okteto/pull/2653

I think it was the first attempt of having variable expansion for build env vars instead of injecting them in the environment when executing the commands. At some point, we had some error in deploy commands and we removed it, but we didn't do the same for destroy commands

pchico83: Thanks @ifbyol for this brilliant analysis :-)
That gives us even more confidence that this change is totally safe.

AdrianPedriza (approved):

ifbyol (approved):

PR #3478, opened 2023-03-24
Always pull & export default cache when building an image

### LINKED ISSUE #3426: new `cache_from` logic when building an image
## Parent epic/feature

## Description

In order to optimize the builds so that if a previous image has already been created, when building an image we will add to the `cache_from` field the appropriate registries according to the scenarios where we are. This implementation will not modify the manifest but internally it will take into account the proper registries to used them as cache when building an image.
## Acceptance criteria

- [ ] if global registry is active, adds in `cache_from` field a new entry pointing to global registry using current git commit (no matter if commit is dirty) 
- [ ] adds in `cache_from` field a new entry pointing to dev registry using `cache` as image tag.

#### Scenario 1: Global registry active

1. Create a new namespace
2. Create the following manifest:
```yaml
build:
  frontend:
    context: frontend
    dockerfile: Dockerfile
    image: okteto.dev/frontend:1.0.0
```
3. Run the following build command: `okteto build` (this will use cache from `okteto.dev/frontend:cache` and `okteto.global/frontend:cache` registry)
4. Internally it transform the manifest to:
```yaml
build:
  frontend:
    context: frontend
    dockerfile: Dockerfile
    cache_from:
    - okteto.global/frontend:cache
    - okteto.dev/frontend:cache
```

##### Scenario 2 - Global registry not active

1. Create a new namespace
2. Create the following manifest:
```yaml
build:
  frontend:
    context: frontend
    dockerfile: Dockerfile
    image: okteto.dev/frontend:1.0.0
```
3. Run the following build command: `okteto build` (this will use cache from `okteto.dev/frontend:cache`registry)
4. Internally it transform the manifest to:
```yaml
build:
  frontend:
    context: frontend
    dockerfile: Dockerfile
    cache_from:
    - okteto.dev/frontend:cache
```

##### Scenario 3 - Non okteto dev registry

1. Create a new namespace
2. Create the following manifest:
```yaml
build:
  frontend:
    context: frontend
    dockerfile: Dockerfile
    image: docker.io/my-user/frontend:1.0.0
```
3. Run the following build command: `okteto build`
4. Check that no cache is exported

andreafalzetti:
# Proposed changes

Fixes #3426

## How to test (Okteto Cloud)

1. `git clone git@github.com:okteto/movies-frontend.git`
2. Edit the `okteto.yml` to be like:
```YAML
build:
  frontend:
    context: .
    dockerfile: Dockerfile
    image: okteto.dev/frontend:1.0.0
    export_cache: okteto.dev/frontend:cache

deploy:
  - name: Deploy Frontend

dev:
  frontend:
    image: ${OKTETO_BUILD_FRONTEND_DEV_IMAGE}
    command: yarn start
    sync:
      - .:/src
```
3. Run `okd build -l=debug`
4. Notice how you get both errors on pulling the cache images (like:  => `ERROR importing cache manifest from registry.cloud.okteto.net/okteto/frontend:cache` for both `/okteto` and `/{your-username`)
5. When the build suceeds it will export the cache to `okteto.dev`, now run again the build command 
6. Notice how this time the cache from `/{your-username}` is pulled successfully with a message like ` => importing cache manifest from registry.cloud.okteto.net/{your-username}/frontend:cache`

## How to test (Dev Cluster)

1. Switch to your dev cluster context
2. Follow the same steps for testing Okteto Cloud with the difference that here you should be able to push and pull the cache from the Global Registry

## How to test (Vanilla Cluster)

1. Using Docker Desktop, turn on the Kubernetes cluster
4. Select that okteto context (ie. `okteto context use docker-desktop`)
5. Run `okd build -l=debug`
6. Notice how you don't see any errors of cache fail being pulled either being imported

### DIFF
--- a/cmd/build/command.go
+++ b/cmd/build/command.go
@@ -51,6 +51,10 @@ type registryInterface interface {
 	IsOktetoRegistry(image string) bool
 	GetImageReference(image string) (registry.OktetoImageReference, error)
 	HasGlobalPushAccess() (bool, error)
+	IsGlobalRegistry(image string) bool
+
+	GetRegistryAndRepo(image string) (string, string)
+	GetRepoNameAndTag(repo string) (string, string)
 }
 
 // NewBuildCommand creates a struct to run all build methods
@@ -109,7 +113,7 @@ func Build(ctx context.Context) *cobra.Command {
 	cmd.Flags().StringVarP(&options.Target, "target", "", "", "set the target build stage to build")
 	cmd.Flags().BoolVarP(&options.NoCache, "no-cache", "", false, "do not use cache when building the image")
 	cmd.Flags().StringArrayVar(&options.CacheFrom, "cache-from", nil, "cache source images")
-	cmd.Flags().StringVarP(&options.ExportCache, "export-cache", "", "", "export cache image")
+	cmd.Flags().StringArrayVar(&options.ExportCache, "export-cache", nil, "export cache images")
 	cmd.Flags().StringVarP(&options.OutputMode, "progress", "", oktetoLog.TTYFormat, "show plain/tty build output")
 	cmd.Flags().StringArrayVar(&options.BuildArgs, "build-arg", nil, "set build-time variables")
 	cmd.Flags().StringArrayVar(&options.Secrets, "secret", nil, "secret files exposed to the build. Format: id=mysecret,src=/local/secret")
--- a/pkg/cache/cache_from.go
+++ b/pkg/cache/cache_from.go
@@ -23,14 +23,12 @@ import (
 // CacheFrom is a list of images to import cache from.
 type CacheFrom []string
 
-type ImageCtrlInterface interface {
+type oktetoRegistryInterface interface {
+	HasGlobalPushAccess() (bool, error)
+	IsGlobalRegistry(image string) bool
+
 	GetRegistryAndRepo(image string) (string, string)
 	GetRepoNameAndTag(repo string) (string, string)
-}
-
-type oktetoRegistryInterface interface {
-	GetImageCtrl() ImageCtrlInterface
-	HasGlobalPushAccess() (bool, error)
 }
 
 // UnmarshalYAML implements the Unmarshaler interface of the yaml pkg.
@@ -68,17 +66,16 @@ func (cf *CacheFrom) AddDefaultPullCache(reg oktetoRegistryInterface, image stri
 		oktetoLog.Infof("error trying to access globalPushAccess: %w", err)
 	}
 
-	imageCtrl := reg.GetImageCtrl()
-	_, repo := imageCtrl.GetRegistryAndRepo(image)
-	imageName, _ := imageCtrl.GetRepoNameAndTag(repo)
+	_, repo := reg.GetRegistryAndRepo(image)
+	imageName, _ := reg.GetRepoNameAndTag(repo)
 
 	if hasAccess {
-		globalCacheImage := fmt.Sprintf("%s/%s:cache", constants.GlobalRegistry, imageName)
+		globalCacheImage := fmt.Sprintf("%s/%s:%s", constants.GlobalRegistry, imageName, defaultCacheTag)
 		cf.addCacheFromImage(globalCacheImage)
 		oktetoLog.Debugf("Dynamically adding cache_from: %s", globalCacheImage)
 	}
 
-	devCacheImage := fmt.Sprintf("%s/%s:cache", constants.DevRegistry, imageName)
+	devCacheImage := fmt.Sprintf("%s/%s:%s", constants.DevRegistry, imageName, defaultCacheTag)
 	cf.addCacheFromImage(devCacheImage)
 	oktetoLog.Debugf("Dynamically adding cache_from: %s", devCacheImage)
 }
--- a/pkg/cache/cache_from.go
+++ b/pkg/cache/cache_from.go
@@ -66,8 +66,8 @@ func (cf *CacheFrom) AddDefaultPullCache(reg oktetoRegistryInterface, image stri
 		oktetoLog.Infof("error trying to access globalPushAccess: %w", err)
 	}
 
-	_, repo := reg.GetRegistryAndRepo(image)
-	imageName, _ := reg.GetRepoNameAndTag(repo)
+	_, imageRepo := reg.GetRegistryAndRepo(image)
+	imageName, _ := reg.GetRepoNameAndTag(imageRepo)
 
 	if hasAccess {
 		globalCacheImage := fmt.Sprintf("%s/%s:%s", constants.GlobalRegistry, imageName, defaultCacheTag)
--- a/pkg/cache/cache_from.go
+++ b/pkg/cache/cache_from.go
@@ -1,0 +1,101 @@
+// Copyright 2023 The Okteto Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cache
+
+import (
+	"fmt"
+
+	"github.com/okteto/okteto/pkg/constants"
+	oktetoLog "github.com/okteto/okteto/pkg/log"
+)
+
+// CacheFrom is a list of images to import cache from.
+type CacheFrom []string
+
+type ImageCtrlInterface interface {
+	GetRegistryAndRepo(image string) (string, string)
+	GetRepoNameAndTag(repo string) (string, string)
+}
+
+type oktetoRegistryInterface interface {
+	GetImageCtrl() ImageCtrlInterface
+	HasGlobalPushAccess() (bool, error)
+}
+
+// UnmarshalYAML implements the Unmarshaler interface of the yaml pkg.
+func (cf *CacheFrom) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	var single string
+	err := unmarshal(&single)
+	if err == nil {
+		*cf = CacheFrom{single}
+		return nil
+	}
+
+	var multi []string
+	err = unmarshal(&multi)
+	if err == nil {
+		*cf = multi
+		return nil
+	}
+
+	return err
+}
+
+// MarshalYAML implements the marshaler interface of the yaml pkg.
+func (cf *CacheFrom) MarshalYAML() (interface{}, error) {
+	if len(*cf) == 1 {
+		return (*cf)[0], nil
+	}
+
+	return cf, nil
+}
+
+// AddDefaultPullCache appends the default cache layers for a given image
+func (cf *CacheFrom) AddDefaultPullCache(reg oktetoRegistryInterface, image string) {
+	hasAccess, err := reg.HasGlobalPushAccess()
+	if err != nil {
+		oktetoLog.Infof("error trying to access globalPushAccess: %w", err)
+	}
+
+	imageCtrl := reg.GetImageCtrl()
+	_, repo := imageCtrl.GetRegistryAndRepo(image)
+	imageName, _ := imageCtrl.GetRepoNameAndTag(repo)
+
+	if hasAccess {
+		globalCacheImage := fmt.Sprintf("%s/%s:cache", constants.GlobalRegistry, imageName)
+		cf.addCacheFromImage(globalCacheImage)
+		oktetoLog.Debugf("Dynamically adding cache_from: %s", globalCacheImage)
+	}
+
+	devCacheImage := fmt.Sprintf("%s/%s:cache", constants.DevRegistry, imageName)
+	cf.addCacheFromImage(devCacheImage)
+	oktetoLog.Debugf("Dynamically adding cache_from: %s", devCacheImage)
+}
+
+// addCacheFromImage appends a cache image to the list if it's not already there
+func (cf *CacheFrom) addCacheFromImage(imageName string) {
+	if !cf.hasCacheFromImage(imageName) {
+		*cf = append(*cf, imageName)
+	}
+}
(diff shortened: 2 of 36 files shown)


cmd/build/command.go
```diff
@@ -138,6 +139,9 @@ func (bc *Command) getBuilder(options *types.BuildOptions) (Builder, error) {
 		}
 	}
 
+	// TODO: check if global registry is active
+	cache.AddDefaultPullCache(manifest)
```
jLopezbarb, line 143: As we talked offline, the okteto image can be empty at this point

go.mod
```diff
@@ -224,6 +224,7 @@ require (
 )
 
 require (
+	gotest.tools v2.2.0+incompatible
```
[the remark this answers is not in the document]
andreafalzetti: auto-imported by the IDE by mistake, removed now

cmd/build/command.go
andreafalzetti: I moved the logic 👍

pkg/cache/cache_from.go
```diff
@@ -0,0 +1,50 @@
...
+
+func AddDefaultPullCache(m *model.Manifest) {
+	for _, build := range m.Build {
+		// Extract the registry and image name from the original Docker image
+		imageCtrl := registry.NewImageCtrl(okteto.Config{})
+		registry, repo := imageCtrl.GetRegistryAndRepo(build.Image)
+		image, _ := imageCtrl.GetRepoNameAndTag(repo)
+
+		// Construct the cache image tag
+		cacheImage := fmt.Sprintf("%s/%s:cache", registry, image)
```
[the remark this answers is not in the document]
andreafalzetti: Updated!

pkg/cache/cache_from.go
```diff
@@ -0,0 +1,51 @@
...
+
+import (
+	"fmt"
+
+	"github.com/okteto/okteto/pkg/model"
+	"github.com/okteto/okteto/pkg/okteto"
+	"github.com/okteto/okteto/pkg/registry"
+)
+
+func AddDefaultPullCache(b *model.BuildInfo) {
```
jLopezbarb, line 24: NIT: I think all these functions should be part of type `CacheFrom`. Maybe we can move the definition and serialisation here. WDYT?

pkg/cache/cache_from.go
```diff
@@ -0,0 +1,51 @@
...
+	// Extract the registry and image name from the original Docker image
+	imageCtrl := registry.NewImageCtrl(okteto.Config{})
+	_, repo := imageCtrl.GetRegistryAndRepo(b.Image)
+	image, _ := imageCtrl.GetRepoNameAndTag(repo)
+
+	// Construct the cache image tags
```
jLopezbarb, line 30: We should only add the global registry whenever the user has access to it. [registry.HasGlobalAccess](https://github.com/okteto/okteto/blob/master/pkg/registry/registry.go#L141) is your friend here 😄

pkg/cache/cache_from.go
```diff
@@ -0,0 +1,51 @@
...
+	globalCacheImage := fmt.Sprintf("%s/%s:cache", "okteto.global", image)
+	devCacheImage := fmt.Sprintf("%s/%s:cache", "okteto.dev", image)
```
jLopezbarb, line 32:
```suggestion
	devCacheImage := fmt.Sprintf("%s/%s:cache", constans.DevRegistry, image)
```

pkg/cache/cache_from.go
andreafalzetti: Done 👍

pkg/cache/cache_from_test.go
```diff
@@ -0,0 +1,128 @@
...
+
+func (m *mockImageCtrl) GetRegistryAndRepo(_ string) (string, string) {
+	return "registry", "test-account/test-image:x.y.z"
+}
+
+func (m *mockImageCtrl) GetRepoNameAndTag(_ string) (string, string) {
+	return "test-account/test-image", "x.y.z"
+}
+
+func Test_AddDefaultPullCache(t *testing.T) {
```
jLopezbarb, line 56: We should add serialisation tests (Unmarshal and Marshal)

jLopezbarb:
Other than that LGTM

andreafalzetti:
@AdrianPedriza @jLopezbarb WDYT we should do with all these anti-patterns? https://deepsource.io/gh/okteto/okteto/run/bf3d2f91-7e75-45d1-8357-0995add2ea44/go/GO-W1029?listindex=0

pkg/cache/cache_from.go
```diff
@@ -0,0 +1,101 @@
...
+func (cf CacheFrom) MarshalYAML() (interface{}, error) {
+	if len(cf) == 1 {
+		return cf[0], nil
+	}
+
+	return cf, nil
+}
+
+// AddDefaultPullCache appends the default cache layers for a given image
+func (cf *CacheFrom) AddDefaultPullCache(reg oktetoRegistryInterface, image string) {
```
ifbyol, line 65: It feels odd that an exported function depends on an unexported interface. The compiler will catch any possible error anyways but it is hard for a client of this package to know which functions has to implement to satisfy the interface if it is unexported unless they access to the source code. 

Imagine if someone is using the CLI as dependency in their project and they want to use the CacheFrom. They would have to go to the source code to see which functions are defined in that interface

pkg/cache/cache_from_test.go
```diff
@@ -0,0 +1,221 @@
...
+		expected interface{}
+	}{
+		{
+			name:     "empty",
+			data:     []byte("[]"),
+			expected: CacheFrom{},
+		},
+		{
+			name:     "one item",
+			data:     []byte(`["test-registry/test-image:cache"]`),
```
AdrianPedriza, line 146:
```suggestion
			data:     []byte(`"test-registry/test-image:cache"`),
```
If we send a list we are using the same test case as the last one

pkg/cache/cache_from_test.go
```diff
@@ -0,0 +1,221 @@
...
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			result := tt.cf.hasCacheFromImage(tt.input)
+			assert.Equal(t, tt.expected, result)
+		})
+	}
+}
+
+func Test_UnmarshalYAML(t *testing.T) {
```
AdrianPedriza, line 133: I think we could add a test case where the input cannot be converted to either a string or a list and we get an error

pkg/cache/cache_from_test.go
```diff
@@ -0,0 +1,221 @@
...
+		t.Run(tt.name, func(t *testing.T) {
+			var cf CacheFrom
+			err := yaml.Unmarshal([]byte(tt.data), &cf)
+			assert.NoError(t, err)
+			assert.Equal(t, tt.expected, cf)
+		})
+	}
+}
+
+func Test_UnmarshalYAML_WithError(t *testing.T) {
```
AdrianPedriza, line 168: I believe we are not reaching the Unmarshal error of a `CacheFrom` object.

pkg/cache/cache_from.go
```diff
@@ -0,0 +1,101 @@
...
+// addCacheFromImage appends a cache image to the list if it's not already there
+func (cf *CacheFrom) addCacheFromImage(imageName string) {
+	if !cf.hasCacheFromImage(imageName) {
+		*cf = append(*cf, imageName)
+	}
+}
+
+// hasCacheFromImage checks if a cache image is already in the list
+func (cf *CacheFrom) hasCacheFromImage(imageName string) bool {
+	for _, cacheFrom := range *cf {
```
AdrianPedriza, line 95: would it be a good idea to make the cachefrom object a map? maybe it would be easier to search for elements in it?

AdrianPedriza:
> @AdrianPedriza @jLopezbarb WDYT we should do with all these anti-patterns? https://deepsource.io/gh/okteto/okteto/run/bf3d2f91-7e75-45d1-8357-0995add2ea44/go/GO-W1029?listindex=0

I would try to be as consistent as possible, using pointers or not if the other methods of that object do

andreafalzetti:
> > @AdrianPedriza @jLopezbarb WDYT we should do with all these anti-patterns? https://deepsource.io/gh/okteto/okteto/run/bf3d2f91-7e75-45d1-8357-0995add2ea44/go/GO-W1029?listindex=0
> 
> I would try to be as consistent as possible, using pointers or not if the other methods of that object do

That touches pretty much the whole serialized, I am not sure it's a good idea to fix that in this PR 🤔 I will have a look

pkg/cache/cache_from.go
andreafalzetti: @AdrianPedriza it would simplify accessing the values, not sure about the other consequences yet (for example, in YAML this has to stay an array so would need to make sure mapping array -> map is not causing any issue).

AdrianPedriza:
> > > @AdrianPedriza @jLopezbarb WDYT we should do with all these anti-patterns? https://deepsource.io/gh/okteto/okteto/run/bf3d2f91-7e75-45d1-8357-0995add2ea44/go/GO-W1029?listindex=0
> > 
> > 
> > I would try to be as consistent as possible, using pointers or not if the other methods of that object do
> 
> That touches pretty much the whole serialized, I am not sure it's a good idea to fix that in this PR 🤔 I will have a look

deepsource is sometimes a bit noisy. I would just stick to the deepsource rules for your entered changes. There is no mandatory check for integration, so we don't need to comply with all the suggestions it gives us. IMO the most important thing is to have consistency in the method receivers of the same struct (for example, that all the receivers of the struct CacheFrom are pointers.

andreafalzetti:
> > > > @AdrianPedriza @jLopezbarb WDYT we should do with all these anti-patterns? https://deepsource.io/gh/okteto/okteto/run/bf3d2f91-7e75-45d1-8357-0995add2ea44/go/GO-W1029?listindex=0
> > > 
> > > 
> > > I would try to be as consistent as possible, using pointers or not if the other methods of that object do
> > 
> > 
> > That touches pretty much the whole serialized, I am not sure it's a good idea to fix that in this PR 🤔 I will have a look
> 
> I was just thinking aloud about that possibility. If it is a commitment that this PR should not make then let's forget it 😃

I've tried updating other receivers too but it very quickly becomes a critical surgery! I think it makes sense to take over time, as we touch the different methods, I agree with the approach you suggested on using DeepSource!

pkg/cache/cache_from_test.go
andreafalzetti: you were right! I've updated the test 🙏

pkg/cache/cache_from_test.go
```diff
@@ -0,0 +1,217 @@
...
+		{
+			name:     "two images",
+			cf:       CacheFrom{"test-registry/test-image-1:cache", "test-registry/test-image-2:cache"},
+			expected: "- test-registry/test-image-1:cache\n- test-registry/test-image-2:cache\n",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			result, err := yaml.Marshal(tt.cf)
```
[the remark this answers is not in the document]
AdrianPedriza: weird. From codcoverage tool in VScode seems is tested. This is the config I have in my `settings.json` :

```json
"go.coverOnSave": true,
"go.coverageDecorator": {
    "type": "highlight",
    "coveredHighlightColor": "rgba(64,128,128,0.5)",
    "uncoveredHighlightColor": "rgba(128,64,64,0.25)",
    "coveredGutterStyle": "blockgreen",
    "uncoveredGutterStyle": "blockred"
},
"go.coverOnSingleTest": true,
```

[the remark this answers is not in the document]
AdrianPedriza: and indeed the test does cover it, I have checked. I would base it on VScode coverage, it seems that `codcoverage` is not up to date with the latest changes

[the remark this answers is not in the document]
AdrianPedriza: OK! now `UnmarshalYAML` method is fully covered! but `MarshalYAML` not :/. 
```suggestion
			result, err := tt.cf.MarshalYAML()
```
maybe if we call it directly in test we reach it

AdrianPedriza, line 212:
```suggestion
			result, err := yaml.Marshal(&tt.cf)
```
I think it was the problem to not reach `Marshal()` func. The signature that we use didn't fit with the real one.

pkg/cache/cache_from.go
```diff
@@ -0,0 +1,101 @@
...
+import (
+	"fmt"
+
+	"github.com/okteto/okteto/pkg/constants"
+	oktetoLog "github.com/okteto/okteto/pkg/log"
+)
+
+// CacheFrom is a list of images to import cache from.
+type CacheFrom []string
+
```
AdrianPedriza, line 25:
```suggestion
// 
```

pkg/cache/cache_from.go
```diff
@@ -0,0 +1,101 @@
...
+type ImageCtrlInterface interface {
```
AdrianPedriza, line 26: I think we should add a description here since it is exported to the registry pkg

AdrianPedriza (approved):

AdrianPedriza:
I have only small comments or suggestions. Overall I think it looks good! 👍

pkg/registry/registry.go
```diff
@@ -145,3 +146,8 @@ func (or OktetoRegistry) HasGlobalPushAccess() (bool, error) {
 	image := or.imageCtrl.ExpandOktetoGlobalRegistry(globalTestImage)
 	return or.client.HasPushAccess(image)
 }
+
+// GetImageCtrl returns the image controller
+func (or OktetoRegistry) GetImageCtrl() cache.ImageCtrlInterface {
```
jLopezbarb, line 151: I find odd to return an interface from another package. It should return the ImageCtrl from this package and the clients implement their own interface

andreafalzetti: we resolve this together and merged the other PR into this branch

pkg/cache/cache_from.go
andreafalzetti: It's no longer exported now, but it was a valid comment 🙏

pkg/cache/cache_from_test.go
andreafalzetti: good catch @AdrianPedriza, I've updated the code! 🙏

PR #3489, opened 2023-03-29
inject OKTETO_ACTION_NAME in remote

AdrianPedriza:
# Proposed changes
- Inject `OKTETO_ACTION_NAME` in remote destroy

Fixes #[6123](https://github.com/okteto/app/issues/6123)

When we perform a remote destroy using Okteto UI we check if there is an operation already running. To do so, we check `OKTETO_ACTION_NAME` value. If we don't inject this env in remote, we will not update the config map where information about the destroy action is stored.

### DIFF
--- a/cmd/destroy/remote.go
+++ b/cmd/destroy/remote.go
@@ -50,6 +50,7 @@ ENV {{ .NamespaceEnvVar }} {{ .NamespaceValue }}
 ENV {{ .ContextEnvVar }} {{ .ContextValue }}
 ENV {{ .TokenEnvVar }} {{ .TokenValue }}
 ENV {{ .RemoteDeployEnvVar }} true
+ENV {{ .ActionNameEnvVar }} {{ .ActionNameValue }}
 
 COPY . /okteto/src
 WORKDIR /okteto/src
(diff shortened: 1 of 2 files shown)


cmd/destroy/remote.go
```diff
@@ -50,6 +50,7 @@ ENV {{ .NamespaceEnvVar }} {{ .NamespaceValue }}
 ENV {{ .ContextEnvVar }} {{ .ContextValue }}
 ENV {{ .TokenEnvVar }} {{ .TokenValue }}
 ENV {{ .RemoteDeployEnvVar }} true
+ENV {{ .ActionNameEnvVar }} {{ .ActionNameValue }}
```
jLopezbarb, line 53: Can we add a test covering that we are adding or not this value?

AdrianPedriza: done a840a698ababb23144caa9d54e0f890d7e74a9b3

ifbyol (approved):

jLopezbarb (approved):

AgustinRamiroDiaz (approved):

teresaromero (approved):

PR #3493, opened 2023-03-29
fix: add remote build image building

jLopezbarb:
# Proposed changes

Fixes https://github.com/okteto/app/issues/6122

- Checks if the image that is in the `destroy.image` is already built and build it if it's not built

### DIFF
--- a/cmd/destroy/build_control.go
+++ b/cmd/destroy/build_control.go
@@ -46,7 +46,7 @@ func (bc buildCtrl) buildImageIfNecessary(ctx context.Context, manifest *model.M
 	oktetoLog.Debug("checking if destroy.image is already built")
 	imageToBuild := manifest.Destroy.Image
 
-	reg := regexp.MustCompile(`OKTETO_BUILD_(\w+)_IMAGE`)
+	reg := regexp.MustCompile(`OKTETO_BUILD_(\w+)_`)
 	matches := reg.FindStringSubmatch(imageToBuild)
 	if len(matches) == 0 {
 		oktetoLog.Debugf("image '%s' is not an okteto build variable", imageToBuild)
--- a/cmd/destroy/build_control.go
+++ b/cmd/destroy/build_control.go
@@ -1,0 +1,87 @@
+// Copyright 2023 The Okteto Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package destroy
+
+import (
+	"context"
+	"fmt"
+	"regexp"
+	"strings"
+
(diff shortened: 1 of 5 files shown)


cmd/destroy/build_control.go
```diff
@@ -0,0 +1,87 @@
...
+		}
+		svcsToBuild, err := bc.builder.GetServicesToBuild(ctx, manifest, []string{svc})
+		if err != nil {
+			return fmt.Errorf("error getting services to build: %w", err)
+		}
+		if len(svcsToBuild) == 0 {
+			oktetoLog.Infof("image is already built: %s", imageToBuild)
+			return nil
+		}
+		buildOptions := &types.BuildOptions{
```
teresaromero, line 76: are this the only options we want to pass though to build? what about image, etc?

cmd/destroy/destroy.go
```diff
@@ -244,6 +245,16 @@ func (dc *destroyCommand) getDestroyer(opts *Options) (destroyInterface, error)
 
 		destroyImage := ""
 		if manifest.Destroy != nil {
+			if opts.Name == "" {
+				if manifest.Name == "" {
+					manifest.Name = dc.buildCtrl.name
```
teresaromero, line 250: where does `dc.builderCtrl` get `name` ?

jLopezbarb: In the initializer. in the `destroy.go`. The name is the inferred name (from the cmap or the folder name)

cmd/destroy/build_control.go
jLopezbarb: We get the image information from the manifest.Build section. I'm adding `CommandArgs: svcsToBuild,` with the service that needs to be built to have the logs with the stage

jLopezbarb:
> testing this change I notice that if we need to create the image we don't display any message related to that process. Maybe the user expects some log related to his destroy commands but we only see a 'stuck' destroy command (because is building the image). Should we add some info to the user saying that destroy image is being built?
> 
> Steps to repro:
> 
> * clone https://github.com/okteto/movies/tree/adrian/test-3493
> * exec `okteto deploy`
> * exec `okteto destroy`
> * check no info displayed related to building destroy image (it may take time)
> 
> Even a message of already built images is displayed when this is not the case.

Hi @AdrianPedriza  couldn't reproduce your issue, but checking the build there was an optimisation to be done at the image check. We were checking all the images even if we only needed one of them. Here is [a loom video](https://www.loom.com/share/0ffd7417d42a4976aded116fb8200b52) to show you that is working as expected

AdrianPedriza:
> > testing this change I notice that if we need to create the image we don't display any message related to that process. Maybe the user expects some log related to his destroy commands but we only see a 'stuck' destroy command (because is building the image). Should we add some info to the user saying that destroy image is being built?
> > Steps to repro:
> > 
> > * clone https://github.com/okteto/movies/tree/adrian/test-3493
> > * exec `okteto deploy`
> > * exec `okteto destroy`
> > * check no info displayed related to building destroy image (it may take time)
> > 
> > Even a message of already built images is displayed when this is not the case.
> 
> Hi @AdrianPedriza couldn't reproduce your issue, but checking the build there was an optimisation to be done at the image check. We were checking all the images even if we only needed one of them. Here is [a loom video](https://www.loom.com/share/0ffd7417d42a4976aded116fb8200b52) to show you that is working as expected

nevermind, I think I tested it wrong. I have tested it again and it works as it should! :)

AdrianPedriza (approved):

andreafalzetti (approved):
LGTM

PR #3494, opened 2023-03-29
chore(destroy): better error message when image not accessible

andreafalzetti:
# Proposed changes

Fixes [#6121](https://github.com/okteto/app/issues/6121)

### How to test

1. Create a manifest like this:
```yaml
destroy:
  image: image-that-doesnt-exist:nope
  commands:
    - name: hello
      command: echo "hello world!"
```
1. Run `okteto destroy` 
1. Notice the error message being: "Error during development environment deployment"
1. Now build this branch and run again
1. Notice the error being: "Error during destroying the development environment: error building image: failed to pull image 'image-that-doesnt-exist:nope'. The repository is not accessible or it does not exist."

| Before | After |
|--------|-------|
|        |        |

### DIFF
--- a/cmd/destroy/destroy.go
+++ b/cmd/destroy/destroy.go
@@ -252,7 +252,7 @@ func (dc *destroyCommand) getDestroyer(opts *Options) (destroyInterface, error) 
 		runInRemote := !isRemote && (destroyImage != "" || opts.RunInRemote)
 
 		if runInRemote {
-			deployer = newRemoteDestroyer(destroyImage)
+			deployer = newRemoteDestroyer(manifest)
 			oktetoLog.Info("Destroying remotely...")
 		} else {
 			destroyerAll, err := newLocalDestroyerAll(dc.k8sClientProvider, dc.executor, dc.nsDestroyer, dc.oktetoClient)
--- a/cmd/destroy/remote.go
+++ b/cmd/destroy/remote.go
@@ -81,16 +81,18 @@ type remoteDestroyCommand struct {
 	fs                   afero.Fs
 	workingDirectoryCtrl filesystem.WorkingDirectoryInterface
 	temporalCtrl         filesystem.TemporalDirectoryInterface
+	manifest             *model.Manifest
 }
 
-func newRemoteDestroyer(destroyImage string) *remoteDestroyCommand {
+func newRemoteDestroyer(manifest *model.Manifest) *remoteDestroyCommand {
 	fs := afero.NewOsFs()
 	return &remoteDestroyCommand{
 		builder:              remoteBuild.NewBuilderFromScratch(),
-		destroyImage:         destroyImage,
+		destroyImage:         manifest.Destroy.Image,
 		fs:                   fs,
 		workingDirectoryCtrl: filesystem.NewOsWorkingDirectoryCtrl(),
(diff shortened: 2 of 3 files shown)


cmd/destroy/destroy.go
```diff
@@ -65,6 +65,7 @@ type Options struct {
 	ManifestPath        string
 	Name                string
 	Variables           []string
+	Manifest            *model.Manifest
```
andreafalzetti, line 68: we need to propagate the Manifest to be able to get the image name in the error message

jLopezbarb: Let's try to not use the options to store the manifest. Maybe we can store it in the `remoteDestroyCommand` struct adding it [here](https://github.com/okteto/okteto/pull/3494/files#diff-a1fde665f65ea0f657c3cc6aff6529b33ec5b8c7cd15596b7418a3ea5d8669d3R259). This will reduce a lot the complexity of the changes.

We can change `newRemoteDestroyer` signature to receive a manifest instead of the destroyImage name and that will should be simpler. WDYT?

andreafalzetti: I have used the same approach for the `build` and did the change fairly quickly so I didn't stop too much thinking about other ways to store, so thanks, I will update it now

andreafalzetti: @jLopezbarb I've started making the change, but realized that the error message is composed [here](https://github.com/okteto/okteto/blob/master/pkg/cmd/build/build.go#L177-L184) and if `buildOptions.Manifest` is nil, it won't work - that's because `build` already had the Manifest in the options. 

I will think about your approach, but I foresee other changes needed to make it work

andreafalzetti: I looked with fresh eyes at the code after lunch and your suggestion makes 100% sense. I've updated the code now. Thanks @jLopezbarb!

jLopezbarb (approved):

cmd/destroy/remote.go
```diff
@@ -132,14 +134,15 @@ func (rd *remoteDestroyCommand) destroy(ctx context.Context, opts *Options) erro
...
+	buildOptions.Manifest = rd.manifest
 
 	// we need to call Build() method using a remote builder. This Builder will have
 	// the same behavior as the V1 builder but with a different output taking into
 	// account that we must not confuse the user with build messages since this logic is
 	// executed in the deploy command.
 	if err := rd.builder.Build(ctx, buildOptions); err != nil {
 		return oktetoErrors.UserError{
-			E: fmt.Errorf("error during development environment deployment"),
+			E: fmt.Errorf("error during destroying the development environment: %w", err),
```
teresaromero, line 145:
```suggestion
			E: fmt.Errorf("error during destroy of the development environment: %w", err),
```

teresaromero (approved):
minor comment regarding the copy of the error :/ sounds wierd

AdrianPedriza (approved):

ifbyol:
Do you have an screenshot or example of how the error message looks like in this case?

andreafalzetti:
> Do you have an screenshot or example of how the error message looks like in this case?

@ifbyol  Here you go!

| Before | After |
|--------|-------|
|        |        |

PR #3496, opened 2023-03-29
remove external resource validation

AdrianPedriza:
# Proposed changes
- remove external resource validation. 

Fixes #https://github.com/okteto/app/issues/6124

validation removed was creating an external CRD (inexpertly) before the proxy (the one adding `deploy-by` label) running so this caused the external to be seen in the UI outside the environment that displayed it.


jLopezbarb (approved):

pkg/externalresource/control.go
```diff
@@ -79,22 +79,6 @@ func (c *K8sControl) List(ctx context.Context, ns string, labelSelector string)
...
 
-func (c *K8sControl) Validate(ctx context.Context, name, ns string, externalInfo *ExternalResource) error {
-	k8sclient, err := c.ClientProvider(c.Cfg)
-	if err != nil {
-		return fmt.Errorf("error creating external CRD client: %s", err.Error())
-	}
-
-	externalResourceCRD := translate(name, externalInfo)
-
-	_, err = k8sclient.ExternalResources(ns).Create(ctx, externalResourceCRD, metav1.CreateOptions{DryRun: []string{metav1.DryRunAll}})
```
ifbyol, line 90: Why this fixes the issue if the create is done with a `DryRun` which means that the changes won't be persisted in Kubernetes?

AdrianPedriza: even having that option, the CRD was being created. I will create a follow up issue to try to introduce a validation again properly

AdrianPedriza:
follow up issue tu add again validation https://github.com/okteto/okteto/issues/3510

pkg/externalresource/control.go
ifbyol: Cool, thanks!!

PR #3504, opened 2023-03-30
fix destroy remote when adding envs with blank values

AdrianPedriza:
# Proposed changes
- do not add ENV to img if value is blank

when an image is built for running destroy operations, some tests added an ENV with a blank value so this caused the destroy operation to fail because the image that would execute the command could not be created.

### DIFF
--- a/cmd/deploy/remote.go
+++ b/cmd/deploy/remote.go
@@ -63,6 +63,9 @@ ENV {{ .NamespaceEnvVar }} {{ .NamespaceValue }}
 ENV {{ .ContextEnvVar }} {{ .ContextValue }}
 ENV {{ .TokenEnvVar }} {{ .TokenValue }}
 ENV {{ .RemoteDeployEnvVar }} true
+{{ if ne .ActionNameValue "" }}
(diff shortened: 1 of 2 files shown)


jLopezbarb:
Should this change be included in the remote deploy too? I guess the issue is affecting also the deployments. What happen if two different pipelines are deployed at the same time?

AdrianPedriza:
> Should this change be included in the remote deploy too? I guess the issue is affecting also the deployments. What happen if two different pipelines are deployed at the same time?

yes, config map updates are also performed in the destroy so it may also be necessary to make use of the `OKTETO_ACTION_NAME` value in the remote deploy. Done 702ae99f7cfec01b8b52c99f13243a1b02517b43

ifbyol (approved):
We should make sure that the CLI when reads that env variable doesn't treat differently the absence of that env than an empty value

AdrianPedriza:
> We should make sure that the CLI when reads that env variable doesn't treat differently the absence of that env than an empty value

only scenario that does not have this env set was the failing test. Others include this env therefore include its value to the remote

jLopezbarb (approved):

PR #3507, opened 2023-03-30
fix: extra step internal server error on remote

jLopezbarb:
# Proposed changes

Fixes https://github.com/okteto/app/issues/6159

- Propagate command error from buildkit logs until the remote command. Then set CLI  stage to the same it was happening in buildkit
- This also fixes an issue with the spinner


andreafalzetti:
Code changes LGTM.

I've tested it and this is was I see:
Screenshot 2023-03-30 at 14 24 22

Is it expected to have an Internal Server Error? I find that very confusing. Why not only keep the error on the actual step that failed?

andreafalzetti (approved):
Approving as my comment is a genuine question whether we can further improve it, but it's already an improvement that we can ship

AdrianPedriza:
> Code changes LGTM.
> 
> I've tested it and this is was I see: Screenshot 2023-03-30 at 14 24 22
> 
> Is it expected to have an Internal Server Error? I find that very confusing. Why not only keep the error on the actual step that failed?

I think we should not have that step, it is not helpful, it contains the same cause of the error as the failed command step. Should we keep it? @jLopezbarb

PR #3513, opened 2023-03-31
infer dev name before create remote image

AdrianPedriza:
# Proposed changes

Fixes #https://github.com/okteto/app/issues/6165

the dev name inference on the remote does not work like on local because it has no information about the `.git` (we avoid adding that folder to the remote) so when inferring the name on remote it does not match the real name.

### DIFF
--- a/cmd/destroy/destroy.go
+++ b/cmd/destroy/destroy.go
@@ -136,14 +136,13 @@ func Destroy(ctx context.Context) *cobra.Command {
 				return fmt.Errorf("failed to get the current working directory: %w", err)
 			}
 
-			name := options.Name
 			if options.Name == "" {
 				c, _, err := okteto.NewK8sClientProvider().Provide(okteto.Context().Cfg)
 				if err != nil {
 					return err
 				}
 				inferer := devenvironment.NewNameInferer(c)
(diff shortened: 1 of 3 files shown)


cmd/destroy/destroy.go
```diff
@@ -191,6 +191,15 @@ func Destroy(ctx context.Context) *cobra.Command {
 			os.Setenv("KUBECONFIG", kubeconfigPath)
 			defer os.Remove(kubeconfigPath)
 
+			if options.Name == "" {
```
teresaromero, line 194: why are we moving this here? was just being done at local destroyer?

jLopezbarb, line 194: We should also have this logic in the deploy.
We can use the same repo switching the deploy with the destroy sections and running `okteto deploy`

ifbyol: Why don't we just use the logic in line 140 instead of adding it here? I think it would be enough to update options.Name there

AdrianPedriza: I think deploy command will not face this issue because there has been nothing prior to the deploy that has created resources with a different name (as is the case with the destroy)

ifbyol (approved):

cmd/destroy/destroy.go
ifbyol: Done!

PR #3515, opened 2023-03-31
Restore "Okteto kubeconfig will use kubetoken client auth if available on okteto server""

AgustinRamiroDiaz:
# Changes

- Reverts okteto/okteto#3490
- Fixes e2e tests with https://github.com/okteto/okteto/pull/3553
- Requires https://github.com/okteto/okteto/pull/3602

# Note

The breaking change is that `okteto` needs to be in `PATH` for `okteto kubeconfig` to work properly

# Release plan
https://github.com/okteto/okteto/pull/3602 needs to be merged in a previous version than this PR, so that okteto developers are not forced to add their development binary to path


pkg/types/credential.go
```diff
@@ -17,6 +17,6 @@ package types
 type Credential struct {
 	Server      string `json:"server" yaml:"server"`
 	Certificate string `json:"certificate" yaml:"certificate"`
-	Token       string `json:"token" yaml:"token"`
+	Token       string `json:"token" yaml:"token"` // TODO: Deprecate once we move to kubetoken auth
```
jLopezbarb, line 20: We should maintain retrocompatibility as much as we can but I agree that this should be deprecated.

jLopezbarb (approved):

cmd/context/create.go
```diff
@@ -52,6 +54,21 @@ func NewContextCommand() *ContextCommand {
 		LoginController:      login.NewLoginController(),
 		OktetoClientProvider: okteto.NewOktetoClientProvider(),
 		OktetoContextWriter:  okteto.NewContextConfigWriter(),
+		IsOktetoKubeTokenPresent: func(oktetoURL string) (bool, error) {
```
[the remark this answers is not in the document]
AgustinRamiroDiaz: In a semantic way, I'd say yeah it makes a lot of sense. But seeing around the code dependencies, I feel that we'd be overloading the `OktetoInterface`. 

I think that we could refactor the code so that:
- `IsOktetoKubeTokenPresent` is in the `OktetoClient`
- [Provide](https://github.com/okteto/okteto/blob/1341b4675e0e546a3ce5d8ca86df54319933e83f/pkg/okteto/client.go#L62) returns an instance instead of an interface
- Reduce dependencies from the OktetoClient interface in other packages (optional, related)

Maybe we should do it in a follow up PR

WDYT?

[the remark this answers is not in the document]
jLopezbarb: I like that approach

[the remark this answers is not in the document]
AgustinRamiroDiaz: I've created https://github.com/okteto/okteto/issues/3581to tackle this

cmd/context/create.go
```diff
@@ -52,6 +54,21 @@ func NewContextCommand() *ContextCommand {
...
+			if err != nil {
+				return false, fmt.Errorf("failed to parse kubetoken url: %w", err)
+			}
+
+			resp, err := http.Get(kubeTokenURL)
+			if err != nil {
+				return false, fmt.Errorf("failed to get kubetoken url: %w", err)
+			}
+
+			return resp.StatusCode == http.StatusUnauthorized, nil
```
AdrianPedriza, line 70:
```suggestion
			return resp.StatusCode != http.StatusUnauthorized, nil
```
if endpoint is not available backend response will be `401` but then we return `true` I think. We should return `true` when response is not `401` following func name

AgustinRamiroDiaz: @AdrianPedriza I got your confusion, but it's the other way around

> if endpoint is not available backend response will be 401

In fact, it's the opposite. if endpoint IS AVAILABLE backend response will be 401, since it can handle the request but we didn't add the auth mechanism. You can check this out by going to https://cloud.okteto.com/auth/kubetoken/default

If the endpoint is NOT AVAILABLE we'd get a 404

AdrianPedriza: cool then! 🚀

PR #3521, opened 2023-03-31
fix: duplicated logs on same stage

jLopezbarb:
# Proposed changes

Fixes Duplicated logs on the same stage.

- We were splitting the signal that buildkit send us to save it and put it in the output. This change only uses one channel that is sending the information back to the CLI and is the CLI the one updating the cmap

## Screenshoot:
Captura de Pantalla 2023-04-04 a las 11 13 17
Captura de Pantalla 2023-03-31 a las 18 19 27


AdrianPedriza:
can you post a "before and after" about how looks like your change?

jLopezbarb:
> can you post a "before and after" about how looks like your change?

Updated @AdrianPedriza

AdrianPedriza (approved):

PR #3534, opened 2023-04-05
fix .env removal locally

### LINKED ISSUE #3532: Okteto CLI removes .env file
I have an .env file with many environment variables for our project. When I use okteto CLI to deploy, build or destroy environments using a custom docker-file, okteto cli randomly deletes that file (which is very annoying). It's started happening after upgrading to 2.14.1 version.

**To Reproduce**

1. Upgrade to 2.14.1
2. okteto deploy -f docker-kompose.yaml
3. The .env file on the project root is ramdonly deleted.

After downgrading to 2.12.0 the problem has disappeared.

jLopezbarb:
# Proposed changes

Fixes #3532

- local was not cleaning the correct path and could lead to removing the `.env` file
- Added a test to make sure this is not happening again

### DIFF
--- a/cmd/deploy/local_test.go
+++ b/cmd/deploy/local_test.go
@@ -22,32 +22,21 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func TestRunDeploySection(t *testing.T) {
+func TestDeployNotRemovingEnvFile(t *testing.T) {
 	fs := afero.NewMemMapFs()
 
 	_, err := fs.Create(".env")
 	require.NoError(t, err)
-	tt := []struct {
-		name string
-		opts *Options
-	}{
-		{
-			name: "check that deploy doesn't remove .env file",
-			opts: &Options{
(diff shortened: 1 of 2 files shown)


teresaromero (approved):

cmd/deploy/local_test.go
```diff
@@ -0,0 +1,53 @@
...
+import (
+	"context"
+	"testing"
+
+	"github.com/okteto/okteto/pkg/model"
+	"github.com/spf13/afero"
+	"github.com/stretchr/testify/require"
+)
+
+func TestRunDeploySection(t *testing.T) {
```
AdrianPedriza, line 25:
```suggestion
func TestDeployNotRemovingEnvFile(t *testing.T) {
```
I think this name is more descriptive for the test. WDYT? 😄

AdrianPedriza (approved):

ifbyol (approved):

PR #3535, opened 2023-04-05
remove sensitive data in git url displayed

AdrianPedriza:
# Proposed changes
- remove sensitive data when we add git URL as annotation

Fixes #https://github.com/okteto/app/issues/6193

### DIFF
--- a/pkg/cmd/pipeline/translate.go
+++ b/pkg/cmd/pipeline/translate.go
@@ -32,6 +32,7 @@ import (
 	"github.com/okteto/okteto/pkg/k8s/configmaps"
 	oktetoLog "github.com/okteto/okteto/pkg/log"
 	"github.com/okteto/okteto/pkg/model"
+	giturls "github.com/whilp/git-urls"
 	apiv1 "k8s.io/api/core/v1"
 	k8sErrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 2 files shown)


pkg/cmd/pipeline/translate.go
```diff
@@ -260,11 +261,27 @@ func AddDevAnnotations(ctx context.Context, manifest *model.Manifest, c kubernet
...
+			app.ObjectMeta().Annotations[model.OktetoRepositoryAnnotation] = removeSensitiveDataFromGitURL(repo)
 		}
 		app.ObjectMeta().Annotations[model.OktetoDevNameAnnotation] = devName
 		if err := app.PatchAnnotations(ctx, c); err != nil {
 			oktetoLog.Infof("could not add %s dev annotations due to: %s", devName, err.Error())
 		}
 	}
 }
+
+func removeSensitiveDataFromGitURL(gitURL string) string {
```
teresaromero, line 273:
```suggestion
func redactedGitURL(gitURL string) string {
```

pkg/cmd/pipeline/translate.go
```diff
@@ -260,11 +261,27 @@ func AddDevAnnotations(ctx context.Context, manifest *model.Manifest, c kubernet
...
+	if gitURL == "" {
+		return gitURL
+	}
+
+	parsedRepo, err := giturls.Parse(gitURL)
+	if err != nil {
+		return ""
```
AdrianPedriza, line 280: I couldn't see any line that can be based on this value so I guess is safe to set it as `""` for that cases.

teresaromero: i found this https://github.com/okteto/app/master/backend/pkg/model/redact.go#L13 at the backend, where logic is more or less the sale, in case of err, the gitURL is returned, could it be that the redacted function here is not working?

AdrianPedriza: Seems this annotations in set from backend installer so this fix need to be done from CLI side. I tested my solution with my change in a custom installer with proposed changes and works as expected 👍

pkg/cmd/pipeline/translate_test.go
```diff
@@ -110,8 +110,49 @@ func Test_AddDevAnnotations(t *testing.T) {
 	assert.Equal(t,
 		d.Annotations,
 		map[string]string{
-			model.OktetoRepositoryAnnotation: "git-repo",
+			model.OktetoRepositoryAnnotation: "file://git-repo",
```
AdrianPedriza, line 113: Following my assumption that changing the value of this annotation does not affect other existing logic, I have modified this test instead of adding more logic to the introduced function just to pass the test

teresaromero (approved):

pkg/cmd/pipeline/translate.go
AgustinRamiroDiaz: @AdrianPedriza does that mean that this PR is not needed?

AdrianPedriza: no no, means that change needs to be done in CLI not it backend

ifbyol (approved):

jLopezbarb (approved):

PR #3539, opened 2023-04-05
build empty `.dockerignore` in tmp dir for remote build if no `.oktetodeployignore` provided

AdrianPedriza:
# Proposed changes

- always create `.dockerignore` in tmp dir if no `.oktetodeployignore` provided. If we don't create this file in tmp dir, buildkit lib will use as possible `.dockerignore` the one in the context set it up for the build (folder where user exec okteto deploy)

Fixes #[6214](https://github.com/okteto/app/issues/6214)

### DIFF
--- a/cmd/deploy/remote.go
+++ b/cmd/deploy/remote.go
@@ -223,21 +223,22 @@ func (rd *remoteDeployCommand) createDockerfile(tmpDir string, opts *Options) (s
 }
 
 func (rd *remoteDeployCommand) createDockerignoreIfNeeded(cwd, tmpDir string) error {
+	dockerignoreContent := []byte(``)
 	dockerignoreFilePath := filepath.Join(cwd, oktetoDockerignoreName)
 	if _, err := rd.fs.Stat(dockerignoreFilePath); err != nil {
 		if !errors.Is(err, os.ErrNotExist) {
 			return err
 		}
-	} else {
-		dockerignoreContent, err := afero.ReadFile(rd.fs, dockerignoreFilePath)
-		if err != nil {
-			return err
-		}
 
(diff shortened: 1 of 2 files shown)


cmd/deploy/remote.go
```diff
@@ -223,22 +223,23 @@ func (rd *remoteDeployCommand) createDockerfile(tmpDir string, opts *Options) (s
 }
 
 func (rd *remoteDeployCommand) createDockerignoreIfNeeded(cwd, tmpDir string) error {
+	dockerignoreContent := []byte(``)
```
jLopezbarb, line 226: Can we add a comment explaining why do we create it empty

cmd/deploy/remote.go
```diff
@@ -223,22 +223,23 @@ func (rd *remoteDeployCommand) createDockerfile(tmpDir string, opts *Options) (s
 }
 
 func (rd *remoteDeployCommand) createDockerignoreIfNeeded(cwd, tmpDir string) error {
```
jLopezbarb, line 225: Can we add a test for this function? We should check that we create the right file with the content without leaking the content from the `.dockerignore`

cmd/deploy/remote.go
AdrianPedriza: done b91e64af29e071cbc55246556b66f024435f2ea3

cmd/deploy/remote.go
AdrianPedriza: done b91e64af29e071cbc55246556b66f024435f2ea3

teresaromero (approved):

jLopezbarb (approved):

PR #3540, opened 2023-04-05
export cache always

### LINKED ISSUE #3424: Building an image exports that image to the cache
## Parent epic/feature

## Description

In order to optimize the builds so that if a previous image has already been created, we should cache the layers so that when the same build is called it always has the cache of the previous builded images. For this, when we build an image we must store the layers in the cache. This way the user could import from the image for other builds.

This issue seeks to always export the cache of the image built as `inline` type and with mode set to `min` ([as it's the only value that is available with inline](https://github.com/moby/buildkit#export-cache)).

We should add always the image that the user is using to the export cache option programatically(note that the user has already set an image in the `export_cache`):
```yaml
build:
  api:
    context: .
    dockerfile: Dockerfile
    export_cache: okteto.dev/my-app:cache
    image: okteto.dev/api:okteto
```

should become programatically:
```yaml
build:
  api:
    context: .
    dockerfile: Dockerfile
    export_cache:
    - okteto.dev/my-app:cache
    - okteto.dev/api:okteto
    image: okteto.dev/api:okteto
```

In order to do that we need to allow using a list in the `export_cache` field and modify the builder to export all the caches in the manifest

### Acceptance criteria

- [ ] `export_cache` field admits a list instead a single string
- [ ]  adds to `export_cache` field the needed values: always export cache to specified registry plus the one user has access to. 

##### Scenario 1: User can provide a list

1. Create a new namespace
2. Create the following manifest:
```yaml
build:
  frontend:
    context: frontend
    dockerfile: Dockerfile
    export_cache:
    - okteto.dev/frontend:1.0.0
    - okteto.dev/frontend:2.0.0
```
3. Run the following build command: `okteto build` (this will export cache to `okteto.dev/frontend:1.0.0` and `okteto.dev/frontend:2.0.0` registry)
4. Add `cache_from` field pointing to same values used in `export_cache`:
```yaml
build:
  frontend:
    context: frontend
    dockerfile: Dockerfile
    cache_from:
    - okteto.dev/frontend:1.0.0
    - okteto.dev/frontend:2.0.0
```
5. Create a new namespace
6. Run `okteto build` and check image is cached

##### Scenario 2 - User can provide an string (non breaking previous scenarios)

1. Create a new namespace
2. Create the following manifest:
``` yaml
build:
  frontend:
    context: frontend
    dockerfile: Dockerfile
    export_cache: okteto.dev/frontend:1.0.0
```
3. Run the following build command: `okteto build` (this will export cache to `okteto.dev/frontend:1.0.0`)
4. Add `cache_from` field pointing to same values used in `export_cache`:
```yaml
build:
  frontend:
    context: frontend
    dockerfile: Dockerfile
    cache_from:
    - okteto.dev/frontend:1.0.0
```
5. Create a new namespace
6. Run `okteto build` and check image is cached

##### Scenario 1 - Always export what the user is building

1. Create a new namespace
2. Create the following manifest:
```yaml
build:
  frontend:
    context: frontend
    dockerfile: Dockerfile
    export_cache:
    - okteto.dev/frontend:1.0.0
    image: okteto.global/frontend:1.0.0

```
4. Run the following build command: `okteto build` (This will export cache to `okteto.dev/frontend:1.0.0` and `okteto.global/frontend:1.0.0`)
5. Add `cache_from` field pointing to the global registry:
```yaml
build:
  frontend:
    context: frontend
    dockerfile: Dockerfile
    cache_from:
    - okteto.global/frontend:1.0.0
```
7. Create a new namespace
8. Run `okteto build` and check image is using the cache properly
9. Modify again the manifest to:
```yaml
build:
  frontend:
    context: frontend
    dockerfile: Dockerfile
    cache_from:
    - okteto.dev/frontend:1.0.0
```
7. Create a new namespace
8. Run `okteto build` and check image is using the cache properly

##### Scenario 2 - Always export what the user is building

1. Create a new namespace
2. Create the following manifest:
```yaml
build:
  frontend:
    context: frontend
    dockerfile: Dockerfile
    image: myregistry/ns/frontend:1.0.0

```
4. Run the following build command: `okteto build` (This will export cache to `myregistry/ns/frontend:1.0.0`)
5. Add `cache_from` field pointing to the global registry:
```yaml
build:
  frontend:
    context: frontend
    dockerfile: Dockerfile
    cache_from:
    - myregistry/ns/frontend:1.0.0
```
7. Create a new namespace
8. Run `okteto build` and check image is using the cache properly

jLopezbarb:
# Proposed changes

Fixes #3424

- `export_cache` field admits a string or a list
- `export cache` is added automatically with the same registry (if it's an okteto one) and using the cache tag


pkg/cache/export_cache.go
```diff
@@ -0,0 +1,78 @@
...
+	imageCtrl := reg.GetImageCtrl()
+	imageReg, imageRepo := imageCtrl.GetRegistryAndRepo(image)
+	imageName, _ := imageCtrl.GetRepoNameAndTag(imageRepo)
+
+	if reg.IsGlobalRegistry(image) {
+		newCache := fmt.Sprintf("%s/%s:cache", constants.GlobalRegistry, imageName)
+		pec.add(newCache)
+		return
+	}
+	newDevCache := fmt.Sprintf("%s/%s:cache", imageReg, imageName)
```
andreafalzetti, line 65: In the CacheFrom we always use `constants.DevRegistr` instead of `imageReg` - What's the correct approach?

pkg/cache/export_cache.go
```diff
@@ -0,0 +1,78 @@
...
+package cache
+
+import (
+	"fmt"
+
+	"github.com/okteto/okteto/pkg/constants"
+)
+
+// ExportCache is a list of images that will be created to export the cache.
+type ExportCache []string
```
AdrianPedriza, line 23: How complicated is it to handle it as a map? I think it would improve the data handling internally (for example when adding values we don't need to check if they already exist)

pkg/cache/export_cache.go
```diff
@@ -0,0 +1,78 @@
...
+// AddDefaultPushCache appends the default cache layers for a given image
+func (pec *ExportCache) AddDefaultPushCache(reg oktetoRegistryInterface, image string) {
+	imageCtrl := reg.GetImageCtrl()
+	imageReg, imageRepo := imageCtrl.GetRegistryAndRepo(image)
+	imageName, _ := imageCtrl.GetRepoNameAndTag(imageRepo)
+
+	if reg.IsGlobalRegistry(image) {
+		newCache := fmt.Sprintf("%s/%s:cache", constants.GlobalRegistry, imageName)
+		pec.add(newCache)
+		return
```
AdrianPedriza, line 63: this would not limit the addition of both `okteto.dev` and `okteto.global` if they are not previously defined? with this logic if they are not previously defined only `okteto.global` is added I think if it is active, no?

pkg/cache/export_cache_test.go
```diff
@@ -0,0 +1,136 @@
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			result, err := yaml.Marshal(tt.ec)
+			assert.NoError(t, err)
+			assert.Equal(t, tt.expected, string(result))
+		})
+	}
+}
+
+func Test_AddDefaultPushCache(t *testing.T) {
```
AdrianPedriza, line 94: how about add a test where we need to add a cache layer but is already in `ExportCache` struct? I think that case is not covered

pkg/cache/export_cache_test.go
```diff
@@ -0,0 +1,136 @@
...
+package cache
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	yaml "gopkg.in/yaml.v2"
+)
+
+func TestUnmarshalExportCache(t *testing.T) {
```
AdrianPedriza, line 23: I think we need another case where ExportCache is neither a single item or a list so we have a error trying to unmarshal it

pkg/cache/export_cache.go
```diff
@@ -0,0 +1,78 @@
...
+}
+
+// AddDefaultPushCache appends the default cache layers for a given image
+func (pec *ExportCache) AddDefaultPushCache(reg oktetoRegistryInterface, image string) {
+	imageCtrl := reg.GetImageCtrl()
+	imageReg, imageRepo := imageCtrl.GetRegistryAndRepo(image)
+	imageName, _ := imageCtrl.GetRepoNameAndTag(imageRepo)
+
+	if reg.IsGlobalRegistry(image) {
+		newCache := fmt.Sprintf("%s/%s:cache", constants.GlobalRegistry, imageName)
```
AdrianPedriza, line 61: should we move that tag as a const? I saw it in too different places

pkg/cache/export_cache.go
jLopezbarb: it doesn't matter. I used the same as the user introduced because it could be in its long syntax and avoid the expansion later on

pkg/cache/export_cache_test.go
jLopezbarb: `Test_ExportCacheUnmarshalYAML_WithError` is the one you're looking for :)

pkg/cache/export_cache.go
jLopezbarb: That logic is in the add method. If we find the element we don't add it

andreafalzetti (commented):
LGTM! I will merge into my branch and we can continue there with any chance, if required

PR #3541, opened 2023-04-07
feat(build): improve missing cache error message

### LINKED ISSUE #3420: Warning instead of error on `from_cache` issues while building
## Parent epic/feature

## Description

Right now if we build an image importing a cache from an image that it doesn't exist, Buildkit shows an error that might be misleading by the user:
Captura de Pantalla 2023-03-02 a las 13 48 44

As you can see in the logs above there is an error importing the cache:
```
=> ERROR importing cache manifest from registry.staging.okteto.net/jlopezbarb/aa:okteto                    0.9s
```

That error must be **skipped** and not shown to the user but we must **log** that the user is not able to access the image.

## Acceptance Criteria

- [ ] The log is not shown in tty builds 
- [ ] The log is not shown in plain builds 
- [ ] okteto logs must contain that the user doesn't have access to that image

### Scenarios

To reproduce the issue follow the next scenario and you should be able to get the 

1. Be connected to a okteto cluster (cloud.okteto.com)
2. Create a new namespace
3. Create the following Dockerfile:
```Dockerfile
FROM alpine

RUN echo hola

RUN echo $RANDOM
```
4. Run the following build command: `okteto build --cache-from=okteto.dev/test:okteto`
5. See error 

#### Scenario 1

1. Connect to an okteto cluster
2. Create a new namespace
3. Create the following Dockerfile:
```Dockerfile
FROM alpine

RUN echo hola
```
4. Run the following build command: `okteto build --cache-from=okteto.dev/test:okteto`
5. Check that the error is not shown
 
#### Scenario 2

1. Connect to an okteto cluster
2. Create a new namespace
3. Create the following Dockerfile:
```Dockerfile
FROM alpine

RUN echo hola
```
4. Run the following build command: `okteto build --cache-from=okteto.dev/test:okteto --progress=plain`
5. Check that the error is not shown

#### Scenario 3

1. Connect to an okteto cluster
2. Create a new namespace
3. Create the following Dockerfile:
```Dockerfile
FROM alpine

RUN echo hola
```
4. Run the following build command: `okteto build --cache-from=okteto.dev/test:okteto --progress=plain`
5. Check okteto.log file
6. Check that the error is shown there

andreafalzetti:
Fixes #3420

# Proposed changes

This PR introduces a way to manipulate BuildKit logs. The first use-case is rewriting the error when a cache image is not available.

Example (TTY):

| Before | After |
|--------|-------|
|        |        |

Example (Plain):

| Before | After |
|--------|-------|
|        |        |

## Considerations

I dug into the source code of BuildKit to understand how to display best the logs, however, it feels very hacky to do so. The print utility of BuildKit has a lot of logic to sort and group log entries, so any attempt to manipulate them, will result in a long-term "hack" to maintain. By default, the two messages related to the failed cache-from would be printed together in the same group. (Groups are represented with the prefix "#").

BuildKit doesn't offer any hook or utility to manipulate the logs so one way was adding a synthetic log entry (a Vertex) but I wasn't able to put it in the same group, it will always show in the next group.

Happy to see if there are suggestions to improve the change!

### DIFF
--- a/pkg/cmd/build/log_filter.go
+++ b/pkg/cmd/build/log_filter.go
@@ -7,7 +7,7 @@ import (
 	"regexp"
 )
 
-type ConditionFunc func(vertex *client.Vertex, ss *client.SolveStatus) bool
+type ConditionFunc func(vertex *client.Vertex) bool
 type TransformerFunc func(vertex *client.Vertex, ss *client.SolveStatus, progress string)
 
 type Rule struct {
@@ -28,7 +28,7 @@ func NewBuildKitLogsFilter(rules []Rule) *BuildKitLogsFilter {
 func (lf *BuildKitLogsFilter) Run(ss *client.SolveStatus, progress string) {
 	for _, vertex := range ss.Vertexes {
 		for _, rule := range lf.rules {
-			if rule.condition(vertex, ss) {
+			if rule.condition(vertex) {
 				rule.transformer(vertex, ss, progress)
 			}
 		}
@@ -36,7 +36,7 @@ func (lf *BuildKitLogsFilter) Run(ss *client.SolveStatus, progress string) {
 }
 
 // BuildKitMissingCacheCondition checks if the log contains a missing cache error
-var BuildKitMissingCacheCondition ConditionFunc = func(vertex *client.Vertex, ss *client.SolveStatus) bool {
+var BuildKitMissingCacheCondition ConditionFunc = func(vertex *client.Vertex) bool {
 	importPattern := `importing cache manifest from .*`
 
 	if !regexp.MustCompile(importPattern).MatchString(vertex.Name) {
--- a/pkg/cmd/build/log_filter.go
+++ b/pkg/cmd/build/log_filter.go
@@ -1,0 +1,79 @@
+package build
+
+import (
+	"fmt"
+	"github.com/moby/buildkit/client"
+	oktetoLog "github.com/okteto/okteto/pkg/log"
+	"regexp"
+)
+
+type ConditionFunc func(vertex *client.Vertex, ss *client.SolveStatus) bool
+type TransformerFunc func(vertex *client.Vertex, ss *client.SolveStatus, progress string)
+
+type Rule struct {
+	condition   ConditionFunc
+	transformer TransformerFunc
+}
+
+type BuildKitLogsFilter struct {
+	rules []Rule
+}
+
+func NewBuildKitLogsFilter(rules []Rule) *BuildKitLogsFilter {
+	return &BuildKitLogsFilter{
+		rules: rules,
+	}
+}
+
+func (lf *BuildKitLogsFilter) Run(ss *client.SolveStatus, progress string) {
+	for _, vertex := range ss.Vertexes {
+		for _, rule := range lf.rules {
+			if rule.condition(vertex, ss) {
+				rule.transformer(vertex, ss, progress)
+			}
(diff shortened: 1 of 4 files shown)


pkg/cmd/build/log_filter.go
```diff
@@ -0,0 +1,79 @@
...
+		for _, rule := range lf.rules {
+			if rule.condition(vertex, ss) {
+				rule.transformer(vertex, ss, progress)
+			}
+		}
+	}
+}
+
+// BuildKitMissingCacheCondition checks if the log contains a missing cache error
+var BuildKitMissingCacheCondition ConditionFunc = func(vertex *client.Vertex, ss *client.SolveStatus) bool {
```
AdrianPedriza, line 39:
```suggestion
var BuildKitMissingCacheCondition ConditionFunc = func(vertex *client.Vertex) bool {
```
do we need that param?

pkg/cmd/build/log_filter.go
```diff
@@ -0,0 +1,79 @@
...
+	return &BuildKitLogsFilter{
+		rules: rules,
+	}
+}
+
+func (lf *BuildKitLogsFilter) Run(ss *client.SolveStatus, progress string) {
+	for _, vertex := range ss.Vertexes {
+		for _, rule := range lf.rules {
+			if rule.condition(vertex, ss) {
+				rule.transformer(vertex, ss, progress)
```
jLopezbarb, line 32:
```suggestion
				oktetoLog.Info("could not detect cache image")
				rule.transformer(vertex, ss, progress)
```

jLopezbarb:
Love the approach!! Check to inject the log as debug or info

AdrianPedriza:
overall LGTM! I've one concern about introducing this new logic which is if we should always follow this approach when we face an error relative to using a cache or just use it when the registries involved belong to `okteto.dev `and `okteto.global` (the ones used for us as cache). WDYT? cc @jLopezbarb

jLopezbarb:
> overall LGTM! I've one concern about introducing this new logic which is if we should always follow this approach when we face an error relative to using a cache or just use it when the registries involved belong to `okteto.dev `and `okteto.global` (the ones used for us as cache). WDYT? cc @jLopezbarb

Yes, we should do this as it is (for every registry). This issue was created because most of our users saw the red text and thought that something went wrong and his image was not built properly, but the only meaning of the red text is that buildkit couldn't import the cache from the `cache_from` field

AdrianPedriza (approved):

pkg/cmd/build/log_filter.go
andreafalzetti: actually not, it was used at the beginnign and then removed, good shout! Removed in https://github.com/okteto/okteto/pull/3541/commits/05c6061c2d53e4b0d54addfd2555bce7ef178c0e

pkg/cmd/build/log_filter.go
andreafalzetti: @jLopezbarb the code is generic but the error log would be specific to the cache image. We could move it to the `BuildKitMissingCacheTransformer` func but I think it's already available in logs the message: `[skip] cache image not available: %s`. Do you think we need an extra log? Let me know!

jLopezbarb (approved):

pkg/cmd/build/log_filter.go
andreafalzetti: log added!

PR #3543, opened 2023-04-11
Apply divert transformation in the local deploy proxy

pchico83:
This PR applies the virtual service transformation to inject the divert header on every virtual service created by an `okteto deploy` command using the local kubernetes proxy. This way, the okteto manifest configuration is simplified because these virtual services don't need to be configured in the [deploy.divert.hosts](https://www.okteto.com/docs/reference/manifest/#divert) section of the okteto manifest.

The PR does a refactor on the local kubernetes proxy configuration to pass the divert driver. The divert driver interface is aumented with two new methods to perform pods/virtualservices transformations.

### DIFF
--- a/pkg/divert/istio/divert.go
+++ b/pkg/divert/istio/divert.go
@@ -20,7 +20,9 @@ import (
 	"github.com/okteto/okteto/pkg/k8s/virtualservices"
 	oktetoLog "github.com/okteto/okteto/pkg/log"
 	"github.com/okteto/okteto/pkg/model"
+	istioNetworkingV1beta1 "istio.io/api/networking/v1beta1"
 	istioclientset "istio.io/client-go/pkg/clientset/versioned"
+	apiv1 "k8s.io/api/core/v1"
 	k8sErrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/client-go/kubernetes"
 )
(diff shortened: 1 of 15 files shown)


pkg/divert/istio/divert.go
```diff
@@ -91,8 +93,12 @@ func (d *Driver) Destroy(ctx context.Context) error {
 	return err
 }
 
-func (d *Driver) GetDivertNamespace() string {
-	return ""
+func (d *Driver) UpdatePod(pod apiv1.PodSpec) apiv1.PodSpec {
+	return pod
```
jLopezbarb, line 97: Just to clarify for future revisit of the PR. We have to update the pod only when the weaver implementation and update the vs in the istio implementation

pkg/divert/istio/virtualservices.go
```diff
@@ -80,37 +80,42 @@ func (d *Driver) restoreDivertService(vs *istioV1beta1.VirtualService) *istioV1b
 
 func (d *Driver) translateDivertHost(vs *istioV1beta1.VirtualService) *istioV1beta1.VirtualService {
 	result := vs.DeepCopy()
+	labels.SetInMetadata(&result.ObjectMeta, model.DeployedByLabel, d.name)
+	labels.SetInMetadata(&result.ObjectMeta, model.OktetoAutoCreateAnnotation, "true")
```
ifbyol, line 84: Why we set the auto create annotation now when translating the divert host?

ifbyol (approved):
A question just to know why some change is done but the PR looks good

pkg/divert/istio/divert.go
pchico83: Yes, that's right!

pkg/divert/istio/virtualservices.go
pchico83: This was a bug. If the virtual service already exists in the dev namespace, we shouldn't modify it (other than injecting the header). `model.OktetoAutoCreateAnnotation` header means the virtual service was created by divert from another namespace. In that case, if the virtual service in the other namespace is updated, we need to update also the copy in the developer namespace

ifbyol: Thanks for the explanation!

PR #3544, opened 2023-04-11
Support for custom divert headers

pchico83:
This PR adds support to configure the headers used in the istio divert driver to redirect traffix from the shared namespace to the developer namespace.

There are use cases to use different headers than `x-okteto-divert`. For example, a user might have their application already instrumented using the standard `baggage` header. Their could resue that using this configuration:
```
deploy:
   divert:
      header:
          name: baggage
          match: regex
          value: ^.*\bokteto-divert\\s*=\\s*${OKTETO_NAMESPACE}\b.*$
```
and configuring a `SIDECAR_INBOUND` istio envoy filter to transform the `x-okteto-divert` header into a baggage header in incoming requests.

### DIFF
--- a/pkg/model/const.go
+++ b/pkg/model/const.go
@@ -270,25 +270,4 @@ const (
 
 	// OktetoImageTagWithVolumes is the tag assigned to an image with volume mounts
 	OktetoImageTagWithVolumes = "okteto-with-volume-mounts"
-
-	// OktetoDivertWeaverDriver is the divert driver for weaver
-	OktetoDivertWeaverDriver = "weaver"
-
-	// OktetoDivertIstioDriver is the divert driver for istio
-	OktetoDivertIstioDriver = "istio"
-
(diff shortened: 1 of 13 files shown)


pkg/model/const.go
```diff
@@ -277,6 +277,18 @@ const (
 	// OktetoDivertIstioDriver is the divert driver for istio
 	OktetoDivertIstioDriver = "istio"
 
-	// OktetoDivertHeader the header used by okteto to divert traffic
-	OktetoDivertHeader = "x-okteto-divert"
+	// OktetoDivertIstioExactMatch represents the exact header match type
+	OktetoDivertIstioExactMatch = "exact"
```
jLopezbarb, line 281: Can we moved the constants related to the divert implementation to `pkg/constants`. We want to stop growing the model package

jLopezbarb (approved):

ifbyol (approved):

AdrianPedriza (approved):

PR #3545, opened 2023-04-11
Support for diverting a list of virtual services

pchico83:
This PR adds support to divert several virtual services (potentially from different namespaces) using a single okteto manifest. It also supports the option to filter the destination routes you want to divert, instead of matching the destination ruotes by service name. Both changes increase the flexibility of the divert implementation to support new use cases.

The PR includes a refactor of the spinner messages because we cannot rely on `deploy.divert.namespace` to be available anymore. `deploy.divert.namespace` will be empty in the istio driver

### DIFF
--- a/pkg/divert/istio/divert.go
+++ b/pkg/divert/istio/divert.go
@@ -51,9 +51,15 @@ func New(m *model.Manifest, c kubernetes.Interface, ic istioclientset.Interface)
 }
 
 func (d *Driver) Deploy(ctx context.Context) error {
-	oktetoLog.Spinner(fmt.Sprintf("Diverting service %s/%s...", d.divert.Namespace, d.divert.Service))
-	if err := d.retryTranslateDivertService(ctx); err != nil {
-		return err
+	for i := range d.divert.VirtualServices {
+		oktetoLog.Spinner(fmt.Sprintf("Diverting virtual service %s/%s...", d.divert.VirtualServices[i].Namespace, d.divert.VirtualServices[i].Name))
+		oktetoLog.StartSpinner()
+		defer oktetoLog.StopSpinner()
+		if err := d.retryTranslateDivertVirtualService(ctx, d.divert.VirtualServices[i]); err != nil {
+			return err
+		}
+		oktetoLog.StopSpinner()
+		oktetoLog.Success("Virtual service '%s/%s' successfully diverted", d.divert.VirtualServices[i].Namespace, d.divert.VirtualServices[i].Name)
 	}
 
 	for i := range d.divert.Hosts {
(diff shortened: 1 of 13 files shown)


pkg/divert/istio/divert.go
```diff
@@ -51,9 +51,15 @@ func New(m *model.Manifest, c kubernetes.Interface, ic istioclientset.Interface)
...
-	if err := d.retryTranslateDivertService(ctx); err != nil {
-		return err
+	for i := range d.divert.VirtualServices {
+		oktetoLog.Spinner(fmt.Sprintf("Diverting virtual service %s/%s...", d.divert.VirtualServices[i].Namespace, d.divert.VirtualServices[i].Name))
+		oktetoLog.StartSpinner()
+		defer oktetoLog.StopSpinner()
+		if err := d.retryTranslateDivertVirtualService(ctx, d.divert.VirtualServices[i]); err != nil {
+			return err
+		}
+		oktetoLog.StopSpinner()
```
ifbyol, line 61: Is it ok to stop the spinner twice? Line 57 is using defer to stop it

ifbyol (approved):
Could you include an example on how the divert section would look like with different services? I think it would help to understand better the changes.

PR looks good!

jLopezbarb (approved):

pkg/divert/istio/divert.go
pchico83: yes, it's idempotent

pkg/model/manifest.go
```diff
@@ -194,14 +194,14 @@ type DestroyInfo struct {
 
 // DivertDeploy represents information about the deploy divert configuration
 type DivertDeploy struct {
```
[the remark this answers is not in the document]
pchico83: Not sure if it's doable, the manifest model needs to import these definitions, and the divert package needs to import `model`

pchico83:
> Could you include an example on how the divert section would look like with different services? I think it would help to understand better the changes.
> 
> PR looks good!

@ifbyol this would be a sample:
```
  divert:
    driver: istio
    virtualServices:
      - name: service-b
        namespace: staging
    hosts:
      - virtualService: service-a
        namespace: staging
```
at some point we could also support a simpler syntax for the common case of diverting a single virtual service.
Something like this:
```
  divert:
    driver: istio
    virtualService: service-b
    namespace: staging
    hosts:
      - virtualService: service-a
        namespace: staging
```
but we can do it later. The main priority at the moment is to cover the needed use cases

ifbyol:
> > Could you include an example on how the divert section would look like with different services? I think it would help to understand better the changes.
> > PR looks good!
> 
> @ifbyol this would be a sample:
> 
> ```
>   divert:
>     driver: istio
>     virtualServices:
>       - name: service-b
>         namespace: staging
>     hosts:
>       - virtualService: service-a
>         namespace: staging
> ```
> 
> at some point we could also support a simpler syntax for the common case of diverting a single virtual service. Something like this:
> 
> ```
>   divert:
>     driver: istio
>     virtualService: service-b
>     namespace: staging
>     hosts:
>       - virtualService: service-a
>         namespace: staging
> ```
> 
> but we can do it later. The main priority at the moment is to cover the needed use cases

👍  thanks!

PR #3546, opened 2023-04-11
Translate virtual service in okteto mutation webhook using an annotation

pchico83:
This PR moves the virtual service transformation to the okteto mutation webhook. This is needed to handle changes in the virtual service done by CI tools. Otherwise, an update in the virtual service destroys the divert configuration.

The new implementation injects an annotation in the virtual service, and the transformation is done by the okteto mutation webhook.

### DIFF
--- a/pkg/model/const.go
+++ b/pkg/model/const.go
@@ -291,4 +291,7 @@ const (
 
 	// OktetoDivertDefaultHeaderValue the default divert header value
 	OktetoDivertDefaultHeaderValue = "${OKTETO_NAMESPACE}"
+
+	//OktetoDivertAnnotationTemplate annotation for the okteto mutation webhook to divert a virtual service
+	OktetoDivertAnnotationTemplate = "divert.okteto.com/%s-%s"
 )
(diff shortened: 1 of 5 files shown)


ifbyol (commented):
How are we going to handle the backend and CLI upgrade? I mean, a new CLI with an old backend won't do any kind of translation for virtual services

pchico83:
> How are we going to handle the backend and CLI upgrade? I mean, a new CLI with an old backend won't do any kind of translation for virtual services

@ifbyol it's fine, I already mentioned this to the istio divert users and they will coordinate CLI & backend changes

ifbyol (approved):

pkg/model/const.go
```diff
@@ -291,4 +291,7 @@ const (
 
 	// OktetoDivertDefaultHeaderValue the default divert header value
 	OktetoDivertDefaultHeaderValue = "${OKTETO_NAMESPACE}"
+
+	//OktetoDivertAnnotationTemplate annotation for the okteto mutation webhook to divert a virtual service
+	OktetoDivertAnnotationTemplate = "divert.okteto.com/%s-%s"
```
AdrianPedriza, line 296: can we move this const to` pkg/constants`? we are trying to reduce `model` package

pchico83: sure, I will update the PR :-)

PR #3547, opened 2023-04-12
Fix multicluster preview deploy

jLopezbarb:
# Proposed changes

Fixes https://github.com/okteto/app/issues/5460

- Use the preview used by the user in the flags instead of setting it in the context and using it across the code.
- @maroshii can you create a pipeline installer with these changes. I think the only that could recreate the namespace is creating a namespace, but I'll need to check it out before making the changes

### DIFF
--- a/cmd/preview/deploy.go
+++ b/cmd/preview/deploy.go
@@ -23,7 +23,6 @@ import (
 	"sync"
 	"time"
 
-	contextCMD "github.com/okteto/okteto/cmd/context"
 	"github.com/okteto/okteto/cmd/pipeline"
 	"github.com/okteto/okteto/cmd/utils"
 	"github.com/okteto/okteto/pkg/analytics"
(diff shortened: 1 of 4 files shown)


cmd/preview/deploy.go
```diff
@@ -198,7 +196,7 @@ func (pw *Command) waitUntilRunning(ctx context.Context, name, namespace string,
 	return nil
 }
 func (pw *Command) waitToBeDeployed(ctx context.Context, name string, a *types.Action, timeout time.Duration) error {
-	return pw.okClient.Pipeline().WaitForActionToFinish(ctx, name, okteto.Context().Namespace, a.Name, timeout)
+	return pw.okClient.Pipeline().WaitForActionToFinish(ctx, name, name, a.Name, timeout)
```
ifbyol, line 199: Is this correct? Does the name contains the namespace to be used to query the API?

jLopezbarb: Yes, If we use the namespace stored in the context is not going to be the preview env anymore...

ifbyol: but the flag `--wait` won't work. I mean, that namespace is being used to do a query to the API

jLopezbarb: name is the preview namespace, it should work as before

jLopezbarb:
@maroshii This PR will be merge into your PR. Let me know when this should be merged

maroshii (approved):

PR #3551, opened 2023-04-13
Added dev.okteto.com/namespace label to external resource objects when they are being created

ifbyol:
Partially Fixes # https://github.com/okteto/app/issues/6269

Include the label `dev.okteto.com/namespace` in the external resources when they are created/updated as that would help us to list all external resources within different namespaces in mor efficient way

### DIFF
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -85,4 +85,7 @@ const (
 
 	// OktetoGitCommitEnvVar is the SHA1 hash of the last commit of the branch.
 	OktetoGitCommitEnvVar = "OKTETO_GIT_COMMIT"
+
+	// OktetoNamespaceLabel is the label used to identify the namespace where the resource lives
+	OktetoNamespaceLabel = "dev.okteto.com/namespace"
 )
--- a/pkg/externalresource/control.go
+++ b/pkg/externalresource/control.go
@@ -27,7 +27,7 @@ func (c *K8sControl) Deploy(ctx context.Context, name, ns string, er *ExternalRe
 		return fmt.Errorf("error creating external CRD client: %s", err.Error())
 	}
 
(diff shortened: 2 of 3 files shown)


pkg/externalresource/control.go
```diff
@@ -99,9 +99,13 @@ func translate(name string, externalResource *ExternalResource) *k8s.External {
 			APIVersion: fmt.Sprintf("%s/%s", k8s.GroupName, k8s.GroupVersion),
 		},
 		ObjectMeta: metav1.ObjectMeta{
-			Name: format.ResourceK8sMetaString(name),
+			Name:      format.ResourceK8sMetaString(name),
+			Namespace: ns,
```
ifbyol, line 103: We were not specifying the namespace but now that we have it, it doesn't hurt to include it

AdrianPedriza (approved):
LGTM!

pkg/constants/constants.go
```diff
@@ -85,4 +85,7 @@ const (
 
 	// OktetoGitCommitEnvVar is the SHA1 hash of the last commit of the branch.
 	OktetoGitCommitEnvVar = "OKTETO_GIT_COMMIT"
+
+	// OktetoNamespaceLabel is the label used to identify the namespace where the resource lives
+	OktetoNamespaceLabel = "dev.okteto.com/namespace"
```
jLopezbarb, line 90: What's the idea with this new constant? Should we include it in every resource? Or just on CRDs?

ifbyol: Just for the external resources or any object in which we want to do queries by that. In this case, it is to optimize how we retrieve the endpoints for external resources from a list of namespaces

jLopezbarb (approved):

PR #3552, opened 2023-04-13
Improve kubetoken performance with a cache #5858

AgustinRamiroDiaz:
# Proposed changes

Fixes https://github.com/okteto/app/issues/5858

Uses a file cache that will look something like this:
```
{
	"{\"context\":\"https://okteto.agustin.dev.okteto.net\",\"namespace\":\"agustinramirodiaz\"}": {
		"token": {...}
         }

	"{\"context\":\"https://cloud.okteto.net\",\"namespace\":\"my-cloud-ns\"}": {
		"token": {...}
         }

}
```

It's not the prettiest, but I think it has advantages:
- We can use structs as keys
- We rely on a lot of "automatic" code from json marshallers, so we don't need to create our custom marshalling
- It's a map so it's got efficient lookup

---

Tested scenarios:
- token expired
- token did not expire
- cache file does not exist
- cache file is corrupted
- some cache registry token field is wrongly marshaled 

All of them work correctly and self-heal from the bad states by re-requesting the token and cleaning up the cache file

---

The implementation details have 3 levels of abstraction that build on top of each other:
- A `byte[]` storage
- A cache
- And the kubetoken client uses the cache

### DIFF
--- a/cmd/kubetoken/kubetoken.go
+++ b/cmd/kubetoken/kubetoken.go
@@ -43,13 +43,13 @@ You can find more information on 'ExecCredential' and 'client side authenticatio
 
 	cmd.RunE = func(_ *cobra.Command, args []string) error {
 		ctx := context.Background()
-		context := args[0]
+		contextName := args[0]
 		namespace := args[1]
 
 		cacheFileName := path.Join(config.GetDotKubeFolder(), oktetoTokenCacheFileName)
 		cache := kubetoken.NewCache(cacheFileName)
 		// Return early if we have a valid token in the cache before we run the context command to improve performance
-		if token, err := cache.Get(context, namespace); err == nil && token != "" {
+		if token, err := cache.Get(contextName, namespace); err == nil && token != "" {
 			cmd.Print(token)
 			return nil
 		} else {
@@ -57,18 +57,19 @@ You can find more information on 'ExecCredential' and 'client side authenticatio
 		}
 
 		err := contextCMD.NewContextCommand().Run(ctx, &contextCMD.ContextOptions{
-			Context:   context,
+			Context:   contextName,
 			Namespace: namespace,
 		})
 		if err != nil {
 			return err
 		}
 
-		if !okteto.Context().IsOkteto {
+		okctx := okteto.Context()
+		if !okctx.IsOkteto {
 			return errors.ErrContextIsNotOktetoCluster
 		}
 
-		c, err := kubetoken.NewClient(okteto.Context().Name, okteto.Context().Token, namespace, cache)
+		c, err := kubetoken.NewClient(okctx.Name, okctx.Token, namespace, okctx.Certificate, okctx.IsInsecure, cache)
 		if err != nil {
 			return fmt.Errorf("failed to initialize the kubetoken client: %w", err)
 		}
--- a/cmd/kubetoken/kubetoken.go
+++ b/cmd/kubetoken/kubetoken.go
@@ -24,6 +24,7 @@ import (
 	"github.com/okteto/okteto/pkg/config"
 	"github.com/okteto/okteto/pkg/errors"
 	"github.com/okteto/okteto/pkg/okteto"
+	"github.com/okteto/okteto/pkg/okteto/kubetoken"
 	"github.com/spf13/cobra"
 )
 
@@ -45,7 +46,7 @@ You can find more information on 'ExecCredential' and 'client side authenticatio
 		namespace := args[1]
 
 		cacheFileName := path.Join(config.GetDotKubeFolder(), oktetoTokenCacheFileName)
-		cache := &okteto.KubeTokenCache{StringStore: &okteto.FileByteStore{FileName: cacheFileName}}
+		cache := &kubetoken.Cache{StringStore: &kubetoken.FileByteStore{FileName: cacheFileName}}
 		// Return early if we have a valid token in the cache before we run the context command to improve performance
 		if token, err := cache.Get(context, namespace); err == nil && token != "" {
 			cmd.Print(token)
@@ -64,7 +65,7 @@ You can find more information on 'ExecCredential' and 'client side authenticatio
 			return errors.ErrContextIsNotOktetoCluster
 		}
 
-		c, err := okteto.NewKubeTokenClient(okteto.Context().Name, okteto.Context().Token, namespace, cache)
+		c, err := kubetoken.NewClient(okteto.Context().Name, okteto.Context().Token, namespace, cache)
 		if err != nil {
 			return fmt.Errorf("failed to initialize the kubetoken client: %w", err)
 		}
--- a/cmd/kubetoken/kubetoken.go
+++ b/cmd/kubetoken/kubetoken.go
@@ -45,8 +45,8 @@ You can find more information on 'ExecCredential' and 'client side authenticatio
 		context := args[0]
 		namespace := args[1]
 
-		cacheFile := path.Join(config.GetDotKubeFolder(), oktetoTokenCacheFileName)
-		cache := &okteto.FileCache{File: cacheFile}
+		cacheFileName := path.Join(config.GetDotKubeFolder(), oktetoTokenCacheFileName)
+		cache := &okteto.KubeTokenFileCache{File: cacheFileName}
 		if token, err := cache.Get(context, namespace); err == nil && token != nil {
 			tokenString, _ := json.MarshalIndent(token, "", "\t")
 
--- a/cmd/kubetoken/kubetoken.go
+++ b/cmd/kubetoken/kubetoken.go
@@ -46,7 +46,7 @@ You can find more information on 'ExecCredential' and 'client side authenticatio
 		namespace := args[1]
 
 		cacheFileName := path.Join(config.GetDotKubeFolder(), oktetoTokenCacheFileName)
-		cache := &kubetoken.Cache{StringStore: &kubetoken.FileByteStore{FileName: cacheFileName}}
+		cache := &kubetoken.Cache{StringStore: kubetoken.NewFileByteStore(cacheFileName)}
 		// Return early if we have a valid token in the cache before we run the context command to improve performance
 		if token, err := cache.Get(context, namespace); err == nil && token != "" {
 			cmd.Print(token)
--- a/cmd/kubetoken/kubetoken.go
+++ b/cmd/kubetoken/kubetoken.go
@@ -15,7 +15,6 @@ package kubetoken
 
 import (
 	"context"
-	"encoding/json"
 	"fmt"
 	"os"
 	"path"
@@ -47,10 +46,8 @@ You can find more information on 'ExecCredential' and 'client side authenticatio
 
 		cacheFileName := path.Join(config.GetDotKubeFolder(), oktetoTokenCacheFileName)
 		cache := &okteto.KubeTokenFileCache{File: cacheFileName}
-		if token, err := cache.Get(context, namespace); err == nil && token != nil {
-			tokenString, _ := json.MarshalIndent(token, "", "\t")
-
-			cmd.Print(string(tokenString))
+		if token, err := cache.Get(context, namespace); err == nil {
+			cmd.Print(token)
 			return nil
 		}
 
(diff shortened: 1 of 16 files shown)


cmd/kubetoken/kubetoken.go
```diff
@@ -37,22 +43,33 @@ You can find more information on 'ExecCredential' and 'client side authenticatio
 
 	cmd.RunE = func(_ *cobra.Command, args []string) error {
 		ctx := context.Background()
-		context := args[0]
+		contextName := args[0]
 		namespace := args[1]
 
+		cacheFileName := path.Join(config.GetDotKubeFolder(), oktetoTokenCacheFileName)
```
jLopezbarb, line 49: What if I have set my `KUBECONFIG` env var pointing to another location?

pkg/okteto/kubetoken/does-not-exist
jLopezbarb, line 1: Do we need this file?

pkg/okteto/kubetoken/cache.go
```diff
@@ -0,0 +1,122 @@
+package kubetoken
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"time"
+
+	"github.com/okteto/okteto/pkg/log"
```
jLopezbarb, line 9: Let's use the same name as in the rest of the code for consistency
```suggestion
	oktetoLog "github.com/okteto/okteto/pkg/log"
```

pkg/okteto/kubetoken/kubetoken_client.go
```diff
@@ -11,42 +11,54 @@
...
 		httpClient:  httpClient,
 		url:         url,
 		contextName: contextName,
+		namespace:   namespace,
+		cache:       cache,
 	}, nil
 }
 
-func (c *KubeTokenClient) GetKubeToken() (string, error) {
+func (c *Client) GetKubeToken() (string, error) {
```
jLopezbarb, line 61: Maybe we should have this under the OktetoClient since it's related and we don't export another client for just one function. WDYT?

pkg/okteto/kubetoken/cache.go
```diff
@@ -0,0 +1,122 @@
...
+	authenticationv1 "k8s.io/api/authentication/v1"
+)
+
+const cacheIsCorrupted = "cache is corrupted"
```
jLopezbarb, line 13: If we want to have it as a comparable error to use errors.Is on this we should define it as `errors.New("cache is corrupted")` and var. After that you can compare it on your tests and in other aprts of the code

pkg/okteto/kubetoken/cache.go
```diff
@@ -0,0 +1,122 @@
...
+type storeRegister struct {
+	ContextName string                        `json:"context"`
+	Namespace   string                        `json:"namespace"`
+	Token       authenticationv1.TokenRequest `json:"token"`
+}
+
+func (c *Cache) read() ([]storeRegister, error) {
+	contents, err := c.StringStore.Get()
+	if err != nil {
+		return nil, err
```
jLopezbarb, line 43: Let's add some information about the errors. Otherwise it's missing information when we try to debug them:
```suggestion
		return nil, fmt.Errorf("error while trying to get kubetoken cache: %w", err)
```

pkg/okteto/kubetoken/cache.go
```diff
@@ -0,0 +1,122 @@
...
+				return "", nil
+			}
+		}
+	}
+
+	return "", nil
+}
+
+func updateStore(store []storeRegister, contextName, namespace string, token authenticationv1.TokenRequest) []storeRegister {
+	existed := false
```
jLopezbarb, line 84: Looking at this function I'm thinking that storeRegister is more like a map than a list and could reduce the complexity of iterating always to look for the right one. Let me know if I'm wrong

pkg/okteto/kubetoken/store.go
```diff
@@ -0,0 +1,52 @@
...
+		osCreate:   os.Create,
+		osReadFile: os.ReadFile,
+		writeFile: func(filename string, data []byte) error {
+			return os.WriteFile(filename, data, 0600)
+		},
+	}
+}
+
+type FileByteStore struct {
+	FileName   string
```
jLopezbarb, line 22: Just a recommendation. We use `afero` to mock all the OS calls. Check it out because you will just need one field

pkg/okteto/kubetoken/cache.go
```diff
@@ -0,0 +1,122 @@
...
+	newStore, err := json.MarshalIndent(store, "", "\t")
+	if err != nil {
+		return err
+	}
+
+	return c.StringStore.Set(newStore)
+}
+
+func (c *Cache) Set(contextName, namespace string, token authenticationv1.TokenRequest) {
+	if err := c.setWithErr(contextName, namespace, token); err != nil {
```
jLopezbarb, line 119: It seems odd that the only thing that the `Set` function does is calling to this other function. Should we move the code from `setWithErr` here

AgustinRamiroDiaz: @jLopezbarb I wanted to test the errors from `setWithErr` to guarantee the functionality, but on the other hand the `kubetoken_client` won't care of any cache errors so the interface is with `Set`.

An alternative could be to have the interface with `Set(...) error` and then ignore it, but I feel we'd be forcing the interface from the side of the creator and not from the consumer

WDYT?

pkg/okteto/kubetoken/cache.go
AgustinRamiroDiaz: Yeah that's true, the thing is that we'd need to merge the `contextName + namespace` into a single string in order to be able to store it as a key. I was afraid of getting into that due to the operator, but I think a `:` should be fine right? like `okteto_cloud:agustin`

pkg/okteto/kubetoken/does-not-exist
AgustinRamiroDiaz: Misscommited, fixed!

cmd/kubetoken/kubetoken.go
AgustinRamiroDiaz: @jLopezbarb the `KUBECONFIG` envvar points to a file, not a folder. In order to keep things correctly separated, I think it's better to have a custom file. Do you think of a way that `KUBECONFIG` could get in the way here?

This comment has made me realize that we are not handling when the `.kube` folder doesn't exist though

pkg/okteto/kubetoken/store.go
```diff
@@ -0,0 +1,59 @@
...
+)
+
+func NewFileByteStore(fileName string) *FileByteStore {
+	return &FileByteStore{
+		FileName: fileName,
+		osStat:   os.Stat,
+		createFile: func(filename string) (*os.File, error) {
+			fp := filepath.Dir(filename)
+			folder := filepath.Base(filename)
+			if err := os.MkdirAll(fp, 0764); err != nil {
```
AgustinRamiroDiaz, line 16: I couldn't get these permission to `0664` because I was hitting `stat /home/az/.kubetest/okteto_auth_plugin_cache.json: permission denied 0xc0006a9e00` when the folder didn't exist

pkg/okteto/kubetoken/store.go
AgustinRamiroDiaz: I've been trying to migrate to `afero` but I had an issue. How do I mock a call to give me an error? Like for example, I want to test if `os.Stat` returns an error

pkg/okteto/kubetoken/kubetoken_client.go
AgustinRamiroDiaz: Personally I prefer it as it is since we get a more defined boundary on dependencies. I'm flexible to discuss :D

maroshii (commented):
I jumped into this PR when I read "cache" because it's one of those things that seem simple but are not and could lead to stuff breaking in unexpected ways, specially if there's no opt-out. JWTs are stateless and this makes them stateful, this change doesn't come for free.

Leaving aside the implementation, the issue is about "performance". How are we measuring it? Is this currently a bottleneck? The epic says _slowing down commands_ but by how much? What's the added benefit of this and is it noticeable? I don't see any metrics and I think we should understand that before anything else and compare.


---
A few comments regarding the implementation:

- If I'm following correctly when a context is deleted, we are not deleting the token from the cache  since I don't see a `Delete` method. Also, how are we making sure that the cache doesn't grow indefinitely and that it's in sync with the namespace? If I delete a namespace, and create it again with the same name, will it work?
- As Javi pointed out, using a map instead of a list might be a better approach since we would always read at `O(1)`. For this a calculated key could work but you could also consider a nested structure `map[ctx]map[ns][token]` to avoid calculated keys.
- An alternative approach is to have a `.cache` folder and store the tokens directly as files and just read the token from the file (if it exists):
```
cmd/.cache/
├── ctx1
│   ├── ns1 <-------- this is the jwt
│   └── ns2
├── ctx2
│   ├── ns1
│   └── ns2
└── ctx3
    ├── ns1
    └── ns2
```

When a namespace gets deleted we just delete that file. If a context gets deleted, we delete the folder.

AgustinRamiroDiaz:
@maroshii thanks for the feedback! Yeah, I agree that cache is something that sounds easy but it's super delicate

Regarding the time performance:
- without the cache, the round trip to the backend is approx 500ms for getting the token
- with the cache, it's approx 50ms

This is done for each kubernetes request in the CLI

I think the biggest advantage is with tools like kubectl, where we do 2 roundtrips (1 to get the token + another one to get the data). In a classic `okteto deploy, okteto up` we won't see a significant improvement since we have other container stuff that's way slower (pull image, build image, create deployment)

I'll try to get better samples for tomorrow

AgustinRamiroDiaz:
@maroshii regarding implementation:
- Yeah, I didn't take into account the deletion of contexts and namespaces. I think we could go for the deletion depending on the token expiry state
- I like the idea of the nested map. I'm a bit afraid with null maps here since they tend to panic, but I'll check it out
- I'm not a fan of creating a whole directory structure here. I think it's more complicated that it needs to be since we need to handle a lot of cases (we can already see a lot of error handling with only one file). I think in memory strings shouldn't be a problem for this cache, but please comment if you think differently.
  - also, when issues related to files arise with other OSs (cough windows) it's harder to debug

AgustinRamiroDiaz:
@maroshii I really liked your comment about "possibility to opt out" so I implemented it in my last commit. This allowed me to easily benchmark the command.

Here's the benchmar for the `kubetoken` command:
```
$ time for run in {1..100}; do lokteto kubetoken https://cloud.okteto.com b-agustinramirodiaz > /dev/null; done

real    0m4,910s
user    0m6,892s
sys     0m2,605s

$ export OKTETO_KUBETOKEN_NO_CACHE=true
$ time for run in {1..100}; do lokteto kubetoken https://cloud.okteto.com b-agustinramirodiaz > /dev/null; done

real    2m15,137s
user    0m9,650s
sys     0m4,243s
```

For other okteto commands, we need to add https://github.com/okteto/okteto/pull/3515 into the mix. I'll get some metrics for that

AgustinRamiroDiaz:
:point_up: Additional metrics

```
$ time for run in {1..100}; do k get pod > /dev/null; done

real    0m20,965s
user    0m12,348s
sys     0m4,041s

$ time for run in {1..100}; do k get pod > /dev/null; done

real    1m14,800s
user    0m15,784s
sys     0m4,760s
```

AgustinRamiroDiaz:
:point_up: Additional metrics for `okteto deploy` with `go-getting-started` already deployed in the ns
```
$ time for run in {1..20}; do ok deploy > /dev/null; done 
real    1m6,863s
user    0m12,642s
sys     0m3,508s

$ export OKTETO_KUBETOKEN_NO_CACHE=true
$ time for run in {1..20}; do ok deploy > /dev/null; done 
real    1m14,283s
user    0m12,711s
sys     0m3,500s
```

AgustinRamiroDiaz:
:point_up: Additional metrics for `okteto deploy + okteto destroy` with go-getting started

```
$ time okteto deploy
real    0m3,773s
user    0m0,626s
sys     0m0,200s
$ time okteto destroy
real    0m2,408s
user    0m0,302s
sys     0m0,091s

$ export OKTETO_KUBETOKEN_NO_CACHE=true
$ time okteto deploy
real    0m4,412s
user    0m0,702s
sys     0m0,185s
$ time okteto destroy
real    0m3,061s
user    0m0,310s
sys     0m0,106s
```

AgustinRamiroDiaz:
@maroshii from the metrics I conclude that there's a noticable improvement in time, enough to feel it. BUT we should evaluate the risks we are taking from this since the difference is not abysmal

PR #3557, opened 2023-04-14
fix: e2e test for remote deploy

jLopezbarb:
# Proposed changes

Fixes e2e test that is failing

- Use a different getOpts for the remote deploy. There is no point in getting all the options for the remote build if we are going to unset it after it

### DIFF
--- a/pkg/cmd/build/build.go
+++ b/pkg/cmd/build/build.go
@@ -341,6 +341,7 @@ func OptsFromBuildInfo(manifestName, svcName string, b *model.BuildInfo, o *type
 	return opts
 }
 
+// OptsFromBuildInfoForRemoteDeploy returns the options for the remote deploy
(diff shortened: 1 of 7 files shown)


andreafalzetti (approved):

pkg/cmd/build/build.go
```diff
@@ -341,6 +341,16 @@ func OptsFromBuildInfo(manifestName, svcName string, b *model.BuildInfo, o *type
 	return opts
 }
 
+func OptsFromBuildInfoForRemoteDeploy(b *model.BuildInfo, o *types.BuildOptions) *types.BuildOptions {
```
ifbyol, line 344: Nit: being an exported function we could add a comment in the function

ifbyol (approved):

PR #3559, opened 2023-04-14
Automatically create a namespace if needed while doing okteto up

### LINKED ISSUE #3452: Automatically create a namespace if needed
**Describe the bug**
When I have `namespace` in my okteto manifest and the namespace doesn't exist, we are automcatically creating the namespace in Okteto clusters when running an Okteto CLI command. But this is not the case if using the `-n namespace` flag. We should be consistent and create the namespace when using the `-n` flag too.

**To Reproduce**

1. Clone https://github.com/okteto/go-getting-started
1. Run `okteto up -n my-new-namespace`
1. It fails with a low level error because the namespace doesn't exist

**Expected behavior**
The namespace is create and the command works

AnshulAngaria:
# Proposed changes

Fixes #3452

Added the logic to automatically create namespace if it does not exists, while running `okteto up` command.

### DIFF
--- a/cmd/up/up.go
+++ b/cmd/up/up.go
@@ -30,6 +30,7 @@ import (
 	contextCMD "github.com/okteto/okteto/cmd/context"
 	"github.com/okteto/okteto/cmd/deploy"
 	"github.com/okteto/okteto/cmd/manifest"
+	"github.com/okteto/okteto/cmd/namespace"
 	"github.com/okteto/okteto/cmd/utils"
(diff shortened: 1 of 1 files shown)


AdrianPedriza:
thanks for your contribution! I used it on my machine and it worked great 👍

AdrianPedriza (approved):

teresaromero (approved):

pchico83 (approved):
Thanks @AnshulAngaria for your contribution!

PR #3561, opened 2023-04-17
load envs in cmap when deploy

### LINKED ISSUE #3482: Expose `dependency` envvars to the main pipeline
I want to access the environment variables of my dependencies. A common use case is to access a randomly generated password.

The variables of a dependency will be available as `OKTETO_DEPENDENCY_{DEPENDENCY_NAME}_VARIABLE_{VARIABLE_NAME}`.

Same thing for external resource endpoints, using the envvar `OKTETO_DEPENDENCY_{DEPENDENCY_NAME}_ENDPOINT_{ENDPOINT_NAME}`.

If `wait` is false for the dependency, only the original variables/endpoints are available.
If `wait` is true, we will also get the value of dynamic envvars generated using `OKTETO_ENV`.
That means we will wait until the dependency is deployed, and then, get its variables.

AdrianPedriza:
# Proposed changes

- load in config map envs from OKTETO_ENV file d8fa6ed43496081864a2aaf3317e28535360c2d8
- load in config map envs defined by deploy flag `-v` d8fa6ed43496081864a2aaf3317e28535360c2d8
- consume envs from main pipeline 1584eec97fa746c58ba216824236f846c82e5604

Fixes #3482

# Test scenario (with an installer using this changes)
- clone https://github.com/okteto/movies
- modify `okteto.yml` adding a dependency:
```
dependencies:
  depend:
    repository: https://github.com/okteto/go-getting-started
    branch: adrian/test-3482
    wait: true
```
- consume an env generated in dependency deployment (add `echo $OKTETO_DEPENDENCY_DEPEND_VARIABLE_OKTETO_RANDOM2_PASS`) as a deploy step
- check value is `P@sSvV0rD333`


cmd/deploy/cfghandler.go
```diff
@@ -29,6 +29,7 @@ import (
 type configMapHandler interface {
 	translateConfigMapAndDeploy(context.Context, *pipeline.CfgData) (*apiv1.ConfigMap, error)
 	updateConfigMap(context.Context, *apiv1.ConfigMap, *pipeline.CfgData, error) error
+	updateEnvs(context.Context, string, string, []string) error
```
jLopezbarb, line 32: Can we be more explicit? 
```suggestion
	updateEnvsFromCommands(context.Context, string, string, []string) error
```

pkg/cmd/pipeline/translate.go
```diff
@@ -136,6 +136,26 @@ func UpdateConfigMap(ctx context.Context, cmap *apiv1.ConfigMap, data *CfgData,
 	return configmaps.Deploy(ctx, cmap, cmap.Namespace, c)
 }
 
+// UpdateConfigMap updates the configmaps fields
+func UpdateEnvs(ctx context.Context, name, namespace string, envs []string, c kubernetes.Interface) error {
```
jLopezbarb, line 140: Can we add a test for this func

pkg/k8s/configmaps/crud.go
```diff
@@ -83,7 +83,7 @@ func Create(ctx context.Context, cf *apiv1.ConfigMap, namespace string, c kubern
 	return nil
 }
 
-func update(ctx context.Context, cf *apiv1.ConfigMap, namespace string, c kubernetes.Interface) error {
+func Update(ctx context.Context, cf *apiv1.ConfigMap, namespace string, c kubernetes.Interface) error {
```
jLopezbarb, line 86: We should probably have to use the deploy since it¡s calling the update inside and we want to check line 62

cmd/deploy/local.go
```diff
@@ -224,6 +226,11 @@ func (ld *localDeployer) runDeploySection(ctx context.Context, opts *Options) er
 		oktetoLog.SetStage("")
 	}
 
+	err = ld.ConfigMapHandler.updateEnvs(ctx, opts.Name, opts.Manifest.Namespace, opts.Variables)
```
jLopezbarb, line 229: Are you using the right variables? opts.Variables are the ones defined by the user, or am I wrong?

jLopezbarb: Oh I didn't see that we were adding all the variables from the okteto env file to opts.variables

cmd/deploy/cfghandler.go
AdrianPedriza: done f3b410440639b89a763e32b632327c3fd024fbc9

cmd/pipeline/deploy.go
```diff
@@ -203,10 +204,35 @@ func (pc *Command) ExecuteDeployPipeline(ctx context.Context, opts *DeployOption
...
 
+	if err := pc.setEnvsFromDependency(ctx, opts.Name, opts.Namespace, c); err != nil {
+		return fmt.Errorf("could not set environment variable generated by dependency '%s'", opts.Name)
+	}
+
 	oktetoLog.Success("Repository '%s' successfully deployed", opts.Name)
 	return nil
 }
 
+func (pc *Command) setEnvsFromDependency(ctx context.Context, name, namespace string, c kubernetes.Interface) error {
```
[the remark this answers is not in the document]
AdrianPedriza: done c0ca84d004da1001196d542c893991495cf1a4ea

pkg/k8s/configmaps/crud.go
AdrianPedriza: done 7778daef659b3822220e44128db9a7119a214a70

pkg/cmd/pipeline/translate.go
AdrianPedriza: done cb978af6de11e2423a14ec3d7cb70306b6ec7e9c

cmd/deploy/cfghandler.go
```diff
@@ -81,6 +82,18 @@ func (ch *defaultConfigMapHandler) updateConfigMap(ctx context.Context, cfg *api
 	return errMain
 }
 
+func (h *defaultConfigMapHandler) updateEnvsFromCommands(ctx context.Context, name, namespace string, envs []string) error {
```
teresaromero, line 85: can we add a comment on what does this func do?

AdrianPedriza: done b811281808fb903ba71b9d926f73e713a2141479

pkg/cmd/pipeline/translate.go
```diff
@@ -136,6 +136,27 @@ func UpdateConfigMap(ctx context.Context, cmap *apiv1.ConfigMap, data *CfgData,
...
+	}
+
+	if cmap != nil {
+		for _, env := range envs {
+			result := strings.Split(env, "=")
+			if len(result) != 2 {
+				return fmt.Errorf("invalid env format: '%s'", env)
+			}
+
+			cmap.Data[fmt.Sprintf("%s_%s", constants.OktetoDependencyEnvPreffix, result[0])] = result[1]
```
AdrianPedriza: we need to add a prefix in order to know which values in the config map should be retrieved and set it as an environment variable

pkg/cmd/pipeline/translate.go
```diff
@@ -136,6 +136,27 @@ func UpdateConfigMap(ctx context.Context, cmap *apiv1.ConfigMap, data *CfgData,
...
+	}
+
+	if cmap != nil {
+		for _, env := range envs {
+			result := strings.Split(env, "=")
+			if len(result) != 2 {
+				return fmt.Errorf("invalid env format: '%s'", env)
+			}
+
+			cmap.Data[fmt.Sprintf("%s_%s", constants.OktetoDependencyEnvPrefix, result[0])] = result[1]
```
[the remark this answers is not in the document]
AdrianPedriza: I like the idea. We could use a more descriptive name for all the envs stored over a unique key. Something like `dependency_envs`..?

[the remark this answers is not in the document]
AdrianPedriza: Thinking more about it, I don't see a great added value with this change. In the end, the config mas storage limit will be the same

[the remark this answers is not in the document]
AdrianPedriza: done b3e34686678c4b3821eb38ba484d1fb20002dd5b

cmd/pipeline/deploy.go
```diff
@@ -221,13 +227,24 @@ func setEnvsFromDependency(ctx context.Context, name, namespace string, c kubern
 
 	if cmap != nil {
 		for k, v := range cmap.Data {
```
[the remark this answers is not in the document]
AdrianPedriza: you are right!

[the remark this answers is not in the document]
AdrianPedriza: done 6b9f5288ec3f4043f4092c8f0389b4dc5b108839

jLopezbarb (approved):

PR #3566, opened 2023-04-18
fix device resource parsing

### LINKED ISSUE #3563: Okteto silently discards most Device Manager resource claims
**Describe the bug**

Any device management statement in the `resource: limits` section that is not either `nvidia.com/gpu` or `amd.com/gpu` is silently discarded.

**To Reproduce**

We use for example the FUSE plugin:

```yaml

context: $OKTETO_CONTEXT
dev:
  container-sdk:
    autocreate: true
    initContainer:
      image: okteto/bin:1.4.2
    image: quay.io/buildah:latest
    command:
    - bash
    - -l
    persistentVolume:
      enabled: false
    sync:
    - .:/src
    resources:
      limits:
        github.com/fuse: 1
    volumes:
    - /var/lib/containers
    workdir: /src
```

**Expected behavior**

Device Management and it's general syntax is GA since 1.26; `<device namespace>/<resource>: <value>` is the more generic syntax.
See also for example also the [Generic Device Plugin](https://github.com/squat/generic-device-plugin) (`squat.ai/video: 1`) or [Intel's drivers](https://github.com/intel/intel-device-plugins-for-kubernetes) (`sgx.intel.com: epc`)

- OS: RHEL
- Version 7/8, Okteto 2.14.2

**Additional context**
`okteto` seems hardcoded to only evaluate and pass on `nvidia.com/gpu`  and `amd.com/gpu`:
https://github.com/okteto/okteto/blob/71bd808464b80980bf5fc361cb1bc8156835f7c8/pkg/k8s/apps/translate.go#L263
https://github.com/okteto/okteto/blob/71bd808464b80980bf5fc361cb1bc8156835f7c8/pkg/model/const.go#L159

The-Loeki:
# Proposed changes

Fixes #3563

* rather than taking two hardcoded values, all keys in resources are selected by containing '/' and considered as device links
* typo fix for the `syncFieldDocsURL` const

### DIFF
--- a/pkg/k8s/apps/translate.go
+++ b/pkg/k8s/apps/translate.go
@@ -243,53 +243,33 @@ func TranslateLifecycle(c *apiv1.Container, l *model.Lifecycle) {
 }
 
 // TranslateResources translates the resources attached to a container
-func TranslateResources(c *apiv1.Container, r model.ResourceRequirements) {
-	if c.Resources.Requests == nil {
-		c.Resources.Requests = make(map[apiv1.ResourceName]resource.Quantity)
-	}
+func TranslateResources(c *apiv1.Container, rr model.ResourceRequirements) {
 
-	if v, ok := r.Requests[apiv1.ResourceMemory]; ok {
-		c.Resources.Requests[apiv1.ResourceMemory] = v
-	}
+	rrl := []model.ResourceList{rr.Requests, rr.Limits}
+	cr := []*apiv1.ResourceList{&c.Resources.Requests, &c.Resources.Limits}
 
-	if v, ok := r.Requests[apiv1.ResourceCPU]; ok {
-		c.Resources.Requests[apiv1.ResourceCPU] = v
-	}
+	for i, crl := range cr {
+		rl := rrl[i]
 
-	if v, ok := r.Requests[apiv1.ResourceEphemeralStorage]; ok {
-		c.Resources.Requests[apiv1.ResourceEphemeralStorage] = v
-	}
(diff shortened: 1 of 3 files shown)


pkg/k8s/apps/translate.go
```diff
@@ -260,14 +262,13 @@ func TranslateResources(c *apiv1.Container, r model.ResourceRequirements) {
...
 	}
 
-	if v, ok := r.Requests[model.ResourceAMDGPU]; ok {
-		c.Resources.Requests[model.ResourceAMDGPU] = v
-	}
-
-	if v, ok := r.Requests[model.ResourceNVIDIAGPU]; ok {
-		c.Resources.Requests[model.ResourceNVIDIAGPU] = v
+	for resname, v := range r.Requests {
+		if strings.Contains(string(resname), model.ResourceNamespaceSeparator) {
```
jLopezbarb, line 266: Let's move the constant to the top of the file, I think it'd be better since we are just using it here. WDYT?

jLopezbarb (approved):
Nice PR and nice addition @The-Loeki. Thanks a lot for your contribution

AdrianPedriza (approved):

pkg/k8s/apps/translate.go
The-Loeki: How 'bout we reduce the duplication by a simple loop; now it's not even used more than once,  and it doesn't need to be a const at all?

pchico83 (approved):
Thanks for your contribution @The-Loeki , great PR!

PR #3568, opened 2023-04-19
Remove preview from deployPreview mutation response

maroshii:
We don't really need the preview object in the deploy mutation given that the mutation returns the id at the top level.

### DIFF
--- a/pkg/okteto/preview.go
+++ b/pkg/okteto/preview.go
@@ -107,8 +107,8 @@ type previewEnv struct {
 }
 
 type deployPreviewResponse struct {
-	Action  actionStruct
-	Preview previewIDStruct
+	Id     graphql.String
+	Action actionStruct
 }
 
 type previewIDStruct struct {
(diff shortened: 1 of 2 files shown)


maroshii:
> This change is not retrocompatible with previous versions. Maybe we can catch the error and create a deprecated function for it. We have done that [here](https://github.com/okteto/okteto/blob/master/pkg/okteto/secrets.go#L95).

This change is compatible with any Okteto backend. The `id` was always being returned although we were not referencing it here. We can go as back as 0.10 and see that the payload from deployPreview never changed.

Tested this change with the current okteto installation version and deploy/destroy works

pkg/okteto/preview.go
```diff
@@ -107,8 +107,8 @@ type previewEnv struct {
 }
 
 type deployPreviewResponse struct {
-	Action  actionStruct
-	Preview previewIDStruct
+	Id     graphql.String
```
ifbyol, line 110: I thinks this is not the same and it's a breaking change

We are changing this response:

```
{
    "preview": {
         "id": ...,
     },
     "action": {
           .....
       }
}
```

by this:

```
{
    "id": ....,
    "action": {
         .....
     }
}
```

maroshii: We are, but the return value from `deployPreview` has always been:

```
{
  "id": "my-id",
  "action": {
    ...
  },
  "preview": {
    "id": "my-id"
    ...
  }
}
```

So the top level `id` will exist in **any** okteto backend version

ifbyol: Oh, ok, yeah, I didn't check the preview response shape. It seems it's the same

ifbyol (approved):

jLopezbarb (approved):

PR #3570, opened 2023-04-21
feat: add okteto preview `sleep|wake` command

### LINKED ISSUE #2463: Support for `okteto preview sleep|wake`
I want to sleep a preview environment from the Okteto CLI. For example, I want to create a preview environment, run some tests, and sleep it to save resources in the cluster.

`okteto preview sleep` to mimic the function of the Okteto UI  **sleep** button.
`okteto preview wake` to mimic the function of the Okteto UI  **wake** button.

rakshitgondwal:
# Proposed changes

Fixes #2463 

- Added `okteto preview sleep` command
- Added `okteto preview wake` command


cmd/preview/sleep.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/okteto/okteto/pkg/okteto"
+	"github.com/spf13/cobra"
+)
+
+// Sleep sleeps a preview
+func Sleep(ctx context.Context) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "sleep <name>",
+		Short: "Sleeps a preview",
+		Args:  utils.ExactArgsAccepted(1, ""),
```
NitishKumar06, line 33:
```suggestion
		Args:  utils.ExactArgsAccepted(1, ""),
```
We may need to set the preview as default if no argument is passed.

cmd/preview/wake.go
```diff
@@ -0,0 +1,67 @@
...
+	oktetoLog "github.com/okteto/okteto/pkg/log"
+	"github.com/okteto/okteto/pkg/okteto"
+	"github.com/spf13/cobra"
+)
+
+func Wake(ctx context.Context) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "wake <name>",
+		Short: "Wakes a preview",
+		Args:  utils.ExactArgsAccepted(1, ""),
```
NitishKumar06, line 32: same suggestion for `wake` as `sleep`

NitishKumar06 (commented):
Added some suggestions.

cmd/preview/wake.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+func (pr *Command) ExecuteWakePreview(ctx context.Context, preview string) error {
+	// Spinner to be loaded before waking a preview
+	oktetoLog.Spinner(fmt.Sprintf("Waking %s preview", preview))
+	oktetoLog.StartSpinner()
+	defer oktetoLog.StopSpinner()
+
+	// trigger preview to wake
+	if err := pr.okClient.Previews().WakePreview(ctx, preview); err != nil {
```
jLopezbarb, line 61: The endpoint is the same as Namespace so it would be:
```suggestion
	if err := pr.okClient.Namespaces().Wake(ctx, preview); err != nil {
```

cmd/preview/sleep.go
rakshitgondwal: Hey @NitishKumar06! The other commands such as `okteto destroy` is also following this practice of accepting one argument atleast and almost, that's why I did the same for `preview` and `sleep`. If we are to switch to a default preview if no arg is provided, then maybe we should implement this logic for other commands as well. What do you think?

cmd/preview/wake.go
rakshitgondwal: Hey @jLopezbarb! Sorry I didn't understand, why we need a `Wake` endpoint over here. The wake endpoint requires a `namespaceClient` whereas for the preview command, don't we need to wake a `previewClient`?
If we want to use a `Wake` then are we trying to wake a namespace over here?
Sorry, I am new to this project and just have some doubts.

rakshitgondwal: Thank you for clearing up my doubt!

rakshitgondwal:
Hey @jLopezbarb @pchico83! Sorry for the tag but can I get a review on this PR, please? :smiley:

cmd/preview/sleep.go
```diff
@@ -0,0 +1,67 @@
...
+	oktetoLog "github.com/okteto/okteto/pkg/log"
+	"github.com/okteto/okteto/pkg/okteto"
+	"github.com/spf13/cobra"
+)
+
+// Sleep sleeps a preview
+func Sleep(ctx context.Context) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "sleep <name>",
+		Short: "Sleeps a preview",
```
pchico83, line 32:
```suggestion
		Short: "Sleeps a preview environment",
```

cmd/preview/sleep.go
pchico83: @NitishKumar06 there is no concept of default preview environment. `<name>` is mandatory in this case

cmd/preview/sleep.go
```diff
@@ -0,0 +1,67 @@
...
+			err = prCmd.ExecuteSleepPreview(ctx, prToSleep)
+			return err
+		},
+	}
+	return cmd
+}
+
+func (pr *Command) ExecuteSleepPreview(ctx context.Context, preview string) error {
+	// Spinner to be loaded sleeping a preview
+	oktetoLog.Spinner(fmt.Sprintf("Sleeping %s preview", preview))
```
pchico83, line 57:
```suggestion
	oktetoLog.Spinner(fmt.Sprintf("Sleeping preview environment '%s'...", preview))
```

cmd/preview/sleep.go
```diff
@@ -0,0 +1,67 @@
...
+	oktetoLog.StartSpinner()
+	defer oktetoLog.StopSpinner()
+
+	if err := pr.okClient.Namespaces().Sleep(ctx, preview); err != nil {
+		return fmt.Errorf("%w: %v", errFailedSleepPreview, err)
+	}
+
+	oktetoLog.Success("Preview '%s' is sleeping", preview)
```
pchico83, line 65:
```suggestion
	oktetoLog.Success("Preview environment '%s' is sleeping", preview)
```

cmd/preview/wake.go
```diff
@@ -0,0 +1,67 @@
...
+	oktetoErrors "github.com/okteto/okteto/pkg/errors"
+	oktetoLog "github.com/okteto/okteto/pkg/log"
+	"github.com/okteto/okteto/pkg/okteto"
+	"github.com/spf13/cobra"
+)
+
+func Wake(ctx context.Context) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "wake <name>",
+		Short: "Wakes a preview",
```
pchico83, line 31:
```suggestion
		Short: "Wakes a preview environment",
```

cmd/preview/wake.go
```diff
@@ -0,0 +1,67 @@
...
+			err = prCmd.ExecuteWakePreview(ctx, prToWake)
+			return err
+		},
+	}
+	return cmd
+}
+
+func (pr *Command) ExecuteWakePreview(ctx context.Context, preview string) error {
+	// Spinner to be loaded before waking a preview
+	oktetoLog.Spinner(fmt.Sprintf("Waking %s preview", preview))
```
pchico83, line 56:
```suggestion
	oktetoLog.Spinner(fmt.Sprintf("Waking preview environment '%s'...", preview))
```

cmd/preview/wake.go
```diff
@@ -0,0 +1,67 @@
...
+	oktetoLog.StartSpinner()
+	defer oktetoLog.StopSpinner()
+
+	// trigger preview to wake
+	if err := pr.okClient.Namespaces().Wake(ctx, preview); err != nil {
+		return fmt.Errorf("%w: %v", errFailedWakePreview, err)
+	}
+
+	oktetoLog.Success("Preview '%s' is awake now", preview)
```
pchico83, line 65:
```suggestion
	oktetoLog.Success("Preview environment '%s' is awake", preview)
```

pchico83 (approved):

cmd/preview/sleep.go
NitishKumar06: got it @pchico83 ! You are correct @rakshitgondwal but it depends on what is the end goal of the command.

For eg. We can pass additional arguments to `sleep` and `wake` command if we don't want to go with the default. However, this may/may not be a perfect fit for a command such as `delete`in which if we don't pass an argument, the command deletes the default namespace. Your thoughts about it @pchico83 ?

pchico83: `okteto ns delete` could default to the current namespace, but `okteto preview destroy` should have the name as argument. Both should ask for confirmation. @rakshitgondwal do you want to create another PR to cover those cases?

NitishKumar06: @pchico83 Here are the changes for `delete` I've implemented:

Issue: #3597
PR: https://github.com/okteto/okteto/pull/3598

rakshitgondwal:
Hey @pchico83 @jLopezbarb!
When are we planning to release this feature?

cmd/preview/sleep.go
```diff
@@ -0,0 +1,67 @@
...
+
+	contextCMD "github.com/okteto/okteto/cmd/context"
+	"github.com/okteto/okteto/cmd/utils"
+	oktetoErrors "github.com/okteto/okteto/pkg/errors"
+	oktetoLog "github.com/okteto/okteto/pkg/log"
+	"github.com/okteto/okteto/pkg/okteto"
+	"github.com/spf13/cobra"
+)
+
+// Sleep sleeps a preview
```
AdrianPedriza, line 28:
```suggestion
// Sleep sleeps a preview environment
```

internal/test/client/preview.go
```diff
@@ -69,3 +71,11 @@ func (c *FakePreviewsClient) Destroy(_ context.Context, _ string) error {
...
 	return nil, nil
 }
+
+func (c *FakePreviewsClient) SleepPreview(_ context.Context, _ string) error {
+	return c.response.ErrSleepPreview
+}
+
+func (c *FakePreviewsClient) WakePreview(_ context.Context, _ string) error {
+	return c.response.ErrWakePreview
+}
```
AdrianPedriza, line 81:
```suggestion

```
do we really need these lines in this PR? I cannot see any interface modification adding this new behavior so test shouldn't need this.

AdrianPedriza:
after all, we are sleeping/awakening a namespace. Could we use the existing logic in the `cmd/namespace` pkg for sleep and wake a preview? Seems like we are duplicating code. So the e2e from https://github.com/okteto/okteto/pull/3605 would be sufficient for the sleep/wake preview command.

rakshitgondwal:
> after all, we are sleeping/awakening a namespace. Could we use the existing logic in the `cmd/namespace` pkg for sleep and wake a preview? Seems like we are duplicating code. So the e2e from #3605 would be sufficient for the sleep/wake preview command.

Oh, so should I just write test cases for `preview sleep|wake` with the logic used from the `namespace` package right?

AdrianPedriza:
> > after all, we are sleeping/awakening a namespace. Could we use the existing logic in the `cmd/namespace` pkg for sleep and wake a preview? Seems like we are duplicating code. So the e2e from #3605 would be sufficient for the sleep/wake preview command.
> 
> Oh, so I should just write test cases for `preview sleep|wake` with the logic used from the `namespace` package right?

I think sleep/wake namespace logic is same (https://github.com/okteto/okteto/blob/master/cmd/namespace/sleep.go#LL59C11-L59C11). Could we use it? If so, we already have an integration covering it :)

rakshitgondwal:
Please check now.
I moved the logic of creating a `NewFakeNamespaceCommand` from `delete_test.go` to `namespace.go` so that it could be used in other packages too. I was not being able to use that function in `preview` package while it was in `delete_test.go`.
If it needs to be moved into some other place, do let me know.

cmd/preview/sleep.go
```diff
@@ -0,0 +1,68 @@
...
+	defer oktetoLog.StopSpinner()
+
+	// trigger preview environment to sleep
+	if err := pr.okClient.Namespaces().Sleep(ctx, preview); err != nil {
+		return fmt.Errorf("%w: %v", errFailedSleepPreview, err)
+	}
+
+	oktetoLog.Success("Preview environment '%s' is sleeping", preview)
+	return nil
+}
```
AdrianPedriza, line 68:
```suggestion
			nsCmd, err := namespace.NewCommand()
			if err != nil {
				return err
			}
			err = nsCmd.ExecuteSleepNamespace(ctx, prToSleep)
			return err
		},
	}
	return cmd
}
```

Same case as above. We can use `ExecuteSleepNamespace` func from namespace package instead of duplicate it.

cmd/preview/wake_test.go
AdrianPedriza, line 1: With my suggestions this test is already covered by `namespace` package

rakshitgondwal:
Just one question, do you really think it would be a good idea to use the functions from `namespace` package, as if we are doing so, we are showing the user an error `ErrFailedWakeNamespace`/`ErrFailedSleepNamespace`  or promoting the user that `namespace xyz is sleeping/awake now` when he/she is trying to `sleep/wake` an xyz preview.

TLDR; would it be a good idea to mention `namespace` in the error message or the success message in case of `preview` too?

AdrianPedriza:
> Just one question @AdrianPedriza, do you really think it would be a good idea to use the functions from `namespace` package as if we are doing so, we are showing the user an error `ErrFailedWakeNamespace`/`ErrFailedSleepNamespace` or promoting the user that `namespace xyz is sleeping/awake now` when he/she is trying to `sleep/wake` an xyz preview.
> 
> TLDR; would it be a good idea to mention `namespace` in the error message or the success message in the case of `preview` too? Screenshot from 2023-05-25 19-37-21

oh I see :/ . Then let's come back to your previous approach because in the end I think functions is not exactly the same. So sorry for the noise! :(

rakshitgondwal:
Hey @AdrianPedriza!
Please review.

jLopezbarb (approved):

AdrianPedriza (approved):
Great work! 👍

rakshitgondwal:
> Great work! +1

Thank you!

rakshitgondwal:
Hey, when are we planing to release this?

cmd/namespace/namespace.go
```diff
@@ -64,3 +71,34 @@ func Namespace(ctx context.Context) *cobra.Command {
 	cmd.AddCommand(Wake(ctx))
 	return cmd
 }
+
+type fakeK8sProvider struct {
```
teresaromero, line 75: can we move this to the test file where they are used?

teresaromero (changes requested):
@rakshitgondwal  Hi! I've been having a look to it a it looks ready to merge but i first please change the fake helpers so they are under a test file instead of `namespace.go` file. 

Thanks!

cmd/namespace/namespace.go
jLopezbarb: +1 @rakshitgondwal let us know if you can move the `fakeK8sProvider` to the test file, if not we will move it and merge the PR

rakshitgondwal: Oh yes, so sorry, just busy with other stuff, will fix this in 1-2 days.

teresaromero (approved):

teresaromero:
Update the branch removing the preview tests as they where duplicated from namespace tests

AdrianPedriza (approved):
LGTM

jLopezbarb (approved):

teresaromero:
Added docs PR https://github.com/okteto/docs/pull/444

rakshitgondwal:
Thank you for the merge. Also, sorry for being inactive, got caught up witg something else.

jLopezbarb:
> Thank you for the merge. Also, sorry for being inactive, got caught up with something else.

don't worry, you don't have to apologise. We know people have other things to attend to. Thanks for your contribution!!

Issue #3444, opened 2023-03-14
Hybrid development: Command runner


jLopezbarb:
## Parent epic/feature

* Feature: [Okteto CLI 3.0](https://www.notion.so/okteto/Okteto-CLI-3-0-7fbd3a6a566145d3a4930d3c15fc642d?pvs=4)

### Depends on:

* #3425  

## Description

We want our users to be able to develop in the most comfortable way possible. That is why we are going to add the option to develop following a hybrid approach. This means that instead of synchronizing the code and developing with the code living in the cluster, the code lives in local and through a reverse the connections are made.

This issue aims to run the service that the user is trying to run in hybrid mode locally, creating the connections with the clusters specified by reverse/forward ports:
- The command must be run locally
- The command must have the `workdir` specified. If not specified the project `workdir` will be the one used
- The environment variables must be the same ones as if the command was run in the cluster. (Defined in the environment variables) 
- Anew field is added to the `up` metrics to specify which mode the user is using
- Only one `okteto up` can be running at the same time for the same service
- We should add env vars from:
  - main image. Inspect the image to retrieve the env vars
  - config map. 
  - okteto secrets
 

### Tips:
- `okteto up` command execution should be working as it was previously. Creating an interface that runs the command could be a good option 

## Acceptance Criteria

- [ ] `okteto up` continues working with a manifest in `sync` mode
- [ ] `okteto up` in `hybrid` mode runs locally with the env vars defined
- [ ] `okteto down` restores the original status of the application in `sync` mode
- [ ] `okteto down` restores the original status of the application in `hybrid` mode
- [ ] Add `okteto up` in `hybrid` e2e test

## Test strategy and scenarios

All the tests scenarios will use the yaml used below or any example but keep in mind to use the `sync` mode:
```yaml
dev:
  frontend:
    mode: hybrid
    selector:
      app.kubernetes.io/part-of: okteto
      app.kubernetes.io/component: frontend
    command: ["sh", "-c", "echo KEY=$KEY && yarn start"]
    environment: 
      KEY: VALUE
    reverse:
      - 8080:8080
  api:
    mode: sync
    selector:
      app.kubernetes.io/part-of: okteto
      app.kubernetes.io/component: api
    image: okteto/golang:1
    environment:
      - LOG_FORMATTER=text
    command: sh
    sync:
      - ./api:/usr/src/app
    forward:
      - 2345:2345
  producer:
    selector:
      app.kubernetes.io/part-of: okteto
[...]

PR #3571, opened 2023-04-21
support hybrid development

### LINKED ISSUE #3425: Hybrid development: add mode field dev section
## Parent epic/feature

* Feature: [Okteto CLI 3.0](https://www.notion.so/okteto/Okteto-CLI-3-0-7fbd3a6a566145d3a4930d3c15fc642d?pvs=4)

## Description

We want our users to be able to develop in the most comfortable way possible. That is why we are going to add the option to develop following a hybrid approach. This means that instead of synchronizing the code and developing with the code living in the cluster, the code lives in local and through a reverse the connections are made.

This issue will try to add this new mode in a way that does not modify any of the functionality we have for the dev section. To do so, we will modify the way of making unmarshal of each of the dev by checking the value of the new mode field (sync, hybrid).

```yaml
dev:
  frontend:
    selector:
      app.kubernetes.io/part-of: okteto
      app.kubernetes.io/component: frontend
    command: ["sh", "-c", "yarn start"]
    reverse:
      - 8080:8080
  api:
    mode: sync
    selector:
      app.kubernetes.io/part-of: okteto
      app.kubernetes.io/component: api
    image: okteto/golang:1
    environment:
      - LOG_FORMATTER=text
    command: sh
    sync:
      - ./api:/usr/src/app
    forward:
      - 2345:2345
  producer:
    selector:
      app.kubernetes.io/part-of: okteto
      app.kubernetes.io/component: producer
    image: okteto/golang:1
    command: sh
    sync:
      - ./producer:/usr/src/app
    forward:
      - 2345:2345
```

The only fields supported in the hybrid mode are:
- `reverse`
- `command`
- `environment`
- `forward`
- `selector`
- `workdir`

After serializing we need to convert the hybrid dev into a normal dev. This will allow us to continue using the same translation rules that are used in `sync` mode:
- Using a small image like busybox 
- Use an infinite command(`while true ; do continue ; done`)
- Use an empty sync folder.

### Tips:
- We could parse the dev section in two steps. The first one we unmarshal the `mode` and then 

## Acceptance Criteria

- [ ] `okteto up` continues working with a manifest without mode specified
- [ ] `okteto up` continues working if mode is `sync`
- [ ] `okteto up` tracks a new property that indicates the mode the user is developing  
- [ ] `okteto up` creates a new app that is running the small image with the infinite command

## Test strategy and scenarios

All the tests scenarios will use the yaml used below:
```yaml
dev:
  frontend:
    mode: hybrid
    selector:
      app.kubernetes.io/part-of: okteto
      app.kubernetes.io/component: frontend
    command: ["sh", "-c", "yarn start"]
    reverse:
      - 8080:8080
  api:
    mode: sync
    selector:
      app.kubernetes.io/part-of: okteto
      app.kubernetes.io/component: api
    image: okteto/golang:1
    environment:
      - LOG_FORMATTER=text
    command: sh
    sync:
      - ./api:/usr/src/app
    forward:
      - 2345:2345
  producer:
    selector:
      app.kubernetes.io/part-of: okteto
      app.kubernetes.io/component: producer
    image: okteto/golang:1
    command: sh
    sync:
      - ./producer:/usr/src/app
    forward:
      - 2345:2345
```

#### Scenario 1
1. Copy the okteto manifest above
2. Create a folder named `producer` on the path
3. Run `okteto up producer`
4. It should work as before

#### Scenario 2
1. Copy the okteto manifest above
2. Create a folder `api` on the path
3. Run `okteto up api`
4. It should work as before

AdrianPedriza:
# Proposed changes
- add new field `mode` for `dev` section
- parse more restrictive mode (`hybrid`). If err when unmarshaling then fallback to sync mode unmarshal.
- add test
- sync dummy folder if `hybrid` enbled
- use `busybox` image for dev container if `hybrid` enabled
- split execution behavior for hybrid development vs sync development. 
- add e2e covering use of available envs in local
- exec `okteto exec` against local
- replicate envs in remote to local command

Fixes #3425 #3444


pkg/model/serializer.go
```diff
@@ -724,10 +724,27 @@ func checkFileAndNotDirectory(path string) error {
 func (d *Dev) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	type devType Dev // Prevent recursion
 	dev := devType(*d)
-	err := unmarshal(&dev)
+	type hybridModeInfo struct {
```
AdrianPedriza, line 727: we could make this struct public and use in inside `Dev` struct to avoid duplicate fields (in `Dev` and in `hybridModeInfo`). 
````go
type HybridModeInfo struct {
  Workdir string `json:"workdir,omitempty" yaml:"workdir,omitempty"`
  Selector Selector `json:"selector,omitempty" yaml:"selector,omitempty"`
  . . .
}

type Dev struct {
. . .
HybridModeInfo
}
````
that means modify the use of the all fields related to HybridMode. (i.e Dev{HybridModeInfo.Selector: ...}. WDYT? @jLopezbarb

File: pkg/model/serializer.go [hunk 2]
```diff
@@ -724,10 +724,27 @@ func checkFileAndNotDirectory(path string) error {
...
+		Reverse     []Reverse         `json:"reverse,omitempty" yaml:"reverse,omitempty"`
+		Mode        string            `json:"mode,omitempty" yaml:"mode,omitempty"`
+	}
+	hybridModeDev := &hybridModeInfo{}
+	err := unmarshal(hybridModeDev)
 	if err != nil {
-		return fmt.Errorf("Unmarshal error: '%s'", err)
+		if hybridModeDev.Mode == "hybrid" {
+			return err
+		}
```
AgustinRamiroDiaz, line 741: Is it guaranteed that this will get filled? 

If not, I think we might need to do this in 2 steps, first unmarshall into a struct with only `Mode` and check there to see how to follow


---

Also, we could wrap the error here to notify the user that the hybrid mode has fewer fields

AgustinRamiroDiaz:
@AdrianPedriza by just adding the unmarshalling this works?

pkg/model/serializer.go
AdrianPedriza: following spec, if hybrid mode is not filled we assume that we are working with the classic mode: sync

AdrianPedriza:
> @AdrianPedriza by just adding the unmarshalling this works?

this PR is just to be able to read and set dev fields according dev mode

pkg/model/serializer.go
AgustinRamiroDiaz: @AdrianPedriza agreed. I'm talking about the unmarshalling. In this block we are checking `hybridModeDev.Mode` but the marshalling has failed. Are we guaranteed to have it filled when `mode` is set?

AgustinRamiroDiaz:
@AdrianPedriza so then it's not a full fix of https://github.com/okteto/okteto/issues/3425?

pkg/model/serializer.go
AdrianPedriza: if the unmarshal fail because the dev is not hybrid or because it contains field that doesn't belong to hybrid mode I'm still able to check this value if user filled it in. So let's think on an scenario where user use this dev:
``````yaml

dev:
  frontend:
    mode: hybrid
    selector:
      app.kubernetes.io/part-of: okteto
      app.kubernetes.io/component: frontend
    command: ["sh", "-c", "yarn start"]
    reverse:
      - 8080:8080
    fordward:
      - 8090:8090
``````
(this will fail because `fordward` field is not supported for hybrid mode)
first of all, we will try to unmarshal it using hybrid field. It will fail but `mode` value will be filled in the struct because that value is present. If there is an error and the value is hybrid (this case) there is nothing more we can do, we return err. On the other hand, if there is an error trying to unmarshal it to hybrid mode and there is no indication that mode is hybrid then we fall back to the classic unmarshal. Is that what you are asking?

File: pkg/model/serializer.go (hunk 1 above)
Comment (jLopezbarb) [reply pkg/model/serializer.go:727]: I'd have tehm separate. They can evolve different and it doesn't make sense that the hybrid mode is inside the `Dev` since it's not a composition.

**Comment by jLopezbarb** [comment]
@AdrianPedriza I think we are missing this part from the spec:

```
After serializing we need to convert the hybrid dev into a normal dev. This will allow us to continue using the same translation rules that are used in sync mode:

Using a small image like busybox
Use an infinite command(while true ; do continue ; done)
Use an empty sync folder.
```

File: pkg/model/dev.go [hunk 3]
```diff
@@ -94,6 +94,7 @@ type Dev struct {
 	EnvFiles             EnvFiles              `json:"envFiles,omitempty" yaml:"envFiles,omitempty"`
 	Environment          Environment           `json:"environment,omitempty" yaml:"environment,omitempty"`
 	Volumes              []Volume              `json:"volumes,omitempty" yaml:"volumes,omitempty"`
+	Mode                 string                `json:"mode,omitempty" yaml:"mode,omitempty"`
```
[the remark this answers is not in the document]
AgustinRamiroDiaz: nit: Also an enum could be an option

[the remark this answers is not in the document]
AdrianPedriza: for the time being and taking into account that only two modes are supported, I think best option is to check a boolean field. Probably something like `dev.isHybridDev()` is the best fit for now. I still need to refactor that

pkg/model/serializer.go
```diff
@@ -724,10 +724,32 @@ func checkFileAndNotDirectory(path string) error {
...
 	if err != nil {
-		return fmt.Errorf("Unmarshal error: '%s'", err)
+		return err
 	}
+
+	if dev.Mode == "hybrid" {
+		dev.Image.Name = "busybox"
+	}
+
 	*d = Dev(dev)
```
jLopezbarb, line 753: I can't find it where but If we didn't set the sync folders we were syncthing a path, Can you double check that we are syncthing only one folder and that folder is empty?

jLopezbarb (approved):

pkg/model/serializer.go
AdrianPedriza: here https://github.com/okteto/okteto/blob/adrian/add-mode-field-in-dev-section/cmd/up/up.go#L373 we creates a dummy folder to sync

cmd/up/exec.go
```diff
@@ -18,18 +18,64 @@ import (
...
+	runCommand(ctx context.Context, cmd []string) error
+}
+
+type hybridExecutor struct {
+	workdir string
+}
+
+func (he *hybridExecutor) runCommand(_ context.Context, cmd []string) error {
+	e := executor.NewExecutor(oktetoLog.GetOutputFormat(), false, he.workdir)
+	command := model.DeployCommand{
```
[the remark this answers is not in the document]
AdrianPedriza: done dd0018e9260f91c506dbec5176523ae56c2252c7

cmd/up/exec.go
```diff
@@ -18,18 +18,64 @@ import (
...
+		Name:    "hybrid development command",
+		Command: strings.Join(cmd, " "),
+	}
+	return e.Execute(command, os.Environ())
```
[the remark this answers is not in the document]
AdrianPedriza: done 8c5dab5a81937c992f8f778698ad1a7985226de3

pkg/okteto/secrets.go
```diff
@@ -139,6 +143,27 @@ func (c *userClient) GetContext(ctx context.Context, ns string) (*types.UserCont
 	return result, nil
 }
 
+// GetSecrets returns the secrets from Okteto API
+func (c *userClient) GetUserSecrets(ctx context.Context) ([]types.Secret, error) {
```
jLopezbarb, line 147: Can we add a unit test for this new function?

cmd/up/up.go
```diff
@@ -503,6 +508,24 @@ func loadManifestOverrides(dev *model.Dev, upOptions *UpOptions) error {
 	return nil
 }
 
+func setSyncDefaultsByDevMode(dev *model.Dev, fs afero.Fs) error {
```
jLopezbarb, line 511: Can we add a test for this new function?

cmd/up/stignore.go
```diff
@@ -122,18 +122,19 @@ func addSyncFieldHash(dev *model.Dev) error {
 }
 
 func checkStignoreConfiguration(dev *model.Dev) error {
+	isHybridModeEnabled := dev.IsHybridModeEnabled()
```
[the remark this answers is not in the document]
AdrianPedriza: done e7c65a4f8a8a0f9ca94aa3a121f9f8c0472fd6da

cmd/up/exec.go
```diff
@@ -16,27 +16,168 @@ package up
...
+type syncExecutor struct {
+	iface      string
+	remotePort int
+}
+
+func (se *syncExecutor) runCommand(ctx context.Context, cmd []string) error {
+	return ssh.Exec(ctx, se.iface, se.remotePort, true, os.Stdin, os.Stdout, os.Stderr, cmd)
+}
+
+func newExecutorByDevMode(ctx context.Context, up *upContext) (devExecutor, error) {
```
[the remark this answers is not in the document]
AdrianPedriza: done 169cd17fbdd495ab4b7494aa7f40732ba51a7aad

ifbyol:
@AdrianPedriza Is there a way to know if a development in dev mode is using hybrid development or a normal one?

cmd/up/up.go
AdrianPedriza: done b4f7ad31312400db69a6e935dfc5b831526e9069

pkg/okteto/secrets.go
AdrianPedriza: done c9e2b7deea5b348a5f9cddef298219dc33be523e

cmd/up/exec.go
```diff
@@ -16,27 +16,168 @@ package up
...
+			envs:    envs,
+		}, nil
+	}
+	return &syncExecutor{
+		iface:      up.Dev.Interface,
+		remotePort: up.Dev.RemotePort,
+	}, nil
+}
+
+func getEnvsFromContext(ctx context.Context, dev *model.Dev, c kubernetes.Interface, name string, namespace string) ([]string, error) {
```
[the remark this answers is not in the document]
AdrianPedriza: done 10d16993331524e0cf9bfeb928a00129e6cf16d8

pkg/model/serializer.go
AdrianPedriza: ok, tell me if this a9e3e62cd08289de9b575f09efa40e4f7b97d70f covers your concerns :)

go.mod
```diff
@@ -85,7 +85,7 @@ require (
 	github.com/containerd/containerd v1.5.18 // indirect
 	github.com/containerd/continuity v0.3.0 // indirect
 	github.com/containerd/typeurl v1.0.2 // indirect
-	github.com/creack/pty v1.1.17 // indirect
+	github.com/creack/pty v1.1.18
```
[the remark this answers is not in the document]
jLopezbarb: Another thing to have in mind.

How does this work with interactive commands like `sh`? I've seen that we are not piping the stdin so I think the users won't be able to write if they set a command like that. Can you confirm my concerns?

cmd/up/exec.go
```diff
@@ -16,27 +16,268 @@ package up
...
+		return nil, err
+	}
+	envs = append(envs, secretsEnvs...)
+
+	app, err := apps.Get(ctx, eg.dev, eg.namespace, eg.client)
+	if err != nil {
+		return nil, err
+	}
+
+	imageEnvs, err := eg.imageEnvsGetter.getEnvsFromImage(app.PodSpec().Containers[0].Image)
```
jLopezbarb, line 195: We shouldn't be checking `app.PodSpec().Containers[0].Image`. We can use the function `GetDevContainer` to get the container with the container name.

cmd/up/exec.go
```diff
@@ -16,27 +16,268 @@ package up
...
+		return nil, err
+	}
+
+	envs, err := envsGetter.getEnvs(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	return &hybridExecutor{
+		workdir: filepath.Join(wd, up.Dev.Workdir),
```
jLopezbarb, line 99: Q: Are we checking that the workdir is a relative path? I didn't see it neither in the serializer nor the `dev.go` file. What if the user set it as an absolute path? I think this scenario could be broken

cmd/up/exec.go
```diff
@@ -16,27 +16,268 @@ package up
...
+	runCommand(ctx context.Context, cmd []string) error
+}
+
+type hybridExecutor struct {
+	workdir string
+	envs    []string
+}
+
+func (he *hybridExecutor) runCommand(_ context.Context, cmd []string) error {
+	c := exec.Command("bash", "-c", strings.Join(cmd, " "))
```
jLopezbarb, line 57: I think the bash here could break other scenarios (Windows, docker images with alpine...) Should we just run it without the `bash -c`?

cmd/up/exec.go
```diff
@@ -16,27 +16,268 @@ package up
...
+	c.Dir = he.workdir
+	c.Env = append(os.Environ(), he.envs...)
```
jLopezbarb, line 59: We shouldn't be adding the os.Environ variables. We need to be as close as it would be running on the cloud. I'd add only needed environment variables such as `PATH` but definetely not adding all of them since we are retrieving them from the cluster

AdrianPedriza: done 9c398bf87ad1cf49cfdde25f350c7da00c2b2671

cmd/up/exec.go
AdrianPedriza: I followed what we have now where bash is always used unless remote deploys. https://github.com/okteto/okteto/blob/master/cmd/utils/executor/executor.go#L61. It is OK?

jLopezbarb: On the deploy Windows users can omit the bash/sh by setting the flag `--no-bash`. Otherwise we were redirecting them to WSL. Here is where we do that:  https://github.com/okteto/okteto/blob/master/cmd/utils/executor/executor.go#L78

cmd/up/exec.go
```diff
@@ -16,27 +16,268 @@ package up
...
+func (cmg *configMapGetter) getEnvsFromConfigMap(ctx context.Context, name string, namespace string, client kubernetes.Interface) ([]string, error) {
+	var envs []string
+
+	devCmap, err := configmaps.Get(ctx, pipeline.TranslatePipelineName(name), namespace, client)
+	if err != nil && !oktetoErrors.IsNotFound(err) {
+		return nil, err
+	}
+
+	if devCmap != nil && devCmap.Data != nil {
+		if devVariables, ok := devCmap.Data[constants.OktetoConfigMapVariablesField]; ok {
```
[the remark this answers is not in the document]
AdrianPedriza: I think it is more cumbersome to make a function for it than to duplicate two lines in this case

cmd/up/exec.go
AdrianPedriza: done dd202d86cf1ce42babc90f207f706c127e0a4170

cmd/up/exec.go
AdrianPedriza: done fbfde67a4651acf3c7676ff3f1b71be1788d94d1

go.mod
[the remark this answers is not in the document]
AdrianPedriza: done 3d7fd83e441bcd3ea1e650dcca4243da5e4290e3 I added a snippet from creack/pty repo. It is OK?

cmd/up/exec.go
AdrianPedriza: done 38016f667ab2144ab79b7fe6e4faf031d85e687f

AdrianPedriza:
> @AdrianPedriza Is there a way to know if a development in dev mode is using hybrid development or a normal one?

not at the moment. Is there any case where we might need it? If so I can add it as a follow up or address it in this issue, depending on the criticality.

cmd/up/exec.go
```diff
@@ -16,27 +16,286 @@ package up
...
+		return nil, err
+	}
+
+	imageEnvs, err := eg.imageEnvsGetter.getEnvsFromImage(apps.GetDevContainer(app.PodSpec(), "").Image)
+	if err != nil {
+		return nil, err
+	}
+	envs = append(envs, imageEnvs...)
+
+	for _, env := range app.PodSpec().Containers[0].Env {
```
jLopezbarb, line 214:
```suggestion
	for _, env := range apps.GetDevContainer(app.PodSpec(), "").Env {
```

cmd/up/exec.go
```diff
@@ -16,27 +16,286 @@ package up
...
+		envs = append(envs, fmt.Sprintf("%s=%s", env.Name, env.Value))
+	}
+
+	pathValue := os.Getenv("PATH")
```
jLopezbarb, line 218: Should we create another func for this so we can mock it in the test? WDYT

cmd/up/exec.go
AdrianPedriza: done a46c142a76338abe33f016b004c4533e246f2b35

ifbyol:
> > @AdrianPedriza Is there a way to know if a development in dev mode is using hybrid development or a normal one?
> 
> not at the moment. Is there any case where we might need it? If so I can add it as a follow up or address it in this issue, depending on the criticality.

@AdrianPedriza  Yes, the backend might need it to execute different logic to consider a session active or not to know if a namespace has to be scale to zero or not

AdrianPedriza:
> > > @AdrianPedriza Is there a way to know if a development in dev mode is using hybrid development or a normal one?
> > 
> > 
> > not at the moment. Is there any case where we might need it? If so I can add it as a follow up or address it in this issue, depending on the criticality.
> 
> @AdrianPedriza Yes, the backend might need it to execute different logic to consider a session active or not to know if a namespace has to be scale to zero or not

added as a follow up issue your suggestion  it was not initially contemplated among the acceptance criteria. -> https://github.com/okteto/okteto/issues/3653

Issue #3572, opened 2023-04-21
Wrong spinner on `okteto deploy --remote`


pchico83:
**Describe the bug**
Run `okteto deploy --remote` and the spinner says "Destroying your development environment..."

**Expected behavior**
- The spinner should say "Deploying your development environment..." if I run `okteto deploy --remote`.
- If I run `okteto destroy --remote`, the spinner should say "Destroying your development environment..."

NitishKumar06:
This [PR](https://github.com/okteto/okteto/pull/3573) solves this issue.

pchico83:
> This [PR](https://github.com/okteto/okteto/pull/3573) solves this issue.

Wow, that was fast @NitishKumar06 !

NitishKumar06:
Thanks @pchico83! We are ready to close this issue now.

AdrianPedriza:
Description updated. Now instead of display always `Deploying..` we are displaying `Destroying..`

NitishKumar06:
I tried running this locally, and here's what I got:
![image](https://user-images.githubusercontent.com/86051118/234349480-b47de281-9dd0-469a-bfe6-6e1a221e9174.png)

NitishKumar06:
I think the spinner is correct for `okteto deploy --remote`. I tried running the command `okteto destroy --remote` which showed spinner text as _Deploying your development environment_ instead of _Destroying your development environment_.

The spinner was way fast to analyse the output. Hence i screen recorded to see what text was showing up.

Hence, I think the issue description raised by @pchico83 was correct rather than the current one. 
![image](https://user-images.githubusercontent.com/86051118/234350710-bf1fbd23-499c-4831-91c9-ad9b83248234.png)

jLopezbarb:
which Okteto CLI version are you using? It's only happening with the master build of the CLI. To reproduce it, run make build on the CLI project and then use that binary (probably ok alias if you have it pointing to the binary)

AdrianPedriza:
[...]

PR #3573, opened 2023-04-22
Modified the spinner text to mention about destroying dev env

NitishKumar06:
Fixes:#3572

### DIFF
--- a/pkg/cmd/build/logger.go
+++ b/pkg/cmd/build/logger.go
@@ -145,7 +145,7 @@ func (t *trace) display() {
 			}
 		}
 		if t.hasCommandLogs(v) {
-			oktetoLog.Spinner("Deploying your development environment...")
+			oktetoLog.Spinner("Destroying your development environment...")
 			for _, log := range v.logs {
 				var text oktetoLog.JSONLogFormat
 				if err := json.Unmarshal([]byte(log), &text); err != nil {


pchico83 (approved):
Thanks @NitishKumar06 !

AdrianPedriza:
I believe that this PR is not correct. The problem is that we were not distinguishing between deploy and destroy in the message when displaying the spinner. If previously we always showed the `Deploying` message... now we only show the `Destroying..` message. You can check it running `okteto deploy --remote`. I will reopen the issue again

pchico83:
> I believe that this PR is not correct. The problem is that we were not distinguishing between deploy and destroy in the message when displaying the spinner. If previously we always showed the `Deploying` message... now we only show the `Destroying..` message. You can check it running `okteto deploy --remote`. I will reopen the issue again

@NitishKumar06 my bad, we should show "Deploying..." on `okteto deploy` and "Dstroying..." on `okteto destroy`

jLopezbarb:
> > @NitishKumar06 my bad, we should show "Deploying..." on `okteto deploy` and "Dstroying..." on `okteto destroy`
> 
> I'm sorry for not pushing the right changes. I'd like to take this issue up and fix it.

No problem @NitishKumar06, it was hard to notice that both command where using the same logger (probably it's bad design in our end 🤦‍♂️ ). The simplest solution would be to pass to the logger if it's a deploy or a destroy. We can add it [here](https://github.com/okteto/okteto/blob/master/cmd/build/remote/build.go#L39) and then use that value [here](https://github.com/okteto/okteto/blob/master/pkg/cmd/build/buildkit.go#L283). Let me know if you have any further help and we will help you 😄

NitishKumar06:
> > > @NitishKumar06 my bad, we should show "Deploying..." on `okteto deploy` and "Dstroying..." on `okteto destroy`
> > 
> > 
> > I'm sorry for not pushing the right changes. I'd like to take this issue up and fix it.
> 
> No problem @NitishKumar06, it was hard to notice that both command where using the same logger (probably it's bad design in our end man_facepalming ). The simplest solution would be to pass to the logger if it's a deploy or a destroy. We can add it [here](https://github.com/okteto/okteto/blob/master/cmd/build/remote/build.go#L39) and then use that value [here](https://github.com/okteto/okteto/blob/master/pkg/cmd/build/buildkit.go#L283). Let me know if you have any further help and we will help you smile

- Could you please provide some context how [this function](https://github.com/okteto/okteto/blob/master/cmd/build/remote/build.go#L39) is related to logger? 

- By passing the value [here](https://github.com/okteto/okteto/blob/master/pkg/cmd/build/buildkit.go#L283), do we intend to add a new case?

jLopezbarb:
> * Could you please provide some context how [this function](https://github.com/okteto/okteto/blob/master/cmd/build/remote/build.go#L39) is related to logger?

We run the `deploy/destroy` in a buildkit and this is the struct constructor for all the remote deploy/destroy, but we don't know if it's a deploy or destroy. I found another way of doing it with a better approach. I'll explain it in the next question

> * By passing the value [here](https://github.com/okteto/okteto/blob/master/pkg/cmd/build/buildkit.go#L283), do we intend to add a new case?

Yes we should have two different cases (`deploy` and `destroy`). These values comes from  [this line](https://github.com/okteto/okteto/blob/master/cmd/deploy/remote.go#L151) for deploy and [this other line](https://github.com/okteto/okteto/blob/master/cmd/destroy/remote.go#L149) for destroy. We can change the deploy for destroy in the destroy file, add a new case in the switch(`case "deploy", "destroy":`) and use that for the deploying destroying

NitishKumar06:
> Yes we should have two different cases (`deploy` and `destroy`). These values comes from [this line](https://github.com/okteto/okteto/blob/master/cmd/deploy/remote.go#L151) for deploy and [this other line](https://github.com/okteto/okteto/blob/master/cmd/destroy/remote.go#L149) for destroy. We can change the deploy for destroy in the destroy file, add a new case in the switch(`case "deploy", "destroy":`) and use that for the deploying destroying

Thanks for the help. Here is the new PR: https://github.com/okteto/okteto/pull/3584
cc @pchico83 @AdrianPedriza @jLopezbarb

PR #3479, opened 2023-03-26
Added new commands for okteto namespace - sleep and wake

NitishKumar06:
Fixes: #3470

**Note:** This is a WIP PR - the changes applied are running perfectly. However, I believe that code can be improved considering the file _delete.go_. The test cases need to be added as well. I'm raising this PR before hand to make sure we don't keep bunch of suggestions at the end to be implemented all at once rather it's better if suggestions are implemented side by side.

PR #3574, opened 2023-04-24
Set the sleep and wake command to perform actions on current  namespace

NitishKumar06:
# Proposed changes

This PR adds a little functionality to the new commands `sleep` and `wake` that have been added to Oketo CLI after [this](https://github.com/okteto/okteto/pull/3479) PR being merged. Initially, `sleep` and  `wake` commands required an argument to work. However, this is not how we want the user to use it.

With this PR being merged, the user can use `sleep` and `wake` commands without passing any argument. **If no argument is passed, the command will be executed on the current namespace.**

For more reference, refer to [this](https://github.com/okteto/docs/pull/366#issuecomment-1518517252) conversation.

### DIFF
--- a/cmd/namespace/sleep.go
+++ b/cmd/namespace/sleep.go
@@ -29,17 +29,18 @@ func Sleep(ctx context.Context) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "sleep <name>",
 		Short: "Sleeps a namespace",
-		Args:  utils.ExactArgsAccepted(1, ""),
+		Args:  utils.MaximumNArgsAccepted(1, ""),
 		RunE: func(cmd *cobra.Command, args []string) error {
-			nsToSleep := args[0]
-			if err := contextCMD.NewContextCommand().Run(ctx, &contextCMD.ContextOptions{Namespace: nsToSleep, Show: true}); err != nil {
-				return err
-			}
-
 			if !okteto.IsOkteto() {
 				return oktetoErrors.ErrContextIsNotOktetoCluster
 			}
-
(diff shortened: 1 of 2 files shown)


AdrianPedriza (approved):

cmd/namespace/sleep.go
```diff
@@ -29,9 +29,12 @@ func Sleep(ctx context.Context) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "sleep <name>",
 		Short: "Sleeps a namespace",
-		Args:  utils.ExactArgsAccepted(1, ""),
+		Args:  utils.MaximumNArgsAccepted(1, ""),
 		RunE: func(cmd *cobra.Command, args []string) error {
-			nsToSleep := args[0]
+			nsToSleep := okteto.Context().Namespace
```
jLopezbarb, line 34: What happens if the user doesn't have the `okteto ctx` set? I think this solve be moved after the ctx command run

AgustinRamiroDiaz: +1

NitishKumar06: image

NitishKumar06: `sleep` and `wake` command will only work for the clusters that have okteto installed.

jLopezbarb: What I meant is that I think that if you remove `~/.okteto/context/config.json` (or have it malformed) it would throw a panic and the user won't have any output. If we move the code after the context command, the user will get the selector for choosing the cluster to connect to

PR #3584, opened 2023-04-28
Fixes wrong spinner on okteto deploy --remote

NitishKumar06:
Fixes:https://github.com/okteto/okteto/issues/3572

Attached images:
Screenshot from 2023-04-29 00-18-56
image

### DIFF
--- a/pkg/cmd/build/buildkit.go
+++ b/pkg/cmd/build/buildkit.go
@@ -31,7 +31,6 @@ import (
 	"github.com/okteto/okteto/pkg/config"
 	oktetoLog "github.com/okteto/okteto/pkg/log"
 	"github.com/okteto/okteto/pkg/okteto"
-	"github.com/okteto/okteto/pkg/registry"
 	"github.com/okteto/okteto/pkg/types"
 	"github.com/pkg/errors"
 	"golang.org/x/oauth2"
@@ -121,11 +120,7 @@ func getSolveOpt(buildOptions *types.BuildOptions) (*client.SolveOpt, error) {
 			},
 		}
 	}
-	reg := registry.NewOktetoRegistry(okteto.Config{})
 	for _, cacheFromImage := range buildOptions.CacheFrom {
-		if _, err := reg.GetImageTagWithDigest(cacheFromImage); err != nil {
-			continue
-		}
 		opt.CacheImports = append(
 			opt.CacheImports,
 			client.CacheOptionsEntry{
@@ -286,7 +281,11 @@ func solveBuild(ctx context.Context, c *client.Client, opt *client.SolveOpt, pro
 			// not using shared context to not disrupt display but let it finish reporting errors
 			return progressui.DisplaySolveStatus(context.TODO(), "", nil, w, plainChannel)
 		case "deploy":
-			err := deployDisplayer(context.TODO(), plainChannel)
+			err := deployDisplayer(context.TODO(), plainChannel, &types.BuildOptions{OutputMode: "deploy"})
+			commandFailChannel <- err
+			return err
+		case "destroy":
+			err := deployDisplayer(context.TODO(), plainChannel, &types.BuildOptions{OutputMode: "destroy"})
 			commandFailChannel <- err
 			return err
 		default:
--- a/pkg/cmd/build/logger.go
+++ b/pkg/cmd/build/logger.go
@@ -17,6 +17,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"github.com/okteto/okteto/pkg/types"
 	"strings"
 	"time"
 
@@ -30,7 +31,7 @@ const (
 	largeContextThreshold = 50000000
 )
 
-func deployDisplayer(ctx context.Context, ch chan *client.SolveStatus) error {
+func deployDisplayer(ctx context.Context, ch chan *client.SolveStatus, o *types.BuildOptions) error {
 	// TODO: import build timeout
 	timeout := time.NewTicker(10 * time.Minute)
 	defer timeout.Stop()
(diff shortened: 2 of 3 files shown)


NitishKumar06:
This is not actually a big line change. I'm not sure why my local branch is pushing those extra commits and a merged commit at the same time. However, I've squashed these commits.

Apart from that, this PR solves the issue mentioned above!

NitishKumar06:
[This](https://github.com/okteto/okteto/pull/3584) PR is closed.

pkg/cmd/build/logger.go
```diff
@@ -54,7 +54,47 @@ func deployDisplayer(ctx context.Context, ch chan *client.SolveStatus) error {
...
+				if t.err != nil {
+					return t.err
+				}
+				return nil
+			}
+		}
+	}
+}
+
+func destroyDisplayer(ctx context.Context, ch chan *client.SolveStatus) error {
```
jLopezbarb, line 73: I think the only thing that change is the spinner text. Could we pass an argument and reuse the function to not have duplicated code (whenever we need to update one we'll need to update the other and it's very likely that we forget 😅 ).

NitishKumar06: Here's what I've understood up till now:

- I think in the following code:

```buildOptions := build.OptsFromBuildInfoForRemoteDeploy(buildInfo, &types.BuildOptions{OutputMode: "deploy"})```

When we pass the _OutputMode_ as "deploy" or "destroy", this leads to the case we've mentioned in the switch statement in _buildkit.go_:

```
case "deploy", "destroy":
		err := deployDisplayer(context.TODO(), plainChannel)
		commandFailChannel <- err
		return err
```

 - The _deployDisplayer()_ method at [L:33](https://github.com/okteto/okteto/blob/master/pkg/cmd/build/logger.go#L33) calls [t.display() ](https://github.com/okteto/okteto/blob/master/pkg/cmd/build/logger.go#LL135C7-L135C7) method at L:135 that is responsible for displaying the text which we need to modify. 
 
 Correct?
 
 **Problem that I'm facing:**
 
 - I'm thinking of adding an _if_ statement to check about which cmd is being executed - whether it's "destroy" or "deploy". I think, one way of doing this is by fetching the value from [here](https://github.com/okteto/okteto/blob/master/cmd/destroy/remote.go#L149). However, this is not directly possible to access its value, how should I proceed to fetch that value? 
 
- If we were to pass an argument, what should it look like?

Comment (jLopezbarb) [reply pkg/cmd/build/logger.go:73]: Sorry I didn't see this message. Everthing you posted is correct. 

> I'm thinking of adding an if statement to check about which cmd is being executed - whether it's "destroy" or "deploy". I think, one way of doing this is by fetching the value from [here](https://github.com/okteto/okteto/blob/master/cmd/destroy/remote.go#L149). However, this is not directly possible to access its value, how should I proceed to fetch that value?

You can add a new argument to `displayer` the `progress` variable which contains `deploy` or `destroy`

I hope that answer your question if not let me know

Comment (AdrianPedriza) [reply pkg/cmd/build/logger.go:73]: I also believe that adding an argument is the cleanest approach in this case. I see that there are details still to be covered in the PR. If you don't mind I will change it to draft to consider it's ready for review status and make a new revision. 👍

Comment (NitishKumar06) [reply pkg/cmd/build/logger.go:73]: - Should the type of `progress` variable be `*types.BuildOptions`?
- Just to confirm, do we intend to add the `progress` argument to `t.display()` func, right?

Comment (jLopezbarb) [reply pkg/cmd/build/logger.go:73]: > Should the type of progress variable be *types.BuildOptions?

I think we can work with it being a string since it's the only thing we need from `*types.BuildOptions`

> Just to confirm, do we intend to add the progress argument to t.display() func, right?

Yes, we should add the argument there

Comment (NitishKumar06) [reply pkg/cmd/build/logger.go]: Fixed!

**Comment by NitishKumar06** [comment]
@jLopezbarb Ready for review!

**Review by jLopezbarb — APPROVED** [review]

File: pkg/cmd/build/buildkit.go [hunk 2]
```diff
@@ -286,7 +281,11 @@ func solveBuild(ctx context.Context, c *client.Client, opt *client.SolveOpt, pro
 			// not using shared context to not disrupt display but let it finish reporting errors
 			return progressui.DisplaySolveStatus(context.TODO(), "", nil, w, plainChannel)
 		case "deploy":
-			err := deployDisplayer(context.TODO(), plainChannel)
+			err := deployDisplayer(context.TODO(), plainChannel, &types.BuildOptions{OutputMode: "deploy"})
```
jLopezbarb, line 284: NIT: I'd just go with a single string `deploy` instead of creating the buildOptions but It's ok like this too

pkg/cmd/build/logger.go
```diff
@@ -54,7 +62,7 @@ func deployDisplayer(ctx context.Context, ch chan *client.SolveStatus) error {
 					oktetoLog.Info(err.Error())
 					continue
 				}
-				t.display()
+				t.display(outputMode)
```
AdrianPedriza, line 65:
```suggestion
				t.display(o.OutputMode)
```

can we use directly `o.OutputMode`?

AdrianPedriza (approved):
only minor comments, LGTM 👍

pkg/cmd/build/logger.go
NitishKumar06: Yes, we can. I had this thought at the start while making changes. However, I felt the code would be more readable and understandable if we create a variable rather than directly passing value. Do you want to pass the value directly?

AdrianPedriza: if you think is more readable your approach go ahead with it 👍

PR #3586, opened 2023-05-02
Add external resources to preview endpoints

maroshii:
External resources are not being listed when running: `okteto preview endpoints {name}`

Two different fixes here:

- We first try to read the `preview.Endpoints` value which is already being returned by the api and fallback to listing by resource. 
- In the fallback, we now read the `externals` value from the preview response to resolve the external resources endpoints.

Note that currently, the api is not returning correctly the endpoints in `preview.endpoints` so we'll always list endpoints from the fallback. This change goes along with an api change that takes care of correctly returning the canonical `previews.endpoints`

### DIFF
--- a/pkg/okteto/preview.go
+++ b/pkg/okteto/preview.go
@@ -84,8 +84,10 @@ type resourceInfo struct {
 	DeployedBy graphql.String
 }
 type previewEndpoints struct {
+	Endpoints    []endpointURL
 	Deployments  []deploymentEndpoint
 	Statefulsets []statefulsetEdnpoint
+	Externals    []externalEndpoints
 }
 
 type deploymentEndpoint struct {
@@ -93,6 +95,9 @@ type deploymentEndpoint struct {
 }
 
 type statefulsetEdnpoint struct {
+	Endpoints []endpointURL
+}
+type externalEndpoints struct {
 	Endpoints []endpointURL
 }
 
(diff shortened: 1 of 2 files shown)


pkg/okteto/preview.go
```diff
@@ -240,21 +245,40 @@ func (c *previewClient) ListEndpoints(ctx context.Context, previewName string) (
...
+
+	for _, endpoint := range queryStruct.Response.Endpoints {
+		endpointsMap[endpoint.Url] = true
+	}
+
+	// @francisco - 2023//05/02
+	// The backend Endpoints field was being return empty in some older versions
+	// of the okteto api. Lets make sure we correctly resolve the endpoints from
+	// the known resources. All this below should be safe to remove in newer versions
+	// of the okteto cli
```
jLopezbarb, line 258: Can we specify the version when we can remove this code? For example `when all okteto cluster are > 1.7`

jLopezbarb (approved):

pkg/okteto/preview.go
maroshii, line 258:
```suggestion
	// The backend Endpoints field was being return empty in okteto clusters <1.9
	// Lets make sure we correctly resolve the endpoints from the known resources
	// All this below should be safe to remove in newer versions of the okteto cli
```

PR #3593, opened 2023-05-05
fix: consume envs from dependency

AdrianPedriza:
# Proposed changes
- remove default variables to be pushed as `OKTETO_DEPENDENCY` vars into the config map
- add unit test
- add e2e test covering the following scenarios:
  - consume as env a value coming from dependency variable
  - consume as env a value coming from dependency `OKTETO_ENV`

In order to be able to pass the e2e test, we need first of all to have an installer with these changes and [this fix](https://github.com/okteto/pipeline-installer-image/pull/94)

### DIFF
--- a/cmd/deploy/local.go
+++ b/cmd/deploy/local.go
@@ -186,25 +186,22 @@ func (ld *localDeployer) deploy(ctx context.Context, deployOptions *Options) err
 
 func (ld *localDeployer) setDefaultVariables(opts *Options) {
 	for _, k := range defaultVariables {
-		if k == model.OktetoDomainEnvVar {
-			continue
-		}
-		if k == constants.KubeConfigEnvVar {
-			opts.Variables = append(opts.Variables, fmt.Sprintf("%s=%s", constants.KubeConfigEnvVar, ld.TempKubeconfigFile))
-		} else if k == model.OktetoNamespaceEnvVar {
-			opts.Variables = append(opts.Variables, fmt.Sprintf("%s=%s", model.OktetoNamespaceEnvVar, okteto.Context().Namespace))
-		} else if k == constants.OktetoAutodiscoveryReleaseName {
-			opts.Variables = append(opts.Variables, fmt.Sprintf("%s=%s", constants.OktetoAutodiscoveryReleaseName, format.ResourceK8sMetaString(opts.Name)))
-
-		} else {
+		switch k {
+		case model.OktetoDomainEnvVar:
+			if okteto.IsOkteto() {
+				opts.Variables = append(
+					opts.Variables, fmt.Sprintf("%s=%s", k, okteto.GetSubdomain()),
+				)
+			}
+		case constants.KubeConfigEnvVar:
+			opts.Variables = append(opts.Variables, fmt.Sprintf("%s=%s", k, ld.TempKubeconfigFile))
+		case model.OktetoNamespaceEnvVar:
+			opts.Variables = append(opts.Variables, fmt.Sprintf("%s=%s", k, okteto.Context().Namespace))
+		case constants.OktetoAutodiscoveryReleaseName:
+			opts.Variables = append(opts.Variables, fmt.Sprintf("%s=%s", k, format.ResourceK8sMetaString(opts.Name)))
+		default:
 			opts.Variables = append(opts.Variables, fmt.Sprintf("%s=%s", k, "true"))
 		}
-	}
-
-	if okteto.IsOkteto() {
-		opts.Variables = append(
-			opts.Variables, fmt.Sprintf("%s=%s", model.OktetoDomainEnvVar, okteto.GetSubdomain()),
-		)
 	}
 }
 
--- a/cmd/deploy/local.go
+++ b/cmd/deploy/local.go
@@ -55,6 +55,27 @@ type localDeployer struct {
 	DivertDriver divert.Driver
 }
 
+var (
+	defaultVariables = []string{
+		// Set KUBECONFIG environment variable as environment for the commands to be executed
+		constants.KubeConfigEnvVar,
+		// Set OKTETO_WITHIN_DEPLOY_COMMAND_CONTEXT env variable, so all okteto commands ran inside this deploy
+		// know they are running inside another okteto deploy
+		constants.OktetoWithinDeployCommandContextEnvVar,
+		// Set OKTETO_SKIP_CONFIG_CREDENTIALS_UPDATE env variable, so all the Okteto commands executed within this command execution
+		// should not overwrite the server and the credentials in the kubeconfig
+		constants.OktetoSkipConfigCredentialsUpdate,
+		// Set OKTETO_DISABLE_SPINNER=true env variable, so all the Okteto commands disable spinner which leads to errors
+		oktetoLog.OktetoDisableSpinnerEnvVar,
+		// Set OKTETO_NAMESPACE=namespace-name env variable, so all the commandsruns on the same namespace
+		model.OktetoNamespaceEnvVar,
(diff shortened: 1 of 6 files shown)


cmd/deploy/validate.go
```diff
@@ -23,14 +23,27 @@ type envVar struct {
 	value string
 }
 
-func validateAndSet(variables []string, setEnv func(key, value string) error) error {
-	envVars, err := parse(variables)
+func validateAndSetVariables(opts *Options, setEnv func(key, value string) error) error {
```
AdrianPedriza, line 26: we need to remove escape values that comes from installer

integration/deploy/pipeline_test.go
```diff
@@ -154,6 +165,38 @@ func TestDeployPipelineManifestInsidePipeline(t *testing.T) {
 	require.True(t, k8sErrors.IsNotFound(err))
 }
 
+// TestDeployPipelineAndConsumerEnvsFromDependency tests the following scenario:
+// - Deploying a dependency
+// - Consume dependency's variable from main pipeline
+func TestDeployPipelineAndConsumerEnvsFromDependency(t *testing.T) {
```
jLopezbarb, line 171: This test should be in the `integration/okteto` folder. That's the only folder where we can store tests that requires a cluster with okteto installed

cmd/deploy/validate.go
```diff
@@ -23,14 +23,23 @@ type envVar struct {
...
+	cleanEscapedVariables(opts)
```
jLopezbarb, line 27: NIT: Can we not modify options, return the variables in a list and assign it later. I think it's more verbose. WDYT

cmd/deploy/validate_test.go
```diff
@@ -65,7 +65,9 @@ func Test_validateAndSet(t *testing.T) {
 				return nil
 			}
 
-			err := validateAndSet(tt.variables, setEnvStorage)
+			opts := &Options{Variables: tt.variables}
+
+			err := validateAndSetVariables(opts, setEnvStorage)
```
jLopezbarb, line 70: Can we test `cleanEscapedVariables`

cmd/deploy/validate.go
```diff
@@ -23,14 +23,23 @@ type envVar struct {
...
 	if err != nil {
 		return err
 	}
 	return setOptionVarsAsEnvs(envVars, setEnv)
 }
 
+func cleanEscapedVariables(opts *Options) {
+	cleanedVariables := []string{}
+	for _, v := range opts.Variables {
+		cleanedVariables = append(cleanedVariables, strings.ReplaceAll(v, "\"", ""))
```
jLopezbarb, line 38: How can I reproduce this case? When we do this for the name we don't have to do it

cmd/deploy/local.go
```diff
@@ -185,6 +184,30 @@ func (ld *localDeployer) deploy(ctx context.Context, deployOptions *Options) err
 	return err
 }
 
+func (ld *localDeployer) setDefaultVariables(opts *Options) {
+	for _, k := range defaultVariables {
+		if k == model.OktetoDomainEnvVar {
+			continue
+		}
+		if k == constants.KubeConfigEnvVar {
```
jLopezbarb, line 192: Can it be a switch instead of if else

integration/deploy/pipeline_test.go
AdrianPedriza: resolving suggestion to addressed it in https://github.com/okteto/okteto/pull/3594

cmd/deploy/validate.go
AdrianPedriza: you can check the e2e test for it. We need to escape values in the installer in order to forward variables using `exec.Command`. If we do not, you will not be able to use the proper value, instead of that, you will use something like: `\"a value\"`. In the following example, config map name cannot contain `\`.
````yaml
deploy:
  - kubectl create cm $OKTETO_DEPENDENCY_TEST_VARIABLE_CMNAME --from-literal=key1="$OKTETO_DEPENDENCY_TEST_VARIABLE_MY_DYNAMIC_APP_ENV"
dependencies:
  test:
    repository: https://github.com/okteto/go-getting-started
    branch: generating-dynamic-envs
    variables:
      CMNAME: test
    wait: true
````

File: cmd/deploy/validate.go (hunk 3 above)
Comment (AdrianPedriza) [reply cmd/deploy/validate.go]: done 4e2b7b8ee0f399d8715119761e4b511797e8d6b6

File: cmd/deploy/validate_test.go (hunk 4 above)
Comment (AdrianPedriza) [reply cmd/deploy/validate_test.go]: refactored. Done 4e2b7b8ee0f399d8715119761e4b511797e8d6b6

File: integration/deploy/pipeline_test.go (hunk 2 above)
Comment (AdrianPedriza) [reply integration/deploy/pipeline_test.go]: btw, done 9d45dc7e2c0a589e4648e41982a19cc143347704

File: cmd/deploy/local_test.go [hunk 7]
```diff
@@ -41,3 +43,78 @@ func TestDeployNotRemovingEnvFile(t *testing.T) {
 	require.NoError(t, err)
 
 }
+
+func TestAddOktetoEnvsValuesAsDesployVariables(t *testing.T) {
```
andreafalzetti, line 47:
```suggestion
func TestAddOktetoEnvsValuesAsDeployVariables(t *testing.T) {
```

cmd/deploy/local.go
```diff
@@ -280,6 +287,41 @@ func (ld *localDeployer) runDeploySection(ctx context.Context, opts *Options) er
 	return nil
 }
 
+func addOktetoEnvsValuesAsDesployVariables(opts *Options, oktetoEnvFile string, addedEnvs map[string]string) {
```
andreafalzetti, line 290:
```suggestion
func addOktetoEnvsValuesAsDeployVariables(opts *Options, oktetoEnvFile string, addedEnvs map[string]string) {
```

cmd/deploy/local.go
```diff
@@ -208,26 +228,13 @@ func (ld *localDeployer) runDeploySection(ctx context.Context, opts *Options) er
...
-			envsFromOktetoEnvFile = append(envsFromOktetoEnvFile, fmt.Sprintf("%s=%s", k, v))
-		}
-
-		// the variables in the $OKTETO_ENV file are added as environment variables
-		// to the executor. If there is already a previously set value for that
-		// variable, the executor will use in next command the last one added which
-		// corresponds to those coming from $OKTETO_ENV.
-		opts.Variables = append(opts.Variables, envsFromOktetoEnvFile...)
+		// we need to set available envs in OKTETO_ENV as deploy variables for next command
+		addOktetoEnvsValuesAsDesployVariables(opts, oktetoEnvFile.Name(), addedOktetoEnvs)
```
andreafalzetti, line 232:
```suggestion
		addOktetoEnvsValuesAsDeployVariables(opts, oktetoEnvFile.Name(), addedOktetoEnvs)
```

cmd/deploy/validate.go
jLopezbarb: It's weird because If I work directly with the terminal cobra doesn't add the "" but if I executed it with exec it does 🤔.

Maybe we can try to use `--` to let the `cmd.Exec` that is a single argument. 

The problem with the code above is that if a user wants to add a variable that contains a double quote it will get removed...

jLopezbarb:
@AdrianPedriza What's the status of this PR? It has been open for 2 months now

AdrianPedriza:
> @AdrianPedriza What's the status of this PR? It has been open for 2 months now

ÎIneed to to retake it and see what the status is since there have been changes in components that could affect the PR. I'm converting it to draft for now

PR #3595, opened 2023-05-05
Modify okteto  ns create command to prevent selecting the  new namespace as current namespace

### LINKED ISSUE #3579: Option to create namespace without setting it as current
The description of `okteto namespace create` (version 2.14.0) is currently:

```
Usage:
  okteto namespace create <name> [flags]

Flags:
  -h, --help                  help for create
  -m, --members stringArray   members of the namespace, it can the username or email
```

When you use this to create a new namespace, it automatically sets this as the current namespace (same as applying `okteto namespace use <namespace>`).

We want to script things on top of `okteto namespace create` to automate create a namespace (or sometimes multiple) without having to remember to switch back to the previous namespace.

# Proposed solution

To have an option `--keep-current` or `--no-change-current` or something, to disable automatically changing the current namespace

NitishKumar06:
# Proposed changes

Fixes #3579

image

### DIFF
--- a/cmd/namespace/create.go
+++ b/cmd/namespace/create.go
@@ -29,15 +29,16 @@ import (
 
 // CreateOptions represents the options that namespace create has
 type CreateOptions struct {
-	Members   *[]string
-	Namespace string
-	Show      bool
+	Members      *[]string
+	Namespace    string
+	Show         bool
+	SetCurrentNs bool
 }
 
 // Create creates a namespace
 func Create(ctx context.Context) *cobra.Command {
 	options := &CreateOptions{
-		Show: true,
+		Show: false,
 	}
 	cmd := &cobra.Command{
 		Use:   "create <name>",
@@ -46,12 +47,10 @@ func Create(ctx context.Context) *cobra.Command {
 			if err := contextCMD.NewContextCommand().Run(ctx, &contextCMD.ContextOptions{}); err != nil {
 				return err
 			}
-
 			options.Namespace = args[0]
 			if !okteto.IsOkteto() {
 				return oktetoErrors.ErrContextIsNotOktetoCluster
 			}
-
 			nsCmd, err := NewCommand()
 			if err != nil {
 				return err
@@ -64,6 +63,7 @@ func Create(ctx context.Context) *cobra.Command {
 	}
 
 	options.Members = cmd.Flags().StringArrayP("members", "m", []string{}, "members of the namespace, it can the username or email")
+	cmd.Flags().BoolVarP(&options.SetCurrentNs, "use", "", true, "use the newly created namespace as the current namespace")
 	return cmd
 }
 
@@ -84,11 +84,17 @@ func (nc *NamespaceCommand) Create(ctx context.Context, opts *CreateOptions) err
 	ctxOptions := &contextCMD.ContextOptions{
 		IsCtxCommand: opts.Show,
 		IsOkteto:     true,
-		Save:         true,
-		Show:         false,
 		Token:        okteto.Context().Token,
 		Namespace:    oktetoNS,
 		Context:      okteto.Context().Name,
+	}
+
+	if opts.SetCurrentNs == true {
+		ctxOptions.Save = true
+		ctxOptions.Show = true
+	} else {
+		ctxOptions.Save = false
+		ctxOptions.Show = false
 	}
 
 	if err := nc.ctxCmd.Run(ctx, ctxOptions); err != nil {
--- a/cmd/context/options.go
+++ b/cmd/context/options.go
@@ -30,6 +30,7 @@ type ContextOptions struct {
 	OnlyOkteto            bool
 	Show                  bool
 	Save                  bool
+	SetCurrentNs          bool
 	IsCtxCommand          bool
 	CheckNamespaceAccess  bool
 	IsOkteto              bool


jLopezbarb:
@NitishKumar06 We'd need documentation for this new flag. Can you work on that when you finish the implementation?

NitishKumar06:
> @NitishKumar06 We'd need documentation for this new flag. Can you work on that when you finish the implementation?

Yes, I'd be happy to address this too.

cmd/namespace/create.go
```diff
@@ -29,15 +29,16 @@ import (
...
+	Namespace  string
+	Show       bool
+	SetCurrent bool
 }
 
 // Create creates a namespace
 func Create(ctx context.Context) *cobra.Command {
 	options := &CreateOptions{
-		Show: true,
+		Show: false,
```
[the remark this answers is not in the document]
NitishKumar06: Then for a particular namespace we create by specifying argument, we can pass `Show: True` for it. Wdyt?

cmd/namespace/create.go
```diff
@@ -60,7 +60,7 @@ func Create(ctx context.Context) *cobra.Command {
 			analytics.TrackCreateNamespace(err == nil)
 			return err
 		},
-		Args: utils.ExactArgsAccepted(1, ""),
+		Args: utils.MaximumNArgsAccepted(2, ""),
 	}
 
 	options.Members = cmd.Flags().StringArrayP("members", "m", []string{}, "members of the namespace, it can the username or email")
```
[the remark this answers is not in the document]
NitishKumar06: Added!

cmd/namespace/create.go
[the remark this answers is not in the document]
NitishKumar06: Fixed.

cmd/namespace/create.go
```diff
@@ -84,13 +86,19 @@ func (nc *NamespaceCommand) Create(ctx context.Context, opts *CreateOptions) err
...
 		IsCtxCommand: opts.Show,
 		IsOkteto:     true,
-		Save:         true,
-		Show:         false,
 		Token:        okteto.Context().Token,
 		Namespace:    oktetoNS,
 		Context:      okteto.Context().Name,
 	}
 
+	if opts.SetCurrent == true {
```
NitishKumar06, line 94: @jLopezbarb Would you mind taking a look, how can i improve this piece of code?

jLopezbarb: Right now there is no other way of doing this... So I guess this is the correct approach

cmd/namespace/create.go
```diff
@@ -60,10 +61,11 @@ func Create(ctx context.Context) *cobra.Command {
 			analytics.TrackCreateNamespace(err == nil)
 			return err
 		},
-		Args: utils.ExactArgsAccepted(1, ""),
+		Args: utils.MaximumNArgsAccepted(2, ""),
 	}
 
 	options.Members = cmd.Flags().StringArrayP("members", "m", []string{}, "members of the namespace, it can the username or email")
+	cmd.Flags().BoolVarP(&options.SetCurrent, "set current", "u", false, "use to set ns as default ns while creating")
```
jLopezbarb, line 68: @pchico83 @rlamana Can you help with the naming of this new flag? 

We need a way of preserving the current approach (changing to the new namespace created if the user doesn't set a flag). If we set it default to true to set it to false the user would do something like `okteto namespace create --set-current=false`. I guess we could do something like we do for the `--no-bash` and name it something like `--no-switch` so the default value is `false` keeping the current behavior untouched and the new behavior would look like `okteto namespace create --no-switch`. WDYT?

NitishKumar06: or more simply we can use `--use`. If `--use` flag is specified, we can set the newly created namespace as the current namespace. And if not, the namespace will be created but not be set as current namespace.

pchico83: It's explain [here](https://github.com/okteto/okteto/issues/3579#issuecomment-1524709937).
The flag is `--use`, to match the `okteto ctx use` command. The default value is `true` to avoid breaking changes and keep the current behavior. And you can disable it with `okteto ns create --use=false`

NitishKumar06: @jLopezbarb I'll work on adding a new flag name. Could you help me on how to set the flag as `--use`. The shorthand is exceeding the length if i take more than 1 character.

jLopezbarb: Let's only use the longer syntax then:
```suggestion
	cmd.Flags().BoolVarP(&options.SetCurrent, "use", "", true, "use to set ns as default ns while creating")
```

cmd/namespace/create.go
```diff
@@ -29,15 +29,17 @@ import (
...
+	Show         bool
+	SetCurrentNs bool
 }
 
 // Create creates a namespace
 func Create(ctx context.Context) *cobra.Command {
 	options := &CreateOptions{
-		Show: true,
+		Show:         false,
+		SetCurrentNs: false,
```
NitishKumar06, line 42:
```suggestion
		SetCurrentNs: false,
```

I'm assuming that `SetCurrentNs` value will be taken as `false` when we pass no flag , something like `okteto namespace create <ns-name>`. However, No matter what, the value is automatically taken as true same as default value for flag. Any idea why? cc @jLopezbarb

NitishKumar06 (commented):
When we pass the flag `--use` and `--use=false`, the command runs well as expected. This is because for the flag, we have specified default value as `true` that will hit conditions in the _if-else_ block on line 93.

However, when we pass no flag such as `okteto ns create <ns-create>`, the default value of `setCurrentNs` should be taken as false as line 42. But it is taken as true exactly like the default value for the flag (we're not passing a flag here).

cmd/namespace/create.go
NitishKumar06: Fixed it!

NitishKumar06:
**Update:** This PR is ready to be reviewed now.
- Squashed the commits
- Added the test image in the PR description

cmd/namespace/create.go
```diff
@@ -60,10 +60,11 @@ func Create(ctx context.Context) *cobra.Command {
 			analytics.TrackCreateNamespace(err == nil)
 			return err
 		},
-		Args: utils.ExactArgsAccepted(1, ""),
+		Args: utils.MaximumNArgsAccepted(2, ""),
 	}
 
 	options.Members = cmd.Flags().StringArrayP("members", "m", []string{}, "members of the namespace, it can the username or email")
+	cmd.Flags().BoolVarP(&options.SetCurrentNs, "use", "", false, "use the newly created namespace as the current namespace")
```
AdrianPedriza, line 67: using `false` as the default value change the behavior we have until now. Should we use as default value `true` in order to keep the same behavior when the input is the same as before? cc @jLopezbarb

NitishKumar06: Could you specify what do you mean by 'behaviour' here? 

Earlier the default value was set as `true` but then, new ns was selected as current ns with flag being passed and not being passed. After some try, I changed it to `false` and it's running well as we expect it now.

cmd/namespace/create.go
jLopezbarb, line 42: I'd remove this line since it's already being set by the flag. Having it here just overwrites the flag

cmd/namespace/create.go
```diff
@@ -60,10 +60,11 @@ func Create(ctx context.Context) *cobra.Command {
 			analytics.TrackCreateNamespace(err == nil)
 			return err
 		},
-		Args: utils.ExactArgsAccepted(1, ""),
+		Args: utils.MaximumNArgsAccepted(2, ""),
```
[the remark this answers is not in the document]
NitishKumar06: I forgot to change it. My bad! 

I thought the flag could have been taken as another argument but no. Thanks for pointing this!

NitishKumar06:
I think we're ready with the next set of review:)

cmd/namespace/create.go
AdrianPedriza: I mean, if the user executes `okteto ns create my-namespace` then the namespace will not be changed because by default `--use` opt is false. All users using this command would be affected since now by default the namespace will not be switched following the changes of this PR. I think we could set `--use` to true to keep the same behavior if the user does not set the `--use` flag in the command. Is this the behavior we want? cc @jLopezbarb

NitishKumar06: Oh! I thought we want to make it work something like I mentioned earlier. If the behavior is the same as you're saying, this PR would have merged earlier because I was thinking of implement different behavior rather than you mentioned. 

Th behaviour you mentioned was already working. I'll push the new changes now. Thanks for confirming!

NitishKumar06: Fixed!

NitishKumar06:
cc @jLopezbarb

jLopezbarb:
I'll review it on Monday but code looks good. I need to check if this last change is affecting other commands when creating a new namespace

jLopezbarb:
I think this breaks another scenario where we add the `--namespace` to another command like `okteto up --namespace non-existent-ns`, but I'm not able to reproduce it right now. I'll create another issue for that and only check the ctx command

jLopezbarb (approved):

NitishKumar06:
> I think this breaks another scenario where we add the `--namespace` to another command like `okteto up --namespace non-existent-ns`, but I'm not able to reproduce it right now. I'll create another issue for that and only check the ctx command

I'll try to look at this in more detail and see what other command is being affected.

AdrianPedriza:
code looks good but I observe a different output in this PR. 

| Before   |      After      |
|----------|:-------------:|
| Captura de Pantalla 2023-05-31 a las 10 33 51 |  Captura de Pantalla 2023-05-31 a las 10 33 07 | 

we have changed the success log to info log. Is that what we want? cc @rlamana @jLopezbarb

rlamana:
@AdrianPedriza it looks like the "after" version should be the good one 👌

AdrianPedriza:
> @AdrianPedriza it looks like the "after" version should be the good one 👌

yess! it looks better now to me too

PR #3601, opened 2023-05-08
Add okteto certificates in remote deploy

maroshii:
# Proposed changes

Fetch the certificate from the server and add it to `/etc/ssl/certs` in the deploy remote. This will allow buildkit to trust okteto certificates even if certs are self signed.

### DIFF
--- a/cmd/deploy/remote.go
+++ b/cmd/deploy/remote.go
@@ -16,9 +16,13 @@ package deploy
 import (
 	"context"
 	"crypto/rand"
+	"crypto/tls"
+	"encoding/base64"
+	"encoding/pem"
 	"errors"
 	"fmt"
 	"math/big"
+	"net/url"
 	"os"
 	"path/filepath"
 	"regexp"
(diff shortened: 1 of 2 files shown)


cmd/deploy/remote.go
```diff
@@ -74,6 +78,8 @@ COPY . /okteto/src
 WORKDIR /okteto/src
 
 ENV OKTETO_INVALIDATE_CACHE {{ .RandomInt }}
+ARG OKTETO_TLS_CERT_BASE64
```
jLopezbarb, line 81: We should also do this fro `cmd/destroy/remote.go`

maroshii: done

teresaromero (approved):

jLopezbarb:
> @jLopezbarb is `okteto deploy --remote` only for Okteto clusters? If not, should we catch that exception in the current implementation?

Post review answer: Yes, we only support `--remote` on okteto clusters because the remote happens in the buildkit instance of okteto

PR #3603, opened 2023-05-09
Okteto deploy with only dependencies

### LINKED ISSUE #3364: `okteto deploy` shouldn't fail if manifest has `dependencies` but not `deploy` commands
**Describe the bug**
It's a common use case to deploy an okteto manifest with only `dependencies` to schedule the deployment of several repos together.

**To Reproduce**

1. Create a manifest with only `dependencies` section
1. `okteto deploy`

**Expected behavior**
It should work.

AnshulAngaria:
# Proposed changes

Fixes #3364

Okteto deploy fails if manifest only has dependencies section. This PR makes changes so that it does not fail.


AnshulAngaria:
@AgustinRamiroDiaz I will open a new PR, I messed up this branch while rebasing

PR #3605, opened 2023-05-09
feat: add test cases for `namespace sleep|wake`

### LINKED ISSUE #3604: Add test cases for `namespace sleep|wake`
We don't have test cases for `namespace sleep|wake`

With the addition of the `sleep|wake` commands for `namespace`, I think it would be a good idea to add test cases for each.

rakshitgondwal:
# Proposed changes

Fixes #3604 

Added test cases for `namespace sleep|wake`.

### DIFF
--- a/cmd/namespace/sleep_test.go
+++ b/cmd/namespace/sleep_test.go
@@ -1,0 +1,110 @@
+// Copyright 2023 The Okteto Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package namespace
+
+import (
+	"context"
+	"testing"
+
+	"github.com/okteto/okteto/internal/test/client"
+	"github.com/okteto/okteto/pkg/okteto"
+	"github.com/okteto/okteto/pkg/types"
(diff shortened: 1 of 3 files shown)


cmd/namespace/sleep_test.go
```diff
@@ -0,0 +1,110 @@
...
+			initialNamespacesAtOktetoClient: initNamespaces,
+			fakeOkClient: &client.FakeOktetoClient{
+				Namespace:    client.NewFakeNamespaceClient(initNamespaces, nil),
+				Users:        client.NewFakeUsersClient(usr),
+				StreamClient: client.NewFakeStreamClient(&client.FakeStreamResponse{}),
+			},
+			fakeK8sClient: fake.NewSimpleClientset(),
+		},
+		{
+			name:                            "sleep non existing ns",
```
rakshitgondwal, line 73: Hey @jLopezbarb @pchico83!
In this case, it expects an error to be returned as there is no `test-non-existing` namespace present but while trying to run this test, it still sleeps a `test-non-existing` namespace, idk why. Can you help me out here?
Screenshot from 2023-05-10 00-09-24

cmd/namespace/sleep_test.go
```diff
@@ -0,0 +1,110 @@
...
+						Namespace: currentNamespace,
+						UserID:    "1",
+					},
+				},
+				CurrentContext: "test-context",
+			}
+			nsFakeCommand := newFakeNamespaceCommand(tt.fakeOkClient, tt.fakeK8sClient, usr)
+			err := nsFakeCommand.ExecuteSleepNamespace(ctx, tt.toSleepNs)
+			if tt.err != nil {
+				assert.EqualError(t, err, tt.err.Error())
```
jLopezbarb, line 104: Let's use `assert.ErrorIs` We wrap errors and we should check that the error is in the chain

jLopezbarb (approved):

AdrianPedriza (approved):
LGTM!

PR #3606, opened 2023-05-10
Fix kubetoken usage message

AgustinRamiroDiaz:
Signed-off-by: Agustín Díaz <agustin.ramiro.diaz@gmail.com>

In the last changes to the command we changed the positional arguments to be flags

### DIFF
--- a/cmd/kubetoken/kubetoken.go
+++ b/cmd/kubetoken/kubetoken.go
@@ -27,7 +27,7 @@ import (
 
 func KubeToken() *cobra.Command {
 	cmd := &cobra.Command{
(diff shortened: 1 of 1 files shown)


teresaromero:
@AgustinRamiroDiaz can we update status on this PR? is this still needed for the kubetoken feature?

AgustinRamiroDiaz:
> @AgustinRamiroDiaz can we update status on this PR? is this still needed for the kubetoken feature?

@teresaromero there's still an ongoing refinement for this feature, we might remove the command entirely

This in particular is just a message change, that I think we should do since it's currently wrong. It's not priority since it's only documentation for an unused command ATM

PR #3610, opened 2023-05-10
feat: improve `.gitignore`

### LINKED ISSUE #3609: Improve `.gitignore`
We use `make` to build the binary which builds it in `bin/oketeo` but I accidentally used `go build` to build the binary. I was about to commit the binary built in `./okteto`  but then suddenly noticed it. I think this can happen with other folks too.

To prevent this from happening, we can add `okteto` in the `.gitignore` file which would prevent this from happening. 
We can also add some more IDE files which are automatically generated when using a particular IDE such as `.atom` for Atom.

rakshitgondwal:
# Proposed changes

Fixes #3609

Changes made-
- Added `okteto` in `.gitignore`.
- Added `.atom` in `.gitognore`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 # IDEs
 .vscode
 .idea
+.atom
 
 # Golang binary
 cli
@@ -11,3 +12,4 @@ bin/*
 coverage.txt
 coverage.html
 dist
+okteto


jLopezbarb (approved):

NitishKumar06:
/lgtm
Ready to be meregd.

andreafalzetti (approved):

PR #3611, opened 2023-05-11
feat: add label flag to deploy preview command

### LINKED ISSUE #3607: Deploy preview accepts a list of labels
## Parent epic/feature

* Feature: [Labels for preview environment](https://www.notion.so/okteto/Labels-for-Preview-Environments-Epic-1-CLI-Support-4178b8e637a34829972c6baf9247be42?pvs=4)

## Description

- A new flag `label` (`-l` alias) is supported in the `deploy preview` command
- The new flag label is array type and can store several labels
- The labels added as flags are passed as query arguments to the gql `deployPreview` call
- If there is no label flag the behavior must be the same as it is right now. For example, the gql call should not receive a label argument (to keep retrocompatibility)

## Acceptance Criteria

- [ ]  A user can add several `labels` as flags
- [ ]  The flag `—label` can be used with the `-l` shortcut
- [ ]  If the backend return an error, the user must get a friendly error (validation error)
- [ ]  If there is no labels the gql call shouldn’t have any mention to labels
- [ ]  The output should show the labels that every preview env has

## Test strategy

- Unit tests covering:
    - Validation error
    - deployPreview with several labels
    - deployPreview with label
- E2e test covering
    - deploy with several labels

jLopezbarb:
# Proposed changes

Fixes #3607

- Add a `label` flag to deploy command to deploy a preview environment using the labels passed
- Refactor the deploy preview graphql code so it's more readable

### DIFF
--- a/pkg/okteto/client_test.go
+++ b/pkg/okteto/client_test.go
@@ -46,6 +46,9 @@ func (fc fakeGraphQLClient) Query(ctx context.Context, q interface{}, _ map[stri
 }
 func (fc fakeGraphQLClient) Mutate(ctx context.Context, m interface{}, _ map[string]interface{}) error {
 	if fc.mutationResult != nil {
+		if isAnInterface(m) {
+			m = reflect.ValueOf(m).Elem().Interface()
+		}
 		entityType := reflect.TypeOf(m).Elem()
 		for i := 0; i < entityType.NumField(); i++ {
 			value := entityType.Field(i)
@@ -55,6 +58,10 @@ func (fc fakeGraphQLClient) Mutate(ctx context.Context, m interface{}, _ map[str
 		}
 	}
 	return fc.err
+}
+
+func isAnInterface(m interface{}) bool {
+	return reflect.TypeOf(m).Kind() == reflect.Ptr && reflect.TypeOf(m).Elem().Kind() == reflect.Interface
 }
 
 type fakeGraphQLMultipleCallsClient struct {
--- a/pkg/okteto/preview.go
+++ b/pkg/okteto/preview.go
@@ -50,8 +50,26 @@ type InputVariable struct {
 
 type PreviewScope graphql.String
 
+type labelList []graphql.String
+
+type deployPreviewResponseInterface interface {
+	response() deployPreviewResponse
+}
+
 type deployPreviewMutation struct {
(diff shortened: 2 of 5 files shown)


pkg/okteto/preview_test.go
```diff
@@ -166,14 +167,48 @@ func TestDeployPreview(t *testing.T) {
 				err: nil,
 			},
 		},
+		{
```
andreafalzetti, line 170: did you consider checking if labels are returned?

pkg/okteto/client_test.go
```diff
@@ -46,6 +46,9 @@ func (fc fakeGraphQLClient) Query(ctx context.Context, q interface{}, _ map[stri
 }
 func (fc fakeGraphQLClient) Mutate(ctx context.Context, m interface{}, _ map[string]interface{}) error {
 	if fc.mutationResult != nil {
+		if isAnInterface(m) {
```
andreafalzetti, line 49: genuine question, why do we need this check?

pkg/okteto/preview.go
```diff
@@ -121,83 +139,74 @@ type previewIDStruct struct {
...
+		"variables":  variablesVariable,
+		"filename":   graphql.String(filename),
+	}
+
+	if len(labels) > 0 {
+		labelsVariable := make(labelList, 0)
+		for _, l := range labels {
+			labelsVariable = append(labelsVariable, graphql.String(l))
 		}
+		vars["labels"] = labelsVariable
```
andreafalzetti, line 207: do I understand correctly that this request now will send `labels` but we also need a change on the backend to receive them and store them in k8s?

andreafalzetti (approved):
code change LGTM, just a couple of questions to clarify for my own benefit of understanding 🙏

AdrianPedriza:
have we considered using labels to destroy a preview also? or to list endpoints related to a preview that has specific labels..?

AdrianPedriza (approved):
LGTM

pkg/okteto/client_test.go
jLopezbarb: This is to get the struct that implements the interface. Just for mocking the behavior of gql call

pkg/okteto/preview_test.go
jLopezbarb: We will when we list the previews

PR #3614, opened 2023-05-12
Deploy command works even if only dependencies are specified in manifest

### LINKED ISSUE #3364: `okteto deploy` shouldn't fail if manifest has `dependencies` but not `deploy` commands

AnshulAngaria:
# Proposed changes

Fixes #3364 

Okteto deploy fails if manifest only has dependencies section. This PR makes changes so that it does not fail.

### DIFF
--- a/cmd/deploy/deploy_test.go
+++ b/cmd/deploy/deploy_test.go
@@ -136,6 +136,25 @@ var fakeManifest *model.Manifest = &model.Manifest{
 	},
 }
 
+var fakeManifestWithDependency *model.Manifest = &model.Manifest{
+	Dependencies: model.ManifestDependencies{
+		"a": &model.Dependency{
+			Namespace: "b",
+		},
+		"b": &model.Dependency{},
+	},
+}
+
+var noDeployNorDependenciesManifest *model.Manifest = &model.Manifest{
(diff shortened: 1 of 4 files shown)


andreafalzetti (approved):
I've reviewed the change and tested different scenarios, LGTM! Thanks for your valuable contribution @AnshulAngaria 🙏

AdrianPedriza (approved):
LGTM

cmd/deploy/deploy_test.go
```diff
@@ -290,6 +309,45 @@ func TestDeployWithErrorReadingManifestFile(t *testing.T) {
 	assert.False(t, p.started)
 }
 
+func TestDeployWithNeitherDeployNorDependencyInManifestFile(t *testing.T) {
```
jLopezbarb, line 312: Really nice test 🚀

jLopezbarb (approved):
Code looks good and already tested!!

PR #3615, opened 2023-05-12
fix endpoints retrieval e2e

jLopezbarb:
# Proposed changes

Fixes https://github.com/okteto/app/issues/6484

- Use a wrapper so we don't break any unit tests
- Remove the use of the file and use the config that we need
- Add e2e test to check that the endpoints are working correctly


AdrianPedriza (approved):
LGTM

andreafalzetti (approved):

PR #3616, opened 2023-05-15
Change value for OktetoPipelineRunnerImage

teresaromero:
Related to https://github.com/okteto/app/issues/6395. 

# Proposed changes

Default OktetoPipelineRunnerImage was using the installer image instead of the pipeline runner image.
This fixes the default value.

### DIFF
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -60,7 +60,7 @@ const (
 	OktetoCLIImageForRemoteTemplate = "okteto/okteto:%s"
 
 	// OktetoPipelineRunnerImage defines image to use for remote deployments if empty
-	OktetoPipelineRunnerImage = "okteto/installer:1.8.9"
+	OktetoPipelineRunnerImage = "okteto/pipeline-runner:1.0.1"
 
 	// OktetoEnvFile defines the name for okteto env file
 	OktetoEnvFile = "OKTETO_ENV"


ifbyol:
It's not that simple. Doing this, we would lose some scenarios as pipeline runner image is missing some binaries that pipeline installer has. We should do what Pablo mentions in the ticket altogether (use pipeline runner image but getting the binaries from the installer image)

teresaromero:
> It's not that simple. Doing this, we would lose some scenarios as pipeline runner image is missing some binaries that pipeline installer has. We should do what Pablo mentions in the ticket altogether (use pipeline runner image but getting the binaries from the installer image)

By `getting the binaries from the installer` what is that meaning? Just to confirm i might/not understand correctly. 

https://github.com/okteto/okteto/blob/master/cmd/destroy/remote.go#L45 

The `OktetoCLIImage` here is used to extract the binary of okteto. By this change, you mean this binary should not be from `OktetoCLIImage` but from the installer used at the context?

teresaromero:
also, @ifbyol  can you please add the scenarios that would be broken with this change so i can have them in mind when testing?

maroshii:
@teresaromero I think the image we need to change is this one: https://github.com/okteto/okteto/blob/master/cmd/deploy/remote.go#L50

We are copying all the binaries from `/usr/local/bin/*` from the okteto-cli base image. That location is specific to the installer image and we need those binaries as well.

The pipeline-runner is a debian image so it might require other changes (at least `RUN apk update && apk add ca-certificates`)

maroshii:
> also, @ifbyol can you please add the scenarios that would be broken with this change so i can have them in mind when testing?

The installer image has `kubectl`, `kustomize`, `helm` etc and the pipeline-runner does not. I'm assuming we want the binaries from both images

ifbyol:
@teresaromero Exactly, as Fran mentioned, any scenario using `kubectl`, `helm` `kustomize`or any other binary we have in the pipeline-installer-image would be broken just changing the image to be used. Those binaries were the reason we changed at some point the pipeline-runner (it was the image originally used in the remote deploy) by the installer image (the currently one).

The main change requested in the linked ticket is to get the image version for the backend. Then, Pablo added the comment of doing the same we do in the installer, which is, copying the binaries from the installer image and use the runner as the base image for the remote commands.

So, we need to know both images from the API: installer and runner. And then, add the logic to copy the binaries

PR #3626, opened 2023-05-18
Use chart runner and installer images for remote deploy/destroy

teresaromero:
# Proposed changes

Resolves partially https://github.com/okteto/app/issues/6395

Currently when `okteto deploy --remote` and `okteto destroy --remote` the base image used by default was the installer.

This PR changes this, so the base image by default, is the pipeline runner. The binaries are taken from the installer, and okteto cli is overriden by the actual version that is running the command, not the installer.
The references of the images are now set by the response of the `metadata` query, which returns from the backend the values of the images to use.

**Refactor of GetDeployer**
There has been a refactor done in order to set the image to use for `Deploy.Image` as for the remote deploy command this was set way before remote flow was running, and for destroy was done at the remote flow.
Also, there is a fix under `GetDeployer` which determine if the deploy is remote or not. 3 states determine if the deployment should be remote:
- Based on the flag `--remote`  (if true, deploy should be remote)
- Based on env `OKTETO_REMOTE_CLI_IMAGE` which indicates if the flow is running inside a remote (if true, deploy should be local, over any other condition)
- Based on the `Deploy.Image` at the manifest, if exisit, deploy should be remote, even if there is no flag at the command

Additionally, unused `cwd` has been remove from this function.

## Testing done
In order to test the changes `go-getting-started` has been the repo used. Here the manifest does not have a Deploy.Image by default

- Running `okteto deploy --remote` with commands at the deploy manifest

```
# manifest.Deploy
deploy:
  commands:
    - okteto version
    - kustomize version
    - helm version
    - bazel version
    - kubectl apply -f k8s.yml


❯ ok deploy --remote --log-output plain
...
INFO: Running stage 'okteto version'
okteto version latest
INFO: Running stage 'kustomize version'
v5.0.0
INFO: Running stage 'helm version'
version.BuildInfo{Version:"v3.11.1", GitCommit:"293b50c65d4d56187cd4e2f390f0ada46b4c4737", GitTreeState:"clean", GoVersion:"go1.18.10"}
INFO: Running stage 'bazel version'
Extracting Bazel installation...
Build label: 4.2.2
Build target: bazel-out/k8-opt/bin/src/main/java/com/google/devtools/build/lib/bazel/BazelServer_deploy.jar
Build time: Thu Dec 2 18:15:58 2021 (1638468958)
Build timestamp: 1638468958
Build timestamp as int: 1638468958
INFO: Running stage 'kubectl apply -f k8s.yml'
deployment.apps/hello-world unchanged
service/hello-world unchanged
ingress.networking.k8s.io/hello-world configured
...
```

- Running `okteto deploy` with `manifest.Deploy.Image` runs the pipeline in remote mode with the base image set under the manifest
- Running `okteto deploy` without flag or image runs the process locally
- Running `okteto destroy --remote` with commands referencing the binaries, also work as expected
- Running remote deploys with an old backend not having the query implemented returns the default values present at the cli constants
- Running remote deploys when backend does not return the images triggers an error

### DIFF
--- a/cmd/deploy/remote.go
+++ b/cmd/deploy/remote.go
@@ -125,7 +125,7 @@ func newRemoteDeployer(builder *buildv2.OktetoBuilder) *remoteDeployCommand {
 
 func (rd *remoteDeployCommand) deploy(ctx context.Context, deployOptions *Options) error {
 
-	if deployOptions.Manifest.Deploy.Image == "" {
+	if deployOptions.Manifest != nil && deployOptions.Manifest.Deploy != nil && deployOptions.Manifest.Deploy.Image == "" {
 		deployOptions.Manifest.Deploy.Image = constants.OktetoPipelineRunnerImage
 	}
 
--- a/cmd/deploy/remote.go
+++ b/cmd/deploy/remote.go
@@ -124,6 +124,11 @@ func newRemoteDeployer(builder *buildv2.OktetoBuilder) *remoteDeployCommand {
 }
 
 func (rd *remoteDeployCommand) deploy(ctx context.Context, deployOptions *Options) error {
+
+	if deployOptions.Manifest.Deploy.Image == "" {
+		deployOptions.Manifest.Deploy.Image = constants.OktetoPipelineRunnerImage
+	}
+
 	cwd, err := rd.getOriginalCWD(deployOptions.ManifestPathFlag)
 	if err != nil {
 		return err
--- a/cmd/deploy/remote.go
+++ b/cmd/deploy/remote.go
@@ -43,7 +43,7 @@ import (
 
 const (
 	templateName           = "dockerfile"
-	dockerfileTemporalNane = "deploy"
+	dockerfileTemporalName = "deploy"
 	oktetoDockerignoreName = ".oktetodeployignore"
 	dockerfileTemplate     = `
 FROM {{ .OktetoCLIImage }} as okteto-cli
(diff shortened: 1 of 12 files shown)


cmd/deploy/remote.go
```diff
@@ -124,6 +128,16 @@ func newRemoteDeployer(builder *buildv2.OktetoBuilder) *remoteDeployCommand {
...
 
 func (rd *remoteDeployCommand) deploy(ctx context.Context, deployOptions *Options) error {
+
+	sc, err := rd.clusterMetadata(ctx)
+	if err != nil {
+		return err
+	}
+
+	if deployOptions.Manifest != nil && deployOptions.Manifest.Deploy != nil && deployOptions.Manifest.Deploy.Image == "" {
+		deployOptions.Manifest.Deploy.Image = sc.PipelineRunnerImage
```
maroshii, line 138: Unless I'm missing something, I don't think this change is backwards compatible. With an old backend, `sc` will be `types.ClusterMetadata{}` and no image will be defined there.

teresaromero: yup, i was asking this yesteday, what would the behavour be desired in this case. this is pending to solve right now

teresaromero: updated.

>If the query doesn’t exist in the backend, we should fallback to a harcoded version of the images in the CLI
If the query returns an empty value for the images (because the backend is not returning them), we should fail

PR #3635, opened 2023-05-19
remove secret content from build hash

### LINKED ISSUE #3622: Build hash should not check secret content to create it
**Describe the feature**
When we are generating the [build hash](https://github.com/okteto/okteto/blob/master/cmd/build/v2/config.go#L96). We should only check the key value and not the content of the secret, since buildkit is just checking the key/value to use the cache.

**Expected behavior**
I want that the build hash is generated only with the key value of secrets, without checking the content of secret file.

sankalp-r:
# Proposed changes

Fixes #3622 

### DIFF
--- a/cmd/build/v2/config.go
+++ b/cmd/build/v2/config.go
@@ -93,7 +93,12 @@ func (oc oktetoBuilderConfig) getTextToHash(buildInfo *model.BuildInfo, sha stri
 		args = append(args, arg.String())
 	}
 	argsText := strings.Join(args, ";")
-	secretsText := buildInfo.Secrets.GetContent(oc.fs)
+
+	secrets := []string{}
+	for key, value := range buildInfo.Secrets {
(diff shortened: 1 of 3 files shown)


jLopezbarb:
@sankalp-r I tested it and it's working perfectly 😉. Thanks for the addition

jLopezbarb (approved):

cmd/build/v2/config.go
```diff
@@ -93,7 +93,12 @@ func (oc oktetoBuilderConfig) getTextToHash(buildInfo *model.BuildInfo, sha stri
 		args = append(args, arg.String())
 	}
 	argsText := strings.Join(args, ";")
-	secretsText := buildInfo.Secrets.GetContent(oc.fs)
```
AdrianPedriza, line 96: I think this func is not needed anymore. Can we remove it?

PR #3639, opened 2023-05-19
Add destinationURL.Host to request Host at proxy handler

teresaromero:
# Proposed changes

Fixes https://github.com/okteto/app/issues/6356


cmd/deploy/proxy.go
```diff
@@ -255,6 +255,7 @@ func (ph *proxyHandler) getProxyHandler(token string, clusterConfig *rest.Config
 			r.Body = io.NopCloser(bytes.NewBuffer(b))
 		}
 
+		r.Host = destinationURL.Host
```
maroshii, line 258: Check line 242. We are proxying differently for PUT and POST so we might need to move this to line 231

teresaromero: yes you are right, changed it

maroshii (approved):

PR #3648, opened 2023-05-22
fix: performance issues on `git status`

andreafalzetti:
# Proposed changes

Fixes https://github.com/okteto/app/issues/6518

We've noticed a huge delay in the function that gets the repo status using the golang library `git-go` so we've decided to implement a different mechanism. If the host running the CLI has `git` installed, we will use `git status` and parse the output. If git is not installed, we rely on the library. However, if the library takes more than 1 second, we assume the commit is not clean and we try to build which if already built it should be quick.

## How to test

1. Using a large repo, try running `okteto deploy` and observe how slow it can be between the messages ` i  Using <okteto user> @ <okteto url> as context` and `⠋ Checking images to build...`
2. Try it again using the CLI built with this branch and observe the difference in time

### DIFF
--- a/pkg/repository/git.go
+++ b/pkg/repository/git.go
@@ -70,7 +70,6 @@ func (r gitRepoController) isClean(ctx context.Context) (bool, error) {
 		}
 
 		ch <- cleanStatus{status.IsClean(), nil}
-		return
 	}()
 
 	select {
@@ -186,7 +185,7 @@ func (ogr oktetoGitWorktree) Status(ctx context.Context) (oktetoGitStatus, error
 		return oktetoGitStatus{status: git.Status{}}, fmt.Errorf("failed to get git status: %w", err)
 	}
 
-	lines := strings.Split(string(output), "\000")
+	lines := strings.Split(output, "\000")
 	stat := make(map[string]*git.FileStatus, len(lines))
 	for _, line := range lines {
 		// line example values can be: "M modified-file.go", "?? new-file.go", etc
--- a/pkg/repository/git.go
+++ b/pkg/repository/git.go
@@ -17,6 +17,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	oktetoLog "github.com/okteto/okteto/pkg/log"
 	"os/exec"
 	"strings"
 	"time"
@@ -36,7 +37,7 @@ func newGitRepoController() gitRepoController {
 	}
 }
 
-type CleanStatus struct {
+type cleanStatus struct {
 	IsClean bool
 	Err     error
 }
@@ -46,28 +47,35 @@ func (r gitRepoController) isClean(ctx context.Context) (bool, error) {
 	ctxWithTimeout, cancel := context.WithTimeout(ctx, 1*time.Second)
 	defer cancel()
 
-	ch := make(chan CleanStatus)
+	ch := make(chan cleanStatus)
+	defer close(ch)
 
 	go func() {
 		repo, err := r.repoGetter.get(r.path)
 		if err != nil {
-			ch <- CleanStatus{false, fmt.Errorf("failed to analyze git repo: %w", err)}
+			ch <- cleanStatus{false, fmt.Errorf("failed to analyze git repo: %w", err)}
+			return
 		}
+
 		worktree, err := repo.Worktree()
 		if err != nil {
-			ch <- CleanStatus{false, fmt.Errorf("failed to infer the git repo's current branch: %w", err)}
+			ch <- cleanStatus{false, fmt.Errorf("failed to infer the git repo's current branch: %w", err)}
+			return
 		}
 
-		status, err := worktree.Status()
+		status, err := worktree.Status(ctx)
 		if err != nil {
-			ch <- CleanStatus{false, fmt.Errorf("failed to infer the git repo's status: %w", err)}
+			ch <- cleanStatus{false, fmt.Errorf("failed to infer the git repo's status: %w", err)}
+			return
 		}
 
-		ch <- CleanStatus{status.IsClean(), nil}
+		ch <- cleanStatus{status.IsClean(), nil}
+		return
 	}()
 
 	select {
 	case <-ctxWithTimeout.Done():
+		oktetoLog.Warning("Timeout exceeded calculating git status: assuming dirty commit")
 		return false, ctxWithTimeout.Err()
 	case res := <-ch:
 		return res.IsClean, res.Err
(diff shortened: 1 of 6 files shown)


pkg/repository/git.go
```diff
@@ -113,12 +129,55 @@ type oktetoGitWorktree struct {
...
+		}
+		fixAttempt++
+		return runGitStatusCommand(path, fixAttempt)
+	}
+
+	return string(output), err
+}
+
+func (ogr oktetoGitWorktree) Status() (oktetoGitStatus, error) {
+	output, err := runGitStatusCommand(ogr.GetRoot(), 0)
```
andreafalzetti, line 163: the first iteration removed the old logic, but I am considering keeping the git status from `git-go` so that we can keep the CLI backward compatible (ie. do not require the user/host to have the `git` binary installed).

The logic would be:

- try find `git`, if present, then use the new fast approach
- if `git` is missing, then fallback to `git-go`

pkg/repository/repository.go
```diff
@@ -53,7 +53,7 @@ func NewRepository(path string) Repository {
 	}
 	return Repository{
 		path:    path,
-		url:     url,
+		url:     repoUrl,
```
andreafalzetti, line 56: there is a lib imported called `url`, which could overlap with this var.

pkg/repository/git.go
jLopezbarb: Let's do that and go with the fallback. Also add a log (`oktetoLog.Infof()`) informing that we are using the old method then

pkg/repository/git.go
```diff
@@ -113,12 +129,55 @@ type oktetoGitWorktree struct {
...
+	lines := strings.Split(string(output), "\000")
+	stat := make(map[string]*git.FileStatus, len(lines))
+	for _, line := range lines {
+		// line example values can be: "M modified-file.go", "?? new-file.go", etc
+		parts := strings.SplitN(strings.TrimLeft(line, " "), " ", 2)
+		if len(parts) == 2 {
+			stat[strings.Trim(parts[1], " ")] = &git.FileStatus{
+				Staging: git.StatusCode([]byte(parts[0])[0]),
+			}
+		}
```
jLopezbarb, line 177: should we return an error `if len(parts) != 2`

pkg/repository/git.go
```diff
@@ -33,29 +36,42 @@ func newGitRepoController() gitRepoController {
 	}
 }
 
+type CleanStatus struct {
```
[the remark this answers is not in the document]
andreafalzetti: no need for now, I'll rename it

pkg/repository/git.go
```diff
@@ -113,12 +129,55 @@ type oktetoGitWorktree struct {
...
+		// line example values can be: "M modified-file.go", "?? new-file.go", etc
+		parts := strings.SplitN(strings.TrimLeft(line, " "), " ", 2)
+		if len(parts) == 2 {
+			stat[strings.Trim(parts[1], " ")] = &git.FileStatus{
+				Staging: git.StatusCode([]byte(parts[0])[0]),
+			}
+		}
+	}
+
+	return oktetoGitStatus{status: stat}, err
```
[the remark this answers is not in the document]
andreafalzetti: nope, I will make it explicit

pkg/repository/repository_test.go
```diff
@@ -23,12 +25,18 @@ import (
...
 
 type fakeRepositoryGetter struct {
-	repository *fakeRepository
-	err        error
+	repository []*fakeRepository
+	err        []error
+	callCount  int
 }
 
-func (frg fakeRepositoryGetter) get(_ string) (gitRepositoryInterface, error) {
```
andreafalzetti, line 30: This is required because we call this function several times and we need to mock a sequennce of responses/errors instead of a mock value all the time

pkg/repository/repository_test.go
```diff
@@ -23,12 +25,18 @@ import (
...
-	return frg.repository, frg.err
+func (frg *fakeRepositoryGetter) get(_ string) (gitRepositoryInterface, error) {
```
andreafalzetti, line 33: This is required because we call this function several times and we need to mock a sequennce of responses/errors instead of a mock value all the time

pkg/repository/git_test.go
```diff
@@ -151,28 +203,49 @@ func TestGetSHA(t *testing.T) {
...
 				err: nil,
 			},
 		},
 		{
 			name: "error getting repository",
 			config: config{
-				repositoryGetter: fakeRepositoryGetter{
-					repository: nil,
-					err:        assert.AnError,
+				repositoryGetter: &fakeRepositoryGetter{
```
andreafalzetti, line 244: I had to refactor some stuff because this test and the one below were not working properly. They would pass but not for the reason we thought they would.

They were actualky the same code (expecting to test different things) which is a bit odd! 

They were both failing early in the function so not reaching their respective logic.

It should be fixed now!

pkg/repository/git.go
```diff
@@ -33,29 +34,47 @@ func newGitRepoController() gitRepoController {
...
+type cleanStatus struct {
+	IsClean bool
+	Err     error
+}
+
 // IsClean checks if the repository have changes over the commit
 func (r gitRepoController) isClean(ctx context.Context) (bool, error) {
 	ctxWithTimeout, cancel := context.WithTimeout(ctx, 1*time.Second)
 	defer cancel()
-	select {
```
andreafalzetti, line 40: the previous timeout was ineffective because all function calls below did not actually use this context and they were not in a coroutine either so they would just execute and the timeout would kick in but without any effect.

pkg/repository/git.go
```diff
@@ -33,29 +34,47 @@ func newGitRepoController() gitRepoController {
...
 }
 
+type cleanStatus struct {
+	IsClean bool
+	Err     error
+}
+
 // IsClean checks if the repository have changes over the commit
 func (r gitRepoController) isClean(ctx context.Context) (bool, error) {
 	ctxWithTimeout, cancel := context.WithTimeout(ctx, 1*time.Second)
```
andreafalzetti, line 44: the new logic keeps having a timeout for safety but the local execution of `git status` should be fairly reliable, however in case the bin of `git` is not installed on host, we fall back on the old logic using `git-go`. In that case, it will be slower (I've seen it take over 20 seconds). In that case the timeout will be useful.

pkg/repository/git_test.go
jLopezbarb: Thanks!!

pkg/repository/local_git_test.go
```diff
@@ -0,0 +1,166 @@
...
+	tests := []struct {
+		name     string
+		mockExec func() *mockLocalExec
+		err      error
+	}{
+		{
+			name: "success",
+			mockExec: func() *mockLocalExec {
+				return &mockLocalExec{
+					runCommand: func(ctx context.Context, dir string, name string, arg ...string) ([]byte, error) {
```
jLopezbarb, line 81:
```suggestion
					runCommand: func(_ context.Context, _ string, _ string, _ ...string) ([]byte, error) {
```

jLopezbarb (approved):

pkg/repository/git.go
```diff
@@ -33,29 +34,47 @@ func newGitRepoController() gitRepoController {
 	}
 }
 
+type cleanStatus struct {
+	IsClean bool
+	Err     error
```
AdrianPedriza, line 39: I think we can keep fields private

pkg/repository/local_git.go
```diff
@@ -0,0 +1,105 @@
...
+}
+
+func NewLocalGit(gitPath string, exec CommandExecutor) *LocalGit {
+	return &LocalGit{
+		gitPath: gitPath,
+		exec:    exec,
+	}
+}
+
+func (lg *LocalGit) Status(ctx context.Context, dirPath string, fixAttempt int) (git.Status, error) {
```
AdrianPedriza, line 54: instead of run `git status` locally how about call graphql github API in order to get repo status and to not have two implementations depending of if user has `git` installed or not? It is feasible?

pkg/repository/git.go
andreafalzetti: done in https://github.com/okteto/okteto/pull/3648/commits/d61f715e597663ba5a3903089869e803400b4277

pkg/repository/local_git.go
AdrianPedriza, line 54: can we add a description for this func?

pkg/repository/local_git.go
```diff
@@ -0,0 +1,105 @@
...
+
+	status, err := lg.parseGitStatus(string(output))
+	if err != nil {
+		return git.Status{}, err
+	}
+
+	return status, err
+}
+
+func (lg *LocalGit) FixDubiousOwnershipConfig(path string) error {
```
AdrianPedriza, line 79: can we add a description for this func?

pkg/repository/local_git.go
```diff
@@ -0,0 +1,105 @@
...
+	_, err := lg.exec.RunCommand(context.Background(), lg.gitPath, "config", "--global", "--add", "safe.directory", path)
+	return err
+}
+
+func (lg *LocalGit) Exists() (string, error) {
```
AdrianPedriza, line 84: can we add a description for this func?

andreafalzetti: done in 0bbf828

PR #3649, opened 2023-05-22
Rewrite Host header in okteto deploy proxy

maroshii:
If `cluster.endpoint` is not defined the kube server is expose at `kubernetes.<subdomain>`. Since we are running behind the ingress we need to set the `Host` header so that we can route requests to the appropriate backend, in this case the kube service. The `NewSingleHostReverseProxy` doesn't rewrite the Host header by default so we need to do it ourselves.

This errors happens in staging and **WILL** happen in cloud.


ifbyol:
I think this is already done in https://github.com/okteto/okteto/pull/3639.

maroshii:
closing in favor of #3639

PR #3655, opened 2023-05-23
Fix RC in divert (weaver) when the cache has outdated information

pchico83:
Some users have reported RC with the divert implementation. I was able to reproduce this by having outdated data in the cache. 
This PR refreshes the info in the catch when a create/update operation fails because the object has been modified.
I updated the tests to simulate this scenario

### DIFF
--- a/pkg/divert/weaver/divert_test.go
+++ b/pkg/divert/weaver/divert_test.go
@@ -679,4 +679,55 @@ func Test_divertIngresses(t *testing.T) {
 	if !reflect.DeepEqual(expectedE3, resultE3) {
 		t.Fatalf("Didn't compute e3 correctly: \n%v\n%v", expectedE3, resultE3)
 	}
+
+	// Eliminate elements from the cache to force RCs
+	d.cache.developerIngresses = map[string]*networkingv1.Ingress{}
+	d.cache.developerServices = map[string]*apiv1.Service{}
+	d.cache.developerEndpoints = map[string]*apiv1.Endpoints{}
+	d.Deploy(ctx)
+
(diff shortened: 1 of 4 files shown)


jLopezbarb (approved):

pkg/divert/weaver/divert_test.go
```diff
@@ -679,4 +679,55 @@ func Test_divertIngresses(t *testing.T) {
...
 	}
+
+	// Eliminate elements from the cache to force RCs
+	d.cache.developerIngresses = map[string]*networkingv1.Ingress{}
+	d.cache.developerServices = map[string]*apiv1.Service{}
+	d.cache.developerEndpoints = map[string]*apiv1.Endpoints{}
+	d.Deploy(ctx)
+
+	resultI1, _ = c.NetworkingV1().Ingresses("cindy").Get(ctx, "i1", metav1.GetOptions{})
+	if !reflect.DeepEqual(expectedI1, resultI1) {
```
jLopezbarb, line 690: Can we replace all the reflect.DeepEqual with assert.NotEqual? In case it fails it provides a much better error

AdrianPedriza (approved):

pkg/divert/weaver/divert_test.go
pchico83: @jLopezbarb  `assert` isn't working for me, `reflect.DeepEqual` checks each value in the whole object, should ` assert.NotEqual` work the same way?

jLopezbarb: assert uses `reflect.DeepEqual` inside, I don't know why it's not working 🤔

pchico83: @jLopezbarb It works! I was using a different `assert` library :-)

PR #3658, opened 2023-05-24
fix(build): vanilla cluster should not reach out to okteto registry

andreafalzetti:
# Proposed changes

This issue came up while testing the `2.16.x` release. When working with vanilla clusters, the CLI will try to reach out to the Okteto Registry and the build fails.

| Before | After |
|--------|-------|
|   before  |  after  |

## How to test
2. Build this branch and run against the repo "okteto build"
3. Observe how the validation kicks in, instead of a weird connection issue

### DIFF
--- a/cmd/build/v2/tagger.go
+++ b/cmd/build/v2/tagger.go
@@ -15,6 +15,7 @@ package v2
 
 import (
 	"fmt"
+	"github.com/okteto/okteto/pkg/okteto"
 
 	"github.com/okteto/okteto/pkg/constants"
 	"github.com/okteto/okteto/pkg/format"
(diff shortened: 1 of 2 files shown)


cmd/build/v2/tagger.go
```diff
@@ -35,7 +36,15 @@ type imageWithVolumesTagger struct {
 	cfg oktetoBuilderConfigInterface
 }
 
-var targetRegistries = []string{constants.GlobalRegistry, constants.DevRegistry}
+func getTargetRegistries() []string {
```
jLopezbarb, line 39: Should we add a test for this?

jLopezbarb (approved):

AdrianPedriza (approved):

cmd/build/v2/tagger.go
andreafalzetti: Added!

PR #3666, opened 2023-05-26
add local envs for hybrid command

### LINKED ISSUE #3651: hybrid dev command output loses the original output colors
**Describe the bug**
if you run a command that generates output with colors, at the beginning these colors related to the executed command are present but the next ones don't have colors.
### Current output
Screenshot 2023-05-22 at 12 58 30

**To Reproduce**

1. Add `yarn start` to dev exec command using hybrid mode
1. See colors are present
1. See appended output lost the colors

**Expected behavior**
Screenshot 2023-05-09 at 10 26 29

AdrianPedriza:
# Proposed changes

Fixes #3651

- add local envs to cmd hybrid command
- local envs will have the highest preference if one of its variables is already contained in the config map, image used or in the secrets.

### DIFF
--- a/cmd/up/exec.go
+++ b/cmd/up/exec.go
@@ -157,7 +157,7 @@ type envsGetter struct {
 	configMapEnvsGetter configMapEnvsGetterInterface
 	secretsEnvsGetter   secretsEnvsGetterInterface
 	imageEnvsGetter     imageEnvsGetterInterface
-	pathGetter          func(string) string
+	getLocalEnvs        func() []string
 }
 
 func newEnvsGetter(hybridCtx *HybridExecCtx) (*envsGetter, error) {
@@ -183,7 +183,7 @@ func newEnvsGetter(hybridCtx *HybridExecCtx) (*envsGetter, error) {
 		imageEnvsGetter: &imageEnvsGetter{
 			imageGetter: registry.NewOktetoRegistry(okteto.Config{}),
 		},
-		pathGetter: os.Getenv,
+		getLocalEnvs: os.Environ,
 	}, nil
 }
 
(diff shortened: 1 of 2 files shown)


cmd/up/exec.go
```diff
@@ -183,7 +183,7 @@ func newEnvsGetter(hybridCtx *HybridExecCtx) (*envsGetter, error) {
 		imageEnvsGetter: &imageEnvsGetter{
 			imageGetter: registry.NewOktetoRegistry(okteto.Config{}),
 		},
-		pathGetter: os.Getenv,
+		getLocalEnvs: os.Environ,
```
jLopezbarb, line 186: Do we really need all of them? This can cause that there might be some bugs that are hidden because local has the right env vars. I'd limit it to as few as possible

AdrianPedriza: ok, I will add the needed one to solve this issue

AdrianPedriza: 3f6c57b193ef7489324e42b0a318e486ee059ae0

jLopezbarb (approved):

jLopezbarb:
Can you post an screenshoot on how this looks like with the reproduction steps on the issue?

cmd/up/exec.go
```diff
@@ -242,14 +242,27 @@ func (eg *envsGetter) getEnvs(ctx context.Context) ([]string, error) {
...
 
+func getDefaultLocalEnvs() []string {
+	var envs []string
+
+	path := os.Getenv("PATH")
+	if path != "" {
+		envs = append(envs, fmt.Sprintf("PATH=%s", path))
+	}
+
+	term := os.Getenv("TERM")
```
andreafalzetti, line 258: @AdrianPedriza is `TERM` enough to fix colors, etc?

I ask because in other projects I've seen both `TERM` and `COLORTERM` are used for this

andreafalzetti (approved):

AdrianPedriza:
@andreafalzetti @jLopezbarb this is the output using this PR 
Captura de Pantalla 2023-05-29 a las 12 52 43
I only add `TERM` to fix this issue. I've `COLORTERM` in my system but seems to be not needed for this purpose

PR #3667, opened 2023-05-26
Mask variables at destroy

### LINKED ISSUE #658: Support for "okteto rescan"
Sometimes my syncthing indexes get corrupted. The only option to solve this problem now is to run `okteto down -v`.  

`okteto rescan` would compute syncthing indexes on demand (by default, this is done each 5min).

teresaromero:
# Proposed changes

Fixes #658

### DIFF
--- a/cmd/destroy/local.go
+++ b/cmd/destroy/local.go
@@ -2,6 +2,8 @@ package destroy
 
 import (
 	"context"
+	"encoding/base64"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"os"
@@ -19,6 +21,7 @@ import (
 	oktetoLog "github.com/okteto/okteto/pkg/log"
 	"github.com/okteto/okteto/pkg/model"
 	"github.com/okteto/okteto/pkg/okteto"
+	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/selection"
 )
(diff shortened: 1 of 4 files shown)


cmd/destroy/local.go
```diff
@@ -59,20 +62,35 @@ func (ld *localDestroyCommand) destroy(ctx context.Context, opts *Options) error
...
+	if !ok {
+		return nil
+	}
+
+	decodedStringVariables, err := base64.StdEncoding.DecodeString(b64Variables)
+	if err != nil {
+		return nil
+	}
+
+	var variables []model.EnvVar
```
teresaromero, line 80: this should be a dedicated Variables model

cmd/destroy/cfghandler.go
```diff
@@ -31,6 +31,7 @@ type configMapHandler interface {
 	translateConfigMapAndDeploy(context.Context, *pipeline.CfgData) (*apiv1.ConfigMap, error)
 	destroyConfigMap(context.Context, *apiv1.ConfigMap, string) error
 	setErrorStatus(context.Context, *apiv1.ConfigMap, *pipeline.CfgData, error) error
+	getConfigmap(ctx context.Context, name, namespace string) (*apiv1.ConfigMap, error)
```
jLopezbarb, line 34: I'd go with getting directly the variables, that way we abstract the get variables instead of retrieving the cmap and the if exists retrieving the vars. Same way we translate and deploy or set the error status. I don't like the fact that the deploy function needs to know that `getcmap` can return `nil, nil` WDYT?

ifbyol:
Could we update the PR description to match with the purpose of it?

teresaromero:
> Could we update the PR description to match with the purpose of it?

@ifbyol done

pkg/cmd/pipeline/translate.go
```diff
@@ -87,6 +89,21 @@ type CfgData struct {
 	Filename   string
 	Manifest   []byte
 	Icon       string
+	Variables  []string
+}
+
+// GetConfigmapVariables returns Data["variables"] content from Configmap
+func GetConfigmapVariables(ctx context.Context, name, namespace string, c kubernetes.Interface) (string, error) {
```
AdrianPedriza, line 96:
```suggestion
func GetConfigmapVariablesEncoded(ctx context.Context, name, namespace string, c kubernetes.Interface) (string, error) {
```

AdrianPedriza: using `GetConfigmapVariables` I expect a list of variables instead of string

jLopezbarb (approved):

maroshii (approved):

AdrianPedriza (approved):

PR #3677, opened 2023-05-30
fix send on closed channel panic in isClean

maroshii:
Fix `send on closed channel` error that occurs when we exceed the timeout for git clean status.

This issue can be reproduced locally by changing the `ctxWithTimeout` deadline to 1 millisecond [here](https://github.com/okteto/okteto/blob/master/pkg/repository/git.go#L44).

```
okteto build --progress plain
 !  Calculating git status: git is not installed, for better performances consider installing it
 !  Timeout exceeded calculating git status: assuming dirty commit
 i  Building ‘Dockerfile.okteto’ in tcp://buildkit.qwerty.okteto.dev:443...
panic: send on closed channelgoroutine 23 [running]:
github.com/okteto/okteto/pkg/repository.gitRepoController.isClean.func1()
  github.com/okteto/okteto/pkg/repository/git.go:69 +0x299
created by github.com/okteto/okteto/pkg/repository.gitRepoController.isClean
  github.com/okteto/okteto/pkg/repository/git.go:50 +0x1a5
```

This happens because when we hit the timeout the function returns and we call `defer close(ch)`.  However, the internal go routines is still active. Eventually the goroutine tries to write to `ch` but it's closed so it panics.

The fix is making sure we race `ch` against a `timeoutCh` and write to `ch` in a single place for readability (hence the new `calculateIsClean` func).

One other thing that changed, is that we are now passing a context with a timeout to `worktree.Status`. Before we were passing the top level function context. Not sure if this has any actual effect but it's worth mentioning since this could've been the original issue.

I didn't get into the nitty gritty of preventing `calculateIsClean` and children from blocking forever and assumed that the context deadline is respected. We are passing down the right context with a 1s timeout to `worktree.Status` so if that works it should be ok. Otherwise the goroutine could leak (if calculateIsClean blocks forever).

### DIFF
--- a/pkg/repository/git.go
+++ b/pkg/repository/git.go
@@ -15,9 +15,11 @@ package repository
 
 import (
 	"context"
+	"errors"
 	"fmt"
-	oktetoLog "github.com/okteto/okteto/pkg/log"
 	"time"
+
+	oktetoLog "github.com/okteto/okteto/pkg/log"
 
 	"github.com/go-git/go-git/v5"
 	"github.com/go-git/go-git/v5/plumbing"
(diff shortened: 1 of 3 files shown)


jLopezbarb (approved):

andreafalzetti (approved):

andreafalzetti:
converting to draft while we work on the unit tests

andreafalzetti:
ready to be merged, waiting for CI

jLopezbarb (approved):

PR #3683, opened 2023-06-01
gracefull shutdown for hybrid process

### LINKED ISSUE #3652: process running locally is not properly cleaned up if connection is lost
**Describe the bug**
using hybrid mode if the connection is lost we attempt to connect again with the remote and execute the command again (locally). If local process use a port to serve some kind of service the reconnection logic will execute the command again but it will fail because the port is already in use.
Screenshot 2023-05-23 at 10 15 55

**To Reproduce**

1. Run hybrid mode with command `yarn start` 
1. simulate bad network conditions
1. okteto up logic will run again but port is already in use

**Expected behavior**
We should gracefully shutdown the local process in order to be able to use same port if reconnection is needed.

AdrianPedriza:
# Proposed changes
- add gracefull shutdown for local process if hybrid mode is running.
- close hybrid process and childs.

Fixes #3652

### DIFF
--- a/cmd/up/exec_unix.go
+++ b/cmd/up/exec_unix.go
@@ -1,0 +1,41 @@
+//go:build !windows
+// +build !windows
+
+package up
+
+import (
+	"context"
+	"os"
+	"os/exec"
+	"runtime"
+	"strings"
+	"syscall"
+)
+
+func (he *hybridExecutor) GetCommandToExec(ctx context.Context, cmd []string) (*exec.Cmd, error) {
+	var c *exec.Cmd
+	if runtime.GOOS != "windows" {
+		c = exec.Command("bash", "-c", strings.Join(cmd, " "))
+	} else {
+		binary, err := expandExecutableInCurrentDirectory(cmd[0], he.workdir)
+		if err != nil {
+			return nil, err
+		}
+		c = exec.Command(binary, cmd[1:]...)
+	}
+
+	c.Env = he.envs
+
+	c.Stdin = os.Stdin
+	c.Stdout = os.Stdout
+	c.Stderr = os.Stderr
+
+	c.Dir = he.workdir
+
+	c.SysProcAttr = &syscall.SysProcAttr{
+		Setpgid:    true,
+		Foreground: true,
+	}
+
+	return c, nil
+}
--- a/cmd/up/exec_unix.go
+++ b/cmd/up/exec_unix.go
@@ -34,8 +34,8 @@ func (he *hybridExecutor) GetCommandToExec(ctx context.Context, cmd []string) (*
 
 	c.SysProcAttr = &syscall.SysProcAttr{
 		Setpgid:    true,
+		Foreground: true,
 	}
-	c.SysProcAttr.Foreground = true
 
 	return c, nil
 }
--- a/cmd/up/exec_unix.go
+++ b/cmd/up/exec_unix.go
@@ -34,8 +34,8 @@ func (he *hybridExecutor) GetCommandToExec(ctx context.Context, cmd []string) (*
 
 	c.SysProcAttr = &syscall.SysProcAttr{
 		Setpgid:    true,
-		Foreground: true,
 	}
+	c.SysProcAttr.Foreground = true
 
 	return c, nil
 }
--- a/cmd/up/exec_windows.go
+++ b/cmd/up/exec_windows.go
@@ -1,0 +1,32 @@
+//go:build windows
+// +build windows
+
+package up
+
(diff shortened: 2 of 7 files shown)


cmd/up/up.go
```diff
@@ -904,11 +913,54 @@ func (up *upContext) shutdown() {
...
+
 	oktetoLog.Info("completed shutdown sequence")
 	up.ShutdownCompleted <- true
 
 }
 
+func shutdownHybridMode(pid int) {
+	if existProcess(pid) {
+		if err := syscall.Kill(-pid, syscall.SIGINT); err != nil {
+			oktetoLog.Warning("failed to stop gracefully local process related to command executed in hybrid mode: %s", err.Error())
```
viddo, line 929: Could we output what the commands were and some CTA to terminate them manually? Might also need to inform that not terminating might cause spurious errors on next `okteto` usages.

rlamana: That's when the user specifies the command, but a common scenario is also to have an interactive terminal (using `bash` or whatever), what would we show in that case?

cmd/up/exec_unix.go
```diff
@@ -0,0 +1,41 @@
+//go:build !windows
+// +build !windows
```
andreafalzetti, line 2: This file is going to be ignored when built for Windows using this directive, however below at line 17, we check for Windows - perhaps this line is a left over of previous attempts?

cmd/up/exec_unix.go
```diff
@@ -0,0 +1,41 @@
...
+	"os/exec"
+	"runtime"
+	"strings"
+	"syscall"
+)
+
+func (he *hybridExecutor) GetCommandToExec(ctx context.Context, cmd []string) (*exec.Cmd, error) {
+	var c *exec.Cmd
+	if runtime.GOOS != "windows" {
+		c = exec.Command("bash", "-c", strings.Join(cmd, " "))
```
andreafalzetti, line 18: I believe we're not passing the `ctx` to `exec.Command` to avoid sending `SIGKILL` and handle gracefully the shutdown of child processes?

cmd/up/exec_windows.go
```diff
@@ -0,0 +1,40 @@
...
+
+	c.Env = he.envs
+
+	c.Stdin = os.Stdin
+	c.Stdout = os.Stdout
+	c.Stderr = os.Stderr
+
+	c.Dir = he.workdir
+
+	c.SysProcAttr = &syscall.SysProcAttr{
```
andreafalzetti, line 35: I am wondering if we couldn't make `c.SysProcAttr` conditional based on the OS, and remove the double file.

cmd/up/up.go
```diff
@@ -904,11 +916,66 @@ func (up *upContext) shutdown() {
...
+
+func existProcess(pid int) bool {
+	found, err := ps.FindProcess(pid)
+	if err == nil && found == nil {
+		return false
+	}
+	return true
+}
+
+func hasHybridCommandFinishes(pid int, timeout time.Duration) bool {
```
andreafalzetti, line 963:
```suggestion
func hasHybridCommandFinished(pid int, timeout time.Duration) bool {
```

It sounds like a typo?

cmd/up/up.go
```diff
@@ -904,11 +916,66 @@ func (up *upContext) shutdown() {
...
+func (up *upContext) shutdownHybridMode() {
+	if up.hybridCommand != nil {
+		pid := up.hybridCommand.Process.Pid
+		if existProcess(pid) {
+			if err := killProcess(pid); err != nil {
+				oktetoLog.Warning("failed to stop gracefully local process related to command executed in hybrid mode: %s", err.Error())
+			}
+
+			waitTimeout := 15 * time.Second
+			hasFinished := hasHybridCommandFinishes(pid, waitTimeout)
```
andreafalzetti, line 938:
```suggestion
			hasFinished := hasHybridCommandFinished(pid, waitTimeout)
```

typo?

andreafalzetti (approved):
I have not tested it, only code reviwed and besides a couple of comments, LGTM!

cmd/up/exec_windows.go
teresaromero: i think this can be done at runtime but not on compilation. The error appearing was because when building for windows the compiler did not found this function for the platform...

andreafalzetti: Not blocking comment, just a thought

andreafalzetti: I see, thanks @teresaromero

PR #3688, opened 2023-06-01
Fix variables deploy installer

teresaromero:
# Proposed changes

When introducing the variables for deploy command at the configmap, there was no condition to check if this was running at the installer or remote or locally.

This fixes the case when the variables where removed when the pipeline was running the deploy.

Tests done:

- Deploy locally with variables, the configmap include them
- Redeploy From UI, the configmap still has them, they are not removed
- Deploy remote with variables, the confirmap include them


maroshii (approved):

jLopezbarb (approved):

cmd/deploy/deploy.go
```diff
@@ -587,3 +605,18 @@ func (dc *DeployCommand) recreateFailedPods(ctx context.Context, name string) er
 	}
 	return nil
 }
+
+// getVariablesFromCfgmap given a cfgmap this returns the variables as []EnvVar in it
+func decodeConfigMapVariables(stringVariables string) []types.DeployVariable {
```
jLopezbarb, line 610: I don't see a test for this. Should we added?

AdrianPedriza (approved):
LGTM

cmd/deploy/deploy.go
teresaromero: i've moved the func to types as is common to the one at destroy...

pkg/cmd/pipeline/translate.go
```diff
@@ -278,6 +275,7 @@ func updateCmap(cmap *apiv1.ConfigMap, data *CfgData) error {
 	cmap.Data[yamlField] = base64.StdEncoding.EncodeToString(data.Manifest)
 	cmap.Data[iconField] = data.Icon
 	cmap.Data[actionNameField] = actionName
+	cmap.Data[variablesField] = translateVariables(data.Variables)
```
jLopezbarb, line 278: Post-merge review:
```suggestion
vars := translateVariables(data.Variables)
if vars != "" {
  cmap.Data[variablesField] = translateVariables(data.Variables)
}
```

jLopezbarb: We depend on checking if the [variables field is set](https://github.com/okteto/okteto/blob/master/cmd/up/exec.go#L275). That's why e2e tests are failing

PR #3689, opened 2023-06-01
fix: variable expansion on buildargs that only have one env

### LINKED ISSUE #3684: Build Args refering to another build are not expanded
**Describe the bug**
If I use an okteto build environment as a build arg the build arg is not getting the correct value. I expect that this build arg uses the okteto env var that I provided

**To Reproduce**

1. Create the following Dockerfile:
```Dockerfile
FROM alpine

ARG KEY
RUN echo testing: $KEY
```
1. Create the following okteto manifest:
```yaml
build:
  test:
    context: .
    args:
      KEY: hello
  test-2:
    context: .
    args:
      KEY: $OKTETO_BUILD_TEST_IMAGE
    depends_on:
      - test
```
1. Run `okteto build --progress plain -f okteto.yml --no-cache`
1. See that the `$OKTETO_BUILD_TEST_IMAGE` is not propagated

**Expected behavior**
I want that the build arg gets into the image

### LINKED ISSUE #3686: `OKTETO_` env vars extended too early making manifest not work
See: https://community.okteto.com/t/docker-image-no-longer-receiving-okteto-build-arg/841

Image

jLopezbarb:
# Proposed changes

Fixes #3686 
Fixes #3684

- Update dependency
- Add unit tests covering a value with a single env var
- Add unit test for getting the build hash with a single env var
- If the value is a single env var on serializer don't expand it

## Tests done
- Using #3684 reproduction steps. Check that the variable is printed inside buildkit
- Check that the build hash changes with different values of the build env var

### DIFF
--- a/pkg/model/dev.go
+++ b/pkg/model/dev.go
@@ -141,7 +141,11 @@ type BuildArg struct {
 }
 
 func (v *BuildArg) String() string {
-	return fmt.Sprintf("%s=%s", v.Name, v.Value)
+	value, err := ExpandEnv(v.Value, true)
+	if err != nil {
+		return fmt.Sprintf("%s=%s", v.Name, v.Value)
+	}
+	return fmt.Sprintf("%s=%s", v.Name, value)
 }
 
 // BuildArgs is a list of arguments used on the build step.
--- a/cmd/build/v2/config_test.go
+++ b/cmd/build/v2/config_test.go
@@ -173,7 +173,7 @@ func TestGetGitCommit(t *testing.T) {
 func TestGetTextToHash(t *testing.T) {
 	fs := afero.NewMemMapFs()
 	afero.WriteFile(fs, "secret", []byte("bar"), 0600)
-
+	t.Setenv("BAR", "bar")
 	type input struct {
 		repo      fakeConfigRepo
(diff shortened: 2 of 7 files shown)


andreafalzetti (approved):

pkg/model/dev.go
```diff
@@ -141,7 +141,11 @@ type BuildArg struct {
 }
 
 func (v *BuildArg) String() string {
-	return fmt.Sprintf("%s=%s", v.Name, v.Value)
+	value, err := ExpandEnv(v.Value, true)
+	if err != nil {
+		return fmt.Sprintf("%s=%s", v.Name, v.Value)
+	}
+	return fmt.Sprintf("%s=%s", v.Name, value)
```
AdrianPedriza, line 148:
```suggestion
	value, err := ExpandEnv(v.Value, true)
	if err != nil {
		value = v.Value
	}
	
	return fmt.Sprintf("%s=%s", v.Name, value)
```

AdrianPedriza (approved):

AdrianPedriza (approved):

pkg/model/dev.go
AdrianPedriza: minor suggestion

PR #3692, opened 2023-06-01
feat: add preview labels to preview list

### LINKED ISSUE #3608: Add a flag to support list with labels
## Parent epic/feature

* Feature: [Labels for preview environment](https://www.notion.so/okteto/Labels-for-Preview-Environments-Epic-1-CLI-Support-4178b8e637a34829972c6baf9247be42?pvs=4)

## Description

- A new flag `label` is supported in the `deploy list` command
- The new flag label is array type and can store several labels
- The labels added as flags are passed as query arguments to the gql `previews` call
- The result of the command should only show the previews with the labels
- The result of the command should show the labels that a preview has
- If there is no label flag the behavior must be the same as it is right now. For example, the gql call should not receive a label argument (to keep retrocompatibility)

## Acceptance Criteria

- [x]  A user can add several `labels` as flags
- [x]  The flag `—label` can be used 
- [x]  If the backend return an error, the user must get a friendly error
- [x]  If there is no labels the gql call shouldn’t have any mention to labels

## Test strategy

- Unit tests covering:
    - query with several labels
    - query with label

## User scenario

As a user that list all my preview environment I want to filter it by labels so I can have only specific preview environments

jLopezbarb:
# Proposed changes

Fixes #3608 

- Add a label flag to preview list command
- If the backend don't support it, return a user friendly error
- The output shows the labels that a preview has

### Acceptance criteria

- [x]  A user can add several labels as flags
- [x]  The flag —label can be used
- [x]  If the backend return an error, the user must get a friendly error
- [x]  If there is no labels the gql call shouldn’t have any mention to labels

### DIFF
--- a/pkg/okteto/preview.go
+++ b/pkg/okteto/preview.go
@@ -81,7 +81,11 @@ type destroyPreviewMutation struct {
 	Response previewIDStruct `graphql:"destroyPreview(id: $id)"`
 }
 
-type listPreviewQuery struct {
+type listPreviewQueryWithLabels struct {
+	Response []previewEnvWithLabels `graphql:"previews(labels: $labels)"`
+}
+
+type listPreviewQueryDeprecated struct {
 	Response []previewEnv `graphql:"previews"`
 }
 
@@ -132,6 +136,13 @@ type previewEnv struct {
 	Id       graphql.String
 	Sleeping graphql.Boolean
 	Scope    graphql.String
+}
+
+type previewEnvWithLabels struct {
+	Id            graphql.String
+	Sleeping      graphql.Boolean
+	Scope         graphql.String
+	PreviewLabels []graphql.String
 }
 
 type deployPreviewResponse struct {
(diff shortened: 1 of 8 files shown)


pkg/okteto/preview.go
```diff
@@ -81,7 +81,11 @@ type destroyPreviewMutation struct {
 	Response previewIDStruct `graphql:"destroyPreview(id: $id)"`
 }
 
-type listPreviewQuery struct {
+type listPreviewQueryWithLabels struct {
```
AdrianPedriza, line 84: I dont think we should include `WithLabels` in this struct. We could add more params to the query but we should add each param for the struct name

pkg/okteto/preview.go
```diff
@@ -134,6 +138,13 @@ type previewEnv struct {
 	Scope    graphql.String
 }
 
+type previewEnvWithLabels struct {
```
AdrianPedriza, line 141: same here

AdrianPedriza (commented):
can you post here loom demo covering the case when fall back to the deprecated method?

AdrianPedriza:
how about add e2e for this PR?

jLopezbarb:
> listPreviewQuery

The backend is not prepared to support it. We should do it as a follow up

AdrianPedriza:
> > how about add e2e for this PR?
> 
> The backend is not prepared to support it. We should do it as a follow up

oh ok. I thought we did an abstraction about backend layer in order to mock it

AdrianPedriza:
> > > how about add e2e for this PR?
> > 
> > 
> > The backend is not prepared to support it. We should do it as a follow up
> 
> oh ok. I thought we did an abstraction about backend layer in order to mock it to do some integration test

AdrianPedriza (approved):

PR #3703, opened 2023-06-09
Fix depends on when loading stack from file

teresaromero:
# Proposed changes

Fixes https://github.com/okteto/app/issues/6669

When reading a compose from a file using `okteto deploy -f docker-compose.yaml` the function that loads the stack was not performing the validation required.

This PR enables validation when the stack is read from file as it is done when inferring the stack when the command is `okteto deploy` and there is no `okteto.yaml` but `docker-compose.yaml`

Depends_on fields point into services within the same compose.

Output is now the same in both cases:

Example using `movies-with-compose` and removing the service `frontend` but leaving `depends_on` at `movies` service.

```
API server listening at: 127.0.0.1:16484
 i  Using teresaromero @ okteto.teresa.dev.okteto.net as context
 x   Service 'movies' depends on service 'frontend' which is undefined.
```

### DIFF
--- a/pkg/model/manifest.go
+++ b/pkg/model/manifest.go
@@ -612,7 +612,7 @@ func getOktetoManifest(devPath string) (*Manifest, error) {
 	}
 
 	if isEmptyManifestFile(b) {
-		return nil, fmt.Errorf("%w: %w", oktetoErrors.ErrInvalidManifest, oktetoErrors.ErrEmptyManifest)
+		return nil, fmt.Errorf("%s: %w", oktetoErrors.ErrInvalidManifest, oktetoErrors.ErrEmptyManifest)
 	}
 
 	manifest, err := Read(b)
--- a/pkg/model/manifest.go
+++ b/pkg/model/manifest.go
@@ -612,7 +612,7 @@ func getOktetoManifest(devPath string) (*Manifest, error) {
 	}
 
 	if isEmptyManifestFile(b) {
-		return nil, fmt.Errorf("%w: %s", oktetoErrors.ErrInvalidManifest, oktetoErrors.ErrEmptyManifest)
+		return nil, fmt.Errorf("%w: %w", oktetoErrors.ErrInvalidManifest, oktetoErrors.ErrEmptyManifest)
 	}
 
 	manifest, err := Read(b)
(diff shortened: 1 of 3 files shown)


pkg/model/stack.go
```diff
@@ -432,7 +432,7 @@ func (svc *Service) ToDev(svcName string) (*Dev, error) {
 	return d, nil
 }
 
-func (s *Stack) Validate() error {
```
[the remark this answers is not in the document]
teresaromero: it was just being used inside the pkg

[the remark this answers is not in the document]
teresaromero: 👍🏻  removed the commit

pkg/model/manifest.go
```diff
@@ -410,12 +410,9 @@ func getManifestFromFile(cwd, manifestPath string) (*Manifest, error) {
 		for _, composeInfo := range stackManifest.Deploy.ComposeSection.ComposesInfo {
 			composeFiles = append(composeFiles, composeInfo.File)
 		}
-		s, stackErr := LoadStack("", composeFiles, false)
-
-		// We failed to load a stack file and a manifest file, so we need to return
-		// only the original manifest error
```
ifbyol, line 416: According to the code comment, this was done on purpose. Have we checked or asked about why we should be returning the original manifest error?

And why the code was getting a panic? My understanding is that this should be returning an error (not the one we wanted), but as it's an error, I would expect that the execution is broken instead of keep going and get a panic

ifbyol: talked offline, we will run e2e tests to see if they pass or not with these changes

ifbyol: @teresaromero should we also change it here? https://github.com/okteto/okteto/blob/bf5f3cde626c900e021f3b626c82a952bea2d77c/pkg/model/manifest.go#L440

I think that scenario is when we refer to a docker compose from an okteto manifest V2. If that docler-compose also have a depends_on, it might fail also as we are not validating the stack there.

teresaromero: Yes, that is right, i've just debugged this and added validation. In this case there is also a panic if an okteto manifest has a compose section with an invalid compose...

In this case the err returned is this one

teresaromero:
> LGTM, it is possible to add a test for `getManifestFromFile` with an invalid stack?

I've refactored the tests for this function and added the case for the error on depends on

teresaromero:
Deepsource are already existent issues, i wont fix them as they imply changes not relative to this PR

maroshii (approved):

Issue #3699, opened 2023-06-07
Orphaned .git/index.lock file running `okteto up [svc]`


mkraft:
**Describe the bug**

I'm using git version 2.41.0 (which was released last week so granted it's possible this is a bug with git itself) and I'm not using an IDE or any other software that is interacting with git other than okteto cli.

**To Reproduce**

1. Go to the root of the project
1. Ensure there is no .git/index.lock file (`rm .git/index.lock` if there is)
1. Run `okteto up [svc]`
1. An orphaned .git/index.lock file appears

**Expected behavior**

No orphaned .git/index.lock file remains.

**Desktop (please complete the following information):**

- OS: macOS
- Version 13.4

**Additional context**

The .git/index.lock file is orphaned by multiple services we run (e.g. web, api).

Here's a screenshot of `okteto up` running in one terminal and `sudo fs_usage | grep .git/index.lock` in another. I confirmed that before running `okteto up` the index.lock did not exist and afterwards the file is orphaned (ie exists).

<img width="1234" alt="Screenshot 2023-06-07 at 11 30 17 AM" src="https://github.com/okteto/okteto/assets/1149597/238807dd-92af-4e01-a858-9c5f5fd504c2">

teresaromero:
@mkraft we are trying to reproduce the scenario but for now we have not been able, with latest okteto cli and git versions.

However, we would like to know more regarding the operation `git status` run at your repository, as okteto has a 1 second timeout for this operation and we might be having something there.

Could you run `time git status --porcelain -z` at your repository and let us know the time the operation spends? This can be useful for investigation.

Thanks!

PR #3704, opened 2023-06-09
Fix context cancelled with reading git status operation

teresaromero:
# Proposed changes

Fixes: https://github.com/okteto/app/issues/6687

Relates: https://github.com/okteto/okteto/issues/3699

- Add custom Cancel function to `exec.CommandContext` so when the context is cancelled by external functions, the funcion exits with a termination signal instead of kill, that way git can remove the lock file and recover from the cancelation.

EDIT: after research, for windows we have to keep killing the process as there is no termination signal implementation

### DIFF
--- a/pkg/repository/local_git.go
+++ b/pkg/repository/local_git.go
@@ -3,10 +3,12 @@ package repository
 import (
 	"context"
 	"errors"
-	"github.com/go-git/go-git/v5"
 	"os"
 	"os/exec"
 	"strings"
+	"syscall"
+
+	"github.com/go-git/go-git/v5"
 )
 
 var (
@@ -24,6 +26,10 @@ type LocalExec struct{}
 
 func (*LocalExec) RunCommand(ctx context.Context, dir string, name string, arg ...string) ([]byte, error) {
 	c := exec.CommandContext(ctx, name, arg...)
+	c.Cancel = func() error {
+		return c.Process.Signal(syscall.SIGTERM)
+	}
+
 	c.Dir = dir
 	c.Env = os.Environ()
 	return c.Output()
--- a/pkg/repository/local_git.go
+++ b/pkg/repository/local_git.go
@@ -7,6 +7,9 @@ import (
 	"os/exec"
 	"strings"
 	"syscall"
+	"time"
+
+	oktetoLog "github.com/okteto/okteto/pkg/log"
 
 	"github.com/go-git/go-git/v5"
 )
@@ -27,7 +30,20 @@ type LocalExec struct{}
 func (*LocalExec) RunCommand(ctx context.Context, dir string, name string, arg ...string) ([]byte, error) {
 	c := exec.CommandContext(ctx, name, arg...)
 	c.Cancel = func() error {
-		return c.Process.Signal(syscall.SIGTERM)
+		oktetoLog.Debugf("terminating %s...", c.String())
+		if err := c.Process.Signal(syscall.SIGTERM); err != nil {
+			oktetoLog.Debugf("err at signal SIGTERM: %v", err)
+		}
+
+		time.Sleep(5 * time.Second)
+		if err := c.Process.Signal(syscall.Signal(0)); err != nil {
+			if errors.Is(err, os.ErrProcessDone) {
+				return nil
+			}
+			oktetoLog.Debugf("reading signal with error %v", err)
+		}
+		oktetoLog.Debugf("killing %s...", c.String())
+		return c.Process.Signal(syscall.SIGKILL)
 	}
 
 	c.Dir = dir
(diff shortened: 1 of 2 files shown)


cmd/deploy/deploy.go
```diff
@@ -95,6 +96,7 @@ type DeployCommand struct {
 	PipelineType       model.Archetype
 	isRemote           bool
 	runningInInstaller bool
+	repositorySHA      string
```
ifbyol, line 99: Define properties in a struct which won't be initialized as part of the instance creation is misleading. It might lead to panics (nil pointers. In this case it doesn't matter as it's not a pointer) or a wrong logic based on checks when the property wasn't set yet. I would try to use this pattern to pass the sha to the command configuration

cmd/deploy/deploy.go
```diff
@@ -297,6 +299,18 @@ func (dc *DeployCommand) RunDeploy(ctx context.Context, deployOptions *Options)
 		return fmt.Errorf("failed to get the current working directory: %w", err)
 	}
 
+	repo := repository.NewRepository(cwd)
+	sha, err := repo.GetSHA()
```
ifbyol, line 303: Why are these changes required to fix the issue?

We should review where this is being called. Checking the code, it seems that during a deploy command execution, this function or the analogous `IsClean` might be called at least 3 times. Also, before this change, this logic was on `command_configuration` and it was being executed only when `OktetoGitCommitEnvVar` wasn't already set, so now, we will be calling it more times potentially impacting the performance.

If these changes are required to fix the issue (I don't see `repositorySHA` being used in the function which will kill the process), let's revert it.

pkg/repository/local_git.go
```diff
@@ -24,6 +29,23 @@ type LocalExec struct{}
 
 func (*LocalExec) RunCommand(ctx context.Context, dir string, name string, arg ...string) ([]byte, error) {
 	c := exec.CommandContext(ctx, name, arg...)
+	c.Cancel = func() error {
```
ifbyol, line 32: I think CircleCI is complaining about the existence of `Cancel`

teresaromero: yes, i am investigating

andreafalzetti (approved):

teresaromero:
working parallel in https://github.com/okteto/okteto/tree/tere/isclean-sha-optimization_win to resolve the windows integration failing test when updating go

maroshii:
Looking at the documentation and exec code, it looks like `cmd.Cancel` simply sends a signal to the parent process upon context cancellation. 

I haven't tested this but in case it helps we could do this ~manually (at least for testing) with something like this:

```golang
func (*LocalExec) RunCommand(ctx context.Context, dir string, name string, arg ...string) ([]byte, error) {
	c := exec.Command(name, arg...)
	c.Dir = dir
	c.Env = os.Environ()

	go func() {
		<-ctx.Done() // TODO: Add timeout so that this doesn't block forever if ctx is never cancelled
		oktetoLog.Debugf("terminating %s - %s/%s", c.String(), dir, name)
		if err := c.Process.Signal(syscall.SIGTERM); err != nil {
			oktetoLog.Debugf("err at signal SIGTERM: %v", err)
		}

		time.Sleep(5 * time.Second)
		if err := c.Process.Signal(syscall.Signal(0)); err != nil {
			if errors.Is(err, os.ErrProcessDone) {
				return nil
			}
			oktetoLog.Debugf("reading signal with error %v", err)
		}
		oktetoLog.Debugf("killing %s - %s/%s", c.String(), dir, name)
		return c.Process.Signal(syscall.SIGKILL)

	}()

	return c.Output()
}
```

AdrianPedriza (approved):

pkg/repository/local_git.go
```diff
@@ -24,6 +29,23 @@ type LocalExec struct{}
...
+		oktetoLog.Debugf("terminating %s - %s/%s", c.String(), dir, name)
+		if err := c.Process.Signal(syscall.SIGTERM); err != nil {
```
jLopezbarb, line 34: Do we know if this is working on windows? We have been facing issues with signals on windows before and leaving zombie processes because of that

pkg/repository/local_git.go
```diff
@@ -24,6 +29,23 @@ type LocalExec struct{}
...
+			oktetoLog.Debugf("err at signal SIGTERM: %v", err)
+		}
+
+		time.Sleep(5 * time.Second)
```
jLopezbarb, line 38: Why 5 seconds? I'd try to avoid delaying it for that long

pkg/repository/local_git.go
jLopezbarb, line 32: Could we somehow add a unit/integration test testing that this part of the code gets the proper execution?

Maybe a `RunCommand` of sleeping infinite and cancelling the context 1 ms after it's executing?

jLopezbarb (commented):
Code looks good but check the comments before merging

andreafalzetti:
@teresaromero I am moving this to draft until the other branch is ready, wdyt?

andreafalzetti (approved):

PR #3707, opened 2023-06-12
fix: no optional locks on git read only commands

jpf-okteto:
May resolve:
- https://github.com/okteto/okteto/issues/3699

Complementary or alternative to:
- https://github.com/okteto/okteto/pull/3704

This PR focuses on `git` configuration to avoid creating optional locks
on read only operations, specifically `git status`.

This comes with a cost, as the `.git/index` must be recalculated if it
is not up to date or if it was generated from a different `git` arch
such as running `git status` sequentially from a `linux/amd64` and then
from a `darwin/arm64`.

Because of it, `git` command timeout has been extended to 5 seconds.

This change is complementary to https://github.com/okteto/okteto/pull/3704 as it doesn't touch process
management and signals, so with the current `HEAD`, `git` command will
still be killed if it goes beyond the specified timeout.

`--no-optional-locks` was introduced in `git` v2.15:
- https://github.com/git/git/blob/master/Documentation/RelNotes/2.15.0.txt#L97-L100
- https://github.com/git/git/blob/fe86abd7511a9a6862d5706c6fa1d9b57a63ba09/Documentation/RelNotes/2.15.0.txt#L97-L100

### DIFF
--- a/pkg/repository/git.go
+++ b/pkg/repository/git.go
@@ -73,7 +73,7 @@ func (r gitRepoController) isClean(ctx context.Context) (bool, error) {
 	timeoutErr := errors.New("timeout exceeded")
 
 	go func() {
-		time.Sleep(time.Second)
+		time.Sleep(5 * time.Second)
 		close(timeoutCh)
 		cancel()
 		ch <- cleanStatus{false, timeoutErr}
--- a/pkg/repository/git.go
+++ b/pkg/repository/git.go
@@ -73,7 +73,7 @@ func (r gitRepoController) isClean(ctx context.Context) (bool, error) {
 	timeoutErr := errors.New("timeout exceeded")
 
 	go func() {
-		time.Sleep(5 * time.Second)
+		time.Sleep(time.Second)
 		close(timeoutCh)
 		cancel()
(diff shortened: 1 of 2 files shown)


pkg/repository/git.go
```diff
@@ -73,7 +73,7 @@ func (r gitRepoController) isClean(ctx context.Context) (bool, error) {
 	timeoutErr := errors.New("timeout exceeded")
 
 	go func() {
-		time.Sleep(time.Second)
+		time.Sleep(5 * time.Second)
```
ifbyol, line 76: Reviewing the flow, this might be called at least 3 times in a `deploy` sequence. This could have an impact of 15 seconds on those deployments and we already had performance issues because of it. I would avoid to increase this timeout

andreafalzetti: I agree with @ifbyol, the timeout should be 1s, otherwise it defeats the purpose of being there. Once we refactor the code and avoid calling this function many times, I agree in bumping up the timeout to 5s

andreafalzetti: That change has not been scheduled in out sprint yet, I will talk to @rlamana about it

andreafalzetti (approved):

PR #3721, opened 2023-06-18
Execute `okteto destroy --remote` successfully when there is no okteto manifest found

### LINKED ISSUE #3656: Panic executing `okteto destroy --remote` when there is no okteto manifest
**Describe the bug**

I was able to deploy my repository using `okteto deploy --remote` but when I executed `okteto destroy --remote` I got a panic

```
╰─ okteto destroy --remote
 i  Using test-nacho @ okteto.mario18rc.dev.okteto.net as context
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x2c43057]

goroutine 1 [running]:
github.com/okteto/okteto/cmd/destroy.newRemoteDestroyer(0xc001c74000)
	github.com/okteto/okteto/cmd/destroy/remote.go:104 +0x217
github.com/okteto/okteto/cmd/destroy.(*destroyCommand).getDestroyer(0xc001a3fbc0, {0x3b54878, 0xc0001a0000}, 0xc0001e5880)
	github.com/okteto/okteto/cmd/destroy/destroy.go:269 +0x345
github.com/okteto/okteto/cmd/destroy.Destroy.func1(0xc001862f00?, {0xc00184c840?, 0x0?, 0x1?})
	github.com/okteto/okteto/cmd/destroy/destroy.go:193 +0x990
github.com/spf13/cobra.(*Command).execute(0xc001862f00, {0xc00184c830, 0x1, 0x1})
	github.com/spf13/cobra@v1.5.0/command.go:872 +0x694
github.com/spf13/cobra.(*Command).ExecuteC(0xc001848780)
	github.com/spf13/cobra@v1.5.0/command.go:990 +0x3b4
	github.com/spf13/cobra@v1.5.0/command.go:918
main.main()
	github.com/okteto/okteto/main.go:148 +0xa25
```

I think the error might come from the fact I'm not having an `okteto.yml` file in my root folder. Instead, I have a `k8s.yml` file.

The repository is https://github.com/ifbyol/test-pipeline and the branch `destroy-custom-manifest`

**To Reproduce**

1. Clone the repository https://github.com/ifbyol/test-pipeline
2. Change to the branch `destroy-custom-manifest`
3. Execute `okteto deploy --remote`. You should see how the application is successfully deployed
4. Execute `okteto destroy --remote`. You should see the panic error

**Expected behavior**

As I was able to deploy the application correctly, I would expect to be able to destroy it successfully. If something is wrong, I would expect a clear error message instead of a stack trace

- Version: `2.15.3` and master

**Additional context**
N/A

nitishfy:
# Proposed changes

Fixes #3656 

image

**Additional Information:**

1. **Why is the panic occurring?**

- In the [directory](https://github.com/ifbyol/test-pipeline), we don't have any `okteto.yml` file. Okteto was able to deploy because of the `k8s.yml` file and executed the command `kubectl apply -f...` that is [mentioned here](https://github.com/okteto/okteto/blob/master/pkg/model/manifest.go#L554). it applied the commands successfully and hence was able to deploy. 

- When it comes to destroy, okteto tried to look for okteto manifet such as `okteto.yml` but in the repository, `okteto-custom.yml` was present. Hence it detected that [no okteto manifest was found](https://github.com/okteto/okteto/blob/master/pkg/discovery/manifest.go#L24). However, the K8s manifest was there but there were no commands on how to delete the services produced by k8s manifest. Generally, we have Okteto manifest mentioning how to destroy services produced by k8s manifest but here `okteto-custom.yml` doesn't fit the naming convention for the okteto manifest.

2. **Have you tested your changes on some other application that contains both `okteto.yml` and `k8s.yml`?**

- Yes, I've tried testing changes on [this go application](https://github.com/okteto/go-getting-started) that contains both `k8s.yml` and `okteto.yml`. The changes run well as expected. Here's the attached image:
Screenshot from 2023-06-22 11-11-42

3. **Is there anything else you'd like to add?**

- I think we may need to add separate commands to destroy services generated by _compose, helm_ when no okteto manifest has been detected. The application will be deployed but will execute panic when destroying if there is no okteto manifest and no command specified on how to delete services generated by _compose, helm etc,_

### DIFF
--- a/cmd/destroy/destroy.go
+++ b/cmd/destroy/destroy.go
@@ -235,14 +235,11 @@ func (dc *destroyCommand) getDestroyer(ctx context.Context, opts *Options) (dest
 	} else {
 		manifest, err := model.GetManifestV2(opts.ManifestPath)
 		if err != nil {
-			// error will only be available when there is no valid manifest found
-			// hence, we need not to destroy because the manifest would never be deployed
-			// Return the error message
+			// Log error message but application can still be deleted
 			oktetoLog.Infof("could not find manifest file to be executed: %s", err)
 			manifest = &model.Manifest{
 				Destroy: &model.DestroyInfo{},
 			}
-			return nil, err
 		}
 
 		isRemote := utils.LoadBoolean(constants.OktetoDeployRemote)
--- a/cmd/destroy/destroy.go
+++ b/cmd/destroy/destroy.go
@@ -270,7 +270,7 @@ func (dc *destroyCommand) getDestroyer(ctx context.Context, opts *Options) (dest
 		// if the command is executed remotely and destroy section is nil, this will cause a panic
 		// Hence consider the command locally for successful execution.
 		// This will make sure all the resources are deleted.
-		if runInRemote == true && manifest.Destroy == nil {
+		if manifest.Destroy == nil {
 			runInRemote = false
 		}
 
--- a/cmd/destroy/destroy.go
+++ b/cmd/destroy/destroy.go
@@ -16,10 +16,6 @@ package destroy
 import (
 	"context"
 	"fmt"
-	"os"
-	"path/filepath"
-	"time"
-
 	contextCMD "github.com/okteto/okteto/cmd/context"
 	"github.com/okteto/okteto/cmd/utils"
 	"github.com/okteto/okteto/cmd/utils/executor"
@@ -27,7 +23,6 @@ import (
 	"github.com/okteto/okteto/pkg/constants"
 	"github.com/okteto/okteto/pkg/devenvironment"
 	oktetoErrors "github.com/okteto/okteto/pkg/errors"
-
 	"github.com/okteto/okteto/pkg/k8s/kubeconfig"
 	"github.com/okteto/okteto/pkg/k8s/namespaces"
 	"github.com/okteto/okteto/pkg/k8s/secrets"
@@ -37,6 +32,9 @@ import (
 	oktetoPath "github.com/okteto/okteto/pkg/path"
 	"github.com/spf13/cobra"
 	v1 "k8s.io/api/core/v1"
+	"os"
+	"path/filepath"
+	"time"
 )
 
 const (
@@ -55,7 +53,7 @@ type secretHandler interface {
 	List(ctx context.Context, ns, labelSelector string) ([]v1.Secret, error)
 }
 
-// Options destroy commands options
+// Options represents the options for destroy command
 type Options struct {
 	// ManifestPathFlag is the option -f as introduced by the user when executing this command.
 	// This is stored at the configmap as filename to redeploy from the ui.
@@ -220,15 +218,15 @@ func getTempKubeConfigFile(name string) string {
 
 func (dc *destroyCommand) getDestroyer(ctx context.Context, opts *Options) (destroyInterface, error) {
 	var (
-		deployer destroyInterface
-		err      error
+		destroyer destroyInterface
+		err       error
 	)
 
 	if opts.DestroyAll {
 		if !okteto.Context().IsOkteto {
 			return nil, oktetoErrors.ErrContextIsNotOktetoCluster
 		}
-		deployer, err = newLocalDestroyerAll(dc.k8sClientProvider, dc.executor, dc.nsDestroyer, dc.oktetoClient)
+		destroyer, err = newLocalDestroyerAll(dc.k8sClientProvider, dc.executor, dc.nsDestroyer, dc.oktetoClient)
 		if err != nil {
 			return nil, err
 		}
@@ -237,11 +235,14 @@ func (dc *destroyCommand) getDestroyer(ctx context.Context, opts *Options) (dest
 	} else {
 		manifest, err := model.GetManifestV2(opts.ManifestPath)
 		if err != nil {
-			// Log error message but application can still be deleted
+			// error will only be available when there is no valid manifest found
+			// hence, we need not to destroy because the manifest would never be deployed
+			// Return the error message
 			oktetoLog.Infof("could not find manifest file to be executed: %s", err)
 			manifest = &model.Manifest{
 				Destroy: &model.DestroyInfo{},
 			}
+			return nil, err
 		}
 
 		isRemote := utils.LoadBoolean(constants.OktetoDeployRemote)
@@ -265,8 +266,16 @@ func (dc *destroyCommand) getDestroyer(ctx context.Context, opts *Options) (dest
 		}
 		runInRemote := !isRemote && (destroyImage != "" || opts.RunInRemote)
 
+		// destroy section should not be nil whether the command is executed remotely or locally
+		// if the command is executed remotely and destroy section is nil, this will cause a panic
+		// Hence consider the command locally for successful execution.
+		// This will make sure all the resources are deleted.
+		if runInRemote == true && manifest.Destroy == nil {
+			runInRemote = false
+		}
+
 		if runInRemote {
-			deployer = newRemoteDestroyer(manifest)
+			destroyer = newRemoteDestroyer(manifest)
 			oktetoLog.Info("Destroying remotely...")
 		} else {
 			destroyerAll, err := newLocalDestroyerAll(dc.k8sClientProvider, dc.executor, dc.nsDestroyer, dc.oktetoClient)
@@ -274,10 +283,10 @@ func (dc *destroyCommand) getDestroyer(ctx context.Context, opts *Options) (dest
 				return nil, err
 			}
 
-			deployer = newLocalDestroyer(manifest, destroyerAll)
+			destroyer = newLocalDestroyer(manifest, destroyerAll)
 			oktetoLog.Info("Destroying locally...")
 		}
 	}
 
-	return deployer, nil
+	return destroyer, nil
 }
--- a/cmd/destroy/destroy.go
+++ b/cmd/destroy/destroy.go
@@ -266,14 +266,6 @@ func (dc *destroyCommand) getDestroyer(ctx context.Context, opts *Options) (dest
 		}
 		runInRemote := !isRemote && (destroyImage != "" || opts.RunInRemote)
 
-		// destroy section should not be nil whether the command is executed remotely or locally
-		// if the command is executed remotely and destroy section is nil, this will cause a panic
-		// Hence consider the command locally for successful execution.
-		// This will make sure all the resources are deleted.
-		if manifest.Destroy == nil {
-			runInRemote = false
-		}
-
 		if runInRemote {
 			destroyer = newRemoteDestroyer(manifest)
 			oktetoLog.Info("Destroying remotely...")
--- a/cmd/destroy/remote.go
+++ b/cmd/destroy/remote.go
@@ -104,9 +104,12 @@ type remoteDestroyCommand struct {
 func newRemoteDestroyer(manifest *model.Manifest) *remoteDestroyCommand {
 	fs := afero.NewOsFs()
 	builder := remoteBuild.NewBuilderFromScratch()
+	if manifest.Destroy == nil {
+		manifest.Destroy = &model.DestroyInfo{}
+	}
 	return &remoteDestroyCommand{
 		builder:              builder,
-		destroyImage:         "",
+		destroyImage:         manifest.Destroy.Image,
 		fs:                   fs,
 		workingDirectoryCtrl: filesystem.NewOsWorkingDirectoryCtrl(),
 		temporalCtrl:         filesystem.NewTemporalDirectoryCtrl(fs),
--- a/cmd/destroy/remote.go
+++ b/cmd/destroy/remote.go
@@ -106,7 +106,7 @@ func newRemoteDestroyer(manifest *model.Manifest) *remoteDestroyCommand {
 	builder := remoteBuild.NewBuilderFromScratch()
 	return &remoteDestroyCommand{
 		builder:              builder,
-		destroyImage:         manifest.Destroy.Image,
+		destroyImage:         "",
 		fs:                   fs,
 		workingDirectoryCtrl: filesystem.NewOsWorkingDirectoryCtrl(),
 		temporalCtrl:         filesystem.NewTemporalDirectoryCtrl(fs),
--- a/cmd/destroy/remote.go
+++ b/cmd/destroy/remote.go
@@ -1,3 +1,16 @@
+// Copyright 2023 The Okteto Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 package destroy
 
 import (
--- a/pkg/model/manifest.go
+++ b/pkg/model/manifest.go
@@ -524,14 +524,14 @@ func GetInferredManifest(cwd string) (*Manifest, error) {
 		}
 		oktetoLog.AddToBuffer(oktetoLog.InfoLevel, "Found helm chart on %s", chartPath)
 		tags := inferHelmTags(chartPath)
-		command := fmt.Sprintf("helm upgrade --install ${%s} %s %s", constants.OktetoAutodiscoveryReleaseName, chartPath, tags)
+		deployHelm := fmt.Sprintf("helm upgrade --install ${%s} %s %s", constants.OktetoAutodiscoveryReleaseName, chartPath, tags)
 		chartManifest := &Manifest{
 			Type: ChartType,
 			Deploy: &DeployInfo{
 				Commands: []DeployCommand{
 					{
-						Name:    command,
-						Command: command,
+						Name:    deployHelm,
+						Command: deployHelm,
 					},
 				},
 			},
--- a/pkg/model/manifest_test.go
+++ b/pkg/model/manifest_test.go
@@ -1297,3 +1297,78 @@ func Test_Manifest_HasBuildSection(t *testing.T) {
 
 	}
 }
+
+func Test_getInferredManifestFromK8sManifestFile(t *testing.T) {
+	wd := t.TempDir()
+	fullpath := filepath.Join(wd, "k8s.yml")
+	f, err := os.Create(fullpath)
+	assert.NoError(t, err)
+	defer func() {
+		if err := f.Close(); err != nil {
+			t.Fatalf("Error closing file %s: %s", fullpath, err)
+		}
+	}()
+	_, err = GetInferredManifest(wd)
+	assert.NoError(t, err)
+}
+
+func Test_getInferredManifestFromK8sManifestFolder(t *testing.T) {
+	wd := t.TempDir()
+	fullpath := filepath.Join(wd, "manifests")
+	assert.NoError(t, os.MkdirAll(filepath.Dir(fullpath), 0750))
+	f, err := os.Create(fullpath)
+	assert.NoError(t, err)
+	defer func() {
+		if err := f.Close(); err != nil {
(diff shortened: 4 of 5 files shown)


teresaromero (commented):
hi, thanks for contributing!
I've left some questions in order to understand the changes. Have you identify when is the panic occuring? I would start by identifying where, and then on top probably making some refactor. 

In order to debug and find the panic i use the following setup under VSCode .launch.json to run a command at debug mode so it can stop on the breakpoint i set:

```
 {
            "name": "Launch okteto destroy --remote",
            "type": "go",
            "request": "launch",
            "mode": "debug",
            "program": "${workspaceFolder}/main.go",
            "cwd": “<path to a repository to use for debugging, in your case one without okteto.yaml> ,
            "args": [
                "destroy",
                "--remote"
            ]
        },

```

nitishfy:
> In order to debug and find the panic i use the following setup under VSCode .launch.json to run a command at debug mode so it can stop on the breakpoint i set:
> 
> ```
>  {
>             "name": "Launch okteto destroy --remote",
>             "type": "go",
>             "request": "launch",
>             "mode": "debug",
>             "program": "${workspaceFolder}/main.go",
>             "cwd": “<path to a repository to use for debugging, in your case one without okteto.yaml> ,
>             "args": [
>                 "destroy",
>                 "--remote"
>             ]
>         },
> ```

Hi, this was precisely what I was looking for. I've gone through your suggestions and would be deep diving into this issue via debugging. I'd like to know something from you:

1. Is there a way to see how the flow of code is happening on certain manifests that I provide? How do you provide a manifest for the okteto to work upon in the debug mode?
2. I'm using Goland as a primary Golang code editor. Do you have any experience with debugging with Goland?

teresaromero:
> > In order to debug and find the panic i use the following setup under VSCode .launch.json to run a command at debug mode so it can stop on the breakpoint i set:
> > ```
> >  {
> >             "name": "Launch okteto destroy --remote",
> >             "type": "go",
> >             "request": "launch",
> >             "mode": "debug",
> >             "program": "${workspaceFolder}/main.go",
> >             "cwd": “<path to a repository to use for debugging, in your case one without okteto.yaml> ,
> >             "args": [
> >                 "destroy",
> >                 "--remote"
> >             ]
> >         },
> > ```
> 
> Hi, this was precisely what I was looking for. I've gone through your suggestions and would be deep diving into this issue via debugging. I'd like to know something from you:
> 
> 1. Is there a way to see how the flow of code is happening on certain okteto manifests that I provide? If yes, How do I provide a manifest for the okteto to work upon in the debug mode?
> 2. I'm using Goland as a primary Golang code editor. Do you have any experience with debugging with Goland?It is asking to provide Host and Port number for debugging.

Regarding Goland i am not too much familiar but i've tested this to be able to debug:
- Under configurations, add configuration and select `Go build`
- Choose a descriptive name for the configuration, in my case `Run exec`
- Run on Local machine, Run kind: package, Package path: github.com/okteto/okteto
- Working directory: here i added the path to the repository i use for testing, you can create a test repository with an okteto manifest or use one existant and iterate with different manifest configurations. this field will tell Goland to use "run" the command in this working directory as if you where running there okteto.
- Program arguments: here you should add the argument to run, in my case `exec echo "hello"` for example, if you want to debug the destroy --remote, this would be `destroy --remote`
- Click Ok to save this config and then run it by selecting the config and clicking the bug icon for debug.

It should look something like this, if breakpoints are set then the execution will stop where ever they are set.
Screenshot 2023-06-20 at 16 12 14

Hope this helps! let me know if you have further questions!

nitishfy:
Hi @teresaromero , thanks for your help! I was able to push the new changes and have mentioned about the details of this PR in the Pull Request description.

teresaromero:
@nitishfy I've gone though the solution and i see it like a good fit with what we have. So the issue was that the `inferred` manifest we use as `k8s.yaml` is present, did not have a Destroy section default defined, so this failed.

The `GetInferredManifest` has also the case of compose, chart and pipeline files, and similar problem, so when no okteto manifest per se exist, the inferred one will be used. Can you check if this also happens when we have this cases?

For this you can use the same example and instead of having a k8s.yaml file, have a docker-compose and a helm chart definition.

Great work! 👏🏻

nitishfy:
> @nitishfy I've gone though the solution and i see it like a good fit with what we have. So the issue was that the `inferred` manifest we use as `k8s.yaml` is present, did not have a Destroy section default defined, so this failed.
> 
> The `GetInferredManifest` has also the case of compose, chart and pipeline files, and similar problem, so when no okteto manifest per se exist, the inferred one will be used. Can you check if this also happens when we have this cases?
> 
> For this you can use the same example and instead of having a k8s.yaml file, have a docker-compose and a helm chart definition.
> 
> Great work! 👏🏻

Thank you! I think the panic will occur for each, I'll try working with other manifests and push the changes in this Pull Request only.

pkg/model/manifest.go
```diff
@@ -559,6 +559,15 @@ func GetInferredManifest(cwd string) (*Manifest, error) {
 					},
 				},
 			},
+
+			Destroy: &DestroyInfo{
+				Commands: []DeployCommand{
+					{
+						Name:    fmt.Sprintf("kubectl delete -f %s", manifestPath),
```
pchico83, line 566: @jLopezbarb what should be the expected behavior on `destroy` if there is no okteto manifest?

nitishfy: If the Okteto manifest isn't there, it'll look at whether there are other manifests such as Kubernetes, helm, and docker-compose. If none exist, it'll panic like in the issue this PR is trying to solve. 

Even if there are manifests other than okteto manifest existing, we've only provided commands in the `manifest.go` on how to deploy (like _kubectl apply -f ..._) but not provided commands on how to destroy these manifests if no okteto manifest is there.

nitishfy:
I'm working on to checking about _compose_ and found that _The application will not be deployed if we don't have compose but it can be destroyed even if we don't have a compose file. I think this is happening because of this part of the code:
image

_If we have encountered an error such that the manifest is not available, we should return a clear error message and not wait to destroy it. Isn't it?_ It is strange that application can't be deployed but can be destroyed when using compose.

nitishfy:
The changes have been pushed and solve the issue for _helm, compose, okteto manifest_ completely. I was getting the same error for destroying helm chart since there was no destroy command, it is also resolved now. 

I'm working on writing test cases for this and would push it very soon.
cc @teresaromero @pchico83

pkg/model/manifest_test.go
```diff
@@ -15,18 +15,17 @@ package model
...
 	"github.com/okteto/okteto/pkg/discovery"
 	oktetoErrors "github.com/okteto/okteto/pkg/errors"
 	"github.com/okteto/okteto/pkg/model/forward"
 	"github.com/stretchr/testify/assert"
 	apiv1 "k8s.io/api/core/v1"
 	"k8s.io/utils/pointer"
+	"os"
+	"path/filepath"
+	"testing"
+	"time"
```
nitishfy, line 28:
```suggestion
	"time"
```
IDE automatically sorted the imports.

pkg/model/manifest_test.go
```diff
@@ -1297,3 +1297,78 @@ func Test_Manifest_HasBuildSection(t *testing.T) {
...
+				assert.NoError(t, os.MkdirAll(filepath.Dir(fullpath), 0750))
+				f, err := os.Create(fullpath)
+				assert.NoError(t, err)
+				defer func() {
+					if err := f.Close(); err != nil {
+						t.Fatalf("Error closing file %s: %s", fullpath, err)
+					}
+				}()
+			}
+			_, err := GetInferredManifest(wd)
```
teresaromero, line 1363: can we check the manifest is the one we are expecting?

nitishfy: I went through the other test cases as well mentioned in `manifest_test.go` such as _Test_getManifestFromFile_ in which no manifest is checked.

Regarding the tst case that I've written above, it's running well and checking if the Inferred manifest files of specific names are present or not.  Let me know your thoughts.

pkg/model/manifest.go
jLopezbarb: > @jLopezbarb what should be the expected behavior on `destroy` if there is no okteto manifest?

If the name is passed or we can infer it from one of the manifest we should destroy that environment

jLopezbarb:
Hi @nitishfy,

I don't know if adding the commands to helm/kubectl is the best option since that requires to have it installed on the image. But I could deploy the environment using my local setup with helm but then when I run `okteto destroy --remote` the image may not have helm installed by default

nitishfy:
**Update:** The previous changes were affecting the `okteto init` command since the previous changes would have added the new destroy section for `okteto init` command which we don't want as by now.

Hence, I made latest changes with the following approach:

Once the manifest has been detected, we'll check whether the destroy section for that manifest is nil. If it comes out to be nil, we'll add the destroy commands depending on the _type_ of the manifest.

cc @jLopezbarb @teresaromero

cmd/destroy/destroy.go
```diff
@@ -237,11 +238,48 @@ func (dc *destroyCommand) getDestroyer(ctx context.Context, opts *Options) (dest
...
 				Destroy: &model.DestroyInfo{},
 			}
+			return nil, err
+		}
+
+		// add the commands for the destroy section depending upon which type of manifest it is
+		if manifest.Destroy == nil {
+			if manifest.Type == model.ChartType {
+				chartPath, _ := discovery.GetHelmChartPath(opts.ManifestPath)
+				chartPath, _ = filepath.Rel(opts.ManifestPath, chartPath)
```
nitishfy, line 255:
```suggestion
				chartPath, _ = filepath.Rel(opts.ManifestPath, chartPath)
```
The reason we're not checking the error here is that these errors would have already got detected while trying to deploy the application [here](https://github.com/okteto/okteto/blob/master/pkg/model/manifest.go#L518).  _The application won't be destroyed if it has not been deployed on Okteto cloud initially. Hence a user needs to deploy the application first, if the error is found it'll automatically be returned and no destroy section will be executed. When the context gets over to destroy section after being deployed, there will be no errors expected._

jLopezbarb (approved):

cmd/destroy/destroy.go
```diff
@@ -265,19 +266,27 @@ func (dc *destroyCommand) getDestroyer(ctx context.Context, opts *Options) (dest
 		}
 		runInRemote := !isRemote && (destroyImage != "" || opts.RunInRemote)
 
+		// destroy section should not be nil whether the command is executed remotely or locally
+		// if the command is executed remotely and destroy section is nil, this will cause a panic
+		// Hence consider the command locally for successful execution.
+		// This will make sure all the resources are deleted.
+		if runInRemote == true && manifest.Destroy == nil {
+			runInRemote = false
+		}
```
AdrianPedriza, line 275:
```suggestion
		if manifest.Destroy == nil {
			runInRemote = false
		}
```

AdrianPedriza: maybe redundant if `runInRemote == false` but I think is clearer

nitishfy: I tested it already. This will cause a panic to `okteto init` command. 

We need to have destroy section not as nil to execute the destroy command remotely. This destroy section was added automatically in case of valid manifests before checking whether we are running the command remotely.
Hence, if we find that even after executing the command remotely, the destroy section is nil, this means there is a chance of panic being raised. Hence consider this as the local case and set the value to false.

nitishfy: **Update:** we've a separate issue for panic of _okteto init_ command. The above changes will cause no destroy panic. We can tackle the _okteto init_ panic in other issue.
cc @jLopezbarb

nitishfy: Changes applied. Resolved!

AdrianPedriza: on second thought, let's avoid adding logic in this function, check my last comment :)

AdrianPedriza (changes requested):
requesting changes about run command in remote instead locally

nitishfy:
> In general I think we are using the wrong approach if we do the destroy locally. The user is identifying where to do it, taking into account the advantages of doing it remotely, and we are using a destroy from local. I think what we should do in this case is to fix the panic when creating a remote destroyer. That is to say, when creating a remote destroyer we should pass an empty image to use the default one https://github.com/okteto/okteto/blob/master/cmd/destroy/remote.go#L113. In this way we continue to apply the destroy remotely and we do not introduce new logic to the getDestroyer (prone to introduce regressions).

The changes would run exactly as remotely if the Inferred manifest is spotted. However, in the case when the manifest is not found, we need to make sure that Destroy section is not nil. For manifest not found, the destroy section is always nil which will not execute the command remotely and won't delete the resources. In order to delete the resources, the command is shifted to locally. As an end user, i think we'll be using a correct manifest file name. LMK your thoughts?

AdrianPedriza:
> > In general I think we are using the wrong approach if we do the destroy locally. The user is identifying where to do it, taking into account the advantages of doing it remotely, and we are using a destroy from local. I think what we should do in this case is to fix the panic when creating a remote destroyer. That is to say, when creating a remote destroyer we should pass an empty image to use the default one https://github.com/okteto/okteto/blob/master/cmd/destroy/remote.go#L113. In this way we continue to apply the destroy remotely and we do not introduce new logic to the getDestroyer (prone to introduce regressions).
> 
> The changes would run exactly as remotely if the Inferred manifest is spotted. However, in the case when the manifest is not found, we need to make sure that Destroy section is not nil. For manifest not found, the destroy section is always nil which will not execute the command remotely and won't delete the resources. In order to delete the resources, the command is shifted to locally. As an end user, i think we'll be using a correct manifest file name. LMK your thoughts?

but at the end we are running the command locally, right? even if the user is setting destroy option to remote. I think just setting `manifest.destroy.Image` ([here](https://github.com/okteto/okteto/blob/master/cmd/destroy/remote.go#L96)) would be enough to run destroy command remotely

nitishfy:
> but at the end we are running the command locally, right? even if the user is setting destroy option to remote. I think just setting `manifest.destroy.Image` ([here](https://github.com/okteto/okteto/blob/master/cmd/destroy/remote.go#L96)) would be enough to run destroy command remotely

This works. However, what about the panic for _okteto init_ command?

AdrianPedriza:
> > but at the end we are running the command locally, right? even if the user is setting destroy option to remote. I think just setting `manifest.destroy.Image` ([here](https://github.com/okteto/okteto/blob/master/cmd/destroy/remote.go#L96)) would be enough to run destroy command remotely
> 
> This works. However, what about the panic for _okteto init_ command?

is that panic still happen if we don't introduce logic into `getDestroyer` func and just set image to `""`? if it is not related to the introduced changes I'd create a new issue for that bug

nitishfy:
> is that panic still happen if we don't introduce logic into `getDestroyer` func and just set image to `""`? if it is not related to the introduced changes I'd create a new issue for that bug

Playing with _movies-with-helm_ for this. _ok init_ Although causes a panic, however, it creates the `okteto.yml` file as well.
Screenshot from 2023-07-04 22-19-42
image

We can create a separate issue for this. In this repo, i just modified the existing okteto.yml file to change it's name to demo.yml and tried changes there. Here's something more, _ok init_ command don't cause panic when we have a valid manifest path (i tried on the repository that has been mentioned in the [issue description](https://github.com/okteto/okteto/issues/3656).

cmd/destroy/remote.go
```diff
@@ -93,7 +106,7 @@ func newRemoteDestroyer(manifest *model.Manifest) *remoteDestroyCommand {
 	builder := remoteBuild.NewBuilderFromScratch()
 	return &remoteDestroyCommand{
 		builder:              builder,
-		destroyImage:         manifest.Destroy.Image,
+		destroyImage:         "",
```
AdrianPedriza, line 109: if we set this always to `""` we are not allowing to set a destroy image in manifest. A solution would be something like: 
`````go

func newRemoteDestroyer(manifest *model.Manifest) *remoteDestroyCommand {
	fs := afero.NewOsFs()
	builder := remoteBuild.NewBuilderFromScratch()
	if manifest.Destroy == nil {
		manifest.Destroy = &model.DestroyInfo{}
	}
	return &remoteDestroyCommand{
		builder:              builder,
		destroyImage:         manifest.Destroy.Image,
		fs:                   fs,
		workingDirectoryCtrl: filesystem.NewOsWorkingDirectoryCtrl(),
		temporalCtrl:         filesystem.NewTemporalDirectoryCtrl(fs),
		manifest:             manifest,
		registry:             builder.Registry,
		clusterMetadata:      fetchClusterMetadata,
	}
}
`````

**Comment by AdrianPedriza** [comment]
> > is that panic still happen if we don't introduce logic into `getDestroyer` func and just set image to `""`? if it is not related to the introduced changes I'd create a new issue for that bug
> 
> Playing with _movies-with-helm_ for this. _ok init_ causes a panic, however, it creates the `okteto.yml` file as well. Screenshot from 2023-07-04 22-19-42 image
> 
> We can create a separate issue for this. In this repo, i just modified the existing okteto.yml file to change it's name to demo.yml and tried changes there. Here's something more, _ok init_ command don't cause panic when we have a valid manifest path (i tried on the repository that has been mentioned in the [issue description](https://github.com/okteto/okteto/issues/3656).

nice research! can you create an issue with adding all this info? :)

File: cmd/destroy/remote.go (hunk 6 above)
Comment (nitishfy) [reply cmd/destroy/remote.go]: Good one! I have applied those changes here.

I think one way of doing the same another way would have been initializing the DestroInfo at the time when the manifest has been found but it'd consist of more code. This works best.

File: cmd/destroy/destroy.go [hunk 7]
```diff
@@ -237,11 +235,14 @@ func (dc *destroyCommand) getDestroyer(ctx context.Context, opts *Options) (dest
...
 		if err != nil {
-			// Log error message but application can still be deleted
+			// error will only be available when there is no valid manifest found
+			// hence, we need not to destroy because the manifest would never be deployed
+			// Return the error message
 			oktetoLog.Infof("could not find manifest file to be executed: %s", err)
 			manifest = &model.Manifest{
 				Destroy: &model.DestroyInfo{},
 			}
+			return nil, err
```
AdrianPedriza, line 245:
```suggestion
			// error will only be available when there is no valid manifest found
			// hence, we need not to destroy because the manifest would never be deployed
			// Return the error message
			oktetoLog.Infof("could not find manifest file to be executed: %s", err)
			manifest = &model.Manifest{
				Destroy: &model.DestroyInfo{},
			}
			return nil, err
```
I think that for now this change is not necessary. Let's keep the changes to a minimum to avoid possible regressions.

nitishfy:
> nice research! can you create an issue adding all this info? :)

Yes for sure, here's the [issue](https://github.com/okteto/okteto/issues/3791).

cmd/destroy/destroy.go
nitishfy: I understand, but how would this look that you've not deployed anything on Okteto cloud and you don't have a valid manifest as well yet the `okteto destroy` command successfully destroys without having any manifest file.

AdrianPedriza: if we can't get the destroy commands from the manifest we use the deploy-by label as well as the configmap belonging to the dev environment. With your change, we would not be removing the dev environment but showing an error. However, this change is not related to the one reported in the issue. We should stick to the definition of the [scope of a PR](https://github.com/okteto/okteto/blob/master/docs/code-review.md#pr-scope) in this case to facilitate the review and possible resolution of future bugs.

AdrianPedriza: `okteto deploy` cannot be executed due to there is no manifest detected so `okteto` doesn't know how to deploy the environment. However, `okteto destroy` although it does not detect it either, it performs a destroy action removing all resources within namespace which has the proper `deployed-by` label. After that, `destroy` removes also configmap related to that environment. For that reason, `destroy` doesn't need a manifest to use the default `destroy`. Maybe we could improve our logs to let to know the user what happened clearly. If you think that we could improve our dev experience in this case please feel free to fill a new issue :smile:

nitishfy: I see, that's how it's working underneath. Thanks for explaining! I shall mark these changes now.

nitishfy: Resolved!

AdrianPedriza (approved):
Great job! thank you for your contribution and dedication 👍 @nitishfy

jLopezbarb (approved):

PR #3723, opened 2023-06-19
Don't copy ca-certs on deploy image

pchico83:
Since we pass the [right certificate](https://github.com/okteto/okteto/pull/3618) on `okteto deploy --remote` , there is no need to copy `ca-certificates` on the deploy image at runtime.
The user is responsible of adding the right certificates to her image if needed.

### DIFF
--- a/cmd/deploy/remote.go
+++ b/cmd/deploy/remote.go
@@ -48,15 +48,11 @@ const (
 	dockerfileTemplate     = `
 FROM {{ .OktetoCLIImage }} as okteto-cli
 
-FROM alpine as certs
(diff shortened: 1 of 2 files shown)


teresaromero (approved):

andreafalzetti:
@pchico83 is there a way to validate/test this PR? How did you test it?

andreafalzetti (approved):

pchico83:
> @pchico83 is there a way to thoroughly validate/test this PR? How did you test it?
> 
> I've tested this branch with a couple of sample repos and they worked ✅

I have tested it using a cluster with self-signed certificates

PR #3725, opened 2023-06-19
Use ARG instead of ENV to pass metadata for "--remote"

pchico83:
This PR refactors the `Dockerfile` generation for `--remote` to pass information using `ARG`instead of `ENV`. This way, sensible information like the okteto token, is not in plain in the dynamic Dockerfile.

### DIFF
--- a/cmd/destroy/remote.go
+++ b/cmd/destroy/remote.go
@@ -31,7 +31,7 @@ import (
 
 const (
 	templateName           = "destroy-dockerfile"
-	dockerfileTemporalNane = "deploy"
+	dockerfileTemporalNane = "Dockerfile.destroy"
 	oktetoDockerignoreName = ".oktetodeployignore"
 	dockerfileTemplate     = `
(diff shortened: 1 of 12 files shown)


cmd/destroy/remote.go
```diff
@@ -48,50 +48,42 @@ COPY --from=certs /etc/ssl/certs /etc/ssl/certs
 COPY --from=installer /app/bin/* /okteto/bin/
 COPY --from=okteto-cli /usr/local/bin/* /okteto/bin/
 
-{{range $key, $val := .OktetoBuildEnvVars }}
```
teresaromero, line 51: we are removing the envs then?

pchico83: Based on my experiments, the remote `okteto deploy` also computes these envvars, so it doesn't look like they bring any value here. Are you aware of any optimization that skips the computation of the envvars if they are already defined?

teresaromero (approved):

PR #3736, opened 2023-06-20
Fix silent error when build images as part of "okteto deploy"

pchico83:
If `buildImages` is failing as part of `okteto deploy`, we are silencing the error if updating the configmap doesn't fail.

### DIFF
--- a/cmd/deploy/deploy.go
+++ b/cmd/deploy/deploy.go
@@ -356,7 +356,10 @@ func (dc *DeployCommand) RunDeploy(ctx context.Context, deployOptions *Options) 
 	}
 
 	if err := buildImages(ctx, dc.Builder.Build, dc.Builder.GetServicesToBuild, deployOptions); err != nil {
-		return dc.CfgMapHandler.updateConfigMap(ctx, cfg, data, err)
+		if errStatus := dc.CfgMapHandler.updateConfigMap(ctx, cfg, data, err); errStatus != nil {
+			return errStatus
+		}
+		return err
 	}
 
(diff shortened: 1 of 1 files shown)


cmd/deploy/deploy.go
```diff
@@ -356,7 +356,10 @@ func (dc *DeployCommand) RunDeploy(ctx context.Context, deployOptions *Options)
 	}
 
 	if err := buildImages(ctx, dc.Builder.Build, dc.Builder.GetServicesToBuild, deployOptions); err != nil {
-		return dc.CfgMapHandler.updateConfigMap(ctx, cfg, data, err)
+		if errStatus := dc.CfgMapHandler.updateConfigMap(ctx, cfg, data, err); errStatus != nil {
+			return errStatus
+		}
+		return err
```
pchico83, line 362: this returns the original error when `updateConfigMap` doesn't fail

cmd/deploy/deploy.go
```diff
@@ -407,8 +410,8 @@ func (dc *DeployCommand) RunDeploy(ctx context.Context, deployOptions *Options)
 		data.Status = pipeline.DeployedStatus
 	}
 
-	if err := dc.CfgMapHandler.updateConfigMap(ctx, cfg, data, err); err != nil {
-		return err
+	if errStatus := dc.CfgMapHandler.updateConfigMap(ctx, cfg, data, err); errStatus != nil {
```
teresaromero, line 413: is this rename for better reading? `err` is local variable of the if statement, so this changes just the name of the err

teresaromero (approved):

PR #3742, opened 2023-06-21
Update hybrid analytics

teresaromero:
# Proposed changes

Fixes https://github.com/okteto/app/issues/6684

- update Track Metadata structs for Up & Exec compliant with internal documentation
- refactor methods to get metadata, Dev and Manifest new methods to extract information


teresaromero:
Some of the deepsource comments are not code related with this PR. I won't be fixing them as they are just raised because i changed the file.

pkg/model/manifest.go
```diff
@@ -1360,3 +1360,19 @@ func (m *Manifest) GetStack() *Stack {
...
+
+func (m *Manifest) HasDependenciesSection() bool {
+	return m.IsV2 && len(m.Dependencies) > 0
+}
+
+func (m *Manifest) HasBuildSection() bool {
+	return m.IsV2 && len(m.Build) > 0
+}
+
+func (m *Manifest) HasDeploySection() bool {
```
andreafalzetti, line 1372: Do you think could we add some unit tests for this function? I've just noticed that the old implementation causes a panic in this block and it's affecting `2.16.x` at least.

It'd be good to make sure it's not happening again in the future

teresaromero: I've added test for 3 helpers and updated the functions to contemplate the case of manifest being nil

cmd/up/up.go
```diff
@@ -617,16 +617,12 @@ func (up *upContext) start() error {
 	pidFileCh := make(chan error, 1)
 
 	analytics.TrackUp(analytics.TrackUpMetadata{
-		IsInteractive:          up.getInteractive(),
+		Mode:                   up.Dev.GetMode(),
 		IsOktetoRepository:     utils.IsOktetoRepo(),
-		IsV2:                   up.Manifest.IsV2,
```
rlamana, line 622: I'd keep the v2 parameter, since older version of the CLI are going to keep sending it, we will not have an easy way to distinguish old ones and new ones without it. I understand we are trying to make it compliant with the spec, but even if this was not added there I would keep it just in case it affects our dashboards in mixpaneI. Is there any other reason why we are removing it?

So I gues what I'm trying to say is to add all those that are defined in the dictionary, and at the end keep old those that were already there for retro-compatibility.

teresaromero: i've updated, so i kept the old ones and added the new ones in both exec and up

rlamana (approved):
Looks good to me from product perspective and the parameter we expect, I'm not reviewing from a code point of view.

andreafalzetti (approved):
# 🚀

pkg/model/dev.go
```diff
@@ -529,6 +529,14 @@ func (dev *Dev) IsHybridModeEnabled() bool {
 	return dev.Mode == constants.OktetoHybridModeFieldValue
 }
 
+func (dev *Dev) GetMode() string {
```
pchico83, line 532: could we set the default value when reading the manifest? I think there is already a `setDefaults` method.
That way we make sure if someone access `dev.Mode` directly it will have the right value and we don't need the `GetMode` method

teresaromero: Done!

pchico83 (approved):

PR #3747, opened 2023-06-22
Remove git credentials from debug logs

### LINKED ISSUE #3560: remove git credentials from logs
**Describe the bug**
If I run a command using `-l=debug`, logs will display git credentials. In order to debug issues is not safe to display this information with them. This credential are in the repo URL -> https://github.com/okteto/okteto/blob/master/pkg/devenvironment/name.go#L85

teresaromero:
# Proposed changes

Fixes #3560 

In order to display de repo url without user information, new method attached to `Repository` and use it for displying the logs.

### DIFF
--- a/pkg/devenvironment/name.go
+++ b/pkg/devenvironment/name.go
@@ -82,12 +82,12 @@ func (n NameInferer) InferNameFromDevEnvsAndRepository(ctx context.Context, repo
 		optsRepo := repository.NewRepository(repoURL)
 		cmapRepo := repository.NewRepository(repo)
 		if !optsRepo.IsEqual(cmapRepo) {
-			oktetoLog.Infof("configmap %s with repo %s doesn't match with found repo %s", cmap.Name, repo, repoURL)
+			oktetoLog.Infof("configmap %s with repo %s doesn't match with found repo %s", cmap.Name, cmapRepo.GetAnonymizedRepo(), repoURL)
 			continue
 		}
 
 		if filename := cmap.Data["filename"]; filename != manifestPath {
-			oktetoLog.Infof("configmap %s with manifest %s doesn't match with provided manifest %s", filename, repo, manifestPath)
+			oktetoLog.Infof("configmap %s with manifest %s doesn't match with provided manifest %s", filename, cmapRepo.GetAnonymizedRepo(), manifestPath)
 			continue
 		}
 
--- a/pkg/repository/repository.go
+++ b/pkg/repository/repository.go
@@ -44,8 +44,9 @@ type repositoryURL struct {
 
 // String is a custom implementation for the url where User is removed
 func (r repositoryURL) String() string {
-	r.URL.User = nil
-	return r.URL.String()
+	repo := r.URL
+	repo.User = nil
+	return repo.String()
 }
 
 func getURLFromPath(path string) repositoryURL {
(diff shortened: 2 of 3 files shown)


andreafalzetti (approved):
LGTM

pkg/repository/repository.go
```diff
@@ -89,3 +89,18 @@ func (r Repository) IsEqual(otherRepo Repository) bool {
 func cleanPath(path string) string {
 	return strings.TrimSuffix(strings.TrimPrefix(path, "/"), ".git")
 }
+
+// GetAnonymizedRepo returns a clean repo url string without sensible information
+func (r Repository) GetAnonymizedRepo() string {
```
pchico83, line 94: would be possible to overwrite the `String()` method? or would it have other side effects? if it's possible, we would make sure the credentials are never printed

teresaromero: you mean url.URL.String() in this package? the url prop for Repository is private so it will need a method to expose it, not sure i understand the benefits

pchico83: The current solution is error-prone in the sense that it's very easy to print `url.URL.String()` for a new developer and expose the credentials. If `Repository` has a `String` method that masks the credentials, no one will be able to print the credentials. Ideally, that means `URL` will be private

teresaromero: `Repository` struct has already private url, so this is the only method exposing the url, and removing the user information if it has.

I will take a look and make some changes

teresaromero: updated code with changes

pkg/devenvironment/name.go
```diff
@@ -82,12 +82,12 @@ func (n NameInferer) InferNameFromDevEnvsAndRepository(ctx context.Context, repo
 		optsRepo := repository.NewRepository(repoURL)
 		cmapRepo := repository.NewRepository(repo)
 		if !optsRepo.IsEqual(cmapRepo) {
-			oktetoLog.Infof("configmap %s with repo %s doesn't match with found repo %s", cmap.Name, repo, repoURL)
+			oktetoLog.Infof("configmap %s with repo %s doesn't match with found repo %s", cmap.Name, cmapRepo.GetAnonymizedRepo(), repoURL)
```
pchico83, line 85: I was thinking more about something like:
```
oktetoLog.Infof("configmap %s doesn't match with found repo %s", cmap, repoURL)
```
and configmap would have a `String()` method that shows what's needed, for example, the configmap name, repo URL and okteto manifest path, without credentials.
But take it as a suggestion, whatever works better for you

pchico83 (approved):

pkg/repository/repository.go
```diff
@@ -38,13 +38,31 @@ type repositoryInterface interface {
...
-// NewRepository creates a repository controller
-func NewRepository(path string) Repository {
-	repoUrl, err := giturls.Parse(path)
+type repositoryURL struct {
+	url.URL
+}
+
+// String is a custom implementation for the url where User is removed
+func (r repositoryURL) String() string {
+	r.URL.User = nil
```
AdrianPedriza, line 47: it is safe to do it? if we remove the user from the repo could not be affected future uses about the repository object ? could we work with a copy for this case ?

teresaromero: makes sense, updating the PR

teresaromero: @AdrianPedriza i've added a test for this function

AdrianPedriza (approved):

PR #3749, opened 2023-06-23
Fix integration hybrid

teresaromero:
# Proposed changes

When introducing https://github.com/okteto/okteto/pull/3683 at master the integration test failed.
The cause identified where 2:
- nil pointer when Process is nil when shutingdown
- Foreground property at SysProcAttr (after long investigation i found that using this property caused the binary for the shell to not work properly)

This changes have been also tested with the hybrid feature, this means the functionality is still working.

Also added logs in case of error for the Hybrid integration as the error shown at the test was not clear enough

### DIFF
--- a/cmd/up/exec_unix.go
+++ b/cmd/up/exec_unix.go
@@ -34,7 +34,6 @@ func (he *hybridExecutor) GetCommandToExec(ctx context.Context, cmd []string) (*
 
 	c.SysProcAttr = &syscall.SysProcAttr{
 		Setpgid:    true,
-		Foreground: true,
 	}
 
 	return c, nil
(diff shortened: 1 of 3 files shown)


cmd/up/exec_unix.go
```diff
@@ -34,7 +34,6 @@ func (he *hybridExecutor) GetCommandToExec(ctx context.Context, cmd []string) (*
 
 	c.SysProcAttr = &syscall.SysProcAttr{
 		Setpgid:    true,
-		Foreground: true,
```
andreafalzetti, line 37: > Foreground property at SysProcAttr (after long investigation i found that using this property caused the binary for the shell to not work properly)

Nice catch, I am interested to understand what wasn't working properly?

andreafalzetti (approved):

PR #3753, opened 2023-06-26
Always sync okteto manifest

### LINKED ISSUE #3727: Always sync the okteto manifest for `--remote`
**Describe the bug**
I have a `.oktetodeployignore` file that ignores my okteto manifest. `okteto deploy --remote` and `okteto destroy --remote` don't fail, but they don't execute my `deploy|destroy` commands.

**Expected behavior**
My okteto manifest should be always synched.

**Additional context**
We can always add this line to the `.dockerignore` file
```
![path-to-my-okteto-manifest]
```

teresaromero:
# Proposed changes

Fixes #3727 

When having an `.oktetodeployignore` file which is ignoring the okteto manifest used for deployment, the `remote` deploy is unable to use the file as this is ignore and the dockerfile used for remote does not include the manifest.

To fix this:
- discovery of the manifest file name, in case the file flag is used, this is the file included at the ingore file
- exclude this at the end of the existent `.oktetodeployignore` 
- unify creation of dockerignore into new `remote` pkg so deploy and destroy share config

## Tests

Using [this](https://github.com/teresaromero/hybrid-test) repository and including the manifest file at a `.oktetodeployignore` when running `okteto deploy --remote` the command deploys with the manifest. Without the fix, the deploy remote was inferring the deployment from the `k8s.yaml` found. 

Same when using `okteto destroy --remote`

Also, adding a deploy/destroy command with echo "hello" and remote command should show that command to be sure the manifest is being used


cmd/deploy/remote_test.go
```diff
@@ -350,7 +350,7 @@ func TestCreateDockerignore(t *testing.T) {
 			rdc := remoteDeployCommand{
 				fs: fs,
 			}
-			err := rdc.createDockerignore(tt.config.wd, tempDir)
+			err := rdc.createDockerignore(tt.config.wd, tempDir, "")
```
andreafalzetti, line 353:
```suggestion
			err := rdc.createDockerignore(tt.config.wd, tempDir, tt.config.manifestPathFlag)
```

WDYT if we add a unit test that at least exec the new logic?

```
{
			name: "with okteto manifest",
			config: config{
				wd:               dockerignoreWd,
				manifestPathFlag: "okteto.yml",
			},
		},

...

type config struct {
		wd               string
		manifestPathFlag string
	}
```

andreafalzetti:
@teresaromero I am not familiar with `.oktetodeployignore`. Do you mind sharing steps on how to reproduce the bug witht he current version and steps to test the new build?

This is what I've tried:
1. Clone https://github.com/teresaromero/hybrid-test
2. Create `.oktetodeployignore` and insert the following content: "okteto.yml"
3. Build this branch
4. Run `okteto deploy --remote`

However I am not sure what to look for.

teresaromero:
> @teresaromero I am not familiar with `.oktetodeployignore`. Do you mind sharing steps on how to reproduce the bug witht he current version and steps to test the new build?
> 
> This is what I've tried:
> 
> 1. Clone https://github.com/teresaromero/hybrid-test
> 2. Create `.oktetodeployignore` and insert the following content: "okteto.yml"
> 3. Build this branch
> 4. Run `okteto deploy --remote`
> 
> However I am not sure what to look for.

yes! i updated the test section at description. Adding this file to your repository, prevents the dockerfile to include any file on this ignorefile to the build, so adding `okteto.yaml` will prevent the remote dockerfile to include the manifest.

What happens with this is that the deploy running remotly does not have a manifest, so it tries to infer with the files it has. Doing this with a repo with a `k8s.yaml` does the trick as the manifest is inferred from this file, but adding any other command to the deploy section or the destroy, you will see (without the fix) that this commands are not the ones running, but the inferred ones.
With the fix, the commands run by remote should be the ones on your manifest

cmd/destroy/remote.go
```diff
@@ -242,7 +243,25 @@ func (rd *remoteDestroyCommand) createDockerignore(cwd, tmpDir string) error {
 			return err
 		}
 	}
-	return afero.WriteFile(rd.fs, fmt.Sprintf("%s/%s", tmpDir, ".dockerignore"), dockerignoreContent, 0600)
+
+	// write the content into the .dockerignore used for building the remote image
+	filename := fmt.Sprintf("%s/%s", tmpDir, ".dockerignore")
```
jLopezbarb, line 248: Maybe not to address on this PR but we need to have a common place for deploy and destroy

pkg/discovery/discovery.go
```diff
@@ -0,0 +1,34 @@
...
+	if chartPath, err := GetHelmChartPath(cwd); err == nil {
+		return chartPath
+	}
+	if k8sManifestPath, err := GetK8sManifestPath(cwd); err == nil {
+		return k8sManifestPath
+	}
+	return ""
+}
+
+func FindManifestName(cwd string) string {
```
jLopezbarb, line 24: Should we add a unit test for this?

cmd/destroy/remote.go
teresaromero: I believe the whole remote thing can be in a common place as we do have a lot of duplicates

pkg/discovery/discovery.go
```diff
@@ -0,0 +1,34 @@
+package discovery
+
+import "path/filepath"
+
+func getInferredManifestFilePath(cwd string) string {
```
AdrianPedriza, line 5: if I use a different name than the ones in [this list](https://github.com/okteto/okteto/blob/master/pkg/discovery/manifest.go#L24) I'm still ignoring my okteto manifest. To test it:
- create okteto manifest with name `okteto-test.yml`
- add to `.oktetodeployignore` an entry for `okteto-test.yml`

Should we infer the manifest base on the content of the file (maybe checking if yaml file contains main keys as `dev`, `build`or `deploy`) instead rely on the name? Following the principle that my manifest can have any name respecting the accepted format I think it would be optimal.

teresaromero: when we have a custom name for the manifest, we have to pass the -f flag, as the actual infer of the manifest is done with this lists of files.

As far as i know this would change behavour and is not scope of this PR to change the infer manifest behavour

AdrianPedriza (approved):

jLopezbarb (approved):

pkg/discovery/discovery.go
teresaromero: iupdated and added test

cmd/destroy/remote.go
teresaromero: i've added changes to unify at least the dockerignore logic

cmd/deploy/remote_test.go
teresaromero: updated and added test

pkg/remote/remote_test.go
```diff
@@ -0,0 +1,86 @@
...
+
+import (
+	"path/filepath"
+	"testing"
+
+	"github.com/spf13/afero"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestCreateDockerignore(t *testing.T) {
```
andreafalzetti, line 11:
```suggestion
func TestCreateDockerignoreFileWithFilesystem(t *testing.T) {
```

cmd/destroy/remote.go
```diff
@@ -228,21 +228,8 @@ func (rd *remoteDestroyCommand) createDockerfile(tempDir string, opts *Options)
...
-		}
-
-	} else {
-		dockerignoreContent, err = afero.ReadFile(rd.fs, dockerignoreFilePath)
-		if err != nil {
-			return err
-		}
-	}
-	return afero.WriteFile(rd.fs, fmt.Sprintf("%s/%s", tmpDir, ".dockerignore"), dockerignoreContent, 0600)
+func (rd *remoteDestroyCommand) createDockerignore(cwd, tmpDir, manifestPathFlag string) error {
```
jLopezbarb, line 231: do we really need this func? Can we call directly `remote.CreateDockerignoreFileWithFilesystem(cwd, tmpDir, manifestPathFlag, rd.fs)`?

jLopezbarb (approved):

pkg/remote/remote.go
```diff
@@ -0,0 +1,52 @@
...
+
+const (
+	oktetoDockerignoreName = ".oktetodeployignore"
+)
+
+func CreateDockerignoreFileWithFilesystem(cwd, tmpDir, manifestPathFlag string, fs afero.Fs) error {
+	dockerignoreContent := []byte(``)
+	dockerignoreFilePath := filepath.Join(cwd, oktetoDockerignoreName)
+	if _, err := fs.Stat(dockerignoreFilePath); err != nil {
+		if !errors.Is(err, os.ErrNotExist) {
```
andreafalzetti, line 21: As we discussed this if needs an update, it's not going in this condition when passing a non existing `cwd`.

```suggestion
		if os.IsNotExist(err) {
```

I've found this one works, but please double check 🙏

andreafalzetti: This is a test case I'd add:

```go
func TestCreateDockerignoreFileWithFilesystem_ErrNotExist(t *testing.T) {
	tempDir := t.TempDir()
	fs := afero.NewMemMapFs()
	err := CreateDockerignoreFileWithFilesystem("/non-existing-test-folder", tempDir, "okteto.yml", fs)
	assert.ErrorIs(t, err, os.ErrNotExist)
}
```

andreafalzetti: I am not 100% sure of my suggestion because afero seems to return `fs.PathError` for not found, and inspecting the code of `os.IsNotExist` it checks against `fs.ErrNotExist`

andreafalzetti: I'd also check the implementation of `FileExistsWithFilesystem` which seems doing what I was suggesting (using `os.IsNotExist`

teresaromero: I've checked by debugging and the err is wrapping an non exist error. We want to return err only if the err is not a not exist one.

In order to add a test case i am not sure how i can mock afero to return other err ?

andreafalzetti: @teresaromero if we pass a non existing path, shouldn't that be a not found error?

AdrianPedriza (approved):

pkg/remote/remote.go
teresaromero: It is a non-exist error. The file does not exist on that path, so we create an empty dockerignore for the build 

Screenshot 2023-07-03 at 10 22 49

andreafalzetti (approved):

PR #3754, opened 2023-06-27
add envs from manifest to local process

AdrianPedriza:
# Proposed changes
- support `envFiles` field for hybrid dev
- add envs from `environment` field into local process

Fixes #[6715](https://github.com/okteto/app/issues/6715)

### DIFF
--- a/cmd/up/exec.go
+++ b/cmd/up/exec.go
@@ -214,6 +214,10 @@ func (eg *envsGetter) getEnvs(ctx context.Context) ([]string, error) {
 
 	envs = append(envs, eg.getDefaultLocalEnvs()...)
 
+	for _, env := range eg.dev.Environment {
+		envs = append(envs, fmt.Sprintf("%s=%s", env.Name, env.Value))
+	}
+
 	return envs, nil
 }
 
--- a/cmd/up/exec.go
+++ b/cmd/up/exec.go
@@ -185,18 +185,6 @@ func newEnvsGetter(hybridCtx *HybridExecCtx) (*envsGetter, error) {
 func (eg *envsGetter) getEnvs(ctx context.Context) ([]string, error) {
 	var envs []string
 
-	configMapEnvs, err := eg.configMapEnvsGetter.getEnvsFromConfigMap(ctx, eg.name, eg.namespace, eg.client)
-	if err != nil {
-		return nil, err
-	}
-	envs = append(envs, configMapEnvs...)
-
-	secretsEnvs, err := eg.secretsEnvsGetter.getEnvsFromSecrets(ctx)
-	if err != nil {
-		return nil, err
-	}
-	envs = append(envs, secretsEnvs...)
-
 	app, err := apps.Get(ctx, eg.dev, eg.namespace, eg.client)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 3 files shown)


cmd/up/exec.go
```diff
@@ -214,6 +214,10 @@ func (eg *envsGetter) getEnvs(ctx context.Context) ([]string, error) {
 
 	envs = append(envs, eg.getDefaultLocalEnvs()...)
 
+	for _, env := range eg.dev.Environment {
```
pchico83, line 217: can we have a tests covering this?

cmd/up/exec.go
```diff
@@ -214,6 +214,10 @@ func (eg *envsGetter) getEnvs(ctx context.Context) ([]string, error) {
 
 	envs = append(envs, eg.getDefaultLocalEnvs()...)
```
teresaromero, line 215: when i tested this last week the envs from the field `environment` where not appearing at the dev container.... have we check this is working?

AdrianPedriza: 1. Since the command is executed locally I don't think we need those values in the dev container. Also when doing the `exec` cmd we get them in the command so I think we cover the cases that require them

cmd/up/exec.go
AdrianPedriza: 928f11577a8c9184d686db338d7998a06473943f

pchico83 (approved):

teresaromero (approved):

cmd/up/exec.go
```diff
@@ -214,6 +214,10 @@ func (eg *envsGetter) getEnvs(ctx context.Context) ([]string, error) {
...
 
+	for _, env := range eg.dev.Environment {
+		envs = append(envs, fmt.Sprintf("%s=%s", env.Name, env.Value))
```
jLopezbarb, line 218: Did we check what happens if the same variable is defined in multiple sections? I mean if I have one environment set in the pod but when I run `okteto up` with a manifest that is overwriting it, is it working as expected?
Could we add a test for that scenario?

AdrianPedriza: then variable that is set using `environment` section is the one used. Same approach described in docs https://www.okteto.com/docs/reference/manifest/#environment-string-optional:~:text=Environment%20variables%20declared%20in%20the%20environment%20section%20override%20these%20values.

jLopezbarb (approved):

cmd/up/exec.go
AdrianPedriza: I will add a test covering it

AdrianPedriza: b2da3c8657915cfb17cd8bba0b05b369fa0a2928 @jLopezbarb test to check overwrites

cmd/up/exec_test.go
```diff
@@ -324,6 +324,63 @@ func TestGetEnvsError(t *testing.T) {
...
+		imageEnvsGetter:     &fakeImageEnvsGetter,
+		getDefaultLocalEnvs: func() []string { return []string{} },
+	}
+	envs, err := eg.getEnvs(ctx)
+	require.NoError(t, err)
+
+	// following docs, if exec.cmd.Env contains duplicate environment keys, only the last value in the
+	// slice for each duplicate key is used so we can check overwrittes by getting last value
+	// in envs list.
+	require.Equal(t, "ENVTOOVERWRITE=FROMMANIFEST", envs[len(envs)-1])
```
jLopezbarb, line 381: Could we ensure that the order is the correct instead of checking just the last one? That way we also cover the other cases.

AdrianPedriza: checking your suggestions I think we don't have any preferred order except those coming from the manifest, so a test checking the order of all of them seems strange to me. I think that with the test we have now we cover the only preference we have. WDYT?

AdrianPedriza: @pchico83 What should be the order of priority between the shipments coming from the configmap, secrets, image pod and manifest?

jLopezbarb: My 2¢ is (being 1 the top priority):
1. manifest
2. pod
3. configmap
4. secrets
5. image
But let's double check before going with an approach

pchico83: +1

AdrianPedriza: 4fbd68b96a114bb496c9cc95116a387ba43cde32

jLopezbarb (approved):

Issue #3589, opened 2023-05-04
SSH forwarding in okteto build


guydomb:
**Is your feature request related to a problem? Please describe.**
Our use case for building Docker images requires that we forward SSH agent into the build context. This is used during build time to resolve project dependencies from private Git repositories. This option is supported by Docker BuildKit and used like so:
```docker build --ssh default -t local/service1:1 -f Dockerfile .```

assuming our Dockerfile makes use of the SSH agent being available in the context of the RUN command:
```RUN --mount=type=ssh do-something-with-ssh```

This approach doesn't work for us with `okteto build` as SSH forwarding isn't supported
```
# okteto.yml
build:
  service1:
    context: .
    dockerfile: Dockerfile
```

**Describe the solution you'd like**
Support for BuildKit's SSH forwarding in Okteto Manifest, similar to how secret mounting are currently supported:
```
# suggested okteto.yml
build:
  service1:
    context: .
    dockerfile: Dockerfile
    ssh: <ssh-mount-id (default: "default")> 
```

**Describe alternatives you've considered**
A partial workaround we're trying involves injecting the invoker SSH key as a Docker build secret and mounting during build time. This is less preferred because it means Dockerfile definition need to refer to the mounted secret explicitly.

**Additional context**
N/A

pchico83:
@jLopezbarb @AdrianPedriza this has implications with `okteto deploy --remote`. 
As part of this issue we should make sure that:
- `okteto deploy --remote` can clone a private git repo in the `installer` if the okteto ssh key is configured
- `okteto deploy --remote` can clone a private git repo executed from local if the user has a valid ssh key to clone the repo. or at least the default one. if the second, we should have a flag in `okteto deploy` to specify the ssh key
- we support `ssh: <ssh-mount-id (default: "default")>` in the build section

PR #3755, opened 2023-06-27
Mount ssh agent in remote deploy

maroshii:
Mounts the ssh agent in the remote deploy command allowing to clone private repos and make authenticated ssh operations.

There are a few things pending (see below) but wanted to open early to get feedback

## Testing

Prior to this change this doesn't work:

```Dockerfile
FROM okteto/pipeline-runner:1.0.3
```
```yaml
build:
  base:
    image: okteto.dev/my-repo:latest

deploy:
  image: okteto.dev/my-repo:latest
  commands:
    - git clone git@github.com:maroshii/okteto-tester.git
```

```bash
$ okteto deploy --remote
 i  Using fran @ okteto.fran.dev.okteto.net as context
 i  Running stage 'git clone git@github.com:maroshii/okteto-tester.git'
Cloning into 'okteto-tester'...
Host key verification failed.
fatal: Could not read from remote repository.
Please make sure you have the correct access rights
and the repository exists.
 x  Error executing command 'git clone git@github.com:maroshii/okteto-tester.git': exit status 128
```

After this change:

```bash
$ okteto deploy --remote
 i  Using fran @ okteto.fran.dev.okteto.net as context
 i  Running stage 'git clone git@github.com:maroshii/okteto-tester.git'
Executing command 'git clone git@github.com:maroshii/okteto-tester.git'...
Cloning into 'okteto-tester'...
Command 'git clone git@github.com:maroshii/okteto-tester.git' successfully executed
 i  Endpoints available:
  - https://hello-world-fran.fran.dev.okteto.net/
 ✓  Development environment 'okteto-tester' successfully deployed
```

## Open Questions

- [x] How should we add known hosts? Currently github.com and gitlab.com are hardcoded but we should support other hosts that may be needed
- [x] Should we automatically mount `$HOME/.ssh/id_rsa` if `SSH_AUTH_SOCK` is empty? I think no but it was listed in the requirement

## Pending

- [x] Add ~same logic `okteto destroy --remote`
- [x] Write tests

### DIFF
--- a/cmd/deploy/deploy.go
+++ b/cmd/deploy/deploy.go
@@ -74,8 +74,6 @@ type Options struct {
 	Timeout    time.Duration
 
 	ShowCTA bool
-
-	SshIdentityKey string
 }
 
 // DeployCommand defines the config for deploying an app
@@ -264,7 +262,6 @@ func Deploy(ctx context.Context) *cobra.Command {
 	cmd.Flags().StringVarP(&options.Namespace, "namespace", "n", "", "overwrites the namespace where the development environment is deployed")
 	cmd.Flags().StringVarP(&options.K8sContext, "context", "c", "", "context where the development environment is deployed")
 	cmd.Flags().StringArrayVarP(&options.Variables, "var", "v", []string{}, "set a variable (can be set more than once)")
-	cmd.Flags().StringVarP(&options.SshIdentityKey, "identity-file", "i", "", "the identity key to use for ssh sessions")
 	cmd.Flags().BoolVarP(&options.Build, "build", "", false, "force build of images when deploying the development environment")
 	cmd.Flags().BoolVarP(&options.Dependencies, "dependencies", "", false, "deploy the dependencies from manifest")
 	cmd.Flags().BoolVarP(&options.RunWithoutBash, "no-bash", "", false, "execute commands without bash")
--- a/cmd/deploy/deploy.go
+++ b/cmd/deploy/deploy.go
@@ -74,6 +74,8 @@ type Options struct {
 	Timeout    time.Duration
 
 	ShowCTA bool
+
+	SshIdentityKey string
 }
 
 // DeployCommand defines the config for deploying an app
@@ -262,6 +264,7 @@ func Deploy(ctx context.Context) *cobra.Command {
 	cmd.Flags().StringVarP(&options.Namespace, "namespace", "n", "", "overwrites the namespace where the development environment is deployed")
 	cmd.Flags().StringVarP(&options.K8sContext, "context", "c", "", "context where the development environment is deployed")
 	cmd.Flags().StringArrayVarP(&options.Variables, "var", "v", []string{}, "set a variable (can be set more than once)")
+	cmd.Flags().StringVarP(&options.SshIdentityKey, "identity-file", "i", "", "the identity key to use for ssh sessions")
 	cmd.Flags().BoolVarP(&options.Build, "build", "", false, "force build of images when deploying the development environment")
 	cmd.Flags().BoolVarP(&options.Dependencies, "dependencies", "", false, "deploy the dependencies from manifest")
 	cmd.Flags().BoolVarP(&options.RunWithoutBash, "no-bash", "", false, "execute commands without bash")
--- a/cmd/deploy/remote.go
+++ b/cmd/deploy/remote.go
@@ -100,8 +100,11 @@ type remoteDeployCommand struct {
 	temporalCtrl         filesystem.TemporalDirectoryInterface
 	clusterMetadata      func(context.Context) (*types.ClusterMetadata, error)
 
-	// defaults to SSH_AUTH_SOCK. Provided mostly for testing
+	// sshAuthSockEnvvar is the default for SSH_AUTH_SOCK. Provided mostly for testing
 	sshAuthSockEnvvar string
+
+	// knownHostsPath  is the default known_hosts file path. Provided mostly for testing
+	knownHostsPath string
 }
 
 // newRemoteDeployer creates the remote deployer from a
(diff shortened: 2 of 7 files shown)


cmd/deploy/remote.go
```diff
@@ -69,7 +69,8 @@ WORKDIR /okteto/src
 ARG {{ .GitCommitArgName }}
 ARG {{ .InvalidateCacheArgName }}
 
-RUN okteto deploy --log-output=json --server-name="${{ .InternalServerName }}" {{ .DeployFlags }}
+RUN ssh-keyscan {{ join .KnownSshHosts " " }} >> ~/.ssh/known_hosts
```
pchico83, line 72: @maroshii could we also apply the changes to `okteto destroy`. All the remote code is duplicated between `cmd/deploy` and `cmd/destroy`.
@jLopezbarb I think we should refactor ^^ in common interfaces, wdyt?

ifbyol, line 72: If we are using local ssh key, would it make sense to copy the local known_hosts file instead of generating it from a list we provide (with the implications of it referring to one of the questions you posted in the description like how users would be able to configure it)? cc @pchico83 @jpf-okteto

jLopezbarb: Yes, we have to refactor so both codes can share as much as possible. Something that we'll look into it on next sprints

cmd/deploy/deploy.go
```diff
@@ -262,6 +264,7 @@ func Deploy(ctx context.Context) *cobra.Command {
 	cmd.Flags().StringVarP(&options.Namespace, "namespace", "n", "", "overwrites the namespace where the development environment is deployed")
 	cmd.Flags().StringVarP(&options.K8sContext, "context", "c", "", "context where the development environment is deployed")
 	cmd.Flags().StringArrayVarP(&options.Variables, "var", "v", []string{}, "set a variable (can be set more than once)")
+	cmd.Flags().StringVarP(&options.SshIdentityKey, "identity-file", "i", "", "the identity key to use for ssh sessions")
```
jLopezbarb, line 267: Why do we need a new flag? Could we use the one locally and pass it through an arg to the remote?

 I think adding a new flag with a shortcut that is only going to be used on this command can led to future problems. WDYT?

maroshii: Users should be able to provide a different identity key to authenticate:

```bash
okteto deploy --remote -i "$HOME/.ssh/id_rsa"
```

I agree it's confusing since this flag will only work with `--remote` and we should probably add a guard for that and fail if it's not provided.

Can you think of a different way of achieving/doing this?

pchico83: `-i "$HOME/.ssh/id_rsa"` would probably break replicability between developers.
My 2$ is to ship the first version without it. A workaround, if someone needs this, is to add the private image to the deploy `image` itself. But my understanding is that the ssh agent should cover most of the use cases

jLopezbarb: I don't know if we are using it in any other place rather than the creation of the Dockerfile, at least is the only place I can see it. 

We can always use the environment variables/args/secrets to inject things into buildkit to skip adding the flag

maroshii:
Once we validate that this is OK, I'll replicate it for destroy. Should I do it in different PR or same PR?

cmd/deploy/remote.go
```diff
@@ -86,6 +90,7 @@ type dockerfileTemplateProperties struct {
 	GitCommitArgName       string
 	InvalidateCacheArgName string
 	DeployFlags            string
+	KnownSshHosts          []string
```
pchico83, line 93: do we need this?

cmd/deploy/remote.go
```diff
@@ -169,6 +180,20 @@ func (rd *remoteDeployCommand) deploy(ctx context.Context, deployOptions *Option
...
+
+	sshSock := os.Getenv(rd.sshAuthSockEnvvar)
+	if sshSock == "" {
+		sshSock = os.Getenv("SSH_AUTH_SOCK")
+	}
+
+	if sshSock != "" {
+		sshSession := types.BuildSshSession{Id: "remote", Target: sshSock}
+		buildOptions.SshSessions = append(buildOptions.SshSessions, sshSession)
+		buildOptions.Secrets = append(buildOptions.Secrets, fmt.Sprintf("id=known_hosts,src=%s/.ssh/known_hosts", home))
```
pchico83, line 192: does this work in windows? I think we we have `--remote` e2e tests, it would be great to have an e2e test covering this scenario in linux/window
I haven't tried, but I think the path is slightly different:
https://superuser.com/questions/311886/where-is-the-known-hosts-file-for-openssh-for-windows

cmd/deploy/remote.go
```diff
@@ -219,6 +247,7 @@ func (rd *remoteDeployCommand) createDockerfile(tmpDir string, opts *Options) (s
 		GitCommitArgName:       constants.OktetoGitCommitEnvVar,
 		InvalidateCacheArgName: constants.OktetoInvalidateCacheEnvVar,
 		DeployFlags:            strings.Join(getDeployFlags(opts), " "),
+		KnownSshHosts:          []string{"github.com", "gitlab.com"},
```
pchico83, line 250: do we still need this?

cmd/deploy/remote.go
pchico83: another option, if this approach gets too complicated, is to use a static file. and the user can always update their own known host file to the deploy `image` if needed
or infer the file in a similar way as we do in the installer:
https://github.com/okteto/pipeline-installer-image/blob/b0d72026c80eb8947adfc9c5b4f4f47a2f7bca8a/cmd/git/ssh.go#L53

maroshii: Yeah, this was incorrect and started failing in windows once I updated the test to `filepath.Join(home, ".ssh", "known_hosts")`. It should be fixed now.

maroshii: I was trying to create an integration test but it's tricky since it requires credentials or a private repo to test an ssh connection from inside the deploy. Adding these credentials as envvars would make `go test github.com/okteto/okteto/integration/... -tags="common integration actions"`  stop working out of the box. Is there a recommended way or do we have something in place for doing this? cc @okteto/lake.

cmd/deploy/remote.go
maroshii: removed

maroshii:
@okteto/lake @okteto/rainbow this PR is now fully complete and tested. Can I get a review 🙏 ?

jLopezbarb (approved):

maroshii:
> LGTM
> 
> some questions, what happens if:
> 
> * the ssh agent is not listening?
> * the ssh known hosts file doesn't exists?
> 
> does buildkit fail? as there are no overrides, these are important as they can block the usage of the cli.

if `SSH_SOCK_AUTH` doesn't exist we don't do aything.

Otherwise,
if it points to a file that doesn't exist we don't mount it,
if it points to a file that is not a ssh socket it mounts correctly but ssh will fail internally in the RUN step,
if `~/.ssh/know_hosts` doesn't exist we don't mount it
If `~/.ssh/know_hosts` points to an invalid file (but exist) we mount it and the RUN step fails.

jLopezbarb:
Just to mention that with this PR it would be very easy to implement https://github.com/okteto/okteto/issues/3589

PR #3756, opened 2023-06-28
place exec command in foreground when hybrid mode

AdrianPedriza:
## Context
when a command that requires interaction is executed on a dev environment in hybrid mode, the user does not get interaction with the command.

Captura de Pantalla 2023-06-28 a las 13 01 31

# Proposed changes
- place exec command in foreground
Captura de Pantalla 2023-06-28 a las 13 05 04

### DIFF
--- a/cmd/exec.go
+++ b/cmd/exec.go
@@ -227,6 +227,8 @@ func executeExec(ctx context.Context, dev *model.Dev, args []string) error {
 				return err
 			}
 
+			cmd.SysProcAttr.Foreground = true
+
 			return executor.RunCommand(cmd)
 		}
 


andreafalzetti:
I suggest to chat with @teresaromero because it was a recent change which apparently was causing some issues.. check out this chat: https://github.com/okteto/okteto/pull/3749#discussion_r1241809693

AdrianPedriza:
> I suggest to chat with @teresaromero because it was a recent change which apparently was causing some issues.. check out this chat: [#3749 (comment)](https://github.com/okteto/okteto/pull/3749#discussion_r1241809693)

thanks Andrea! I will take a look if I face same problem discussed on the thread

jLopezbarb:
> > I suggest to chat with @teresaromero because it was a recent change which apparently was causing some issues.. check out this chat: [#3749 (comment)](https://github.com/okteto/okteto/pull/3749#discussion_r1241809693)
> 
> thanks Andrea! I will take a look if I face same problem discussed on the thread

The problem is the implementation of sysproc is different from [windows](https://cs.opensource.google/go/go/+/master:src/syscall/exec_windows.go;l=243)  and [darwin](https://cs.opensource.google/go/go/+/master:src/syscall/exec_libc2.go;l=15), so we must differentiate between both adding a [build tag](https://pkg.go.dev/go/build#hdr-Build_Constraints) to have both versions running.

Since this will be shared between all commands on exec and up. How do you see creating a package where its specified how to run the commands in each OS?

AdrianPedriza:
> > > I suggest to chat with @teresaromero because it was a recent change which apparently was causing some issues.. check out this chat: [#3749 (comment)](https://github.com/okteto/okteto/pull/3749#discussion_r1241809693)
> > 
> > 
> > thanks Andrea! I will take a look if I face same problem discussed on the thread
> 
> The problem is the implementation of sysproc is different from [windows](https://cs.opensource.google/go/go/+/master:src/syscall/exec_windows.go;l=243) and [darwin](https://cs.opensource.google/go/go/+/master:src/syscall/exec_libc2.go;l=15), so we must differentiate between both adding a [build tag](https://pkg.go.dev/go/build#hdr-Build_Constraints) to have both versions running.
> 
> Since this will be shared between all commands on exec and up. How do you see creating a package where its specified how to run the commands in each OS?

yes, seems implementation need to change depending on the OS. Since I originally thought it would be a quick fix I will close the PR and create an issue to address it when needed

PR #3757, opened 2023-06-28
Improve "cannot create pid" error message

andreafalzetti:
# Proposed changes

I've hit this error today and it took me some time to figure out why it was failing.

It was caused accidentally because I've used `sudo` to run some okteto commands which I then interrupted leaving an orphan `okteto.pid` owned by `root`, so that any following command was failing.

Besides spitting out an error here, we culd offer the user an actionable suggestion to heal and solve the situation.

It might be sligly different for Windows but I think as a generic advice, the one I've used could be sufficient for solving this error in most scenarios.. Let me know your thought on:

1. The approach used / style of comms
2. The actual suggested fix

| Before | After |
|--------|-------|
|    Screenshot 2023-06-15 at 17 55 35    |    Screenshot 2023-06-15 at 17 55 28   |

### DIFF
--- a/cmd/up/up.go
+++ b/cmd/up/up.go
@@ -606,7 +606,14 @@ func (up *upContext) start() error {
 
 	if err := up.pidController.create(); err != nil {
 		oktetoLog.Infof("failed to create pid file for %s - %s: %s", up.Dev.Namespace, up.Dev.Name, err)
-		return fmt.Errorf("couldn't create pid file for %s - %s", up.Dev.Namespace, up.Dev.Name)
+
+		return oktetoErrors.UserError{
+			E: fmt.Errorf("couldn't create a pid file for %s - %s", up.Dev.Namespace, up.Dev.Name),
+			Hint: `This error can occur if the ".okteto" folder in your home has misconfigured permissions.
+    To resolve, try 'sudo chown -R <your-user>: ~/.okteto'
+
+    Alternatively, check the permissions of that directory and its content to ensure your user has write permissions.`,
+		}
 	}
 
 	defer up.pidController.delete()


cmd/up/up.go
```diff
@@ -606,7 +606,12 @@ func (up *upContext) start() error {
 
 	if err := up.pidController.create(); err != nil {
 		oktetoLog.Infof("failed to create pid file for %s - %s: %s", up.Dev.Namespace, up.Dev.Name, err)
-		return fmt.Errorf("couldn't create pid file for %s - %s", up.Dev.Namespace, up.Dev.Name)
+		return fmt.Errorf(`The command failed because the okteto CLI couldn't create a pid file for %s - %s.
+
+    This error can occur if the ".okteto" folder in your home has misconfigured permissions.
+    To resolve, try "sudo chown -R <your-user>: ~/.okteto"
+
+    Alternatively, check the permissions of that directory and its content to ensure your user has write permissions."`, up.Dev.Namespace, up.Dev.Name)
```
AdrianPedriza, line 614:
```suggestion
    Alternatively, check the permissions of that directory and its content to ensure your user has write permissions.`, up.Dev.Namespace, up.Dev.Name)
```

AdrianPedriza: Maybe we could use [UserError](https://github.com/okteto/okteto/blob/master/pkg/errors/errors.go#L23C12-L23C12) in this case as we expose the error with a suggestion

jLopezbarb: +1 on using `UserError` to give the user a hint on how to fix it.

andreafalzetti: thanks, PR updated

AdrianPedriza (approved):

Issue #3685, opened 2023-06-01
Ignore/warn about unsupported dev fields in the manifest instead of failing 


rlamana:
To ease switching to a hybrid mode instead of sync, we should take into account unsupported manifest fields and warn the users instead of failing the execution. 

Current output on `okteto up`
<img width="594" alt="Screenshot 2023-06-01 at 10 34 10" src="https://github.com/okteto/okteto/assets/237819/956c1b31-a461-4f13-a053-832c9145e4c5">

It should instead warn about it, something like this:
<img width="699" alt="Screenshot 2023-06-01 at 10 34 58" src="https://github.com/okteto/okteto/assets/237819/07decdc4-188b-434f-9ee2-d05b8bde30f1">

This is the table of the fields that should be supported in hybrid:

Sync | Hybrid | Currently Addressed
-- | -- | --
name | SHOULD WORK |  
selector | selector (ALREADY SUPPORTED) | YES
annotations | SHOULD WORK |  
tolerations | Not Supported |  
context | Not Supported |  
namespace | Not Supported |  
container | SHOULD WORK |  
image | Not Supported |  
push | Not Supported |  
imagePullPolicy | Not Supported |  
secrets | Not Supported |  
command | command (ALREADY SUPPORTED) | YES
args | SHOULD WORK |  
probes | SHOULD WORK |  
lifecycle | Not Supported |  
workdir | workdir (ALREADY SUPPORTED) |  
securityContext | Not Supported |  
serviceAccount | Not Supported |  
remote | SHOULD WORK -> local port where everything is exposed |  
sshServerPort | Not Supported |  
externalVolumes | Not Supported |  
sync | Not Supported |  
forward | forward (ALREADY SUPPORTED) |  
reverse | reverse (ALREADY SUPPORTED) |  
interface | SHOULD WORK |  
resources | Not Supported |  
services | SHOULD FAIL ? |  
persistentVolume | Not Supported |  
initContainer | Not Supported → Make sure we DON’T CREATE init container |  
initFromImage | Not Supported |  
timeout | SHOULD WORK |  
nodeSelector | Not Supported |  
affinity | Not Supported |  
metadata | SHOULD WORK |  
autocreate | SHOULD WORK |  
envFiles | SHOULD WORK |  
environment | environment (ALREADY SUPPORTED) |  
volumes | Not Supported |  
replicas | Not Supported |  
healthchecks | SHOULD WORK |  
labels | SHOULD WORK |

PR #3764, opened 2023-06-29
[Hybrid] Warn on unused fields but don't fail

andreafalzetti:
# Proposed changes

Fixes https://github.com/okteto/okteto/issues/3685

## Description

This change adds warnings that will be displayed to the users when they enable hybrid mode in a manifest where the `dev` section has fields unused by hybrid mode, but that could belong to the default mode. This can occur when trying hybrid mode on an existing configured manifest.

## How to test the old behavioir

1. Clone a sample repo: `git clone git@github.com:andreafalzetti/hybrid-test.git`
2. Go inside and using the latest okteto CLI version, run: `cd hybrid-test && okteto up`
3. Notice a series of errors and the process interrupted

## How to test the new behaviour

1. Checkout this branch and build the CLI
2. Go to the previous sample repo and run `okteto up` again.
3. Notice that the command is not failing anymore and you should see a series of warn messages saying: "In hybrid mode, the field "<field>" specified in your manifest is ignored".


pkg/model/serializer.go
```diff
@@ -742,13 +765,21 @@ func (d *Dev) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
+				UnsupportedFields map[string]interface{} `yaml:",inline" json:"-"`
 			}
 
 			hybridModeDev := &hybridModeInfo{}
 			err := unmarshal(hybridModeDev)
 			if err != nil {
 				return err
 			}
+
+			for _, field := range hybridUnsupportedFields {
```
jLopezbarb, line 778: Could we iterate over `hybridModeDev.UnsupportedFields[field]`. That way we don't need the list of unsupported fields which does not scale and if we add a new one we may forget to add it to the list.

```suggestion
for field, _ := range hybridModeDev.UnsupportedFields {
    oktetoLog.Warning("In hybrid mode, the field '%s' specified in your manifest is ignored", field)
}
```

jLopezbarb (changes requested):

pkg/model/serializer.go
andreafalzetti: addressed in https://github.com/okteto/okteto/pull/3764/commits/9b62488e14fd06420b5ef33a49f1b6cff9ea9194#diff-4f6045f32d12e1bf0cf56de0a6ec8e8937dca49ea38367f0f9dbf11e359ff4c9R778

pkg/model/serializer.go
```diff
@@ -757,27 +775,17 @@ func (d *Dev) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
+					if key == unsupportedFieldName {
+						isSupported = false
+						break
+					}
+				}
 
-			for _, field := range hybridUnsupportedFields {
-				if hybridModeDev.UnsupportedFields[field] != nil {
-					oktetoLog.Warning("In hybrid mode, the field '%s' specified in your manifest is ignored", field)
+				if !isSupported {
```
AdrianPedriza, line 787: @andreafalzetti @jLopezbarb WDYT about show just one warning that contains all the unsupported fields used instead of repeat same message for each field?

jLopezbarb: Yes, we can do something like the message that we have for compose.

andreafalzetti: Now that you guys mentioned it, and I also thought it was a good idea, I've done the change ✅

jLopezbarb (approved):

integration/commands/dev.go
```diff
@@ -87,25 +87,24 @@ func RunOktetoUp(oktetoPath string, upOptions *UpOptions) (*UpCommandProcessResu
 	}, nil
 }
 
-func RunOktetoUpAndWait(oktetoPath string, upOptions *UpOptions) error {
-
+func RunOktetoUpAndWait(oktetoPath string, upOptions *UpOptions) (bytes.Buffer, error) {
```
teresaromero, line 90:
```suggestion
func RunOktetoUpAndWaitWithOutput(oktetoPath string, upOptions *UpOptions) (bytes.Buffer, error) {
```

could we rename the function if now is returning the output? or perhaps create a similar one with output and keep the original if is used on other places?

andreafalzetti: @teresaromero this function is only used for this test, would you still go for the `WithOutput` suffix in that case?

pkg/model/serializer.go
```diff
@@ -722,6 +722,63 @@ func checkFileAndNotDirectory(path string) error {
 	return fmt.Errorf("Secret '%s' is not a regular file", path)
 }
 
+type hybridModeInfo struct {
```
teresaromero, line 725: would it make sense to move the hybrid "models" and variables to a pkg `hybrid` in order to avoid making model bigger? 

cc @jLopezbarb

pkg/model/serializer.go
```diff
@@ -734,22 +791,16 @@ func (d *Dev) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
-
 			hybridModeDev := &hybridModeInfo{}
 			err := unmarshal(hybridModeDev)
 			if err != nil {
 				return err
 			}
+
+			warningMsg := hybridModeDev.warnHybridUnsupportedFields()
+			if warningMsg != "" {
+				oktetoLog.Warning(warningMsg)
```
teresaromero, line 802: i am thinking maybe we could just call the function that shows or not the warning instead of checking the output.

`warnHybridUnsupportedFields` i feel the function is going to warn, if we are returning output and then warning i would change naming maybe `getHybridUnsupportedFields` and with those then warn? instead of returning the message, return the fields and then create the message out. WDYT?

teresaromero (approved):
minor comments but not blockers, just discussion :) LGTM

pkg/model/serializer.go
andreafalzetti: @teresaromero I was thinking about the same, like we did for `pkg/cache`, howerer the Hybrid type depends on types from `model` that are shared with `Dev` so it turns into a cycle dependency so I gave up moving only Hybrid, I think once we move the whole "Dev" type outside of the model it will be easier

pkg/model/serializer.go
andreafalzetti: @teresaromero we need to keep `oktetoLog.Warning` calls outside of the function so that it's easily testable.

I see your point about the naming, I am open for suggestions! I think `getHybridUnsupportedFields` implies it returns the actual fields? The word "warn" in the current one makes me think about logging/warnings and having the return value called `warningMsg` feels quite clear.

I can't think of alternatives now but if you have one, happy to follow-up on this 🙏

integration/commands/dev.go
teresaromero: "only for test" but i think we should be consistent. We dont have a guidelines for naming yet but i think is better if we change it like that. No hard feelings

andreafalzetti: @teresaromero "only for _this_ test" was a reply to your feedback: "or perhaps create a similar one with output and keep the original if is used on other places" - my point is, no need to create a new one because it's only used once

andreafalzetti: I agree on the name, it's more explicit 👍 https://github.com/okteto/okteto/pull/3764/commits/512e05d846698856ab86dd1978732d9216e61c6e

PR #3766, opened 2023-06-30
add implementation to retrieve cmap variables when running in remote

### LINKED ISSUE #3735: `okteto deploy --remote` doesn't mask variables specified with `--var`
**Describe the bug**
Run `okteto deploy --remote --var TEST=111` with a deploy command `echo $TEST`.
The variable value isn't masked

**Expected behavior**
The variable is masked as any other variable/secret

AdrianPedriza:
# Context

when a variable is printed in deploy section using remote deploy feature it is not masked as expected. This is because implementation related to retrieve variables from config map in remote deploy mode is empty

Fixes #3735

# Proposed changes
- add implementation about retrieve variables from configmap when deploy is running remotely

# How to check the fix

1. Using latest CLI release
2. clone https://github.com/okteto/movies
3. set deploy section as:
````yaml
deploy:
  - name: test
    command: echo $TEST
````
4. run `okteto deploy --remote --var TEST=111`
5. Value should be masked

### DIFF
--- a/cmd/deploy/cfghandler.go
+++ b/cmd/deploy/cfghandler.go
@@ -35,10 +35,14 @@ type configMapHandler interface {
 
 // deployInsideDeployConfigMapHandler is the runner used when the okteto is executed
 // inside an okteto deploy command
-type deployInsideDeployConfigMapHandler struct{}
+type deployInsideDeployConfigMapHandler struct {
+	k8sClientProvider okteto.K8sClientProvider
+}
 
-func newDeployInsideDeployConfigMapHandler() *deployInsideDeployConfigMapHandler {
-	return &deployInsideDeployConfigMapHandler{}
+func newDeployInsideDeployConfigMapHandler(provider okteto.K8sClientProvider) *deployInsideDeployConfigMapHandler {
+	return &deployInsideDeployConfigMapHandler{
+		k8sClientProvider: provider,
+	}
 }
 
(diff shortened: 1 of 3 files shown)


pchico83:
@AdrianPedriza is it possible to add a test that covers this scenario? does the fix work for `okteto destroy --remote` too?

AdrianPedriza:
> @AdrianPedriza is it possible to add a test that covers this scenario? does the fix work for `okteto destroy --remote` too?

I can add e2e test for it. On the other hand, `okteto destroy` doesn't accept variables so I'm afraid we can't cover that case for the destroy command

pchico83:
> @AdrianPedriza is it possible to add a test that covers this scenario? does the fix work for `okteto destroy --remote` too?

@AdrianPedriza `okteto destroy --remote` reuses the variables used at deployment time. I want to test if those envvars are masked at destroy time, even if they were added with the `--var` option.

AdrianPedriza:
> > @AdrianPedriza is it possible to add a test that covers this scenario? does the fix work for `okteto destroy --remote` too?
> 
> @AdrianPedriza `okteto destroy --remote` reuses the variables used at deployment time. I want to test if those envvars are masked at destroy time, even if they were added with the `--var` option.

👍 I will check that case for `destroy` command then. I will add an e2e test covering both cases

jLopezbarb (approved):

teresaromero (approved):
👍🏻

PR #3771, opened 2023-07-03
add dev mode annotation (sync or hybrid)

### LINKED ISSUE #3653: add dev mode (hybrid or sync) into deployment/sfs
**Describe the bug**
if an `okteto up` sessions runs using hybrid mode, backend doesn't knows dev mode used. In order to let know the backend which mode is used, we should add a new annotation into the deployment using values `sync` or `hybrid`.

AdrianPedriza:
# Context
when a dev environment starts the backend doesn't have the possibility to know which mode is in use. Available dev modes are `sync` or `hybrid`.

# Proposed changes
- add `dev.okteto.com/mode` to app and dev app to make available dev mode in use for the backend
- delete annotation when dev mode off.

Fixes #3653 

# How to test changes
1. develop in https://github.com/okteto/movies by running `okteto up api` 
2. check that development `api` and `api-okteto` have annotations `dev.okteto.com/mode = sync`
3. run `okteto down api` and check that annotation is removed

### DIFF
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -23,7 +23,7 @@ const (
 	OktetoURLAnnotation = "dev.okteto.com/url"
 
 	// OktetoDevModeAnnotation indicates the development mode in use (sync or hybrid)
-	OktetoDevModeAnnotation = "dev.okteto.com/mode"
+	OktetoDevModeAnnotation = "dev.okteto.com/dev-mode"
 
 	// OktetoExtension identifies the okteto extension in kubeconfig files
 	OktetoExtension = "okteto"
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -22,6 +22,9 @@ const (
 	// OktetoURLAnnotation indicates the okteto cluster public url
 	OktetoURLAnnotation = "dev.okteto.com/url"
 
+	// OktetoDevModeAnnotation indicates the development mode in use (sync or hybrid)
+	OktetoDevModeAnnotation = "dev.okteto.com/mode"
+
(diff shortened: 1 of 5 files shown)


andreafalzetti (approved):
code change LGTM

tested and works as expected in both cases (sync/hybrid) 👍

teresaromero (approved):

pkg/constants/constants.go
```diff
@@ -22,6 +22,9 @@ const (
 	// OktetoURLAnnotation indicates the okteto cluster public url
 	OktetoURLAnnotation = "dev.okteto.com/url"
 
+	// OktetoDevModeAnnotation indicates the development mode in use (sync or hybrid)
+	OktetoDevModeAnnotation = "dev.okteto.com/mode"
```
jLopezbarb, line 26: @ifbyol Do we agree that this is the correct name for the annotation? 

The possible values are `sync` and `hybrid`

jLopezbarb (approved):
Wait until @ifbyol confirms the name of the annotation but code looks good to me!!

pkg/constants/constants.go
ifbyol: Given that `dev.okteto.com` is the common prefix for our labels and annotations, `mode` doesn't specify too much about what it means or what it refers to. What about `dev.okteto.com/dev-mode`? 

Anyways, no strong opinion on the name of the annotation, so you can leave it as it is. The key part is being able to identify when an okteto up was sync or hybrid mode.

AdrianPedriza: 👍  b0832a86ca453c1fced70b11f89487ce79ff7b18

PR #3776, opened 2023-07-03
feat: add `--wake` flag for `okteto preview destroy`

### LINKED ISSUE #2057: Support for `okteto preview destroy --wait`
I am scripting some tests and I need to know when a preview environment is destroyed. Right now, the preview namespace can stay in `Terminating` state for a while.

Implement a `--wait` flag that makes sure the namespace is gone.

nitishfy:
# Proposed changes

Fixes #2057

- Added the `--wake` flag for `okteto preview destroy`
Screenshot from 2023-07-03 20-55-16


cmd/preview/destroy.go
```diff
@@ -83,29 +87,37 @@ func Destroy(ctx context.Context) *cobra.Command {
...
 
-			err = c.executeDestroyPreview(ctx, name)
+			err = c.executeDestroyPreview(ctx, opts)
 			analytics.TrackPreviewDestroy(err == nil)
 			return err
 		},
 	}
-
+	cmd.Flags().BoolVarP(&opts.wait, "wait", "w", false, "wait until the preview environment gets destroyed (defaults to false)")
+	if err := cmd.Flags().MarkHidden("filename"); err != nil {
```
andreafalzetti, line 96: `MarkHidden` for `filename` is not needed here

AdrianPedriza:
if the default behavior now is to schedule a destroy and not wait we will change the previous default behavior when we wait for the `preview destroy` to finish. Should we change the default behavior in order to wait to finish setting `wait` flag to `true` ? cc @pchico83

pchico83:
> if the default behavior now is to schedule a destroy and not wait we will change the previous default behavior when we wait for the `preview destroy` to finish. Should we change the default behavior in order to wait to finish setting `wait` flag to `true` ? some users may be affected as has happened in the past by changing the default behavior cc @pchico83

@AdrianPedriza what's the current default behavior, is it to always wait?

AdrianPedriza:
> > if the default behavior now is to schedule a destroy and not wait we will change the previous default behavior when we wait for the `preview destroy` to finish. Should we change the default behavior in order to wait to finish setting `wait` flag to `true` ? some users may be affected as has happened in the past by changing the default behavior cc @pchico83
> 
> @AdrianPedriza what's the current default behavior, is it to always wait?

yes it is

pchico83:
> > > if the default behavior now is to schedule a destroy and not wait we will change the previous default behavior when we wait for the `preview destroy` to finish. Should we change the default behavior in order to wait to finish setting `wait` flag to `true` ? some users may be affected as has happened in the past by changing the default behavior cc @pchico83
> > 
> > 
> > @AdrianPedriza what's the current default behavior, is it to always wait?
> 
> yes it is. Well there is no flag ATM for this command so it is the only behavior, to wait to finish

@AdrianPedriza one option is to keep the current behavior, without a flag, but wait until the namespace of the preview is gone. If we want to change the default behavior, it would be a breaking change, and we would need to wait for CLI 3.0

nitishfy:
> @AdrianPedriza one option is to keep the current behavior, without a flag, but wait until the namespace of the preview is gone. If we want to change the default behavior, it would be a breaking change, and we would need to wait for CLI 3.0

When we don't use the `--wait` flag, the destroy is scheduled with prompt appearing on the CLI. However, when the `--wait` flag is used (value is true in this case), then we can see all the steps running on the CLI. This is similar to how `okteto preview deploy` works with and without `--wait` flag.

AdrianPedriza:
> > @AdrianPedriza one option is to keep the current behavior, without a flag, but wait until the namespace of the preview is gone. If we want to change the default behavior, it would be a breaking change, and we would need to wait for CLI 3.0
> 
> When we don't use the `--wait` flag, the destroy is scheduled with prompt appearing on the CLI. However, when the `--wait` flag is used (value is true in this case), then we can see all the steps running on the CLI. This is similar to how `okteto preview deploy` works with and without `--wait` flag.

@pchico83 I agree, that way we make sure that the namespace no longer exists and is not a breaking change

@nitishfy I think your implementation is the right one, similar to the preview deploy command, but it would be a breaking change since for now, without any flag, we wait for the preview to be destroyed, and with the proposed implementation this behavior is under the `--wait=true` flag. Some users may be affected. The new proposal is to keep the same behavior we have now when the `wait` flag is not present but add logic that waits until the `Terminating` state disappears for the namespace to destroy, which allows to recreate a preview under a namespace with the same name.

nitishfy:
> > > @AdrianPedriza one option is to keep the current behavior, without a flag, but wait until the namespace of the preview is gone. If we want to change the default behavior, it would be a breaking change, and we would need to wait for CLI 3.0
> > 
> > 
> > When we don't use the `--wait` flag, the destroy is scheduled with prompt appearing on the CLI. However, when the `--wait` flag is used (value is true in this case), then we can see all the steps running on the CLI. This is similar to how `okteto preview deploy` works with and without `--wait` flag.
> 
> @pchico83 I agree, that way we make sure that the namespace no longer exists and is not a breaking change
> 
> @nitishfy I think your implementation is the right one, similar to the preview deploy command, but it would be a breaking change since for now, without any flag, we wait for the preview to be destroyed, and with the proposed implementation this behavior is under the `--wait=true` flag. Some users may be affected. The new proposal is to keep the same behavior we have now when the `wait` flag is not present but add logic that waits until the `Terminating` state disappears for the namespace to destroy, which allows to recreate a preview under a namespace with the same name.

Could you please explain this once again: _add logic that waits until the Terminating state disappears for the namespace to destroy, which allows to recreate a preview under a namespace with the same name._

When we don't use the `--wait` flag, we want the execution to happen smoothly as it used to happen before the `--wait` flag was introduced. When `--wait` flag is provided, we would like to schedule it for destroy. Is that correct?

AdrianPedriza:
> Could you please explain this once again: _add logic that waits until the Terminating state disappears for the namespace to destroy, which allows to recreate a preview under a namespace with the same name._

Currently, if a user executes `okteto preview destroy <preview-name>` we are sending the command to destroy the namespace and after that we terminate the command. This waiting logic has to be modified because sometimes the namespace deletion remains in `Terminating` state for a while preventing to create previews under the same name. We should add logic that waits until the namespace no longer exists instead, that is to say, that it is no longer in `Terminating` state. 

> When we don't use the `--wait` flag, we want the execution to happen smoothly as it used to happen before the `--wait` flag was introduced. When `--wait` flag is provided, we would like to schedule it for destroy. Is that correct?

For now we must maintain the same behavior that we have today which is if a user executes `okteto preview destroy` we are waiting for it to finish (this is the logic that must be modified taking into account that the namespace is not in `Terminating` state), we are not scheduling its destruction. With your new implementation we will be able to schedule its destruction and not wait for it if the user indicates it by executing `okteto preview destroy --wait=false`

nitishfy:
> > Could you please explain this once again: _add logic that waits until the Terminating state disappears for the namespace to destroy, which allows to recreate a preview under a namespace with the same name._
> 
> @nitishfy Currently, if a user executes `okteto preview destroy <preview-name>` we are sending the command to destroy the namespace and after that we terminate the command. This waiting logic has to be modified because sometimes the namespace deletion remains in `Terminating` state for a while preventing to create previews under the same name. We should add logic that waits until the namespace no longer exists instead, that is to say, that it is no longer in `Terminating` state.
> 
> > When we don't use the `--wait` flag, we want the execution to happen smoothly as it used to happen before the `--wait` flag was introduced. When `--wait` flag is provided, we would like to schedule it for destroy. Is that correct?
> 
> For now we must maintain the same behavior that we have today which is if a user executes `okteto preview destroy` we are waiting for it to finish (this is the logic that must be modified taking into account that the namespace is not in `Terminating` state), we are not scheduling its destruction. With your new implementation we will be able to schedule its destruction and not wait for it if the user indicates it by executing `okteto preview destroy --wait=false`

Understood well. Then what's the purpose of using `--wait` flag? I think we can just modify our current logic otherwise implement this logic to be used with the `--wait` flag. WDYT?

AdrianPedriza:
> > > Could you please explain this once again: _add logic that waits until the Terminating state disappears for the namespace to destroy, which allows to recreate a preview under a namespace with the same name._
> > 
> > 
> > @nitishfy Currently, if a user executes `okteto preview destroy <preview-name>` we are sending the command to destroy the namespace and after that we terminate the command. This waiting logic has to be modified because sometimes the namespace deletion remains in `Terminating` state for a while preventing to create previews under the same name. We should add logic that waits until the namespace no longer exists instead, that is to say, that it is no longer in `Terminating` state.
> > > When we don't use the `--wait` flag, we want the execution to happen smoothly as it used to happen before the `--wait` flag was introduced. When `--wait` flag is provided, we would like to schedule it for destroy. Is that correct?
> > 
> > 
> > For now we must maintain the same behavior that we have today which is if a user executes `okteto preview destroy` we are waiting for it to finish (this is the logic that must be modified taking into account that the namespace is not in `Terminating` state), we are not scheduling its destruction. With your new implementation we will be able to schedule its destruction and not wait for it if the user indicates it by executing `okteto preview destroy --wait=false`
> 
> Understood well. Then what's the purpose of using `--wait` flag? I think we can just modify our current logic otherwise implement this logic to be used with the `--wait` flag. WDYT?

For now the wait logic should be kept without using the flag in order not to introduce braking changes but allowing not to wait for it to be removed can be useful using the `--wait=false` flag. With the introduction of CLI `3.0` the wait logic will be under the `--wait` flag so I think introducing this flag will create the basis for 3.0.

andreafalzetti:
Moving the PR to draft until ready for review again

nitishfy:
**Update:** This PR is ready for the next set of reviews. 
- Added the test case
- All the unit tests are passing 
- Default value of `--wait` flag is set to true this time. This will make sure default behavior is followed.

AdrianPedriza (approved):
LGTM

teresaromero (approved):

teresaromero:
Added docs PR to add this new flag https://github.com/okteto/docs/pull/443

nitishfy:
Awesome! We're ready to merge this PR🎉

jLopezbarb (approved):

andreafalzetti (approved):
Tested and works as expected!

Thanks @nitishfy 🙇

Issue #3780, opened 2023-07-04
Expand variable in the build.image


jLopezbarb:
**Describe the bug**
We are not expanding the environment variables on the `build.image` field. If I run `okteto build` with an image that has the environment variables I get the following error:
` x  Error building service 'hello-world': error building image 'registry.xxxxxxxx.okteto.net/jlopezbarb/python-hello-world:${OKTETO_NAMESPACE}': build failed: failed to solve: failed to push registry.xxxxxxx.okteto.net/jlopezbarb/python-hello-world:${OKTETO_NAMESPACE}: invalid reference format
`

**To Reproduce**

1. Create the following `Dockerfile`:
```Dockerfile
FROM alpine
```
2. Create the following `okteto.yml`:
```yaml
build:
  hello-world:
    image: okteto.dev/python-hello-world:${OKTETO_NAMESPACE}
```
3. Run `okteto build`

**Expected behavior**
I want that the environment variable get expanded and the build run successfully

**Desktop (please complete the following information):**

- Version 2.16.6

PR #3784, opened 2023-07-04
feat(build): expand image before build

andreafalzetti:
# Proposed changes

Expand build image just before doing the actual build to make sure if we use `depends_on` and other vars coming from other services, it won't break that behavioir. 

Fixes https://github.com/okteto/okteto/issues/3780

## How to reproduce the bug

- Follow steps in the [issue](https://github.com/okteto/okteto/issues/3780)

## How to test the fix

- Build this branch, run against the example in the issue
- Notice how the CLI can successfully build the image now
- Also, notice that the expansion is done just before the build maintaining the original manifest so that if the user uses variables coming from other builds, it won't break. To verify this, you can use the following manifest:

```yaml
build:
  api2:
    context: .
  k8s-status-api:
    image: okteto.dev/k8s-status-api:${OKTETO_BUILD_API2_TAG}
    context: .
    depends_on:
      - api2
```
- In this case, the CLI will build `api2` first, and when building `k8s-status-api`, it will replace `${OKTETO_BUILD_API2_TAG}` with the actual tag built from `api2`.

### DIFF
--- a/pkg/model/dev_test.go
+++ b/pkg/model/dev_test.go
@@ -1063,43 +1063,56 @@ func Test_ExpandEnv(t *testing.T) {
 		value         string
 		expandIfEmpty bool
 		result        string
+		expectedErr   error
 	}{
+		{
+			name:          "broken var - missing closing curly bracket",
+			value:         "value-${BAR",
+			expandIfEmpty: true,
+			result:        "",
+			expectedErr:   fmt.Errorf("error expanding environment on 'value-${BAR': closing brace expected"),
+		},
 		{
 			name:          "no-var",
 			value:         "value",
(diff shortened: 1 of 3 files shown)


pkg/model/dev_test.go
```diff
@@ -1063,43 +1063,56 @@ func Test_ExpandEnv(t *testing.T) {
 		value         string
 		expandIfEmpty bool
 		result        string
+		expectedErr   error
 	}{
+		{
```
andreafalzetti, line 1068: adding this test to `ExpandEnv` because the failure case was not covered yet by tests

jLopezbarb (approved):

PR #3798, opened 2023-07-06
fix: context switch regression

### LINKED ISSUE #3334: Can't switch to previous context with non existent / deleted namespace

jLopezbarb:
# Proposed changes

Fixes #3334 regression caused by #3343 introduced in [CLI release 2.16.0](https://github.com/okteto/okteto/releases/tag/2.16.0)

The bug was introduced as part of the multicluster fixes. We end up going with another approach but this part of the code was a nice to have (always send the namespace the user will be working on) regarding if the user was running in multicluster or single cluster.

I've tested it with different commands (context and namespace)

Also I've added an unit test for the code that part where we were not covering the errors that we could face and the scenario that we fixed. 

| Before | After |
|--------|-------|
|   Captura de Pantalla 2023-07-06 a las 18 23 53  |  Captura de Pantalla 2023-07-06 a las 18 24 25 |

### DIFF
--- a/cmd/context/create.go
+++ b/cmd/context/create.go
@@ -289,8 +289,12 @@ func getLoggedUserContext(ctx context.Context, c *ContextCommand, ctxOptions *Co
 	if ctxOptions.Namespace == "" {
 		ctxOptions.Namespace = user.Namespace
 	}
+
 	userContext, err := c.getUserContext(ctx, ctxOptions.Namespace)
 	if err != nil {
+		if oktetoErrors.IsNotFound(err) {
+
+		}
 		return nil, err
 	}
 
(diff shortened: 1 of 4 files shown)


cmd/context/create.go
```diff
@@ -289,8 +289,12 @@ func getLoggedUserContext(ctx context.Context, c *ContextCommand, ctxOptions *Co
 	if ctxOptions.Namespace == "" {
 		ctxOptions.Namespace = user.Namespace
 	}
+
 	userContext, err := c.getUserContext(ctx, ctxOptions.Namespace)
 	if err != nil {
+		if oktetoErrors.IsNotFound(err) {
+
```
andreafalzetti, line 296: missing a return?

AdrianPedriza:
why a fix of the use logic is not in the `use.go` file but in `create.go`? shouldn't we refactor to place the code where it makes more sense to locate it?

AdrianPedriza:
> why a fix of the use logic is not in the `use.go` file but in `create.go`? shouldn't we refactor to place the code where it makes more sense to locate it?

maybe we could move [this function](https://github.com/okteto/okteto/blob/jlopezbarb/fix-context-switch-regression/cmd/context/create.go#L106) into `use.go`. WDYT?

cmd/context/create.go
```diff
@@ -336,17 +340,28 @@ func (c ContextCommand) getUserContext(ctx context.Context, ns string) (*types.U
...
-		// If there is a TLS error, don't continue the loop and return the raw error
-		if err != nil && oktetoErrors.IsX509(err) {
-			return nil, err
+			// If there is a TLS error, don't continue the loop and return the raw error
+			if oktetoErrors.IsX509(err) {
+				return nil, err
+			}
+
+			if oktetoErrors.IsNotFound(err) {
+				userContext, err = client.User().GetContext(ctx, "")
```
teresaromero, line 360:
```suggestion
                                 // fallback to personal namesapce using empty string as param
				userContext, err = client.User().GetContext(ctx, "")
```

jLopezbarb:
> > why a fix of the use logic is not in the `use.go` file but in `create.go`? shouldn't we refactor to place the code where it makes more sense to locate it?
> 
> maybe we could move [this function](https://github.com/okteto/okteto/blob/jlopezbarb/fix-context-switch-regression/cmd/context/create.go#L106) into `use.go`. WDYT?

I do think that we should refactor it, but I don't think that should be covered in this PR to keep it simple

teresaromero (approved):

andreafalzetti (approved):
Tested and works as expected, thanks for fixing this 🎖️

PR #3799, opened 2023-07-06
Add cluster private credentials to build and deploy remote

maroshii:
# Context

If the current context points to an okteto cluster that has private credentials defined we now use those credentials to authenticate builds and remote deploys if those images are used in the builds.

The use cases defined below wouldn't work before this change unless you had another set of credentials locally in your docker config:

1) Deploy remote with private image:
```yaml
deploy:
  image: my-private-registry.amazonaws.com/base
```
```bash
# fails because buildkit fails to pull base image for deploy
okteto deploy --remote
```

2) Build with private image as base image:
```Dockerfile
FROM my-private-registry.amazonaws.com/base
```
```bash
# fails because buildkit fails to pull base image
okteto build
```

3) Build without private images but push to private registry:
```Dockerfile
FROM alpine
```

```bash
# build and push, even with public base
okteto build -t "my-private-registry.amazonaws.com/target"
```

After this PR all the above use cases should work.

# Implementation

After building the image with buildkit (or deploy remote) we go to the registry to check that the image is actually there. This made this feature much more complicated because adding a buildkit session is not enough since we also need to provide the credentials to the registry which is not directly tied to buildkit.
Specifically I'm referring to [this piece of code](https://github.com/okteto/okteto/blob/master/pkg/cmd/build/build.go#L160-L175). Not sure if we can remove that or not but if we can the scope of this will be reduce significantly.

For buildkit, we extended the Auth provider to try and read the credentials from the okteto api if such functionality is desired (for cluster with okteto enabled).

# Test Cases

I tested most of the cases I know of. Please let me know if there's anything scenario I'm missing.

#### Compatibility
- [x] Deploying and building remote with private registries on old backend with no `RegistryCredentials` endpoint fail but doesn't panic

#### Build in okteto cluster with private registry configured
  - [x] push to cluster defined private registry SUCCEEDS _`okteto build -t {private_img}`_
  - [x] build to okteto registry SUCCEEDS
  - [x] build with private `FROM` SUCCEEDS

#### Deploy remote in okteto cluster
  - [x] `deploy --remote` with private `image: {image}` in manifest SUCCEEDS
  - [x] `deploy --remote --build` with private `FROM` SUCCEEDS

#### Destroy remote in okteto cluster
  - [x] `deploy --remote` with private `image: {image}` in manifest SUCCEEDS

#### Build in okteto cluster without private registry configured
  - [x] push to cluster defined private registry FAILS _`okteto build -t {private_img}`_
  - [x] build to okteto registry SUCCEEDS
  - [x] build with private `FROM` FAILS

#### Build in vanilla cluster
  - [x] build to okteto registry SUCCEEDS
  - [x] build with private `FROM` toto okteto registry FAILS _(shouldn't leak credentials)_

#### Deploying via pipeline installer
 - [x] deploy works with private `image` defined
 - [x] destroy works with private `image` defined

### DIFF
--- a/cmd/registrytoken/registrytoken.go
+++ b/cmd/registrytoken/registrytoken.go
@@ -1,0 +1,68 @@
+// Copyright 2023 The Okteto Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package registrytoken
+
+import (
+	"context"
+	"os"
+
+	contextCMD "github.com/okteto/okteto/cmd/context"
+
+	"github.com/okteto/okteto/pkg/auth/dockercredentials"
+	"github.com/okteto/okteto/pkg/errors"
+	"github.com/okteto/okteto/pkg/okteto"
+	"github.com/spf13/cobra"
+
+	"github.com/docker/docker-credential-helpers/credentials"
+)
+
+type regCreds struct {
+	okteto.Config
+}
+
+func (r regCreds) GetRegistryCredentials(host string) (string, string, error) {
+	return r.GetExternalRegistryCredentials(host)
+}
+
+func RegistryToken() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "registrytoken",
+		Short: "docker credentials helper for private registries registered in okteto",
+		Long: `Acts as a docker credentials helper printing out private registry credentials previously configured in okteto:
+
+Usage: echo gcr.io | okteto registrytoken get
+
+More info about docker credentials helpers here: https://github.com/docker/docker-credential-helpers
+  `,
+		Hidden:    true,
+		ValidArgs: []string{"get"},
+		Args:      cobra.OnlyValidArgs,
+	}
+
+	cmd.RunE = func(_ *cobra.Command, args []string) error {
+		action := args[0]
+		ctx := context.Background()
+		if err := contextCMD.NewContextCommand().Run(ctx, &contextCMD.ContextOptions{}); err != nil {
+			return err
+		}
+		conf := okteto.Config{}
+		if !conf.IsOktetoCluster() {
+			return errors.ErrContextIsNotOktetoCluster
+		}
+		h := dockercredentials.NewOktetoClusterHelper(regCreds{conf})
+		return credentials.HandleCommand(h, action, os.Stdin, os.Stdout)
+	}
+
+	return cmd
+}
--- a/cmd/registrytoken/registrytoken.go
+++ b/cmd/registrytoken/registrytoken.go
@@ -43,11 +43,15 @@ func RegistryToken() *cobra.Command {
 
 Usage: echo gcr.io | okteto registrytoken get
 
+Valid arguments are: store, get, erase, list, version.
+
+At this time only "get" is supported
+
 More info about docker credentials helpers here: https://github.com/docker/docker-credential-helpers
   `,
 		Hidden:    true,
-		ValidArgs: []string{"get"},
-		Args:      cobra.OnlyValidArgs,
+		ValidArgs: []string{"store", "get", "erase", "list", "version"},
+		Args:      cobra.ExactValidArgs(1),
 	}
 
 	cmd.RunE = func(_ *cobra.Command, args []string) error {
--- a/internal/test/client/user.go
+++ b/internal/test/client/user.go
@@ -45,7 +45,7 @@ func (c *FakeUserClient) GetUserSecrets(_ context.Context) ([]types.Secret, erro
 	return c.userSecrets, nil
 }
 
-func (*FakeUserClient) GetClusterCertificate(_ context.Context, cluster, ns string) ([]byte, error) {
+func (*FakeUserClient) GetClusterCertificate(_ context.Context, _, _ string) ([]byte, error) {
 	return nil, nil
 }
 
--- a/internal/test/client/user.go
+++ b/internal/test/client/user.go
@@ -49,10 +49,10 @@ func (c *FakeUserClient) GetClusterCertificate(_ context.Context, cluster, ns st
 	return nil, nil
 }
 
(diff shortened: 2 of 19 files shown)


teresaromero:
Regarding testing scenarios, have we checked this CLI changes integrated with the backend via installer?

maroshii:
> Regarding testing scenarios, have we checked this CLI changes integrated with the backend via installer?

Good point. In the installer we inject the node docker config file as a volume so all the credentials will already be there which is analogous to having the creds in your local docker store. Nonetheless, it needs to be tested to make sure there are no regression. I'll add those test scenarios  👍

pchico83:
> > Regarding testing scenarios, have we checked this CLI changes integrated with the backend via installer?
> 
> Good point. In the installer we inject the node docker config file as a volume so all the credentials will already be there which is analogous to having the creds in your local docker store. Nonetheless, it needs to be tested to make sure there are no regression. I'll add those test scenarios 👍

@maroshii once we integrate this CLI in the installer, could we remove the logic to inject the docker credentials directly in the installer job?

ifbyol:
> > > Regarding testing scenarios, have we checked this CLI changes integrated with the backend via installer?
> > 
> > 
> > Good point. In the installer we inject the node docker config file as a volume so all the credentials will already be there which is analogous to having the creds in your local docker store. Nonetheless, it needs to be tested to make sure there are no regression. I'll add those test scenarios 👍
> 
> @maroshii once we integrate this CLI in the installer, could we remove the logic to inject the docker credentials directly in the installer job?

@pchico83 no, more work is still needed and it will be done in following epics. We need the CLI behaving as a docker-credential-helper and inject a docker config using such helper.

jLopezbarb:
Reading the context you mentioned `If the current context points to an okteto cluster that has private credentials defined `. I've a question related to PR scope since you are specifying that is an okteto cluster. Should this PR cover a context with the private credentials defined in the keychain instead of  username + password?

maroshii:
> Should this PR cover a context with the private credentials defined in the keychain instead of username + password?

These values are returned from the backend and we generate them dynamically so we know for sure they are username and password (for now). As long as the `registryCredentials` query returns user and pass we should be good.

pchico83 (approved):

maroshii:
NOTE: private docker hub registries are failing because the format of the default registry. It's a simple fix but it needs to get fixed before merging

jLopezbarb (approved):

internal/test/client/user.go
```diff
@@ -51,3 +52,7 @@ func (c *FakeUserClient) GetClusterCertificate(_ context.Context, cluster, ns st
 func (c *FakeUserClient) GetClusterMetadata(ctx context.Context, ns string) (types.ClusterMetadata, error) {
 	return types.ClusterMetadata{}, nil
 }
+
+func (c *FakeUserClient) GetRegistryCredentials(ctx context.Context, registry string) (dockertypes.AuthConfig, error) {
```
andreafalzetti, line 56:
```suggestion
func (c *FakeUserClient) GetRegistryCredentials(_ context.Context, _ string) (dockertypes.AuthConfig, error) {
```

shall we make it explicit that those are not used?

I see that the func above is inconsistent too with the functions above.

pkg/okteto/registry_credentials.go
```diff
@@ -0,0 +1,29 @@
...
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	oktetoLog "github.com/okteto/okteto/pkg/log"
+)
+
+func GetExternalRegistryCredentials(ctx context.Context, registryOrImage string) (string, string, error) {
```
andreafalzetti, line 11: Should we have a unit test for this func?

andreafalzetti (approved):
I haven't tested the funcionality itself, I've only reviewed the code. I left some non-blocking feedback.

pkg/okteto/registry_credentials.go
```diff
@@ -0,0 +1,29 @@
+package okteto
```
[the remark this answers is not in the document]
maroshii: I don't think we can afford to do that. This function has a strong dependency on the okteto package and the registry package doesn't depend on the okteto package. Moving it there would make it depend on it assuming it doesn't create a circular dependency.

pkg/auth/dockercredentials/helper.go
```diff
@@ -0,0 +1,56 @@
...
+// limitations under the License.
+package dockercredentials
+
+import (
+	"errors"
+
+	"github.com/docker/docker-credential-helpers/credentials"
+)
+
+var ErrNotImplemented = errors.New("not implemented")
```
AdrianPedriza, line 21: can we keep this error private?

pkg/okteto/config.go
```diff
@@ -27,3 +30,6 @@ func (Config) GetContextCertificate() (*x509.Certificate, error) { return GetCon
 func (Config) IsInsecureSkipTLSVerifyPolicy() bool               { return Context().IsInsecure }
 func (Config) GetServerNameOverride() string                     { return GetServerNameOverride() }
 func (Config) GetContextName() string                            { return Context().Name }
+func (Config) GetExternalRegistryCredentials(registryHost string) (string, string, error) {
+	return GetExternalRegistryCredentials(context.TODO(), registryHost)
```
AdrianPedriza, line 34:
```suggestion
	return GetExternalRegistryCredentials(context.TODO(), registryHost)
```
should we use `context.TODO()` ?

AdrianPedriza:
just minor comments from my side

cmd/registrytoken/registrytoken.go
AdrianPedriza, line 1: I think we should add `get` as a `registrytoken` command same as other (i.e https://github.com/okteto/okteto/blob/master/cmd/preview/preview.go#L46-L51) . If not, executing `okteto registrytoken` will panic instead give the proper suggestion:
Captura de Pantalla 2023-07-12 a las 10 33 52

maroshii: Nice catch! The cobra command configuration was incorrect. This fixes it:

```golang
ValidArgs: validArgs,
Args:      cobra.ExactValidArgs(1),
```

I don't think we should create dedicated commands for these since they would all be exactly the same. For now we are using only `get` but we'll need `store, get, erase, list, version`

pkg/okteto/config.go
maroshii: This function is mostly executed by internal packages outside of a "context" (registry, credential helpers, etc).

Changing this signature to add a  context would require more more code to account for it and adding `context.TODO()` to the callers.

The `func(registryHost string) (string, string, error)` is a known signatures by all these packages that we need to respect

maroshii: I made some changes to improve this since, at least for buildkit's auth provider, we do have access to the context and we now respect it: https://github.com/okteto/okteto/pull/3799/commits/a537160c89da114b757050b660fc828b5e2102dc

PR #3803, opened 2023-07-07
Initialize registrytoken command as docker credentials helper

maroshii:
Creates a new (hidden) command that can be used as a [docker credentials helper](https://github.com/docker/docker-credential-helpers):

```golang
echo "1234567890.dkr.ecr.us-east-2.amazonaws.com" | okteto registrytoken get
{"ServerURL": "1234567890.dkr.ecr.us-east-2.amazonaws.com", "Username": "AWS", "Secret": "eyJwZqwer......"}
```

### DIFF
--- a/cmd/registrytoken/registrytoken.go
+++ b/cmd/registrytoken/registrytoken.go
@@ -1,0 +1,69 @@
+// Copyright 2023 The Okteto Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package registrytoken
+
+import (
+	"context"
+	"os"
+
+	contextCMD "github.com/okteto/okteto/cmd/context"
+
+	"github.com/okteto/okteto/pkg/auth/dockercredentials"
+	"github.com/okteto/okteto/pkg/errors"
+	"github.com/okteto/okteto/pkg/okteto"
+	"github.com/spf13/cobra"
+
+	"github.com/docker/docker-credential-helpers/credentials"
+)
+
+type regCreds struct {
+	okteto.Config
+}
+
+func (r regCreds) GetRegistryCredentials(host string) (string, string, error) {
+	return r.GetExternalRegistryCredentials(host)
+}
+
+func RegistryToken() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "registrytoken",
+		Short: "docker credentials helper for private registries registered in okteto",
+		Long: `Acts as a docker credentials helper printing out private registry credentials previously configured in okteto:
+
+Usage: echo gcr.io | okteto registrytoken get
+
+More info about docker credentials helpers here: https://github.com/docker/docker-credential-helpers
+  `,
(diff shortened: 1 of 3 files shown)


cmd/registrytoken/registrytoken.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/okteto/okteto/pkg/auth/dockercredentials"
+	"github.com/okteto/okteto/pkg/errors"
+	"github.com/okteto/okteto/pkg/okteto"
+	"github.com/spf13/cobra"
+
+	"github.com/docker/docker-credential-helpers/credentials"
+)
+
+type regCreds struct {
+	okteto.Config
```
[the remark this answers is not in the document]
ifbyol: `GetRegistryCredentials` is internally calling to `r.GetExternalRegistryCredentials` which is a function defined for the `okteto.Config` struct. Embedding the struct there allows to call to `r.GetExternalRegistryCredentials` (and any other possible function)

ifbyol (approved):

cmd/registrytoken/registrytoken.go
```diff
@@ -0,0 +1,69 @@
...
+		Use:   "registrytoken",
+		Short: "docker credentials helper for private registries registered in okteto",
+		Long: `Acts as a docker credentials helper printing out private registry credentials previously configured in okteto:
+
+Usage: echo gcr.io | okteto registrytoken get
+
+More info about docker credentials helpers here: https://github.com/docker/docker-credential-helpers
+  `,
+		Hidden:    true,
+		ValidArgs: []string{"get"},
```
andreafalzetti, line 49: I was curious on the approach used here. We pass `get` as argument to this command. Is it a simpler approach instead of building `get` as sub-command of `registrytoken`?

Are we planning to have different arguments in the future?

andreafalzetti: Oh, I see the action go straight into the docker HandleCommand func, makes sense

pkg/auth/dockercredentials/helper.go
```diff
@@ -0,0 +1,44 @@
+package dockercredentials
```
andreafalzetti, line 1: Shall we add the open-source header here?

cmd/registrytoken/registrytoken.go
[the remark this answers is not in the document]
maroshii: Yeah,  we could have avoided this intermediate struct by naming the internal interface with the exact name but "external" is not really relevant in that context

cmd/registrytoken/registrytoken.go
```diff
@@ -0,0 +1,69 @@
...
+		Args:      cobra.OnlyValidArgs,
+	}
+
+	cmd.RunE = func(_ *cobra.Command, args []string) error {
+		action := args[0]
+		ctx := context.Background()
+		err := contextCMD.NewContextCommand().Run(ctx, &contextCMD.ContextOptions{})
```
teresaromero, line 56: nit: can we use `if err := ... ; err != nil{return err}`

teresaromero (approved):

pkg/auth/dockercredentials/helper.go
```diff
@@ -0,0 +1,56 @@
...
+
+type RegistryCredentialsGetter interface {
+	GetRegistryCredentials(host string) (string, string, error)
+}
+
+type OktetoClusterHelper struct {
+	getter RegistryCredentialsGetter
+}
+
+var _ credentials.Helper = (*OktetoClusterHelper)(nil)
```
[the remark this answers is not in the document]
maroshii: This asserts that `OktetoClusterHelper` in fact satisfies the `credentials.Helper` interface. It's not strictly necessary as it would fail to compile we try to use it or pass it down as such but the error will be less cryptic.

PR #3804, opened 2023-07-10
feat: add `--output` flag for `okteto preview list`

### LINKED ISSUE #3772: add `--output` flag to `okteto preview list` command
**Describe the bug**
```
ok preview list --log-output json
```

```
Name                                        Scope     Sleeping  Labels
test-quickstart                             personal  true      -
msd-pr-20                                   global    true      -
oktaco-pr-3                                 global    true      -
oktaco-pr-4                                 global    true      -
oktaco-pr-6                                 global    true      -
preview-8-update-voting-app-to-version-2-0  global    true      -
```

**Expected behavior**
The output is formated in json

nitishfy:
# Proposed changes

Fixes #3772

image

- Added the `--output` flag for `json` & `yaml` to `okteto preview list` command
- Added the test cases to check whether the output format is correct

### DIFF
--- a/cmd/preview/list.go
+++ b/cmd/preview/list.go
@@ -15,34 +15,46 @@ package preview
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
-	"os"
-	"strings"
-	"text/tabwriter"
-
 	contextCMD "github.com/okteto/okteto/cmd/context"
 	oktetoErrors "github.com/okteto/okteto/pkg/errors"
+	oktetoLog "github.com/okteto/okteto/pkg/log"
 	"github.com/okteto/okteto/pkg/okteto"
+	"github.com/okteto/okteto/pkg/types"
 	"github.com/spf13/cobra"
+	"gopkg.in/yaml.v2"
+	"os"
+	"text/tabwriter"
 )
 
 // ListFlags are the flags available for list commands
 type ListFlags struct {
 	labels []string
+	output string
+}
+
+type PreviewOutput struct {
+	Name     string   `json:"name" yaml:"name"`
+	Scope    string   `json:"scope" yaml:"scope"`
+	Sleeping bool     `json:"sleeping" yaml:"sleeping"`
+	Labels   []string `json:"labels" yaml:"labels"`
 }
 
 // List lists all the previews
 func List(ctx context.Context) *cobra.Command {
 	flags := &ListFlags{}
-
 	cmd := &cobra.Command{
 		Use:   "list",
 		Short: "List all preview environments",
 		RunE: func(cmd *cobra.Command, args []string) error {
+			ctxOptions := &contextCMD.ContextOptions{}
 
-			if err := contextCMD.NewContextCommand().Run(ctx, &contextCMD.ContextOptions{
-				Show: true,
-			}); err != nil {
+			if flags.output == "" {
+				ctxOptions.Show = true
+			}
+
+			if err := contextCMD.NewContextCommand().Run(ctx, ctxOptions); err != nil {
 				return err
 			}
 
@@ -50,12 +62,16 @@ func List(ctx context.Context) *cobra.Command {
 				return oktetoErrors.ErrContextIsNotOktetoCluster
 			}
 
+			if err := ValidateOutput(flags.output); err != nil {
+				return err
+			}
+
 			err := executeListPreviews(ctx, *flags)
 			return err
-
 		},
 	}
 	cmd.Flags().StringArrayVarP(&flags.labels, "label", "", []string{}, "tag and organize preview environments using labels (multiple --label flags accepted)")
+	cmd.Flags().StringVarP(&flags.output, "output", "o", "", "output format. One of: ['json', 'yaml']")
 
 	return cmd
 }
@@ -72,16 +88,65 @@ func executeListPreviews(ctx context.Context, opts ListFlags) error {
 		}
 		return fmt.Errorf("failed to get preview environments: %s", err)
 	}
-	w := tabwriter.NewWriter(os.Stdout, 1, 1, 2, ' ', 0)
-	fmt.Fprintf(w, "Name\tScope\tSleeping\tLabels\n")
-	for _, preview := range previewList {
-		previewLabels := "-"
-		if len(preview.PreviewLabels) > 0 {
-			previewLabels = strings.Join(preview.PreviewLabels, ", ")
+	switch opts.output {
+	case "json":
+		previewListOutput, err := getPreviewOutput(ctx, opts, oktetoClient)
+		if err != nil {
+			return err
 		}
-		fmt.Fprintf(w, "%s\t%s\t%v\t%s\n", preview.ID, preview.Scope, preview.Sleeping, previewLabels)
+		bytes, err := json.MarshalIndent(previewListOutput, "", " ")
+		if err != nil {
+			return err
+		}
+		oktetoLog.Println(string(bytes))
+	case "yaml":
+		previewListOutput, err := getPreviewOutput(ctx, opts, oktetoClient)
+		if err != nil {
+			return err
+		}
+		bytes, err := yaml.Marshal(previewListOutput)
+		if err != nil {
+			return err
+		}
+		oktetoLog.Println(string(bytes))
+	default:
+		w := tabwriter.NewWriter(os.Stdout, 1, 1, 2, ' ', 0)
+		fmt.Fprintf(w, "Name\tScope\tSleeping\tLabels\n")
+		for _, preview := range previewList {
+			previewLabels := "-"
+			fmt.Fprintf(w, "%s\t%s\t%v\t%s\n", preview.ID, preview.Scope, preview.Sleeping, previewLabels)
+		}
+		w.Flush()
 	}
-
-	w.Flush()
 	return nil
+}
+
+func getPreviewOutput(ctx context.Context, opts ListFlags, oktetoClient types.OktetoInterface) ([]PreviewOutput, error) {
+	var previewSlice []PreviewOutput
+	previewList, err := oktetoClient.Previews().List(ctx, opts.labels)
+	if err != nil {
+		if uErr, ok := err.(oktetoErrors.UserError); ok {
+			return nil, uErr
+		}
+		return nil, fmt.Errorf("failed to get preview environments: %s", err)
+	}
+	for _, preview := range previewList {
+		previewOutput := PreviewOutput{
+			Name:     preview.ID,
+			Scope:    preview.Scope,
+			Sleeping: preview.Sleeping,
+			Labels:   preview.PreviewLabels,
+		}
+		previewSlice = append(previewSlice, previewOutput)
+	}
+	return previewSlice, nil
+}
+
+func ValidateOutput(output string) error {
+	switch output {
+	case "", "json", "yaml":
+		return nil
+	default:
+		return fmt.Errorf("output format is not accepted. Value must be one of: ['json', 'yaml']")
+	}
 }
--- a/cmd/preview/list_test.go
+++ b/cmd/preview/list_test.go
@@ -1,0 +1,143 @@
+package preview
+
+import (
+	"context"
+	"fmt"
+	"github.com/okteto/okteto/internal/test/client"
+	oktetoErrors "github.com/okteto/okteto/pkg/errors"
+	"github.com/okteto/okteto/pkg/okteto"
+	"github.com/okteto/okteto/pkg/types"
+	"github.com/stretchr/testify/assert"
+	"testing"
+)
+
+func Test_listPreview(t *testing.T) {
+	ctx := context.Background()
+	opts := ListFlags{}
+	var tests = []struct {
+		name           string
+		previews       []types.Preview
+		expectedResult []PreviewOutput
+		expectedErr    error
+	}{
+		{
+			name: "List all previews",
+			previews: []types.Preview{
+				{
+					ID:            "test",
+					PreviewLabels: []string{"label-1", "label-2"},
+				},
+				{
+					ID:            "test-1",
+					PreviewLabels: []string{"label-3"},
+				},
+				{
+					ID:            "test-2",
+					Sleeping:      true,
+					PreviewLabels: []string{"-"},
+				},
+			},
+			expectedResult: []PreviewOutput{
+				{
+					Name:     "test",
+					Scope:    "",
+					Sleeping: false,
+					Labels:   []string{"label-1", "label-2"},
+				},
+				{
+					Name:     "test-1",
+					Scope:    "",
+					Sleeping: false,
+					Labels:   []string{"label-3"},
+				},
+				{
+					Name:     "test-2",
+					Scope:    "",
+					Sleeping: true,
+					Labels:   []string{"-"},
+				},
+			},
+		},
+		{
+			name:           "error retrieving preview list",
+			previews:       nil,
+			expectedResult: nil,
+			expectedErr:    fmt.Errorf("error retrieving previews"),
+		},
+		{
+			name:           "user error retrieving preview list",
+			previews:       nil,
+			expectedResult: nil,
+			expectedErr: oktetoErrors.UserError{
+				E:    fmt.Errorf("error retrieving previews"),
+				Hint: "please try again",
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			okteto.CurrentStore = &okteto.OktetoContextStore{
+				Contexts: map[string]*okteto.OktetoContext{
+					"test": {
+						Name:     "test",
+						Token:    "test",
+						IsOkteto: true,
+						UserID:   "1",
+					},
+				},
+				CurrentContext: "test",
+			}
+			usr := &types.User{
+				Token: "test",
+			}
+			fakeOktetoClient := &client.FakeOktetoClient{
+				Preview: client.NewFakePreviewClient(&client.FakePreviewResponse{PreviewList: tt.previews, ErrList: tt.expectedErr}),
+				Users:   client.NewFakeUsersClient(usr),
+			}
+			result, err := getPreviewOutput(ctx, opts, fakeOktetoClient)
+			if tt.expectedErr != nil {
+				assert.Error(t, err)
+			} else {
+				assert.NoError(t, err)
+				assert.ElementsMatch(t, result, tt.expectedResult)
+			}
+		})
+	}
+}
+
+func Test_PreviewListOutputValidation(t *testing.T) {
+	var tests = []struct {
+		name        string
+		output      ListFlags
+		expectedErr error
+	}{
+		{
+			name: "ouput format is yaml",
+			output: ListFlags{
+				output: "yaml",
+			},
+			expectedErr: nil,
+		},
+		{
+			name: "output format is json",
+			output: ListFlags{
+				output: "json",
+			},
+			expectedErr: nil,
+		},
+		{
+			name: "output format is not valid",
+			output: ListFlags{
+				output: "xml",
+			},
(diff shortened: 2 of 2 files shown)


cmd/preview/list.go
```diff
@@ -73,15 +88,66 @@ func executeListPreviews(ctx context.Context, opts ListFlags) error {
 		return fmt.Errorf("failed to get preview environments: %s", err)
 	}
 	w := tabwriter.NewWriter(os.Stdout, 1, 1, 2, ' ', 0)
```
andreafalzetti, line 90: do we need a new instance of `tabwriter` when we output JSON/YAML?

cmd/preview/list.go
```diff
@@ -73,15 +88,66 @@ func executeListPreviews(ctx context.Context, opts ListFlags) error {
...
+			previewLabels := "-"
+			if len(preview.PreviewLabels) > 0 {
+				previewLabels = strings.Join(preview.PreviewLabels, ", ")
+			}
+			fmt.Fprintf(w, "%s\t%s\t%v\t%s\n", preview.ID, preview.Scope, preview.Sleeping, previewLabels)
 		}
-		fmt.Fprintf(w, "%s\t%s\t%v\t%s\n", preview.ID, preview.Scope, preview.Sleeping, previewLabels)
 	}
-
 	w.Flush()
```
andreafalzetti, line 142: if we output in JSON/YAML, is this `w.Flush()` really needed?

cmd/preview/list.go
```diff
@@ -73,15 +88,66 @@ func executeListPreviews(ctx context.Context, opts ListFlags) error {
...
 	}
 	w := tabwriter.NewWriter(os.Stdout, 1, 1, 2, ' ', 0)
-	fmt.Fprintf(w, "Name\tScope\tSleeping\tLabels\n")
-	for _, preview := range previewList {
-		previewLabels := "-"
-		if len(preview.PreviewLabels) > 0 {
-			previewLabels = strings.Join(preview.PreviewLabels, ", ")
+	switch opts.output {
+	case "json":
+		var previewSlice []PreviewJson
```
andreafalzetti, line 93: It would be nice to keep it [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) and extract this duplicated code to a func

cmd/preview/list.go
nitishfy: Probably no, should i put this in default case?

cmd/preview/list.go
nitishfy: without `w.Flush()` the default behaviour will not be executed. However, we don't need it for `json` or `yaml`

andreafalzetti: @nitishfy I suggest to move all tablewriter logic inside the default case

cmd/preview/list_test.go
```diff
@@ -0,0 +1,23 @@
...
+func TestExecutePreviewListWithError(t *testing.T) {
+	opts := ListFlags{
+		output: "mdx",
+	}
+	err := ValidateOutput(opts.output)
+	require.Error(t, err)
+}
+
+func TestExecutePreviewListWithoutError(t *testing.T) {
+	//ctx := context.Background()
```
andreafalzetti, line 17: shall we remove this commented line?

cmd/preview/list.go
```diff
@@ -73,15 +88,66 @@ func executeListPreviews(ctx context.Context, opts ListFlags) error {
...
+			}
+			previewSlice = append(previewSlice, previewJson)
+		}
+		bytes, err := json.MarshalIndent(previewSlice, "", " ")
+		if err != nil {
+			return err
+		}
+		oktetoLog.Println(string(bytes))
+	case "yaml":
+		var previewSlice []PreviewJson
```
AdrianPedriza, line 113:
```suggestion
		var previewSlice []PreviewOutput
```
since it is used for control the output in both cases

cmd/preview/list.go
```diff
@@ -15,20 +15,30 @@ package preview
...
+	"text/tabwriter"
 )
 
 // ListFlags are the flags available for list commands
 type ListFlags struct {
 	labels []string
+	output string
+}
+
+type PreviewJson struct {
```
AdrianPedriza, line 37:
```suggestion
type PreviewOutput struct {
```

cmd/preview/list_test.go
```diff
@@ -0,0 +1,23 @@
+package preview
+
+import (
+	"github.com/stretchr/testify/require"
+	"testing"
+)
+
+func TestExecutePreviewListWithError(t *testing.T) {
```
AdrianPedriza, line 8:
```suggestion
func TestPreviewListOutputValidationWithError(t *testing.T) {
```

cmd/preview/list_test.go
```diff
@@ -0,0 +1,23 @@
...
+	opts := ListFlags{
+		output: "mdx",
+	}
+	err := ValidateOutput(opts.output)
+	require.Error(t, err)
+}
+
+func TestExecutePreviewListWithoutError(t *testing.T) {
```
AdrianPedriza, line 16:
```suggestion
func TestPreviewListOutputValidation(t *testing.T) {
```

cmd/preview/list.go
nitishfy: Resolved. I used a little different approach

cmd/preview/list_test.go
nitishfy: Resolved

cmd/preview/list.go
nitishfy: Resolved

cmd/preview/list.go
nitishfy: Resolved

nitishfy:
**Update:** Ready for the next set of review.

cmd/preview/list.go
```diff
@@ -72,16 +86,72 @@ func executeListPreviews(ctx context.Context, opts ListFlags) error {
...
+		}
+		return nil, fmt.Errorf("failed to get preview environments: %s", err)
+	}
 	for _, preview := range previewList {
 		previewLabels := "-"
 		if len(preview.PreviewLabels) > 0 {
 			previewLabels = strings.Join(preview.PreviewLabels, ", ")
 		}
-		fmt.Fprintf(w, "%s\t%s\t%v\t%s\n", preview.ID, preview.Scope, preview.Sleeping, previewLabels)
+		previewJson := PreviewOutput{
```
AdrianPedriza, line 139:
```suggestion
		previewOutput := PreviewOutput{
```

cmd/preview/list.go
```diff
@@ -72,16 +86,72 @@ func executeListPreviews(ctx context.Context, opts ListFlags) error {
 		}
 		return fmt.Errorf("failed to get preview environments: %s", err)
 	}
-	w := tabwriter.NewWriter(os.Stdout, 1, 1, 2, ' ', 0)
-	fmt.Fprintf(w, "Name\tScope\tSleeping\tLabels\n")
+	previewListOutput, err := getPreviewSlice(ctx, opts)
```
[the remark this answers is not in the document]
nitishfy: Resolved

cmd/preview/list.go
```diff
@@ -72,16 +86,72 @@ func executeListPreviews(ctx context.Context, opts ListFlags) error {
 		}
 		return fmt.Errorf("failed to get preview environments: %s", err)
 	}
-	w := tabwriter.NewWriter(os.Stdout, 1, 1, 2, ' ', 0)
-	fmt.Fprintf(w, "Name\tScope\tSleeping\tLabels\n")
+	previewListOutput, err := getPreviewOutput(ctx, opts)
```
andreafalzetti, line 89: why not moving the err check just after this line and avoid repeating it inside each switch case?

cmd/preview/list.go
```diff
@@ -72,16 +86,72 @@ func executeListPreviews(ctx context.Context, opts ListFlags) error {
...
+	switch opts.output {
+	case "json":
+		if err != nil {
+			return err
+		}
```
andreafalzetti, line 94: I think we should move `previewListOutput, err := getPreviewOutput(ctx, opts)` inside the switch, for both JSON and YAML.

When the user runs the "default" case, which is the most common, we don't need to process this output. It will be an extra line repeated but then each case is handled separately properly

cmd/preview/list.go
```diff
@@ -72,16 +86,72 @@ func executeListPreviews(ctx context.Context, opts ListFlags) error {
...
+			Labels:   previewLabels,
+		}
+		previewSlice = append(previewSlice, previewOutput)
 	}
+	return previewSlice, nil
+}
 
-	w.Flush()
-	return nil
+func ValidateOutput(output string) error {
```
andreafalzetti, line 150:
```suggestion
func ValidateOutputFlag(output string) error {
```

Just to improve readability of this function, and avoiding the reader to think it's about validating the actual output

cmd/preview/list.go
```diff
@@ -15,20 +15,30 @@ package preview
...
 )
 
 // ListFlags are the flags available for list commands
 type ListFlags struct {
 	labels []string
+	output string
+}
+
+type PreviewOutput struct {
+	Name     string `json:"Name" yaml:"name"`
```
andreafalzetti, line 38: the attribute names should be lowercase

cmd/preview/list.go
nitishfy: Isn't it a kind of little redundancy we're adding?

cmd/preview/list.go
nitishfy: Resolved

cmd/preview/list.go
nitishfy: > I don't think so, just both cases need that function call. What I do believe we should do is to generate an output only if it is strictly necessary. In the end we are executing a logic that in most cases is not going to be necessary as Andrea says. Of course we could refactor the code to call the function once for both cases (JSON and YAML) but I think in this case calling the function in both cases makes it pretty clear what logic to follow in both cases avoiding to generate the output in the default cases.

Sounds good. Thanks!

cmd/preview/list.go
```diff
@@ -72,16 +86,73 @@ func executeListPreviews(ctx context.Context, opts ListFlags) error {
 		}
 		return fmt.Errorf("failed to get preview environments: %s", err)
 	}
-	w := tabwriter.NewWriter(os.Stdout, 1, 1, 2, ' ', 0)
-	fmt.Fprintf(w, "Name\tScope\tSleeping\tLabels\n")
+	switch opts.output {
+	case "json":
+		if err != nil {
```
andreafalzetti, line 91: @nitishfy the line `previewListOutput, err := getPreviewOutput(ctx, opts)` should go before you check for error nil

nitishfy: Aah! sorry, didn't notice this thing. It's resolved now.

cmd/preview/list.go
```diff
@@ -72,16 +86,73 @@ func executeListPreviews(ctx context.Context, opts ListFlags) error {
...
+		}
+		bytes, err := json.MarshalIndent(previewListOutput, "", " ")
+		if err != nil {
+			return err
+		}
+		oktetoLog.Println(string(bytes))
+	case "yaml":
+		previewListOutput, err := getPreviewOutput(ctx, opts)
+		bytes, err := yaml.Marshal(previewListOutput)
+		if err != nil {
```
andreafalzetti, line 103: we need to check if `getPreviewOutput` failed before marshalling `previewListOutput` - if the previous function failed, this most likely will result in some panic

nitishfy: Resovled

cmd/preview/list.go
```diff
@@ -72,16 +86,76 @@ func executeListPreviews(ctx context.Context, opts ListFlags) error {
...
+		previewOutput := PreviewOutput{
+			Name:     preview.ID,
+			Scope:    preview.Scope,
+			Sleeping: preview.Sleeping,
+			Labels:   previewLabels,
+		}
+		previewSlice = append(previewSlice, previewOutput)
 	}
+	return previewSlice, nil
+}
```
andreafalzetti, line 152: It should be fairly easy to add tests for this function - we're currently focusing a lot of attention to CLI quality so if you could add some more tests to this PR, it'll be very much appreciated

nitishfy: working on this! I'll update once it's done.

AdrianPedriza: I will change the state to draft then 👍

nitishfy: Added!

nitishfy:
Sorry for being late, the test case has been added. This PR is ready for next set of review!
cc @AdrianPedriza @andreafalzetti

jLopezbarb:
I've been testing it and it works perfectly, just one thing. Can we not show the context information at the beginning when it's json? If we do, it's not parseable. You have an example of covering this [here](https://github.com/okteto/okteto/blob/master/cmd/deploy/endpoints.go#L105).



Captura de Pantalla 2023-07-13 a las 17 21 57

nitishfy:
> I've been testing it and it works perfectly, just one thing. Can we not show the context information at the beginning when it's json? If we do, it's not parseable. You have an example of covering this [here](https://github.com/okteto/okteto/blob/master/cmd/deploy/endpoints.go#L105).

Why do we specifically need for `json`? We already have this piece of code in the cmd:
```
if err := contextCMD.NewContextCommand().Run(ctx, &contextCMD.ContextOptions{
				Show: true,
			});
```

cmd/preview/list.go
```diff
@@ -15,20 +15,31 @@ package preview
...
 type ListFlags struct {
 	labels []string
+	output string
+}
+
+type PreviewOutput struct {
+	Name     string `json:"name" yaml:"name"`
+	Scope    string `json:"scope" yaml:"scope"`
+	Sleeping bool   `json:"sleeping" yaml:"sleeping"`
+	Labels   string `json:"labels" yaml:"labels"`
```
AdrianPedriza, line 42: do we really need make public this fields?

cmd/preview/list_test.go
```diff
@@ -0,0 +1,103 @@
...
+				{
+					ID: "test-1",
+				},
+			},
+		},
+		{
+			name:     "error retrieving preview list",
+			previews: nil,
+			err:      fmt.Errorf("error retrieving previews"),
+		},
```
AdrianPedriza, line 36: I think we are missing a test case that checks the flow when the preview list returns a user error
``````go

{
	name:     "user error retrieving preview list",
	previews: nil,
	err:      oktetoErrors.UserError{},
}
``````

File: cmd/preview/list_test.go [hunk 20]
```diff
@@ -0,0 +1,103 @@
...
+				CurrentContext: "test",
+			}
+			usr := &types.User{
+				Token: "test",
+			}
+			fakeOktetoClient := &client.FakeOktetoClient{
+				Preview: client.NewFakePreviewClient(&client.FakePreviewResponse{PreviewList: tt.previews, ErrList: tt.err}),
+				Users:   client.NewFakeUsersClient(usr),
+			}
+			_, err := getPreviewOutput(ctx, opts, fakeOktetoClient)
```
AdrianPedriza, line 59: we need to check the result too. You can use `assert.ElementsMatch(t, result, tt.expectedResult)`

jLopezbarb:
> > I've been testing it and it works perfectly, just one thing. Can we not show the context information at the beginning when it's json? If we do, it's not parseable. You have an example of covering this [here](https://github.com/okteto/okteto/blob/master/cmd/deploy/endpoints.go#L105).
> 
> Why do we specifically need for `json`? We already have this piece of code in the cmd:
> 
> ```
> if err := contextCMD.NewContextCommand().Run(ctx, &contextCMD.ContextOptions{
> 				Show: true,
> 			});
> ```

The thing is that the `-o json` must be parseable so I can do `okteto preview list -o json | jq -r '.[].name'` or something similar. If we have the context message users won't be able to do that. You can split the `context.run` in definition, add show if the output is "" and then execute it. WDYT?

nitishfy:
> The thing is that the `-o json` must be parseable so I can do `okteto preview list -o json | jq -r '.[].name'` or something similar. If we have the context message users won't be able to do that. You can split the `context.run` in definition, add show if the output is "" and then execute it. WDYT?

This needs to be done [here](https://github.com/okteto/okteto/blob/master/cmd/preview/list.go#L41) or specifically in the new function I've created?

cmd/preview/list.go
AdrianPedriza: oh I see, if not we cannot Marshal it. 👍

cmd/preview/list.go
```diff
@@ -72,16 +87,72 @@ func executeListPreviews(ctx context.Context, opts ListFlags) error {
...
+		if uErr, ok := err.(oktetoErrors.UserError); ok {
+			return nil, uErr
+		}
+		return nil, fmt.Errorf("failed to get preview environments: %s", err)
+	}
 	for _, preview := range previewList {
 		previewLabels := "-"
 		if len(preview.PreviewLabels) > 0 {
 			previewLabels = strings.Join(preview.PreviewLabels, ", ")
 		}
```
AdrianPedriza, line 139: I don't think we should convert this list to string. We should marshal the labels list as it

jLopezbarb:
> > The thing is that the `-o json` must be parseable so I can do `okteto preview list -o json | jq -r '.[].name'` or something similar. If we have the context message users won't be able to do that. You can split the `context.run` in definition, add show if the output is "" and then execute it. WDYT?
> 
> This needs to be done [here](https://github.com/okteto/okteto/blob/master/cmd/preview/list.go#L41) or specifically in the new function I've created?

There yes, on the[ line 43](https://github.com/okteto/okteto/blob/master/cmd/preview/list.go#L43)

cmd/preview/list.go
nitishfy: How about removing this line completely?

AdrianPedriza: absolutely!

cmd/preview/list_test.go
nitishfy: Resolved

nitishfy:
**Update:** This PR is ready for the next set of review

cmd/preview/list.go
```diff
@@ -72,16 +88,66 @@ func executeListPreviews(ctx context.Context, opts ListFlags) error {
...
+	}
 	for _, preview := range previewList {
 		previewLabels := "-"
-		if len(preview.PreviewLabels) > 0 {
-			previewLabels = strings.Join(preview.PreviewLabels, ", ")
+		previewOutput := PreviewOutput{
+			Name:     preview.ID,
+			Scope:    preview.Scope,
+			Sleeping: preview.Sleeping,
+			Labels:   previewLabels,
```
AdrianPedriza, line 139:
```suggestion
			Labels:   preview.PreviewLabels.,
```
we are always set `-` as value

cmd/preview/list_test.go
```diff
@@ -0,0 +1,130 @@
...
+					Name:     "test",
+					Scope:    "",
+					Sleeping: false,
+					Labels:   "-",
+				},
+				{
+					Name:     "test-1",
+					Scope:    "",
+					Sleeping: false,
+					Labels:   "-",
```
AdrianPedriza, line 44: we need to add more info to the preview. At least enough to check that the conversion to `previewOutput` works correctly. With the current information we were not catching the bug that we always set to `-` the label field

nitishfy: Resolved

jLopezbarb:
Hi again @nitishfy and thanks for your contributions and hard work!!!

Now it's perfect and sorry for the back and forward messaging, We keep improving to try to give all the feedback at once but is something that we have to improve.

Tested and the code looks and works correctly and the tests are amazing 😍 

Captura de pantalla 2023-07-18 a las 19 07 41

andreafalzetti (approved):
Thanks @nitishfy, well done!

nitishfy:
Great to get feedback from folks I take inspiration from, looking forward to contributing more!

PR #3806, opened 2023-07-10
fix: fail if dev mode is not valid

### LINKED ISSUE #3778: Add warning when falling back to "sync" mode
When parsing the manifest, if the `mode` field contains an unknown value (different from `hybrid` or `sync`), it will fallback to `sync` silently. The CLI should instead warn the user about this behaviour so that as a user I can realize the mistake and fix my manifest.

AdrianPedriza:
# Context
when a development mode is neither `sync` nor `hybrid`, we use `sync` mode as default. We should inform the user that the mode is not correct instead fallback `sync` mode.

# Proposed changes
- raise an error if development mode is not `sync` nor `hybrid``
- add test covering new logic

Fixes #3778

# How to test it
1. create a manifest with an invalid development mode:
``````yaml
dev:
  frontend:
    mode: synco-invalid
    image: okteto/node:18
    sync:
      - ./frontend:/app
    workdir: /app
    command: ["bash"]
``````
2. Check error is raised:
Captura de Pantalla 2023-07-10 a las 14 32 03

### DIFF
--- a/pkg/errors/errors.go
+++ b/pkg/errors/errors.go
@@ -210,6 +210,9 @@ var (
 
 	// ErrTimeout is raised when an operation has timed out
 	ErrTimeout = fmt.Errorf("operation timed out")
+
+	//ErrDevModeNotValid is raised when development mode in manifest is not 'sync' nor 'hybrid'
+	ErrDevModeNotValid = errors.New("development mode not valid. Value must be one of: ['sync', 'hybrid']")
 )
 
 // IsAlreadyExists raised if the Kubernetes API returns AlreadyExists
--- a/pkg/model/serializer.go
+++ b/pkg/model/serializer.go
@@ -28,6 +28,7 @@ import (
 	"github.com/kballard/go-shellquote"
 	"github.com/okteto/okteto/pkg/cache"
 	"github.com/okteto/okteto/pkg/constants"
+	"github.com/okteto/okteto/pkg/errors"
 	"github.com/okteto/okteto/pkg/externalresource"
 	oktetoLog "github.com/okteto/okteto/pkg/log"
 	"github.com/okteto/okteto/pkg/model/forward"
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: pkg/errors/errors.go [hunk 1]
```diff
@@ -210,6 +210,9 @@ var (
 
 	// ErrTimeout is raised when an operation has timed out
 	ErrTimeout = fmt.Errorf("operation timed out")
+
+	//ErrDevModeNotValid is raised when development mode in manifest is not 'sync' nor 'hybrid'
```
jLopezbarb, line 214: I want to open a discussion here about if the errors should be defined on the package that is producing the error or in a common/utils file. I think it's better to have it in the serializer file but I'm nitpicking

AdrianPedriza: +1, if is not necessary out of the package let's keep it inside. I will update it

andreafalzetti: about the error, should this be a `UserError` instead?

andreafalzetti (approved):
tested and works as expected, I understand that we'll move the error to its own pkg but I am already pre-approving it 👍

pkg/errors/errors.go
andreafalzetti: sounds good, let's keep the change simple

AdrianPedriza: refactor done d21bcfa484a558045b528d7db392286dd7bd9b42

pkg/model/serializer.go
```diff
@@ -38,6 +38,12 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+var (
+
+	//ErrDevModeNotValid is raised when development mode in manifest is not 'sync' nor 'hybrid'
+	ErrDevModeNotValid = errors.New("development mode not valid. Value must be one of: ['sync', 'hybrid']")
```
andreafalzetti, line 44: do we still want to expose this error outside the pkg?

AdrianPedriza: 362a5e9417a534ad172bd68ca404ef550df74fa3

pkg/model/serializer.go
```diff
@@ -37,6 +38,12 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+var (
+
+	// ErrDevModeNotValid is raised when development mode in manifest is not 'sync' nor 'hybrid'
```
andreafalzetti, line 43:
```suggestion
	// errDevModeNotValid is raised when development mode in manifest is not 'sync' nor 'hybrid'
```

or consider removing it

jLopezbarb (approved):

pkg/model/serializer.go
AdrianPedriza: 🙏 e3485e01ea5c2cab6bd82bacf0b5ba733ba5b1c7 . I keep it because I think it is not superfluous to have a description of the error.

PR #3808, opened 2023-07-11
[Hybrid] Support additional Dev fields

### LINKED ISSUE #3739: Supported fields for hybrid development
As part of the hybrid development feature we should support the following fields at the okteto dev manifest:

```
// annotations
// container
// args
// probes
// remote
// interface
// timeout
// metadata
// autocreate
// envFiles
// healthchecks
// labels
// environment
```

current supported:
```
type hybridModeInfo struct {
				Workdir     string            `json:"workdir,omitempty" yaml:"workdir,omitempty"`
				Selector    Selector          `json:"selector,omitempty" yaml:"selector,omitempty"`
				Forward     []forward.Forward `json:"forward,omitempty" yaml:"forward,omitempty"`
				Environment Environment       `json:"environment,omitempty" yaml:"environment,omitempty"`
				Command     Command           `json:"command,omitempty" yaml:"command,omitempty"`
				Reverse     []Reverse         `json:"reverse,omitempty" yaml:"reverse,omitempty"`
				Mode        string            `json:"mode,omitempty" yaml:"mode,omitempty"`
			}
```

andreafalzetti:
# Proposed changes

Fixes #3739

With this PR I've tested all manifest properties that should also work for services in Hybrid mode, as specified in #3739.

Most of them are already working out of the box with recent changes that we made in https://github.com/okteto/okteto/pull/3764, so, as part of this PR, I am mainly extending E2E tests to cover for fields:

- `envFiles`
- `environment`
- `metadata.annotations`
- `metadata.labels`
- `annotations` 

Sure, here's the information as a table:

| Fields                | Status                                                 |
|-----------------------|--------------------------------------------------------|
| annotations           | Works ✅                                               |
| container             | Works ✅                                               |
| args                  | Not implemented in sync either. Moved to separate issue: [Link](https://github.com/okteto/okteto/issues/3836) |
| probes                | Works ✅ |
| interface             | Works ✅ |
| remote                | Works ✅                                               |
| timeout               | Not working properly for sync either (Created dedicated issue: [Link](https://github.com/okteto/okteto/issues/3849)) |
| metadata.annotations  | Works ✅                                               |
| metadata.labels       | Works ✅                                               |
| autocreate            | Works ✅                                               |
| envFiles              | Works ✅                                               |
| environment           | Works ✅                                               |
| healthchecks          | Should work, but deprecated                            |
| labels                | Should work, but deprecated                            |

## How to test

I've tested manually all of those with the ✅  - for some of them I was able to modify the e2e test to include those in the automated test.

See passing E2E tests for Windows: https://app.circleci.com/pipelines/github/okteto/okteto/11166/workflows/0c8e7538-0a87-4537-8cdb-8ac1c0169a90/jobs/35422

### DIFF
--- a/integration/up/hybridmode_test.go
+++ b/integration/up/hybridmode_test.go
@@ -24,7 +24,6 @@ import (
 	"github.com/okteto/okteto/pkg/k8s/kubeconfig"
 	"github.com/okteto/okteto/pkg/okteto"
 	"github.com/stretchr/testify/require"
-	"log"
 	"os"
 	"path/filepath"
 	"testing"
@@ -122,7 +121,6 @@ func TestUpUsingHybridMode(t *testing.T) {
 	require.NoError(t, commands.RunOktetoKubeconfig(oktetoPath, dir))
 	c, _, err := okteto.NewK8sClientProvider().Provide(kubeconfig.Get([]string{filepath.Join(dir, ".kube", "config")}))
 	require.NoError(t, err)
-	log.Printf("Using kubeconfig: '%s'", filepath.Join(dir, ".kube", "config"))
 
 	require.NoError(t, writeFile(filepath.Join(dir, "okteto.yml"), hybridManifest))
 	require.NoError(t, writeFile(filepath.Join(dir, "docker-compose.yml"), hybridCompose))
--- a/integration/up/hybridmode_test.go
+++ b/integration/up/hybridmode_test.go
@@ -50,6 +50,8 @@ dev:
         e2e/test-1: annotation-1
       labels:
         custom.label/e2e: "true"
+    annotations:
+      deprecated.annotation.format: deprecated-annotation-1
 `
 	hybridCompose = `services:
  svc:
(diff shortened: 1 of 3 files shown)


integration/up/hybridmode_test.go
```diff
@@ -37,9 +38,21 @@ dev:
     context: svc
     namespace: user
     mode: hybrid
-    command: "{{ .Shell }} checker.sh"
+    command: ./checker.sh
```
jLopezbarb, line 41: I think we need the shell for windows tests. Probably @AdrianPedriza can provide more information about this but we saw that if we used `./script.sh` or `sh script.sh` wasn't working

andreafalzetti: For now I've set `bash` to simplify this test, until we can remove it entirely in the future.

integration/up/hybridmode_test.go
```diff
@@ -49,9 +61,25 @@ dev:
 
 	svcDockerfile = `FROM busybox
 ENV ENV_IN_IMAGE value_from_image`
+	envFile      = `TEST_ENV_FILE_VAR1=from-file-1`
 	localProcess = `
 #!/bin/bash
 
+if [ "$#" -ne 0 ]; then
```
andreafalzetti, line 68: this check is not particularly useful now, but it doesn't harm. It was initially helpful because as part of this PR I added the support for `args` which I moved to a separate  issue and I would like to re-use this check in the future once we introduce args.

teresaromero (approved):

jLopezbarb (approved):

AdrianPedriza:
Executing an okteto up using these fields I see a warning message repeated about deprecated fields. I don't know if this PR needs to address it but commenting it here since I discovered it testing the fields
Captura de Pantalla 2023-08-01 a las 13 35 24

integration/up/hybridmode_test.go
```diff
@@ -49,9 +61,25 @@ dev:
 
 	svcDockerfile = `FROM busybox
 ENV ENV_IN_IMAGE value_from_image`
```
AdrianPedriza, line 63: if I change Dockerfile env key to `ANOTHER_ENV` e2e test still passes. Could be that some envs from dockerfile or pod are not checked properly in the script?

andreafalzetti: As we discussed offline, this seems to be a bug in how we calculate the digest for rebuilding the dev container image: https://github.com/okteto/okteto/issues/3871

PR #3809, opened 2023-07-11
Fix win e2e

teresaromero:
# Proposed changes

Home directory was hardcoded to a cirlceci folder, which is changing. OKTETO_PATH should be referenced using $HOME

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -213,36 +213,39 @@ jobs:
           name: Run deprecated integration tests
           environment:
             - OKTETO_URL: https://staging.okteto.dev/
-            - OKTETO_PATH: 'C:\Users\circleci.PACKER-64370BA5\project\artifacts\bin\okteto.exe'
             - OKTETO_SKIP_CLEANUP: 'true'
             - OKTETO_APPS_SUBDOMAIN: staging.okteto.net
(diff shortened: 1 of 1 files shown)


andreafalzetti (approved):

AdrianPedriza (approved):

teresaromero:
> is a good idea to use `Prepare env` step to export `OKTETO_PATH` env into `BASH_ENV` and use it directly in the commands used by other steps? https://circleci.com/docs/env-vars/#using-parameters-and-bash-environment:~:text=Project%20Settings%20page.-,Example%20configuration%20of%20environment%20variables,-Consider%20the%20example

I've pushed some changes, lets see if this works :)

teresaromero:
> > is a good idea to use `Prepare env` step to export `OKTETO_PATH` env into `BASH_ENV` and use it directly in the commands used by other steps? https://circleci.com/docs/env-vars/#using-parameters-and-bash-environment:~:text=Project%20Settings%20page.-,Example%20configuration%20of%20environment%20variables,-Consider%20the%20example
> 
> I've pushed some changes, lets see if this works :)

Reverted as the env variable was not transmited to each job.

Lets review this within the CI initiative for CLI Quality

PR #3810, opened 2023-07-11
Add docker-credential-okteto to cli image

maroshii:
Creates an executable file that should be used as a docker credential helper.

For now, this is not added locally to the user path meaning that it won't work locally since the UX for this needs to be decided. TL;DR: `~/.docker/config.json`  reads this executable automatically

However, this works out of the box in `--remote` deploy because all binaries in `/usr/local/bin/*` are already copied to the target image. After this change this should work:

```yaml
deploy:
  commands:
    - echo my-registry.com | docker-credential-okteto get
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -41,6 +41,8 @@ RUN chmod +x /okteto/bin/okteto
 # Test
 RUN /okteto/bin/okteto version
 
+COPY docker-credential-okteto /okteto/bin/docker-credential-okteto
+
 FROM alpine:3
 
 RUN apk add --no-cache bash ca-certificates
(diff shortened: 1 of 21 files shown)


Dockerfile
```diff
@@ -41,6 +41,8 @@ RUN chmod +x /okteto/bin/okteto
 # Test
 RUN /okteto/bin/okteto version
 
+COPY docker-credential-okteto /okteto/bin/docker-credential-okteto
```
pchico83, line 44: is it possible to make it work in the installer too?
That way people could use third party tooling when not using `--remote?

maroshii: I just opened this PR: https://github.com/okteto/pipeline-installer-image/pull/114. Is that what you mean?

maroshii:
Too many conflicts due to a rebase in the base branch. Closing this PR and cherry-picking this change on top of `fran/build-external-credentials`

Dockerfile
pchico83: Exactly, thanks @maroshii !

PR #3811, opened 2023-07-12
Fix hybrid interactive terminal and processes

### LINKED ISSUE #3775: Hybrid mode doesn't work with a `bash` terminal
**Describe the bug**
When I am using `hybrid` with a bash terminal, execution get's stuck

**To Reproduce**
Run `okteto up` with the following settings:
```
    mode: hybrid
    command: bash
```    

**Expected behavior**
I get an interactive bash terminal

teresaromero:
# Proposed changes

Fixes #3775 

The grouping of processes done for setting up the command to run at the dev environment was causing that when a command was interactive, the terminal was not shown. That was the case for bash or sh. Also this was affecting the `okteto exec` command when a shell was added as command

```
c.SysProcAttr = &syscall.SysProcAttr{	
		Setpgid: true,	
	}
```

Once we investigated more on the processes, we evaluated the use of the property `Foreground` at the command attribute but that caused the child processes not to end when the parent was ending.

This PR proposes the removal con unix and windows implementation and removing the SusProcAttr for the command.

In order to kill/terminate the child processes running inside the parent (okteto up) we iterate though them and terminate them if the run of the command does not clean them properly.

Tested:

- using `frontend-local` using as command `bash` and `yarn start` processes start, and when control+c or exit, ports are released. When using bash the interactive terminal is shown.
- once `frontend-local` is running on dev, run `okteto exec`

Captured windows hybrid with this PR changes, processes are killed. Due i dont have bash on the windows instance i used sh as command on the dev manifest

https://github.com/okteto/okteto/assets/40767058/c78ada6d-1412-40bc-b8d1-eea7cf4bc4ba


cmd/up/exec_unix.go
```diff
@@ -7,22 +7,13 @@ import (
...
-	if runtime.GOOS != "windows" {
-		c = exec.Command("bash", "-c", strings.Join(cmd, " "))
-	} else {
-		binary, err := expandExecutableInCurrentDirectory(cmd[0], he.workdir)
-		if err != nil {
-			return nil, err
-		}
-		c = exec.Command(binary, cmd[1:]...)
-	}
+	c := exec.Command("bash", "-c", strings.Join(cmd, " "))
```
pchico83, line 16: why are we forcing `bash`?

teresaromero: This was already implemented, i've havent changed the code

cc @AdrianPedriza for context

AdrianPedriza: It was the same approach used by our default [_executors_](https://github.com/okteto/okteto/blob/master/cmd/utils/executor/executor.go#L61-L64) (run without bash is false by default so in most of the cases we are using `bash`) where we are always using `bash` unless we are running remotely deploy/detroy operation.

jLopezbarb, line 16: I'd use something like `shellquote.Join` instead of `strings.Join`. We could have several issues with `strings.Join`, for example with backslash characters

cmd/up/exec_unix.go
```diff
@@ -7,22 +7,13 @@ import (
 	"context"
 	"os"
 	"os/exec"
-	"runtime"
 	"strings"
 	"syscall"
 )
 
+// GetCommandToExec for non-windows GOOS
 func (he *hybridExecutor) GetCommandToExec(ctx context.Context, cmd []string) (*exec.Cmd, error) {
```
jLopezbarb, line 15: Do we need the context here?

cmd/up/exec_windows.go
```diff
@@ -32,9 +29,32 @@ func (he *hybridExecutor) GetCommandToExec(ctx context.Context, cmd []string) (*
...
 	c.SysProcAttr = &syscall.SysProcAttr{
+		// SysProcAttr for windows are different from linux,
+		// use CreationFlags to group the processes
 		CreationFlags: syscall.CREATE_NEW_PROCESS_GROUP,
 	}
 
 	return c, nil
 }
+
+func expandExecutableInCurrentDirectory(args0, dir string) (string, error) {
```
jLopezbarb, line 42: NIT: I'd go with `binaryName` instead of `args0`, but both of them are readable for me
```suggestion
func expandExecutableInCurrentDirectory(binaryName, dir string) (string, error) {
```

cmd/up/exec_windows.go
```diff
@@ -5,24 +5,21 @@ package up
...
-	var c *exec.Cmd
-	if runtime.GOOS != "windows" {
-		c = exec.Command("bash", "-c", strings.Join(cmd, " "))
-	} else {
-		binary, err := expandExecutableInCurrentDirectory(cmd[0], he.workdir)
-		if err != nil {
-			return nil, err
-		}
-		c = exec.Command(binary, cmd[1:]...)
+	binary, err := expandExecutableInCurrentDirectory(cmd[0], he.workdir)
```
jLopezbarb, line 18: Should we check that `len(cmd)>0`? I don't know if we could force that error and if it can happen but We had some runtime errors and we should try to prevent them

cmd/up/exec_windows.go
```diff
@@ -5,24 +5,21 @@ package up
...
+	if err != nil {
+		return nil, err
 	}
+	c := exec.Command(binary, cmd[1:]...)
```
jLopezbarb, line 22: What if the command is just the binary? I think `cmd[1:]...` will be out of index and return a runtime error

cmd/up/exec_unix.go
andreafalzetti: we agreed to revisit this logic and work on it to unblock all related changes / PRs. This is the new issue: https://github.com/okteto/okteto/issues/3816

teresaromero: just for the record, this PR does not change this piece of code, it was a refactor (move of code) that was previously implemented, does this new issue replace this one? i am not understanding.

teresaromero:
> I am experiencing a weird behaviour with this change, which breaks the execution of `okteto up` with the following error:
> 
> ```
>  x  Couldn't connect to your development container: local port 2345 is already in-use in your local machine: port is already allocated
> ```
> 
> When running `okteto up` there is a child process (I think syncthing) started and listening on port `2345`, the issue seems to be that when the okteto process ends, the child process remains alive and port `2345` is busy, so if you run `okteto up` again it will fail witht his error.
> 
> To reproduce I am using a random manifest with a simple command such as `echo` or `date`. Then in another terminal run `watch lsof -i tcp:2345` (if you don't have `watch`: either remove it or run `brew install watch`) to see the port beign listened on. With stable release, when the okteto CLI ends, the port is released. With the Foreground change, for some reasons it's not happening.
> 
> It looks like we need to investigate further which flags may help us here.

The Attributes added where to group the processes and terminate all the child and parent ones. I thought this was fixed as i am unable to reproduce this, I will try as you say and get back. Thanks

andreafalzetti:
@teresaromero I have been trying the latest version and it works as expected, not leaving any child procees behind! 👏 

By running `okteto up -l trace` I've noticed we always return this error, which is odd. 

```
DEBU[0013] error terminating child process %!s(int=15174): os: process already finished
DEBU[0013] error terminating process %!s(int=15174): os: process already finished
```

After investigating a bit I suspect that we might be killing the parent pid before the child processes, so that by the time the parent process is killed, the parent itself or the OS might have already terminated those child pids, so when we attempt, we error.

It's not a big deal per se, because the code works, but if we can get rid of the persistent error I think we would go the extra mile. 

Another thing I've noticed but I am not sure we can do much about it, is the following scenario:
- have `command: bash`
- once you are in the bash shell, run any command that returns a non-zero exit status. You can etier explicitelly type `exit 123` or for example simulate a `127` exit status by typing incorrectly a cmd "helloooo"
- notice the error "helloooo command not found"
- then type `exit`
- notice the CLI reporting that the `okteto up` failed with `127`

For some reason we are "inheriting" the exit code even if it's not necessarily connected.

teresaromero:
@andreafalzetti regarding the errors on trace, i added the condition to skip if the err is already done, the processes itself are terminated when shutdown, this is a gracefull in case its not responding... 


For the exit with command:bash i see the same behavour with hybrid and with sync. I think this is a expected when the command is a shell, but maybe @jLopezbarb can confirm? if not we should address a new issue

cmd/up/up.go
```diff
@@ -924,51 +924,56 @@ func (up *upContext) shutdownHybridMode() {
...
+
+	terminateChildProcess(up.hybridCommand.Process.Pid, pList)
+
+	if err := terminateProcess(up.hybridCommand.Process.Pid); err != nil {
+		oktetoLog.Debugf("error terminating process %s: %v", up.hybridCommand.Process.Pid, err)
+	}
+}
+
+func terminateChildProcess(parent int, pList []ps.Process) {
+	// assure all the child process are terminated when command is exited
```
andreafalzetti, line 941:
```suggestion
	// assure all the child processes are terminated when command is exited
```

andreafalzetti (approved):

cmd/up/up.go
```diff
@@ -924,51 +924,56 @@ func (up *upContext) shutdownHybridMode() {
...
-		select {
-		case <-to.C:
-			return false
-		case <-ticker.C:
-			existProcess := existProcess(pid)
-			if !existProcess {
-				return true
-			}
+	if err := p.Signal(syscall.SIGTERM); err != nil {
+		if errors.Is(err, os.ErrProcessDone) {
```
AdrianPedriza, line 966: Do you think we should check the error here? I see it is handled [here](https://github.com/okteto/okteto/pull/3811/files#diff-c1b929a650f782cf212e9b4b0ced5a7a298598bcb7473be505e7af1f06d6cf30R951) so maybe we can send up the error and just checking it in the caller function. WDYT?

AdrianPedriza (approved):
minor comment. LGTM

PR #3814, opened 2023-07-12
fix(up): skip deploy properly if deploy and depencencies are empty

### LINKED ISSUE #3812: okteto `up` fails in `2.17.0` for SA without permissions to create configmaps
**Describe the bug**
Running `okteto up` will fails if the serviceaccount used does not have permissions to create a configmap. This mostly affects vanilla clusters which require configuring SA manually and this permission can be missing. Or if someone is granuraly removing this permission. Error example:

```
x  Configmaps is forbidden: User "arn:aws:iam::******:role/****_***" cannot create resource "configmaps" in API group "" in the namespace "*****"
```

The user who reported this, downgraded to `2.16.5` and it's working for them. This means that something changes in `2.17.0` that requires this permission now.

Internal chat: https://okteto.slack.com/archives/C044FNE5B4Z/p1689004260078839

**To Reproduce**

1. In a dev cluster (regardless if it's vanilla or Okteto), create a new namespace
1. Create a service account
1. Create a ClusteRole without permissions to create configmaps (you can copy one of the existing ones and remove configmaps) 
1. Create a RoleBinding 
1. Create a token and assign it to the SA
1. Create a secret for the token and the SA
1. Make a backup of your `.kube/config`
1. Get the token and use it as credentials `kubectl config set-credentials bug -n debug --token=$TOKEN`
1. Specify the namespace that you have created earlier as your default: `kubectl config set-context --current --namespace=<namespace name>`
1. Now verify that you cannot create configmaps with that  `kubectl create configmap <name>`
1. Now you can use `okteto up` with the following manifest and it should pass for this build:
2. 
```yaml
dev:
  api:
    name: test
    image: busybox
    command: sh
    forward:
    - 2222:2222
    sync:
    - .:/usr/src/app
    autocreate: true  
```

**Expected behavior**
`okteto up` should work as normal

andreafalzetti:
# Proposed changes

Fixes #3812

This PR is fixing the unwanted behaviour introduced in https://github.com/okteto/okteto/pull/3614, causing the `up` cmd to fail in scenarios where the ServiceAccount used did not have permissions to create a configmap. Howerver, creating a configmap is not a thing the cmd should do, in fact the code didn't stop where it was supposed to.

## How to test

Refer to the issue #3812 for instructions on how to test it, or feel free to reach out to me to test it together because it might take a while to setup the environment.

### DIFF
--- a/cmd/deploy/deploy.go
+++ b/cmd/deploy/deploy.go
@@ -284,8 +284,7 @@ func (dc *DeployCommand) RunDeploy(ctx context.Context, deployOptions *Options) 
 	oktetoLog.Debug("found okteto manifest")
 	dc.PipelineType = deployOptions.Manifest.Type
 
-	dependenciesEmpty := deployOptions.Manifest.Dependencies == nil || len(deployOptions.Manifest.Dependencies) == 0
-	if deployOptions.Manifest.Deploy == nil && dependenciesEmpty {
+	if deployOptions.Manifest.Deploy == nil && !deployOptions.Manifest.HasDependencies() {
 		return oktetoErrors.ErrManifestFoundButNoDeployAndDependenciesCommands
 	}
 
--- a/cmd/deploy/deploy.go
+++ b/cmd/deploy/deploy.go
@@ -284,9 +284,11 @@ func (dc *DeployCommand) RunDeploy(ctx context.Context, deployOptions *Options) 
 	oktetoLog.Debug("found okteto manifest")
 	dc.PipelineType = deployOptions.Manifest.Type
 
-	if deployOptions.Manifest.Deploy == nil && deployOptions.Manifest.Dependencies == nil {
+	dependenciesEmpty := deployOptions.Manifest.Dependencies == nil || len(deployOptions.Manifest.Dependencies) == 0
+	if deployOptions.Manifest.Deploy == nil && dependenciesEmpty {
 		return oktetoErrors.ErrManifestFoundButNoDeployAndDependenciesCommands
 	}
+
 	if len(deployOptions.servicesToDeploy) > 0 && deployOptions.Manifest.Deploy != nil && deployOptions.Manifest.Deploy.ComposeSection == nil {
 		return oktetoErrors.ErrDeployCantDeploySvcsIfNotCompose
 	}
--- a/pkg/errors/errors.go
+++ b/pkg/errors/errors.go
@@ -157,9 +157,6 @@ var (
 	// ErrNoServicesinOktetoManifest raised when no services are defined in the okteto manifest
 	ErrNoServicesinOktetoManifest = fmt.Errorf("'okteto restart' is only supported when using the field 'services'")
 
-	// ErrManifestFoundButNoDeployCommands raised when a manifest is found but no deploy commands are defined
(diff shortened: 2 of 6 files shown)


cmd/deploy/deploy.go
```diff
@@ -284,9 +284,11 @@ func (dc *DeployCommand) RunDeploy(ctx context.Context, deployOptions *Options)
 	oktetoLog.Debug("found okteto manifest")
 	dc.PipelineType = deployOptions.Manifest.Type
 
-	if deployOptions.Manifest.Deploy == nil && deployOptions.Manifest.Dependencies == nil {
+	dependenciesEmpty := deployOptions.Manifest.Dependencies == nil || len(deployOptions.Manifest.Dependencies) == 0
```
andreafalzetti, line 287: when `.Dependencies` is an empty array (not `nil`), it would skip this condition instead of stopping here

pkg/errors/errors.go
```diff
@@ -157,9 +157,6 @@ var (
 	// ErrNoServicesinOktetoManifest raised when no services are defined in the okteto manifest
 	ErrNoServicesinOktetoManifest = fmt.Errorf("'okteto restart' is only supported when using the field 'services'")
 
-	// ErrManifestFoundButNoDeployCommands raised when a manifest is found but no deploy commands are defined
```
andreafalzetti, line 160: This error is no longer used, so it can be removed.

cmd/deploy/deploy.go
jLopezbarb: NIT: I'd change the variable name to `isDependenciesEmpty` or extract `deployOptions.Manifest.Dependencies` and use it directly. Looking at the variable name I don't know if it's a boolean or not

cmd/deploy/deploy.go
```diff
@@ -284,9 +284,11 @@ func (dc *DeployCommand) RunDeploy(ctx context.Context, deployOptions *Options)
...
+	if deployOptions.Manifest.Deploy == nil && dependenciesEmpty {
```
jLopezbarb, line 288: Can we add a test that goes over here? This way we won't face the error again

jLopezbarb (changes requested):
Need to add test for the deploy file, Code looks good but I didn't try how it works I'll try it later

cmd/deploy/deploy.go
teresaromero: NIT:NIT: `areDependenciesEmpty`

teresaromero: also, maybe a manifest method attached to the struct? so `manifest.HasDependencies()` ? same for deploy section

teresaromero (approved):
Reviewed with andrea in a huddle and looks good. Just minor comments.

jLopezbarb:
Tested and it's working correctly.

cmd/deploy/deploy.go
andreafalzetti: good idea @teresaromero 👍

cmd/deploy/deploy.go
jLopezbarb: That's ok. We'll revisit that `up.go` test when we start with the refactor 😉

PR #3815, opened 2023-07-13
Update info messages for hybrid `okteto up`

### LINKED ISSUE #3773: Update info messages shown to the user when `hybrid` is used
**Describe the bug**
When `hybrid` mode is used, we should remove the following messages:

- Activating your development container...
- Pulling images...
- Configuring SSH tunnel to your development container...
- Starting the file synchronization service...
- Scanning file system...
- Synchronizing your files...
and the success message:
- Files synchronized

**To Reproduce**

Run `okteto up` with hybrid mode

**Expected behavior**
Instead of those messages, we should show:

- Configuring reverse tunnel to your development environment...
and the success message:
- Reverse tunnel configured

teresaromero:
# Proposed changes

Fixes #3773 

When running `okteto up` in `hybrid` mode we were displaying spinners with messages that are not applicable for this new mode.

This PR adds conditions to show one spinner or the other depending on the mode used for run `okteto up`

Screen recording, first the changes, after how it was. Have in mind currently there is a bug where the remote terminal is not shown when running `up` so that is why at the end of the command there is no terminal.

https://github.com/okteto/okteto/assets/40767058/5f62cb22-c840-4d13-a468-b563a7d621bc

### DIFF
--- a/cmd/up/activate.go
+++ b/cmd/up/activate.go
@@ -250,9 +250,11 @@ func (up *upContext) devMode(ctx context.Context, app apps.App, create bool) err
 }
 
 func (up *upContext) createDevContainer(ctx context.Context, app apps.App, create bool) error {
-	oktetoLog.Spinner("Activating your development container...")
-	oktetoLog.StartSpinner()
-	defer oktetoLog.StopSpinner()
+	if !up.Dev.IsHybridModeEnabled() {
+		oktetoLog.Spinner("Activating your development container...")
+		oktetoLog.StartSpinner()
+		defer oktetoLog.StopSpinner()
+	}
 
 	if err := config.UpdateStateFile(up.Dev.Name, up.Dev.Namespace, config.Starting); err != nil {
 		return err
@@ -316,17 +318,18 @@ func (up *upContext) createDevContainer(ctx context.Context, app apps.App, creat
 }
 
 func (up *upContext) waitUntilDevelopmentContainerIsRunning(ctx context.Context, app apps.App) error {
-	msg := "Pulling images..."
-	if up.Dev.PersistentVolumeEnabled() {
-		msg = "Attaching persistent volume..."
-		if err := config.UpdateStateFile(up.Dev.Name, up.Dev.Namespace, config.Attaching); err != nil {
-			oktetoLog.Infof("error updating state: %s", err.Error())
+	if !up.Dev.IsHybridModeEnabled() {
+		msg := "Pulling images..."
+		if up.Dev.PersistentVolumeEnabled() {
+			msg = "Attaching persistent volume..."
+			if err := config.UpdateStateFile(up.Dev.Name, up.Dev.Namespace, config.Attaching); err != nil {
+				oktetoLog.Infof("error updating state: %s", err.Error())
+			}
 		}
+		oktetoLog.Spinner(msg)
+		oktetoLog.StartSpinner()
+		defer oktetoLog.StopSpinner()
 	}
-
-	oktetoLog.Spinner(msg)
-	oktetoLog.StartSpinner()
-	defer oktetoLog.StopSpinner()
 
 	optsWatchPod := metav1.ListOptions{
 		Watch:         true,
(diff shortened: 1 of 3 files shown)


cmd/up/syncthing.go
```diff
@@ -61,7 +61,11 @@ func (up *upContext) sync(ctx context.Context) error {
 		return err
 	}
 
-	oktetoLog.Success("Files synchronized")
+	msg := "Files synchronized"
+	if up.Dev.IsHybridModeEnabled() {
+		msg = "Reverse tunnel configured"
```
AdrianPedriza, line 66: same as above

cmd/up/forwards.go
```diff
@@ -28,7 +28,11 @@ import (
 )
 
 func (up *upContext) forwards(ctx context.Context) error {
-	oktetoLog.Spinner("Configuring SSH tunnel to your development container...")
+	msg := "Configuring SSH tunnel to your development container..."
+	if up.Dev.IsHybridModeEnabled() {
+		msg = "Configuring reverse tunnel to your development environment..."
```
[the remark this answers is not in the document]
pchico83: @omnomagonz any suggestion for this message? this is the spinner we show while activating hybrid development mode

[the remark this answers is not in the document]
omnomagonz: @pchico83 for my clarity: are you asking for a suggestion to the copy as-is or should I also consider what @AdrianPedriza is saying that this message shows even if there is no `reverse` configured? 

Also: when we say "configuring reverse tunnel to your dev env" is that saying it's configuring a reverse tunnel to the dev environment _on the developer's local laptop_?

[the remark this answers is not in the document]
pchico83: @omnomagonz right. Maybe we can start with a single spinner message to show while enabling the hybrid development mode. Configuring the reverse tunnel is just one of the stages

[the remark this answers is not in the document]
omnomagonz: I might just be ignorant to the impact on DevX here but I think the current copy is probably fine unless we have feedback direct from customers that it's causing confusion.

I think we _should_ strive for clear UX any chance we get, but unless someone has a strong opinion on what to change this to I'm supportive of leaving it as-is for now 👍

cmd/up/activate.go
```diff
@@ -250,9 +250,11 @@ func (up *upContext) devMode(ctx context.Context, app apps.App, create bool) err
 }
 
 func (up *upContext) createDevContainer(ctx context.Context, app apps.App, create bool) error {
-	oktetoLog.Spinner("Activating your development container...")
-	oktetoLog.StartSpinner()
-	defer oktetoLog.StopSpinner()
+	if !up.Dev.IsHybridModeEnabled() {
```
jLopezbarb, line 253: I think we are not showing anything here if it's hybrid which can cause some friction with the user. Should we show something to let the user know that okteto is still running?

jLopezbarb (approved):

omnomagonz (approved):

cmd/up/activate.go
teresaromero: What would you show?

jLopezbarb: I'd show the same message as we have for sync, We are activating their development container by getting the env vars, but I don't have a strong opinion on the message we should be showing. WDYT?

AdrianPedriza: In the issue description, it is stated that we do not display this message. In my opinion, even if we activate the dev container, I think it can be confusing for the user to see this message if we are in `sync` mode. On the other hand, if we don't show a spinner at this point, we could leave the user with no feedback on how the command startup process is going (plus the activation of the dev container is the most time-consuming, I think). We could show something like `preparing development environment` or something like that? WDYT?

teresaromero: 👍🏻  ok, i've updated the PR to add this message, also updated the video at the description

jLopezbarb (approved):

AdrianPedriza (approved):

PR #3822, opened 2023-07-14
fix:infer name with default okteto manifest

### LINKED ISSUE #3748: Catalog name is not recognized on `okteto up`
**Describe the bug**
When I deploy a catalog item with a `name`, and then run `okteto up`, `okteto up` doesn't recognize the catalog name and I end up with two dev environments:
Screenshot 2023-06-23 at 08 47 13

**To Reproduce**

1. Deploy the Catalog item `Oktaco Shop` in demo.okteto.dev
1. Clone https://github.com/okteto/external-resources-tf-aws locally
1. Run `okteto up`

**Expected behavior**
`okteto up` should reuse the `Oktaco Shop` dev environment instead of creating a new one.
One way to detect this is by checking the repo name and okteto manifest path against existing dev environments.

jLopezbarb:
# Proposed changes

Fixes #3748 

- Add a new check while infering the name
- If the flag name is empty and the cmap filename is the default manifest we assign the name to it

## Before and after:

| Before | After |
|--------|-------|
|  image   |  Captura de pantalla 2023-07-14 a las 10 38 34 |

## Reproduction steps
1. Deploy the Catalog item `Oktaco Shop` in `demo.okteto.dev` adding `okteto.yml` as manifestPath
2. Clone https://github.com/okteto/external-resources-tf-aws locally: `git clone https://github.com/okteto/external-resources-tf-aws locally`
3. Run `okteto up`

### DIFF
--- a/pkg/devenvironment/name.go
+++ b/pkg/devenvironment/name.go
@@ -26,6 +26,11 @@ import (
 	"k8s.io/client-go/kubernetes"
 )
 
+const (
+	// defaultOktetoFilename is the default name of the manifest file
+	defaultOktetoFilename = "okteto.yml"
+)
+
 // DeprecatedInferName infers the dev environment name from the folder received as parameter.
 // It is deprecated as it doesn't take into account deployed dev environments to get the non-sanitized name.
 // This is only being effectively used in push command, which will be deleted in the next major version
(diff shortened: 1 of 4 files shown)


AdrianPedriza:
Is there another way to test it without an AWS account? Another simpler repo maybe?

teresaromero:
Tested:
- clone https://github.com/okteto/go-getting-started
- checkout this branch and build this binary
- okteto deploy --name "my go project"
- okteto up

Output looks as expected, there is no new dev environment with repository name
Screenshot 2023-07-17 at 07 58 01

pkg/devenvironment/name.go
```diff
@@ -26,6 +26,11 @@ import (
 	"k8s.io/client-go/kubernetes"
 )
 
+const (
+	// defaultOktetoFilename is the default name of the manifest file
+	defaultOktetoFilename = "okteto.yml"
```
teresaromero, line 31: what if is `okteto.yaml` how did we determine that this is the default one?

teresaromero (approved):
solution tested and works

pkg/devenvironment/name.go
ifbyol: IIRC, we handle `okteto.yaml` and `okteto.yml` in the same way, so there is no actually a single default one. Also, this won't work when the "default" manifest is `.okteto/okteto.yml`,  `.okteto/okteto.yaml` or a docker-compose instead of an Okteto manifest.

The main problem of this issues is that when we deploy something specifying the manifest path, and that is one of the ones we automatically infer, when we execute `okteto up` without specifying the path, it doesn't get the name it should.

For example, you can still reproduce the issue with this repository: https://github.com/ifbyol/go-getting-started and the branch `ifbyol/move-default-manifest`

jLopezbarb: We decided to go with `okteto.yml` since it's the default one we use for `okteto up` command

jLopezbarb: Reopening discussion:

After speaking with @ifbyol he suggested to go with all the possible okteto manifest paths that the `okteto up` command can infer. I've gone ahead and added the support for all the possible files, so if the user set one of them in the configmap such as `.okteto/okteto.yaml` whenever he runs `okteto up` he'll get the correct dev environment. 

cc: @pchico83 since this new solution is a little more complex than the original request. Do you agree with the proposed solution?

pchico83: @jLopezbarb sounds good

andreafalzetti (approved):

Code changes look good ✅

teresaromero (approved):
👍🏻

jLopezbarb:
Fixed unit tests. The error was that the `.okteto/okteto.yaml` and `.okteto/okteto.yml` in windows was `.okteto\\okteto.yaml` and `.okteto\\okteto.yml`. The solution was to convert them to UNIX path programatically style so they match with the filepaths stored in the configmap

Issue #3823, opened 2023-07-14
Using context message is not shown on okteto build


jLopezbarb:
**Describe the bug**
When I run `okteto build` I don't get the message `Using namespace @ cluster as context`. This makes me check which context I was pointing to know where I've built the images

**To Reproduce**

1. Copy the following `Dockerfile`:
```Dockerfile
FROM alpine

RUN echo hello world
```
2. Run `okteto build`
3. See that the message is missing

**Expected behavior**
I want to have the using context message at the beggning

nitishfy:
when we're loading the context, we can set the value of _show_ to True. WDYT?

PR #3831, opened 2023-07-17
fix: show context message on okteto build

nitishfy:
Fixes: #3823

253727889-1868b2cd-ee70-4ea5-97df-b9a6ac30345d


jLopezbarb (approved):
Tested and it's working. I couldn't detect any case where we shouldn't be showing this because of showing it two times or similar

andreafalzetti (approved):
Tested and works as expected. No side effects found. Thanks @nitishfy! 🙏

PR #3832, opened 2023-07-18
fix: command unmarshal wrapping command

### LINKED ISSUE #3777: `hybrid` mode command working different locally and remote
**Describe the bug**
I am running `yarn start` as my hybrid dev command, but the behavior is very different than running the command locally.

**To Reproduce**

1. Clone https://github.com/okteto/movies-with-helm
1. Switch to the `hybrid` branch
1. Run `okteto deploy`
1. `cd api`
1. `okteto up` shows:
```
 i  Using pchico83 @ demo.okteto.dev as context
 i  'api' was already deployed. To redeploy run 'okteto deploy' or 'okteto up --deploy'
 i  Build section is not defined in your okteto manifest
 ✓  Images successfully pulled
 ✓  Files synchronized
    Context:   demo.okteto.dev
    Namespace: pchico83
    Name:      api
    Reverse:   8080 <- 8080

yarn install v1.22.18
[1/4] 🔍  Resolving packages...
success Already up-to-date.
✨  Done in 0.20s.
```
1. But running `yarn start` locally shows:
```
yarn run v1.22.18
$ nodemon server.js
[nodemon] 2.0.4
[nodemon] to restart at any time, enter `rs`
[nodemon] watching path(s): *.*
[nodemon] watching extensions: js,mjs,json
[nodemon] starting `node server.js`
```

**Expected behavior**
`okteto up` should show:
```
yarn run v1.22.18
$ nodemon server.js
[nodemon] 2.0.4
[nodemon] to restart at any time, enter `rs`
[nodemon] watching path(s): *.*
[nodemon] watching extensions: js,mjs,json
[nodemon] starting `node server.js`
```

jLopezbarb:
# Proposed changes

Fixes #3777

- Fix serializer to not wrap the command around "sh -c"
- Use `shelquote.Split` instead of splitting by whitespace. This has the advantage of having args with double quotes as a single arg instead of several ones which could cause errors

### DIFF
--- a/cmd/up/exec_unix.go
+++ b/cmd/up/exec_unix.go
@@ -8,14 +8,13 @@ import (
 	"os"
 	"os/exec"
 	"runtime"
-	"strings"
 	"syscall"
 )
 
 func (he *hybridExecutor) GetCommandToExec(ctx context.Context, cmd []string) (*exec.Cmd, error) {
 	var c *exec.Cmd
 	if runtime.GOOS != "windows" {
-		c = exec.Command("bash", "-c", strings.Join(cmd, " "))
+		c = exec.Command(cmd[0], cmd[1:]...)
 	} else {
 		binary, err := expandExecutableInCurrentDirectory(cmd[0], he.workdir)
 		if err != nil {
@@ -33,7 +32,7 @@ func (he *hybridExecutor) GetCommandToExec(ctx context.Context, cmd []string) (*
 	c.Dir = he.workdir
 
 	c.SysProcAttr = &syscall.SysProcAttr{
-		Setpgid:    true,
+		Setpgid: true,
 	}
 
 	return c, nil
--- a/pkg/model/serializer.go
+++ b/pkg/model/serializer.go
@@ -241,20 +241,22 @@ func (e Entrypoint) MarshalYAML() (interface{}, error) {
 func (c *Command) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	var multi []string
 	err := unmarshal(&multi)
-	if err != nil {
-		var single string
-		err := unmarshal(&single)
-		if err != nil {
-			return err
-		}
-		if strings.Contains(single, " ") {
-			c.Values = []string{"sh", "-c", single}
-		} else {
-			c.Values = []string{single}
-		}
-	} else {
+	if err == nil {
 		c.Values = multi
+		return nil
 	}
+
+	var single string
+	err = unmarshal(&single)
+	if err != nil {
+		return err
+	}
+
+	cmd, err := shellquote.Split(single)
+	if err != nil {
+		return err
+	}
+	c.Values = cmd
 	return nil
 }
 
--- a/pkg/model/serializer.go
+++ b/pkg/model/serializer.go
@@ -252,11 +252,11 @@ func (c *Command) UnmarshalYAML(unmarshal func(interface{}) error) error {
 		return err
 	}
 
-	cmd, err := shellquote.Split(single)
-	if err != nil {
-		return err
+	if strings.Contains(single, " ") {
+		c.Values = []string{"sh", "-c", single}
+	} else {
+		c.Values = []string{single}
 	}
-	c.Values = cmd
 	return nil
 }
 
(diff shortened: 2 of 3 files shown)


jLopezbarb:
There is an e2e test failing consistently. I'll set it ready for review when that test is fixed

pkg/model/serializer.go
```diff
@@ -241,20 +241,22 @@ func (e Entrypoint) MarshalYAML() (interface{}, error) {
 func (c *Command) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	var multi []string
 	err := unmarshal(&multi)
-	if err != nil {
```
andreafalzetti, line 244: I think we should do the change of the serializer in a different PR so that we can keep the scope as small as possible, it's a change required here but also in the [PR](https://github.com/okteto/okteto/pull/3833). for refactoring the hybrid executors. It would also be better from a git history point of view. WDYT?

jLopezbarb: I agree but when we do that the issue will be resolved. Should we change the PR to remove the wrapper in the serializer and fix the hybrid working different as a side effect?

jLopezbarb: Included it on the title to make it more visible. Let me know if that's enough

jLopezbarb:
Currently working on fixing e2e tests. It seems that `okteto up with deploy` has been affected by this change. Trying to figure out why

cmd/up/exec_unix.go
```diff
@@ -8,14 +8,13 @@ import (
...
 	"runtime"
-	"strings"
 	"syscall"
 )
 
 func (he *hybridExecutor) GetCommandToExec(ctx context.Context, cmd []string) (*exec.Cmd, error) {
 	var c *exec.Cmd
 	if runtime.GOOS != "windows" {
-		c = exec.Command("bash", "-c", strings.Join(cmd, " "))
+		c = exec.Command(cmd[0], cmd[1:]...)
```
andreafalzetti, line 17: I think we should make this change in another PR (the one about Hybrid executors), also because by removing the wrapping, we break the use case where `command: script.sh`

AdrianPedriza:
I remember some issues reported where we cannot send more than one arg to `okteto exec <svc> [args]` without quoting the args. It is related to this fix?

AdrianPedriza:
can we add more context to the description? Do you have any easy steps to repro the bug? I see `bash -c` was replaced by the solution instead `sh -c`. What am I missing?

jLopezbarb:
> I remember some issues reported where we cannot send more than one arg to `okteto exec <svc> [args]` without quoting the args. It is related to this fix?

It's related but this one doesn't solve the other

> can we add more context to the description? 

Yes, we have to do but first we need to decide the PR scope. If we want to make it simple and just remove the bash is one description, if it's fixing the bug that has been linked is another one. So I'll wait until we clarify that

> Do you have any easy steps to repro the bug? 

To reproduce follow the instructions in the issue:

1. Clone https://github.com/okteto/movies-with-helm
2. Switch to the hybrid branch
3. Run okteto deploy
4. cd api
5. Run `okteto up` 
6. Run `yarn start` in another terminal in the same PATH
7. You'll see that the output was different

> I see `bash -c` was replaced by the solution instead `sh -c`. What am I missing?

`bash` was not replaced by `sh`. We were using bash in the hybrid executor but the change that we did was unmarshal different the hybrid commands and the sync commands. So the sync commands continue to be wrapped by `sh -c`. They needed to be run in k8s if you set more than one command in one line for example: `ls && echo "hello world"`. We removed the `bash -c` from the hybrid so it reuses the same terminal as the user was using and not switch to another one for example from `powershell` to `bash`

cmd/up/exec_unix.go
andreafalzetti: We agred to keep this change and resolve the issue targetted by this PR and address other use-cases in a follow up PR

pkg/model/serializer.go
```diff
@@ -734,13 +736,38 @@ type hybridModeInfo struct {
...
+	Command     hybridCommand     `json:"command,omitempty" yaml:"command,omitempty"`
 	Reverse     []Reverse         `json:"reverse,omitempty" yaml:"reverse,omitempty"`
 	Mode        string            `json:"mode,omitempty" yaml:"mode,omitempty"`
 
 	UnsupportedFields map[string]interface{} `yaml:",inline" json:"-"`
 }
 
+type hybridCommand Command
+
+// UnmarshalYAML Implements the Unmarshaler interface of the yaml pkg.
```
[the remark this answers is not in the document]
andreafalzetti: Discussed offline and we agree to move forward with this approach to take less risks as possible for now.

andreafalzetti (approved):
Tested and works as expected!

AdrianPedriza (approved):

PR #3833, opened 2023-07-18
[Hybrid] Refactor Command Executors 

### LINKED ISSUE #3816: Refactor Hybrid executor to avoid enforcing a shell
## Description

Similarly to the deploy executor, the hybrid one should not enforce a shell but we detect the user shell and use that one (eg. Powershell, Bash, etc)

## Context

Currently we enforce `bash`

andreafalzetti:
# Proposed changes

Fixes #3816

WIP

### DIFF
--- a/cmd/exec.go
+++ b/cmd/exec.go
@@ -222,7 +222,7 @@ func executeExec(ctx context.Context, dev *model.Dev, args []string) error {
 				return err
 			}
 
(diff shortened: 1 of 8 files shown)


andreafalzetti:
Closing as we are keeping `bash` as dependency until we have time to focus on it properly.

PR #3843, opened 2023-07-21
fix: build args flags populated into buildkit

### LINKED ISSUE #3746: How is okteto build [service ...] --build-arg supposed to work?!
**Describe the bug**
I'm trying to pass in a build argument that will set a build argument when building a docker image via the okteto command line.  Ideally, I would like it to override whatever is defined in the okteto manifest.  I've tried various combinations of passing --build-arg to okteto build to no avail.  How is it supposed to be used?

**To Reproduce**

1. Here is an example okteto manifest: 
```
...
build:
  web:
    context: .
    dockerfile: Dockerfile.web
    args:
      MY_BUILD_ARG: test123
```
1. Try and build the web image and override the value of MY_BUILD_ARG: `okteto build web --build-arg "MY_BUILD_ARG=plzwork"`
1. The docker image is still built with the value of test123
1. If I remove the `args` section from okteto manifest, the docker build fails because the arg is not specified!

**Expected behavior**
Specifying `--build-arg` with the `okteto build` command should override whatever value is specified statically in the manifest.  Currently it doesn't do that, and frankly I'm not sure how it is supposed to be used because even without an `args` section defined, it still isn't passed to the docker build.

- OS: macOS
- Version 2.16.3

**Additional context**

jLopezbarb:
# Proposed changes

Fixes #3746 

- build arg flags were not being populated to buildkit instances. This PR fixes that
- If the build arg is passed as flag surrounded by double quotes, this PR cleans those double quotes and generate the correct build arg
- If there is a flag that conflicts with manifest build arg the one in the build arg flag takes precedence
- Added unit test covering the new scenarios. Previous scenarios keeps the same so we don't need to modify them

## Tests:
Copy the following Dockerfile
```Dockerfile
ARG MY_BUILD_ARG=alpine
FROM $MY_BUILD_ARG
```
Create the following `okteto.yml`
```yaml
build:
  web:
    context: .
    args:
      MY_BUILD_ARG: ubuntu
```
Run `ok build --progress plain --no-cache --build-arg "MY_BUILD_ARG=plzwork` and check that it fails because the image is not found

### DIFF
--- a/pkg/cmd/build/build.go
+++ b/pkg/cmd/build/build.go
@@ -286,7 +286,7 @@ func OptsFromBuildInfo(manifestName, svcName string, b *model.BuildInfo, o *type
 	args := []model.BuildArg{}
 	optionsBuildArgs := map[string]string{}
 	for _, arg := range o.BuildArgs {
-		splittedArg := strings.SplitN(strings.Trim(arg, "\""), "=", 2)
+		splittedArg := strings.SplitN(arg, "=", 2)
 		if len(splittedArg) == 1 {
 			optionsBuildArgs[splittedArg[0]] = ""
 			args = append(args, model.BuildArg{
--- a/pkg/cmd/build/build.go
+++ b/pkg/cmd/build/build.go
@@ -283,6 +283,23 @@ func OptsFromBuildInfo(manifestName, svcName string, b *model.BuildInfo, o *type
 		file = extractFromContextAndDockerfile(b.Context, b.Dockerfile, svcName)
 	}
 
+	args := []model.BuildArg{}
+	optionsBuildArgs := map[string]string{}
+	for _, arg := range o.BuildArgs {
+		splittedArg := strings.SplitN(strings.Trim(arg, "\""), "=", 2)
+		optionsBuildArgs[splittedArg[0]] = splittedArg[1]
+		args = append(args, model.BuildArg{
+			Name: splittedArg[0], Value: splittedArg[1],
+		})
+	}
+
+	for _, e := range b.Args {
+		if _, exists := optionsBuildArgs[e.Name]; exists {
+			continue
+		}
+		args = append(args, e)
+	}
+
 	if reg.IsOktetoRegistry(b.Image) {
 		defaultBuildArgs := map[string]string{
 			model.OktetoContextEnvVar:   okteto.Context().Name,
@@ -302,11 +319,10 @@ func OptsFromBuildInfo(manifestName, svcName string, b *model.BuildInfo, o *type
 				continue
 			}
 
-			b.Args = append(b.Args, model.BuildArg{
+			args = append(args, model.BuildArg{
 				Name: key, Value: val,
 			})
 		}
-
 	}
 
 	opts := &types.BuildOptions{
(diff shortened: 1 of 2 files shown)


pkg/cmd/build/build.go
```diff
@@ -283,6 +283,23 @@ func OptsFromBuildInfo(manifestName, svcName string, b *model.BuildInfo, o *type
 		file = extractFromContextAndDockerfile(b.Context, b.Dockerfile, svcName)
 	}
 
+	args := []model.BuildArg{}
+	optionsBuildArgs := map[string]string{}
+	for _, arg := range o.BuildArgs {
+		splittedArg := strings.SplitN(strings.Trim(arg, "\""), "=", 2)
```
AdrianPedriza, line 289: in which cases do we need to `trim` the arg? Tested with `"MY_BUILD_ARG=ubuntu" seems to catch it without that func call.

AdrianPedriza (approved):
just minor concerns. I tested and works properly

pkg/cmd/build/build.go
```diff
@@ -283,6 +283,23 @@ func OptsFromBuildInfo(manifestName, svcName string, b *model.BuildInfo, o *type
...
+		optionsBuildArgs[splittedArg[0]] = splittedArg[1]
```
andreafalzetti, line 290: we have to be careful here, if the user doesn't pass the value of the argument, it will throw a panic. To reproduce:

`okteto build --build-arg MY_BUILD_ARG`

andreafalzetti (changes requested):
What do you think of adding extra test for the "invalid" arguments such as empty string or empty value to make sure we don't panic?

jLopezbarb:
> What do you think of adding extra test for the "invalid" arguments such as empty string or empty value to make sure we don't panic?

Nice catch, I'll add those unit tests 👍

pkg/cmd/build/build.go
```diff
@@ -287,10 +287,19 @@ func OptsFromBuildInfo(manifestName, svcName string, b *model.BuildInfo, o *type
 	optionsBuildArgs := map[string]string{}
 	for _, arg := range o.BuildArgs {
 		splittedArg := strings.SplitN(strings.Trim(arg, "\""), "=", 2)
-		optionsBuildArgs[splittedArg[0]] = splittedArg[1]
-		args = append(args, model.BuildArg{
-			Name: splittedArg[0], Value: splittedArg[1],
-		})
+		if len(splittedArg) == 1 {
```
AdrianPedriza, line 290: I understand that other cases are covered in this if condition like `"MY-KEY="` but `--build-arg MY-KEY` should be managed as `"MY-KEY="`? It is not more appropriate to log that is invalid or is it something common to set an empty key?

pkg/cmd/build/build.go
AdrianPedriza: without `trim()` the `arg` and executing `ok build --progress plain --no-cache --build-arg "MY_BUILD_ARG=ubuntu"` I'm able to build the image

AdrianPedriza: using the following files:

````

ARG MY_BUILD_ARG=
FROM $MY_BUILD_ARG
````

````

build:
  web:
    context: .
    args:
      MY_BUILD_ARG: fdgdgsg
````

pkg/cmd/build/build.go
jLopezbarb: I wouldn't fail. It's valid on docker and since we are trying to emulate the same experience but running it on remote I'd  support it. WDYT?

AdrianPedriza: then go ahead! 👍

andreafalzetti (approved):
I think it works without the `trim` too, but everything else looks good and works as expected. Nice job!

teresaromero (approved):

pkg/cmd/build/build.go
jLopezbarb: I faced it using the debugger so  after checking it's working without it I'll go ahead and remove the trim. Thanks for spotting it

PR #3845, opened 2023-07-24
Add CA to kubeconfig in insecure mode

maroshii:
Fixes `okteto kubeconfig` when using okteto self signed certificates.

When using self signed certs, the backend doesn't return the CA so the kube config file generated is not valid producing the following output:

```
...
Unable to connect to the server: tls: failed to verify certificate: x509: “*.{SUBDOMAIN}” certificate is not standards compliant
```

This PR adds the context certificate to the kubeconfig is the cluster is "insecure"

# How to test

Run you dev environment with self sign certificates by making these changes:
```
UPGRADE_DEV_ARGS="--set wildcardCertificate.create=true" make upgrade-dev 
```

Check that you get a cert error when accessing https://okteto.{me}.dev.okteto.net on the browser

Build the cli from master and run:

```
bin/okteto context use okteto.{me}.dev.okteto.net
bin/okteto kubeconfig
kubectl get pods
```

You should get:

```
Unable to connect to the server: tls: failed to verify certificate: x509: “*.{SUBDOMAIN}” certificate is not standards compliant
```

Switch to this branch, re-build the cli and re-run:

```
bin/okteto context use okteto.{me}.dev.okteto.net
bin/okteto kubeconfig
kubectl get pods
```

You should now get:

```
No resources found in {me} namespace.
```

# Changes

Note that this change is in the critical path (context), so everything else should keep working in the same way it was. It's a sensitive change because prior to this change, we were reading the context from the file and overriding it in memory and discarding those values, so the json unmarshal was useless (at least in all the paths I tested). This requires a thorough review to understand that this "fix" doesn't break something else and why it was the way and still functioning correctly.

This can be tested by adding a property to one of the contexts in the context file and also adding it to the `OktetoContext` struct with the correct unmarshal. That property would ALWAYS have the golang zero value and never be included correctly in the context. This PR also fixes that.

### DIFF
--- a/cmd/context/create.go
+++ b/cmd/context/create.go
@@ -191,6 +191,9 @@ func (c *ContextCommand) UseContext(ctx context.Context, ctxOptions *ContextOpti
 			currentCtx.Namespace = currentCtx.PersonalNamespace
 		}
 
+		currentCtx := ctxStore.Contexts[ctxOptions.Context]
+		currentCtx.IsStoredAsInsecure = okteto.IsInsecureSkipTLSVerifyPolicy()
+
 		if err := c.OktetoContextWriter.Write(); err != nil {
 			return err
 		}
--- a/cmd/context/update-kubeconfig.go
+++ b/cmd/context/update-kubeconfig.go
@@ -15,6 +15,7 @@ package context
 
 import (
 	"context"
+	"encoding/base64"
 
 	"github.com/okteto/okteto/cmd/utils"
 	"github.com/okteto/okteto/pkg/config"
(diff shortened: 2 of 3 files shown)


andreafalzetti:
@maroshii can you add an "how to test" section to this PR?

maroshii:
> @maroshii can you add an "how to test" section to this PR?

@andreafalzetti done. Also added a description of the change and why it requires a through review. Alternative ways of fixing this will be received with open arms 🙌

maroshii:
@jLopezbarb This PR doesn't change any of that behavior and only changes the `kubeconfig` command. Do you get the same result from master and this branch?

You should try with a token which is what we put in installation NOTES.txt:

```
okteto context use https://my.cluster --token <token> --insecure-skip-tls-verify
```

cc @jpf-okteto

ifbyol:
> > @jLopezbarb This PR doesn't change any of that behavior and only changes the `kubeconfig` command. Do you get the same result from master and this branch?
> > You should try with a token which is what we put in installation NOTES.txt:
> > ```
> > okteto context use https://my.cluster --token <token> --insecure-skip-tls-verify
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > cc @jpf-okteto
> 
> I was able to test the scenario posted above but I wasn't able to deploy on remote (`okteto deploy --remote` on https://github.com/okteto/movies). We are also modifying the kubeconfig in that scenario and wanted to test that was working correctly. Probably we'll need to revisit that after we fix the scenario for deploying with insecure flag but for now looks good to me.

And were you able to do it before this change? That scenario should work too. We should review it

Issue #3847, opened 2023-07-25
Dependency update request: `mixpanel/mixpanel-go`


jLopezbarb:
# Dependency Update Request

We should update from https://github.com/dukex/mixpanel to the [official mixpanel library](https://github.com/mixpanel/mixpanel-go). The current dependency that we are using has not been updated for a year.

**Current Dependency: https://github.com/dukex/mixpanel**

**Official Library: https://github.com/mixpanel/mixpanel-go**

## Reason for Update:

- Ability to add a context with timeout to the request. We shouldn't be stuck on sending metrics for too long since it's a bad experience
- The dependency used doesn't have the latest updates 
- It's more secure to depend on the official library
- We'll be getting the latest updates
- The official release code has unit test which makes it more stable

## Steps to Update:

1. Remove [github.com/dukex/mixpanel](https://github.com/dukex/mixpanel) from `go.mod` and replace it with [github.com/mixpanel/mixpanel-go](https://github.com/mixpanel/mixpanel-go).
2. Update the `analytics` package to use the new module

## Testing Plan:

- Create unit test for each of the events we send 
- Check that the events are being sent after the upgrade

## Potential Risks:

- Affects all the commands tracking system

andreafalzetti:
I think it would be helpful to introduce an interface to track events and the library is one implementation. This will allow moving to a different analytics tool or library very easily in the future. @jLopezbarb wdyt?

PR #3848, opened 2023-07-25
track deploy and destroy on remote

### LINKED ISSUE #3646: Track if deploy/destroy is done remotely
## Expected Behavior
We need to track in Okteto CLI whether the user deploys or destroys using the remote feature. To do that, we need to add a new field in our track logic related to whether the remote feature is enabled.

jLopezbarb:
# Proposed changes

Fixes #3646 

## Deploy

- Created `isRemoteDeployer` in order to know if it's running on remote or locally at several points of the code
- Created `DeployAnalyticsTracker` interface in order to mock the calls in the deploy command(tested that it can not throw a panic since I don't know how to test with this module correctly)
- Added the `isRemote` property in deploy command
- Added the `isRemote` property in deploy from up command
- Added tests for the analytics.Deploy

## Destroy

- Created `DestroyAnalyticsTracker` interface in order to mock the calls in the destroy command
- Added the `isRemote` property in destroy command
- Added tests for the analytics.Destroy
- I had to cast the destroyer to `destroyAll` and `destroyRemote` to get the right analytics

## Bug fixes

- Found that we were only tracking events for destroy locally. Now tracks all the different destroy methods(all, remote, local)

## Tests done

- Debugging check that the destroy metadata is correctly injected
- Debugging check that the deploy metadata is correctly injected

Both test I have not send new data to mixpanel

### Related issues

As part of the testing I found that the mixpanel package that we were using is outdated. I created https://github.com/okteto/okteto/issues/3847 to track and address this issue

### DIFF
--- a/cmd/deploy/deploy.go
+++ b/cmd/deploy/deploy.go
@@ -90,7 +90,7 @@ type DeployCommand struct {
 	Fs                 afero.Fs
 	DivertDriver       divert.Driver
 	PipelineCMD        pipelineCMD.PipelineDeployerInterface
-	AnalyticsTracker   *analytics.AnalyticsTracker
+	AnalyticsTracker   DeployAnalyticsTracker
 
 	PipelineType       model.Archetype
 	isRemote           bool
--- a/cmd/deploy/deploy.go
+++ b/cmd/deploy/deploy.go
@@ -204,6 +204,7 @@ func Deploy(ctx context.Context) *cobra.Command {
 				Fs:                 afero.NewOsFs(),
 				PipelineCMD:        pc,
 				runningInInstaller: config.RunningInInstaller(),
+				AnalyticsTracker:   analytics.NewAnalyticsTracker(),
 			}
 			startTime := time.Now()
 
--- a/cmd/deploy/deploy.go
+++ b/cmd/deploy/deploy.go
@@ -212,12 +212,15 @@ func Deploy(ctx context.Context) *cobra.Command {
 				deployType := "custom"
 				hasDependencySection := false
 				hasBuildSection := false
+				isRunningOnRemoteDeployer := false
 				if options.Manifest != nil {
 					if options.Manifest.IsV2 &&
-						options.Manifest.Deploy != nil &&
-						options.Manifest.Deploy.ComposeSection != nil &&
-						options.Manifest.Deploy.ComposeSection.ComposesInfo != nil {
-						deployType = "compose"
+						options.Manifest.Deploy != nil {
+						isRunningOnRemoteDeployer = isRemoteDeployer(options.RunInRemote, options.Manifest.Deploy.Image)
+						if options.Manifest.Deploy.ComposeSection != nil &&
+							options.Manifest.Deploy.ComposeSection.ComposesInfo != nil {
+							deployType = "compose"
+						}
 					}
 
 					hasDependencySection = options.Manifest.IsV2 && len(options.Manifest.Dependencies) > 0
@@ -233,6 +236,7 @@ func Deploy(ctx context.Context) *cobra.Command {
 					IsPreview:              os.Getenv(model.OktetoCurrentDeployBelongsToPreview) == "true",
 					HasDependenciesSection: hasDependencySection,
 					HasBuildSection:        hasBuildSection,
+					IsRemote:               isRunningOnRemoteDeployer,
 				})
 
 				exit <- err
@@ -532,12 +536,7 @@ func getDefaultTimeout() time.Duration {
 
 func GetDeployer(ctx context.Context, manifest *model.Manifest, opts *Options, builder *buildv2.OktetoBuilder, cmapHandler configMapHandler) (deployerInterface, error) {
 
-	// isDeployRemote represents wheather the process is comming from a remote deploy
-	// if true it should get the local deployer
-	isDeployRemote := utils.LoadBoolean(constants.OktetoDeployRemote)
-
-	// remote deployment should be done when flag RunInRemote is active OR deploy.image is fulfilled
-	if !isDeployRemote && (opts.RunInRemote || opts.Manifest.Deploy.Image != "") {
+	if isRemoteDeployer(opts.RunInRemote, opts.Manifest.Deploy.Image) {
 		// run remote
 		oktetoLog.Info("Deploying remotely...")
 		return newRemoteDeployer(builder), nil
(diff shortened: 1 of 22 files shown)


cmd/destroy/destroy.go
```diff
@@ -193,7 +202,18 @@ func Destroy(ctx context.Context) *cobra.Command {
 				return err
 			}
 
-			return destroyer.destroy(ctx, options)
+			err = destroyer.destroy(ctx, options)
+
+			metadata := &analytics.DestroyMetadata{
+				Success: err == nil,
+			}
+			if _, ok := destroyer.(*localDestroyAllCommand); ok {
```
teresaromero, line 210: 🤔 could be add some method to `destroyer` so it returns the `bool` calling `destroyer.IsDestroyAll()` for example, `destroyer.IsRemote()`
 so we remove the if/elses

cmd/up/up.go
```diff
@@ -583,15 +569,32 @@ func (up *upContext) deployApp(ctx context.Context) error {
...
+
+	// tracking deploy either its been successful or not
+	c.AnalyticsTracker.TrackDeploy(analytics.DeployMetadata{
+		Success:                err == nil,
+		IsOktetoRepo:           utils.IsOktetoRepo(),
+		Duration:               time.Since(startTime),
+		PipelineType:           up.Manifest.Type,
+		DeployType:             "automatic",
+		IsPreview:              os.Getenv(model.OktetoCurrentDeployBelongsToPreview) == "true",
+		HasDependenciesSection: up.Manifest.IsV2 && len(up.Manifest.Dependencies) > 0,
```
teresaromero, line 592: I added couple commits ago a method at manifest to check this prop without conditionals here 

https://github.com/okteto/okteto/blob/master/cmd/exec.go#L104

Maybe we can add one for the `IsRemote` as it depends also from Manifest ?

scripts/ci/release.sh
teresaromero, line 1: 💭  wot?

jLopezbarb: pre-commit was complaining about empty lines😅

cmd/destroy/destroy.go
jLopezbarb: I thought it in the beginning but opted to do it this way for several reasons:
- keep interface as small as possible. If we add a new destroyer we'd have to implement an empty function if we don't want to be tracked
- [Interfaces should have no knowledge of satisfying types](https://blog.boot.dev/golang/golang-interfaces/#2-interfaces-should-have-no-knowledge-of-satisfying-types). The next time we implement a new destroyer we'd have to implement these two new methods even if both of them are false
- Follow [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) principle and just have it on one place 

I though about implementing the way you described since it was very descriptive but I think this approach is more maintainable in the long term. WDYT?

cmd/deploy/deploy.go
```diff
@@ -606,3 +590,35 @@ func (dc *DeployCommand) recreateFailedPods(ctx context.Context, name string) er
...
+				manifest.Deploy.ComposeSection.ComposesInfo != nil {
+				deployType = "compose"
+			}
+		}
+
+		hasDependencySection = manifest.HasDependencies()
+		hasBuildSection = manifest.HasBuildSection()
+	}
+
+	dc.AnalyticsTracker.TrackDeploy(analytics.DeployMetadata{
```
andreafalzetti, line 613: I am getting a panic here when I try to deploy the movies repo:

```
Executing command 'Deploy Kafka'...
Release "kafka" does not exist. Installing it now.
Error: 1 error occurred:
	* Internal error occurred: failed calling webhook "serviceaccount.webhook.okteto.com": failed to call webhook: the server could not find the requested resource
[signal SIGSEGV: segmentation violation code=0x2 addr=0x18 pc=0x10682348c]

goroutine 55 [running]:
github.com/okteto/okteto/cmd/deploy.(*DeployCommand).trackDeploy(0x1400184e780, 0x1400184e840?, 0x1?, {0x14001724ea0?, 0x0?, 0x10949dcc0?}, {0x107a91280, 0x140006307a0?})
	github.com/okteto/okteto/cmd/deploy/deploy.go:613 +0x27c
github.com/okteto/okteto/cmd/deploy.Deploy.func1.1()
	github.com/okteto/okteto/cmd/deploy/deploy.go:217 +0x8c
created by github.com/okteto/okteto/cmd/deploy.Deploy.func1
	github.com/okteto/okteto/cmd/deploy/deploy.go:214 +0x738
```

cmd/destroy/destroy.go
teresaromero: makes sense :)

scripts/ci/release.sh
teresaromero: that is something to keep in mind when we implement the linting

cmd/deploy/deploy.go
jLopezbarb: Fixed. Somehow I removed on a previous commit in the declaration 🤷 . Thanks for spotting it

pkg/analytics/analyticstracker.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+func NewAnalyticsTracker() *AnalyticsTracker {
+	return &AnalyticsTracker{
+		TrackFn: track,
+	}
+}
+
+// DeployMetadata contains the metadata of a deploy event
+type DeployMetadata struct {
```
teresaromero, line 22: can we separate the deploy/destroy to a `track_deploy` `track_destroy` file and leave here just the Analytics Tracker? i can use this for the Up events....

jLopezbarb: Addressed. Let me know what you think about the change

scripts/ci/release.sh
teresaromero, line 1: :) again this changes

cmd/deploy/deploy.go
```diff
@@ -606,3 +591,35 @@ func (dc *DeployCommand) recreateFailedPods(ctx context.Context, name string) er
...
+		Success:                err == nil,
+		IsOktetoRepo:           utils.IsOktetoRepo(),
+		Duration:               time.Since(startTime),
+		PipelineType:           dc.PipelineType,
+		DeployType:             deployType,
+		IsPreview:              os.Getenv(model.OktetoCurrentDeployBelongsToPreview) == "true",
+		HasDependenciesSection: hasDependencySection,
```
teresaromero, line 621: we can call directly to the `manifest.HasDependencies()` here and remove the var being declared, also for the BuildSection one. Could we make a helper method from manifest, maybe `IsCompose` to get the deployType same way? then we can remove all the nested ifs from here and delegate that to the manifest to give us that information? WDYT?

pkg/analytics/tracker.go
```diff
@@ -0,0 +1,11 @@
+package analytics
+
+type AnalyticsTracker struct {
+	TrackFn func(event string, success bool, props map[string]interface{})
```
teresaromero, line 4: Can this be private?
```suggestion
	trackFn func(event string, success bool, props map[string]interface{})
```

cmd/deploy/deploy.go
jLopezbarb: I don't think so, what if it fails before manifest is already loaded it will fail because manifest will be nil

teresaromero (approved):
LGTM 👍🏻

Issue #3859, opened 2023-07-29
"okteto preview endpoints -o md" includes system messages


rberrelleza:
**Describe the bug**

``` 
[ramiro-dev-root|default] rberrelleza@Ramiros-MBP getting-started % okteto preview endpoints pr-getting-started-4 -omd
 !  $AWS_ACCESS_KEY_ID secret is being overridden by a local environment variable by the same name.
 !  $AWS_SECRET_ACCESS_KEY secret is being overridden by a local environment variable by the same name.
 i  Using pr-getting-started-4 @ ramiro.okteto.dev as context
Available endpoints for preview [pr-getting-started-4](https://ramiro.okteto.dev/#/previews/pr-getting-started-4):

 - [https://todo-list-pr-getting-started-4.ramiro.okteto.dev](https://todo-list-pr-getting-started-4.ramiro.okteto.dev)
```

The fact that the output includes system messages makes it hard to use during automation scenarios (e.g. include the endpoints on a PR or in other Systems). 

Could we have a way to only output the endpoints in markdown? Similar to what '-o json' does?

**To Reproduce**

1. Click on '....'
1. Scroll down to '....'
1. See error

**Expected behavior**

**Desktop (please complete the following information):**

- OS: [e.g. macOS]

**Additional context**
Add any other context about the problem here (e.g. okteto.yml, k8s manifests, logs)

Issue #3805, opened 2023-07-10
Extend `--output` flag with type default choice `table` and `plaintext`


andreafalzetti:
At the moment some commands support multiple output formats such as `json , `yaml`, `md`, however, if you don't pass the flag, it will output in plaintext and sometimes using a table format.

I propose `plaintext` and `table` could should be valid values for the output flag, and they should be the default one for such commands. Similarly to [aws-cli](https://docs.aws.amazon.com/cli/latest/userguide/cli-usage-output-format.html).

By adding the new value and make it the default, we'll keep the current behaviour so it'd not be a breaking change

andreafalzetti:
not a priority but I thought it was a good time to create this considering we're [working on this](https://github.com/okteto/okteto/issues/3772).

andreafalzetti:
Moved to https://okteto.atlassian.net/browse/LAKE-92

PR #3861, opened 2023-07-31
fix: preview output json

### LINKED ISSUE #3860: "okteto preview endpoints $NAME -ojson" produces invalid json
**Describe the bug**

**To Reproduce**
1. Set a secret in Okteto
2. Create a local environment variable with the same name
3. execute "okteto preview endpoints $NAME -ojson

Output will be something like this:
```
!  $AWS_SECRET_ACCESS_KEY secret is being overridden by a local environment variable by the same name.
 !  $AWS_ACCESS_KEY_ID secret is being overridden by a local environment variable by the same name.
[
  {
    "url": "https://todo-list-pr-getting-started-4.ramiro.okteto.dev",
    "private": false
  }
]
```

**Expected behavior**
the output should be valid JSON 

- OS: macos
- Version 2.18.0

jLopezbarb:
# Proposed changes

Fixes #3860 

- Save the logs that should appear while initializing the context
- Print them as information messages on the log file if the ouptut is json

## Before and after:

| Before | After |
|--------|-------|
| Captura de pantalla 2023-07-31 a las 13 53 24  | Captura de pantalla 2023-07-31 a las 13 54 19 |

## Reproduction steps
1. Set Secret `KEY` on okteto cloud settings
2. Create a local environment variable `$KEY` with different value
3. Run `okteto preview deploy issue-USERNAME --scope personal --branch master --repository https://github.com/okteto/movies`
4. execute "okteto preview endpoints issue-USERNAME -ojson

### DIFF
--- a/cmd/preview/endpoints.go
+++ b/cmd/preview/endpoints.go
@@ -14,9 +14,11 @@
 package preview
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
+	"os"
 	"sort"
 	"strings"
 
@@ -46,11 +48,19 @@ func Endpoints(ctx context.Context) *cobra.Command {
 				return err
 			}
 
+			jsonContextBuffer := bytes.NewBuffer([]byte{})
+			if output == "json" {
+				oktetoLog.SetOutput(jsonContextBuffer)
+			}
+
 			if err := contextCMD.NewContextCommand().Run(ctx, &contextCMD.ContextOptions{}); err != nil {
 				return err
 			}
 			if output != "json" {
 				oktetoLog.Information("Using %s @ %s as context", previewName, okteto.RemoveSchema(okteto.Context().Name))
+			} else {
+				oktetoLog.Info(jsonContextBuffer.String())
+				oktetoLog.SetOutput(os.Stdout)
 			}
 
 			if !okteto.IsOkteto() {


cmd/preview/endpoints.go
```diff
@@ -46,11 +48,19 @@ func Endpoints(ctx context.Context) *cobra.Command {
 				return err
 			}
 
+			jsonContextBuffer := bytes.NewBuffer([]byte{})
+			if output == "json" {
```
andreafalzetti, line 52: I know it's not part of the scope of the bug reported but I was wondering:

We support two output formats: `json` and `md` (besides the default one which would be something like: `plaintext` or `text`)

Would it make sense to slightly adjust this change to also fix `md`? I imagine the `md` output is useful to be injected in a Markdown document, but at the moment it would also output and inject the logs, making the feature kind of broken.

My proposal would be change this condition to be `if output != "text" {..}` and add `text` as valid output format.

I actually think we should add `plaintext` or `text` to all output flags, I've created an [issue about it](https://github.com/okteto/okteto/issues/3805) some time ago

andreafalzetti: I don't want to block the PR, the change works and LGTM. If we can squeeze the change good, or I can create a follow-up issue (I didn't because I think the change could be very tiny, let me know!)

andreafalzetti (approved):

cmd/preview/endpoints.go
AdrianPedriza: IMO I think it is safe and has the sense to address your suggestion in this PR 👍

cmd/preview/endpoints.go
AdrianPedriza, line 1: should we move output validation at the beginning of the command instead run logic related to output before validating it?

cmd/preview/endpoints.go
jLopezbarb: It'd be great to extend them but I'll address in another PR. In fact there is a [current issue](https://github.com/okteto/okteto/issues/3859) for the md

cmd/preview/endpoints.go
jLopezbarb: I wouldn't do it as part of this PR to have an specific context when reviewing the code later on. 
I do think that's something we need to do in the future not just on this command but all the commands that shows tables or other outputs that are not the typical ones

cmd/preview/endpoints.go
andreafalzetti: sounds good to me, already approved, so feel free to ship it 🚢

PR #3874, opened 2023-08-03
fix: compute build envs in local

AdrianPedriza:
# Proposed changes
- image tag calculation on some occasions is different between local vs remote. This change is to calculate image tags locally and send them to remote


AdrianPedriza:
> I think there was a test checking the output for the Dockerfile. Can we add a test to check that the build env vars are injected properly?

OK, I will take a look 👍

jLopezbarb (approved):

PR #3875, opened 2023-08-03
Up event refactor

### LINKED ISSUE #3644: Track duration of every relevant step of `okteto up`
We want to track the duration of every relevant step of the up command to understand where to invest more time to speed up the execution of the `okteto up` command.

We are already tracking the time it takes the initial sync, but we can track more things like time to configure the okteto context, fire another event if there is a `build`, fire another event if there is a `deploy`, time to create the dev container, time to synchronize the the context, time to scan local folders, ... and the total execution time.

teresaromero:
# Proposed changes

Partially Fixes #3644 

This PR changes the way we send the `Up` event to our mixpanel platform. Currently we triggered an event during the run of the `up` command, instead of wrapping up the metadata and sending all when the command ends.

This PR tracks metadata and sends the event at the end, either its been successful or not. 
Note that `success` property was always sent `false` since `1.13.11` with this change the meaning of `success` is fixed, so only when there has been an error it will be `false`.

The events that where tracking conditions on the running command are now props, so this can be measured within the event triggered.

Changes:

- Remove all `up` event functions and const to its own file `up.go` under `analytics` pkg
- `TrackUp` is triggered when the command stops, either because an error or because its success. Also a panic in the code will trigger the event to be sent.
- Created `UpMetadata` methods to keep adding the information into the struct

Events that are now props inside the `Up` event
  - `Up Error` event is now prop `failActivate`
  - `Up Duration Time` event is now prop `activateDurationSeconds` and its tracked in seconds, currently it was being sent as a timestamp of duration which was not possible to interpret at mixpanel
  - `Initial Sync Duration Time `event is now prop `initialSyncDurationSeconds`, also tracked in seconds, same as previous it was sent as timestamp
  - `Sync Error` event is now prop `errSync`
  - `Sync Reset Database` event is now prop `errResetDatabase`
  - `Reconnect` event is now prop `isReconnect` and `reconnectCause` with a string of the cause

### Tests

After changes, compiled the binary and running the command up. With the version tag of the build, check at mixpanel the event and props where getting correctly.
- Force a panic , event is sent success false
- Force exit with error, event is sent success false
- Exit normally, event is send success true

### Mixpanel update

Take in mind that this change is reflected with newer versions of CLI when this is released. Any dashboard using the Up event and others should be updated

### DIFF
--- a/cmd/up/syncthing.go
+++ b/cmd/up/syncthing.go
@@ -19,7 +19,6 @@ import (
 	"time"
 
 	"github.com/okteto/okteto/cmd/utils"
-	"github.com/okteto/okteto/pkg/analytics"
 	"github.com/okteto/okteto/pkg/config"
 	oktetoErrors "github.com/okteto/okteto/pkg/errors"
 	oktetoLog "github.com/okteto/okteto/pkg/log"
@@ -68,7 +67,7 @@ func (up *upContext) sync(ctx context.Context) error {
 	oktetoLog.Success(msg)
 
 	elapsed := time.Since(start)
-	analytics.TrackDurationInitialSync(elapsed)
+	up.analyticsTracker.TrackDurationInitialSync(elapsed)
 	maxDuration := time.Duration(1) * time.Minute
 	if elapsed > maxDuration {
 		minutes := elapsed / time.Minute
@@ -177,7 +176,7 @@ func (up *upContext) synchronizeFiles(ctx context.Context) error {
 	}()
 
 	if err := up.Sy.WaitForCompletion(ctx, reporter); err != nil {
-		analytics.TrackSyncError()
+		up.analyticsTracker.TrackSyncError()
 		switch err {
 		case oktetoErrors.ErrLostSyncthing:
 			return err
--- a/cmd/up/syncthing.go
+++ b/cmd/up/syncthing.go
@@ -67,7 +67,7 @@ func (up *upContext) sync(ctx context.Context) error {
 	oktetoLog.Success(msg)
 
 	elapsed := time.Since(start)
-	up.analyticsMeta.AddInitialSyncDuration(elapsed)
+	up.analyticsMeta.InitialSyncDuration(elapsed)
 	maxDuration := time.Duration(1) * time.Minute
 	if elapsed > maxDuration {
 		minutes := elapsed / time.Minute
@@ -176,14 +176,14 @@ func (up *upContext) synchronizeFiles(ctx context.Context) error {
 	}()
 
 	if err := up.Sy.WaitForCompletion(ctx, reporter); err != nil {
-		up.analyticsMeta.AddErrSync()
+		up.analyticsMeta.ErrSync()
 		switch err {
 		case oktetoErrors.ErrLostSyncthing:
 			return err
 		case oktetoErrors.ErrInsufficientSpace:
 			return up.getInsufficientSpaceError(err)
 		case oktetoErrors.ErrNeedsResetSyncError:
-			up.analyticsMeta.AddErrResetDatabase()
+			up.analyticsMeta.ErrResetDatabase()
 			return oktetoErrors.UserError{
 				E:    fmt.Errorf("the synchronization service state is inconsistent"),
 				Hint: `Try running 'okteto up --reset' to reset the synchronization service`,
--- a/cmd/up/syncthing.go
+++ b/cmd/up/syncthing.go
@@ -67,7 +67,7 @@ func (up *upContext) sync(ctx context.Context) error {
 	oktetoLog.Success(msg)
 
 	elapsed := time.Since(start)
-	up.analyticsTracker.TrackDurationInitialSync(elapsed)
+	up.analyticsMeta.AddInitialSyncDuration(elapsed)
 	maxDuration := time.Duration(1) * time.Minute
 	if elapsed > maxDuration {
 		minutes := elapsed / time.Minute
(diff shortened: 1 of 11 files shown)


pkg/analytics/up.go
```diff
@@ -0,0 +1,120 @@
...
+
+func (u *UpMetadata) AddManifestProps(m *model.Manifest) {
+	u.isV2 = m.IsV2
+	u.manifestType = m.Type
+	u.hasDependenciesSection = m.HasDependenciesSection()
+	u.hasBuildSection = m.HasBuildSection()
+	u.hasDeploySection = m.HasDeploySection()
+}
+
+func (u *UpMetadata) AddDevProps(d *model.Dev) {
```
jLopezbarb, line 69: Can we agree on the convention for all these names? I think it's odd to have add/Set/CommandSuccess I think all of them should follow the same structure since it's modifying the struct `UpMetadata` with the data from the function. 

Also if we add props to one I'd do it for every function, but I think it's better to remove the suffix

pkg/analytics/up.go
```diff
@@ -0,0 +1,120 @@
...
+	"github.com/okteto/okteto/pkg/model"
+)
+
+const (
+	// Event that tracks when a user activates a development container
+	upEvent = "Up"
+)
+
+// UpMetadata defines the properties an up can have
+type UpMetadata struct {
```
jLopezbarb, line 15: NIT:
```suggestion
type UpMetricsMetadata struct {
```

jLopezbarb (commented):
Code looks good!! I've added several comments (nitpicking). It's ready to merge as soon as we have tests for `pkg/analytics/up.go`

cmd/up/syncthing.go
```diff
@@ -177,13 +176,14 @@ func (up *upContext) synchronizeFiles(ctx context.Context) error {
...
 	if err := up.Sy.WaitForCompletion(ctx, reporter); err != nil {
-		analytics.TrackSyncError()
+		up.analyticsMeta.AddErrSync()
 		switch err {
 		case oktetoErrors.ErrLostSyncthing:
 			return err
 		case oktetoErrors.ErrInsufficientSpace:
 			return up.getInsufficientSpaceError(err)
 		case oktetoErrors.ErrNeedsResetSyncError:
+			up.analyticsMeta.AddErrResetDatabase()
```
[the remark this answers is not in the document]
teresaromero: i had the same feeling, this is something we already where tracking with the ResetDatabase event, which just sent boolean when happened, in order to track it into the event i add the prop as i understood it caused an error :/ 

in terms of error in the cmd, we are just tracking if is successful or not, with just some kind of errors. I think it would be interesting to track why if failed in every scenario but we need to refactor the code more, wdyt?

pkg/analytics/up.go
```diff
@@ -0,0 +1,120 @@
...
+
+const (
+	// ReconnectCauseDefault is the default cause for a reconnection
+	ReconnectCauseDefault = "unrecognised"
+
+	// ReconnectCauseDevPodRecreated is cause when pods UID change between retrys
+	ReconnectCauseDevPodRecreated = "dev-pod-recreated"
+)
+
+func (u *UpMetadata) AddReconnect(cause string) {
```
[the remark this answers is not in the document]
teresaromero: yup, i will split into 2 and make the strings private

cmd/up/syncthing.go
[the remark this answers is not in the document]
jLopezbarb: I'd add it. This way we'll have all the sync errors that we know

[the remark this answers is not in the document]
teresaromero: i added them to the follow up PR with new props as this was refactor of existent ones

https://github.com/okteto/okteto/pull/3876/commits/9489d017bfa4627f4205a2acf95a6ac7a0a7db93

cmd/up/up.go
```diff
@@ -372,21 +381,20 @@ func Up() *cobra.Command {
...
 					oktetoLog.Infof("CommandError: %v", err)
+					return err
 				case oktetoErrors.UserError:
 					return err
 				}
-
 			}
 
-			return err
+			up.analyticsMeta.CommandSuccess()
```
AdrianPedriza, line 396: if `up` fails that metric is sent? or do we assume that if the `success` metric does not exist it has failed?

pkg/analytics/up.go
```diff
@@ -0,0 +1,150 @@
...
+	u.initialSyncDuration = duration
+}
+
+const (
+	// reconnectCauseDefault is the default cause for a reconnection
+	reconnectCauseDefault = "unrecognised"
+
+	// reconnectCauseDevPodRecreated is cause when pods UID change between retrys
+	reconnectCauseDevPodRecreated = "dev-pod-recreated"
+)
```
AdrianPedriza, line 118: could we move it at the top along with the others consts?

pkg/analytics/up.go
```diff
@@ -0,0 +1,150 @@
...
+
+// ErrSync sets to true the property errSync
+func (u *UpMetricsMetadata) ErrSync() {
+	u.errSync = true
+}
+
+// ErrResetDatabase sets to true the property errResetDatabase
+func (u *UpMetricsMetadata) ErrResetDatabase() {
+	u.errResetDatabase = true
+}
```
AdrianPedriza, line 140:
```suggestion
func (u *UpMetricsMetadata) ErrResetSyncDatabase() {
	u.errResetSyncDatabase = true
}
```
WDYT about add more context to the func, vars and metric id about the "reset database" event related to Syncthing?

AdrianPedriza (approved):
code looks good. Just minor comments

cmd/up/up.go
teresaromero: i've wrapped the err check into line 384, `CommandSuccess` sets the property `success` to true, if some error occur this is not set so the cmd is reportted as success false.

pkg/analytics/up.go
teresaromero: yes i did that at the followup PR https://github.com/okteto/okteto/pull/3876/files#diff-2a9738d7a31893670e28d1ce2d082069a5af8a0f5de048a12b02e175bee04becR155 🙏🏻

PR #3876, opened 2023-08-03
Add new props to up event

### LINKED ISSUE #3644: Track duration of every relevant step of `okteto up`

teresaromero:
# Proposed changes

Fixes #3644 

After refactoring the Up event logic, this PR adds the new props required to track duration and other events inside the Up event.

- "hasRunDeploy": when the up command triggers a deploy app inside
- "oktetoCtxConfigDurationSeconds": duration in seconds of configuring okteto context
- "devContainerCreationDurationSeconds":  duration in seconds to create a dev container
- "contextSyncDurationSeconds": duration in seconds to sync context
- "localFoldersScanDurationSeconds": duration in seconds to scan local folders
- "execDurationSeconds": duration in seconds of command execution

this PR does not include the build as requested in the issue. Build is done inside a deploy app process, to trigger this in the up context will require further refactor so i propose to revisit this when its possible.

### Testing

With binary build and using the up command, checking at mixpanel the durations an events are being sent


AdrianPedriza:
I've deleted a question about tracking the build, I didn't see your comment in the description. Should we create a folloup issue to track it in the future?

cmd/up/activate.go
```diff
@@ -243,9 +246,11 @@ func (up *upContext) shouldRetry(ctx context.Context, err error) bool {
 }
 
 func (up *upContext) devMode(ctx context.Context, app apps.App, create bool) error {
```
jLopezbarb, line 248: I'd also time the wait in `DevContainerCreation`. I don't know if it should be another prop because it depends on the connection with the cluster

teresaromero:
> I've deleted a question about tracking the build, I didn't see your comment in the description. Should we create a folloup issue to track it in the future?

I think this should be taken into account when planning the refactor, to be able to track different processes within a same command ... not sure if an isolated issue or some doc?

cmd/up/activate.go
teresaromero: this is trakced with `devContainerCreationDurationSeconds ` maybe i added the count in a missplace?

jLopezbarb: Oh, I see it now. Sorry for the noise

AdrianPedriza (approved):

AdrianPedriza:
> > I've deleted a question about tracking the build, I didn't see your comment in the description. Should we create a folloup issue to track it in the future?
> 
> I think this should be taken into account when planning the refactor, to be able to track different processes within a same command ... not sure if an isolated issue or some doc?

I can added in our quality initiatives 👍

PR #3880, opened 2023-08-03
fix panic when displayer has no cancel function

### LINKED ISSUE #3854: Panic when `bash` is not installed
**Describe the bug**
Right now `bash` is a hard dependency in the CLI and if it is not install it results in a **panic**. This happen in any platform, Windows, Linux and Mac.

**Expected behavior**
Until we decouple `bash`, the CLI should gracefully exit and inform users that `bash` is a dependency and should be installed.

AdrianPedriza:
# Proposed changes

- declare cancel function before start command to avoid panics 2819e5a486cf125f6ca6ad0441dc65a701ecd170
- add message about required shells bd10752562fcb85ec3fd8c3654c302428aae89a9

Fixes #3854 

## How looks like

`okteto deploy --no-bash`

| Before | After |
|----------|----------|
| Captura de Pantalla 2023-08-07 a las 13 35 04| Captura de Pantalla 2023-08-07 a las 13 34 12|

`okteto deploy`

| Before | After |
|----------|----------|
|Captura de Pantalla 2023-08-07 a las 13 45 30| Captura de Pantalla 2023-08-07 a las 13 44 12|

to be able to repro the last case I've tested it over Powershell so the output format is not so friendly but the case is covered by the message ` "bash": executable file not found in <ugly format (ignore it)> "bash" is a required dependency for executing the command`

### DIFF
--- a/cmd/utils/displayer/json.go
+++ b/cmd/utils/displayer/json.go
@@ -43,14 +43,17 @@ func newJSONDisplayer(stdout, stderr io.Reader) *jsonDisplayer {
 		stderrScanner = bufio.NewScanner(stderr)
 	}
 
+	commandContext, cancel := context.WithCancel(context.Background())
+
 	return &jsonDisplayer{
-		stdoutScanner: stdoutScanner,
-		stderrScanner: stderrScanner,
+		stdoutScanner:  stdoutScanner,
+		stderrScanner:  stderrScanner,
+		commandContext: commandContext,
+		cancel:         cancel,
 	}
 }
 
 func (d *jsonDisplayer) Display(_ string) {
-	d.commandContext, d.cancel = context.WithCancel(context.Background())
 	var wg sync.WaitGroup
 	wgDelta := 0
 	if d.stdoutScanner != nil {
--- a/cmd/utils/displayer/plain.go
+++ b/cmd/utils/displayer/plain.go
@@ -43,14 +43,17 @@ func newPlainDisplayer(stdout, stderr io.Reader) *plainDisplayer {
 		stderrScanner = bufio.NewScanner(stderr)
 	}
 
+	commandContext, cancel := context.WithCancel(context.Background())
+
 	return &plainDisplayer{
-		stdoutScanner: stdoutScanner,
-		stderrScanner: stderrScanner,
+		stdoutScanner:  stdoutScanner,
+		stderrScanner:  stderrScanner,
+		commandContext: commandContext,
+		cancel:         cancel,
 	}
 }
 
 func (d *plainDisplayer) Display(_ string) {
-	d.commandContext, d.cancel = context.WithCancel(context.Background())
 	var wg sync.WaitGroup
 	wgDelta := 0
 	if d.stdoutScanner != nil {
--- a/cmd/utils/displayer/tty.go
+++ b/cmd/utils/displayer/tty.go
@@ -47,17 +47,20 @@ func newTTYDisplayer(stdout, stderr io.Reader) *ttyDisplayer {
 		stderrScanner = bufio.NewScanner(stderr)
 	}
 
+	commandContext, cancel := context.WithCancel(context.Background())
+
 	return &ttyDisplayer{
-		stdoutScanner: stdoutScanner,
-		stderrScanner: stderrScanner,
-		screenbuf:     screenbuf.New(os.Stdout),
+		stdoutScanner:  stdoutScanner,
+		stderrScanner:  stderrScanner,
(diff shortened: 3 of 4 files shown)


teresaromero:
@AdrianPedriza have you taken a look to the comment i linked to the issue? https://github.com/okteto/okteto/commit/6dcb4a336846735fdcd292bd7183c5f303be9154 

when using the flag (no using bash) we did have an other panic

AdrianPedriza:
> @AdrianPedriza have you taken a look to the comment i linked to the issue? [6dcb4a3](https://github.com/okteto/okteto/commit/6dcb4a336846735fdcd292bd7183c5f303be9154)
> 
> when using the flag (no using bash) we did have an other panic

which command gives you a panic in that block? maybe the fix solve it. I've tested your suggestion using `env | greo IMAGE` command but seems it didn't parse the command properly. Am I doing something wrong?

teresaromero:
> > @AdrianPedriza have you taken a look to the comment i linked to the issue? [6dcb4a3](https://github.com/okteto/okteto/commit/6dcb4a336846735fdcd292bd7183c5f303be9154)
> > when using the flag (no using bash) we did have an other panic
> 
> which command gives you a panic in that block? maybe the fix solve it. I've tested your suggestion using `env | greo IMAGE` command but seems it didn't parse the command properly. Am I doing something wrong?

If i recall correctly this happened at Windows when running `deploy` command with `no-bash` flag. As an alternative to "not using bash" this flag was also sending a panic because of the command composition. Could be happening also in other platforms

AdrianPedriza:
> > > @AdrianPedriza have you taken a look to the comment i linked to the issue? [6dcb4a3](https://github.com/okteto/okteto/commit/6dcb4a336846735fdcd292bd7183c5f303be9154)
> > > when using the flag (no using bash) we did have an other panic
> > 
> > 
> > which command gives you a panic in that block? maybe the fix solve it. I've tested your suggestion using `env | greo IMAGE` command but seems it didn't parse the command properly. Am I doing something wrong?
> 
> If i recall correctly this happened at Windows when running `deploy` command with `no-bash` flag. As an alternative to "not using bash" this flag was also sending a panic because of the command composition. Could be happening also in other platforms

as we talk before, this concern is addressed by the proposed changes (see description)

cmd/utils/displayer/json.go
```diff
@@ -43,14 +43,17 @@ func newJSONDisplayer(stdout, stderr io.Reader) *jsonDisplayer {
...
 	}
 
+	commandContext, cancel := context.WithCancel(context.Background())
+
 	return &jsonDisplayer{
-		stdoutScanner: stdoutScanner,
-		stderrScanner: stderrScanner,
+		stdoutScanner:  stdoutScanner,
+		stderrScanner:  stderrScanner,
+		commandContext: commandContext,
```
jLopezbarb, line 51: Adding the context inside a struct is something that shouldn't be done in Golang.

Contexts have a lifecycle, and they should be created and canceled appropriately. When you store a context inside a struct, you might lose track of when the context should be canceled, leading to potential resource leaks or incorrect behavior.

[More information here](https://go.dev/blog/context-and-structs)

teresaromero: why do we need the context and cancel here?

AdrianPedriza: it was the way I found to reuse the context and cancel function in the needed places. I'm working on a refactor

AdrianPedriza: refactor to avoid store context in struct 687ca34461a0c5c27ec813884fb0c636d7baa01a

AdrianPedriza: do not check. There are bugs with the current proposal

AdrianPedriza: ready to review. As discussed offline, removes `ctx` and `cancel` func from struct requires a refactor in how executors and displayed are designed.

jLopezbarb (approved):

cmd/utils/displayer/plain.go
```diff
@@ -95,7 +91,8 @@ func (d *plainDisplayer) Display(_ string) {
 }
 
 // CleanUp stops displaying
-func (d *plainDisplayer) CleanUp(_ error) {
-	d.cancel()
-	<-d.commandContext.Done()
+func (d *plainDisplayer) CleanUp(ctx context.Context, _ error) {
+	ctx, cancel := context.WithCancel(ctx)
```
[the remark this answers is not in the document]
jLopezbarb: Not the latest commit. Ignore it

teresaromero (approved):

PR #3884, opened 2023-08-04
Use dynamic kube token for k8s requests

andreafalzetti:
# Proposed changes

Fixes https://github.com/okteto/app/issues/6906

WIP

### DIFF
--- a/cmd/context/create.go
+++ b/cmd/context/create.go
@@ -299,7 +299,7 @@ func getLoggedUserContext(ctx context.Context, c *ContextCommand, ctxOptions *Co
 	if err != nil {
 		return nil, err
 	}
-	kubetokenRes, err := kubetokenClient.RequestKubeToken()
+	kubetokenRes, err := kubetokenClient.GetKubeToken()
 	if err != nil {
 		return nil, err
 	}
--- a/cmd/context/create.go
+++ b/cmd/context/create.go
@@ -295,6 +295,16 @@ func getLoggedUserContext(ctx context.Context, c *ContextCommand, ctxOptions *Co
 		return nil, err
 	}
 
+	kubetokenClient, err := okteto.NewKubeTokenClient(ctxOptions.Context, ctxOptions.Token, ctxOptions.Namespace)
+	if err != nil {
+		return nil, err
+	}
+	kubetokenRes, err := kubetokenClient.RequestKubeToken()
+	if err != nil {
+		return nil, err
+	}
+	userContext.Credentials.Token = kubetokenRes.Status.Token
+
 	return userContext, nil
 }
 
--- a/integration/kubetoken/kubetoken_test.go
+++ b/integration/kubetoken/kubetoken_test.go
@@ -1,0 +1,54 @@
+//go:build integration
+// +build integration
+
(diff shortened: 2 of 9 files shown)


cmd/context/create.go
```diff
@@ -295,6 +295,16 @@ func getLoggedUserContext(ctx context.Context, c *ContextCommand, ctxOptions *Co
 		return nil, err
 	}
 
+	kubetokenClient, err := okteto.NewKubeTokenClient(ctxOptions.Context, ctxOptions.Token, ctxOptions.Namespace)
```
jLopezbarb, line 298: NIT: I think this has to be injected from `ContextCommand` otherwise we'll be calling a real environment on tests

pkg/okteto/kubetoken_client.go
```diff
@@ -46,24 +52,40 @@ func NewKubeTokenClient(contextName, token, namespace string) (*KubeTokenClient,
 	}, nil
 }
 
-func (c *KubeTokenClient) GetKubeToken() (string, error) {
+func (c *KubeTokenClient) RequestKubeToken() (*KubeTokenResponse, error) {
 	resp, err := c.httpClient.Get(c.url)
 	if err != nil {
-		return "", fmt.Errorf("failed GET request: %w", err)
+		return &KubeTokenResponse{}, fmt.Errorf("failed GET request: %w", err)
```
jLopezbarb, line 58: The idiomatic way of returning an error would be to return a nil struct or removing the pointer at the func signature:
`func (c *KubeTokenClient) RequestKubeToken() (KubeTokenResponse, error)`

integration/kubetoken/kubetoken_test.go
```diff
@@ -0,0 +1,54 @@
...
+	"github.com/okteto/okteto/pkg/okteto"
+	"github.com/stretchr/testify/require"
+	"testing"
+)
+
+// TestKubetokenHasExpiration test that kubernetes returns a dynamic token with a set expiration
+func TestKubetokenHasExpiration(t *testing.T) {
+	t.Parallel()
+
+	oktetoPath, err := integration.GetOktetoPath()
```
jLopezbarb, line 33: We'll need to ensure that the cluster is on 1.27 or later and that the okteto cluster has the endpoint. We should skip this tests if those criteria are not meet

andreafalzetti: @jLopezbarb don't we run our tests against staging? I assume that if we have control over the target cluster, we can safely assume this feature will be always available. Please correct me if I am wrong

ifbyol: @andreafalzetti  Technically, they can run against any cluster changing properly some environment variable. But check my previous comment. Could you check since which Kubernetes version is TokenRequest API available?

andreafalzetti: @ifbyol the TokenRequest API seems to be available in Kubernetes since 1.14 according to [this](https://jpweber.io/blog/a-look-at-tokenrequest-api/)

andreafalzetti:
Closing in favor of https://github.com/okteto/okteto/pull/3891

PR #3886, opened 2023-08-06
Add support for deleting multiple context

### LINKED ISSUE #3857: support deleteing multiple contexts at once
Right now you can't delete multiple contexts at once when running `okteto ctx delete`

Being able to delete multiple contexts using `okteto ctx delete ctx1 ctx2 ctx3`

**Additional context**
Screenshot 2023-07-28 at 1 35 46 PM

jeremyphua:
# Proposed changes

Fixes #3857  (issue)

Support deletion of multiple context and adding analytic for tracking if more than context is deleted using the command.

## Changes
* Adding for loop in `DeleteCmd` to loop through all `args` and move existing logic inside
* Add `TrackDeleteMultipleContexts` in `track.go` to track if more than one context is deleted

### DIFF
--- a/cmd/context/delete.go
+++ b/cmd/context/delete.go
@@ -37,8 +37,20 @@ func DeleteCMD() *cobra.Command {
 				args[idx] = okteto.AddSchema(arg)
 				args[idx] = strings.TrimSuffix(arg, "/")
 			}
-			analytics.TrackContextDelete(len(args))
-			return Delete(args)
+			errs := Delete(args)
+			var errLen int
+			var totalContextsDeleted int
+			if errs != nil {
+				if merr, ok := errs.(*multierror.Error); ok {
+					errLen = len(merr.Errors)
+				}
+			}
+			totalContextsDeleted = len(args) - errLen
+			if len(args) == errLen {
+				analytics.TrackContextDelete(totalContextsDeleted, false)
+			}
+			analytics.TrackContextDelete(totalContextsDeleted, true)
+			return errs
 		},
 	}
 	return cmd
--- a/cmd/context/delete.go
+++ b/cmd/context/delete.go
@@ -18,6 +18,7 @@ import (
 	"strings"
 
 	"github.com/okteto/okteto/cmd/utils"
+	"github.com/okteto/okteto/pkg/analytics"
 	oktetoErrors "github.com/okteto/okteto/pkg/errors"
 	oktetoLog "github.com/okteto/okteto/pkg/log"
 	"github.com/okteto/okteto/pkg/okteto"
@@ -28,12 +29,18 @@ import (
 func DeleteCMD() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "delete",
-		Args:  utils.ExactArgsAccepted(1, "https://okteto.com/docs/reference/cli/#delete"),
-		Short: "Delete a context",
+		Args:  utils.MinimumNArgsAccepted(1, "https://okteto.com/docs/reference/cli/#delete"),
+		Short: "Delete multiple contexts",
 		RunE: func(cmd *cobra.Command, args []string) error {
-			args[0] = okteto.AddSchema(args[0])
-			args[0] = strings.TrimSuffix(args[0], "/")
-			return Delete(args[0])
+			analytics.TrackDeleteMultipleContexts(len(args))
+			for _, arg := range args {
+				arg = okteto.AddSchema(arg)
+				arg = strings.TrimSuffix(arg, "/")
+				if err := Delete(arg); err != nil {
+					return err
+				}
+			}
+			return nil
 		},
 	}
 	return cmd
--- a/cmd/context/delete.go
+++ b/cmd/context/delete.go
@@ -17,6 +17,7 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/hashicorp/go-multierror"
 	"github.com/okteto/okteto/cmd/utils"
 	"github.com/okteto/okteto/pkg/analytics"
 	oktetoErrors "github.com/okteto/okteto/pkg/errors"
@@ -32,43 +33,44 @@ func DeleteCMD() *cobra.Command {
 		Args:  utils.MinimumNArgsAccepted(1, "https://okteto.com/docs/reference/cli/#delete"),
 		Short: "Delete one or more contexts",
 		RunE: func(cmd *cobra.Command, args []string) error {
-			for _, arg := range args {
-				arg = okteto.AddSchema(arg)
-				arg = strings.TrimSuffix(arg, "/")
-				if err := Delete(arg); err != nil {
-					return err
-				}
+			for idx, arg := range args {
+				args[idx] = okteto.AddSchema(arg)
+				args[idx] = strings.TrimSuffix(arg, "/")
 			}
 			analytics.TrackContextDelete(len(args))
-			return nil
+			return Delete(args)
 		},
 	}
 	return cmd
 }
 
-func Delete(okCtx string) error {
+func Delete(okCtxs []string) error {
 	ctxStore := okteto.ContextStore()
-	if okCtx == ctxStore.CurrentContext {
-		ctxStore.CurrentContext = ""
+	var errs error
+	for _, okCtx := range okCtxs {
+		if okCtx == ctxStore.CurrentContext {
+			ctxStore.CurrentContext = ""
+		}
+
+		if _, ok := ctxStore.Contexts[okCtx]; ok {
+			delete(ctxStore.Contexts, okCtx)
+			if err := okteto.NewContextConfigWriter().Write(); err != nil {
+				return err
+			}
+			oktetoLog.Success("'%s' deleted successfully", okCtx)
+		} else {
+			validOptions := make([]string, 0)
+			for k, v := range ctxStore.Contexts {
+				if v.IsOkteto {
+					validOptions = append(validOptions, k)
+				}
+			}
+			errs = multierror.Append(errs, oktetoErrors.UserError{
+				E:    fmt.Errorf("'%s' context doesn't exist. Valid options are: [%s]", okCtx, strings.Join(validOptions, ", ")),
+				Hint: fmt.Sprintf("To delete a Kubernetes context run 'kubectl config delete-context %s'", okCtx),
+			})
+		}
 	}
 
-	if _, ok := ctxStore.Contexts[okCtx]; ok {
-		delete(ctxStore.Contexts, okCtx)
-		if err := okteto.NewContextConfigWriter().Write(); err != nil {
-			return err
-		}
-		oktetoLog.Success("'%s' deleted successfully", okCtx)
-	} else {
-		validOptions := make([]string, 0)
-		for k, v := range ctxStore.Contexts {
-			if v.IsOkteto {
-				validOptions = append(validOptions, k)
-			}
-		}
-		return oktetoErrors.UserError{
-			E:    fmt.Errorf("'%s' context doesn't exist. Valid options are: [%s]", okCtx, strings.Join(validOptions, ", ")),
-			Hint: fmt.Sprintf("To delete a Kubernetes context run 'kubectl config delete-context %s'", okCtx),
-		}
-	}
-	return nil
+	return errs
 }
--- a/cmd/context/delete.go
+++ b/cmd/context/delete.go
@@ -20,7 +20,6 @@ import (
 	"github.com/hashicorp/go-multierror"
 	"github.com/okteto/okteto/cmd/utils"
 	"github.com/okteto/okteto/pkg/analytics"
-	oktetoErrors "github.com/okteto/okteto/pkg/errors"
 	oktetoLog "github.com/okteto/okteto/pkg/log"
 	"github.com/okteto/okteto/pkg/okteto"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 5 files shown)


cmd/context/delete.go
```diff
@@ -28,12 +29,18 @@ import (
 func DeleteCMD() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "delete",
-		Args:  utils.ExactArgsAccepted(1, "https://okteto.com/docs/reference/cli/#delete"),
-		Short: "Delete a context",
+		Args:  utils.MinimumNArgsAccepted(1, "https://okteto.com/docs/reference/cli/#delete"),
```
andreafalzetti, line 32: it would be nice to update the docs reflecting this change

jeremyphua: Hi @andreafalzetti , is the update of docs supposed to be done on my side? If so, are you able to point me in the right direction to do it because I am not too sure. Thanks in advance!

cmd/context/delete.go
```diff
@@ -28,12 +29,18 @@ import (
...
+		Short: "Delete multiple contexts",
 		RunE: func(cmd *cobra.Command, args []string) error {
-			args[0] = okteto.AddSchema(args[0])
-			args[0] = strings.TrimSuffix(args[0], "/")
-			return Delete(args[0])
+			analytics.TrackDeleteMultipleContexts(len(args))
```
andreafalzetti, line 35:
```suggestion
			analytics.TrackContextDelete(len(args))
```

I would move this function after the actual delete logic and callit `TrackContextDelete` being consistent with `TrackContextUse` and another one already there. I would set the quantity of deleted contexts a properties rather than a different event. WDYT?

cmd/context/delete.go
andreafalzetti: up to you really, the docs is open source too: https://github.com/okteto/docs

I've left the comment so at least we don't forget to do it

cmd/context/delete.go
jeremyphua: Moving after the actual delete logic is a good idea   , does this means that it will only be tracked if there’s no error? 

Yes setting the quantity of deleted contexts as a properties is good! I believed this is inline with the changes I made currently in the `TrackDeleteMultipleContexts` function

andreafalzetti: > Moving after the actual delete logic is a good idea , does this means that it will only be tracked if there’s no error?

If we move it after the logic we could also add a property if the cmd failed

cmd/context/delete.go
```diff
@@ -28,12 +29,18 @@ import (
 func DeleteCMD() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "delete",
-		Args:  utils.ExactArgsAccepted(1, "https://okteto.com/docs/reference/cli/#delete"),
-		Short: "Delete a context",
+		Args:  utils.MinimumNArgsAccepted(1, "https://okteto.com/docs/reference/cli/#delete"),
+		Short: "Delete multiple contexts",
```
AdrianPedriza, line 33:
```suggestion
		Short: "Delete one or more contexts",
```
We allow to delete just one context and that case doesn't fit with the current description

pkg/analytics/track.go
```diff
@@ -69,6 +69,7 @@ const (
 	disableEvent             = "Disable Analytics"
 	stackNotSupportedField   = "Stack Field Not Supported"
 	buildPullErrorEvent      = "BuildPullError"
+	deleteMultipleContexts   = "Delete Multiple Contexts"
```
AdrianPedriza, line 72:
```suggestion
	deleteContexts   = "contexts deletion"
```
as event name I would use something more descriptive in term of event identifier and use a variable which "allow" only the deletion of one context

AdrianPedriza: > nit: as event name I would use something more descriptive in term of event identifier and use a variable which "allow" only the deletion of one context

cmd/context/delete.go
```diff
@@ -28,12 +29,18 @@ import (
...
-			args[0] = strings.TrimSuffix(args[0], "/")
-			return Delete(args[0])
+			for _, arg := range args {
+				arg = okteto.AddSchema(arg)
+				arg = strings.TrimSuffix(arg, "/")
+				if err := Delete(arg); err != nil {
+					return err
+				}
+			}
+			analytics.TrackContextDelete(len(args))
```
AdrianPedriza, line 42: if I understood correctly @andreafalzetti, your proposal is that we could store if there is an error in the delete logic and track it along with the number of contexts to delete right? After that we could return the possible error to the user

pkg/analytics/track.go
```diff
@@ -371,6 +372,14 @@ func TrackBuildPullError(oktetoBuilkitURL string, success bool) {
 	track(buildPullErrorEvent, success, props)
 }
 
+// TrackContextDelete sends a tracking event to mixpanel indicating whether more than one context is deleted
+func TrackContextDelete(ctxs int) {
+	props := map[string]interface{}{
+		"hasMoreThanOneContext": ctxs > 1,
```
andreafalzetti, line 378:
```suggestion
		"totalContextsDeleted": ctxs,
```

pkg/analytics/track.go
```diff
@@ -371,6 +372,14 @@ func TrackBuildPullError(oktetoBuilkitURL string, success bool) {
 	track(buildPullErrorEvent, success, props)
 }
 
+// TrackContextDelete sends a tracking event to mixpanel indicating whether more than one context is deleted
```
andreafalzetti, line 375:
```suggestion
// TrackContextDelete sends a tracking event to mixpanel indicating one or more contexts have been deleted
```

cmd/context/delete.go
andreafalzetti: @AdrianPedriza I think we're good without it, we have no plans to act on that info anyway

pkg/analytics/track.go
```diff
@@ -371,6 +372,14 @@ func TrackBuildPullError(oktetoBuilkitURL string, success bool) {
 	track(buildPullErrorEvent, success, props)
 }
 
+// TrackContextDelete sends a tracking event to mixpanel indicating one or more context have been deleted
+func TrackContextDelete(ctxs int) {
+	props := map[string]interface{}{
+		"hasMoreThanOneContext": ctxs,
```
AdrianPedriza, line 378: we need to add a proper name to the namber of context deleted as suggested in https://github.com/okteto/okteto/pull/3886#discussion_r1288170817. `totalContextsDeleted` would be a good name 👍

cmd/context/delete.go
```diff
@@ -28,12 +29,18 @@ import (
...
+		Args:  utils.MinimumNArgsAccepted(1, "https://okteto.com/docs/reference/cli/#delete"),
+		Short: "Delete one or more contexts",
 		RunE: func(cmd *cobra.Command, args []string) error {
-			args[0] = okteto.AddSchema(args[0])
-			args[0] = strings.TrimSuffix(args[0], "/")
-			return Delete(args[0])
+			for _, arg := range args {
+				arg = okteto.AddSchema(arg)
+				arg = strings.TrimSuffix(arg, "/")
+				if err := Delete(arg); err != nil {
```
AdrianPedriza, line 38: inside `Delete` function we are getting okteto context[ from storage](https://github.com/jeremyphua/okteto/blob/feature/support-deleting-multiple-context/cmd/context/delete.go#L50) for every context to delete. Instead of that we should send to `Delete` function all context to delete and iterate over contexts to delete them once we instantiate context storage.

AdrianPedriza: If deletion of one context fails because contexts doesn't exist it prevents deleting the rest of contexts so I propose to log an error but be able to delete the rest of contexts. WDYT about that UX @rlamana ?

AdrianPedriza:
I think we need to update https://github.com/okteto/docs to reflect this new behavior in the CLI. @jeremyphua can you take care of it? 😃

cmd/context/delete.go
jeremyphua: Hey! To clarify, if more than one context fails we return back all the errors?

rlamana: Agree with @AdrianPedriza, we should in that case delete the contexts that can be deleted but log an error for the one that has failed. If more than one error happens we will show them all, as long as we specify the name of the context that failed. As an example of output:
```
 ❌ 'xxxxx' doesn't exist and couldn't be deleted
 ❌ 'yyyyy' doesn't exist and couldn't be deleted
 ✅ 'zzzzz' deleted successfully
 ```

Comment (rlamana) [reply cmd/context/delete.go:38]: Btw, the `usage` section should also be updated, it's not only flags that this command accepts
Screenshot 2023-08-21 at 14 05 20

File: pkg/analytics/track.go [hunk 10]
```diff
@@ -62,6 +62,7 @@ const (
 	disableEvent             = "Disable Analytics"
 	stackNotSupportedField   = "Stack Field Not Supported"
 	buildPullErrorEvent      = "BuildPullError"
+	deleteContexts           = "Contexts Deletion"
```
andreafalzetti, line 65:
```suggestion
	deleteContexts           = "Contexts Delete"
```

andreafalzetti: @AdrianPedriza I would keep the name simpler, just "Delete" instead of "Deletion". WDYT?

cmd/context/delete.go
```diff
@@ -28,40 +30,47 @@ import (
...
 		RunE: func(cmd *cobra.Command, args []string) error {
-			args[0] = okteto.AddSchema(args[0])
-			args[0] = strings.TrimSuffix(args[0], "/")
-			return Delete(args[0])
+			for idx, arg := range args {
+				args[idx] = okteto.AddSchema(arg)
+				args[idx] = strings.TrimSuffix(arg, "/")
+			}
+			analytics.TrackContextDelete(len(args))
+			return Delete(args)
```
andreafalzetti, line 41: If I run `okteto ctx delete foo foo foo foo foo foo foo foo foo foo foo` the track event will report `success: true` and `totalContextsDeleted: 11` which is misleading.

To maintain integrity, we could:
- `errs := Delete(args)`
- track event after deletion
- pass to `totalContextsDeleted = len(args) - len(errs)`
- if `len(args) == len(errs)` then pass `success: false`
- `return errs`

andreafalzetti: @jeremyphua will you look into the above? if you don't have time, just let me know and I can follow up, cheers!

jeremyphua: Hi @AdrianPedriza, I've made the changes in the latest commit! But I'm not too sure if it is acceptable.

cmd/context/delete.go
```diff
@@ -37,8 +37,20 @@ func DeleteCMD() *cobra.Command {
...
+			if errs != nil {
+				if merr, ok := errs.(*multierror.Error); ok {
+					errLen = len(merr.Errors)
+				}
+			}
+			totalContextsDeleted = len(args) - errLen
+			if len(args) == errLen {
+				analytics.TrackContextDelete(totalContextsDeleted, false)
+			}
+			analytics.TrackContextDelete(totalContextsDeleted, true)
```
AdrianPedriza, line 52: if we send a `success== false` when `len(args) == errLen` to track it then we send `success == true` again to the tracker. I think this block should be, right?:
```go
success := len(args) == errLen
analytics.TrackContextDelete(totalContextsDeleted, success)
```

cmd/context/delete.go
```diff
@@ -28,40 +30,59 @@ import (
...
+			return errs
 		},
 	}
 	return cmd
 }
 
-func Delete(okCtx string) error {
+func Delete(okCtxs []string) error {
```
AdrianPedriza, line 59: In order to return a proper output following the criteria and the hint we want to display, let's use the following `Delete` func:
````go
func Delete(okCtxs []string) error {
	var errs error
	validOptions := make([]string, 0)
	for _, okCtx := range okCtxs {
		if okCtx == ctxStore.CurrentContext {
			ctxStore.CurrentContext = ""
		}

		if _, ok := ctxStore.Contexts[okCtx]; ok {
			delete(ctxStore.Contexts, okCtx)
			if err := okteto.NewContextConfigWriter().Write(); err != nil {
				return err
			}
			oktetoLog.Success("'%s' deleted successfully", okCtx)
		} else {
			for k, v := range ctxStore.Contexts {
				if v.IsOkteto {
					validOptions = append(validOptions, k)
				}
			}
			errs = multierror.Append(errs, fmt.Errorf("'%s' context doesn't exist.", okCtx))
		}
	}

	if errs != nil {
		if err, ok := errs.(*multierror.Error); ok {
			err.ErrorFormat = func(es []error) string {
				points := make([]string, len(es))
				for i, err := range es {
					points[i] = fmt.Sprintf("\t- %s", err)
				}

				hint := fmt.Sprintf("Valid options are: [%s]. To delete a Kubernetes context run 'kubectl config delete-context <k8s-context-name>'", strings.Join(validOptions, ", "))
				return fmt.Sprintf("Following contexts couldn't be deleted:\n%s\n%s\n\n", strings.Join(points, "\n"), hint)
			}
		}
	}

	return errs
}
````
output will be:
Captura de Pantalla 2023-09-06 a las 16 51 32

AdrianPedriza (approved):
LGTM. Good work! 👍

jLopezbarb:
@jeremyphua Changes looks good to me. Could you resolve conflicts? After that we can merge this PR. Thanks in advance!!

jLopezbarb (approved):

PR #3887, opened 2023-08-07
Add exec to kubeconfig when okteto context init

teresaromero:
# Proposed changes

Resolves https://github.com/okteto/app/issues/6907

Changes to include the kubetoken auth plugin instead of token.

- Refactor at `kubeconfig` cmd
- Refactor `ExecuteUpdateKubeconfig` at `update-kubeconfig` injecting the context and the kubeconfig paths from `RunE` function
- Request to kubetoken CheckService in order to use Token or Exec. If service is available, Exec is used.

## Testing
- included integration test to check running `okteto kubeconfig` sets the correct auth info into the kubeconfig file
- manual testing, removing both okteto and kube context files, running `okteto kubeconfig` initialized the okteto context and includes the auth exec command into the kubeconfig file
- used helm and kubectl with the okteto context normally

### DIFF
--- a/cmd/context/update-kubeconfig.go
+++ b/cmd/context/update-kubeconfig.go
@@ -53,15 +53,23 @@ func UpdateKubeconfigCMD() *cobra.Command {
 
 			okCtx := okteto.Context()
 			kubeconfigPath := config.GetKubeconfigPath()
+			kubetokenEnabled := false
 
-			return ExecuteUpdateKubeconfig(okCtx, kubeconfigPath)
+			okClient, err := okteto.NewOktetoClient()
+			if err != nil {
+				return err
+			}
+			if err := okClient.Kubetoken().CheckService(okCtx.Name, okCtx.Namespace); err == nil {
+				kubetokenEnabled = true
+			}
+			return ExecuteUpdateKubeconfig(okCtx, kubeconfigPath, kubetokenEnabled)
 		},
 	}
 
 	return cmd
 }
 
-func ExecuteUpdateKubeconfig(okContext *okteto.OktetoContext, kubeconfigPaths []string) error {
+func ExecuteUpdateKubeconfig(okContext *okteto.OktetoContext, kubeconfigPaths []string, kubetokenEnabled bool) error {
 	contextName := okContext.Name
 	if okContext.IsOkteto {
 		contextName = okteto.UrlToKubernetesContext(contextName)
@@ -73,6 +81,10 @@ func ExecuteUpdateKubeconfig(okContext *okteto.OktetoContext, kubeconfigPaths []
 				return err
 			}
 			okContext.Cfg.Clusters[contextName].CertificateAuthorityData = certPEM
+		}
+
+		if kubetokenEnabled {
+			updateUserAuthInfoWithExec(okContext, okContext.UserID)
 		}
 	}
 
--- a/cmd/context/update-kubeconfig.go
+++ b/cmd/context/update-kubeconfig.go
@@ -35,40 +35,40 @@ func UpdateKubeconfigCMD() *cobra.Command {
 		RunE: func(cmd *cobra.Command, args []string) error {
 			ctx := context.Background()
 
+			// Run context command to get the Cfg into Okteto Context
 			if err := NewContextCommand().Run(ctx, &ContextOptions{}); err != nil {
 				return err
 			}
 
-			if err := ExecuteUpdateKubeconfig(); err != nil {
-				return err
-			}
+			okCtx := okteto.Context()
+			kubeconfigPath := config.GetKubeconfigPath()
 
-			return nil
+			return ExecuteUpdateKubeconfig(okCtx, kubeconfigPath)
 		},
 	}
 
 	return cmd
 }
 
-func ExecuteUpdateKubeconfig() error {
-	ctx := okteto.Context()
-	k8sContext := okteto.Context().Name
+func ExecuteUpdateKubeconfig(okContext *okteto.OktetoContext, kubeconfigPaths []string) error {
+	contextName := okContext.Name
+	if okContext.IsOkteto {
+		contextName = okteto.UrlToKubernetesContext(contextName)
 
-	if ctx.IsOkteto {
-		k8sContext = okteto.UrlToKubernetesContext(k8sContext)
-		if ctx.IsStoredAsInsecure {
+		if okContext.IsStoredAsInsecure {
 			certPEM, err := base64.StdEncoding.DecodeString(okteto.Context().Certificate)
 			if err != nil {
 				oktetoLog.Debugf("couldn't decode context certificate from base64: %s", err)
 				return err
 			}
-			ctx.Cfg.Clusters[k8sContext].CertificateAuthorityData = certPEM
+			okContext.Cfg.Clusters[contextName].CertificateAuthorityData = certPEM
 		}
 	}
-	if err := kubeconfig.Write(ctx.Cfg, config.GetKubeconfigPath()[0]); err != nil {
+
+	if err := kubeconfig.Write(okContext.Cfg, kubeconfigPaths[0]); err != nil {
 		return err
 	}
+	oktetoLog.Success("Updated kubernetes context '%s/%s' in '%s'", contextName, okContext.Namespace, kubeconfigPaths)
 
-	oktetoLog.Success("Updated kubernetes context '%s/%s' in '%s'", k8sContext, ctx.Namespace, config.GetKubeconfigPath())
 	return nil
 }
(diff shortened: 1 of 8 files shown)


cmd/context/create.go
```diff
@@ -265,6 +265,18 @@ func (c *ContextCommand) initOktetoContext(ctx context.Context, ctxOptions *Cont
...
 	}
 	okteto.AddOktetoCredentialsToCfg(cfg, &userContext.Credentials, ctxOptions.Namespace, userContext.User.ID, okteto.Context().Name)
+
+	ktClient, err := okteto.NewKubeTokenClient(ctxOptions.Context, userContext.Credentials.Token, ctxOptions.Namespace)
+	if err != nil {
+		oktetoLog.Debug("unable to retrieve kubetoken client")
+		return err
+	}
+	if err := ktClient.Head(); err == nil {
+		okteto.AddKubetokenExecToCfg(cfg, userContext.User.ID)
```
ifbyol, line 275: I think this has to be done only when `okteto kubeconfig` command is executed. If I'm not wrong, this would modify all the internal Kubeconfig for the CLI forcing users to have `okteto` in the path when the CLI is executed and it's something we don't want. We just want to force it when they use `kubectl` or any other external tool using our generated kubeconfig.

Also, there is intention to add a feature flag to control when to use kubetoken (https://github.com/okteto/app/issues/6886) and we should check it before doing it not

teresaromero: ok, i think i understood then wrong wher to add the implementation :) i will review and make required modifications

cmd/context/update-kubeconfig.go
```diff
@@ -35,40 +46,65 @@ func UpdateKubeconfigCMD() *cobra.Command {
...
+			kubeconfigPath := config.GetKubeconfigPath()
+			kubetokenEnabled := false
+
+			okClient, err := okteto.NewOktetoClient()
+			if err != nil {
 				return err
 			}
-
-			return nil
+			if err := okClient.Kubetoken().CheckService(okCtx.Name, okCtx.Namespace); err == nil {
```
ifbyol, line 62: Should we also check here the feature flag we are adding in other PR?

teresaromero: i believe we will on the PR for feature flag 

cc: @andreafalzetti

andreafalzetti: The feature flag is covered by https://github.com/okteto/app/issues/6886 so as soon as this PR is merged, I can rebase and finish the [feature flag PR](https://github.com/okteto/okteto/pull/3893)

andreafalzetti:
Tested both Mac and Windows and works as expected 👏

cmd/context/update-kubeconfig.go
```diff
@@ -23,6 +23,17 @@ import (
...
+	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
+)
+
+var (
+	oktetoExecConfig = &clientcmdapi.ExecConfig{
+		APIVersion:      "client.authentication.k8s.io/v1",
+		Command:         "okteto",
+		Args:            []string{"kubetoken"},
+		InstallHint:     "Okteto needs to be installed and in your PATH to use this context. Please visit https://www.okteto.com/docs/getting-started/ for more information.",
+		InteractiveMode: "IfAvailable",
```
andreafalzetti, line 35:
```suggestion
		InteractiveMode: "Never",
```

cmd/context/update-kubeconfig.go
```diff
@@ -23,6 +23,17 @@ import (
...
 	"github.com/okteto/okteto/pkg/okteto"
 	"github.com/spf13/cobra"
+	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
+)
+
+var (
+	oktetoExecConfig = &clientcmdapi.ExecConfig{
+		APIVersion:      "client.authentication.k8s.io/v1",
+		Command:         "okteto",
+		Args:            []string{"kubetoken"},
```
andreafalzetti, line 33: I think we should specify the `--context` and `--namespace` to avoid any interaction with the context prompt

cmd/context/update-kubeconfig_test.go
```diff
@@ -106,24 +106,21 @@ func Test_updateKubeconfig(t *testing.T) {
...
+			okContext := okteto.Context()
+			kubeconfigPaths := []string{file}
+
+			err = ExecuteUpdateKubeconfig(okContext, kubeconfigPaths, false)
+			assert.NoError(t, err, "error writing kubeconfig")
+
+			cfg := kubeconfig.Get(kubeconfigPaths)
+			assert.NotNil(t, cfg, "kubeconfig is nil")
+			assert.Equal(t, tt.context.CurrentContext, cfg.CurrentContext, "current context has changed")
+			assert.Equal(t, tt.context.Contexts[tt.context.CurrentContext].Namespace, cfg.Contexts[tt.context.CurrentContext].Namespace, "namesapce has changed")
```
jLopezbarb, line 122:
```suggestion
			assert.Equal(t, tt.context.Contexts[tt.context.CurrentContext].Namespace, cfg.Contexts[tt.context.CurrentContext].Namespace, "namespace has changed")
```

cmd/context/update-kubeconfig.go
jLopezbarb: Agree, let's specify them so we can change k8s context without changing okteto context

cmd/context/update-kubeconfig.go
```diff
@@ -23,6 +23,17 @@ import (
 	oktetoLog "github.com/okteto/okteto/pkg/log"
 	"github.com/okteto/okteto/pkg/okteto"
 	"github.com/spf13/cobra"
+	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
+)
+
+var (
+	oktetoExecConfig = &clientcmdapi.ExecConfig{
```
AdrianPedriza, line 30: could we use the value directly [here](https://github.com/okteto/okteto/pull/3887/files#diff-e68e63e6d73f6931e3cc1904794fbb6f72948431573b0d73036c90e69f2022ffR109) instead declare a global variable?

cmd/context/update-kubeconfig.go
```diff
@@ -35,40 +46,65 @@ func UpdateKubeconfigCMD() *cobra.Command {
...
+			okClient, err := okteto.NewOktetoClient()
+			if err != nil {
 				return err
 			}
-
-			return nil
+			if err := okClient.Kubetoken().CheckService(okCtx.Name, okCtx.Namespace); err == nil {
+				kubetokenEnabled = true
+			}
+			return ExecuteUpdateKubeconfig(okCtx, kubeconfigPath, kubetokenEnabled)
```
AdrianPedriza, line 65: this function has different behaviors if the auth kubetoken is activated or not. Avoid using argument flags is a proposal in improving the quality of the CLI I believe. Could we refactor this logic by removing `kubetokenEnabled` arg flag and separate the behavior in different functions when updating the kubeconfig depending if kubetoken auth is enabled or not?

cmd/context/update-kubeconfig.go
teresaromero: i created the func to pass the params as flags

cmd/context/update-kubeconfig.go
teresaromero: yes, i removed the global and added a function to include this with params

cmd/context/update-kubeconfig.go
```diff
@@ -35,40 +36,77 @@ func UpdateKubeconfigCMD() *cobra.Command {
...
+	if okCtx.Cfg.AuthInfos == nil {
+		okCtx.Cfg.AuthInfos = clientcmdapi.NewConfig().AuthInfos
+		okCtx.Cfg.AuthInfos[userID] = clientcmdapi.NewAuthInfo()
+	}
+
+	if token := okCtx.Cfg.AuthInfos[userID].Token; token != "" {
+		okCtx.Cfg.AuthInfos[userID].Token = ""
+	}
+
+	okCtx.Cfg.AuthInfos[userID].Exec = getExecConfigForContextAndNamespace(okCtx.Name, okCtx.Namespace)
```
AdrianPedriza, line 99: Considering the context, would it be feasible to directly utilize the value of `clientcmdapi.ExecConfig` at this point, rather than introducing a separate function? It appears that directly incorporating the value might enhance readability, given that the function doesn’t encapsulate any specific logic and isn’t employed elsewhere for reuse.

cmd/context/update-kubeconfig.go
teresaromero: i've refactored ExecuteUpdateKubeconfig in order to inject the client provider and do the operations inside. Also i added test cases for when kubetoken in enabled.

AdrianPedriza (approved):
scenarios have been checked and the code looks good!

jLopezbarb (approved):
Code looks good and the scenarios are working!! Nice job

PR #3888, opened 2023-08-07
Wrap kubetoken client into OktetoClient

teresaromero:
# Proposed changes

this PR wraps the existent kubetoken client code into the OktetoClient interface in order to be able to create testeable flows with this client.

Also, added new function CheckService to request a HEAD call to check if the service is available at the cluster

### DIFF
--- a/pkg/okteto/kubetoken_client.go
+++ b/pkg/okteto/kubetoken_client.go
@@ -27,9 +27,10 @@ const (
 )
 
 var (
-	errRequest      = errors.New("failed request")
-	errStatus       = errors.New("status error")
-	errUnauthorized = errors.New("unauthorized")
+	errRequest               = errors.New("failed request")
+	errStatus                = errors.New("status error")
+	errUnauthorized          = errors.New("unauthorized")
+	errKubetokenNotAvailable = errors.New("kubetoken service not found")
 )
 
 type kubeTokenClient struct {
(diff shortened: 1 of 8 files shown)


andreafalzetti (approved):

pkg/types/kubetoken.go
```diff
@@ -0,0 +1,18 @@
...
+import (
+	"encoding/json"
+	authenticationv1 "k8s.io/api/authentication/v1"
+)
+
+type KubeTokenResponse struct {
+	authenticationv1.TokenRequest
+}
+
+func (k *KubeTokenResponse) ToJson() (string, error) {
```
jLopezbarb, line 12: NIT: types package should not contain any method. ToJson should be a struct that transform a kubetoken to json. If we want to extend functionality and create a `ToYAML`  we don't have to modify `KubeTokenResponse` WDYT?

pkg/types/kubetoken.go
```diff
@@ -0,0 +1,18 @@
+package types
+
+import (
+	"encoding/json"
+	authenticationv1 "k8s.io/api/authentication/v1"
+)
+
+type KubeTokenResponse struct {
```
jLopezbarb, line 8: Should we add a comment here?

pkg/okteto/kubetoken_client.go
```diff
@@ -14,56 +14,92 @@
...
+	var kubeTokenResponse types.KubeTokenResponse
+	err = json.Unmarshal(body, &kubeTokenResponse)
+	if err != nil {
+		return nil, fmt.Errorf("failed to unmarshal kubetoken response: %w", err)
+	}
+
+	return &kubeTokenResponse, nil
+}
+
+func (c *kubeTokenClient) CheckService(baseURL, namespace string) error {
```
jLopezbarb, line 85: Just to make sure. This function is only used on tests.Should we call it when we init the kubetoken client?

jLopezbarb (approved):
Minor comments only. LGTM!!

pkg/okteto/kubetoken_client.go
teresaromero: this is going to be used in the following PRs, we added as both andrea and me will needed for our issues

pkg/types/kubetoken.go
andreafalzetti: makes sense, done in 1945273c3da13950ac31a1fe3162716b6e580d2e

teresaromero (commented):
👍🏻

pkg/okteto/kubetoken_client.go
jLopezbarb: Thanks for explaining it :)

jLopezbarb (approved):

PR #3890, opened 2023-08-07
fix: compose not getting image expansion from build

### LINKED ISSUE #3878: I can't deploy a compose with a generated image from the build
**Describe the bug**
I can't deploy my application because whenever I try to deploy my compose I get the following error:
` x  Invalid service 'service': image cannot be empty`. This bug is a regression because i can make it work downgrading my CLI to 2.16.5 by running: `curl https://get.okteto.com -sSfL | OKTETO_CHANNEL=stable OKTETO_VERSION=2.16.5 sh`

**To Reproduce**
1. Create the following okteto manifest:
```yaml
build:
  service:
    context: .
    dockerfile: ./Dockerfile
deploy:
  compose:
    - file: ./docker-compose.yml
```
2. Create the following docker compose:
```yaml
services:
  service:
    image: ${OKTETO_BUILD_SERVICE_IMAGE}
    command: sleep 100
```
3. Create the following Dockerfile:
```Dockerfile
FROM alpine

RUN echo hello
```
4. Download CLI 2.18.0:
```bash
curl https://get.okteto.com -sSfL | OKTETO_CHANNEL=stable OKTETO_VERSION=2.18.0 sh
```
5. Run `okteto deploy` and check that it fails:
```
 i  Using jlopezbarb @ cloud.okteto.com as context
 x  Invalid service 'service': image cannot be empty
```
6. Download CLI 2.16.5:
```bash
curl https://get.okteto.com -sSfL | OKTETO_CHANNEL=stable OKTETO_VERSION=2.16.5 sh
```
7. Run `okteto deploy` and check that it's working:
```
 ✓  Image 'registry.cloud.okteto.net/jlopezbarb/test-service:okteto' successfully pushed
 ✓  Service 'service' created
 ✓  Compose 'test' successfully deployed
 ✓  Development environment 'test' successfully deployed
```

**Expected behavior**
It should work on the latest version

jLopezbarb:
## Bug Introduction and Root Cause

The bug was causing that compose files were not getting variable expansions when the image was set in the docker compose.

### Bug Description

We couldn't deploy applications that have a compose using the build env vars because whenever we deploy a compose using them we got the following error:
```
 x  Invalid service 'service': image cannot be empty.
 ```

## Bug Fix Explanation

The bug was a regression introduced in CLI 2.17.0 by the PR #3703. We fixed a panic when we deploy a compose with a service that had a `depends_on` field pointing to a non existing service. Adding the check we were also expanding all the env vars in the compose and because of that, the issue that is being solved arised

### Fix Description

In order to fix both bugs I tried the following approach:
1. Revert the changes done in the previous PR without doing a revert per se because the previous fix added several unit tests and explanatory comments that are worth maintaining them
2. Check that the issue #3878 is fixed with those changes
3. Try to reproduce the panic that was fixed by #3703
4. Try to fix it in a different way of the approach followed. The approach followed was to validate the services when we were about to deploy and not during the unmarshal of the compose
5. Modify an e2e test in order to not break it again the scenario
6. Add e2e testing that the compose file does't get a panic while running a depends on pointing to a non existen service

### Testing and Validation

There are two scenarios that needs to be addressed:

**Build env vars expansion**
1. Create the following okteto manifest:
```yaml
build:
  service:
    context: .
    dockerfile: ./Dockerfile
deploy:
  compose:
    - file: ./docker-compose.yml
```
2. Create the following docker compose:
```yaml
services:
  service:
    image: ${OKTETO_BUILD_SERVICE_IMAGE}
    command: sleep 100
```
3. Create the following Dockerfile:
```Dockerfile
FROM alpine

RUN echo hello
```
4. Run `okteto deploy` and check that it gets deployed:

**Panic on compose pointing to a non existent service**

1. Clone movies-with-compose:
```bash
git clone https://github.com/okteto/movies-with-compose.git
```
2. Remove the frontend service from the compose file
3. Run `okteto deploy -f docker-compose.yml`. It should return a proper error and not panic
4. Run `okteto deploy`. It should return a proper error and not panic

### Impact and Considerations

This bug was causing a breaking change in 2.17 and we should patch 2.18 and 2.17 in order to not have a breaking scenario in our maintained versions

## Checklist

- [ ] Bug fix code has been reviewed by at least one team member.
- [ ] Automated tests have been added or updated to cover the bug fix.
- [ ] Manual testing has been performed to verify the bug fix.

## Related Issues

Fixes #3878
Follow up of  #3703

### DIFF
--- a/cmd/deploy/deploy.go
+++ b/cmd/deploy/deploy.go
@@ -278,7 +278,7 @@ func (dc *DeployCommand) RunDeploy(ctx context.Context, deployOptions *Options) 
 	oktetoLog.SetStage("Load manifest")
 	manifest, err := dc.GetManifest(deployOptions.ManifestPath)
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to load manifest: %w", err)
 	}
 	deployOptions.Manifest = manifest
 	oktetoLog.Debug("found okteto manifest")
--- a/integration/commands/deploy.go
+++ b/integration/commands/deploy.go
@@ -53,27 +53,34 @@ type DestroyOptions struct {
 }
 
 // RunOktetoDeploy runs an okteto deploy command
-func RunOktetoDeploy(oktetoPath string, deployOptions *DeployOptions) error {
+func RunOktetoDeployWithOutput(oktetoPath string, deployOptions *DeployOptions) ([]byte, error) {
 	cmd := getDeployCmd(oktetoPath, deployOptions)
 	log.Printf("Running '%s'", cmd.String())
 	stdout, err := cmd.StdoutPipe()
 	if err != nil {
 		log.Printf("error getting stdout pipe: %s", err)
-		return err
+		return nil, err
 	}
 	if err := cmd.Start(); err != nil {
-		return err
+		return nil, err
 	}
 	go func() {
-
 		scanner := bufio.NewScanner(stdout)
-		// optionally, resize scanner's capacity for lines over 64K, see next example
 		for scanner.Scan() {
 			log.Println(scanner.Text())
 		}
 	}()
-	err = cmd.Wait()
+	output, err := cmd.CombinedOutput()
+	if err != nil {
+		return nil, err
+	}
 	log.Printf("okteto deploy success")
+	return output, nil
+}
+
+// RunOktetoDeploy runs an okteto deploy command
+func RunOktetoDeploy(oktetoPath string, deployOptions *DeployOptions) error {
+	_, err := RunOktetoDeployWithOutput(oktetoPath, deployOptions)
 	return err
 }
 
--- a/integration/commands/deploy.go
+++ b/integration/commands/deploy.go
@@ -14,7 +14,6 @@
 package commands
 
 import (
-	"bufio"
 	"fmt"
 	"log"
 	"os"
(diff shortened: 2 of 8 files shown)


cmd/deploy/deploy.go
```diff
@@ -278,7 +278,7 @@ func (dc *DeployCommand) RunDeploy(ctx context.Context, deployOptions *Options)
 	oktetoLog.SetStage("Load manifest")
 	manifest, err := dc.GetManifest(deployOptions.ManifestPath)
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to load manifest: %w", err)
```
teresaromero, line 281: do we have an error for this? Could we create one?

pkg/model/manifest.go
```diff
@@ -445,7 +445,7 @@ func getManifestFromFile(cwd, manifestPath string) (*Manifest, error) {
 				stackFiles = append(stackFiles, composeInfo.File)
 			}
 			// LoadStack should perform validation of the stack read from the file on compose section
```
teresaromero, line 447: could we re-write the comment to explay why is false instead of true?

teresaromero (approved):
minor comments, LGTM

pkg/model/stack.go
```diff
@@ -499,22 +507,51 @@ func validateStackName(name string) error {
...
 }
 
-func validateDependsOn(s *Stack) error {
-	for svcName, svc := range s.Services {
+type errDependsOnItself struct {
+	service string
+}
+
+func (e *errDependsOnItself) Error() string {
+	return fmt.Sprintf("%s: Service '%s' depends can not depend of itself.", errDependsOn.Error(), e.service)
```
andreafalzetti, line 515:
```suggestion
	return fmt.Sprintf("%s: Service '%s' depends cannot depend of itself.", errDependsOn.Error(), e.service)
```

andreafalzetti: both are correct

integration/commands/deploy.go
```diff
@@ -53,27 +53,34 @@ type DestroyOptions struct {
 }
 
 // RunOktetoDeploy runs an okteto deploy command
-func RunOktetoDeploy(oktetoPath string, deployOptions *DeployOptions) error {
+func RunOktetoDeployWithOutput(oktetoPath string, deployOptions *DeployOptions) ([]byte, error) {
```
AdrianPedriza, line 56:
```suggestion
func GetOktetoDeployCmdOutput(oktetoPath string, deployOptions *DeployOptions) ([]byte, error) {
```

PR #3891, opened 2023-08-07
Use dynamic kube token for k8s requests 

andreafalzetti:
# Proposed changes

Fixes https://github.com/okteto/app/issues/6906

~This branch requires this PR to be merged first: https://github.com/okteto/okteto/pull/3888~ ✅  done

## Description

This PR introduces a request to fetch the dynamic kubetoken from the Okteot backend. 

The dynamic kubetoken will be the default behaviour. If the endpoint is not available, the CLI will continue working with the static token returned in the graphql API call to retrieve credentials.

As part of this PR I've added an e2e test for `okteto kubetoken` which checks that the token returned is a dynamic token (with expiration date set)

## How to test

1. Build this branch and try  to deploy a project such as https://github.com/okteto/movies
2. Everything should work without any noticable effect
3. Now change the `kubetokenPathTemplate` and make the path such as `"%s/auth/kubetoken-notfound/%s"`
4. Build the CLI and try to deploy the same project
5. Everything should work without any noticable effect
6. Check the debug logs of both executions and make sure only in the second execution (with the broken path) you get the following log: `DEBU[0000] Dynamic Kubetoken not available, falling back to static token`


integration/kubetoken/kubetoken_test.go
```diff
@@ -0,0 +1,54 @@
...
+)
+
+// TestKubetokenHasExpiration test that kubernetes returns a dynamic token with a set expiration
+func TestKubetokenHasExpiration(t *testing.T) {
+	t.Parallel()
+
+	oktetoPath, err := integration.GetOktetoPath()
+	require.NoError(t, err)
+
+	out, err := commands.RunOktetoKubetoken(oktetoPath, "")
```
teresaromero, line 36: what if the testing server does not include this feature? :/ not sure if this is a possible scenario as this will be always available from 1.27... just wondering

andreafalzetti: there was a similar discussion on: https://github.com/okteto/okteto/pull/3884#pullrequestreview-1562987362

my understanding is that the dynamic token is the default way of getting a token for the user so I don't see how this endpoint can not be there, especially because this has been part of the API for a few versions already.

integration/kubetoken/kubetoken_test.go
teresaromero, line 1: @jLopezbarb should we create a entry at the Makefile to target this tests ? i see there is a `integration` command and then `integration-<pkg>` should adding this new integration be included at ci or makefile config?

teresaromero (approved):
Tested 👍🏻  and working as expected. 

Integration test is failing but i see is related with buildkit, maybe there is smth at staging. Make sure this is transcient. 

👏🏻  great

integration/kubetoken/kubetoken_test.go
ifbyol: Could you review from which backend version we have that endpoint? For the tests It's fine to assume that the endpoint would be available. In normal flows of the CLI we should default to static token if the endpoint is not available. We cannot assume that customers have updated the backend to the latest version and some of them might be running an older version

cmd/context/create.go
```diff
@@ -380,6 +380,17 @@ func (c ContextCommand) getUserContext(ctx context.Context, ns string) (*types.U
 				return nil, fmt.Errorf(oktetoErrors.ErrCorruptedOktetoContexts, config.GetOktetoContextsStorePath())
 			}
 		}
+
+		kubetoken, err := client.Kubetoken().GetKubeToken(okteto.Context().Name, okteto.Context().Namespace)
```
ifbyol, line 384: Would it make sense to execute this query and the one we do to get the credentials from the backend in parallel to try to optimize it and not impact too much on the performance of every command executino?

AdrianPedriza (approved):
working as expected and code looks good!

cmd/context/create.go
andreafalzetti: @ifbyol I've considered but didn't have time to try refactor, I think we're a bit tight this sprint to hold this now. I propose to have a follow up issue perhaps.

jLopezbarb (approved):

integration/kubetoken/kubetoken_test.go
ifbyol: It seems it went out in our chart release `1.6.0`.

cmd/context/create.go
ifbyol: If we do a follow up issue and we release this impacting performance is a worst experience than delaying the feature. This is a new request that is added in almos every command, we have to be careful about it

integration/kubetoken/kubetoken_test.go
andreafalzetti: ooops, I've missed that, I've moved this e2e to a different PR for now, will add the change to the Makefile there

cmd/context/create.go
andreafalzetti: @ifbyol I've already considered the implications and I believe it's better to keep the code simple instead of adding complexity here now. The call to get the kubetoken takes ~100 ms and it's only executing once at the beginning of a command. 

It's more relevant the performance of `okteto kubetoken` because if you call kubectl maby time, it will add up the latency

jmacelroy, line 384: Are we managing the lifecycle of the token appropriately? This is an api call that is potentially expensive on the backend and also blocking if the token cannot be retrieved.

The token should be good for 1d or a length of time. Do we *always* retrieve the token for every command that runs or is the token gone? What are the failure modes for when we possibly could have a good token and use it over getting a new one either b/c it's very new or the api is not working? In these scenarios what does the kubectl lib do or the AKS open source implementation and are we inline with that.

I want to avoid creating a full incident and outage when we only really need to have a partial outage in the case of the api being down for a bit. It's also possible that this is irrelevant b/c we only need this kube token when we're making api calls anyways?

PR #3893, opened 2023-08-08
 Feature Flag to Opt-In to Static Tokens/Existing Flow

andreafalzetti:
# Proposed changes

Fixes https://github.com/okteto/app/issues/6886

This branch requires the following PRs to be merged first:
- https://github.com/okteto/okteto/pull/3891
- https://github.com/okteto/okteto/pull/3887

## Description

Adding feature flag via env var `OKTETO_USE_STATIC_KUBETOKEN`. This allows opt-in for using the static tokens.

## How to test

- Using https://github.com/okteto/movies
- Set `export OKTETO_USE_STATIC_KUBETOKEN=true`
- Run `okteto deploy`
- Notice the warning
- Set `export OKTETO_USE_STATIC_KUBETOKEN=false`
- Run `okteto deploy`
- Check the warning is not displayed anymore

Then, to verify the `okteto kubeconfig` change:

- Back up your `~/.kube/config` (just in case)
- Set `export OKTETO_USE_STATIC_KUBETOKEN=true`
- Run `okteto kubeconfig`
- Check your `~/.kube/config` and notice how you are using a token
- Set `export OKTETO_USE_STATIC_KUBETOKEN=false`
- Run `okteto kubeconfig`
- Check your `~/.kube/config` and notice how you are using the `exec` and the token has been removed

### DIFF
--- a/cmd/context/create.go
+++ b/cmd/context/create.go
@@ -388,7 +388,7 @@ func (c ContextCommand) getUserContext(ctx context.Context, ns string) (*types.U
 
 		kubetoken, err := client.Kubetoken().GetKubeToken(okteto.Context().Name, okteto.Context().Namespace)
 		if err != nil {
-			oktetoLog.Debug("Dynamic kubernetes tokens not available: falling back to static token")
+			oktetoLog.Debug("Dynamic kubernetes token not available: falling back to static token")
 			return userContext, nil
 		}
 
@@ -396,7 +396,7 @@ func (c ContextCommand) getUserContext(ctx context.Context, ns string) (*types.U
 			userContext.Credentials.Token = kubetoken.Status.Token
 		} else {
 			// TODO: when the static token feature gets removed, we must return an error here instead
-			oktetoLog.Debug("Dynamic kubernetes tokens not available: falling back to static token")
+			oktetoLog.Debug("Dynamic kubernetes token not available: falling back to static token")
 		}
 
 		return userContext, nil
--- a/cmd/context/create.go
+++ b/cmd/context/create.go
@@ -383,6 +383,7 @@ func (c ContextCommand) getUserContext(ctx context.Context, ns string) (*types.U
 
 		kubetoken, err := client.Kubetoken().GetKubeToken(okteto.Context().Name, okteto.Context().Namespace)
 		if err != nil {
+			oktetoLog.Debug("Dynamic Kubetoken not available, falling back to static token")
 			return userContext, nil
 		}
 
--- a/cmd/context/create.go
+++ b/cmd/context/create.go
@@ -17,7 +17,6 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"github.com/okteto/okteto/pkg/constants"
 	"net/url"
 	"os"
 	"strings"
@@ -382,7 +381,7 @@ func (c ContextCommand) getUserContext(ctx context.Context, ns string) (*types.U
 			}
 		}
 
-		if utils.LoadBoolean(constants.OktetoUseStaticKubetokenEnvVar) {
+		if utils.LoadBoolean(oktetoUseStaticKubetokenEnvVar) {
 			oktetoLog.Warning(usingStaticKubetokenWarningMessage)
 			return userContext, nil
 		}
--- a/cmd/context/create.go
+++ b/cmd/context/create.go
@@ -381,17 +381,14 @@ func (c ContextCommand) getUserContext(ctx context.Context, ns string) (*types.U
 			}
 		}
 
-		kubetoken, err := client.Kubetoken().GetKubeToken(okteto.Context().Name, okteto.Context().Namespace)
-		if err != nil {
-			if utils.LoadBoolean(model.OktetoUseStaticKubetokenEnvVar) {
-				oktetoLog.Info("You opted-in to use static kubernetes tokens because you set '%s', however the feature is not available and okteto CLI will be using a static token", model.OktetoUseStaticKubetokenEnvVar)
-			}
-			oktetoLog.Debug("Dynamic kubernetes tokens not available: falling back to static token")
+		if utils.LoadBoolean(model.OktetoUseStaticKubetokenEnvVar) {
+			oktetoLog.Debug("You opted-in to use static kubernetes tokens because you set '%s'. This feature will be removed in the future. We recommend using dynamic kubernetes tokens, to know more check out out documentation: https://www.okteto.com/docs", model.OktetoUseStaticKubetokenEnvVar)
 			return userContext, nil
 		}
 
-		if utils.LoadBoolean(model.OktetoUseStaticKubetokenEnvVar) {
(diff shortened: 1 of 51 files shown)


cmd/context/create.go
```diff
@@ -380,6 +380,28 @@ func (c ContextCommand) getUserContext(ctx context.Context, ns string) (*types.U
 				return nil, fmt.Errorf(oktetoErrors.ErrCorruptedOktetoContexts, config.GetOktetoContextsStorePath())
 			}
 		}
+
+		kubetoken, err := client.Kubetoken().GetKubeToken(okteto.Context().Name, okteto.Context().Namespace)
+		if err != nil {
+			if utils.LoadBoolean(model.OktetoUseStaticKubetokenEnvVar) {
+				oktetoLog.Info("You are using static kubernetes tokens because you set '%s', however the feature is not available and okteto CLI will be using a static token", model.OktetoUseStaticKubetokenEnvVar)
```
ifbyol, line 387: I think this message is not right. We are saying that user will use static tokens because they are using the environment variable, but then, we say that the feature is not available and the CLI will use a static token.

Well, isn't that what we want setting that environment variable?

cmd/context/create.go
```diff
@@ -380,6 +380,28 @@ func (c ContextCommand) getUserContext(ctx context.Context, ns string) (*types.U
...
+			}
+			oktetoLog.Debug("Dynamic kubernetes tokens not available: falling back to static token")
+			return userContext, nil
+		}
+
+		if utils.LoadBoolean(model.OktetoUseStaticKubetokenEnvVar) {
+			oktetoLog.Info("You are using static kubernetes tokens because you set '%s'. This feature will be removed in the future. We recommend using dynamic kubernetes tokens, to know more check out out documentation: https://www.okteto.com/docs", model.OktetoUseStaticKubetokenEnvVar)
```
ifbyol, line 394: @rlamana We need to evaluate which documentation we are going to generate and see if this is the right call to action.

Also, I think these messages **won't** be displayed to the user. Is that what we want? Should they be warnings?

cmd/kubetoken/kubetoken.go
```diff
@@ -53,17 +67,26 @@ You can find more information on 'ExecCredential' and 'client side authenticatio
...
 		}
 
-		cmd.Print(out)
+		serializer := &Serializer{
+			KubeToken: out,
+		}
+
+		jsonStr, err := serializer.ToJson()
+		if err != nil {
+			return fmt.Errorf("failed to marshal KubeTokenResponse: %w", err)
```
ifbyol, line 86: Is this logged somewhere?

cmd/context/create_test.go
```diff
@@ -674,6 +703,10 @@ func TestGetUserContext(t *testing.T) {
 				},
 			}
 
+			if tc.useStaticToken {
+				t.Setenv("OKTETO_USE_STATIC_KUBETOKEN", "true")
```
teresaromero, line 707:
```suggestion
				t.Setenv(model.OktetoUseStaticKubetokenEnvVar, "true")
```

pkg/model/const.go
```diff
@@ -263,4 +263,7 @@ const (
 
 	// OktetoImageTagWithVolumes is the tag assigned to an image with volume mounts
 	OktetoImageTagWithVolumes = "okteto-with-volume-mounts"
+
+	// OktetoUseStaticKubetokenEnvVar is used to opt in to use static kubetoken
+	OktetoUseStaticKubetokenEnvVar = "OKTETO_USE_STATIC_KUBETOKEN"
```
teresaromero, line 268: in order to not increase the model pkg, can we perhaps declare this as const under the package that is used?

cmd/context/update-kubeconfig.go
```diff
@@ -62,10 +63,16 @@ func ExecuteUpdateKubeconfig(okCtx *okteto.OktetoContext, kubeconfigPaths []stri
 	if err != nil {
 		return err
 	}
-	if err := okClient.Kubetoken().CheckService(okCtx.Name, okCtx.Namespace); err == nil {
-		updateCfgAuthInfoWithExec(okCtx)
+
+	if !utils.LoadBoolean(model.OktetoUseStaticKubetokenEnvVar) {
```
jLopezbarb, line 67: I would switch this to be. I think it's clearer to read the positive branch first:
```suggestion
	if utils.LoadBoolean(model.OktetoUseStaticKubetokenEnvVar) {
```

cmd/context/create.go
```diff
@@ -381,14 +381,22 @@ func (c ContextCommand) getUserContext(ctx context.Context, ns string) (*types.U
 			}
 		}
 
+		if utils.LoadBoolean(model.OktetoUseStaticKubetokenEnvVar) {
+			oktetoLog.Warning("Using static Kubernetes token due to env var: '%s'. This feature will be removed in the future. We recommend using a dynamic kubernetes token, to know more check out our documentation: https://www.okteto.com/docs", model.OktetoUseStaticKubetokenEnvVar)
```
jLopezbarb, line 385: Could you post an image of how this message looks like in Warning and Information? I think it should be a warning because it's changing the behavior based on a env var but since we wanted to change it on context initialization, let's show both options and confirm with @rlamana

cmd/context/create.go
andreafalzetti: I've changed it to warning and pinged @rlamana too so we can get approval from product

cmd/context/create.go
andreafalzetti: @jLopezbarb 
Screenshot 2023-08-16 at 11 01 48

cmd/context/update-kubeconfig.go
andreafalzetti: good call, done ✅

pkg/model/const.go
andreafalzetti: good call, done ✅

cmd/context/create.go
andreafalzetti: discussed with @rlamana and agreed to remove the link to docs until it's available

cmd/kubetoken/kubetoken.go
andreafalzetti: unfortunatelty not and we don't have a log able to do it as of today :(

AdrianPedriza (approved):
working as expected! LGTM

teresaromero (approved):

pkg/constants/constants.go
```diff
@@ -127,6 +128,10 @@ const (
 
 	//OktetoConfigMapVariablesField represents the field name related to variables seetion in config map
 	OktetoConfigMapVariablesField = "variables"
+
 	//OktetoDependencyEnvsKey the key on the conqfig map that will store OKTETO_ENV values
 	OktetoDependencyEnvsKey = "dependencyEnvs"
+
+	// OktetoUseStaticKubetokenEnvVar is used to opt in to use static kubetoken
+	OktetoUseStaticKubetokenEnvVar = "OKTETO_USE_STATIC_KUBETOKEN"
```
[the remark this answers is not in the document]
andreafalzetti: good call, I made the change here: https://github.com/okteto/okteto/pull/3893/commits/6f8309b58d40993c1222d6894699acab44fc1b67

PR #3899, opened 2023-08-10
Adding syncthing api unittests.

jmacelroy:
Added a test for the proper insertion of headers by the syncthing client roundtripper and the proper initialization of the request timeout for the http client.

### DIFF
--- a/pkg/syncthing/api.go
+++ b/pkg/syncthing/api.go
@@ -30,8 +30,15 @@ type addAPIKeyTransport struct {
 	T http.RoundTripper
 }
 
+const (
+	APIKeyHeader      = "X-Api-Key"
+	APIKeyHeaderValue = "cnd"
+)
+
+// RoundTrip implements the http.RoundTripper interface and is used to add the
+// desired request headers to http requests.
 func (akt *addAPIKeyTransport) RoundTrip(req *http.Request) (*http.Response, error) {
-	req.Header.Add("X-API-Key", "cnd")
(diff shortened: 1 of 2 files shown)


pkg/syncthing/api.go
```diff
@@ -30,8 +30,15 @@ type addAPIKeyTransport struct {
 	T http.RoundTripper
 }
 
+const (
+	APIKeyHeader      = "X-Api-Key"
+	APIKeyHeaderValue = "cnd"
+)
+
+// RoundTrip implements the http.RoundTripper interface and is used to add the
+// desired request headers to http requests.
```
andreafalzetti, line 39: What's the context behind this change? Where does the value `cdn` come from?

jLopezbarb (approved):
LGTM!! Thanks a lot @jmacelroy

pkg/syncthing/api.go
andreafalzetti: I've seen the same API KEY being set when we init syncthing, so it makes more sense now 👍

andreafalzetti (approved):

pkg/syncthing/api.go
jmacelroy: fwiw I do not know the answer to your question on , "why we are sending this header?" The approach was to keep the code doing what it was doing before, but with a test.

I did the overall thing just to go through the flow of a unittest in okteto/okteto while investigating making changes for our approach to test coverage.

Issue #3906, opened 2023-08-13
Logs when building image seem messy and each time I execute the command the output is in different order


ifbyol:
**Describe the bug**
When I execute `okteto build` command I'm getting the logs in different order each time. Check the following screenshot:

<img width="1250" alt="Captura de pantalla 2023-08-13 a las 19 01 43" src="https://github.com/okteto/okteto/assets/3510171/b96110a7-e406-478e-8d84-85f88dd3ecd9">

As you can see the lines: ` ✓  Build succeeded` and ` i  Your image won't be pushed. To push your image specify the flag '-t'.` are displayed sometimes in the middle of build logs but they should be the last logs to be displayed

**To Reproduce**

1. Execute `okteto build` several times. In my case I didn't specify the flag `-t`

**Expected behavior**
The logs should appear in order.

**Desktop (please complete the following information):**

- Version: `2.18.3`

**Additional context**

PR #3908, opened 2023-08-14
fix: wait until build stops logs to return

jLopezbarb:
### Description
This PR fixes a bug while showing the buildkit logs, On previous versions we were waiting until all the logs where on the UI with the stage but we didn't check if the CLI finished them.

### Root Cause
It could happen that the next CLI logs where shown before the CLI ended showing buildkit logs

### Steps to Reproduce (if applicable)
Run several `okteto build`. You'll end up with some build  

### Solution
Wait for the CLI logs instead of the ones that we store with the stage to show them in the UI

### Testing

| Before | After |
|--------|-------|
|  Captura de pantalla 2023-08-13 a las 19 01 43
 |  Captura de pantalla 2023-08-14 a las 17 19 31 |

### Related Issues
Fixes: #3906

### DIFF
--- a/pkg/cmd/build/buildkit.go
+++ b/pkg/cmd/build/buildkit.go
@@ -289,12 +289,12 @@ func solveBuild(ctx context.Context, c *client.Client, opt *client.SolveOpt, pro
 				oktetoLog.Debugf("could not create console from file: %s ", err)
 			}
 			go func() {
-				if err := progressui.DisplaySolveStatus(context.TODO(), "", c, oktetoLog.GetOutputWriter(), ttyChannel); err != nil {
+				if err := progressui.DisplaySolveStatus(context.TODO(), "", nil, w, plainChannel); err != nil {
 					oktetoLog.Infof("could not display solve status: %s", err)
 				}
 			}()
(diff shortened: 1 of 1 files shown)


pkg/cmd/build/buildkit.go
```diff
@@ -289,12 +289,12 @@ func solveBuild(ctx context.Context, c *client.Client, opt *client.SolveOpt, pro
 				oktetoLog.Debugf("could not create console from file: %s ", err)
 			}
 			go func() {
-				if err := progressui.DisplaySolveStatus(context.TODO(), "", c, oktetoLog.GetOutputWriter(), ttyChannel); err != nil {
+				if err := progressui.DisplaySolveStatus(context.TODO(), "", nil, w, plainChannel); err != nil {
```
ifbyol, line 292: Could we include some comment in the code? It's not intuitive that we have to pass the `plainChannel` to a function within a case for `ttyFormat`

teresaromero: +1. Perhaps we can also include why we use tty/plain and now we are swapping

teresaromero (approved):

PR #3909, opened 2023-08-14
feat: add kubetoken pre-reqs validation

jLopezbarb:
# Proposed changes

Fixes https://github.com/okteto/app/issues/6888

Checks if okteto has will be able to run the `okteto kubetoken` command before executing it:
- Adds four different validations:
  - Context checker that checks if the context is one of the ones defined in okteto context config file. It also ensures that the context is an okteto context
  - ns validator: checks that the ns is not empty and that the user has access to it. Since we are not checking the ns in the backend maybe we need to remove it
  - k8s support validation: This checks that the k8s cluster that the user is connected has the `TokenRequest` resource. Maybe this should be checked in the backend and not in the CLI
  - okteto support validation: Checks that the okteto cluster that we are in have the kubetoken endpoint

Edit:
- Removed `ns validator` and `k8s support validation`. These validation should happen in the backend. This way we will at least have logs.

### Invalid context
1. Run `okteto kubetoken -c https://not-my-cluster.com -n jlopezbarb`
2. See failure

### From another context
1. Run `okteto ctx use https://my-cluster.com` Change https://my-cluster.com for any okteto instance
2. Run `okteto ctx use minikube` (or any other context you may have)
3. Run `okteto kubetoken -c https://my-cluster.com -n jlopezbarb`

### DIFF
--- a/cmd/kubetoken/validator_test.go
+++ b/cmd/kubetoken/validator_test.go
@@ -147,6 +147,7 @@ func TestPreReqValidator(t *testing.T) {
 		},
 	}
 	for _, tc := range tt {
+		tc := tc
 		t.Run(tc.name, func(t *testing.T) {
 			v := newPreReqValidator(
 				withCtxName(tc.input.ctxName),


cmd/kubetoken/kubetoken.go
```diff
@@ -17,13 +17,13 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"github.com/okteto/okteto/pkg/types"
 	"os"
 
+	"github.com/okteto/okteto/pkg/okteto"
```
jLopezbarb, line 22: I'll try to remove as much dependency with the pk/okteto since we are getting the interfaces from there, but they should be declared by the file that will use them

cmd/kubetoken/kubetoken.go
```diff
@@ -17,13 +17,13 @@ import (
...
+	"github.com/okteto/okteto/pkg/types"
```
jLopezbarb, line 23: Same with types. I'll define the interfaces that will be used here

cmd/kubetoken/kubetoken.go
```diff
@@ -54,25 +54,36 @@ You can find more information on 'ExecCredential' and 'client side authenticatio
 	cmd.RunE = func(_ *cobra.Command, args []string) error {
 		ctx := context.Background()
 
-		err := contextCMD.NewContextCommand().Run(ctx, &contextCMD.ContextOptions{
+		k8sClientProvider := okteto.NewK8sClientProvider()
+		okClientProvider := okteto.NewOktetoClientProvider()
```
jLopezbarb, line 58: I think this way is more readable to init a struct with several params. If we forgot to set one of them, the default ones will be used so we could remove `withK8sClientProvider(k8sClientProvider)` and the code still works. I think it's a good way of reducing panics. WDYT?

cmd/kubetoken/kubetoken.go
```diff
@@ -54,25 +54,36 @@ You can find more information on 'ExecCredential' and 'client side authenticatio
...
 			Namespace: namespace,
 		})
 		if err != nil {
 			return err
 		}
 
-		octx := okteto.Context()
-		if !octx.IsOkteto {
-			return errors.ErrContextIsNotOktetoCluster
+		if err := newK8sSupportValidator(k8sClientProvider).Validate(ctx); err != nil {
```
jLopezbarb, line 77: We need to download the k8s cfg from the context initialization. If we end up not having this validation it will be easier

pkg/okteto/context.go
```diff
@@ -191,7 +191,10 @@ func K8sContextToOktetoUrl(ctx context.Context, k8sContext, k8sNamespace string,
 
 	// check the namespace label
 	if k8sNamespace == "" {
-		k8sNamespace = cfg.Contexts[k8sContext].Namespace
+		cfgK8sNamespace, exists := cfg.Contexts[k8sContext]
```
jLopezbarb, line 194: This code was throwing a panic. I though about refactoring it a bit by creating a struct and not using global variables. Let me know if you don't think it should be addressed in this PR but since this PR uses this function several times I think is a good refactor to do it

cmd/kubetoken/kubetoken.go
```diff
@@ -54,25 +54,36 @@ You can find more information on 'ExecCredential' and 'client side authenticatio
 	cmd.RunE = func(_ *cobra.Command, args []string) error {
 		ctx := context.Background()
 
-		err := contextCMD.NewContextCommand().Run(ctx, &contextCMD.ContextOptions{
+		k8sClientProvider := okteto.NewK8sClientProvider()
+		okClientProvider := okteto.NewOktetoClientProvider()
+		err := newPreReqValidator(
```
teresaromero, line 59:
```suggestion
		if err := newPreReqValidator(
```

cmd/kubetoken/kubetoken.go
```diff
@@ -54,25 +54,36 @@ You can find more information on 'ExecCredential' and 'client side authenticatio
...
+			withCtxName(contextName),
+			withNamespace(namespace),
+			withK8sClientProvider(k8sClientProvider),
+			withOktetoClientProvider(okClientProvider),
+		).Validate(ctx)
+		if err != nil {
+			return fmt.Errorf("failed to validate pre-reqs: %w", err)
+		}
+
+		err = contextCMD.NewContextCommand().Run(ctx, &contextCMD.ContextOptions{
```
teresaromero, line 69:
```suggestion
		if err := contextCMD.NewContextCommand().Run(ctx, &contextCMD.ContextOptions{
```

cmd/kubetoken/kubetoken.go
```diff
@@ -54,25 +54,36 @@ You can find more information on 'ExecCredential' and 'client side authenticatio
...
 			Context:   contextName,
 			Namespace: namespace,
 		})
 		if err != nil {
 			return err
 		}
```
teresaromero, line 75:
```suggestion
		if err := contextCMD.NewContextCommand().Run(ctx, &contextCMD.ContextOptions{
			Context:   contextName,
			Namespace: namespace,
		}); err != nil {
			return err
		}
```

cmd/kubetoken/kubetoken.go
```diff
@@ -54,25 +54,36 @@ You can find more information on 'ExecCredential' and 'client side authenticatio
...
+		okClientProvider := okteto.NewOktetoClientProvider()
+		err := newPreReqValidator(
+			withCtxName(contextName),
+			withNamespace(namespace),
+			withK8sClientProvider(k8sClientProvider),
+			withOktetoClientProvider(okClientProvider),
+		).Validate(ctx)
+		if err != nil {
+			return fmt.Errorf("failed to validate pre-reqs: %w", err)
+		}
```
teresaromero, line 67:
```suggestion
		if err := newPreReqValidator(
			withCtxName(contextName),
			withNamespace(namespace),
			withK8sClientProvider(k8sClientProvider),
			withOktetoClientProvider(okClientProvider),
		).Validate(ctx); err != nil {
			return fmt.Errorf("failed to validate pre-reqs: %w", err)
		}
```

cmd/kubetoken/validator.go
```diff
@@ -0,0 +1,429 @@
...
+		ctxName:              cfg.ctxName,
+		ns:                   cfg.ns,
+		k8sClientProvider:    cfg.k8sClientProvider,
+		oktetoClientProvider: cfg.oktetoClientProvider,
+	}
+}
+
+// Validate validates that all the pre-reqs to execute the command are met
+func (v *preReqValidator) Validate(ctx context.Context) error {
+	oktetoLog.Info("validating pre-reqs")
```
teresaromero, line 115:
```suggestion
	oktetoLog.Info("validating pre-reqs for kubetoken")
```

could we add a little bit more context at this log? what are pre-reqs for?

pkg/okteto/context.go
```diff
@@ -191,7 +191,10 @@ func K8sContextToOktetoUrl(ctx context.Context, k8sContext, k8sNamespace string,
...
+		if exists {
+			k8sNamespace = cfgK8sNamespace.Namespace
+		}
```
teresaromero, line 197:
```suggestion
		} else{
		return k8sContext
		}
```

can we add here the else condition to return what is checked on line 200-202. If k8sNamespace still is empty would fall into this case. Instead of checking the same condition twice at 200 and 193 ? WDYT?

pkg/okteto/context.go
teresaromero: I think is worth changing it here

cmd/kubetoken/validator.go
```diff
@@ -0,0 +1,429 @@
...
+		oktetoClientProvider: oktetoClientProvider,
+	}
+}
+
+func (c *nsAccessChecker) hasAccess(ctx context.Context, ns string) (bool, error) {
+	okClient, err := c.oktetoClientProvider.Provide()
+	if err != nil {
+		return false, err
+	}
+	nList, err := okClient.Namespaces().List(ctx)
```
AdrianPedriza, line 252: If we know the name of the ns I think we could try to get it instead list all of them

cmd/kubetoken/validator.go
```diff
@@ -0,0 +1,429 @@
...
+		oktetoClientProvider: oktetoClientProvider,
+	}
+}
+
+func (c *previewAccessChecker) hasAccess(ctx context.Context, preview string) (bool, error) {
+	okClient, err := c.oktetoClientProvider.Provide()
+	if err != nil {
+		return false, err
+	}
+	previewList, err := okClient.Previews().List(ctx, []string{})
```
AdrianPedriza, line 280: If we know the name of the preview I think we could try to get it instead list all of them

cmd/kubetoken/validator.go
```diff
@@ -0,0 +1,429 @@
...
+
+	err := newCtxValidator(v.ctxName, v.k8sClientProvider).Validate(ctx)
+	if err != nil {
+		return fmt.Errorf("invalid context: %w", err)
+	}
+
+	var wg sync.WaitGroup
+	validators := []validator{
+		newNsValidator(v.ns, v.oktetoClientProvider),
+		newOktetoSupportValidator(ctx, v.ctxName, v.ns, v.k8sClientProvider, v.oktetoClientProvider),
```
AdrianPedriza, line 128: there is no dependency between ns validator and ctx validator I think. Could both run in parallel and run support validator once ctx validator is finished? even if ns validator is not finished. Something like:

````
ns validator | ctx validator | (if ctx validator is finished) okteto support validator
````

File: cmd/kubetoken/kubetoken.go (hunk 4 above)
Comment (AdrianPedriza) [reply cmd/kubetoken/kubetoken.go:77]: could be added in parallel with others like ns validation or okteto support validation?

Comment (jLopezbarb) [reply cmd/kubetoken/kubetoken.go:77]: We need to get the k8s config to connect to the cluster. This is done at the context init step. My first though was to do it also all together to reduce the time but I can't do it with this one

File: cmd/kubetoken/validator.go (hunk 14 above)
Comment (jLopezbarb) [reply cmd/kubetoken/validator.go:128]: there is. To create the okteto client we need the okteto url which is the name of the context. If the context doesn't exists it will fail

File: cmd/kubetoken/validator.go (hunk 12 above)
Comment (jLopezbarb) [reply cmd/kubetoken/validator.go:252]: prob you're right, but I don't know if this is something that the CLI must do or if it should be handled by the backend

Comment (jLopezbarb) [reply cmd/kubetoken/validator.go:252]: If the backend is the one that fails we'd probably have more logs so I'm deleting this validation

File: cmd/kubetoken/kubetoken.go (hunk 8 above)
Comment (jLopezbarb) [reply cmd/kubetoken/kubetoken.go:75]: It's usually a smell code to have a large if because it reduces visibility, that's why i prefer to have it separate. You can read more about  this [here](https://blog.jetbrains.com/idea/2017/09/code-smells-if-statements/)

File: cmd/kubetoken/kubetoken.go (hunk 9 above)
Comment (jLopezbarb) [reply cmd/kubetoken/kubetoken.go:67]: Same as previous comment

File: cmd/kubetoken/validator.go [hunk 15]
```diff
@@ -0,0 +1,429 @@
...
+				cancel()
+			}
+			errChan <- err
+		}(preReqValidator)
+	}
+
+	wg.Wait()
+	close(errChan)
+
+	for err := range errChan {
```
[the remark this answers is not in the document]
jLopezbarb: If we do that we'll have to wait until all of them finished. We are not returning the errors to the user in the kubetoken command, it just fail and shows the tip that we added in the kubeconfig. So I think it's better to fail fast

teresaromero:
Can we record a quick video with the changes? Some description on how this has been tested?

jLopezbarb:
> Can we record a quick video with the changes? Some description on how this has been tested?

Updated with the tests done. I didn't record a video in case someone may be biased by my video and could think about another test

teresaromero:
i tried the following, and i am getting a failure... not sure this is expected


```
❯ rm -rf ~/.kube/config
❯ rm -rf ~/.okteto
❯ ok kubetoken
 x  Failed to validate pre-reqs: invalid context: context name cannot be empty
❯ ok context use https://my dev cluster
Authentication will continue in your default browser
You can also open a browser and navigate to the following address:
.........
 ✓  Context '<my dev cluster>' created
 ✓  Using teresa @ '<my dev cluster>' 
❯ ok kubetoken
 x  Failed to validate pre-reqs: invalid context: context name cannot be empty
❯ ok kubetoken
 x  Failed to validate pre-reqs: invalid context: context name cannot be empty
❯ ok kubeconfig
 ✓  Updated kubernetes context 'my-dev-cluster/teresa' in '[/Users/teresaromero/.kube/config]'
❯ ok kubetoken
 x  Failed to validate pre-reqs: invalid context: context name cannot be empty
```

jLopezbarb:
> i tried the following, and i am getting a failure... not sure this is expected
> 
> ```
> ❯ rm -rf ~/.kube/config
> ❯ rm -rf ~/.okteto
> ❯ ok kubetoken
>  x  Failed to validate pre-reqs: invalid context: context name cannot be empty
> ❯ ok context use https://my dev cluster
> Authentication will continue in your default browser
> You can also open a browser and navigate to the following address:
> .........
>  ✓  Context '<my dev cluster>' created
>  ✓  Using teresa @ '<my dev cluster>' 
> ❯ ok kubetoken
>  x  Failed to validate pre-reqs: invalid context: context name cannot be empty
> ❯ ok kubetoken
>  x  Failed to validate pre-reqs: invalid context: context name cannot be empty
> ❯ ok kubeconfig
>  ✓  Updated kubernetes context 'my-dev-cluster/teresa' in '[/Users/teresaromero/.kube/config]'
> ❯ ok kubetoken
>  x  Failed to validate pre-reqs: invalid context: context name cannot be empty
> ```

I don't really know, While we were doing the spec for this task we were only thinking about how it was going to be executed (from `kubeconfig exec`) and we are always adding the context and namespace flag since it can happen that you are on a different okteto context than your kubeconfig context. I think it's better to fail instead of getting the current okteto context but I don't have a clear answer to that.
cc: @ifbyol

cmd/kubetoken/kubetoken.go
```diff
@@ -17,20 +17,25 @@ import (
...
-	"github.com/okteto/okteto/pkg/errors"
-	"github.com/okteto/okteto/pkg/okteto"
 	"github.com/spf13/cobra"
 )
 
 type Serializer struct {
 	KubeToken types.KubeTokenResponse
 }
 
+// oktetoClientProvider provides an okteto client ready to use or fail
```
andreafalzetti, line 34: do we still need `oktetoClientProvider` in types? it seems unused now?

cmd/kubetoken/validator.go
```diff
@@ -0,0 +1,238 @@
...
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
+)
+
+var (
+	errEmptyNamespace = errors.New("namespace cannot be empty")
+	errEmptyContext   = errors.New("context name cannot be empty")
+
+	valdationTimeout = 30000 * time.Second
```
andreafalzetti, line 34:
```suggestion
	validationTimeout = 30000 * time.Second
```

cmd/kubetoken/validator.go
```diff
@@ -0,0 +1,238 @@
...
+
+	oktetoLog "github.com/okteto/okteto/pkg/log"
+	"github.com/okteto/okteto/pkg/okteto"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
+)
+
+var (
+	errEmptyNamespace = errors.New("namespace cannot be empty")
```
andreafalzetti, line 31: `errEmptyNamespace` is not used, the check on the namespace works because an empty one returns not found anyway, but perhaps it would help have an explicit error msg, or shall we remove this var?

cmd/kubetoken/validator.go
```diff
@@ -0,0 +1,238 @@
...
+
+	valdationTimeout = 30000 * time.Second
+)
+
+type k8sClientProvider interface {
+	Provide(clientApiConfig *clientcmdapi.Config) (kubernetes.Interface, *rest.Config, error)
+}
+
+// validator is the interface that wraps the Validate method.
+type validator interface {
```
andreafalzetti, line 42: do we use this type anywhere?

cmd/kubetoken/kubetoken.go
jLopezbarb: Yes, We are using it in the validation to check that the okteto client has the kubetoken API endpoint

cmd/kubetoken/validator.go
jLopezbarb: We were using it when we were executing it in parallel but we don't need it anymore :)

cmd/kubetoken/validator.go
```diff
@@ -0,0 +1,232 @@
...
+	defer cancel()
+
+	err := newCtxValidator(v.ctxName, v.k8sClientProvider).Validate(ctx)
+	if err != nil {
+		return fmt.Errorf("invalid context: %w", err)
+	}
+
+	err = newOktetoSupportValidator(ctx, v.ctxName, v.ns, v.k8sClientProvider, v.oktetoClientProvider).Validate(ctx)
+	if err != nil {
+		return fmt.Errorf("invalid okteto support: %w", err)
```
andreafalzetti, line 118: the fact that the CLI is validating before making the actual request could be a debug log, wdyt?

cmd/kubetoken/validator.go
```diff
@@ -0,0 +1,232 @@
...
+		if v.ctxName == "" {
+			result <- errEmptyContext
+			return
+		}
+		if !isURL(v.ctxName) {
+			v.ctxName = v.k8sCtxToOktetoURL(ctx, v.ctxName, "", v.k8sClientProvider)
+		}
+		okCtx, exists := v.getContextStore().Contexts[v.ctxName]
+		if !exists {
+			result <- errOktetoContextNotFound{v.ctxName}
```
andreafalzetti, line 170: what do you think about making this log more explicit? something like:

```suggestion
	oktetoLog.Info("validating context for dynamic kubernetes token request")
```

cmd/kubetoken/validator.go
```diff
@@ -0,0 +1,232 @@
...
+	err = newOktetoSupportValidator(ctx, v.ctxName, v.ns, v.k8sClientProvider, v.oktetoClientProvider).Validate(ctx)
+	if err != nil {
+		return fmt.Errorf("invalid okteto support: %w", err)
+	}
+	return nil
+}
+
+// ctxValidator checks that the ctx use to execute the command is an okteto context
+// that has already being added to your okteto context
+type ctxValidator struct {
```
andreafalzetti, line 125: NIT: in case of timeout it would be more correct to say something like:

```suggestion
		return fmt.Errorf("context validation failed: %w", err)
```

cmd/kubetoken/kubetoken.go
andreafalzetti: I mean `types.oktetoClientProvider`

andreafalzetti (approved):

cmd/kubetoken/validator.go
andreafalzetti: this comment is in regards to: `oktetoLog.Info("validating the context")`

cmd/kubetoken/validator.go
```diff
@@ -0,0 +1,232 @@
...
+	return &preReqValidator{
+		ctxName:              cfg.ctxName,
+		ns:                   cfg.ns,
+		k8sClientProvider:    cfg.k8sClientProvider,
+		oktetoClientProvider: cfg.oktetoClientProvider,
+	}
+}
+
+// Validate validates that all the pre-reqs to execute the command are met
+func (v *preReqValidator) Validate(ctx context.Context) error {
```
AdrianPedriza, line 105: we dont need to export `Validate` func I think

cmd/kubetoken/validator.go
```diff
@@ -0,0 +1,232 @@
...
+
+type errIsNotOktetoCtx struct {
+	ctxName string
+}
+
+func (e errIsNotOktetoCtx) Error() string {
+	return fmt.Sprintf("context '%s' is not an okteto context", e.ctxName)
+}
+
+func (v *ctxValidator) Validate(ctx context.Context) error {
```
AdrianPedriza, line 157: we dont need to export Validate func I think

cmd/kubetoken/validator.go
```diff
@@ -0,0 +1,232 @@
...
+		ctxName = okteto.K8sContextToOktetoUrl(ctx, ctxName, "", k8sClientProvider)
+	}
+	return &oktetoSupportValidator{
+		ctxName:              ctxName,
+		ns:                   ns,
+		oktetoClientProvider: oktetoClientProvider,
+	}
+}
+
+func (v *oktetoSupportValidator) Validate(ctx context.Context) error {
```
AdrianPedriza, line 205: we dont need to export Validate func I think

ifbyol:
> > i tried the following, and i am getting a failure... not sure this is expected
> > ```
> > ❯ rm -rf ~/.kube/config
> > ❯ rm -rf ~/.okteto
> > ❯ ok kubetoken
> >  x  Failed to validate pre-reqs: invalid context: context name cannot be empty
> > ❯ ok context use https://my dev cluster
> > Authentication will continue in your default browser
> > You can also open a browser and navigate to the following address:
> > .........
> >  ✓  Context '<my dev cluster>' created
> >  ✓  Using teresa @ '<my dev cluster>' 
> > ❯ ok kubetoken
> >  x  Failed to validate pre-reqs: invalid context: context name cannot be empty
> > ❯ ok kubetoken
> >  x  Failed to validate pre-reqs: invalid context: context name cannot be empty
> > ❯ ok kubeconfig
> >  ✓  Updated kubernetes context 'my-dev-cluster/teresa' in '[/Users/teresaromero/.kube/config]'
> > ❯ ok kubetoken
> >  x  Failed to validate pre-reqs: invalid context: context name cannot be empty
> > ```
> 
> I don't really know, While we were doing the spec for this task we were only thinking about how it was going to be executed (from `kubeconfig exec`) and we are always adding the context and namespace flag since it can happen that you are on a different okteto context than your kubeconfig context. I think it's better to fail instead of getting the current okteto context but I don't have a clear answer to that. cc: @ifbyol

Sorry but I'm not getting which is the scenario failing here. Why is it failing after executing `okteto context use <>`? After that command the context should have been read from the current context, right?

teresaromero:
> > > i tried the following, and i am getting a failure... not sure this is expected
> > > ```
> > > ❯ rm -rf ~/.kube/config
> > > ❯ rm -rf ~/.okteto
> > > ❯ ok kubetoken
> > >  x  Failed to validate pre-reqs: invalid context: context name cannot be empty
> > > ❯ ok context use https://my dev cluster
> > > Authentication will continue in your default browser
> > > You can also open a browser and navigate to the following address:
> > > .........
> > >  ✓  Context '<my dev cluster>' created
> > >  ✓  Using teresa @ '<my dev cluster>' 
> > > ❯ ok kubetoken
> > >  x  Failed to validate pre-reqs: invalid context: context name cannot be empty
> > > ❯ ok kubetoken
> > >  x  Failed to validate pre-reqs: invalid context: context name cannot be empty
> > > ❯ ok kubeconfig
> > >  ✓  Updated kubernetes context 'my-dev-cluster/teresa' in '[/Users/teresaromero/.kube/config]'
> > > ❯ ok kubetoken
> > >  x  Failed to validate pre-reqs: invalid context: context name cannot be empty
> > > ```
> > 
> > 
> > I don't really know, While we were doing the spec for this task we were only thinking about how it was going to be executed (from `kubeconfig exec`) and we are always adding the context and namespace flag since it can happen that you are on a different okteto context than your kubeconfig context. I think it's better to fail instead of getting the current okteto context but I don't have a clear answer to that. cc: @ifbyol
> 
> Sorry but I'm not getting which is the scenario failing here. Why is it failing after executing `okteto context use <>`? After that command the context should have been read from the current context, right?

Scenario is:
- remove both files kubeconfig and okteto config
- run okteto kubetoken - fail because no context
- run okteto context use - ok, context selected
- run okteto kubetoken - fail because empty context
- run okteto kubeconfig - ok, configres kubeconfig
- run okteto kubetoken - fails because empty context

i feel that running kubetoken after configuring context and kubeconfig should not fail¿ that is my concern
-

ifbyol:
> > > > i tried the following, and i am getting a failure... not sure this is expected
> > > > ```
> > > > ❯ rm -rf ~/.kube/config
> > > > ❯ rm -rf ~/.okteto
> > > > ❯ ok kubetoken
> > > >  x  Failed to validate pre-reqs: invalid context: context name cannot be empty
> > > > ❯ ok context use https://my dev cluster
> > > > Authentication will continue in your default browser
> > > > You can also open a browser and navigate to the following address:
> > > > .........
> > > >  ✓  Context '<my dev cluster>' created
> > > >  ✓  Using teresa @ '<my dev cluster>' 
> > > > ❯ ok kubetoken
> > > >  x  Failed to validate pre-reqs: invalid context: context name cannot be empty
> > > > ❯ ok kubetoken
> > > >  x  Failed to validate pre-reqs: invalid context: context name cannot be empty
> > > > ❯ ok kubeconfig
> > > >  ✓  Updated kubernetes context 'my-dev-cluster/teresa' in '[/Users/teresaromero/.kube/config]'
> > > > ❯ ok kubetoken
> > > >  x  Failed to validate pre-reqs: invalid context: context name cannot be empty
> > > > ```
> > > 
> > > 
> > > I don't really know, While we were doing the spec for this task we were only thinking about how it was going to be executed (from `kubeconfig exec`) and we are always adding the context and namespace flag since it can happen that you are on a different okteto context than your kubeconfig context. I think it's better to fail instead of getting the current okteto context but I don't have a clear answer to that. cc: @ifbyol
> > 
> > 
> > Sorry but I'm not getting which is the scenario failing here. Why is it failing after executing `okteto context use <>`? After that command the context should have been read from the current context, right?
> 
> Scenario is:
> 
> * remove both files kubeconfig and okteto config
> * run okteto kubetoken - fail because no context
> * run okteto context use - ok, context selected
> * run okteto kubetoken - fail because empty context
> * run okteto kubeconfig - ok, configres kubeconfig
> * run okteto kubetoken - fails because empty context
> 
> ## i feel that running kubetoken after configuring context and kubeconfig should not fail¿ that is my concern

Kubetoken after running `okteto context use <>` should work as the context should be set. @jLopezbarb Why is ti failing?

teresaromero (approved):
👍🏻 retested

jLopezbarb:
> > > > > i tried the following, and i am getting a failure... not sure this is expected
> > > > > ```
> > > > > ❯ rm -rf ~/.kube/config
> > > > > ❯ rm -rf ~/.okteto
> > > > > ❯ ok kubetoken
> > > > >  x  Failed to validate pre-reqs: invalid context: context name cannot be empty
> > > > > ❯ ok context use https://my dev cluster
> > > > > Authentication will continue in your default browser
> > > > > You can also open a browser and navigate to the following address:
> > > > > .........
> > > > >  ✓  Context '<my dev cluster>' created
> > > > >  ✓  Using teresa @ '<my dev cluster>' 
> > > > > ❯ ok kubetoken
> > > > >  x  Failed to validate pre-reqs: invalid context: context name cannot be empty
> > > > > ❯ ok kubetoken
> > > > >  x  Failed to validate pre-reqs: invalid context: context name cannot be empty
> > > > > ❯ ok kubeconfig
> > > > >  ✓  Updated kubernetes context 'my-dev-cluster/teresa' in '[/Users/teresaromero/.kube/config]'
> > > > > ❯ ok kubetoken
> > > > >  x  Failed to validate pre-reqs: invalid context: context name cannot be empty
> > > > > ```
> > > > 
> > > > 
> > > > I don't really know, While we were doing the spec for this task we were only thinking about how it was going to be executed (from `kubeconfig exec`) and we are always adding the context and namespace flag since it can happen that you are on a different okteto context than your kubeconfig context. I think it's better to fail instead of getting the current okteto context but I don't have a clear answer to that. cc: @ifbyol
> > > 
> > > 
> > > Sorry but I'm not getting which is the scenario failing here. Why is it failing after executing `okteto context use <>`? After that command the context should have been read from the current context, right?
> > 
> > 
> > Scenario is:
> > 
> > * remove both files kubeconfig and okteto config
> > * run okteto kubetoken - fail because no context
> > * run okteto context use - ok, context selected
> > * run okteto kubetoken - fail because empty context
> > * run okteto kubeconfig - ok, configres kubeconfig
> > * run okteto kubetoken - fails because empty context
> > 
> > ## i feel that running kubetoken after configuring context and kubeconfig should not fail¿ that is my concern
> 
> Kubetoken after running `okteto context use <>` should work as the context should be set. @jLopezbarb Why is ti failing?

The validation was done with the context and namespace used from the flags. This prevented to initialize those variables from `okteto.context` or from env vars. With the new changes that I've just submitted this problem should be fixed since we are checking what could happend with the initialisation. We couldn't leave the validation after the `context.Run` because it could happen to have a selector.

ifbyol:
> The validation was done with the context and namespace used from the flags. This prevented to initialize those variables from `okteto.context` or from env vars. With the new changes that I've just submitted this problem should be fixed since we are checking what could happend with the initialisation. We couldn't leave the validation after the `context.Run` because it could happen to have a selector.

Ok, thanks for the information. Sounds good the changes you describe

PR #3910, opened 2023-08-14
Add cache to registry credentials

maroshii:
Adds a global cache to registry credentials. Prior to this change, we were making unnecessary roundtrips to the okteto backend to retrieve temporary credentials. We now cache these credentials during the lifecycle of the build/deploy and they are reused.

This was part of the original implementation of buildkit's [default auth provider](https://github.com/moby/buildkit/blob/master/session/auth/authprovider/authprovider.go#L230) but was not included as part of the okteto implementation.

We could not add the cache to the authprovider as in the implementation above because we use these credentials also for the registry api and the docker credentials helper. 

The trade-off for this is adding global state to the execution but there aren't too many options given the state of things.

### DIFF
--- a/pkg/okteto/registry_credentials.go
+++ b/pkg/okteto/registry_credentials.go
@@ -12,10 +12,15 @@ import (
 	oktetoLog "github.com/okteto/okteto/pkg/log"
 )
 
-var globalRegistryCache registryCache
+var globalRegistryCredentialsCache registryCache
+
+type registryCacheItem struct {
+	user string
+	pass string
+}
 
 type registryCache struct {
-	cache map[string][2]string
+	cache map[string]registryCacheItem
 	m     sync.RWMutex
 }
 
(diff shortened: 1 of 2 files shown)


pkg/okteto/registry_credentials.go
```diff
@@ -5,15 +5,56 @@ import (
...
+	defer rc.m.Unlock()
+
+	if rc.cache == nil {
+		rc.cache = make(map[string][2]string)
+	}
+
+	rc.cache[host] = [2]string{user, pass}
+}
+
+func (rc *registryCache) Delete(host string) {
```
jLopezbarb, line 47: Do we need this? I see that they are only used in tests

pkg/okteto/registry_credentials.go
```diff
@@ -5,15 +5,56 @@ import (
...
 	"time"
 
 	dockertypes "github.com/docker/cli/cli/config/types"
 	oktetoLog "github.com/okteto/okteto/pkg/log"
 )
 
+var globalRegistryCache registryCache
+
+type registryCache struct {
+	cache map[string][2]string
```
jLopezbarb, line 18: NIT: Could we create a custom type instead of having `[2]string`. I think it'd be more readable if we have a custom type `registryCacheCredentials` with User and pass. WDYT?

maroshii:
All PR feedback addressed

teresaromero (approved):

jLopezbarb (approved):

PR #3919, opened 2023-08-18
refactor: restore previewLabels at default view of preview list

teresaromero:
# Proposed changes

When running `okteto preview list` and the preview had labels, no labels where shown at the default view

```
❯ okteto preview list
 i  Using - @ - as context
Name                   Scope     Sleeping  Labels
upbeat-feynman-teresa  personal  false     -
```

```
❯ ok preview list -o json
[
 {
  "name": "upbeat-feynman-teresa",
  "scope": "personal",
  "sleeping": false,
  "labels": [
   "other",
   "testing"
  ]
 }
]
```

```
❯ ok preview list -o yaml
- name: upbeat-feynman-teresa
  scope: personal
  sleeping: false
  labels:
  - testing
  - other
```

This was not happening when using `yaml` or `json` output

This bug was [introduced](https://github.com/okteto/okteto/pull/3804/files#diff-1de660ef246a78c71aa7be984185a6ab279c12c708e9d6e27799b2a4e0278ddaL79) when adding the output flag to the command, this PR reverts the logic that added the preview labels to the default view.

## Test

- run `okteto preview list` and see the labels are shown

### DIFF
--- a/cmd/preview/list.go
+++ b/cmd/preview/list.go
@@ -17,6 +17,10 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"os"
+	"strings"
+	"text/tabwriter"
+
 	contextCMD "github.com/okteto/okteto/cmd/context"
 	oktetoErrors "github.com/okteto/okteto/pkg/errors"
 	oktetoLog "github.com/okteto/okteto/pkg/log"
@@ -24,8 +28,6 @@ import (
 	"github.com/okteto/okteto/pkg/types"
 	"github.com/spf13/cobra"
 	"gopkg.in/yaml.v2"
-	"os"
-	"text/tabwriter"
 )
 
 // ListFlags are the flags available for list commands
@@ -114,6 +116,9 @@ func executeListPreviews(ctx context.Context, opts ListFlags) error {
 		fmt.Fprintf(w, "Name\tScope\tSleeping\tLabels\n")
 		for _, preview := range previewList {
 			previewLabels := "-"
+			if len(preview.PreviewLabels) > 0 {
+				previewLabels = strings.Join(preview.PreviewLabels, ", ")
+			}
 			fmt.Fprintf(w, "%s\t%s\t%v\t%s\n", preview.ID, preview.Scope, preview.Sleeping, previewLabels)
 		}
 		w.Flush()


teresaromero:
Bug - regression to 2.19. Adding label to patch 2.19 with this fix

jLopezbarb (commented):
Should we create a test checking that this won't happen again?
Maybe we could encapsulate https://github.com/okteto/okteto/pull/3919/files#diff-1de660ef246a78c71aa7be984185a6ab279c12c708e9d6e27799b2a4e0278ddaR115-R124 in order to make it testable. WDYT?

teresaromero:
> Should we create a test checking that this won't happen again? Maybe we could encapsulate https://github.com/okteto/okteto/pull/3919/files#diff-1de660ef246a78c71aa7be984185a6ab279c12c708e9d6e27799b2a4e0278ddaR115-R124 in order to make it testable. WDYT?

@jLopezbarb I've added some refactor in order to wrap the logic that returns the string of the row to make it testable. Not sure the approach is the best... but i was having doubts on how to test the output 🤷🏻‍♀️ . let me know what you think!

jLopezbarb:
> > Should we create a test checking that this won't happen again? Maybe we could encapsulate https://github.com/okteto/okteto/pull/3919/files#diff-1de660ef246a78c71aa7be984185a6ab279c12c708e9d6e27799b2a4e0278ddaR115-R124 in order to make it testable. WDYT?
> 
> @jLopezbarb I've added some refactor in order to wrap the logic that returns the string of the row to make it testable. Not sure the approach is the best... but i was having doubts on how to test the output 🤷🏻‍♀️ . let me know what you think!

That's a good test, The function only does one thing (translate a preview to table) and we can catch if the error arises again. Nice job!!

cmd/preview/list_test.go
```diff
@@ -142,3 +148,41 @@ func Test_PreviewListOutputValidation(t *testing.T) {
...
+			},
+			expected: "my-preview\tpersonal\tfalse\t-\n",
+		},
+		{
+			name: "preview with labels",
+			input: previewOutput{
+				Name:     "my-preview",
+				Scope:    "personal",
+				Sleeping: false,
+				Labels: map[int]string{
```
jLopezbarb, line 173: Run tests several times because it can happen that the order is not the same. Check https://go.dev/blog/maps `Iteration order`

jLopezbarb (approved):
Check map order, The test can be flaky because of that, but other than that it seems good to me

andreafalzetti (approved):
Code changes LGTM, tested and works as expected!

PR #3925, opened 2023-08-20
Keda support. Pause/unpause ScaledObject

alt-dima:
# Proposed changes

Added support for Keda autoscaler.
With option in Okteto manifest `keda = true` it will pause to 0 ScaledObject with the same name as Dev in time of `okteto up` and unpause in time of `okteto down`

This is needed if some one use Okteto with Kubernetes and Keda autoscaler.
Because after okteto up deployment in k8s will be scaled to 0, but ScaledObject/HPA will scale back to the min number of prods.

Fixes # (issue)


jLopezbarb:
Hi @alt-dima,

Thanks so much for this contribution! We really like the premise of this. We think implementing this feature in a more generic way could be more impactful to our broader customer and user base. As it's written now this would only apply to anyone using Keda.

Would you be open to sharing more detail about your use case so we can figure out what a path forward looks like for this type of feature? We're happy to connect directly with you to chat through your use case(s).

alt-dima:
@jLopezbarb Hi! For sure we could close this PR, because it is exactly not "generic" (maybe other customers have other ScaledObject names for example). I just tried to show the demand of this feature and my "Go skills" :) 

About our specific case. We use Keda autoscaler (to scale up/down our workloads based on CPU, Kafka, rabbitmq).
In the current implementation Okteto sets replicas = 0 in the Deployment.
But Keda/ScaledObject/HPA manages exactly replicas in Deployment and it will bring back original container (and we do not want this, because we want to stop all the original containers and start only development container to handle all the messages/requests).
Please check: https://keda.sh/docs/2.11/concepts/scaling-deployments/#pause-autoscaling

jLopezbarb:
> @jLopezbarb Hi! For sure we could close this PR, because it is exactly not "generic" (maybe other customers have other ScaledObject names for example). I just tried to show the demand of this feature and my "Go skills" :)
> 
> About our specific case. We use Keda autoscaler (to scale up/down our workloads based on CPU, Kafka, rabbitmq). In the current implementation Okteto sets replicas = 0 in the Deployment. But Keda/ScaledObject/HPA manages exactly replicas in Deployment and it will bring back (overwrite replicas in Deployment) original container (and we do not want this, because we want to stop all the original containers and start only development container to handle all the messages/requests). Please check: https://keda.sh/docs/2.11/concepts/scaling-deployments/#pause-autoscaling

I've send your use case to our product team and they are investigating a more generic way of achieving your use case. We'll let you know when we have decided the solution and when it will be ready.

BTW, your code was pretty good!! 😄

jLopezbarb:
Hey again @alt-dima, Thanks for sharing that context!

We've been thinking about adding a new feature that will allow the users to have their own custom hooks. This feature enables users to define and execute scripts or commands as part of the development workflow in Okteto. These hooks could be specified in the Okteto manifest are triggered automatically at specific events during the development lifecycle.

For example, in your case your okteto manifest would look something like this (naming needs to be discussed):
```yaml
dev:
  app:
    hooks:
      preUp:
        - command: kubectl patch scaledobject my-app -p '{"metadata": {"annotations": {"autoscaling.keda.sh/paused-replicas": "0"}}}'
      postDown:
        - command: kubectl patch scaledobject my-app -p '{"metadata": {"annotations": {"autoscaling.keda.sh/paused-replicas": null}}}'
```

Would that fit your workflow and address the challenge here?

PR #3928, opened 2023-08-22
Fix kubetoken remote deploy

teresaromero:
# Proposed changes

Fixes https://github.com/okteto/app/issues/7039

- Updating the token to a dynamic one was done at `getUserContext` where namespace could be empty.
- in order to retrieve the kubetoken we need the namespace, if not the url will return 404 and fallback would be static token.
- new function `updateDynamicTokenForNamespace` wrapping the code that was already in place for updating `userContext.Credentials.Token`
- use `updateDynamicTokenForNamespace` at `initOktetoContext` so the token is updated when namespace has been cleared
- return `getUserContext` to its original behavour

### DIFF
--- a/cmd/context/create.go
+++ b/cmd/context/create.go
@@ -264,7 +264,7 @@ func (c *ContextCommand) initOktetoContext(ctx context.Context, ctxOptions *Cont
 	}
 
 	// once we have namespace and user identify we are able to retrieve the dynamic token for the namespace
-	getAndUseDynamicKubetoken(c.OktetoClientProvider, userContext)
+	replaceCredentialsTokenWithDynamicKubetoken(c.OktetoClientProvider, userContext)
 
 	okteto.AddOktetoContext(ctxOptions.Context, &userContext.User, ctxOptions.Namespace, userContext.User.Namespace)
 	cfg := kubeconfig.Get(config.GetKubeconfigPath())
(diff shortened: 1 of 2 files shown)


cmd/context/create.go
```diff
@@ -279,6 +283,36 @@ func (c *ContextCommand) initOktetoContext(ctx context.Context, ctxOptions *Cont
 	return nil
 }
 
+// updateDynamicTokenForNamespace retrieves a dinamic token for the given userContext and updates Credentials.Token
+// if error while retrieving the dynamic token or flag OKTETO_USE_STATIC_KUBETOKEN is true, value is not updated
+// static token is fallback
+func updateDynamicTokenForNamespace(okClientProvider types.OktetoClientProvider, userContext *types.UserContext) {
```
andreafalzetti, line 289: non-blocking suggestion:

```suggestion
func getAndUseDynamicKubetoken(okClientProvider types.OktetoClientProvider, userContext *types.UserContext) {
```

- the `update` part makes me think of a actual update of the token rather than "overriding" the static token
- the `ForNamespace` part can probably be omitted? Considering that the token is for a combination of Context and Namespace

cmd/context/create.go
```diff
@@ -279,6 +283,36 @@ func (c *ContextCommand) initOktetoContext(ctx context.Context, ctxOptions *Cont
 	return nil
 }
 
+// updateDynamicTokenForNamespace retrieves a dinamic token for the given userContext and updates Credentials.Token
```
andreafalzetti, line 286:
```suggestion
// updateDynamicTokenForNamespace retrieves a dynamic token for the given userContext and updates Credentials.Token
```

andreafalzetti (approved):

jLopezbarb (approved):

PR #3929, opened 2023-08-22
handle 423 api error a trial license expiration

maroshii:
# Proposed changes

Gracefully handle 423 Locked return status from the api. Once an okteto trial license has expired, the okteto API will return an http `423 Locked` status code indicating that the instance is locked. As part of this PR, we now handle this returns status and show the appropriate message.

Note that this change is backwards compatible, doesn't require any specific backend version and can be merge at any time.

### DIFF
--- a/pkg/okteto/client.go
+++ b/pkg/okteto/client.go
@@ -302,7 +302,7 @@ func translateAPIErr(err error) error {
 		case isAPITrialExpiredError(err):
 			return oktetoErrors.UserError{
 				E:    oktetoErrors.ErrTrialExpired,
-				Hint: fmt.Sprintf("This command-line tool will no longer work. Please contact our sales team (sales@okteto.com) to obtain a new license key for %s", Context().Name),
+				Hint: "This command-line tool will no longer work. Please contact our sales team (sales@okteto.com) to obtain a new license key",
 			}
 		}
 
--- a/pkg/okteto/client.go
+++ b/pkg/okteto/client.go
@@ -299,7 +299,7 @@ func translateAPIErr(err error) error {
 				E:    err,
 				Hint: oktetoErrors.ErrX509Hint,
 			}
-		case strings.HasPrefix(err.Error(), "non-200 OK status code: 423"):
+		case isAPITrialExpiredError(err):
 			return oktetoErrors.UserError{
 				E:    oktetoErrors.ErrTrialExpired,
 				Hint: fmt.Sprintf("This command-line tool will no longer work. Please contact our sales team (sales@okteto.com) to obtain a new license key for %s", Context().Name),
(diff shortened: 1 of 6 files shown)


ifbyol:
@maroshii Could you provide some screenshot on how the error looks like when it's triggered, please?

andreafalzetti:
@maroshii How can I modify my license to simulate an expiration and get this error?

maroshii:
@andreafalzetti Backend part is not yet ready, still WIP. The CLI<>Backend contract will be that whenever the backend returns a 423 the CLI will understand it as a license expired event. To reproduce this you can run the backend and add a middleware that returns 423 for all requests

pkg/okteto/client.go
```diff
@@ -258,6 +264,10 @@ func translateAPIErr(err error) error {
 
 }
 
+func isAPITrialExpiredError(err error) bool {
+	return strings.HasPrefix(err.Error(), "non-200 OK status code: 423")
```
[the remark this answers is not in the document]
maroshii: Yes. It is awful but is the exact prefix as returned by the library: https://github.com/shurcooL/graphql/blob/master/graphql.go#L79. For other errors we are just searching for a specific string but I think that is much more error prone

jLopezbarb (approved):

pkg/okteto/client.go
```diff
@@ -245,11 +245,17 @@ func translateAPIErr(err error) error {
...
+		switch {
+		case oktetoErrors.IsX509(err):
 			return oktetoErrors.UserError{
 				E:    err,
 				Hint: oktetoErrors.ErrX509Hint,
 			}
+		case isAPITrialExpiredError(err):
+			return oktetoErrors.UserError{
+				E:    oktetoErrors.ErrTrialExpired,
+				Hint: "This command-line tool will no longer work. Please contact our sales team (sales@okteto.com) to obtain a new license key",
```
AdrianPedriza, line 257: it might not be confusing for the user to say `This command-line tool will no longer work.`?. For vanilla clusters the tool would still work

teresaromero: Maybe the copy should mention "for this context" or "for this instance" as the CLI needs to log to some context, also for cloud will still work i guess

maroshii: Yeah, I think that's a good callout @AdrianPedriza. We might need to reword this cc @ifbyol

ifbyol: @omnomagonz this is related to the error message we talked about to make it a more generic. What are your thoughts about the point raised here?

omnomagonz: From [this Slack thread](https://okteto.slack.com/archives/C043X0H48BZ/p1693224716081279), I believe we landed on:

> The Okteto instance for your current context has an expired or missing license. Please contact your administrator for more information.

PR #3930, opened 2023-08-22
fix: show error for `okteto deploy` when no manifest file exists using `-f` flag 

### LINKED ISSUE #3869: "okteto deploy -f non-existing.yaml" doesnt fail instead it searches for okteto.yaml
**Describe the bug**
When running "okteto deploy -f non-existing.yaml" the command doesn't fail. Instead, of `non-existing.yaml` doesn't exist, it runs `okteto deploy` with `okteto.yaml`.

**To Reproduce**

1. On a repo with an `okteto.yaml` run `okteto deploy -f foo.yaml`

**Expected behavior**
Command should wail with `non-existing.yaml file doesn't exist`

- Version 2.18.0

nitishfy:
# Proposed changes

Fixes #3869

Screenshot from 2023-08-21 23-54-47

### DIFF
--- a/cmd/deploy/deploy.go
+++ b/cmd/deploy/deploy.go
@@ -118,7 +118,9 @@ func NewDeployExternalK8sControl(cfg *rest.Config) ExternalResourceInterface {
 // Deploy deploys the okteto manifest
 func Deploy(ctx context.Context) *cobra.Command {
 	options := &Options{}
-
+	fs := &DeployCommand{
+		Fs: afero.NewOsFs(),
+	}
 	cmd := &cobra.Command{
 		Use:   "deploy [service...]",
 		Short: "Execute the list of commands specified in the 'deploy' section of your okteto manifest",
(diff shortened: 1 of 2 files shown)


nitishfy:
cc: @jLopezbarb @AdrianPedriza

jLopezbarb (approved):
Tested and code looks good. Nice addition with code coverage!!!

cmd/deploy/deploy.go
```diff
@@ -623,3 +608,32 @@ func (dc *DeployCommand) trackDeploy(manifest *model.Manifest, runInRemoteFlag b
 		IsRemote:               isRunningOnRemoteDeployer,
 	})
 }
+
+func checkOktetoManifestPathFlag(options *Options) error {
+	if options.ManifestPath != "" {
```
andreafalzetti, line 613: Exiting early improves overall readability, I would do:

```go
if options.ManifestPath == "" {
	return nil
}
```

cmd/deploy/deploy_test.go
```diff
@@ -1165,3 +1167,44 @@ func TestTrackDeploy(t *testing.T) {
...
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			dir, err := os.Getwd()
+			assert.NoError(t, err)
+			fullpath := filepath.Join(dir, tt.manifest)
+			opts.ManifestPath = fullpath
+			if tt.manifest != "nonexistent.yml" {
+				// create the manifest file only if it's not the nonexistent scenario
+				f, err := os.Create(fullpath)
```
andreafalzetti, line 1198: This will create an actual file in the host running the tests.

It's best practice that all automated test have no side effects on the host. In this case the test file is not cleaned so it can be considered "garbage" for the host.

In other parts of the CLI we use `afero` to create virtual filesystems which makes testing much simpler because you don't have to worry about this cleanup for example. That would be the ideal approach.

If that cannot be implemented for any reason, I would make sure to use a tmp directory and cleanup on completion.

andreafalzetti:
Just a reminder, that who is merging this PR, please create the issues for `destroy` and `endpoints`, context: https://github.com/okteto/okteto/issues/3869#issuecomment-1710328605

cmd/deploy/deploy.go
nitishfy: when the manifest path is not set, why are we returning the error as nil? shouldn't we return an error something like "the manifest path doesn't exist" ?

andreafalzetti: @nitishfy sorry for the late reply, absolutely, I would actually remove this check from the function because it should go before this function is even called. We should only validate the path if defined

nitishfy: Here's one more thing I observed:

The above piece of code you wrote will work if use the command `okteto deploy`. Even if we use this piece of code before the function call, the command `okteto deploy` will always lead to error. In reality, `okteto deploy` is designed to work without even mentioning the path. I think we can remove this check since everything is working fine already. WDYT?

cmd/deploy/deploy_test.go
nitishfy: Implemented! I maybe wrong in using `afero` package here (sorry, if i'm) but let me know if there are changes required.

nitishfy:
**Update:** Ready for the next set of review 
(used `afero` package as suggested by @andreafalzetti)

nitishfy:
Ping!

jLopezbarb (approved):

andreafalzetti (approved):

PR #3931, opened 2023-08-22
fix(hybrid): read secrets from kubernetes

### LINKED ISSUE #3923: [Hybrid development] Secrets stored as k8s secret are not propagated
**Describe the bug**
If I define a secret on k8s and then tries to use a service that uses the secret, the secret that the hybrid command uses is empty.

**To Reproduce**
Steps to reproduce the behavior using movies with helm:

1. Run `git clone https://github.com/okteto/movies-with-helm`
1. Checkout the hybrid branch: `git checkout origin/hybrid`
1. Run `okteto deploy --build`
1. Run `okteto up api`
1. When the interactive shell is open write `printenv | grep MONGODB_PASSWORD`
1. Check that the value is: `MONGODB_PASSWORD=`

**Expected behavior**
I'd expect to have it set with the proper password.

**Additional context**
It seems that we are not checking [here](https://github.com/okteto/okteto/blob/8e7bd2705c659c097ec04e670e60712fff969e73/cmd/up/exec.go#L237) if the env var comes from a secret and we treat all the secrets as if they have `name` and `value`

andreafalzetti:
# Proposed changes

Fixes #3923

## Description

This PR fixes the hybrid mode when an env var is expected to come from a kubernetes secret.

## How to test

1. Run `git clone https://github.com/okteto/movies-with-helm`
2. Checkout the hybrid branch: `git checkout origin/hybrid`
3. Delete the line `MONGODB_PASSWORD` from `okteto.yml`
4. Run `okteto deploy --build`
5. Run `okteto up api`
6. When the interactive shell is open write `printenv | grep MONGODB_PASSWORD`
7. Check that the value of `MONGODB_PASSWORD` is populated.

To verify that this change is fixing the bug, use Okteto CLI `2.19.0`and using the steps above, check that `MONGODB_PASSWORD` is empty instead.

### DIFF
--- a/cmd/up/exec.go
+++ b/cmd/up/exec.go
@@ -245,7 +245,9 @@ func (eg *envsGetter) getEnvs(ctx context.Context) ([]string, error) {
 	if err != nil {
 		return nil, err
 	}
-	envs = append(envs, devContainerEnvs..., eg.getDefaultLocalEnvs()...)
+	envs = append(envs, devContainerEnvs...)
+
+	envs = append(envs, eg.getDefaultLocalEnvs()...)
 
 	for _, env := range eg.dev.Environment {
 		envs = append(envs, fmt.Sprintf("%s=%s", env.Name, env.Value))
--- a/cmd/up/exec.go
+++ b/cmd/up/exec.go
@@ -286,6 +286,14 @@ func (d *devContainerEnvGetter) getEnvsFromDevContainer(ctx context.Context, spe
 			}
 			val = string(secret.Data[env.ValueFrom.SecretKeyRef.Key])
 		}
+
+		if env.ValueFrom != nil && env.ValueFrom.ConfigMapKeyRef != nil {
+			cm, err := configmaps.Get(ctx, env.ValueFrom.ConfigMapKeyRef.Name, namespace, client)
+			if err != nil {
+				return envs, fmt.Errorf("%w: the development container didn't start successfully because the kubernetes configmap '%s' was not found", err, env.ValueFrom.ConfigMapKeyRef.Name)
+			}
+			val = cm.Data[env.ValueFrom.ConfigMapKeyRef.Key]
+		}
 		envs = append(envs, fmt.Sprintf("%s=%s", env.Name, val))
 	}
 
@@ -295,7 +303,8 @@ func (d *devContainerEnvGetter) getEnvsFromDevContainer(ctx context.Context, spe
 func (cmg *configMapGetter) getEnvsFromConfigMap(ctx context.Context, name string, namespace string, client kubernetes.Interface) ([]string, error) {
 	var envs []string
 
-	devCmap, err := configmaps.Get(ctx, pipeline.TranslatePipelineName(name), namespace, client)
+	cmName := pipeline.TranslatePipelineName(name)
+	devCmap, err := configmaps.Get(ctx, cmName, namespace, client)
 	if err != nil && !oktetoErrors.IsNotFound(err) {
 		return nil, err
 	}
--- a/cmd/up/exec.go
+++ b/cmd/up/exec.go
@@ -282,7 +282,7 @@ func (d *devContainerEnvGetter) getEnvsFromDevContainer(ctx context.Context, spe
 		if env.ValueFrom != nil && env.ValueFrom.SecretKeyRef != nil {
 			secret, err := secrets.Get(ctx, env.ValueFrom.SecretKeyRef.Name, namespace, client)
 			if err != nil {
-				return nil, err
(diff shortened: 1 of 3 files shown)


cmd/up/exec.go
```diff
@@ -233,9 +241,11 @@ func (eg *envsGetter) getEnvs(ctx context.Context) ([]string, error) {
...
 
-	for _, env := range apps.GetDevContainer(app.PodSpec(), "").Env {
-		envs = append(envs, fmt.Sprintf("%s=%s", env.Name, env.Value))
+	devContainerEnvs, err := eg.devContainerEnvGetter.getEnvsFromDevContainer(ctx, app.PodSpec(), "", eg.namespace, eg.client)
+	if err != nil {
+		return nil, err
 	}
+	envs = append(envs, devContainerEnvs...)
 
 	envs = append(envs, eg.getDefaultLocalEnvs()...)
```
teresaromero, line 250:
```suggestion
	envs = append(envs, devContainerEnvs..., eg.getDefaultLocalEnvs()...)
```

maybe append both at the same operation?

teresaromero (approved):
LGTM 👍🏻  tested and works as expected, minor comment

cmd/up/exec.go
jLopezbarb: Yes, This code could be optimised by retrieving all the env vars in parallel and then adding them by the prioritisation  order. But I don't think this should be covered in this PR

cmd/up/exec.go
```diff
@@ -262,6 +272,26 @@ func getDefaultLocalEnvs() []string {
...
 }
 
+func (d *devContainerEnvGetter) getEnvsFromDevContainer(ctx context.Context, spec *apiv1.PodSpec, name, namespace string, client kubernetes.Interface) ([]string, error) {
+	var envs []string
+
+	devContainer := apps.GetDevContainer(spec, name)
+
+	for _, env := range devContainer.Env {
+		val := env.Value
+		if env.ValueFrom != nil && env.ValueFrom.SecretKeyRef != nil {
```
jLopezbarb, line 282: NIT: I don't think it should be covered in this PR but wanted to mention that there is a common scenario where a user adds env vars from a configmap by using `configMapKeyRef`, just like the secrets. But let's keep the scope short

cmd/up/exec_test.go
```diff
@@ -494,6 +516,93 @@ func TestGetEnvsFromConfigMap(t *testing.T) {
...
+							},
+						},
+					},
+				},
+			},
+			expectedEnvs: []string{
+				"FROMPOD=VALUE1",
+			},
+		},
+		{
```
jLopezbarb, line 555: Can we also test what happens when the secret doesn't exist? 

Also, testing this scenario I noticed that it could cause a very weird error, that doesn't provide enough information to the user to be able to troubleshoot it by himself:

Captura de pantalla 2023-08-23 a las 16 56 20

cmd/up/exec.go
andreafalzetti: @jLopezbarb good catch, I've created a follow-up issue to keep scope down: https://github.com/okteto/okteto/issues/3936

cmd/up/exec_test.go
andreafalzetti: Nice finding @jLopezbarb - I've wrapped the error with some more explaination for the user. Do you have any further ideas to improve this?
Screenshot 2023-08-23 at 19 07 51

andreafalzetti: > Can we also test what happens when the secret doesn't exist?

💯 I've added a test and also modified the return value to return `envs` calculated up to the moment of the error. In the test added, it showcases that the first env var is returned, then it fails because of the secret, and the last env var is not available either despite it would have been available.

I am wondering if we like this logic, or we rather log the error but continue. I guess it can be revisited in a future PR because we would need to revisit all the other functions that return err too.

Commit to check: https://github.com/okteto/okteto/pull/3931/commits/ec854fc9a72c0031eeae274cd37018930bec1062

cmd/up/exec.go
andreafalzetti: I've opened a new PR for the configmap change: https://github.com/okteto/okteto/pull/3937

cmd/up/exec_test.go
jLopezbarb: Following with what k8s does. It creates the pod but then fails on initialization. I think it's better to fail fast and could be revisited later.

> Do you have any further ideas to improve this?

I think the error is very explicit and very useful, maybe adding a `UserError` with a `Hint` let us shorten the first part but I think is good as it is

cmd/up/exec.go
jLopezbarb: Nice!!

jLopezbarb (approved):

PR #3939, opened 2023-08-24
tests: add kubetoken unit tests

jLopezbarb:
Add kubetoken tests.

## Proposed changes

This PR adds unit tests for kubeconfig command

### DIFF
--- a/cmd/kubetoken/kubetoken_test.go
+++ b/cmd/kubetoken/kubetoken_test.go
@@ -1,0 +1,166 @@
+// Copyright 2023 The Okteto Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package kubetoken
+
+import (
+	"context"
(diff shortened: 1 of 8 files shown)


teresaromero (approved):

cmd/kubetoken/validator_test.go
AdrianPedriza, line 1: One important test case could be to test `errOktetoContextNotFound` when `ctxStore` is `nil`. In `TestCtxValidator` test

cmd/kubetoken/kubetoken_test.go
```diff
@@ -0,0 +1,166 @@
...
+
+type fakeCtxCmdRunner struct {
+	err error
+}
+
+func (f fakeCtxCmdRunner) Run(ctx context.Context, ctxOptions *contextCMD.ContextOptions) error {
+	return f.err
+}
+
+func TestKubetoken(t *testing.T) {
```
AdrianPedriza, line 56: Is it worth adding test cases for errors when providing an okteto client and during serialization?

jLopezbarb: then it will fail on validation

cmd/kubetoken/kubetoken_test.go
```diff
@@ -0,0 +1,166 @@
...
+					},
+				},
+				fakeCtxCmdRunner: fakeCtxCmdRunner{
+					err: assert.AnError,
+				},
+			},
+			expected: assert.AnError,
+		},
+		{
+			name: "error getting kubetoken",
```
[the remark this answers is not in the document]
jLopezbarb: Yep, the validation checks if the kubetoken retrieval will work so that's why it's failing there. I can't think any case where it fails on the other one. I added the test because this could cause problems in the future

AdrianPedriza (approved):

PR #3940, opened 2023-08-24
Tag with global and dev when pushing to global registry

### LINKED ISSUE #3922: [Pull First Approach] Build always to global and `okteto.dev`
When I run `okteto deploy --build` + `okteto up` I might have different images.

1. Clone [movies with helm](https://github.com/okteto/movies-with-helm)
2. Modify `frontend/src/index.html`:
```diff
diff --git a/frontend/src/index.html b/frontend/src/index.html
index 91d047a..366c9f3 100644
--- a/frontend/src/index.html
+++ b/frontend/src/index.html
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html>
   <head>
-    <title>Movies</title>
+    <title>Movies test</title>
     <meta charset="utf-8">
     <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=0"/>
   </head>
```
3. Run `okteto deploy --build`
4. Check that we have the latest changes:
Captura de pantalla 2023-08-18 a las 14 11 50

5.  Modify `frontend/src/index.html`:
```diff
diff --git a/frontend/src/index.html b/frontend/src/index.html
index 91d047a..5369868 100644
--- a/frontend/src/index.html
+++ b/frontend/src/index.html
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html>
   <head>
-    <title>Movies</title>
+    <title>Movies with okteto</title>
     <meta charset="utf-8">
     <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=0"/>
   </head>
```
6. Run `git add . && git commit -m "test"`
7. Run `okteto deploy --build`
8. Modify `frontend/src/index.html`:
```diff
diff --git a/frontend/src/index.html b/frontend/src/index.html
index 91d047a..5369868 100644
--- a/frontend/src/index.html
+++ b/frontend/src/index.html
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html>
   <head>
-    <title>Movies with okteto</title>
+    <title>Movies with okteto global</title>
     <meta charset="utf-8">
     <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=0"/>
   </head>
```
9. Run `okteto deploy`
10. Check that the changes that we have is the one generated in `step 3`

It should point to the latest image built. In this case the one in `step 7`. We should be pushing both tags when building with buildkit, to `okteto.global/image:hash` and `okteto.dev/image:okteto`. This will ensure that everyone uses the latest images

**Additional context**
We'll need to change our Build command to support a list of tags instead of a single tag.

teresaromero:
# Proposed changes

Fixes #3922 

- As Tag option just enables for one tag, added new `DevTag` in order to let know buildkit that the same image has 2 tags.
- translate from global to dev when applies and expand the tag to save it at DevTag
- modify opt.Exports at buildkit so it pushes two tags for the image
- when pushing to global, success message is using the dev tag instead of global
- when pushing to global, envs set for the build are using the DevTag reference

## Tested

Followed steps at issue

Now when pushing to global, we display the following success message

### DIFF
--- a/pkg/cmd/build/buildkit.go
+++ b/pkg/cmd/build/buildkit.go
@@ -125,15 +125,29 @@ func getSolveOpt(buildOptions *types.BuildOptions) (*client.SolveOpt, error) {
 	}
 
 	if buildOptions.Tag != "" {
-		opt.Exports = []client.ExportEntry{
-			{
-				Type: "image",
-				Attrs: map[string]string{
-					"name": buildOptions.Tag,
-					"push": "true",
+		// add additional tag if DevTag is defined
+		if buildOptions.DevTag != "" {
+			opt.Exports = []client.ExportEntry{
+				{
+					Type: "image",
+					Attrs: map[string]string{
+						"name": fmt.Sprintf("%s,%s", buildOptions.Tag, buildOptions.DevTag),
+						"push": "true",
+					},
 				},
-			},
+			}
+		} else {
+			opt.Exports = []client.ExportEntry{
+				{
+					Type: "image",
+					Attrs: map[string]string{
+						"name": buildOptions.Tag,
+						"push": "true",
+					},
+				},
+			}
 		}
+
 	}
 	for _, cacheFromImage := range buildOptions.CacheFrom {
 		opt.CacheImports = append(
--- a/pkg/cmd/build/buildkit.go
+++ b/pkg/cmd/build/buildkit.go
@@ -131,7 +131,7 @@ func getSolveOpt(buildOptions *types.BuildOptions) (*client.SolveOpt, error) {
 				{
 					Type: "image",
 					Attrs: map[string]string{
-						"name": fmt.Sprintf("%s,%s", buildOptions.Tag, buildOptions.DevTag),
+						"name": strings.Join([]string{buildOptions.Tag, buildOptions.DevTag}, ","),
 						"push": "true",
 					},
 				},
--- a/pkg/registry/image.go
+++ b/pkg/registry/image.go
@@ -21,6 +21,7 @@ import (
 	containerv1 "github.com/google/go-containerregistry/pkg/v1"
 	"github.com/okteto/okteto/pkg/constants"
 	oktetoLog "github.com/okteto/okteto/pkg/log"
+	"github.com/okteto/okteto/pkg/model"
 	apiv1 "k8s.io/api/core/v1"
 )
 
@@ -151,14 +152,17 @@ func GetDevTagFromGlobal(image string) string {
 	}
 
 	// separate image reference and tag eg: okteto.dev/image:tag
-	reference, _, found := strings.Cut(image, ":")
+	reference, _, found := strings.Cut(image, "@sha256")
 	if !found {
-		return ""
+		reference, _, found = strings.Cut(image, ":")
+		if !found {
+			return ""
+		}
 	}
 
 	devReference := strings.Replace(reference, constants.GlobalRegistry, constants.DevRegistry, 1)
 	if devReference == reference {
 		return ""
 	}
-	return fmt.Sprintf("%s:okteto", devReference)
+	return fmt.Sprintf("%s:%s", devReference, model.OktetoDefaultImageTag)
 }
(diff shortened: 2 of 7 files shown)


pkg/registry/image.go
```diff
@@ -144,3 +144,21 @@ func (ImageCtrl) getExposedPortsFromCfg(cfg *containerv1.ConfigFile) []Port {
...
+	reference, _, found := strings.Cut(image, ":")
+	if !found {
+		return ""
+	}
+
+	devReference := strings.Replace(reference, constants.GlobalRegistry, constants.DevRegistry, 1)
+	if devReference == reference {
+		return ""
+	}
+	return fmt.Sprintf("%s:okteto", devReference)
```
andreafalzetti, line 163: why do we have a fixed tag `:okteto` here?

andreafalzetti (approved):
I've tested the change and I was able to reproduce the bug, and the fix with the new version 🎖️ 

Code changes look good, just a minor question. 

Good work! 🚀

jLopezbarb:
Could we have the message for pushing the image set to the dev registry? We should try to not show the global registry to the user:
Captura de pantalla 2023-08-28 a las 10 03 44

jLopezbarb:
Could we generate okteto builds images based on the okteto.dev registry image instead of the okteto.global? This way the user will always use an image that is on his dev registry.

pkg/registry/image.go
jLopezbarb: I'd use `OktetoDefaultImageTag` for the devReference. In case we change the name for it we won't forget to change it everywhere

pkg/registry/image.go
```diff
@@ -144,3 +144,21 @@ func (ImageCtrl) getExposedPortsFromCfg(cfg *containerv1.ConfigFile) []Port {
...
 	return result
 }
+
+func GetDevTagFromGlobal(image string) string {
+	if !strings.HasPrefix(image, constants.GlobalRegistry) {
+		return ""
+	}
+
+	// separate image reference and tag eg: okteto.dev/image:tag
+	reference, _, found := strings.Cut(image, ":")
```
jLopezbarb, line 154: I guess it's ok because i don't think it can happen but if the image is in the sha format there is no `:` and this will fail

jLopezbarb (approved):

AdrianPedriza:
> Could we have the message for pushing the image set to the dev registry? We should try to not show the global registry to the user: Captura de pantalla 2023-08-28 a las 10 03 44

@jLopezbarb also not as debug logs? maybe the information about in which registers the images are stored is relevant?

pkg/cmd/build/buildkit.go
```diff
@@ -125,15 +125,29 @@ func getSolveOpt(buildOptions *types.BuildOptions) (*client.SolveOpt, error) {
...
 
 	if buildOptions.Tag != "" {
-		opt.Exports = []client.ExportEntry{
-			{
-				Type: "image",
-				Attrs: map[string]string{
-					"name": buildOptions.Tag,
-					"push": "true",
+		// add additional tag if DevTag is defined
+		if buildOptions.DevTag != "" {
```
AdrianPedriza, line 129: I am not sure about this but `buildOptions.Tag` could have the same value as `buildOptions.DevTag`? and thus export twice with the same value?

pkg/registry/image.go
teresaromero: i changed this to check sha format first and then tag.

jLopezbarb (approved):

jLopezbarb:
> > Could we have the message for pushing the image set to the dev registry? We should try to not show the global registry to the user: Captura de pantalla 2023-08-28 a las 10 03 44
> 
> @jLopezbarb also not as debug logs? maybe the information about in which registers the images are stored is relevant?

I would keep this since its a buildkit log and the user should know that they updated the "global cache", but the one that is used to is the one in his personal namespace

PR #3941, opened 2023-08-25
feat: add silent logger

jLopezbarb:
# Proposed changes

Fixes #https://github.com/okteto/app/issues/6889

Depends on: https://github.com/okteto/okteto/pull/3939

- Adds silent logger to not print logs like initializingf context from OKTETO_URL

### DIFF
--- a/pkg/log/format.go
+++ b/pkg/log/format.go
@@ -57,6 +57,8 @@ const (
 	PlainFormat string = "plain"
 	// JSONFormat represents a json logger
 	JSONFormat string = "json"
+	// SilentFormat represents a silent logger
+	SilentFormat string = "silent"
 )
 
 func (l *logger) getWriter(format string) OktetoWriter {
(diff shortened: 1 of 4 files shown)


pkg/log/format.go
```diff
@@ -71,6 +73,9 @@ func (l *logger) getWriter(format string) OktetoWriter {
 		l.outputMode = JSONFormat
 		l.out.SetFormatter(&JSONLogFormat{})
 		return newJSONWriter(l.out, l.file)
+	case SilentFormat:
```
andreafalzetti, line 76: Perhaps it's more explicit if we call it "Log File" or something related? 

```suggestion
	case LogFileFormat:
```

teresaromero: 🤔 giving a thought to this, for me it makes sense "silent" as is not displaying output.... but, is saving it to a file. We are using this as option internally, so, i would keep the name but perhaps add couple more description to the Format (line 61) or here in the case (line 76) to explain why and how the logs are kept

jLopezbarb: I prefer to use the `Silent` naming but definitely this is something we need to join and discuss about it (everything related to logs).

PR #3946, opened 2023-08-27
fix: registrytoken cmd output processing

jpf-okteto:
The following PR fixes how registrytoken command should process output.

It modifies main.go for overriding logger properties, such as writing to stderr to avoid corruption of docker credential helper output. It also sets the format to JSON and lowers the level to INFO.

Another big change is that registrytoken command no longer returns errors neither it lets main.go error handling execute if there is an error. This is because docker credential helper expected outputs are sensitive.

### DIFF
--- a/cmd/registrytoken/registrytoken.go
+++ b/cmd/registrytoken/registrytoken.go
@@ -28,6 +28,10 @@ import (
 	"github.com/docker/docker-credential-helpers/credentials"
 )
 
+// TODO @jpf-okteto the following commit, included in v0.8.0 defines
+// consts for each action directly at the dependency.
+//
+// https://github.com/docker/docker-credential-helpers/commit/129017a3cdb99cd8190c525a75d4da6e1d8a9506
 const (
 	ActionStore   = "store"
 	ActionGet     = "get"


main.go
```diff
@@ -86,6 +86,11 @@ func init() {
 func main() {
 	ctx := context.Background()
 	oktetoLog.Init(logrus.WarnLevel)
+	if registrytoken.IsRegistryCredentialHelperCommand(os.Args) {
```
teresaromero, line 89: @jLopezbarb just implemented a new type of logger "silent" in order to control stdout... maybe his implementation can be used here too? https://github.com/okteto/okteto/pull/3941

teresaromero (approved):

jLopezbarb (approved):

maroshii (approved):

cmd/registrytoken/registrytoken.go
```diff
@@ -54,19 +55,41 @@ More info about docker credentials helpers here: https://github.com/docker/docke
...
+	if len(args) != 3 {
+		return false
+	}
+
+	if args[1] != "registrytoken" {
+		return false
+	}
+
+	switch args[2] {
+	case "store", "get", "erase", "list", "version":
```
AdrianPedriza, line 90: how about storing the valid args in a const and using that const here and in `cobra.Command.ValidArgs`, in command declaration? less error prone I think

AdrianPedriza (approved):

PR #3949, opened 2023-08-28
chore: add rch install / uninstall cmds

jpf-okteto:
The following PR adds command methods for installing Okteto's docker registry credential helper. These commands are behind registrytoken, hidden by default from top level view.

Install command includes a --force flag if credential store is not empty and set to a value different that okteto.

Uninstall command includes a --force flag if credential store is not empty and set to a value different that okteto.

Additionally, context is now correctly propagated through registrytoken command tree.

Signed-off-by: Javier Provecho Fernández (Okteto) jpf@okteto.com

### DIFF
--- a/cmd/registrytoken/registrytoken.go
+++ b/cmd/registrytoken/registrytoken.go
@@ -15,6 +15,7 @@ package registrytoken
 
 import (
 	"context"
+	"fmt"
 	"os"
 
 	contextCMD "github.com/okteto/okteto/cmd/context"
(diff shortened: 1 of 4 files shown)


teresaromero (approved):

cmd/registrytoken/install.go
```diff
@@ -0,0 +1,56 @@
...
+		Use:   "install",
+		Short: "Install Okteto's docker-credential-helper",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if err := contextCMD.NewContextCommand().Run(ctx, &contextCMD.ContextOptions{}); err != nil {
+				return err
+			}
+			confDir := config.Dir()
+			conf, err := config.Load(confDir)
+			if err != nil {
+				return errors.Wrapf(err, "couldn't load docker config file from %q", confDir)
```
jLopezbarb, line 30: Coiuld we use `fmt.Errorf()`. The library used is deprecated since 2.13 afaik: https://github.com/pkg/errors

maroshii: I don't think the library is deprecated, it just feature complete and locked due to the go2 errors proposal:

- https://github.com/pkg/errors#roadmap
- https://go.googlesource.com/proposal/+/master/design/go2draft.md

jLopezbarb (approved):

AdrianPedriza:
I see in the tests scenarios that executing `okteto registrytoken install --force` force an overwrite of existing credential store but executing it I see:
Captura de Pantalla 2023-08-28 a las 16 15 17
Seems it doesn't force an installation if credential store is same. Is that expected?

PR #3953, opened 2023-08-29
add e2e using static token

AdrianPedriza:
# Proposed changes
- add e2e using feature flag to use static tokens

Fixes #https://github.com/okteto/app/issues/6890 (partially)

### DIFF
--- a/integration/kubeconfig/kubeconfig_test.go
+++ b/integration/kubeconfig/kubeconfig_test.go
@@ -17,6 +17,7 @@ package kubeconfig
 
 import (
 	"path/filepath"
+	"strconv"
 	"testing"
 
 	"github.com/okteto/okteto/integration"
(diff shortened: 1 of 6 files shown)


pkg/k8s/kubeconfig/const.go
```diff
@@ -0,0 +1,6 @@
+package kubeconfig
+
+const (
+	// OktetoUseStaticKubetokenEnvVar is used to opt in to use static kubetoken
```
AdrianPedriza, line 4: this const needs to be public in order to use it in e2e test so moving it here because import `context` pkg can have a bigger impact. This way is more isolated

jLopezbarb: I don't know if we should create a kubeconfig package just to have a variable inside. We are using it just in the context package. Should it live there and import it in the tests?

integration/kubeconfig/kubeconfig_test.go
```diff
@@ -25,24 +26,46 @@ import (
...
 	require.NoError(t, err)
 
 	home := t.TempDir()
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Setenv(kubeconfig.OktetoUseStaticKubetokenEnvVar, strconv.FormatBool(tt.useStaticToken))
 
-	err = commands.RunOktetoKubeconfig(oktetoPath, home)
-	require.NoError(t, err)
+			err = commands.RunOktetoKubeconfig(oktetoPath, home)
```
jLopezbarb, line 53: Can we also test that kubetoken output is json parseable?

pkg/k8s/kubeconfig/const.go
AdrianPedriza: it only creates a new `const.go` file but yeah for now I think it is more reasonable your point

AdrianPedriza: done 43f5e125b93ca0e03eb19bb9659691f6f44e3110

integration/kubeconfig/kubeconfig_test.go
AdrianPedriza: I think we already have that test -> https://github.com/okteto/okteto/blob/master/integration/kubetoken/kubetoken_test.go#L42. Is that what you mean?

jLopezbarb (approved):

PR #3956, opened 2023-08-31
expand branch value if comes from variable

AdrianPedriza:
# Proposed changes
- expand dependencies values if variables values fits
- add unit test

Fixes #https://github.com/okteto/app/issues/7104

# How to test changes
To test everything works as expected, I've created a repo with a custom `dependencies` section. The sections looks like:
`````yml
dependencies:
  test-depen:
    repository: ${DESIRED_REPO}
    branch: ${DESIRED_BRANCH}
    variables:
      ${CUSTOM_VARIABLE}: ${CUSTOM_VARIABLE_VALUE}
`````
The expected behavior is that every variable set in the UI need to be expanded in this section so `repository`, `branch` and values for name and value related to variable set in the section need to be expanded. 
Steps:
1. Using Okteto UI, deploy git repo `https://github.com/AdrianPedriza/movies` using the following values:
Captura de Pantalla 2023-09-04 a las 9 45 04

2. Click "Launch" button.
3. Check variables has been propagated as expected:
- dependencies is using the desired repo and branch:
Captura de Pantalla 2023-09-04 a las 9 47 13

- desired variable name and value is in configmap under `variables` section. Exec the following command:
````cmd
k get cm okteto-git-test-depen -o json | jq -r '.data.variables' | base64 --decode
````

### DIFF
--- a/pkg/model/manifest.go
+++ b/pkg/model/manifest.go
@@ -1021,8 +1021,8 @@ func (d *Dependency) GetTimeout(defaultTimeout time.Duration) time.Duration {
 	return defaultTimeout
 }
 
-// ExpandVarsInSection sets dependencies values if values fits with list params
-func (d *Dependency) ExpandVarsInSection(envKeyPairs []string) error {
+// ExpandVars sets dependencies values if values fits with list params
+func (d *Dependency) ExpandVars(envKeyPairs []string) error {
 	parser := parse.New("string", envKeyPairs, &parse.Restrictions{})
 
 	expandedBranch, err := parser.Parse(d.Branch)
--- a/pkg/model/manifest.go
+++ b/pkg/model/manifest.go
@@ -1022,8 +1022,8 @@ func (d *Dependency) GetTimeout(defaultTimeout time.Duration) time.Duration {
 }
 
 // ExpandVars sets dependencies values if values fits with list params
-func (d *Dependency) ExpandVars(envKeyPairs []string) error {
-	parser := parse.New("string", envKeyPairs, &parse.Restrictions{})
+func (d *Dependency) ExpandVars(variables []string) error {
+	parser := parse.New("string", append(os.Environ(), variables...), &parse.Restrictions{})
 
 	expandedBranch, err := parser.Parse(d.Branch)
 	if err != nil {
--- a/pkg/model/manifest.go
+++ b/pkg/model/manifest.go
@@ -25,6 +25,7 @@ import (
 	"time"
 
 	"github.com/a8m/envsubst"
+	"github.com/a8m/envsubst/parse"
 	"github.com/okteto/okteto/pkg/constants"
 	"github.com/okteto/okteto/pkg/discovery"
 	oktetoErrors "github.com/okteto/okteto/pkg/errors"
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by teresaromero** [comment]
@AdrianPedriza can you expand the description with the testing done and how to repro?

File: pkg/model/manifest.go [hunk 1]
```diff
@@ -1020,6 +1021,70 @@ func (d *Dependency) GetTimeout(defaultTimeout time.Duration) time.Duration {
 	return defaultTimeout
 }
 
+// ExpandVarsInSection sets dependencies values if values fits with list params
+func (d *Dependency) ExpandVarsInSection(envKeyPairs []string) error {
```
teresaromero, line 1025: nit: `ExpandVars` what does the `InSection` mean?

pkg/model/manifest_test.go
```diff
@@ -1147,6 +1148,53 @@ func Test_GetTimeout(t *testing.T) {
...
+				Value: "${AVARVALUE}",
+			},
+			EnvVar{
+				Name:  "$${ANAME}",
+				Value: "MyValue",
+			},
+		},
+	}
+	expected := Dependency{
+		Repository:   "my/repo",
```
teresaromero, line 1169: where are the envs for the tests set? 🤔  are they global?

teresaromero: can we set some env value to multiline to check this behavour?

AdrianPedriza:
> @AdrianPedriza can you expand the description with the testing done and how to repro?

done! check it 😄

pkg/model/manifest_test.go
AdrianPedriza: this function doesn't use environment variables. It just uses variables coming from `deploy variables` retrieved from configmap. The variables are stored in config map by the backend

pkg/model/manifest.go
AdrianPedriza: `InSection` means "in dependencies section" but I don't think at this point is needed to mention it. I think is enough to use `ExpandVars`, being called by `Dependency` struct I think it gives the needed context to understand what the func does. Done at b7009a55146dff577014105f540e9c74c125a621 😄

pkg/model/manifest.go
```diff
@@ -1020,6 +1021,70 @@ func (d *Dependency) GetTimeout(defaultTimeout time.Duration) time.Duration {
 	return defaultTimeout
 }
 
+// ExpandVars sets dependencies values if values fits with list params
+func (d *Dependency) ExpandVars(variables []string) error {
+	parser := parse.New("string", append(os.Environ(), variables...), &parse.Restrictions{})
+
+	expandedBranch, err := parser.Parse(d.Branch)
+	if err != nil {
+		return err
```
jLopezbarb, line 1030: should we wrap the error mentioning which field is the one failing? This way it will be easier to debug:
```suggestion
		return fmt.Errorf("error expaning 'branch': %w", err)
```

jLopezbarb (approved):

teresaromero:
Lets review the tests, `Test_ExpandVars` is failing

AdrianPedriza:
> Lets review the tests, `Test_ExpandVars` is failing.
> 
> Functionality is working, tested at UI and from CLI. Deferring approval until unti-tests are solved

fixed! f7f4e2607e3461101b29aedd1f32374d0caa8dbd . Now that we are including os.Environ() to the expansion I think another use of `NAMESPACE` env var was causing an overlap!

pkg/model/manifest.go
AdrianPedriza: done 027299b5b84341aa1a4f0eb8b6ce2eea2cc007e7

teresaromero (approved):

andreafalzetti (approved):

PR #3960, opened 2023-09-04
chore: update codecov

### LINKED ISSUE #3829: [CLI Quality][CI] Configure codecov for CLI
As part of the CLI Quality initiative, code coverage tooling has to be impoved.

This issue is to review and research how the test coverage could be improved and be adding value into our development flow

## Acceptance criteria

- [ ] Codecov must run on every PR and highlights the same output as the local workflow
- [ ] Codecov must prevent merges if new patch is not covered by unit test

jLopezbarb:
Fixes #3829 

## Proposed changes

This pull request makes several updates to the project's configuration files:

1. **`coverage` Configuration**

   - Changed the `status.project.default` section to set the `target` to `auto` and `threshold` to `'15%'`: Tries to ensure that the project keeps improving with every PR

   - Modified the `status.patch.default` section to set the `target` to `75%` and added `only_pulls: true`: A new PR must have tests covering most of the scenarios introduced
   
   2. **`comment` Configuration**

   - Updated the layout to include `header`, `changes`, `diff`, and `sunburst`: I found that this layout is not as bad as the default one. Let me know wdyt

3. **`ignore` Configuration**

   - Added the following patterns to the `ignore` section in order to skip coverage on packages that don't have actual CLI code:
     - `**/*_test.go`
     - `integration/*`
     - `scripts/*`
     - `internal/*`

4. **`github_checks` Configuration**

   - Removed the entire `github_checks` section, including `annotations` We need to ensure that this new cofiguration is applied for every PR

### Purpose of Changes

These changes aim to improve the project's workflow and code coverage reporting. By setting specific coverage thresholds and ignoring certain files and directories, we can maintain a cleaner and more effective CI/CD pipeline.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,13 +1,22 @@
-# prevent codecov from blocking a PR and remove comments
+comment:
+  layout: header, changes, diff, sunburst
+
 coverage:
   status:
     project:
       default:
-        informational: true
+        target: auto
+        threshold: '15%'
+
     patch:
       default:
-        informational: true
+        target: 75%
+        only_pulls: true
 
-comment: false
-github_checks:
-  annotations: false
+    changes: false
+
+ignore:
+  - '**/*_test.go'
+  - integration/*
+  - scripts/*
+  - internal/*
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -6,7 +6,7 @@ coverage:
     project:
       default:
         target: auto
-        threshold: '15%'
+        threshold: '0%'
 
     patch:
       default:
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -13,8 +13,6 @@ coverage:
         target: 75%
         only_pulls: true
 
-    changes: false
-
 ignore:
   - '**/*_test.go'
   - integration/*


.codecov.yml
teresaromero, line 1: does the circleci file need to be updated? Are we sending the coverage retrieved at the jobs to codecov?

jLopezbarb: If we want to start having metrics for different type of tests unit/integration/e2e we will need to update and send them with the proper tags but right now, given that we only have one kind I'd leave it as simple as possible. WDYT?

AdrianPedriza, line 1: Could this configuration block merge to branches that were not master?

teresaromero (approved):

.codecov.yml
```diff
@@ -1,13 +1,20 @@
...
+comment:
+  layout: header, changes, diff, sunburst
+
 coverage:
   status:
     project:
       default:
-        informational: true
+        target: auto
+        threshold: '15%'
```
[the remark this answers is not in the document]
andreafalzetti: If I understand, this is the threshold of how much we are allowing a PR to decrease the coverage, if that's the case I agree that 15% is high

.codecov.yml
jLopezbarb: That's something that should be done in CI and either not pass the codecov at all or pass it with a different configuration.

I think maintaining two different codecov files can increase complexity, so I think we should start just with the one we have and if we face that problem later we can iterate and fix it. WDYT?

AdrianPedriza: Agree!

AdrianPedriza (approved):

PR #3961, opened 2023-09-04
Basic golangci lint

### LINKED ISSUE #3828: [CLI Quality][CI] Implement basic `golang-ci` linters
As part of the CLI Quality initiative, `golang-ci` is being implemented as a local and remote workflow.

This first issue will implement the basics for the lint tool and as part of it a refinement has to be made in order to fix the possible ignored rules that appear from this implementation.

## Acceptance criteria

- [x] `golang-ci` is run locally, the same config is run for every PR and when integration on master
- [ ] `golang-ci` config rules are the default ones
  - [x] errcheck
  - [x] gosimple
  - [ ] govet
  - [ ] ineffassign
  - [ ] staticcheck
  - [ ] typecheck
  - [ ] unused
- [ ] update the linters research to decide which rules are we applying
- [ ] code failures because of this rules have to be ignored, and an issue has to be created and described its priority to be solved in next sprint

teresaromero:
# Proposed changes

Partially resolves #3828 

As part of the CLI Quality initiative, this PR enables the job for running `golangci-lint` during our dev process. 
The pipeline is run before `build-binaries` or `unit-tests` in order to make the static check before continuing.

Also, disabled couple linterns that where failing locally but as we did not have the ci tool where not being tracked.

Following PRs will fix this and enable the linters so we can track them and inform at the PR level.

Regarding `pre-commit` i tried to add this to our circleci config as we removed form github action workflow but it was kind of tricky. I updated README so its clear to run this locally by configuring the hook in a pre-commit so it uses the configuration we have in place

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -23,7 +23,6 @@ linters:
     - gosimple
     - ineffassign
     - errcheck
-
 # Enabled by your configuration linters:
 # deadcode: Finds unused code [fast: false, auto-fix: false]
 # gosimple (megacheck): Linter for Go source code that specializes in simplifying code [fast: false, auto-fix: false]
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,6 +1,11 @@
 run:
   timeout: 5m
-  go: 1.20
+  
+output:
+  format: checkstyle:golangci-report.xml
+  print-issued-lines: false
+  # print-linter-name: false
+  sort-results: true
 # All possible options can be found here https://github.com/golangci/golangci-lint/blob/master/.golangci.reference.yml
 linters-settings:
   errcheck:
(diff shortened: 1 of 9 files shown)


.golangci.yml
```diff
@@ -1,6 +1,10 @@
 run:
   timeout: 5m
-  go: 1.20
```
jLopezbarb, line 3: Will this use the go version installed if not defined?

teresaromero: https://golangci-lint.run/usage/configuration/

> # Define the Go version limit.
>  # Mainly related to generics support since go1.18.
>  # Default: use Go version from the go.mod file, fallback on the env var `GOVERSION`, fallback on 1.18
>  go: '1.19'

teresaromero: i removed it so it uses the one on the repo and not having to update in more places

jLopezbarb (approved):

andreafalzetti (approved):
Awesome!

PR #3970, opened 2023-09-07
Fix preview list output

teresaromero:
# Proposed changes

Fixes #3904 

When the flag `--output` was implemented in order to get the preview list in different formats (`yaml` or `json`) the function used to show this output was our custom logger `oktetoLog.Println()`.

The logger format is set as a global flag `--log-output`, which default is TTY, but it can change to json, plain, silent, etc...

The command `okteto preview list` by default shows a table with a list of preview environments, when we where setting the `--log-output` to `json` and the `--output` to `json`, the logger has a fixed setting to mrashall the log into specific fields (stage log) so it was not marshalling the previews list, and the output was empty.

To fix this:
- Change `oktetoLog.Println()` to` fmt.Println()` for both `json` and `yaml` outputs, the output flag was linked to the log output format, this dependency has been removed. https://github.com/okteto/okteto/pull/3970/commits/0e1f3cffcfd6a9819f96ad282a0285672d43def2
- Refactor the command `okteto preview list` in order to add unit-tests to all components of the command and extract the oktetoClient which retrieves the previews list at the top of the command.

## How to validate

1. Use okteto context to log into your okteto instance
2. Run `okteto preview deploy` with or without labels (its recommended with so all the fields of the list are filled). This will create a preview environment
3. Run `okteto preview list` check that a table is the output

```
❯ okteto preview list
 i  Using teresaromero @ cloud.okteto.com as context
Name                       Scope     Sleeping  Labels
goofy-leakey-teresaromero  personal  false     testing
```

4. Run `okteto preview --log-output json` shows the default view, as there are only info and debug logs we should change the log level to see the logs in json format running `okteto preview list --log-output json --log-level debug`. This does not conflict with `--output` as the print function to show the table is not tied to `oktetoLog`

```
{"level":"debug","stage":"","message":"server name override: \"\"","timestamp":1693921450}
{"level":"info","stage":"","message":"started okteto preview list --log-output json --log-level debug","timestamp":1693921450}
{"level":"info","stage":"","message":"Using teresaromero @ cloud.okteto.com as context","timestamp":1693921451}
Name                       Scope     Sleeping  Labels
goofy-leakey-teresaromero  personal  false     testing
{"level":"info","stage":"","message":"finished okteto preview list --log-output json --log-level debug","timestamp":1693921451}
```

5. Running `okteto preview list --log-output json --log-level debug --output json` we see that the logs are shown but the preview list is not shown in any format

```
❯ okteto preview list --log-output json --log-level debug --output json
[...]

PR #3971, opened 2023-09-07
feat: add pipeline list cmd

andreafalzetti:
# Proposed changes

Fixes https://github.com/okteto/app/issues/6999

Adds a command `okteto pipeline list` to list all dev environments.

### 1 - Happy Path

1. Build this branch
2. Using an Okteto Cluster, create a new namespace:
```
okteto ns <xyz>
```
3. Deploy the following pipelines:
```bash
okd pipeline deploy --repository https://github.com/okteto/movies --label=movies --label=demo
okd pipeline deploy --repository https://github.com/okteto/go-getting-started --label=go --label=demo
```
4. List pipelines:
```
okteto pipeline list
```
5. Validate that you have both dev environments with the right labels
6. Filter by one of the labels and observe if the table gets updated accordingly.
7. Now destroy one pipeline and quickly run the list command again, to make sure the Status is showing correctly
8. Then, using the UI put to sleep the namespace and run the list pipeline command again, notice how the status is now set to "Sleeping"

### 2 - Flag: Output
1. If you followed the steps above, continue listing in JSON with:
```
okteto pipeline list --output json
```
2. Now YAML
```
okteto pipeline list --output yaml
```
3. Try filtering by label to make sure the list is returned correctly filtered
4. Consider piping the output of the command into `jq` and `yq` to validate it's only outputting JSON or YAML without any other unwanted char.
5. Create a new namespace 
```
okteto ns <xyz>
```
6. List again and validate both JSON and YAML outputs return empty lists
```
okteto pipeline list --namespace <new>
```

### 3 - Flag: Context

1. At this point you should have a context that has at least 2 pipelines deployed. Run the list for another cluster by doing:
```
okteto pipeline list --context <ANOTHER OKTETO CLUSTER>
```
2. Observe that the list returned is correct (use the UI of the other cluster to validate)
3. Validate that your default context has not been changed, the flag should not have that side effect

### 4 - Flag: Namespace

1. Create a new namespace and switch to it using `okteto ctx <new ns>`
2. Now list pipelines of another namespace where you have something already deployed: `okteto pipeline list --namespace <name>`
3. Now try listing a namespace of another cluster:
```
okteto pipeline list --cluster <c> --namespace <n>
```
4. Make sure the data returned is correct, your default context and namespace have not been modified

### 5 - Vanilla Clusters

1. Try running the command against a vanilla cluster. You should see an error, not allowing you.

## CLI Quality Reminders 🔧


cmd/pipeline/list.go
```diff
@@ -0,0 +1,217 @@
...
+			c, _, err := pc.k8sClientProvider.Provide(okteto.Context().Cfg)
+			if err != nil {
+				return fmt.Errorf("failed to load okteto context '%s': %v", okteto.Context().Name, err)
+			}
+
+			return executeListPipelines(cmd.Context(), *flags, configmaps.List, getPipelineListOutput, c, os.Stdout)
+		},
+	}
+
+	cmd.Flags().StringArrayVarP(&flags.labels, "label", "", []string{}, "tag and organize dev environments using labels (multiple --label flags accepted)")
```
jLopezbarb, line 98: Should we be able to run it into a certtain cluster?

```suggestion
	cmd.Flags().StringVarP(&flags.context, "context", "c", "", "context where the pipelines are deployed (defaults to the current context)")
	cmd.Flags().StringVarP(&flags.namespace, "namespace", "n", "", "namespace where the pipelines are deployed (defaults to the current namespace)")
```

cmd/pipeline/list.go
```diff
@@ -0,0 +1,217 @@
...
+
+func list(ctx context.Context) *cobra.Command {
+	flags := &listFlags{}
+
+	cmd := &cobra.Command{
+		Use:   "list",
+		Short: "List all okteto pipelines",
+		Args:  utils.NoArgsAccepted(""),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if flags.namespace == "" {
```
jLopezbarb, line 61: Maybe we can wrap all the context initialisation in order to make it testable.

cmd/pipeline/list.go
andreafalzetti: Do we want to? I am not sure, considering other commands require switching context before executing them.

cmd/pipeline/list.go
andreafalzetti: I made some changes now, not perfect but still it's an improvement. Let me know if you have feedback on that

cmd/preview/list.go
teresaromero, line 1: i did some changes here https://github.com/okteto/okteto/pull/3970 you might want to rebase, i just merged to master

cmd/pipeline/list.go
teresaromero: is the context flag supported by the spec? 🤔  @jLopezbarb

jLopezbarb: > Do we want to? I am not sure, considering other commands require switching context before executing them.

I think it's the same for this one, we need to load a context given the namespace and the context (default to current one). We can talk about this offline before making a decision, but I think it's the same as the rest of pipeline commands

> is the context flag supported by the spec? 

No, but I find it odd to have that flag on all the commands except for this one.

cmd/preview/list.go
andreafalzetti: I've tried rebasing from master but it's giving me a bunch of conflicts to fully rebase. Considering that PR is for `preview list` perhaps I can just go ahead and merge this one, or is there anything particular that I should make sure to check?

teresaromero: i did some changes regarding testing and wrapping functions, the changes regarding make private the flag i did them already so maybe you can remove from ur files? 🤔  what kind of problem are there?

andreafalzetti: @teresaromero ahh ok, I hadn't noticed! Fixed now

cmd/pipeline/list.go
andreafalzetti: as discussed offline, it makes sense to add, so I've updated the PR reflecting this change!!

teresaromero (approved):
👍🏻  Tested and looking good! Also tested the log-output combination to check it works as expected :) 👏🏻 👏🏻

PR #3976, opened 2023-09-11
[golangci-lint] Enable lint unused unparam

### LINKED ISSUE #3968: [CLI Quality][CI] Enable `unused` and fix issues

teresaromero:
# Proposed changes

Resolves #3968 

- Enable both linters and fix the issues raised by them
- Change circleci executor to use own okteto golangci image which already includes `golangci-lint`
- Fix cache for golangci-lint job
- Cleanup golangci.yml configuration: removed timeout (default is 1m and at the CI job is set on the cmd line to 5 as when the cache is pruned the process needs at least this time), added `new-from-rev: HEAD~` as with this PR all the default lints are enabled and this option runs the checks only in the changed code. 

EDIT:

Found on this issue while trying to fix `golangci-lint` cache issues:
- Key for cache was set to keep the wrong directories. When i changed the executor image, our executor was failing to recover the cache path `/go/pkg` that is because that dir belongs to `root` although `/go/pkg/mod` belongs to `circleci` user.
- When fixing the cache paths that have to be restored with every run, the correct ones to keep are `./cache/golangci-lint` and `./cache/go-build` in order to have the dependencies updated and also the lints updated. 
- The lint cache key has been set to `golangci-lint-cache-v2-{{ checksum ".golangci.yml" }}-{{ checksum "go.sum" }}`. This means the cache will be pruned when the golangci.yml changes (adding new lints or changing settings) and the dependencies change

With this finding there where pipelines not being able to restore `/go/pkg` so a followup issue will be added in order to fix this CI problem.

- Follow up https://github.com/okteto/app/issues/7155


.circleci/config.yml
AdrianPedriza, line 1: comparing latest executions of `golangci-lint` job it seems ``restore_cache` is giving us some kind of permission error:
#### Good one
Captura de Pantalla 2023-09-12 a las 13 30 03

#### Proposed one
Captura de Pantalla 2023-09-12 a las 13 31 51

is that expected? the rest looks good to me

teresaromero: yes, i was trying to fix it but i think this didnt work :/ i will take a look again

teresaromero: updated description of the PR with information regarding this. Unfortunatly this has been happening for a while, as the /go/pkg/ folder belongs to root, and user is circleci, any pipeling using a cache that restores this directory is failing (test-integration) 

I was looking to fix this on this PR but i think is worth opening an ISSUE under the CLI Quality CI initiative to review the pipelines and fix the caches

.golangci.yml
```diff
@@ -1,11 +1,10 @@
...
+  modules-download-mode: readonly
+
 output:
-  # format: checkstyle:golangci-report.xml
   print-issued-lines: false
-  # print-linter-name: false
   sort-results: true
-# All possible options can be found here https://github.com/golangci/golangci-lint/blob/master/.golangci.reference.yml
+
 linters-settings:
```
jLopezbarb, line 8: Could we test if the following setting adds a lot of false positives? I think it's a good addition based on our reviews:
```yaml
linters-settings:
  unparam:
    check-exported: true
```

cmd/destroy/destroy.go
```diff
@@ -188,6 +188,7 @@ func Destroy(ctx context.Context) *cobra.Command {
 				oktetoClient:      okClient,
 				buildCtrl:         newBuildCtrl(options.Name),
 				analyticsTracker:  analytics.NewAnalyticsTracker(),
+				getManifest:       model.GetManifestV2,
```
jLopezbarb, line 191: Wow. This could affect us and create a big issue, This is a big improvement!!

jLopezbarb (approved):

AdrianPedriza (approved):

PR #3983, opened 2023-09-18
fix(hybrid): terminate process correctly

andreafalzetti:
# Proposed changes

Fixes https://github.com/okteto/app/issues/7117

After researching, it looks like `bash` cannot be terminated with `SIGTERM`. This is causing issues in hybrid mode because some processes remain orphans (such as `webpack` in the case of the issue reported).

To validate what I've said, try to:

1. Launch `bash` and from that session get its pid by running `echo $$`
2. Copy the PID and from another terminal run `kill -SIGTERM <PID>` notice how the bash session is not being terminated
3. Now try with `kil -SIGKILL <PID>` and notice how bash is successfully terminated.

This PR will change how shutdown procedure to treat bash and any session leader processes differently. This is a safe approach because we already safely send SIGTERM to each child process recursively, we only forcefully kill bash when all child processes have already been terminated.

## How to validate

Using the latest version of `okteto`, reproduce the bug first:

1. Start `okteto up frontend-hybrid` in `okteto/app`
2. Run `ps aux | grep frontend-hybrid` and get the PID of the CLI
3. Run `watch -n 0.1 pstree <PID>` to monitor updates of the process and its children
4. Once `okteto up` is ready, run `yarn start` inside
5. Now run `kill -SIGTERM <PID>` and observe the list of processes from the watch command, has not been fully cleaned

Now repeat the steps above using the version of the CLI from this branch, you should see the list of child processes fully cleaned.

Additionally, repeat the steps above but instead of manually killing the CLI, run an upgrade of your cluster. Observe the processes being fully cleaned up.

## CLI Quality Reminders 🔧

### DIFF
--- a/cmd/up/up.go
+++ b/cmd/up/up.go
@@ -973,6 +973,7 @@ func terminateChildProcess(parent int, pList []ps.Process) {
 
 func terminateProcess(pid int) error {
 	oktetoLog.Debugf("terminating process: %s", pid)
+
 	p, err := os.FindProcess(pid)
 	if err != nil {
 		oktetoLog.Debugf("error getting process %s: %v", pid, err)
@@ -985,11 +986,9 @@ func terminateProcess(pid int) error {
 		return err
 	}
 
-	oktetoLog.Debugf("requesting process termination for pid: %s", pid)
-
 	if pgid == pid {
-		// If the process id (pid) is equal to its process group id (pgid), it means that the process is a session leader (ie. bash)
-		// In that case, sending a SIGTERM will not kill the process, so we need to send a SIGKILL
+		// if the pid equals its process group id (pgid), it's a session leader process such as bash
+		// SIGTERM will not terminate it, so we need to send SIGKILL instead
 		oktetoLog.Debugf("killing session leader process %s", pid)
 		if err := p.Kill(); err != nil {
 			oktetoLog.Debugf("error terminating session leader process %s: %v", p.Pid, err)
--- a/cmd/up/up.go
+++ b/cmd/up/up.go
@@ -891,7 +891,7 @@ func (up *upContext) getInsufficientSpaceError(err error) error {
 
 // Shutdown runs the cancellation sequence. It will wait for all tasks to finish for up to 500 milliseconds
 func (up *upContext) shutdown() {
-	if up.isTerm {
+	if !up.Dev.IsHybridModeEnabled() && up.isTerm {
 		if err := term.RestoreTerminal(up.inFd, up.stateTerm); err != nil {
 			oktetoLog.Infof("failed to restore terminal: %s", err.Error())
 		}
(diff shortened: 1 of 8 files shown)


cmd/up/up.go
```diff
@@ -972,22 +972,43 @@ func terminateChildProcess(parent int, pList []ps.Process) {
...
+	if err != nil {
+		oktetoLog.Debugf("error getting process group id %s: %v", pid, err)
 		return err
 	}
-	if _, err := p.Wait(); err != nil {
-		oktetoLog.Debugf("error waiting for process to exit %s: %v", p.Pid, err)
-		return err
+
+	if pgid == pid {
+		// if the pid equals its process group id (pgid), it's a session leader process such as bash
```
teresaromero, line 990: 🫠 wow! nice research..... i am wondering if using bash is decided by us i recall? i mean, the user just uses "yarn start" but this is executed with bash....
- what would happen in windows without bash?
- have we tried to run the yarn start with sh?

pkg/registry/client.go
```diff
@@ -113,7 +113,7 @@ func (c client) Write(ref name.Reference, image v1.Image) error {
 func (c client) GetDigest(image string) (string, error) {
 	descriptor, err := c.GetDescriptor(image)
 	if err != nil {
-		return "", fmt.Errorf("error getting image digest: %w", err)
```
andreafalzetti, line 116: This is redundant because the calling function already wraps this error making the message look like the following:

`Command execution failed: error getting image metadata: error getting image tag with digest: error getting image digest: error getting image descriptor: get https://gcr.io/XXX`

With this change:

`Command execution failed: error getting image metadata: error getting image tag with digest: error getting image descriptor: get`

cmd/up/up.go
andreafalzetti: - In the example reported, `bash` is the command in `okteto.yml`
- In Windows it's easier because it's directly sending a `SIGKILL`
- I've tried with `yarn start` inside `sh` and works too :)

teresaromero:
Can we run integration? both e2e and win?

cmd/up/up.go
```diff
@@ -947,7 +947,9 @@ func (up *upContext) shutdownHybridMode() {
 
 	terminateChildProcess(up.hybridCommand.Process.Pid, pList)
 
-	if err := terminateProcess(up.hybridCommand.Process.Pid); err != nil {
+	p := newOktetoProcess(up.hybridCommand.Process.Pid)
+
+	if err := terminateProcess(p); err != nil {
```
AdrianPedriza, line 952: nit: how about extract `p.IsProcessSessionLeader()` here and create two different termination methods?: `terminateProcessByKill` vs `terminateProcessBySignal` and use it inside `terminateChildProcess` or something like that? in order to not check recursively if is sessionLeader for child processes, it is always false, right? that's my main concern

AdrianPedriza:
tested several use cases and works fine 👍

cmd/up/up.go
andreafalzetti: @AdrianPedriza we should check each process during the recursion because you might have a child process that is also session leader. The simplest case is when you run `bash` inside `bash` you get two processes that are session leaders.

Example (on MacOS from `zsh` I run `bash,` and from it I run `bash` again)

```
❯ bash
bash-5.2$ echo $$
24513
bash-5.2$ bash
bash-5.2$ echo $$
25704
```

Process tree:

```
-+= 24513 andrea bash
 \--= 25704 andrea bash
```

Process details:

```
❯ ps -j 24513
USER     PID  PPID  PGID   SESS JOBC STAT   TT       TIME COMMAND
andrea 24513 24102 24513      0    1 S+   s000    0:00.01 bash

❯ ps -j 25704
USER     PID  PPID  PGID   SESS JOBC STAT   TT       TIME COMMAND
andrea 25704 24513 25704      0    1 S+   s000    0:00.01 bash
```

In both cases PGID === PID, which makes them process leaders, in fact if I run: `﻿kill -SIGTERM 24513` or `﻿kill -SIGTERM 25704` nothing happens

andreafalzetti:
Moving to draft to follow-up about the process being suspended

viddo:
> Moving to draft to follow-up about the process being suspended

Just realized - the reproduce case when using `upgrade-dev` is probably the easiest/fastest way to find a smaller reproduce/fix? Maybe can find some faster way to provoke the "Connection lost" of dev container and process being put into background. Not sure what `upgrade-dev` does internally but it must be some step at the end that cause the disconnection and effect on running processes, e.g. restarting pod manually using kubectl maybe?

PR #3985, opened 2023-09-21
New option `--reuse-params` at `okteto pipeline deploy`

teresaromero:
# Proposed changes

Resolves https://github.com/okteto/app/issues/6997 

This PR enables a new flag/option at the command `okteto pipeline deploy` where the user can redeploy (as in the UI) an existent dev environment.

## Changes
- Add flag `--reause-param`of type boolean with default value `false`
- Refactor at `ExecuteDeployPipeline` to get configmap for both `ReuseParam` or `SkipIfExists`
- refactor `setEnvsFromDependency` to inject configmap instead of calling from the function and also injecting the env setter to be able to mock under tests, updated tests
- new functions to parse configmap data into deployOptions to be able to trigger the pipeline with the overriden options from the configmap
- unit test for new functions and updated ones
- new integration test `TestPipelineDeployWithReuse` where is also testing the use of Labels at the pipeline deploy

## How to validate

1. run `okteto pipeline deploy` with any repository and wait for the dev environment is ready
1. run `okteto pipeline deploy --reuse-params` this would retrigger the pipeline with the same params the first was triggered.

Note: is more "visual" to deploy a pipeline with some labels, when not using `--reuse-params` the labels disapear from the pipeline if its retriggered without them. When the flag is used, as the deployment reuses the configmap values, the labels would still be there.

## CLI Quality Reminders 🔧


cmd/pipeline/deploy.go
```diff
@@ -142,61 +146,70 @@ func (pc *Command) ExecuteDeployPipeline(ctx context.Context, opts *DeployOption
...
-		cfgName := pipeline.TranslatePipelineName(opts.Name)
-
-		cfg, err := configmaps.Get(ctx, cfgName, opts.Namespace, c)
-		if err != nil && !oktetoErrors.IsNotFound(err) {
+	exists := false
+	cfgName := pipeline.TranslatePipelineName(opts.Name)
+	cfg, err := configmaps.Get(ctx, cfgName, opts.Namespace, c)
+	if err != nil {
+		if opts.ReuseParams && oktetoErrors.IsNotFound(err) {
+			return fmt.Errorf("Dev Environment not found to reuse params. Deploy your pipeline without --reuse-params")
```
teresaromero, line 154: i would need some 👁️  for this copy. Intent is to inform the user that if using the flag but no configmap is found, then they should not use the flag and make a regular `okteto pipeline deploy`

cc @omnomagonz for some copy 🪄 magic

cmd/pipeline/deploy.go
```diff
@@ -38,13 +38,17 @@ import (
...
-	"k8s.io/client-go/kubernetes"
+	v1 "k8s.io/api/core/v1"
 )
 
 const (
 	dependencyEnvTemplate = "OKTETO_DEPENDENCY_%s_VARIABLE_%s"
 )
 
+var (
+	errUnableToReuseParams = errors.New("development environment not found: unable to use --reuse-params option")
```
teresaromero, line 49: i would need some 👁️ for this copy. Intent is to inform the user that if using the flag but no configmap is found, then they should not use the flag and make a regular okteto pipeline deploy

cc @omnomagonz for some copy 🪄 magic

omnomagonz:
```suggestion
	errUnableToReuseParams = errors.New("configmap not found: unable to use --reuse-params option")
```

Maybe I'm not understanding this context but would this suggestion make sense?

I'm assuming the original copy said "development environment" since that's the user-facing resource that's "not found", but would it be more clear to say what we mean about the missing configmap?

"More clear" here could mean easier to tell us "Hey I'm getting this error about a missing configmap when trying to pipeline deploy"

Please correct me if I'm missing the mark 🙏

teresaromero: thanks Matt! The configmap is the k8s component we use to "model" a dev environment. This stores the data that "defines" a dev environment in terms of repository, branch, etc... so... technical interface is a configmap but "popular" name would be dev environment, also known as gitdeploy....

teresaromero:
Closing this PR as pipeline deploy was not intended to be the command running this change.

teresaromero:
Reopening after offline conversation, the flag should be triggered by `okteto pipeline deploy` to reuse the params. An other issue will be filled to implement this redeploy behavour into the UI redeploy

cmd/pipeline/deploy.go
```diff
@@ -142,61 +150,70 @@ func (pc *Command) ExecuteDeployPipeline(ctx context.Context, opts *DeployOption
...
+	}
+	exists = cfg != nil && cfg.Data != nil
 
-				if !opts.Wait && cfg.Data["status"] == pipeline.ProgressingStatus {
-					oktetoLog.Success("Repository '%s' already scheduled for deployment", opts.Name)
-					return nil
-				}
+	if opts.ReuseParams && exists {
+		if overrideOptions := cfgToDeployOptions(cfg); overrideOptions != nil {
+			applyOverrideToOptions(opts, overrideOptions)
```
AdrianPedriza, line 168:
```suggestion
			opts = overrideOptions
```
can we do it like this suggestion?

integration/okteto/pipeline_test.go
```diff
@@ -68,3 +72,61 @@ func TestPipelineCommand(t *testing.T) {
 	}
 	require.NoError(t, commands.RunOktetoPipelineDestroy(oktetoPath, pipelineDestroyOptions))
 }
+
+func TestPipelineDeployWithReuse(t *testing.T) {
```
AdrianPedriza, line 76: nice test 👍

AdrianPedriza (approved):
working as expected, just minor comments

cmd/pipeline/deploy.go
teresaromero: i did it with the function to select which ones to override. There are other options set before this that could be changed to empty if i assign the value directly...

cmd/pipeline/deploy.go
omnomagonz: @teresaromero Thanks so much for the clarification 🙇 

If that's the case then I'm not sure I'd suggest a change here. It seems your original copy conveys what's happening at least in this context.

Eventually we probably want a specific error code but I know that's a separate, documented initiative that lives elsewhere 😄 

I'd say we ship this as-is unless someone else objects 👍

jLopezbarb (approved):

PR #4001, opened 2023-10-02
Implement new event ImageBuild

teresaromero:
# Proposed changes

Resolves https://okteto.atlassian.net/browse/LAKE-46

We want to track every image built using okteto. This means any other build run from the build command itself, or any other command such as deploy, up, destroy, stack that requires the image to be built prior to execute itself.

We want to measure the cachehit when looking into the image and skipping the build in order to make some improvements when using the commit of the build context instead on the commit at the repository itself. For this to happen the metadata and tracking of the event is placed at the V2 Builder Build funcion.

Proposed changes are the following:
- create a new event at `analytics` pkg where we locate the props to send and the event name. Required data is sent hashed and its the tracker who is hashing the information.
- Inject the analyticsTracker into v2 Builder Build though a new property at OktetoBuilder struct `analyticsTracker`. This has caused some refactor around the use of v2 Builder from other commands, as the injection of the tracker is done at `NewBuilderFromScratch`
  - deploy, destroy, init, up, stack commands now include also an analyticsTracker interface and tracker injection so when the build is done we can track this.
- retrieving the metadata for the event has resulted on some refactoring
  - `meta.RepoURL` - repository was not available under the `bc.Confing` object, so made the changes to make it available
  - `meta.RepoHash` - is using the build hash, changed the function that calculates this from being inside the `Config` object into a local function renamed to `getBuildHashFromCommit`
- refactor of OktetoBuilder.areImagesBuilt into local function areAllServicesBuilt using a control map instead of the buildImages property aat the OktetoBuilder struct. The control map is only used locally on the loop so there is no need to tie it to the struct. Also created a local function so its more readeable the condtion `isServiceBuilt` which looks at the control map to check it the service has been built
- refactor around the imageTagger as i found it really difficult to follow what was going on at each function. Mainly the name "sha" used to refer to "tag" on an image referece. 

> The SHA of an image is a unique identifyer and at a docker image is reference like [name]@sha256:[sha] ; on the other hand a tag is a label assigned to the image, which is referenced as [name]:[tag]. Multiple image references with different tags, can be referencing a single SHA.

  - using `tag` when refering to the tag of an image reference: [name]:[tag]
  - rename function `tag` into `getServiceImageReference` as the function returns the image reference in the format [name]:[tag]
  - rename `getImageFromTmpl` into `useReferenceTmplate`
  - rename `getPossibleTags` into `getImageReferencesForTagWithDefaults` as the function returned the references with the tag provided and the default okteto tags
  - rename `getPossibleHashImages` into `getImageReferencesForTag` as the function returned the references wth the tag provided, being this a hash or not.
  - rename `getImageFromTmplWithVolumesAndSHA` into `useReferenceTemplateWithVolumes` as the tag provided could be a hash or a regular tag, either way sha is not the type of value passed here.
  
- refactor at `imageChecker` as in the tagger, i found it difficult so i renamed and added couple comments in order to make it more readeable
  - removed dependency with `GetBuildHash` and now is a param on the functions that need it
  - renaming of sha to tag, and tag to reference
  - rename `getImageSHA` into `lookupReferenceWithDigest` as this is the value returned. An image reference with digest has the following format [name]:[tag]@sha256:[sha/digest]
  - rename `checkIfCommitHashIsBuilt` into `checkIfBuildHashIsBuilt` as the value that is being using for tagging the iamges is the buildHash and not the commit itself, this can be missleading.
  - rename `getImageDigestFromAllPossibleTags` to `getImageDigestReferenceForService` as this is returning the references of images for a service.

- refactor at `GetServicesToBuild`. This function is the link of the build command with other commands. It returns a list of services that need to be built, and for that it checks if they are already built.
  - rename tha chan from `toBuild` into `toBuildCh`
  - rename `checkServicesToBuild` into `checkServiceToBuild` as the function itself checks one by one so its no sense to use the plural
  - inject the `buildHash` into refactored `getImageDigestReferenceForService`

## How to validate

- Using cloud/staging environment, pull first approach is disabled, testing can be done using this context to check that the events are generated and sent
- Using dev cluster, check the metadata for pull first aproach is correctly retrieved

- Create a new namespace and run okteto build for a repository/project to check that on every build an event is sent with the metadata.
- Go to mixpanel to check the event imagebuild filtering by version (local binary commit) or add a printlin at the TrackImageBuild function to see the event being sent with the props if using ur dev cluster.
- Run up, deploy, destroy, stack and all the commands that builder is injected to check that the images are built and the event is sent

## Additional context for reviewers

Attaching diagrams used for the analysis and implementation of the event on how build is done and related with other commands

Builders

Commands using build

## CLI Quality Reminders 🔧


cmd/stack/deploy.go
```diff
@@ -116,8 +116,9 @@ func (c *DeployCommand) RunDeploy(ctx context.Context, s *model.Stack, options *
 	}
 
 	stackDeployer := &stack.Stack{
-		K8sClient: c.K8sClient,
-		Config:    c.Config,
```
jLopezbarb, line 120: Maybe it can come from the main.go file

cmd/build/v2/build.go
```diff
@@ -167,56 +180,92 @@ func (bc *OktetoBuilder) Build(ctx context.Context, options *types.BuildOptions)
...
+			meta.RepoHash = buildHash
+			meta.RepoHashDuration = time.Since(repoHashDurationStart)
+
 			// We only check that the image is built in the global registry if the noCache option is not set
 			if !options.NoCache && bc.Config.IsCleanProject() {
+
 				imageChecker := getImageChecker(buildSvcInfo, bc.Config, bc.Registry)
-				if imageWithDigest, isBuilt := imageChecker.checkIfCommitHashIsBuilt(options.Manifest.Name, svcToBuild, buildSvcInfo); isBuilt {
-					oktetoLog.Information("Skipping build of '%s' image because it's already built for commit %s", svcToBuild, bc.Config.GetGitCommit())
+				cacheHitDurationStart := time.Now()
```
AdrianPedriza, line 201: I think `cacheHitDurationStart` should be time to check if some hash is on the available registries but here there is more logic before inside `checkIfBuildHashIsBuilt` func. I would start tracking it in the loop that checks if the digest is in the possible registries

cmd/build/v2/build.go
```diff
@@ -167,56 +180,92 @@ func (bc *OktetoBuilder) Build(ctx context.Context, options *types.BuildOptions)
...
 					continue
 				}
 			}
 
 			if !okteto.Context().IsOkteto && buildSvcInfo.Image == "" {
+				bc.analyticsTracker.TrackImageBuild(buildsAnalytics...)
 				return fmt.Errorf("'build.%s.image' is required if your cluster doesn't have Okteto installed", svcToBuild)
 			}
-
+			buildDurationStart := time.Now()
```
AdrianPedriza, line 233: same as before, here we don't know what is inside `buildService` func. If we need to check build duration I would track the time we used until external lib call finish the build

cmd/manifest/init-v2.go
```diff
@@ -42,10 +43,15 @@ import (
 	"github.com/spf13/cobra"
 )
 
+type analyticsTrackerInterface interface {
```
AdrianPedriza, line 46: should we move it to analytics package to hold this instead of declaring an interface in each command?

cmd/up/up.go
```diff
@@ -338,7 +339,8 @@ func Up() *cobra.Command {
 				}()
 			}
 
-			if err := setBuildEnvVars(ctx, oktetoManifest); err != nil {
+			// build images and set env vars for the services at the manifest
+			if err := buildAllServices(ctx, oktetoManifest, up.builder); err != nil {
```
AdrianPedriza, line 343: if I call `buildAllService` I expect to build all services but inside the func there is logic to check which services needs to be build. Can we rename the func -> `buildServices` or `buildNeededServices` ?

AdrianPedriza:
In my opinion, any refactor that is not needed for the change should be eliminated. We could address it separately because it makes the review of the PR quite complex and implies more risk in it

cmd/up/up.go
teresaromero: agree on the "all" i was thinking on all as iteration on all the services of the manifest, but the build can be done/not if is already built. I've renamed and modified the comment

cmd/manifest/init-v2.go
teresaromero: I made this decision based on the current implementation of the AnalyticsTracker, which holds the methods to track different commands/events.

With this context, i preferred to declare when i needed the interface with just the methods i will need in that package, a benefit of this is that when mocking this on the tests i only take into account the method that is being used on this package from the whole implementation. Also it makes it more modular between interface/implementation, although the downsize is to declare the interface when the implementation is injected on every package.

cmd/build/v2/build.go
teresaromero: I've added a comment to the function. This is where the built is done, either a single one, or 2 due to the volume mounts. 

If an error occurs the analytics does not include buld time (no build)

@pchico83 what should we track when building with volume mounts? should perhaps a boolean that indicates that volumes have been added? so the time would be of 2 builds?

pchico83: @teresaromero ideally, it should be a single event. but if it simplifies the logic, we can track two independent build events. Or ignore the builds due to volume mounts

cmd/build/v2/build.go
teresaromero: i see your point but the process of checking if we already have the image at any registry is running the loop of the different options, the whole step of the build is to get the references and iterate to see if they are at each registry.

cmd/build/v2/build.go
```diff
@@ -167,56 +180,92 @@ func (bc *OktetoBuilder) Build(ctx context.Context, options *types.BuildOptions)
 
 			buildSvcInfo := buildManifest[svcToBuild]
 
+			// create the meta pointer and append it to the analytics slice
```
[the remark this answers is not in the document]
teresaromero: > This is something that is only useful if we want to track it.

Do you mean to create the pointer or to the calculation of the properties? Most of them are inputs to the functions of the process, the only ones "optional" would be the duration ones AFAIK

andreafalzetti:
> In my opinion, any refactor that is not needed for the change should be eliminated. We could address it separately because it makes the review of the PR quite complex and implies more risk in it

If the refactor is needed for testing the change, I am in favour of keeping in this PR. In the future we could try splitting in 2 Prs (refactor + change) but it's not easy!

cmd/build/v2/build.go
teresaromero: I've been going though this and right now it will trigger a single event.

There are cases where there is just 1 build that is with the volume mounts, or just 1 build with a dockerfile, or both one after the other...
I think its valueable to have the event to track the build time, and a property that also inform if is with volumes or not to be able to distinguish between building times....

cmd/build/v2/build.go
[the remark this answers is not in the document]
teresaromero: after giving a thought on this, i've moved the send of the metrics at a defer when the functon returns (err or success). buildsAnalytics will collect the meta for each iteration of the for loop (each service built) and send them when the process is exited.

the collection of metadata when analytics is not active would need a further refactor, as we do have logic collecting metrics that dont look up this. This check is done prior to make the call to send the data, so i think is worth to create a focus issue on improving the analytics pkg for all the events we have.

cmd/stack/stack.go
```diff
@@ -17,18 +17,23 @@ import (
...
 	"github.com/spf13/cobra"
 )
 
+type analyticsTrackerInterface interface {
+	TrackImageBuild(...*analytics.ImageBuildMetadata)
+}
+
 // Stack stack management commands
-func Stack(ctx context.Context) *cobra.Command {
+func Stack(ctx context.Context, af analyticsTrackerInterface) *cobra.Command {
```
AdrianPedriza, line 29:
```suggestion
func Stack(ctx context.Context, at analyticsTrackerInterface) *cobra.Command {
```
nit

cmd/build/v2/tagger.go
```diff
@@ -46,110 +44,122 @@ func getTargetRegistries() []string {
...
-func getImageFromTmpl(targetRegistry, repoName, svcName, tag string) string {
+// useReferenceTmplate returns the image reference with the given parameters [name]:[tag]
+func useReferenceTmplate(targetRegistry, repoName, svcName, tag string) string {
 	return fmt.Sprintf("%s/%s-%s:%s", targetRegistry, repoName, svcName, tag)
 }
 
-// getImageFromTmpl returns the image name from the template of image sha
-func getImageFromTmplWithVolumesAndSHA(targetRegistry, repoName, svcName, sha string) string {
-	return fmt.Sprintf("%s/%s-%s:%s-%s", targetRegistry, repoName, svcName, model.OktetoImageTagWithVolumes, sha)
+// useTmplateWithVolumes returns the image reference from the template of image with volume mounts [name]:okteto-with-volume-mounts-[tag]
```
AdrianPedriza, line 162:
```suggestion
// useReferenceTemplateWithVolumes returns the image reference from the template of image with volume mounts [name]:okteto-with-volume-mounts-[tag]
```

cmd/build/v2/tagger.go
```diff
@@ -46,110 +44,122 @@ func getTargetRegistries() []string {
...
-		tags = append(tags, getImageFromTmpl(targetRegistry, sanitizedName, svcToBuildName, model.OktetoImageTagWithVolumes))
+		tags = append(tags, useReferenceTmplate(targetRegistry, sanitizedName, svcToBuildName, model.OktetoImageTagWithVolumes))
 	}
 	return tags
 }
 
-// getImageFromTmpl returns the image name from the template of image tag
-func getImageFromTmpl(targetRegistry, repoName, svcName, tag string) string {
+// useReferenceTmplate returns the image reference with the given parameters [name]:[tag]
+func useReferenceTmplate(targetRegistry, repoName, svcName, tag string) string {
```
AdrianPedriza, line 158:
```suggestion
// useReferenceTemplate returns the image reference with the given parameters [name]:[tag]
func useReferenceTemplate(targetRegistry, repoName, svcName, tag string) string {
```

cmd/build/v2/tagger.go
```diff
@@ -46,110 +44,122 @@ func getTargetRegistries() []string {
...
+		return useReferenceTmplate(targetRegistry, sanitizedName, svcName, tag)
 	}
-	return b.Image
+	return useReferenceTmplate(targetRegistry, sanitizedName, svcName, model.OktetoDefaultImageTag)
 }
 
-// getPossibleHashImages returns all the possible images that can be built from a commit hash
-func (i imageTagger) getPossibleHashImages(manifestName, svcToBuildName, sha string) []string {
-	if sha == "" {
+// getPossibleImageReferenceWithTag returns all the possible images references that can be used for build with the given tag
```
AdrianPedriza, line 82:
```suggestion
// getImageReferencesForTag returns all the possible images references that can be used for build with the given tag
```

cmd/build/v2/tagger.go
```diff
@@ -46,110 +44,122 @@ func getTargetRegistries() []string {
...
-	if shouldBuildFromDockerfile(b) && b.Image == "" {
-		if sha != "" {
-			return getImageFromTmpl(targetRegistry, sanitizedName, svcName, sha)
-		}
-		return getImageFromTmpl(targetRegistry, sanitizedName, svcName, model.OktetoDefaultImageTag)
+	if serviceHasDockerfile(b) && tag != "" {
+		return useReferenceTmplate(targetRegistry, sanitizedName, svcName, tag)
 	}
-	return b.Image
+	return useReferenceTmplate(targetRegistry, sanitizedName, svcName, model.OktetoDefaultImageTag)
```
AdrianPedriza, line 79: I think we need to check in this call to `useReferenceTmplate` that `serviceHasDockerfile(b) == true` to follow previous approach

cmd/build/v2/build.go
```diff
@@ -167,16 +185,37 @@ func (bc *OktetoBuilder) Build(ctx context.Context, options *types.BuildOptions)
...
+			meta := analytics.NewImageBuildMetadata()
+			buildsAnalytics = append(buildsAnalytics, meta)
+
+			meta.Name = svcToBuild
+			meta.RepoURL = bc.Config.GetAnonymizedRepo()
```
AdrianPedriza, line 193: in the issue description it says `should be hashed and not readable` . Do we need to hash the value for `name` and `repoUrl`(already anonymized but not hashed) ?

AdrianPedriza:
e2e works as expected and the code looks good, good work! 👍  Just waiting for the latest added comments to approve this PR

cmd/build/v2/build.go
teresaromero: i checked that when running the debugger i get the file root of the repository, but when running binary is the repo.

teresaromero: the hashing is done in the analyrics https://github.com/okteto/okteto/pull/4001/files#diff-5d815810b1f75652c01329c28da0b4fa4c0720e47d31f2c5447c19c52333deb6R37

cmd/build/v2/tagger.go
teresaromero: yes! i have fix this refactor as i was returning the b.Image just when this was not empty, but also, this is the default value if there is no dockerfile in the service. I've uploaded changes.

the change here is that b.Image was returned at the end if no infer was needed, but i changed the order and the output is the same

andreafalzetti (approved):

PR #4003, opened 2023-10-04
Deploy remote extra hosts

maroshii:
# Proposed changes

Allows deploy remote to to talk to the internal IP of the ingress. It does so by adding and `ExtraHosts` to buildOptions to be injected into buildkit solve options. If extra hosts are defined, we use the `gateway.v0` frontend which allows this config to be injected as attributes.

Still missing test Opening early to get reviews from the team

## How to validate

1. Remove .Values.clusterEndpoint from your okteto cluster and run an upgrade
1. Clone the go-getting-started repo
1. Add `cat /etc/hosts` as a deploy step
1. The deploy succeeds 
1. See in the output that all know okteto hosts are mapped to the internal IP

## CLI Quality Reminders 🔧

### DIFF
--- a/cmd/deploy/remote.go
+++ b/cmd/deploy/remote.go
@@ -196,7 +196,10 @@ func (rd *remoteDeployCommand) deploy(ctx context.Context, deployOptions *Option
 	registryUrl := okteto.Context().Registry
 	subdomain := strings.TrimPrefix(registryUrl, "registry.")
 
-	ip, _, _ := net.SplitHostPort(sc.ServerName)
+	ip, _, err := net.SplitHostPort(sc.ServerName)
+	if err != nil {
+		return fmt.Errorf("failed to parse server name network address: %w", err)
+	}
 	buildOptions.ExtraHosts = []types.HostMap{
 		{Hostname: registryUrl, IP: ip},
 		{Hostname: fmt.Sprintf("kubernetes.%s", subdomain), IP: ip},
--- a/cmd/deploy/remote.go
+++ b/cmd/deploy/remote.go
@@ -193,16 +193,17 @@ func (rd *remoteDeployCommand) deploy(ctx context.Context, deployOptions *Option
 		fmt.Sprintf("%s=%d", constants.OktetoInvalidateCacheEnvVar, int(randomNumber.Int64())),
 	)
 
-	registryUrl := okteto.Context().Registry
-	subdomain := strings.TrimPrefix(registryUrl, "registry.")
-
-	ip, _, err := net.SplitHostPort(sc.ServerName)
-	if err != nil {
-		return fmt.Errorf("failed to parse server name network address: %w", err)
-	}
-	buildOptions.ExtraHosts = []types.HostMap{
-		{Hostname: registryUrl, IP: ip},
-		{Hostname: fmt.Sprintf("kubernetes.%s", subdomain), IP: ip},
+	if sc.ServerName != "" {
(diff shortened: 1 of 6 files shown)


cmd/deploy/remote.go
```diff
@@ -191,6 +193,17 @@ func (rd *remoteDeployCommand) deploy(ctx context.Context, deployOptions *Option
...
 	)
 
+	registryUrl := okteto.Context().Registry
+	subdomain := strings.TrimPrefix(registryUrl, "registry.")
+
+	ip, _, _ := net.SplitHostPort(sc.ServerName)
+	buildOptions.ExtraHosts = []types.HostMap{
+		{Hostname: registryUrl, IP: ip},
+		{Hostname: fmt.Sprintf("kubernetes.%s", subdomain), IP: ip},
+		{Hostname: fmt.Sprintf("okteto.%s", subdomain), IP: ip},
```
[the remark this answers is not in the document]
maroshii: ahh you're right 🤦

cmd/deploy/remote.go
```diff
@@ -191,6 +193,17 @@ func (rd *remoteDeployCommand) deploy(ctx context.Context, deployOptions *Option
...
+		{Hostname: fmt.Sprintf("buildkit.%s", subdomain), IP: ip},
```
ifbyol, line 204: BuildKit it is not always exposed behind our ingress controller. If BuildKit it's in its own load balancer, this wouldn't work.

I don't know/recall what information is retrieved from the backend with the cluster metadata, but we should return an specific field with the BuildKit private IP. In some configurations, it will be the same but in other cases, it will be different.

If the server already returns that IP, lets change it, if not, lets not do it as part of this PR and create another ticket for that specific scenario, so at least we fix the scenario for registry, kubernetes and API which are the most common scenario.

cmd/deploy/remote.go
```diff
@@ -191,6 +193,17 @@ func (rd *remoteDeployCommand) deploy(ctx context.Context, deployOptions *Option
 		fmt.Sprintf("%s=%d", constants.OktetoInvalidateCacheEnvVar, int(randomNumber.Int64())),
 	)
 
+	registryUrl := okteto.Context().Registry
+	subdomain := strings.TrimPrefix(registryUrl, "registry.")
+
+	ip, _, _ := net.SplitHostPort(sc.ServerName)
```
ifbyol, line 199: I assume we need this for `cmd/destroy/remote.go` too?

ifbyol (changes requested):

ifbyol:
Glad to see we didn't have to update BuildKit client to make it work :)

cmd/deploy/remote.go
maroshii: I'll remove `okteto.*` (may have public override) and `buildkit.*` (may be in separate LB) for now and just keep `registry.*` and `kubernetes.*`

ifbyol: Ok, I think `okteto.*` might be simpler because the CLI should have a reference to the public domain (even if it is a override) as it has to talk with it. But I'm ok with leaving it for other ticket. Mentioning it just to give context

ifbyol:
I'm doing a test with an instance with TLS termination in the LoadBalancer which was the scenario affected by this before final approval, but changes looks good after addressing the comments

ifbyol (approved):

maroshii:
@jpf-okteto can you take another look 🙏

PR #4005, opened 2023-10-05
feat: add smart builds

AdrianPedriza:
## Description.
This PR pretends to add smart builds which means to manage each service in a repository individually instead of the current approach which is that the build related of a service is based on the repository commit. With this new approach, build context is check and only if the build context related to a service has changes then the image will be build again, but only for that service.

## How to test it
1. using okteto/movies with no changes
2. add a new file (`touch test.txt`) into `rent` folder.
3. run `okteto build` 
4. check images are build for the new generated hash
5. build again (`okteto build`)
6. check only `rent` service is built again because is the one with changes


cmd/build/v2/build.go
```diff
@@ -167,9 +166,10 @@ func (bc *OktetoBuilder) Build(ctx context.Context, options *types.BuildOptions)
 
 			buildSvcInfo := buildManifest[svcToBuild]
 
+			serviceContext := getServiceContext(buildSvcInfo)
```
jLopezbarb, line 169: What is serviceContext? What is the purpose of it? Should we make the name of the function clearer?

cmd/build/v2/servicecontext.go
```diff
@@ -0,0 +1,38 @@
...
+	oktetoLog "github.com/okteto/okteto/pkg/log"
+	"github.com/okteto/okteto/pkg/model"
+	"github.com/okteto/okteto/pkg/repository"
+)
+
+type serviceContextInterface interface {
+	IsCleanBuildContext() bool
+}
+
+type serviceContext struct {
```
jLopezbarb, line 14: I'd like to know what this does from the struct. Maybe `SvcContextCleanlinessChecker`

cmd/build/v2/servicecontext.go
```diff
@@ -0,0 +1,38 @@
...
+	isClean bool
+}
+
+func (sc *serviceContext) IsCleanBuildContext() bool {
+	return sc.isClean
+}
+
+func getServiceContext(buildInfo *model.BuildInfo) *serviceContext {
```
jLopezbarb, line 22: If we are initializing, Should we name this as `newServiceContext`?

cmd/build/v2/servicecontext.go
```diff
@@ -0,0 +1,38 @@
...
+	wdCtrl := filesystem.NewOsWorkingDirectoryCtrl()
```
jLopezbarb, line 23: Should we reuse the one from the build? That way we can test from an integration test if the `wd` is the same

cmd/build/v2/servicecontext.go
```diff
@@ -0,0 +1,38 @@
...
+	wd, err := wdCtrl.Get()
+	if err != nil {
+		oktetoLog.Infof("could not get working dir: %w", err)
+	}
+	gitRepo := repository.NewRepository(wd)
```
jLopezbarb, line 28: Same, should we inject the repository from the build and not creating it again? It will let us test with integration tests in the future

pkg/repository/git.go
```diff
@@ -64,6 +64,25 @@ func (r gitRepoController) calculateIsClean(ctx context.Context) (bool, error) {
 	return status.IsClean(), nil
 }
 
+func (r gitRepoController) calculateIsCleanBuildContext(ctx context.Context, buildContext string) (bool, error) {
```
jLopezbarb, line 67: Can we reduce the naming of the function? I don't think the build context should be referenced in the name/param of the function. 

I'd name it `calculateIsCleanDir` or something similar. That way is more reusable

pkg/repository/git.go
```diff
@@ -100,6 +119,42 @@ func (r gitRepoController) isClean(ctx context.Context) (bool, error) {
 	return s.isClean, s.err
 }
 
+// isClean checks if the repository have changes over the commit
+func (r gitRepoController) isBuildContextClean(ctx context.Context, buildContext string) (bool, error) {
```
jLopezbarb, line 123: Same as previous comment about `buildContext`. gitRepo doesn't know what is a `buildContext`. Let's use something that is into `gitrepo` model

pkg/repository/git.go
```diff
@@ -196,3 +252,25 @@ func (ogr oktetoGitWorktree) Status(ctx context.Context, localGit LocalGitInterf
...
 	return oktetoGitStatus{status: status}, nil
 }
+
+func (ogr oktetoGitWorktree) BuildContextStatus(ctx context.Context, localGit LocalGitInterface, buildContext string) (oktetoGitStatus, error) {
+	// using git directly is faster, so we check if it's available
+	_, err := localGit.Exists()
+	if err != nil {
+		// git is not available, so we fall back on git-go
+		oktetoLog.Debug("Calculating git status: git is not installed, for better performances consider installing it")
+		status, err := ogr.worktree.Status()
```
jLopezbarb, line 262: How does this know if this is referred for a specific context. I can't see clear where it's specifying it

jLopezbarb: I mean using the external dependency

AdrianPedriza: yes, we need to filter only relevant status related to the service context

AdrianPedriza: d6723152baa1fda86f4cf0a53fdfb4da9db1613e

cmd/build/v2/servicecontext.go
AdrianPedriza: eecbe80e10436841929955ebfd994cd41629b530

cmd/build/v2/servicecontext.go
AdrianPedriza: 93cf93c3273f887c4d4bdd3fbad218cfaf667240

pkg/repository/git.go
AdrianPedriza: cf69c7a86197f9bcf2929d9ca9aa1940200334ea

pkg/repository/git.go
AdrianPedriza: cf69c7a86197f9bcf2929d9ca9aa1940200334ea

cmd/build/v2/servicecontext.go
AdrianPedriza: c4ade2c07260cf3dd979a7e0c69bc6bcda87c71d

cmd/build/v2/servicecontext.go
AdrianPedriza: c4ade2c07260cf3dd979a7e0c69bc6bcda87c71d

cmd/build/v2/servicecontext.go
```diff
@@ -0,0 +1,29 @@
...
+
+type SvcContextCleanlinessChecker struct {
+	isClean bool
+}
+
+func (sc *SvcContextCleanlinessChecker) IsCleanBuildContext() bool {
+	return sc.isClean
+}
+
+func newSvcContextCleanlinessChecker(buildInfo *model.BuildInfo, isCleanServiceBuild func(string) (bool, error)) *SvcContextCleanlinessChecker {
```
teresaromero, line 20: can we add comments to new functions?

cmd/build/v2/servicecontext.go
```diff
@@ -0,0 +1,29 @@
...
+import (
+	oktetoLog "github.com/okteto/okteto/pkg/log"
+	"github.com/okteto/okteto/pkg/model"
+)
+
+type serviceContextInterface interface {
+	IsCleanBuildContext() bool
+}
+
+type SvcContextCleanlinessChecker struct {
```
teresaromero, line 12: what is clean lines? :/

pkg/constants/constants.go
```diff
@@ -137,4 +137,10 @@ const (
 
 	// EnvironmentLabelKeyPrefix represents the prefix for the preview and pipeline labels
 	EnvironmentLabelKeyPrefix = "label.okteto.com"
+
+	// OktetoServiceBuildHash represents the template to store a build context hash related to a service
+	OktetoServiceBuildHash = "OKTETO_SERVICE_%s_BUILD_HASH"
+
+	// OktetoServiceBuildHash represents the template to store if a service is clean
```
andreafalzetti, line 144:
```suggestion
	// OktetoServiceIsClean represents the template to store if a service is clean
```

cmd/build/v2/build.go
```diff
@@ -113,6 +113,30 @@ func (*OktetoBuilder) IsV1() bool {
...
 }
 
+func (ob *OktetoBuilder) newServiceContext(buildInfo *model.BuildInfo, repoService repository.RepositoryService) serviceContextChecker {
+	service := buildInfo.Name
+	isClean, err := repoService.IsBuildContextClean(buildInfo.Context)
+	if err != nil {
+		oktetoLog.Infof("failed to check if service '%s' has build context clean: %w", service, err)
+	}
+	ob.lock.Lock()
+	ob.buildEnvironments[fmt.Sprintf(constants.OktetoServiceIsClean, service)] = strconv.FormatBool(isClean)
```
andreafalzetti, line 123: why do we expose this env var?

cmd/build/v2/config.go
```diff
@@ -70,53 +58,3 @@ func (oc oktetoBuilderConfig) IsOkteto() bool {
 func (oc oktetoBuilderConfig) HasGlobalAccess() bool {
 	return oc.hasGlobalAccess
 }
-
-// IsCleanProject checks if the repository is clean(no changes over the last commit)
-func (oc oktetoBuilderConfig) IsCleanProject() bool {
```
teresaromero, line 75: why are we removing this? we still need to track the project hash

pkg/repository/git.go
```diff
@@ -17,16 +17,21 @@ import (
...
 	oktetoLog "github.com/okteto/okteto/pkg/log"
 
 	"github.com/go-git/go-git/v5"
 	"github.com/go-git/go-git/v5/plumbing"
+	"github.com/go-git/go-git/v5/plumbing/object"
 )
 
 var (
 	errNotCleanRepo = errors.New("repository is not clean")
+	errNotCleanDir  = errors.New("directory is not clean")
```
andreafalzetti, line 34: This var is unused, check with `golangci-lint run --new-from-rev d13d0c12d7fcf162c7e9dcc2c13ac3c975bc18b7`

pkg/repository/repository_test.go
```diff
@@ -39,9 +41,25 @@ func (frg *fakeRepositoryGetter) get(_ string) (gitRepositoryInterface, error) {
 }
 
 type fakeRepository struct {
-	worktree *fakeWorktree
-	head     *plumbing.Reference
-	err      error
+	worktree       *fakeWorktree
+	head           *plumbing.Reference
+	commitResponse fakeCommitResponse
+	tree           fakeTreeResponse
```
andreafalzetti, line 47: `tree` seems to be unused

andreafalzetti:
Considering we are changing the way we calculate the hash, this change introduces some sort of breaking change with the registry, all users will have to rebuild their images. Have we considered the impact this may have?

AdrianPedriza:
closing this PR but keeping the branch because atm it is not necessary adopt smart build

PR #4009, opened 2023-10-10
User Friendly Errors

### LINKED ISSUE #2690: User friendly errors when the yaml file can't be deserialized
**Describe the bug**
If I have a badly formatted okteto.yml, I get this error:

```
Invalid manifest:
    - line 2: cannot unmarshal !!str `todo-list` into model.ManifestBuild
    See https://okteto.com/docs/reference/manifest/ for details
```

`model.ManifestBuild` and `!!str`  are not concepts that a regular user would understand. Is it possible to parse that error and show something more user friendly? (e.g. " cannot unmarshal line 2 into a `build` object. See https://www.okteto.com/docs/reference/manifest/#build-object-optional for reference")

- OS: macOS
- Version okteto version 2.3.0-rc.1 

**Additional context**
This is the manifest:
``` 
build:
  todo-list
  
deploy:
  - cd manifest; kustomize edit set image okteto.dev/todo-list=okteto.dev/todo-list:${OKTETO_GIT_COMMIT}
  - kubectl apply -k manifest
dev:
  todo-list:
    image: okteto/golang:1
    command: bash
    securityContext:
      capabilities:
        add:
        - SYS_PTRACE
    volumes:
    - /go/pkg/
    - /root/.cache/go-build/
    sync:
    - .:/app
    forward:
    - 2345:2345
```

andreafalzetti:
Fixes: LAKE-22
Also fixes #2690

# Proposed changes

- Add a new pkg responsible for suggesting user friendly errors
- Integrate new pkg with the `build` command

## How to validate

1. Build the CLI
2. Using any project, try breaking the manifest and run commands such as `build`, `deploy`, `up`, etc
3. Compare the output with the latest stable okteto version and see the improvements against this version

## CLI Quality Reminders 🔧


cmd/build/command.go
```diff
@@ -103,7 +103,12 @@ func Build(ctx context.Context) *cobra.Command {
 				}
 			}
 
-			return builder.Build(ctx, options)
+			err = builder.Build(ctx, options)
+			if err != nil {
+				return suggest.UserFriendlyError(err)
```
jLopezbarb, line 108: I find it odd to transform it here. There are many things that can go wrong in the middle to transform the error here. Maybe we should transform it on the serialization? and check if it's a userError like we do in other parts of the code like [preview list](https://github.com/okteto/okteto/blob/master/cmd/preview/list.go#L103)

go.sum
```diff
@@ -457,6 +457,7 @@ github.com/fullsailor/pkcs7 v0.0.0-20190404230743-d7302db945fa/go.mod h1:KnogPXt
 github.com/fvbommel/sortorder v1.0.1 h1:dSnXLt4mJYH25uDDGa3biZNQsozaUWDSWeKJ0qqFfzE=
 github.com/fvbommel/sortorder v1.0.1/go.mod h1:uk88iVf1ovNn1iLfgUVU2F9o5eO30ui720w+kxuqRs0=
 github.com/garyburd/redigo v0.0.0-20150301180006-535138d7bcd7/go.mod h1:NR3MbYisc3/PwhQ00EMzDiPmrwpPxAn5GI05/YaO1SY=
+github.com/getkin/kin-openapi v0.76.0/go.mod h1:660oXbgy5JFMKreazJaQTw7o+X00qeSyhcnluiMv+Xg=
```
jLopezbarb, line 460: Why do we change go.sum and not go.mod? Can you run go mod tidy in case any version is not sync?

pkg/suggest/rules.go
```diff
@@ -0,0 +1,179 @@
+package suggest
```
jLopezbarb, line 1: Missing copyright

pkg/suggest/rules.go
```diff
@@ -0,0 +1,179 @@
...
+	"errors"
+	"fmt"
+	"github.com/agext/levenshtein"
+	"regexp"
+	"sigs.k8s.io/kustomize/kyaml/yaml"
+	"strings"
+)
+
+// NewStrReplaceRule creates a Rule that finds and replaces a string in the error message
+func NewStrReplaceRule(find, replace string) Rule {
```
jLopezbarb, line 13: Do we need to export all the functions? I think the only one used outside the package is `UserFriendlyError`

jLopezbarb (commented):
I haven't tested yet given that is a draft and probably will change but I put some concerns about the code that it's right now

pkg/suggest/rules.go
andreafalzetti: yes, I've refactored having `UserFriendlyError` so I will adjust before opening the PR :)

cmd/build/command.go
andreafalzetti: this is a good topic to discuss, let me explain my idea: the function `suggest.UserFriendlyError` is responsible for improving all errors (not only manifest), in the near future we might need to improve error messages related to another aspect of the CLI.

I find very intruisive in the code to change literally every single file to wrap error return with a function specific for that error.

This approach is more generic and wrapping the whole CLI instead. I am sure it comes with a downside but from the positive side we have a single point of entrance for this logic.

go.sum
andreafalzetti: Fixed

pkg/suggest/rules.go
```diff
@@ -0,0 +1,113 @@
...
+		replacedMessage := strings.ReplaceAll(e.Error(), find, replace)
+		return errors.New(replacedMessage)
+	}
+
+	return newRule(condition, transformation)
+}
+
+// newRegexRule creates a ruleInterface based on a regex pattern.
+func newRegexRule(pattern string, transform transformFunc) ruleInterface {
+	re := regexp.MustCompile(pattern)
```
maroshii, line 72: We should be careful calling `regex.MustCompile` with a dynamic input since it panics if the regex is invalid. Not sure how we are providing `pattern` here as it seems to be a private function but mentioning just in case

pkg/suggest/suggest.go
```diff
@@ -0,0 +1,58 @@
...
+func (es *errorSuggestion) suggest(err error) error {
+	newErr := err
+	for _, rule := range es.rules {
+		newErr = rule.apply(newErr)
+	}
+	return newErr
+}
+
+// UserFriendlyError generates a user-friendly suggestion for the given error.
+func UserFriendlyError(err error) error {
```
maroshii, line 50: An idea could be to have `UserFriendlyError` be a struct that implements the error interface instead of doing it in place which works but we lose all the context. This way it could be used as a regular error while also allowing to potentially expose extra functionality through type casting and enable the use of `error.Is`

pkg/suggest/rules.go
ifbyol: Agree, the purpose of this PR is to improve the user experience regarding the errors. This will potentially throw a panic which is not very descriptive usually. Couldn't we use `Compile` and handle the error gracefully?

andreafalzetti: Thanks for the feedback, I will address it immediately

pkg/suggest/rules_yaml_test.go
```diff
@@ -0,0 +1,185 @@
...
+					field2 string
+				}
+			}{},
+			expected: map[string][]string{"_": []string{"field1"}},
+		},
+		{
+			name:  "okteto manifest",
+			input: model.Manifest{},
+			expected: map[string][]string{
+				"forward.Forward":            []string{"localPort", "remotePort", "name", "labels"},
```
teresaromero, line 144: nit: according to simplifycompositelit(default) redundant type from array, slice, or map composite literal

here and all the []string inside maps

```suggestion
				"forward.Forward":            {"localPort", "remotePort", "name", "labels"},
```

main.go
```diff
@@ -170,7 +171,7 @@ func main() {
 	err = root.Execute()
 
 	if err != nil {
-		message := err.Error()
+		message := suggest.UserFriendlyError(err).Error()
```
teresaromero, line 174: not sure i agree on having this at the main, so all the error will go though the "suggestor", but not all would be...

Can we throw them from the origin? Related to frans suggestion i think we should raise a UserFriendlyError when is potentially triggered (serializer?) and then that goes around as error until is returned at the client.

pkg/suggest/suggest.go
teresaromero: agree on the struct suggestion, minor comment i think the naming should be perhaps "SuggestionError" as we do have the "UserError" that involves a hint, this new type of error involves a suggestion that is being calculated

main.go
andreafalzetti: @teresaromero This is a good point of dicussion. I look forward to hear more ideas. My point of view is that the "centralized" approach is looking at errors that now we don't want to modify but if in the future we want to extend the "content" we can easily add a new rule, if we change errors at the origin we would need to modify the code in N parts, making each new addition more complicated and in some occasions very intruisive. For example we might want to log "technical" errors inside, but to the user we want to present a different error. 

I also have the feeling that having it here could have an unwanted behaviour but on the other side the benefits above make me prefer the current approach.

pkg/suggest/suggest.go
andreafalzetti: Change done ✅ Thanks @maroshii 🙇 

About renaming, I am not sure about `SuggestionError` because we have `newErrorSuggestion` too 🤔

pkg/suggest/rules_yaml_test.go
andreafalzetti: Updated, thanks ✅

pkg/suggest/rules_test.go
```diff
@@ -0,0 +1,28 @@
...
+
+package suggest
+
+import (
+	"errors"
+	"github.com/stretchr/testify/assert"
+	"testing"
+)
+
+func TestNewLevenshteinRule_ErrWithoutPanic(t *testing.T) {
```
AdrianPedriza, line 22: what is the purpose of this test? Does it assure us that there are no panics in all cases? asking it because there is only one test case

andreafalzetti: It's to make sure we don't panic passing a broken regex

andreafalzetti:
Moving to draft until I integrate the feedback.

pkg/suggest/rules.go
```diff
@@ -0,0 +1,121 @@
...
+// apply executes the rule on the error.
+func (g *rule) apply(err error) error {
+	if g.condition(err) {
+		return g.transformation(err)
+	}
+	return err
+}
+
+// newStrReplaceRule creates a ruleInterface that finds and replaces a string in the error message
+func newStrReplaceRule(find, replace string) ruleInterface {
```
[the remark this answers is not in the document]
andreafalzetti: done ✅

main.go
andreafalzetti: After discussing offline we agreed to change this, the latest commits have this change, please have a look 🙏

pkg/suggest/suggest.go
```diff
@@ -0,0 +1,73 @@
...
+	suggestion *errorSuggestion
+	err        error
+}
+
+// Error allows UserFriendlyError to satisfy the error interface
+func (u *UserFriendlyError) Error() string {
+	if err := u.suggestion.suggest(u.err); err != nil {
+		return err.Error()
+	}
+	return ""
```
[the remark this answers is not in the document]
andreafalzetti: I've changed this in the last commits, please have a look 🙏

pkg/suggest/rules.go
```diff
@@ -0,0 +1,121 @@
...
+		threshold = 1
+	}
+
+	condition := func(e error) bool {
+		if err != nil {
+			return false
+		}
+		matchingErrors := re.FindAllStringSubmatch(e.Error(), -1)
+		for _, matchingError := range matchingErrors {
+			distance := levenshtein.Distance(target, matchingError[2], nil)
```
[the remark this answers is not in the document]
andreafalzetti: I've changed this in the last commits, please have a look 🙏

pkg/suggest/rules.go
```diff
@@ -0,0 +1,121 @@
...
+		}
+		return false
+	}
+
+	transformation := func(e error) error {
+		errorMsg := e.Error()
+		matchingErrors := re.FindAllStringSubmatch(e.Error(), -1)
+
+		for _, matchingError := range matchingErrors {
+			distance := levenshtein.Distance(target, matchingError[2], nil)
```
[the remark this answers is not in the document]
andreafalzetti: I've changed this in the last commits, please have a look 🙏

pkg/suggest/rules_yaml.go
```diff
@@ -0,0 +1,247 @@
...
+// hasLinkToDocs checks if the error already contains a link to our docs
+func hasLinkToDocs(err error) bool {
+	return strings.Contains(err.Error(), "https://www.okteto.com/docs")
+}
+
+// isYamlError check if the error is related to YAML unmarshalling
+func isYamlError(err error) bool {
+	if err == nil {
+		return false
+	}
```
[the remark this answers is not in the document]
andreafalzetti: this is to make sure we don't apply the rule if the pattern provides creates an invalid regex, which returns that `err`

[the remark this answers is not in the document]
andreafalzetti: it's a defensive way to prevent from interrupting the execution or returning any error to the user considered that this is supposed to be transparent for the user

pkg/suggest/rules.go
```diff
@@ -0,0 +1,121 @@
...
+func newLevenshteinRule(pattern string, target string) ruleInterface {
+	re, err := regexp.Compile("(.*?)" + pattern + "(.*)") // Capture everything before and after the pattern
+
+	threshold := 3
+	if len(target) <= threshold {
+		threshold = 1
+	}
+
+	condition := func(e error) bool {
+		if err != nil {
```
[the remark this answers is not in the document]
andreafalzetti: we could check if the pattern is valid outside of this fn? But then we still want to be defensive inside in my opinion

pkg/suggest/rules.go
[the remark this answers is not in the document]
andreafalzetti: > so if distance is < 3 (can we use a const to add more context to this value?)

✅  done!

> should we calculate this rule using go routines and first matching return it? in order to not calculate distance for every key sequently

this could be a performance optimization, not sure if it's worth it, to be sure I've tried comparing execution time between okteto stable and this branch, but I cannot see any difference. Perhaps if we had hundreds of rules like this it would be beneficial to refactor. WHYT? 

Happy to give it a try together if you like

teresaromero (approved):
tested ✅ 

only to comment that also in stable, when running okteto build and the manifest has some kind of invalid key, the build is done and no error is returned.
perhaps to follow along we could open a task to fix this if we want to return an error while running okteto build

pkg/model/manifest_friendly_err.go
```diff
@@ -0,0 +1,165 @@
...
+		return errors.New(newErr)
+	}
+
+	return suggest.NewRule(condition, transformation)
+}
+
+// fieldsNotExistingRule replaces "not found" fields which are unknown to the Okteto manifest specification
+func fieldsNotExistingRule() *suggest.Rule {
+	pattern := `field (\w+) not found`
+	re := regexp.MustCompile(pattern)
```
AdrianPedriza, line 153: https://github.com/okteto/okteto/pull/4009#discussion_r1361118090 this was already addressed before I think

pkg/model/manifest_friendly_err.go
```diff
@@ -0,0 +1,165 @@
...
+		errorWithUrlToDocs := fmt.Sprintf("%s\n    Check out the okteto manifest docs at: %s", e.Error(), docsURL)
+		return errors.New(errorWithUrlToDocs)
+	}
+
+	return suggest.NewRule(isYamlErrorWithoutLinkToDocs, addUrl)
+}
+
+func indentNumLines() *suggest.Rule {
+	pattern := `(?:yaml: )?line (\d+):`
+	re := regexp.MustCompile(pattern)
```
AdrianPedriza, line 136: https://github.com/okteto/okteto/pull/4009#discussion_r1361118090 this will panic if expresion cannot be parsed

pkg/model/manifest_friendly_err.go
```diff
@@ -0,0 +1,165 @@
...
+	}
+
+	return suggest.NewRule(isYamlErrorWithoutLinkToDocs, addUrl)
+}
+
+// addUrlToManifestDocs appends to the error the URL to the Okteto manifest ref docs
+func addUrlToManifestDocs() *suggest.Rule {
+	addUrl := func(e error) error {
+		docsURL := "https://www.okteto.com/docs/reference/manifest"
+		errorWithUrlToDocs := fmt.Sprintf("%s\n    Check out the okteto manifest docs at: %s", e.Error(), docsURL)
```
AdrianPedriza, line 127: should we wrap all errors that dont need an string modification? That way we can check error type unwrapping error chain

pkg/model/manifest_friendly_err.go
```diff
@@ -0,0 +1,165 @@
...
+	}
+	// detect yaml errors by prefix
+	if strings.Contains(err.Error(), "yaml:") {
+		return true
+	}
+	// detect yaml errors by type
+	var typeError *yaml.TypeError
+	return errors.As(err, &typeError)
+}
+
```
AdrianPedriza, line 113: can you add a description about the rule added in this func?

pkg/model/manifest_friendly_err.go
```diff
@@ -0,0 +1,165 @@
...
+func addYamlParseErrorHeading() *suggest.Rule {
+	addUrl := func(e error) error {
+		errorWithUrlToDocs := fmt.Sprintf("Your okteto manifest is not valid, please check the following errors:\n%s", e.Error())
```
AdrianPedriza, line 116: should we wrap all manifest errors with `ErrInvalidManifest` ? If not, some checks for this error will never work. For example -> https://github.com/okteto/okteto/blob/master/cmd/build/command.go#L138

andreafalzetti: Good call @AdrianPedriza 👏 I've made the change here: https://github.com/okteto/okteto/pull/4009/commits/ffe965d4ae9d44b3bffa9161ee1ecf0ccb8f75a4

pkg/model/manifest_friendly_err.go
andreafalzetti: Due to new commit: https://github.com/okteto/okteto/pull/4009/commits/ffe965d4ae9d44b3bffa9161ee1ecf0ccb8f75a4 

This func is no longer necessary so I've removed it 🎉

pkg/model/manifest_friendly_err.go
andreafalzetti: Done in https://github.com/okteto/okteto/pull/4009/commits/ffe965d4ae9d44b3bffa9161ee1ecf0ccb8f75a4 👍

pkg/model/manifest_friendly_err.go
andreafalzetti: as discussed offline the pattern is fixed so it cannot panic, we use this approach in other parts where the pattern is hardcoded

pkg/model/manifest_friendly_err.go
andreafalzetti: as per the comment below about the fixed pattern

AdrianPedriza (approved):
code looks good. Tested e2e and works as expected. Excellent work! 👍

PR #4010, opened 2023-10-10
feat: add `--include-token` flag to `okteto ctx show` command

### LINKED ISSUE #3935: Remove `token` from `okteto context show` by default
When executing the `okteto context show` command the object returns the `token` value, which is not best practice for security.

We should remove the `token` value by default from this object.

We should also consider adding an option `--include-token` to the `okteto context show` command to optionally include that token in scenarios where it's useful. This way the disclosure of the `token` value can be better controlled/managed.

**Additional context**

nitishfy:
# Proposed changes

Fixes #3935

After changes:

image

### DIFF
--- a/cmd/context/show.go
+++ b/cmd/context/show.go
@@ -45,15 +45,8 @@ func Show() *cobra.Command {
 				return err
 			}
 
-			// store the token value to display to user if the flag is set
-			TokenValue := current.Token
-
-			// By default, set the token value to null since default behaviour doesn't require to show token value
-			current.Token = ""
-
-			// If the flag is set, display the token value
-			if includeToken {
-				current.Token = TokenValue
+			if !includeToken {
+				current.Token = ""
 			}
 
 			current.Certificate = ""
@@ -75,8 +68,7 @@ func Show() *cobra.Command {
 		},
 	}
 	cmd.Flags().StringVarP(&output, "output", "o", "json", "output format. One of: ['json', 'yaml']")
-	cmd.Flags().BoolVar(&includeToken, "include-token", false, "show the token value to the user")
-	return cmd
+    cmd.Flags().BoolVar(&includeToken, "include-token", false, "include the token in the output")	return cmd
 }
 
 func validateOutput(output string) error {
--- a/cmd/context/show.go
+++ b/cmd/context/show.go
@@ -28,6 +28,7 @@ import (
 // Show current context
 func Show() *cobra.Command {
 	var output string
+	var includeToken bool
 	cmd := &cobra.Command{
 		Use:   "show",
 		Args:  utils.NoArgsAccepted("https://okteto.com/docs/reference/cli/#show"),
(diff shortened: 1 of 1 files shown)


nitishfy:
There is no test file for `show.go`, Should we try adding test cases for it?  Maybe we can write tests about output validation.
cc: @jLopezbarb @andreafalzetti

cmd/context/show.go
```diff
@@ -43,6 +44,18 @@ func Show() *cobra.Command {
...
+			TokenValue := current.Token
+
+			// By default, set the token value to null since default behaviour doesn't require to show token value
+			current.Token = ""
+
+			// If the flag is set, display the token value
+			if includeToken {
+				current.Token = TokenValue
+			}
+
```
andreafalzetti, line 58: Why not simplify like this?

```suggestion
			if !includeToken {
				current.Token = ""
			}

```

cmd/context/show.go
```diff
@@ -62,7 +75,7 @@ func Show() *cobra.Command {
 		},
 	}
 	cmd.Flags().StringVarP(&output, "output", "o", "json", "output format. One of: ['json', 'yaml']")
-
+	cmd.Flags().BoolVar(&includeToken, "include-token", false, "show the token value to the user")
```
andreafalzetti, line 78:
```suggestion
	cmd.Flags().BoolVar(&includeToken, "include-token", false, "include the token in the output")
```

jLopezbarb:
> There is no test file for `show.go`, Should we try adding test cases for it? Maybe we can write tests about output validation. cc: @jLopezbarb @andreafalzetti

If you want to give it a try it would be awesome!! If you face any problem let us know and we will be willing to help you

cmd/context/show.go
AdrianPedriza: I would use this. Looks clearer

AdrianPedriza:
we need an info log about use `--include-token` when is not set. https://github.com/okteto/okteto/issues/3935#issuecomment-1742786094

andreafalzetti:
> we need an info log about use `--include-token` when is not set. [#3935 (comment)](https://github.com/okteto/okteto/issues/3935#issuecomment-1742786094)

What do you mean? In `okteto context show` the output is JSON or YAML so a log would break it

cmd/context/show.go
andreafalzetti: hey @nitishfy, can you apply this change? wdyt?

cmd/context/show.go
nitishfy: Resolved

cmd/context/show.go
nitishfy: Resolved

cmd/context/show.go
```diff
@@ -62,8 +68,7 @@ func Show() *cobra.Command {
 		},
 	}
 	cmd.Flags().StringVarP(&output, "output", "o", "json", "output format. One of: ['json', 'yaml']")
-
-	return cmd
+    cmd.Flags().BoolVar(&includeToken, "include-token", false, "include the token in the output")	return cmd
```
AdrianPedriza, line 71: a bad format here. does not compile
```suggestion
    cmd.Flags().BoolVar(&includeToken, "include-token", false, "include the token in the output")
    return cmd
```

nitishfy: Sorry! Resolved now

AdrianPedriza (approved):

PR #4019, opened 2023-10-16
Display a friendly error when the local proxy cannot start

teresaromero:
# Proposed changes

Fixes https://okteto.atlassian.net/browse/LAKE-16

When running `okteto deploy` and the local proxy was unable to get a free port or find localhost as host, the error thrown was not being descriptive of what could be happening. 

This PR detects the error and returns a UserError with a hint that can lead the user to troubleshoot what might be happening

## How to validate

- In order to force the error, we need to manually change the host provided [here](https://github.com/okteto/okteto/blob/379524e706b70d60bc8e711aa0963413ee5d6894/cmd/deploy/proxy.go#L78), so instead of `localhost` use a random host.
- Make build
- run `ok deploy`

the repro steps provided at the issue not always reproduce the error, but also can be used to validate:

> Steps to reproduce the behavior:
> 
> Modify your /etc/hosts file and comment any rule referring to localhost
> 
> Deploy a manifest that talks to Kubernetes, for example, in the deploy section add a command to list pods kubectl get pods
> 
> You will get a no such host error

Before:

```
❯ ok deploy
 i  Using - @ - as context
ERRO[0001] could not find a free port to start proxy server: lookup locaost: no such host
 x  Could not initialize local deploy command: lookup locaost: no such host
```

After:

```
❯ okteto deploy
 i  Using - @ - as context
 x  Could not initialize local deploy command: could not find available ports: lookup lalhost: no such host
    Review your configuration to make sure 'localhost' is resolved correctly
```

Expect an error with a hint troubleshooting

## CLI Quality Reminders 🔧

### DIFF
--- a/cmd/deploy/proxy.go
+++ b/cmd/deploy/proxy.go
@@ -43,6 +43,7 @@ import (
 	apiv1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/rest"
+	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
 )
 
 type proxyInterface interface {
@@ -72,14 +73,20 @@ type proxyHandler struct {
 	DivertDriver divert.Driver
 }
 
+type kubeconfigInterface interface {
+	GetCMDAPIConfig() (*clientcmdapi.Config, error)
+	Modify(port int, sessionToken string, destKubeconfigFile string) error
+	Read() (*rest.Config, error)
+}
+
 // NewProxy creates a new proxy
-func NewProxy(kubeconfig *KubeConfig) (*Proxy, error) {
+func NewProxy(kubeconfig kubeconfigInterface, portGetter func(iface string) (int, error)) (*Proxy, error) {
 	// Look for a free local port to start the proxy
-	port, err := model.GetAvailablePort("localhost")
+	port, err := portGetter("localhost")
 	if err != nil {
 		if dnsError, ok := err.(*net.DNSError); ok && dnsError.IsNotFound {
 			return nil, oktetoErrors.UserError{
-				E:    err,
+				E:    dnsError,
 				Hint: "Review your /etc/hosts configuration, make sure there is an entry for localhost",
 			}
 		}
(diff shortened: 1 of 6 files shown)


cmd/deploy/proxy.go
```diff
@@ -75,6 +77,12 @@ func NewProxy(kubeconfig *KubeConfig) (*Proxy, error) {
 	// Look for a free local port to start the proxy
 	port, err := model.GetAvailablePort("localhost")
 	if err != nil {
+		if dnsError, ok := err.(*net.DNSError); ok && dnsError.IsNotFound {
+			return nil, oktetoErrors.UserError{
+				E:    err,
+				Hint: "Review your /etc/hosts configuration, make sure there is an entry for localhost",
```
ifbyol, line 83: This is just an example and, as we saw yesterday, this might not even generate the error, so not sure it provides values. Maybe we just want to wrap here the original error indicating that we couldn't find an available port to initialize the deploy command

cmd/deploy/proxy.go
```diff
@@ -71,10 +73,16 @@ type proxyHandler struct {
...
 // NewProxy creates a new proxy
-func NewProxy(kubeconfig *KubeConfig) (*Proxy, error) {
+func NewProxy(kubeconfig kubeConfigHandler, portGetter func(iface string) (int, error)) (*Proxy, error) {
 	// Look for a free local port to start the proxy
-	port, err := model.GetAvailablePort("localhost")
+	port, err := portGetter("localhost")
 	if err != nil {
+		if dnsError, ok := err.(*net.DNSError); ok && dnsError.IsNotFound {
+			return nil, oktetoErrors.UserError{
+				E:    fmt.Errorf("could not find available ports: %w", dnsError),
```
AdrianPedriza, line 82: I think adding this context to the error might be confusing. If I face the error 
Captura de Pantalla 2023-10-18 a las 10 47 21
I could think that there is some problem with some ports but it is not the case. Adding this context to the original error seems weird for me in every case. I would return just the error that comes from `portGetter` WDYT?

teresaromero: i think that the function throwing the error is to get the available ports on a host, so it kind on makes sense to me. i initially just returned the error but this was modified after the revision of nacho.

cmd/deploy/proxy.go
AdrianPedriza: following https://github.com/okteto/okteto/pull/4019#discussion_r1363500677, is it worth adding information in the error about the ports? i think it could be confusing

ifbyol: @omnomagonz @rlamana Could you help with this, please? What do you think is the best messaging we can display here?

omnomagonz: I think @ifbyol's suggestion works. I think between the error and the hint we're covering quite a bit that's more helpful compared to the current error/experience.

AdrianPedriza (approved):

andreafalzetti (approved):

PR #4020, opened 2023-10-18
[CLI Quality][Lint] add tparallel lint and fix errors

teresaromero:
# Proposed changes

Adding `tparallel` lintern to golangci-lint config in order to track the use of t.Parallel() correctly.
Bad use of this feature for tests can lead to flaky test.

EDIT:
offline conversation @andreafalzetti detected the option `new-from-rev` from the golangci-lint config file was not working properly.
After investigation, the only way it detects errors only on the new code added is referencing by branch name, eg: `master`.
Proposal is to remove this option from the config as the aproach we are using to add lints is to fix the issues when implementing a new one, so all the code should be clean when new code changes come up. From the docs:

https://golangci-lint.run/usage/faq/#how-to-integrate-golangci-lint-into-large-project-with-thousands-of-issues

> We are sure that every project can easily integrate golangci-lint, even the large one. The idea is to not fix all existing issues. Fix only newly added issue: issues in new code. To do this setup CI to run golangci-lint with option --new-from-rev=HEAD ~ 1. Also, take a look at option --new, but consider that CI scripts that generate unstaged files will make --new only point out issues in those files and not in the last commit. In that regard --new-from-rev=HEAD ~ 1 is safer. By doing this you won't create new issues in your code and can choose fix existing issues (or not).

## CLI Quality Reminders 🔧

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -41,6 +41,7 @@ linters:
     - unused
     - unparam
     - tenv
+    - tparallel
 
 issues:
   # Show only new issues created after git revision REV. Run only over the changed files
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -44,8 +44,6 @@ linters:
     - tparallel
 
 issues:
-  # Show only new issues created after git revision REV. Run only over the changed files
-  new-from-rev: HEAD~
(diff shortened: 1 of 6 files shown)


andreafalzetti (approved):
# 🎖️

PR #4023, opened 2023-10-23
add build service context generation to image build event

AdrianPedriza:
## Context
This PR add for every service build new information into the `imageBuild` event in order to collect metrics related to smart builds.

## Changes
- add two properties for `imageBuild` event: `BuildContextHash` and `BuildContextHashDuration`
- create interface `configServiceRepositoryInterface` with the behavior `GetTreeHash`
- inject repo dependency into `OktetoBuilder` which satisfies `configServiceRepositoryInterface` interface
- create `calculateBuildContextHash` func which includes to the tree hash the buildInfo parameters
- add unit/integration tests 

## How to test it
- check new props metrics are sent for every service build


cmd/build/v2/build.go
```diff
@@ -253,6 +260,39 @@ func (bc *OktetoBuilder) Build(ctx context.Context, options *types.BuildOptions)
 	return options.Manifest.ExpandEnvVars()
 }
 
+func (ob *OktetoBuilder) calculateBuildContextHash(buildInfo *model.BuildInfo) string {
```
teresaromero, line 263: can't we re-use the `getBuildHashFromCommit` but with `treeHash` instead of `repoCommit` ?

teresaromero: as line 202

cmd/build/v2/build.go
```diff
@@ -111,6 +113,7 @@ func NewBuilderFromScratch(analyticsTracker analyticsTrackerInterface) *OktetoBu
 		buildEnvironments: map[string]string{},
 		Config:            getConfig(registry, gitRepo),
 		analyticsTracker:  analyticsTracker,
+		repo:              gitRepo,
```
teresaromero, line 116: can we include inside `Config` this new prop instead of injecting the repo?

cmd/build/v2/build.go
```diff
@@ -253,6 +260,39 @@ func (bc *OktetoBuilder) Build(ctx context.Context, options *types.BuildOptions)
 	return options.Manifest.ExpandEnvVars()
 }
 
+func (ob *OktetoBuilder) calculateBuildContextHash(buildInfo *model.BuildInfo) string {
+	buildContext := buildInfo.Context
+	treeHash, err := ob.repo.GetTreeHash(buildContext)
```
teresaromero, line 265:
```suggestion
	treeHash, err := ob.Config.GetTreeHash(buildContext)
```
and calculate it after the repoCommit ?

cmd/build/v2/build.go
AdrianPedriza: I thought about it but some fields in the generated hash has values related to the hash like `commit` vs `tree_hash` that's why I split it. I could inject this value to use only one func.

AdrianPedriza: cebae20a007361f6ea2f43a51e8d273c7b0e78eb to reuse same func (sorry for not isolate the change)

cmd/build/v2/build.go
AdrianPedriza: cebae20a007361f6ea2f43a51e8d273c7b0e78eb

cmd/build/v2/build.go
AdrianPedriza: build context hash cannot be calculated when repo commit is calculated because we need to know the service to generate this data, that is not the case to get git repo commit hash

cmd/build/v2/build.go
teresaromero: but... we need to use a commit, instead of the one on the root project, the one of the build context

AdrianPedriza: we are using it. https://github.com/okteto/okteto/pull/4023/files#diff-291c33d5347afee4c757a4c6efd49c310488a2fc7855fc15187108eda8202293R90-R97 . Is that what you mean?

AdrianPedriza: build hash key is "commit" to build the hash using the repo commit [(like the current one)](https://github.com/okteto/okteto/pull/4023/files#diff-b219a27123f02bc2d87bec867d81a32e91c18ce5edd0dff86aabec32c7647241R461-R463) and "tree_hash" for build hash made by the tree_hash (build context)

AdrianPedriza: input is only `tree_hash` for build context hashes

teresaromero (approved):
✅ tested

cmd/build/v2/config.go
```diff
@@ -84,3 +86,13 @@ func (oc oktetoBuilderConfig) GetGitCommit() string {
 func (oc oktetoBuilderConfig) GetAnonymizedRepo() string {
 	return oc.repository.GetAnonymizedRepo()
 }
+
+func (oc oktetoBuilderConfig) GetBuildContextHash(buildInfo *model.BuildInfo) string {
+	buildContext := buildInfo.Context
+	treeHash, err := oc.repository.GetTreeHash(buildInfo.Context)
+	if err != nil {
+		oktetoLog.Info("error trying to get tree hash for build context '%s': %w", buildContext, err)
```
andreafalzetti, line 94: What do you think if we return `err` and log outside the function? I am concerned it will be harder to test and considering the effort in refactoring the logger to be injected instead of being global, it could be one less instance in the code to change in the future.

cmd/build/v2/config.go
```diff
@@ -84,3 +86,13 @@ func (oc oktetoBuilderConfig) GetGitCommit() string {
 func (oc oktetoBuilderConfig) GetAnonymizedRepo() string {
 	return oc.repository.GetAnonymizedRepo()
 }
+
+func (oc oktetoBuilderConfig) GetBuildContextHash(buildInfo *model.BuildInfo) string {
```
andreafalzetti, line 90: Could we add a test for this fn?

pkg/repository/git.go
```diff
@@ -120,6 +123,51 @@ func (r gitRepoController) getSHA() (string, error) {
...
+	}
+
+	return svcEntry.Hash.String(), nil
+}
+
+func getTreeDirFromRelPath(relPath string) string {
+
+	wdCtrl := filesystem.NewOsWorkingDirectoryCtrl()
+	wd, err := wdCtrl.Get()
+	if err != nil {
```
andreafalzetti, line 164: Same as https://github.com/okteto/okteto/pull/4023#discussion_r1371621771

It feels odd to have logs inside the function, it could up to the calling function to decide wheather to ignore the error and log. WDYT?

cmd/build/v2/config.go
AdrianPedriza: 2bd136b21e02efb9eaf6fdf36e4ef42173197240

cmd/build/v2/config.go
AdrianPedriza: tbh Im not sure where to move it. I've followed the same approach we have now for `configRepositoryInterface` interface like `GetGitCommit` where we just return a string with the value, I think we should try to be as consistent as possible

pkg/repository/git.go
AdrianPedriza: acd7ef13f7f00dedac81e167c82100f3e202eb55

andreafalzetti (approved):

PR #4024, opened 2023-10-23
Update analytics data

teresaromero:
# Proposed changes

Resolves https://okteto.atlassian.net/browse/LAKE-5

This PR updates the common data sent on every event at tracking

- refactor of tracking functions to remove deprecated props from events
- request cluster metadata when initialization of the context
- remove deprecated props from the common fields at track function
- add new props at track funcion

## How to validate

With a cluster which has enabled analytics, run the binary and look for the new properties under the binary version.
The removed ones should not appear.

## CLI Quality Reminders 🔧

### DIFF
--- a/cmd/context/create.go
+++ b/cmd/context/create.go
@@ -247,7 +247,7 @@ func (c *ContextCommand) UseContext(ctx context.Context, ctxOptions *ContextOpti
 func getClusterMetadata(ctx context.Context, namespace string, okClientProvider oktetoClientProvider) (types.ClusterMetadata, error) {
 	okClient, err := okClientProvider.Provide()
 	if err != nil {
-		oktetoLog.Infof("error providing okteto client: %v", err)
+		return types.ClusterMetadata{}, err
 	}
 	return okClient.User().GetClusterMetadata(ctx, namespace)
 }
@@ -302,6 +302,7 @@ func (c *ContextCommand) initOktetoContext(ctx context.Context, ctxOptions *Cont
 
 	clusterMetadata, err := getClusterMetadata(ctx, ctxOptions.Namespace, c.OktetoClientProvider)
 	if err != nil {
+		oktetoLog.Infof("error getting cluster metadata: %v", err)
 		return err
 	}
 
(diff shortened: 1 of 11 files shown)


teresaromero:
Regarding code coverage, the analytics pkg needs more coverage though a refactor that should need its own epic/issue to plan the changes of this code

cmd/context/create.go
```diff
@@ -185,7 +185,11 @@ func (c *ContextCommand) UseContext(ctx context.Context, ctxOptions *ContextOpti
 	c.initEnvVars()
 
 	if ctxOptions.IsOkteto {
-		if err := c.initOktetoContext(ctx, ctxOptions); err != nil {
+		clusterMetadata, err := getClusterMetadata(ctx, ctxOptions.Namespace, c.OktetoClientProvider)
+		if err != nil {
+			return err
+		}
+		if err := c.initOktetoContext(ctx, ctxOptions, clusterMetadata); err != nil {
```
AdrianPedriza, line 192: could we init the context before access Okteto API in `getClusterMetadata`? Using that way if we have an invalid token we renew it and we can access metadata. I wonder if we want that or prefer to fail. You can check by removing the token from the context and running your code

cmd/context/create.go
```diff
@@ -243,6 +243,15 @@ func (c *ContextCommand) UseContext(ctx context.Context, ctxOptions *ContextOpti
 	return nil
 }
 
+// getClusterMetadata runs the user query GetClusterMetadata and returns the response
+func getClusterMetadata(ctx context.Context, namespace string, okClientProvider oktetoClientProvider) (types.ClusterMetadata, error) {
+	okClient, err := okClientProvider.Provide()
+	if err != nil {
+		oktetoLog.Infof("error providing okteto client: %v", err)
```
andreafalzetti, line 250: I think it's better to move logging responsability up, in the calling fn, so that this function is not interacting with logs logic

andreafalzetti (approved):

AdrianPedriza (approved):

cmd/context/create.go
teresaromero: i've reviewed and changed this

PR #4028, opened 2023-10-26
Adds add/delete to docker credentials helper

maroshii:
# Proposed changes

In Okteto 1.14, we add a docker config file with a single `credStore` responsible for negotiating credentials with the Okteto backend. The helper doesn't implement `add` nor `delete` but catches all registries so doing something like `ko login` doesn't work.

Example repo: https://github.com/okteto/go-with-ko/blob/main/okteto.yaml. The example works because `DOCKER_CONFIG` is set to a different file.

This fix uses a separate docker config file to resolve credentials. Note that this usually run in a temporary environment (buildkit or a kuberentes job)  so managing the lifecycle of this file is not strictly necessary.

## How to validate

To reproduce:

1. Clone: https://github.com/okteto/go-with-ko
2. Check out this commit `de1f3c60037a7fcdaa1c6b3f88382f332c5ba60b`
1. Remove this line: https://github.com/okteto/go-with-ko/blob/de1f3c60037a7fcdaa1c6b3f88382f332c5ba60b/okteto.yaml#L7
1. add `set -e` at the top of the script so execution does not continue
1. Run a `okteto deploy --remote`

A `not implemented` error should be returned. This error is returned by the credentials helper

**To test the fix**: Doing the exact same as above with this patch should work. This requires making sure that `okteto deploy --remote` executes this version of the cli which is not explained in this PR for brevity.

## CLI Quality Reminders 🔧

### DIFF
--- a/pkg/auth/dockercredentials/helper.go
+++ b/pkg/auth/dockercredentials/helper.go
@@ -14,8 +14,12 @@ package dockercredentials
 
 import (
 	"errors"
+	"path"
 
+	"github.com/docker/cli/cli/config"
+	dockertypes "github.com/docker/cli/cli/config/types"
 	"github.com/docker/docker-credential-helpers/credentials"
+	okconfig "github.com/okteto/okteto/pkg/config"
 )
 
 var ErrNotImplemented = errors.New("not implemented")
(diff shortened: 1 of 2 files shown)


ifbyol (commented):
At some point, we might want to include the docker config for local executions of `okteto deploy`, If that happens, we will have to be aware of this to not leave behind this files. For now, might be ok.

I'm not sure how we can remember it if that becomes a reality one day

AdrianPedriza:
code looks good, but I want to reproduce before/after behavior following description steps but I'm obtaining same result in both cases despite what result should be following the description:
Captura de Pantalla 2023-10-31 a las 12 39 38

(I've build the cli and set the for remote instance with the changes)

maroshii:
@AdrianPedriza there was a commit to that repo so now line 7 maps refers to a new line. The command that fails is `ko login` which is the one that adds the credentials. Update repro steps to check out the previous commit.

AdrianPedriza (approved):

maroshii:
Closing for now. This works well in remote (buildkit) deploys and instller jobs but running `okteto deploy` from local would leave behind registry credentials in the user host under the `.okteto` folder which is pretty bad.

maroshii:
Discussed with @andreafalzetti @ifbyol and @rlamana. We'll go with this approach for now to unblock this use case until we decide how to better integration local deploy with private registry creds.

andreafalzetti (approved):

ifbyol (approved):

PR #4036, opened 2023-11-06
add feature flag to enable disable pull first approach

AdrianPedriza:
# Proposed changes
- create feature flag to enable smart build usage
- do not use global registry if feature flag is disabled

## How to validate

if feature flag enabled:
- `okteto build` should skip build image if it is already built for that commit

if feature flag is disabled:
- every `okteto build` command should build the image even if the version has already been built

## CLI Quality Reminders 🔧

### DIFF
--- a/cmd/build/v2/build.go
+++ b/cmd/build/v2/build.go
@@ -110,7 +110,7 @@ func NewBuilderFromScratch(analyticsTracker analyticsTrackerInterface) *OktetoBu
 	config := getConfig(registry, gitRepo)
 
 	buildEnvs := map[string]string{}
-	buildEnvs[OktetoEnableSmartBuilds] = strconv.FormatBool(config.isSmartBuildsEnable)
+	buildEnvs[OktetoEnableSmartBuildEnvVar] = strconv.FormatBool(config.isSmartBuildsEnable)
 
 	return &OktetoBuilder{
 		Builder:           builder,
--- a/cmd/build/v2/build.go
+++ b/cmd/build/v2/build.go
@@ -20,6 +20,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
@@ -106,11 +107,16 @@ func NewBuilderFromScratch(analyticsTracker analyticsTrackerInterface) *OktetoBu
 		oktetoLog.Infof("could not get working dir: %w", err)
 	}
 	gitRepo := repository.NewRepository(wd)
+	config := getConfig(registry, gitRepo)
+
+	buildEnvs := map[string]string{}
+	buildEnvs[OktetoEnableSmartBuilds] = strconv.FormatBool(config.isSmartBuildsEnable)
+
 	return &OktetoBuilder{
 		Builder:           builder,
 		Registry:          registry,
 		V1Builder:         buildv1.NewBuilder(builder, registry),
-		buildEnvironments: map[string]string{},
+		buildEnvironments: buildEnvs,
 		Config:            getConfig(registry, gitRepo),
 		analyticsTracker:  analyticsTracker,
 	}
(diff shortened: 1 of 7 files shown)


cmd/build/v2/build.go
```diff
@@ -65,6 +65,7 @@ type oktetoBuilderConfigInterface interface {
 	IsOkteto() bool
 	GetAnonymizedRepo() string
 	GetBuildContextHash(*model.BuildInfo) string
+	IsSmartBuildsEnable() bool
 }
```
andreafalzetti, line 69: Should we call it `IsSmartBuildsEnabled`?

cmd/build/v2/build.go
```diff
@@ -65,6 +65,7 @@ type oktetoBuilderConfigInterface interface {
 	IsOkteto() bool
 	GetAnonymizedRepo() string
 	GetBuildContextHash(*model.BuildInfo) string
+	IsSmartBuildsEnable() bool
```
jLopezbarb, line 68: I don't like adding this in directly on the main struct. Following best practices for interfaces we should be creating another interface for checking the `IsSmartBuildsEnable` and adding it as a composition for the main struct. This way we can create a simple struct for the tests that checks only `IsSmartBuildsEnable` without the need of implementing the rest of the functions

cmd/build/v2/config.go
```diff
@@ -14,12 +14,20 @@
...
+
 	oktetoLog "github.com/okteto/okteto/pkg/log"
 	"github.com/okteto/okteto/pkg/model"
 	"github.com/okteto/okteto/pkg/okteto"
 	"github.com/spf13/afero"
 )
 
+const (
+	// OktetoEnableSmartBuilds key representing whether the feature flag to enable smart builds is enabled or not
+	OktetoEnableSmartBuilds = "OKTETO_SMART_BUILDS_ENABLED"
```
jLopezbarb, line 28: Could we add the EnvVar suffix to make it clear the intention of the constant?
```suggestion
	OktetoEnableSmartBuildEnvVar = "OKTETO_SMART_BUILDS_ENABLED"
```

cmd/build/v2/config.go
```diff
@@ -49,12 +58,26 @@ func getConfig(registry configRegistryInterface, gitRepo configRepositoryInterfa
 	if err != nil {
 		oktetoLog.Infof("error trying to get directory: %w", err)
 	}
+
+	enableSmartBuilds := true
+	enableSmartBuildsStr := os.Getenv(OktetoEnableSmartBuilds)
+	if enableSmartBuildsStr != "" {
+		smartBuildEnabledBool, err := strconv.ParseBool(enableSmartBuildsStr)
+		if err != nil {
+			oktetoLog.Infof("error trying to get feature flag %s: %w", OktetoEnableSmartBuilds, err)
```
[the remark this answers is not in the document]
jLopezbarb: Agree with changing it into warning. Also, Should we encapsulate this code into a function and make the test function to focus on just that? I think the test is getting bigger and bigger and it can be hard to read if we continue adding use cases without having an specific tests

cmd/build/v2/build.go
AdrianPedriza: 36042266fd89ed3e43ee7d21db2803a63c36a43e

cmd/build/v2/config.go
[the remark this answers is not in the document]
AdrianPedriza: 27fbc6b3fe2a8d8a1d6208e1274c004e4bf1af35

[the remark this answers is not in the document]
AdrianPedriza: @jLopezbarb 391fa8f2661001c76650f84bfaa8634fa84497f1 to move feature flag retrieval

cmd/build/v2/build.go
AdrianPedriza: I would keep it that way, I see no reason why this should be interface composition and not the other methods. Maybe there is something that escapes me but I would try to keep consistency.

andreafalzetti (approved):
Code LGTM, tested in different scenario and it works as expected!

Great work!

PR #4037, opened 2023-11-06
RAIN-51 and RAIN-52: Set right certificate in kubernetes client and kubeconfig generated when working with `--insecure-skip-tls-verify`

ifbyol:
# Proposed changes

Fixes RAIN-51 and RAIN-52

This PR fixes 2 different bugs related with the same. The certificate to be used in kubernetes client when the Okteto context is being configured with `--insecure-skip-tls-verify` flag.

Issue RAIN-51: When the context was using `--insecure-skip-tls-verify` and the Okteto instance has defined `cluster.endpoint` property (kubernetes is **not** exposed behind our ingress controller), we were using the wrong certificate in the generated Kubeconfig. We were always using the one stored in the context, but we should use it the context is insecure AND the kubernetes API is exposed behind Okteto's ingress controller

Issue RAIN-52: When the context was using `--insecure-skip-tls-verify` AND the Kubernetes API is exposed behind our ingress controller (helm chart property `cluster.endpoint` is not defined) we were not using the right certificate in the internal Kubernetes client. In this case, we should be using the one in the Okteto context instead of using the one retrieved from the server.

## How to validate

Issue RAIN-51. Steps to reproduce the error:

1. Have an Okteto instance using self-signed certificate and defining `cluster.endpoint` pointing to the Kubernetes API
2. Execute `okteto ctx use <instance-url> --insecure-skip-tls-verify`
3. Execute `okteto kubeconfig`
4. Try to execute `kubectl get pods`. You should get a certificate error

With my current changes, you won't get any error and you should be able to execute any kubernetes command

Issue RAIN-52. Steps to reproduce the error:

1. Have an Okteto installation using self-signed certificates and leave empty the helm setting `cluster.endpoint`
2. Execute `okteto ctx use <instance-url> --insecure-skip-tls-verify`
3. Execute some command like `okteto up` or `okteto deploy`. You should get an error

With my current changes, you won't get any error and the commands should be executed successfully

Apart from that, we should verify the rest of cases keep working fine

## CLI Quality Reminders 🔧

### DIFF
--- a/pkg/okteto/context.go
+++ b/pkg/okteto/context.go
@@ -350,21 +350,35 @@ func (*ContextConfigWriter) Write() error {
 }
 
 // AddOktetoCredentialsToCfg populates the provided kubernetes config using the provided credentials obtained from the Okteto API
-func AddOktetoCredentialsToCfg(cfg *clientcmdapi.Config, cred *types.Credential, namespace, userName, oktetoURL string) {
+func AddOktetoCredentialsToCfg(cfg *clientcmdapi.Config, cred *types.Credential, namespace, userName string, oktetoContext OktetoContext) error {
 	// If the context is being initialized within the execution of `okteto deploy` deploy command it should not
 	// write the Okteto credentials into the kubeconfig. It would overwrite the proxy settings
 	if os.Getenv(constants.OktetoSkipConfigCredentialsUpdate) == "true" {
-		return
+		return nil
 	}
 
-	clusterName := UrlToKubernetesContext(oktetoURL)
+	clusterName := UrlToKubernetesContext(oktetoContext.Name)
 	// create cluster
 	cluster, ok := cfg.Clusters[clusterName]
 	if !ok {
 		cluster = clientcmdapi.NewCluster()
 	}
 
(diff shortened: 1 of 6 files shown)


pkg/okteto/context.go
```diff
@@ -350,21 +350,35 @@ func (*ContextConfigWriter) Write() error {
...
-	clusterName := UrlToKubernetesContext(oktetoURL)
+	clusterName := UrlToKubernetesContext(oktetoContext.Name)
 	// create cluster
 	cluster, ok := cfg.Clusters[clusterName]
 	if !ok {
 		cluster = clientcmdapi.NewCluster()
 	}
 
-	cluster.CertificateAuthorityData = []byte(cred.Certificate)
+	// If the certificat included in the credentials is empty, it means that the server is kubernetes.subdomain.
```
jLopezbarb, line 367:
```suggestion
	// If the certificate included in the credentials is empty, it means that the server is kubernetes.subdomain.
```

jLopezbarb (approved):
Tested and LGTM

PR #4043, opened 2023-11-13
Extract dev `secrets` and `envFiles`  validation from serializer

andreafalzetti:
# Proposed changes

Fixes LAKE-90

Loom demo 👉  https://www.loom.com/share/a15471e125d9411da9ef95c35e3bea27

This change is aiming to improve the developer experience to users using commands such as `okteto deploy --remote` that do not require the `dev` section of the manifest, but because such section has either a non existing `secret` or `envFile`, the `deploy` command fails.

Disclaimer: This is intended to be a low hanging fruit improvement, the most impactful change here would be do not de-serialize the `dev` section of the manifest, which will follow-up in future PRs.

## How to validate

⚠️  Because this change touches on important commands such as `okteto up`, `okteto deploy`, etc. Please do a careful review with several manifests, both well configured and malformed, to make sure we do not have any regression.

To replicate the current undesired behavioir, follow these steps:

1. Using [okteto/movies](https://github.com/okteto/movies)
2. Using latest stable `okteto` release
3. Deploy the project to an okteto cluster
4. Now, edit `okteto.yml` and in the `dev.api` add:
```yaml
    secrets:
      - $HOME/.token:/root/.token:400
```
5. Now run `okteto deploy --remote`
6. Notice how the CLI fails to deploy because `.token` is not found

Another test is:

1. Edit `okteto.yml` and add in `dev.api`
```
    envFiles:
      - .fake
```
2. Now run `okteto deploy --remote`
3. Notice how the CLI fails to deploy because `.fake` is not found

Now, to validate that this change is resolving the issue, repeat the steps above with the cli built from this branch, build a new image of the CLI and replace it in `remote.go`. In both cases, `okteto deploy --remote` should work successfully.

## CLI Quality Reminders 🔧

### DIFF
--- a/pkg/model/dev.go
+++ b/pkg/model/dev.go
@@ -75,10 +75,10 @@ type Dev struct {
 	Namespace            string                `json:"namespace,omitempty" yaml:"namespace,omitempty"`
 	Container            string                `json:"container,omitempty" yaml:"container,omitempty"`
 	ServiceAccount       string                `json:"serviceAccount,omitempty" yaml:"serviceAccount,omitempty"`
-	ParentSyncFolder     string
-	Interface            string           `json:"interface,omitempty" yaml:"interface,omitempty"`
-	Mode                 string           `json:"mode,omitempty" yaml:"mode,omitempty"`
-	ImagePullPolicy      apiv1.PullPolicy `json:"imagePullPolicy,omitempty" yaml:"imagePullPolicy,omitempty"`
+	ParentSyncFolder     string                `json:"-" yaml:"-"`
+	Interface            string                `json:"interface,omitempty" yaml:"interface,omitempty"`
+	Mode                 string                `json:"mode,omitempty" yaml:"mode,omitempty"`
+	ImagePullPolicy      apiv1.PullPolicy      `json:"imagePullPolicy,omitempty" yaml:"imagePullPolicy,omitempty"`
 
 	Tolerations     []apiv1.Toleration `json:"tolerations,omitempty" yaml:"tolerations,omitempty"`
 	Command         Command            `json:"command,omitempty" yaml:"command,omitempty"`
--- a/pkg/model/dev.go
+++ b/pkg/model/dev.go
@@ -52,32 +52,32 @@ var (
 
 // Dev represents a development container
 type Dev struct {
-	Name                 string             `json:"name,omitempty" yaml:"name,omitempty"`
-	Username             string             `json:"-" yaml:"-"`
-	RegistryURL          string             `json:"-" yaml:"-"`
-	Selector             Selector           `json:"selector,omitempty" yaml:"selector,omitempty"`
-	Annotations          Annotations        `json:"annotations,omitempty" yaml:"annotations,omitempty"`
-	Tolerations          []apiv1.Toleration `json:"tolerations,omitempty" yaml:"tolerations,omitempty"`
-	Context              string             `json:"context,omitempty" yaml:"context,omitempty"`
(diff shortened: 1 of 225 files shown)


pkg/model/serializer.go
```diff
@@ -457,17 +457,15 @@ func (s *Secret) UnmarshalYAML(unmarshal func(interface{}) error) error {
 		}
 	}
 
+	// if the secret is not formatted correctly we return an empty secret (i.e. empty LocalPath and RemotePath)
+	// and we rely on the secret validation to return the appropriate error to the user
 	if len(parts) < 2 || len(parts) > 3 {
```
AdrianPedriza, line 462:
```suggestion
	if len(parts) != 2  {
```

pkg/model/serializer_test.go
```diff
@@ -424,26 +421,26 @@ func TestSecretMarshalling(t *testing.T) {
...
 		{
 			name:          "wrong-remote",
 			data:          fmt.Sprintf("%s:remote", file.Name()),
-			expected:      nil,
-			expectedError: true,
+			expected:      &Secret{LocalPath: file.Name(), RemotePath: "remote", Mode: 420},
+			expectedError: false,
 		},
 		{
 			name:          "wrong-mode",
```
AdrianPedriza, line 446: should we fail here and not it validation for a wrong mode?

andreafalzetti: because validation is done outside, a wrong mode will cause the local and remote port to be `nil` only, so it does not fail. We fail when we call `secret.Validate()`

pkg/model/serializer.go
andreafalzetti: if we change this, we have unreachable conditions below, where it checks for `== 3`

AdrianPedriza:
if I run `okteto deploy --remote` in `okteto/movies` I get:
Captura de Pantalla 2023-11-15 a las 11 39 37
seems it is validating dev section for remote deployments?

AdrianPedriza:
> if I run `okteto deploy --remote` in `okteto/movies` I get: Captura de Pantalla 2023-11-15 a las 11 39 37 seems it is validating dev section for remote deployments?

working as expected using the proposed version in remote. LGTM!

AdrianPedriza (approved):

pkg/model/dev.go
```diff
@@ -809,6 +810,21 @@ func (dev *Dev) Validate() error {
 	return nil
 }
 
+// Prepare calls other methods required to have the dev ready to use
+func (dev *Dev) Prepare(manifestPath string) error {
```
jLopezbarb, line 814: Given that this is just expanding relative paths. Should we name it `PreparePaths`?

pkg/model/manifest.go
```diff
@@ -721,6 +711,33 @@ func (b *ManifestBuild) validate() error {
...
+	}
+
+	if !strings.HasPrefix(s.RemotePath, "/") {
+		return fmt.Errorf("secret remote path '%s' must be an absolute path", s.RemotePath)
+	}
+
+	return nil
+}
+
+func checkFileAndNotDirectory(path string, fs afero.Fs) error {
```
jLopezbarb, line 730: I think this code is repeated in `CheckIfRegularFile`. Should we create a new package that takes care of the filesystem? I think that could reduce the `cmd/utils`

jLopezbarb: I've seen that this was moved from the serializer. Up to you if we change it now or when we get to that part

andreafalzetti: I will open a follow up PR for this, because there are a few changes that I want to make there.

jLopezbarb (approved):

PR #4048, opened 2023-11-16
show hint to deploy using remote feature

AdrianPedriza:
# Proposed changes
- add a hint to suggest using the remote feature for deploys only when the user exec `okteto deploy` without deploying using remote feature 

Captura de Pantalla 2023-11-16 a las 15 35 52

## CLI Quality Reminders 🔧

### DIFF
--- a/cmd/deploy/deploy.go
+++ b/cmd/deploy/deploy.go
@@ -541,7 +541,7 @@ func shouldRunInRemote(opts *Options) bool {
 		return true
 	}
 
-	oktetoLog.Information("Have you tried the `--remote` option? You can run the deploy commands in your Okteto cluster, check our docs for more.")
+	oktetoLog.Information("Have you tried the `--remote` option? You can run the deploy commands in your Okteto cluster. More information available here: https://www.okteto.com/docs/reference/manifest/#deploy-remotely")
 	return false
 
 }
--- a/cmd/deploy/deploy.go
+++ b/cmd/deploy/deploy.go
@@ -541,7 +541,7 @@ func shouldRunInRemote(opts *Options) bool {
 		return true
 	}
 
-	oktetoLog.Information("Have you tried the `--remote` option? You can run the deploy commands in your Okteto cluster. More information available here: https://www.okteto.com/docs/reference/manifest/#deploy-remotely")
+	oktetoLog.Information("Use `--remote` to run the deploy commands in your Okteto cluster.\n    More information available here: https://www.okteto.com/docs/reference/manifest/#deploy-remotely")
 	return false
 
 }
--- a/cmd/deploy/deploy.go
+++ b/cmd/deploy/deploy.go
@@ -541,6 +541,7 @@ func shouldRunInRemote(opts *Options) bool {
 		return true
 	}
 
+	oktetoLog.Information("Have you tried the `--remote` option? You can run the deploy commands in your Okteto cluster, check our docs for more.")
 	return false
 
 }


cmd/deploy/deploy.go
```diff
@@ -541,6 +541,7 @@ func shouldRunInRemote(opts *Options) bool {
 		return true
 	}
 
+	oktetoLog.Information("Have you tried the `--remote` option? You can run the deploy commands in your Okteto cluster, check our docs for more.")
```
pchico83, line 544: Could we add a link to the `remote` docs?

AdrianPedriza: 👍 I'll do! First of all, I would like to clarify if it is necessary to add more documentation about the feature to know which link is the right one to use in this suggestion. I think for now we have what we need but I would like to know your opinion. For now we can link or use from the [cli](https://www.okteto.com/docs/reference/cli/#deploy:~:text=bash%20(defaults%20to%20false)-,%2D%2Dremote,-(bool)) or in the [manifest](https://www.okteto.com/docs/reference/manifest/#deploy-remotely)

AdrianPedriza: I pointed to the docuemntation of the manifest which shows the detail about the feature. 91423593ff7f8379dc7dcd9702b30f5d5619b51f

cmd/deploy/deploy.go
```diff
@@ -541,6 +541,7 @@ func shouldRunInRemote(opts *Options) bool {
 		return true
 	}
 
+	oktetoLog.Information("Have you tried the `--remote` option? You can run the deploy commands in your Okteto cluster. More information available here: https://www.okteto.com/docs/reference/manifest/#deploy-remotely")
```
pchico83, line 544: @AdrianPedriza could you attach a screenshot? I think the text is too long for a single line, should we break it?

AdrianPedriza: this is the current hint (option 1):
Captura de Pantalla 2023-11-17 a las 9 46 09

if we break it the output would be (option 2):
Captura de Pantalla 2023-11-17 a las 9 44 39

Another Option 3 would be to rephrase the sentence so that it falls within one line.

What option do you think it is more appropriate?

cmd/deploy/endpoints.go
```diff
@@ -189,14 +189,16 @@ func (eg *EndpointGetter) getEndpoints(ctx context.Context, opts *EndpointsOptio
 		return nil, err
 	}
 
-	externalEps, err := eg.endpointControl.List(ctx, opts.Namespace, labelSelector)
-	if err != nil {
-		return nil, err
-	}
+	if okteto.Context().IsOkteto {
```
teresaromero, line 192: is this part required for the change? if not can we move it to additional PR, the diff coverage is being affected by this change, which is not tested, and reading the issue i think it might be an additional change not required for the messagE?

AdrianPedriza: oh! I think I've pushed that commit by wrong in this branch. :hushed: Good catch Tere! I'll revert that.

teresaromero (approved):
Approved conditioned to resolution of the copy. 👍🏻

cmd/deploy/deploy.go
omnomagonz:
```suggestion
	oktetoLog.Information("Use `--remote` to run the deploy commands in your Okteto cluster. Learn more: https://www.okteto.com/docs/reference/manifest/#deploy-remotely")
```

This is as short as I can get it with keeping the link to docs. We could also just remove the docs link/copy since CLI doesn't seem like the right place for user education though I'm sure precedent exists for this in other CLI tools.

jLopezbarb: Agree on breaking the line and align the text. Other than that the text that Matt proposed looks good to me

AdrianPedriza: great! I've updated the message 470e907e6b4a3210cae1969a5ad7a6f17a0bb119

Captura de pantalla 2023-11-20 a las 17 28 04

pchico83 (approved):

PR #4049, opened 2023-11-17
Refactor: remove `CheckIfRegularFile` in favour of `filesystem.FileExistsAndNotDir`

andreafalzetti:
# Proposed changes

Changes and reasons behind this refactor:

- `CheckIfRegularFile` is located in `cmd/utils/dev.go` and it's a duplicated of `filesystem.FileExistsAndNotDir`
- `GetFilePathFromWdAndFiles` is not used anywhere
- `FileExistsAndNotDir` doesn't have any unit test
- `FileExistsWithFilesystem` doesn't have any unit test

## How to validate

1. Using `okteto/movies`
2. Try to deploy/build etc

## CLI Quality Reminders 🔧

### DIFF
--- a/cmd/build/remote/build.go
+++ b/cmd/build/remote/build.go
@@ -3,6 +3,7 @@ package remote
 import (
 	"context"
 	"fmt"
+	"github.com/okteto/okteto/pkg/filesystem"
 	"path/filepath"
 
 	"github.com/okteto/okteto/cmd/utils"
@@ -64,8 +65,8 @@ func (bc *OktetoBuilder) Build(ctx context.Context, options *types.BuildOptions)
 		options.File = filepath.Join(path, "Dockerfile")
 	}
 
-	if err := utils.CheckIfRegularFile(options.File); err != nil {
-		return fmt.Errorf("%s: %s", oktetoErrors.InvalidDockerfile, err.Error())
+	if exists := filesystem.FileExistsAndNotDir(options.File); !exists {
+		return fmt.Errorf("%s: is not a regular file", oktetoErrors.InvalidDockerfile)
 	}
 
 	if err := bc.Builder.Run(ctx, options); err != nil {
--- a/cmd/build/remote/build.go
+++ b/cmd/build/remote/build.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"github.com/okteto/okteto/pkg/filesystem"
+	"github.com/spf13/afero"
 	"path/filepath"
 
 	"github.com/okteto/okteto/cmd/utils"
@@ -65,7 +66,7 @@ func (bc *OktetoBuilder) Build(ctx context.Context, options *types.BuildOptions)
 		options.File = filepath.Join(path, "Dockerfile")
 	}
 
-	if exists := filesystem.FileExistsAndNotDir(options.File); !exists {
+	if exists := filesystem.FileExistsAndNotDir(options.File, afero.NewOsFs()); !exists {
 		return fmt.Errorf("%s: is not a regular file", oktetoErrors.InvalidDockerfile)
 	}
 
--- a/cmd/build/remote/build.go
+++ b/cmd/build/remote/build.go
@@ -67,7 +67,7 @@ func (bc *OktetoBuilder) Build(ctx context.Context, options *types.BuildOptions)
 	}
 
 	if exists := filesystem.FileExistsAndNotDir(options.File, afero.NewOsFs()); !exists {
-		return fmt.Errorf("%s: is not a regular file", oktetoErrors.InvalidDockerfile)
+		return fmt.Errorf("%s: '%s' is not a regular file", oktetoErrors.InvalidDockerfile, options.File)
 	}
 
 	if err := bc.Builder.Run(ctx, options); err != nil {
--- a/cmd/build/v1/build.go
+++ b/cmd/build/v1/build.go
@@ -16,6 +16,7 @@ package v1
 import (
 	"context"
 	"fmt"
+	"github.com/okteto/okteto/pkg/filesystem"
 	"path/filepath"
 
 	"github.com/okteto/okteto/pkg/model"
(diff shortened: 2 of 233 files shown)


pkg/cmd/build/build.go
```diff
@@ -387,13 +387,15 @@ func extractFromContextAndDockerfile(context, dockerfile, svcName string) string
 		return dockerfile
 	}
 
+	fs := afero.NewOsFs()
```
AdrianPedriza, line 390: can we inject it? in order to being able to test this fn using memmap

AdrianPedriza (approved):
LGTM. Just a minor comment: Could we inject `afero` from top in order to be able to test the internal fns? that way we can use memmap instead filesystem

cmd/build/remote/build.go
```diff
@@ -64,8 +66,8 @@ func (bc *OktetoBuilder) Build(ctx context.Context, options *types.BuildOptions)
 		options.File = filepath.Join(path, "Dockerfile")
 	}
 
-	if err := utils.CheckIfRegularFile(options.File); err != nil {
-		return fmt.Errorf("%s: %s", oktetoErrors.InvalidDockerfile, err.Error())
+	if exists := filesystem.FileExistsAndNotDir(options.File, afero.NewOsFs()); !exists {
+		return fmt.Errorf("%s: is not a regular file", oktetoErrors.InvalidDockerfile)
```
AdrianPedriza, line 70: we could be returning another error with this change. Is this change safe? Maybe the customer who consumes it acts on the error, that's my concern

pkg/filesystem/file.go
```diff
@@ -14,22 +14,21 @@
...
 	oktetoLog "github.com/okteto/okteto/pkg/log"
 	"github.com/spf13/afero"
+	"io"
+	"os"
 )
 
-func FileExistsWithFilesystem(name string, fs afero.Fs) bool {
-	_, err := fs.Stat(name)
+// FileExistsWithFilesystem return true if the file exists
+func FileExistsWithFilesystem(path string, fs afero.Fs) bool {
```
ifbyol, line 24: Nit: Not blocker at all.

Would it make sense to create an struct `FilsystemChecker` or similar with the fs as dependency? It seems odd (as a client) to pass the filesystem to the function

cmd/build/v1/build.go
```diff
@@ -85,8 +87,8 @@ func (bc *OktetoBuilder) Build(ctx context.Context, options *types.BuildOptions)
 		options.File = filepath.Join(path, "Dockerfile")
 	}
 
-	if err := utils.CheckIfRegularFile(options.File); err != nil {
-		return fmt.Errorf("%s: %s", oktetoErrors.InvalidDockerfile, err.Error())
+	if exists := filesystem.FileExistsAndNotDir(options.File, afero.NewOsFs()); !exists {
+		return fmt.Errorf("%s: is not a regular file", oktetoErrors.InvalidDockerfile)
```
ifbyol, line 91: Nit: Not a blocker as it was as this before but, it would make sense to include `options.File` in the error message? (maybe it is not always set and that is way we didn't include it)

ifbyol (approved):
Minor comments but nothing blocker

pkg/filesystem/file.go
jLopezbarb: Agree, this way we could instantiate it at the beginning of the main file and pass it only when it's needed without the need of initialising the `NewOsFs` all along the code

pkg/filesystem/file.go
```diff
@@ -81,23 +79,11 @@ func CopyFile(from, to string) error {
...
 }
 
-// FileExistsAndNotDir checks if the file exists and its not a dir
-func FileExistsAndNotDir(filename string) bool {
-	info, err := os.Stat(filename)
-	if err != nil && os.IsNotExist(err) {
+// FileExistsAndNotDir checks if the file exists, and it's not a dir
+func FileExistsAndNotDir(path string, fs afero.Fs) bool {
+	info, err := fs.Stat(path)
+	if err != nil {
```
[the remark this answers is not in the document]
jLopezbarb: Right now the only error that os.Stat can return is `os.NotExists`. I think it's safe to keep it like this but if they add another error other than that we can face some problems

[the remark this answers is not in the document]
ifbyol: Then, why we discriminate in the other scenarios? I think both functions have to be consistent to what we want. Also, where did you see that it can only return that error?

pkg/filesystem/file.go
ifbyol: sounds good

cmd/build/remote/build.go
andreafalzetti: adding this wrapping actually makes the error returned the same as before when using `CheckIfRegularFile`, if you see the return from that old func:

```go
fmt.Errorf("'%s' is not a regular file", path)
```

What differs is that if `fs.Stat` fails, we return a different error but I think it's ok.

cmd/build/v1/build.go
andreafalzetti: it's a good idea, I've checked and the `options.File` is always set, because if empty it defaults to path + `Dockerfile`. Changed in 21f6cd9e

pkg/filesystem/file.go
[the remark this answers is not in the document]
andreafalzetti: The reason why I've removed the check on `os.IsNotExist` is because in case `fs.Stat` fails, I don't think it's safe to let the execution continue and reach `info.IsDir` because `info` might be `nil` and locally it should not be checked.

At the same time in the other function `FileExistsWithFilesystem` I initially deleted the whole function, and then put it back because I am trying to do the smallest possible PR. 

This file and a lot of parts using this pkg need to be refactored so to unblock this PR now I will put back the check, we know future refactorings will come and we can consolidate the behavioir in the future.

Changed in ff715d15

jLopezbarb (approved):

PR #4051, opened 2023-11-20
feat: add feature flag to enable context build

jLopezbarb:
Fixes LAKE-94

# Proposed changes

**Introduction of serviceHasher and OktetoSmartBuildUsingContextEnvVar:**

- The `serviceHasher` generates hashes based on different criteria, such as the project commit or the build context. If the developer doesn't specify which one to use `serviceHasher` will use the one based on the feature flag
- The introduction of `OktetoSmartBuildUsingContextEnvVar` as a variable to control whether to use the build context for hashing. This provides flexibility in choosing between different hashing strategies.

**Changes in oktetoBuilderConfig and build.go:**

- The `GetBuildContextHash` method in `oktetoBuilderConfig` has been removed, and its functionality has been moved to the serviceHasher.
- The usage of `getBuildHashFromCommit` has been replaced with calls to `serviceHasher` methods (HashProjectCommit, HashBuildContext, and HashService). This makes that we don't have to rely on a magic string that is set inside the `getBuildHashFromCommit` function. 
- The `BuildContextHash` is now calculated using serviceHasher. When we want to specify if we want to hash using an specific method we have those methods but if we want to hash based on the feature flag we can do it also.

**Changes in tests:**

I have modified the tests in order to use the new serviceHasher instead of the previous method.

Ensure that the new hasher still hashes as the way it was before, by not changing the tests

## How to validate

1. Build a project as usual
2. Set the feature flag(`OKTETO_SMART_BUILDS_USING_BUILD_CONTEXT`) to true
3. Build again the project and check that the commit is different

## CLI Quality Reminders 🔧


cmd/build/v2/hasher.go
```diff
@@ -0,0 +1,121 @@
...
+	fmt.Fprintf(&b, "context:%s;", buildInfo.Context)
+	fmt.Fprintf(&b, "dockerfile:%s;", buildInfo.Dockerfile)
+	fmt.Fprintf(&b, "image:%s;", buildInfo.Image)
+
+	oktetoBuildHash := sha256.Sum256([]byte(b.String()))
+	return hex.EncodeToString(oktetoBuildHash[:])
+}
+
+// LoadBoolean loads a boolean environment variable and returns it value
+func LoadBoolean(k string) bool {
```
jLopezbarb, line 109: Waiting for https://github.com/okteto/okteto/pull/4047 to be merged and remove from here

cmd/build/v2/hasher.go
```diff
@@ -0,0 +1,121 @@
...
+
+	// We use a builder to avoid allocations when building the string
+	var b strings.Builder
+
+	fmt.Fprintf(&b, "%s:%s;", commitType, commitHash)
+	fmt.Fprintf(&b, "target:%s;", buildInfo.Target)
+	fmt.Fprintf(&b, "build_args:%s;", argsText)
+	fmt.Fprintf(&b, "secrets:%s;", secretsText)
+	fmt.Fprintf(&b, "context:%s;", buildInfo.Context)
+	fmt.Fprintf(&b, "dockerfile:%s;", buildInfo.Dockerfile)
```
pchico83, line 101: when getting a hash for the build context, are we considering the content of the Dockerfile?
If the Dockerfile isn't in the build context path, we should hash the Dockerfile content, no just the file path

cmd/build/v2/config.go
```diff
@@ -114,16 +113,6 @@ func (oc oktetoBuilderConfig) GetAnonymizedRepo() string {
...
-	buildContext := buildInfo.Context
-	treeHash, err := oc.repository.GetTreeHash(buildInfo.Context)
-	if err != nil {
-		oktetoLog.Info("error trying to get tree hash for build context '%s': %w", buildContext, err)
-	}
-
-	return getBuildHashFromGitHash(buildInfo, treeHash, "tree_hash")
-}
-
 func (oc oktetoBuilderConfig) IsSmartBuildsEnabled() bool {
```
pchico83, line 116: How is `IsSmartBuildsEnabled` implemented? I think we have three different cases:
- Previous pull first approach based on commit -> default option
- Disable any kind of build optimization -> always build -> I think there is an envvar for this
- Smart builds based on build context 
Could we manage the three options with a single envvar?

andreafalzetti:
Let's have a coffee review about this PR and update the pull first approach docs together 🙏

cmd/build/v2/config.go
andreafalzetti: I would also have a single env var to control the behaviour, otherwise as a user I need to know and worry about 2 env vars and all combinations (true+true, true+false, etc)

jLopezbarb: I prefer to have them separate. If the `OKTETO_SMART_BUILDS_ENABLED` is true then we don't check anything about the other variable. I agree that it can cause some troubles in the future but then I think the naming is not the best one.

If we go with only one variable, what should be the naming: `OKTETO_SMART_BUILD` with accepted values (`default`, `project` or `empty`/`none`/`buildcontext`)?

pchico83: @jLopezbarb which envvars names and values are currently implemented in this PR?

jLopezbarb: @pchico83 Right now we have `OKTETO_SMART_BUILDS_USING_BUILD_CONTEXT` in order to use the build context instead the project commit and `OKTETO_SMART_BUILDS_ENABLED` to disable the feature.

cmd/build/v2/hasher.go
jLopezbarb: Done https://github.com/okteto/okteto/pull/4051/commits/2a85625984827c65442e300df84aeeefa5b83968

cmd/build/v2/config.go
andreafalzetti: sounds good to me 👍

teresaromero (approved):

cmd/build/v2/hasher.go
```diff
@@ -0,0 +1,138 @@
...
+	}
+	return sh.hash(buildInfo, projectCommitType, sh.projectCommit)
+}
+
+func (sh serviceHasher) HashBuildContext(buildInfo *model.BuildInfo) string {
+	buildContext := buildInfo.Context
+	if buildContext == "" {
+		buildContext = "."
+	}
+	treeHash, err := sh.gitRepoCtrl.GetTreeHash(buildContext)
```
AdrianPedriza, line 69: worth create a singleton also for treeHash? that way we don't calculate it each time we need that value

cmd/build/v2/hasher.go
```diff
@@ -0,0 +1,138 @@
...
+	if err != nil {
+		oktetoLog.Info("error trying to read Dockerfile: %w", err)
+		return ""
+	}
+	encodedFile := sha256.Sum256(content)
+	return hex.EncodeToString(encodedFile[:])
+}
+
+// LoadBoolean loads a boolean environment variable and returns it value
+func LoadBoolean(k string) bool {
```
AdrianPedriza, line 126: I feel we already have this logic. Is it?

jLopezbarb: Yup https://github.com/okteto/okteto/pull/4051#discussion_r1399350133

AdrianPedriza:
should we update our docs about this change?

cmd/build/v2/hasher.go
```diff
@@ -0,0 +1,129 @@
...
+	"os"
+	"strings"
+
+	"github.com/okteto/okteto/pkg/env"
+	oktetoLog "github.com/okteto/okteto/pkg/log"
+	"github.com/okteto/okteto/pkg/model"
+)
+
+const (
+	OktetoSmartBuildUsingContextEnvVar = "OKTETO_SMART_BUILDS_USING_BUILD_CONTEXT"
```
AdrianPedriza, line 29: can we add a description for this feature flag?

jLopezbarb:
> should we update our docs about this change?

We didn't have any documentation for the `SMART_BUILD_ENABLED` so I also added [here](https://github.com/okteto/docs/pull/534)

AdrianPedriza:
if I set `OKTETO_SMART_BUILDS_USING_BUILD_CONTEXT`=true and run `okteto build` if images are already built it says: 
Captura de pantalla 2023-11-24 a las 9 23 44
`already built for commit ...` but I'm not checking the repo commit, I'm using the build context hash. Can we modify this message according to the commit/build context hash use?

AdrianPedriza:
I'm facing some odd behavior. This is the steps to repro it, not sure if it is related with your changes:
1. using a clean okteto/movies
2. run `okteto build` 
3. run `export OKTETO_SMART_BUILDS_USING_BUILD_CONTEXT=true``
4. run `okteto build` 
5. run `export OKTETO_SMART_BUILDS_ENABLED=false`
6. run `okteto build`
7. run `export OKTETO_SMART_BUILDS_ENABLED=true``
8. run `okteto build`, see:
Captura de pantalla 2023-11-24 a las 9 28 43

jLopezbarb:
> 7\. OKTETO_SMART_BUILDS_ENABLED

Even though this error was outside the scope of the PR I addressed on the PR to fix it

cmd/build/v2/hasher.go
```diff
@@ -0,0 +1,145 @@
...
+}
+
+func newServiceHasher(gitRepoCtrl repositoryCommitRetriever) *serviceHasher {
+	return &serviceHasher{
+		gitRepoCtrl:       gitRepoCtrl,
+		buildContextCache: map[string]string{},
+	}
+}
+
+// HashProjectCommit hashes the
```
andreafalzetti, line 55:
```suggestion
// HashProjectCommit hashes the repository's commit
```

cmd/build/v2/hasher.go
```diff
@@ -0,0 +1,145 @@
...
+		var err error
+		sh.projectCommit, err = sh.gitRepoCtrl.GetSHA()
+		if err != nil {
+			oktetoLog.Infof("could not get repository sha: %w", err)
+		}
+	}
+	return sh.hash(buildInfo, projectCommitType, sh.projectCommit)
+}
+
+func (sh serviceHasher) HashBuildContext(buildInfo *model.BuildInfo) string {
```
andreafalzetti, line 67:
```suggestion
// HashBuildContext returns the hash of the service using its context tree hash
func (sh serviceHasher) HashBuildContext(buildInfo *model.BuildInfo) string {
```

cmd/build/v2/hasher.go
```diff
@@ -0,0 +1,145 @@
...
+		sh.buildContextCache[buildContext], err = sh.gitRepoCtrl.GetTreeHash(buildContext)
+		if err != nil {
+			oktetoLog.Info("error trying to get tree hash for build context '%s': %w", buildContext, err)
+		}
+	}
+
+	return sh.hash(buildInfo, buildContextCommitType, sh.buildContextCache[buildContext])
+}
+
+func (sh serviceHasher) HashService(buildInfo *model.BuildInfo) string {
```
andreafalzetti, line 83:
```suggestion
// HashService returns the hashed project commit by default. If smart-builds use the context it returns the hash of the service given its git tree hash
func (sh serviceHasher) HashService(buildInfo *model.BuildInfo) string {
```

andreafalzetti: feel free to change/adjust it

andreafalzetti: If we expose `HashService`, can we make `HashProjectCommit` and `HashBuildContext` private?

cmd/build/v2/hasher.go
```diff
@@ -0,0 +1,147 @@
...
+
+func (sh serviceHasher) GetCommitHash(buildInfo *model.BuildInfo) string {
+	if env.LoadBoolean(OktetoSmartBuildUsingContextEnvVar) {
+		buildContext := buildInfo.Context
+		if buildContext == "" {
+			buildContext = "."
+		}
+		if commit, ok := sh.buildContextCache[buildContext]; ok {
+			return commit
+		}
```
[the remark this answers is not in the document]
AdrianPedriza: already checked with Javi

AdrianPedriza (approved):
LGTM, taking into account that the calculation of whether the service should be built is always based on whether there are changes with respect to the project commit and that it will adapt to the new implementation taking into account the context of the service only to decide whether to build the image or not.

andreafalzetti (approved):

PR #4052, opened 2023-11-21
fix: use okteto api to retrieve dev env endpoints

AdrianPedriza:
## Changes
- use okteto api to retrieve endpoints using `space` query
- use old logic for vanilla or fallbacks if endpoint is not available
- remove external endpoints retrieval when backend is used to retrieve endpoints

## How to test it
- clone okteto/movies
- cd to the repo
- deploy okteto/movies with an external resource (`okteto deploy`)
- check all endpoints are correct comparing them with endpoint in okteto UI when exec `okteto endpoints`
---
### With/without changes
- clone https://github.com/okteto/divert-with-istio-sample
- cd divert-with-istio-sample
- using my own cluster `okteto ctx use <my-context>`. You need to enable istio and an okteto feature flag in order to use virtual services so is easy to use my cluster already prepared.
- run `okteto deploy -f okteto-staging.yml`
- check endpoint in UI (broken because we need an ingress but enough to test the needed logic)
- run `okteto endpoints -f okteto-staging.yml` will show same endpoints than UI using this PR and no endpoints with previous implementation

### DIFF
--- a/cmd/deploy/endpoints.go
+++ b/cmd/deploy/endpoints.go
@@ -26,6 +26,7 @@ import (
 	"github.com/okteto/okteto/pkg/devenvironment"
 	"github.com/okteto/okteto/pkg/endpoints"
 	oktetoErrors "github.com/okteto/okteto/pkg/errors"
+	"github.com/okteto/okteto/pkg/externalresource"
 	"github.com/okteto/okteto/pkg/format"
 	"github.com/okteto/okteto/pkg/k8s/ingresses"
 	oktetoLog "github.com/okteto/okteto/pkg/log"
@@ -47,14 +48,20 @@ type endpointGetterInterface interface {
 	List(ctx context.Context, ns string, labelSelector string) ([]string, error)
 }
 
+type externalResourceControlInterface interface {
+	List(ctx context.Context, ns string, labelSelector string) ([]externalresource.ExternalResource, error)
+}
+
 type k8sIngressClientProvider interface {
 	GetIngressClient() (*ingresses.Client, error)
 }
 
 type EndpointGetter struct {
-	GetManifest       func(path string) (*model.Manifest, error)
-	endpointControl   endpointGetterInterface
-	K8sClientProvider k8sIngressClientProvider
+	GetManifest                  func(path string) (*model.Manifest, error)
+	endpointControl              endpointGetterInterface
+	externalResourceControl      externalResourceControlInterface
+	getEndpointsInStandaloneMode func(context.Context, *EndpointsOptions, string, k8sIngressClientProvider) ([]string, error)
+	K8sClientProvider            k8sIngressClientProvider
 }
 
 func NewEndpointGetter() (EndpointGetter, error) {
@@ -63,10 +70,20 @@ func NewEndpointGetter() (EndpointGetter, error) {
 		return EndpointGetter{}, fmt.Errorf("error getting okteto client: %w", err)
 	}
 
+	k8sProvider := okteto.NewK8sClientProvider()
+	_, cfg, err := k8sProvider.Provide(okteto.Context().Cfg)
+	if err != nil {
+		return EndpointGetter{}, fmt.Errorf("error getting kubernetes client: %w", err)
+	}
+
+	erc := externalresource.NewExternalK8sControl(cfg)
+
 	return EndpointGetter{
-		GetManifest:       model.GetManifestV2,
-		endpointControl:   ec,
-		K8sClientProvider: okteto.NewK8sClientProvider(),
+		GetManifest:                  model.GetManifestV2,
+		endpointControl:              ec,
+		externalResourceControl:      erc,
+		getEndpointsInStandaloneMode: getEndpointsStandaloneMode,
+		K8sClientProvider:            k8sProvider,
 	}, nil
 
 }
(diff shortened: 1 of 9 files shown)


pchico83:
@AdrianPedriza we don't need to wait until the backend changes. The code should handle the error and default to the previous behavior. Otherwise, we will be breaking the newer CLIs using old versions of the Okteto chart

AdrianPedriza:
> @AdrianPedriza we don't need to wait until the backend changes. The code should handle the error and default to the previous behavior. Otherwise, we will be breaking the newer CLIs using old versions of the Okteto chart

Absolutely right! I will fallback to the old implementation in that case then

AdrianPedriza:
I'll convert it to draft because I need to call `spaces` endpoint which has the needed info in order to not create a new endpoint in backend

pkg/types/interface.go
```diff
@@ -85,3 +85,8 @@ type KubetokenInterface interface {
 	GetKubeToken(baseURL, namespace string) (KubeTokenResponse, error)
 	CheckService(baseURL, namespace string) error
 }
+
+// EndpointInterface represents the endpoint client
+type EndpointInterface interface {
```
teresaromero, line 90:
```suggestion
type EndpointClientInterface interface {
```

pkg/endpoints/control.go
```diff
@@ -0,0 +1,25 @@
...
+	"context"
+
+	"github.com/okteto/okteto/pkg/okteto"
+)
+
+type APIControl struct {
+	OktetoClient *okteto.OktetoClient
+}
+
+func NewEndpointControl() (*APIControl, error) {
```
jLopezbarb, line 13: Should we add the minimum interface as parameter? That way we can test it from the calling function and be able to test it from the calling function.

cmd/deploy/endpoints.go
```diff
@@ -180,23 +195,33 @@ func (eg *EndpointGetter) getEndpoints(ctx context.Context, opts *EndpointsOptio
...
-		return nil, err
-	}
 
-	externalEps, err := eg.endpointControl.List(ctx, opts.Namespace, labelSelector)
-	if err != nil {
-		return nil, err
-	}
+	var eps []string
+	var err error
+	if okteto.Context().IsOkteto {
```
jLopezbarb, line 201: I think this should be an interface. This way we can encapsulate it into a single function making it more readable and the resulting test would be more readable. There is no need if I'm on a Vanilla cluster to have access to `endpointControl` and `externalResourceControl`WDYT?

pkg/okteto/endpoint.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+
+	err := query(ctx, &queryStruct, variables, c.client)
+	if err != nil {
+		return nil, err
+	}
+
+	endpoints := make([]string, 0)
+	for _, deployment := range queryStruct.Response.Deployments {
+		if deployment.DeployedBy == graphql.String(deployedBy) {
```
jLopezbarb, line 55: NIT: Given that all of them are components. Should we encapsulate into a function so we don't duplicate code?

cmd/deploy/endpoints.go
```diff
@@ -180,23 +195,33 @@ func (eg *EndpointGetter) getEndpoints(ctx context.Context, opts *EndpointsOptio
...
+		if err != nil {
+			eps, err = eg.getEndpointsInStandaloneMode(ctx, opts, labelSelector, eg.K8sClientProvider)
+			if err != nil {
+				return nil, err
+			}
 
-	for _, externalEp := range externalEps {
-		for _, ep := range externalEp.Endpoints {
-			eps = append(eps, fmt.Sprintf("%s (external)", ep.Url))
+			externalEps, err := eg.externalResourceControl.List(ctx, opts.Namespace, labelSelector)
```
pchico83, line 209: Do we need to query external resources in this case?
If the endpoints query fails, I would show an error and continue. Showing the endpoints is a nice to have, but the command shouldn't fail

teresaromero (approved):
Functionality tested 👍🏻  deferring approval to address of code feedback

cmd/deploy/endpoints.go
AdrianPedriza: Yes, that fallback I think is not needed anymore. I did it thinking about creating a new endpoint in the backend and scenarios where okteto chart deployed is not the one containing the new endpoint.. but finally I've reuse `space` query so it shouldn't fail. If the use of the backend endpoint fails I think we should return the error and not fallback. 6ebccf1ff64970e5b09516c5eab7313de63aeb35 WDYT?

pchico83: Makes sense

pkg/types/interface.go
AdrianPedriza: 6ebccf1ff64970e5b09516c5eab7313de63aeb35

cmd/deploy/endpoints.go
AdrianPedriza: e51c9a1cd38266d5e515e36628b2c415714d6e85

pkg/okteto/endpoint.go
AdrianPedriza: 6e3346e02dfdfc36d5ab324fda5207cdcd1e3f96

jLopezbarb (approved):

andreafalzetti (approved):

pkg/endpoints/control.go
AdrianPedriza: 52416511b7aff8495f32628b730d7ac78a10671e

PR #4057, opened 2023-11-21
CLI Quality - Lint - bodyclose

teresaromero:
# Proposed changes

Fixes https://okteto.atlassian.net/browse/LAKE-71

This Pr enables `bodyclose` as litern and resolves errors from it. 

## How to validate

Go to any client request and remove the closing of the response body or code a new request. Then run `make lint` to spot where the response hasnt closed the body.

## CLI Quality Reminders 🔧


jLopezbarb (approved):

AdrianPedriza (approved):

andreafalzetti:
Moved to draft to avoid merging without checking deepsource first (or hear from @teresaromero if it's done) and fixing the merge conflict.

pkg/okteto/kubetoken_client.go
```diff
@@ -80,7 +80,7 @@ func (c *kubeTokenClient) GetKubeToken(baseURL, namespace string) (types.KubeTok
 		return types.KubeTokenResponse{}, fmt.Errorf("failed to unmarshal kubetoken response: %w", err)
 	}
 
-	return kubeTokenResponse, nil
+	return kubeTokenResponse, resp.Body.Close()
```
jLopezbarb, line 83: Could we  move the `resp.Body.Close` to the line before and log it into the log file?

We are checking that `err != nil` from the calling function. If we get the response correctly but the body is not able to close we will be returning an error to the user that we could have ignore.

```suggestion
        defer func() {
          if err := resp.Body.Close(); err != nil {
            oktetoLog.Info("could not close the body: %s", err)
          }
        }
	return kubeTokenResponse, nil
```

pkg/okteto/kubetoken_client.go
```diff
@@ -98,5 +98,5 @@ func (c *kubeTokenClient) CheckService(baseURL, namespace string) error {
 		return fmt.Errorf("CheckService %w: %s", errKubetokenNotAvailable, baseURL)
 	}
 
-	return nil
+	return resp.Body.Close()
```
jLopezbarb, line 101: Same as above. We should be ignoring the error but logging it in case we need that information

jLopezbarb (changes requested):

pkg/okteto/kubetoken_client.go
teresaromero: changed the code, let me know if 👍🏻

andreafalzetti (approved):

andreafalzetti (approved):

PR #4061, opened 2023-11-22
CLI Quality - Lint - gomnd

teresaromero:
# Proposed changes

Fixes https://okteto.atlassian.net/browse/LAKE-75

This PR enables and fixes the problems raised by linter `gomnd`. this spots magic numbers.
Configuration has been set so we avoid magic numbers when passing arguments, and adding some ignored functions where we can use numbers as argument.

## How to validate

When passing some argument (int) into a function directly as number without previously declaring it as const

## CLI Quality Reminders 🔧

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -40,7 +40,7 @@ linters-settings:
   gomnd:
     checks:
       - argument
-      # - case
+      - case
       # - condition
       # - operation
       # - return
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -49,7 +49,6 @@ linters-settings:
       - '\.NewWriter'
       - '\.NewTabWriter'
       - '\.WriteFile'
-      - '^strings.'
       - '^time.'
       - '^os.'
       - '^bcrypt.'
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -41,7 +41,7 @@ linters-settings:
     checks:
       - argument
       - case
-      # - condition
+      - condition
       # - operation
       # - return
       # - assign
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -37,6 +37,28 @@ linters-settings:
       const:
         YEAR: '2023'
     template-path: .copyright-header.tmpl
+  gomnd:
+    checks:
+      - argument
+      # - case
+      # - condition
+      # - operation
+      # - return
+      # - assign
+    ignored-functions:
+      - '\.NewWriter'
+      - '\.NewTabWriter'
+      - '\.WriteFile'
+      - '^strings.'
+      - '^time.'
+      - '^os.'
+      - '^bcrypt.'
+      - '^math.'
+      - '^big.'
+      - '\.SetIndent'
+      - int
+      - make
+      - '\.WithTimeout'
 
 # All possible linters can be found https://golangci-lint.run/usage/linters/
 linters:
@@ -51,7 +73,7 @@ linters:
     - tenv
     - tparallel
     - gci
-    - goheader    
+    - goheader
     - unconvert
     - predeclared
     - asciicheck
@@ -60,6 +82,7 @@ linters:
     - decorder
     - exportloopref
     - makezero
+    - gomnd
 
 issues:
   # Maximum issues count per one linter. Set to 0 to disable. Default is 50.
--- a/cmd/utils/syncthing.go
+++ b/cmd/utils/syncthing.go
@@ -22,6 +22,10 @@ import (
 	decor "github.com/vbauerster/mpb/v7/decor"
 )
 
+const (
+	totalProgressValue = 100
+)
+
 // SyncthingProgress tracks the progress of all the files syncthing
 type SyncthingProgress struct {
 	progressContainer *mpb.Progress
@@ -38,7 +42,7 @@ func NewSyncthingProgressBar(width int) *SyncthingProgress {
 
 func (s *SyncthingProgress) initProgressBar() {
 	s.progressBar = s.progressContainer.Add(
-		100,
+		totalProgressValue,
 		nil,
 		mpb.PrependDecorators(
 			decor.OnComplete(decor.Spinner(nil, decor.WCSyncSpace), "Files synchronized"),
@@ -69,7 +73,7 @@ func (s *SyncthingProgress) SetCurrent(v int64) {
 // Finish finishes the progress bar
 func (s *SyncthingProgress) Finish() {
 	if s.progressBar != nil {
-		s.progressBar.SetCurrent(100)
+		s.progressBar.SetCurrent(totalProgressValue)
 	}
 	s.progressContainer.Wait()
 }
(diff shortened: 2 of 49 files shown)


pkg/cmd/down/run.go
```diff
@@ -80,7 +80,8 @@ func Run(dev *model.Dev, app apps.App, trMap map[string]*apps.Translation, wait
 		return nil
 	}
 
-	waitForDevPodsTermination(ctx, c, dev, 30)
+	t := 30
```
jLopezbarb, line 83:
```suggestion
	devPodTerminationRetries := 30
```

pkg/cmd/stack/deploy.go
```diff
@@ -770,7 +770,8 @@ func waitForPodsToBeRunning(ctx context.Context, s *model.Stack, c kubernetes.In
 	}
 
 	ticker := time.NewTicker(100 * time.Millisecond)
-	timeout := time.Now().Add(600 * time.Second)
+	timeoutDuration := 600 * time.Second
```
jLopezbarb, line 773:
```suggestion
	timeoutDuration := 10 * time.Minute
```

pkg/log/spinner.go
```diff
@@ -56,7 +56,8 @@ func (sl *spinnerLogger) unhold() {
 }
 
 func newSpinner() *sp.Spinner {
-	spinner := sp.New(sp.CharSets[14], 100*time.Millisecond, sp.WithHiddenCursor(true))
+	duration := 100 * time.Millisecond
```
jLopezbarb, line 59:
```suggestion
	rotationSpeed := 100 * time.Millisecond
```

.golangci.yml
```diff
@@ -37,6 +37,28 @@ linters-settings:
       const:
         YEAR: '2023'
     template-path: .copyright-header.tmpl
+  gomnd:
+    checks:
+      - argument
+      # - case
+      # - condition
```
jLopezbarb, line 44: I think it's worth checking this too. I think it's where most magic numbers occurs and one of the hardest to know why we set that number in the if statement. WDYT?

pkg/log/spinner.go
```diff
@@ -56,7 +56,8 @@ func (sl *spinnerLogger) unhold() {
...
+	spinner := sp.New(sp.CharSets[14], duration, sp.WithHiddenCursor(true))
```
teresaromero, line 60:
```suggestion
	spinner := sp.New(sp.CharSets[14], rotationSpeed, sp.WithHiddenCursor(true))
```

.golangci.yml
teresaromero: i've activated them and fixed the numbers, i have couple which i am not sure of the meaning, maybe you can give me some feedback?

```
cmd/up/activate.go:191:27: mnd: Magic number: 3, in <condition> detected (gomnd)
pkg/k8s/volumes/crud.go:195:14: mnd: Magic number: 5, in <condition> detected (gomnd)
pkg/registry/image.go:94:34: mnd: Magic number: 2, in <condition> detected (gomnd)
```

pkg/cmd/build/build.go
```diff
@@ -43,6 +43,8 @@ import (
 const (
 	warningDockerfilePath   string = "Build '%s': Dockerfile '%s' is not in a relative path to context '%s'"
 	doubleDockerfileWarning string = "Build '%s': Two Dockerfiles discovered in both the root and context path, defaulting to '%s/%s'"
+	maxArgFormatParts              = 2
+	minArgFormatParts              = 1
```
AdrianPedriza, line 47: I am not sure if it is correct to have this value as a constant. I think you should reduce its scope to the function where it is used but I would not reuse its value under a constant, I see it quite error prone as each function may require strings.SplitN to split it into N.

pkg/k8s/namespaces/namespace.go
```diff
@@ -260,11 +260,13 @@ func NewTrip(restConfig *rest.Config, opts *Options) (*Trip, error) {
 		return nil, err
 	}
 
+	defaultMaxWeight := int64(10)
```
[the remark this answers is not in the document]
andreafalzetti: this var is passed to `semaphore.NewWeighted` which needs a `int64` so we cannot change it

.golangci.yml
jLopezbarb: `cmd/up/activate.go:191:27: mnd: Magic number: 3, in <condition> detected (gomnd)`: We run a start script on the pod and get three lines as the output for getting some information. Should we call it `minStartScriptLines`?
`pkg/k8s/volumes/crud.go:195:14: mnd: Magic number: 5, in <condition> detected (gomnd)`: This magic number is to print the log every 5 times so we don't get too much logs with the same information. Maybe we can call it `logFrequency`?
`pkg/registry/image.go:94:34: mnd: Magic number: 2, in <condition> detected (gomnd)`: This magic number is the one that checks that the image is composed as `registry.com/image`. Maybe we can call it `registryImageParts`?

WDYT?

.golangci.yml
```diff
@@ -37,6 +37,28 @@ linters-settings:
...
     template-path: .copyright-header.tmpl
+  gomnd:
+    checks:
+      - argument
+      - case
+      - condition
+      # - operation
+      # - return
+      # - assign
+    ignored-functions:
```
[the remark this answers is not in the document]
andreafalzetti: I agree, I've changed this in f555daf6c0f7fe7607d692e12b46115c2dee2caf

pkg/cmd/build/build.go
andreafalzetti: There are quite a few usages of these 2 vars. Being 1 and 2 I think it's safe to use them as `const`. How do you see this be error prone?

cmd/utils/syncthing.go
```diff
@@ -22,6 +22,10 @@ import (
 	decor "github.com/vbauerster/mpb/v7/decor"
 )
 
+const (
+	totalProgressValue = 100
```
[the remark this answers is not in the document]
andreafalzetti: I think it's ok to have this duplicated because they are different pkgs. How would you change it?

pkg/cmd/down/run.go
andreafalzetti: changed in fd0fc8a1

pkg/log/spinner.go
andreafalzetti: changed in 53a9af24

andreafalzetti (approved):

pkg/cmd/build/build.go
AdrianPedriza: What I mean is that using this as const for strings.splitN because they share the same value for me is like reusing both values because they happen to share value but not because in the places where they are used they have to have always the same value as to use the same constant in both cases

AdrianPedriza: now that we have added strings.splitn in the list,  I think that using the magic number directly in strings.splitn makes more sense than reusing it through a constant.

AdrianPedriza (approved):

PR #4064, opened 2023-11-22
CLI Quality - Lint-  musttag

teresaromero:
# Proposed changes

Fixes https://okteto.atlassian.net/browse/LAKE-77

This PR enables lint musttag to check all the structs are tagged correctly.
This lint needs the version of the runner upgraded, so updated the image to 2.3.3 at circleci config

## How to validate

Any struct without json or yaml tags wich is serialize should have this properties

## CLI Quality Reminders 🔧

### DIFF
--- a/pkg/model/dev.go
+++ b/pkg/model/dev.go
@@ -230,8 +230,8 @@ type Volume struct {
 
 // Sync represents a sync info in the development container
 type Sync struct {
-	LocalPath      string
-	RemotePath     string
+	LocalPath      string       `json:"-"`
+	RemotePath     string       `json:"-"`
 	Folders        []SyncFolder `json:"folders,omitempty" yaml:"folders,omitempty"`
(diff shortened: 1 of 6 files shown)


AdrianPedriza (approved):

pkg/model/dev.go
```diff
@@ -230,8 +230,8 @@ type Volume struct {
 
 // Sync represents a sync info in the development container
 type Sync struct {
-	LocalPath      string
-	RemotePath     string
+	LocalPath      string       `json:"-"`
```
jLopezbarb, line 233: I don't think we use json for this struct at all. Should we change it to yaml or add it to the tags?

andreafalzetti: agreed, changed to cd97c3a9

andreafalzetti (approved):

PR #4070, opened 2023-11-28
refactor: remove okteto context global var use in okteto build command

AdrianPedriza:
this PR is intended to inject the okteto context values into the build command

### Changes
- create `pkg/build/context.go` to hold the struct to inject in the build command
- read context before start build command in order to inject okteto context
### How to test
- check `build` command works as expected. No panics detected or weird behaviors.


pkg/cmd/build/build.go
```diff
@@ -95,30 +103,45 @@ func setOutputMode(outputMode string) string {
 
 }
 
+func GetRegistryConfigFromOktetoConfig(okCtx buildCtx.OktetoContextInterface) *okteto.Config {
```
jLopezbarb, line 106: Why do we need it to be on `pkg/cmd/build`? I don't think we should grow the `pkg/cmd/build` package. I think it'd be easier if it lives under the registry package.

cmd/build/command.go
```diff
@@ -80,23 +84,26 @@ const (
...
 		Short: "Build and push the images defined in the 'build' section of your okteto manifest",
 		RunE: func(cmd *cobra.Command, args []string) error {
 			options.CommandArgs = args
-			bc := NewBuildCommand(at)
+
 			// The context must be loaded before reading manifest. Otherwise,
 			// secrets will not be resolved when GetManifest is called and
 			// the manifest will load empty values.
-			if err := bc.loadContext(ctx, options); err != nil {
+			oktetoContext, err := getOktetoContext(ctx, options)
```
jLopezbarb, line 99: NIT: Should we separate this in two different functions? one to load and one to get the context? Right now `getOktetoContext` is loading and getting the context. I think it's cleaner if we have it both separate

cmd/build/v2/build.go
```diff
@@ -25,6 +25,7 @@ import (
 
 	buildv1 "github.com/okteto/okteto/cmd/build/v1"
 	"github.com/okteto/okteto/pkg/analytics"
+	buildCtx "github.com/okteto/okteto/pkg/build"
```
jLopezbarb, line 28: I don't think we should be naming this package `buildCtx`. Right now it only has the `buildCtx` but it can grow and have other types like the `buildInfo`

cmd/build/v2/build.go
```diff
@@ -88,11 +91,12 @@ type OktetoBuilder struct {
 }
 
 // NewBuilder creates a new okteto builder
-func NewBuilder(builder OktetoBuilderInterface, registry oktetoRegistryInterface, analyticsTracker analyticsTrackerInterface) *OktetoBuilder {
+func NewBuilder(builder OktetoBuilderInterface, registry oktetoRegistryInterface, analyticsTracker analyticsTrackerInterface, okCtx *buildCtx.OktetoContext) *OktetoBuilder {
```
jLopezbarb, line 94: Why do we need to import `*buildCtx.OktetoContext` I think we should be going with `buildCtx.OktetoContextInterface`

cmd/context/use.go
```diff
@@ -147,6 +149,19 @@ func (c *ContextCommand) Run(ctx context.Context, ctxOptions *ContextOptions) er
 	return nil
 }
 
+// RunStateless is the fn to use until the refactoring of the context command itself if you want to make use
+// of an injected context instead of using the global context variable.
+func (c *ContextCommand) RunStateless(ctx context.Context, ctxOptions *ContextOptions) (*buildCtx.OktetoContext, error) {
```
jLopezbarb, line 153: Should we return an `OktetoContext` that has all the methods and we only need the interface in order to get use of it? Right now `RunStateless` is returning a `buildCtx` object. If we want to have more commands make use of this, we should be returning an object from `context` package

cmd/build/command.go
AdrianPedriza: now the concept *load* disappears in the command. From the `build` command POV, a function from context package returns an okteto context so I think is not bad to call `getOktetoContext` only.

cmd/context/use.go
AdrianPedriza: good point. I think we should return a general context okteto and have each command transform it to its specific context that will expose only the necessary methods. WDYT?

pkg/cmd/build/build.go
AdrianPedriza: IMO `registry` pkg shouldn't know about `okteto` pkg, we only need to know the behavior exposed in `registry.configInterface`

cmd/build/v2/build.go
AdrianPedriza: 31849e63db2647e6db80ddd77dafe83ad0ae5416

pkg/cmd/build/build.go
jLopezbarb: Then this should be on the okteto pkg and be reused by each command by implementing its own interface. WDYT?

cmd/build/command.go
```diff
@@ -28,8 +28,8 @@ import (
 	"github.com/okteto/okteto/cmd/namespace"
 	"github.com/okteto/okteto/cmd/utils"
 	"github.com/okteto/okteto/pkg/analytics"
-	buildCtx "github.com/okteto/okteto/pkg/build"
-	"github.com/okteto/okteto/pkg/cmd/build"
+	build "github.com/okteto/okteto/pkg/build"
```
jLopezbarb, line 31: Why setting an alias? Having without the alias you use `build`

cmd/utils/okteto.go
```diff
@@ -91,3 +91,26 @@ func ShouldCreateNamespace(ctx context.Context, ns string) (bool, error) {
 	}
 	return false, nil
 }
+
+// ShouldCreateNamespace checks if the user has access to the namespace.
+// If not, ask the user if he wants to create it
+func ShouldCreateNamespaceStateless(ctx context.Context, ns string, c *okteto.OktetoClient) (bool, error) {
```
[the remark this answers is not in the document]
andreafalzetti: Also, we could flag the func as deprecated if we want to promote the usage of the new one. WDYT?

cmd/build/v2/config.go
```diff
@@ -67,7 +66,7 @@ func getConfig(registry configRegistryInterface, gitRepo configRepositoryInterfa
 		hasGlobalAccess:     hasAccess,
 		isCleanProject:      isClean,
 		fs:                  afero.NewOsFs(),
-		isOkteto:            okteto.Context().IsOkteto,
```
andreafalzetti, line 70: why is this always `true`?

pkg/build/context.go
```diff
@@ -0,0 +1,131 @@
...
+func (oc *OktetoContext) GetCurrentName() string {
+	return oc.getCurrentOktetoContext().Name
+}
+
+func (oc *OktetoContext) GetCurrentCertStr() string {
+	return oc.getCurrentOktetoContext().Certificate
+}
+
+func (oc *OktetoContext) GetCurrentCert() *x509.Certificate {
+	return nil
```
andreafalzetti, line 72: Why is this always returning `nil`?

cmd/utils/okteto.go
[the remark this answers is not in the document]
jLopezbarb: If we create the client inside the function we can't test it given that it will always create the real client. I think it's ok to pass the client as argument or a provider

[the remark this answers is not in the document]
AdrianPedriza: great, I've refactor the code to reuse one inside other

cmd/build/v2/config.go
AdrianPedriza: I've changed to use global var when build command is not exec, check it!

pkg/build/context.go
AdrianPedriza: this function is not needed so I've removed

cmd/build/v2/build.go
```diff
@@ -89,18 +91,37 @@ type OktetoBuilder struct {
...
+		hasher:            newServiceHasher(gitRepo),
+		oktetoContext:     okCtx,
+	}
 }
 
 // NewBuilderFromScratch creates a new okteto builder
 func NewBuilderFromScratch(analyticsTracker analyticsTrackerInterface, ioCtrl *io.IOController) *OktetoBuilder {
-	builder := &build.OktetoBuilder{}
+	builder := &buildCmd.OktetoBuilder{
+		OktetoContext: &okteto.OktetoContextStateless{
```
andreafalzetti, line 121: why do we create `Okteto.OktetoContextStateless` twice?

cmd/build/v1/build.go
```diff
@@ -59,7 +60,11 @@ func NewBuilder(builder OktetoBuilderInterface, registry oktetoRegistryInterface
 
 // NewBuilderFromScratch creates a new okteto builder
 func NewBuilderFromScratch(ioCtrl *io.IOController) *OktetoBuilder {
```
AdrianPedriza, line 62: This builder uses okteto context global var because is the builder constructor used by other commands which do not have the context injected. As each command is refactored to remove the global context, each command should use a constructor that injects the context. `okteto build` command does not uses this logic,

cmd/build/v2/build_test.go
```diff
@@ -249,15 +260,6 @@ func TestValidateOptions(t *testing.T) {
 
 func TestOnlyInjectVolumeMountsInOkteto(t *testing.T) {
 	ctx := context.Background()
-	okteto.CurrentStore = &okteto.OktetoContextStore{
```
andreafalzetti, line 252: Should we remove all occurences of `okteto.CurrentStore` in the `build` pkg?

AdrianPedriza: done 2ea04fc0bb6661fe3d92345f7270effde1bf0290  c5801b6e8a28749b6d09cba17210e2b0f2f962ba

andreafalzetti (approved):
Great effort 👏 

Approving as I have tested and it works, and the changes LGTM.

If we could merge with positive code coverage it would be great 👍

jLopezbarb (approved):

PR #4080, opened 2023-12-05
Fix: detect top level git rep dir to set `OKTETO_GIT_BRANCH` and `OKTETO_GIT_COMMIT` properly when service is nested

andreafalzetti:
# Proposed changes

When deploying a service in a monorepo with a manifest per service, example:

```
my-app/
└── services/
    ├── api/
    │   └── okteto.yml
    └── frontend/
        └── okteto.yml      
```

And running `okteto deploy -f services/api/okteto.yml` the env vars `OKTETO_GIT_BRANCH` and `OKTETO_GIT_COMMIT` are not properly set.

In case of `OKTETO_GIT_COMMIT` we auto-generate a commit id because we cannot find the related `.git` folder and `OKTETO_GIT_BRANCH` is empty.

This is a problem if the user depends on such variables for deployment. In this PR we change the behaviour to attempt finding the parent `.git` directory

## How to validate

1. Test using a fork of `okteto/movies`: `git clone git@github.com:andreafalzetti/movies.git`
1. Switch branch `git checkout af/fast-12`
1. Run `okteto deploy -f services/api/okteto.yml`)
1. Make sure `OKTETO_GIT_BRANCH` and `OKTETO_GIT_COMMIT` are populed properly
1. Now repest the deploy from the `services/api` directory using `okteto deploy`
1. Check the env vars to make sure they are set correctly

## CLI Quality Reminders 🔧

### DIFF
--- a/cmd/deploy/deploy.go
+++ b/cmd/deploy/deploy.go
@@ -301,7 +301,7 @@ func (dc *DeployCommand) RunDeploy(ctx context.Context, deployOptions *Options) 
 
 	topLevelGitDir, err := repository.FindTopLevelGitDir(cwd, dc.Fs)
 	if err != nil {
-		oktetoLog.Warning("Repository not detected: the env vars 'OKTETO_GIT_BRANCH' and 'OKTETO_GIT_COMMIT' might not be available: %s.\n    For more information, check out: https://www.okteto.com/docs/cloud/okteto-cli/#built-in-environment-variables", err.Error())
+		oktetoLog.Warning("Repository not detected: the env vars '%s' and '%s' might not be available: %s.\n    For more information, check out: https://www.okteto.com/docs/cloud/okteto-cli/#built-in-environment-variables", constants.OktetoGitBranchEnvVar, constants.OktetoGitCommitEnvVar, err.Error())
 	}
 
 	if topLevelGitDir != "" {
--- a/cmd/deploy/deploy.go
+++ b/cmd/deploy/deploy.go
@@ -40,6 +40,7 @@ import (
 	"github.com/okteto/okteto/pkg/model"
 	"github.com/okteto/okteto/pkg/okteto"
 	oktetoPath "github.com/okteto/okteto/pkg/path"
+	"github.com/okteto/okteto/pkg/repository"
 	"github.com/okteto/okteto/pkg/types"
 	"github.com/spf13/afero"
 	"github.com/spf13/cobra"
@@ -286,18 +287,28 @@ func (dc *DeployCommand) RunDeploy(ctx context.Context, deployOptions *Options) 
 		return oktetoErrors.ErrDeployCantDeploySvcsIfNotCompose
 	}
 
-	cwd, err := os.Getwd()
-	if err != nil {
-		return fmt.Errorf("failed to get the current working directory: %w", err)
-	}
-
 	// We need to create a client that doesn't go through the proxy to create
 	// the configmap without the deployedByLabel
 	c, _, err := dc.K8sClientProvider.Provide(okteto.Context().Cfg)
 	if err != nil {
 		return err
 	}
-	dc.addEnvVars(cwd)
+
+	cwd, err := os.Getwd()
+	if err != nil {
+		return fmt.Errorf("failed to get the current working directory: %w", err)
+	}
+
+	topLevelGitDir, err := repository.FindTopLevelGitDir(cwd, afero.NewOsFs())
+	if err != nil {
+		oktetoLog.Warning("Some git env vars might not be available: %s", err.Error())
+	}
+
+	if topLevelGitDir != "" {
+		dc.addEnvVars(topLevelGitDir)
+	} else {
+		dc.addEnvVars(cwd)
+	}
 
 	if err := setDeployOptionsValuesFromManifest(ctx, deployOptions, cwd, c); err != nil {
 		return err
--- a/cmd/deploy/deploy.go
+++ b/cmd/deploy/deploy.go
@@ -299,7 +299,7 @@ func (dc *DeployCommand) RunDeploy(ctx context.Context, deployOptions *Options) 
 		return fmt.Errorf("failed to get the current working directory: %w", err)
 	}
 
-	topLevelGitDir, err := repository.FindTopLevelGitDir(cwd, afero.NewOsFs())
+	topLevelGitDir, err := repository.FindTopLevelGitDir(cwd, dc.Fs)
 	if err != nil {
 		oktetoLog.Warning("Some git env vars might not be available: %s", err.Error())
 	}
--- a/cmd/deploy/deploy.go
+++ b/cmd/deploy/deploy.go
@@ -301,7 +301,7 @@ func (dc *DeployCommand) RunDeploy(ctx context.Context, deployOptions *Options) 
 
 	topLevelGitDir, err := repository.FindTopLevelGitDir(cwd, dc.Fs)
 	if err != nil {
-		oktetoLog.Warning("Some git env vars might not be available: %s", err.Error())
+		oktetoLog.Warning("Repository not detected: the env vars 'OKTETO_GIT_BRANCH' and 'OKTETO_GIT_COMMIT' might not be available: %s.\n    For more information, check out: https://www.okteto.com/docs/cloud/okteto-cli/#built-in-environment-variables", err.Error())
 	}
 
 	if topLevelGitDir != "" {
--- a/cmd/deploy/deploy_test.go
+++ b/cmd/deploy/deploy_test.go
@@ -427,13 +427,13 @@ func TestCreateConfigMapWithBuildError(t *testing.T) {
 		Build:        true,
 	}
 
-	registry := newFakeRegistry()
-	builder := test.NewFakeOktetoBuilder(registry)
+	reg := newFakeRegistry()
+	builder := test.NewFakeOktetoBuilder(reg)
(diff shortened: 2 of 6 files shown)


cmd/deploy/deploy.go
```diff
@@ -286,18 +287,28 @@ func (dc *DeployCommand) RunDeploy(ctx context.Context, deployOptions *Options)
 		return oktetoErrors.ErrDeployCantDeploySvcsIfNotCompose
 	}
 
+	// We need to create a client that doesn't go through the proxy to create
```
andreafalzetti, line 290: I moved this logic up so that `cwd, err := os.Getwd()` is closer to the code using `cwd`

cmd/deploy/deploy.go
```diff
@@ -286,18 +287,28 @@ func (dc *DeployCommand) RunDeploy(ctx context.Context, deployOptions *Options)
...
+	}
+
+	if topLevelGitDir != "" {
+		dc.addEnvVars(topLevelGitDir)
+	} else {
+		dc.addEnvVars(cwd)
 	}
-	dc.addEnvVars(cwd)
 
 	if err := setDeployOptionsValuesFromManifest(ctx, deployOptions, cwd, c); err != nil {
```
andreafalzetti, line 313: In `setDeployOptionsValuesFromManifest` I keep using `cwd` because it infers the correct name of the subfolder, if I pass the`topLevelGitDir` the inferred name would be misleading

cmd/deploy/deploy.go
```diff
@@ -286,18 +287,28 @@ func (dc *DeployCommand) RunDeploy(ctx context.Context, deployOptions *Options)
...
-	// the configmap without the deployedByLabel
-	c, _, err := dc.K8sClientProvider.Provide(okteto.Context().Cfg)
+	topLevelGitDir, err := repository.FindTopLevelGitDir(cwd, afero.NewOsFs())
 	if err != nil {
-		return err
+		oktetoLog.Warning("Some git env vars might not be available: %s", err.Error())
+	}
+
+	if topLevelGitDir != "" {
+		dc.addEnvVars(topLevelGitDir)
```
andreafalzetti, line 308: decided to keep the call to `addEnvVars` twice, I've considered having a string defining the path, but it would make the code less readable IMHO

cmd/deploy/deploy_test.go
```diff
@@ -427,13 +427,13 @@ func TestCreateConfigMapWithBuildError(t *testing.T) {
 		Build:        true,
 	}
 
-	registry := newFakeRegistry()
```
andreafalzetti, line 430: rename to avoid imports name colliding

pkg/repository/git.go
```diff
@@ -256,3 +258,22 @@ func (ogr oktetoGitWorktree) Status(ctx context.Context, localGit LocalGitInterf
 
 	return oktetoGitStatus{status: status}, nil
 }
+
+func FindTopLevelGitDir(workingDir string, fs afero.Fs) (string, error) {
```
andreafalzetti, line 262: I've used this approach because the `git-go` library does not support `git rev-parse --show-toplevel` (See https://github.com/go-git/go-git/issues/74) and we would have a hard dependency on the `git` binary. This approach is also used in this [google project](https://github.com/GoogleContainerTools/skaffold/pull/275/files#diff-566b8bbd20b2d5f058993d060ab8bbd46a938b6186ac5bbcfd11adf271f807eeR96-R113)

cmd/deploy/deploy.go
```diff
@@ -286,18 +287,28 @@ func (dc *DeployCommand) RunDeploy(ctx context.Context, deployOptions *Options)
...
 		return fmt.Errorf("failed to get the current working directory: %w", err)
 	}
 
-	// We need to create a client that doesn't go through the proxy to create
-	// the configmap without the deployedByLabel
-	c, _, err := dc.K8sClientProvider.Provide(okteto.Context().Cfg)
+	topLevelGitDir, err := repository.FindTopLevelGitDir(cwd, afero.NewOsFs())
 	if err != nil {
-		return err
+		oktetoLog.Warning("Some git env vars might not be available: %s", err.Error())
```
andreafalzetti, line 304: This is a warning in case the user did not notice, but we don't stop the execution to allow users run `okteto` in directories that are not git repos.

The warning looks like this:
Screenshot 2023-12-05 at 10 21 12

AdrianPedriza:
I've tested and compare it with `2.22.3`, `OKTETO_GIT_COMMIT` has different value. It is expected?

cmd/deploy/deploy.go
jLopezbarb: This is just the k8s client, we are not calling to get any configmap on this call

cmd/deploy/deploy.go
jLopezbarb: Agree but I would use the info level, since it´s a very common use case that should not be debugged with such a low log level.

cmd/deploy/deploy_test.go
```diff
@@ -480,7 +480,11 @@ func TestCreateConfigMapWithBuildError(t *testing.T) {
 	assert.Equal(t, expectedCfg.Name, cfg.Name)
 	assert.Equal(t, expectedCfg.Namespace, cfg.Namespace)
 	assert.Equal(t, expectedCfg.Labels, cfg.Labels)
-	assert.Equal(t, expectedCfg.Data, cfg.Data)
+
+	keysToCompare := []string{"actionName", "name", "status", "filename", "icon", "yaml"}
```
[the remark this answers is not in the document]
jLopezbarb: Are they empty? We should check what's the UI is doing if we send the key but empty. I remember that we have a bug in the past because of that. Probably we need to only send them if they are not empty

cmd/deploy/deploy.go
AdrianPedriza: saying that because the comment of the fn

cmd/deploy/deploy.go
andreafalzetti: I am wondering how frequently it's a common use case against how many times this could occur by mistake? I suspect that most people use okteto with git repos and that's why I went for warning, because it would be very helpful to let the user know about a mistake, however I don't have the full visibility on usage so I would appreciate input from @rlamana

cmd/deploy/deploy.go
andreafalzetti: @AdrianPedriza it's not a change that I've introduced so I am not sure why the comment is there

cmd/deploy/deploy_test.go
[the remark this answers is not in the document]
andreafalzetti: @jLopezbarb they were empty before, now they are populated but it's a side effect because the path in the test is the okteto/okteto repo so it detects the current branch used in okteto/okteto

cmd/deploy/deploy.go
rlamana: I'm trying to understand this particular case, will this be shown when no `.git` folder is found? What use cases cover this situation?

andreafalzetti: This warning appears when the `.git` cannot be found (up until `/`).

Imagine I work without any version control, and I want to use `okteto`, I would get this warning. The idea is to warn users that do use git, but for some reasons are running `okteto deploy` and the `.git` cannot be found.

andreafalzetti: Updated in 0f5f092e

Screenshot 2023-12-07 at 14 57 54

cmd/deploy/deploy.go
```diff
@@ -286,18 +287,28 @@ func (dc *DeployCommand) RunDeploy(ctx context.Context, deployOptions *Options)
...
 		return fmt.Errorf("failed to get the current working directory: %w", err)
 	}
 
-	// We need to create a client that doesn't go through the proxy to create
-	// the configmap without the deployedByLabel
-	c, _, err := dc.K8sClientProvider.Provide(okteto.Context().Cfg)
+	topLevelGitDir, err := repository.FindTopLevelGitDir(cwd, dc.Fs)
 	if err != nil {
-		return err
+		oktetoLog.Warning("Repository not detected: the env vars 'OKTETO_GIT_BRANCH' and 'OKTETO_GIT_COMMIT' might not be available: %s.\n    For more information, check out: https://www.okteto.com/docs/cloud/okteto-cli/#built-in-environment-variables", err.Error())
```
AdrianPedriza, line 304: can we use the values of the constant for message formatting?

AdrianPedriza (approved):
just a couple of doubts but they are not blocking I think

cmd/deploy/deploy.go
andreafalzetti: good idea, changed in https://github.com/okteto/okteto/pull/4080/commits/50b0fc0cd2b043329f6497ec5776872c741c8ea9#diff-15a334b420072b3d7ab3589a233e57b32c74c7b83ec2af01e0c939cf3ef11afa

pkg/repository/repository.go
```diff
@@ -65,7 +65,7 @@ func getURLFromPath(path string) repositoryURL {
 func NewRepository(path string) Repository {
 	repoURL := getURLFromPath(path)
 
-	var controller repositoryInterface = newGitRepoController()
+	var controller repositoryInterface = newGitRepoController(path)
```
andreafalzetti, line 68: withput passing `path` to the repo controller, when we check if the repo is clean, we check against an empty path and the result is always false (dirty)

PR #4083, opened 2023-12-07
feat: use git commit instead of git tree hash

jLopezbarb:
# Proposed changes
Fixes LAKE-124
## Refactoring from Tree Hash to Latest Commit:
 - Renamed the `GetTreeHash` method to `GetLatestDirCommit` in relevant interfaces and structs.
- Updated method calls and tests across the codebase to reflect the switch from tree hash to the latest commit.
## Removal of Unused Parameters:
- Removed the `commitType` parameter from the hash method in the serviceHasher struct. It will always be `commit` now (removed the `tree_hash`) 
- Eliminated conditional checks related to commitType in the hash method. We are going to calculate the same way for the build context and the project commit. This way if one of the methods has already built the image we don't need to rebuild the image again.
## Introduction of `GetLatestCommit` Method:
- Added the `GetLatestCommit` method to the LocalGitInterface interface.
- Implemented the `GetLatestCommit` method in the LocalGit struct.
- If git is installed and in path we'll use `git --no-optional-locks log -n 1 --pretty=format:%H -- path`. If git doesn't exists we will get the latest commit using the library (can take several minutes to finish, that's why git is the first option).

## How to validate

1. Run `gh repo clone okteto/movies`
1. Run `cd movies`
1. Run `export OKTETO_SMART_BUILDS_ENABLED=true`
1. Run `okteto build` twice 
1. Check that the commit that has been built is the latest commit of the dir.

## CLI Quality Reminders 🔧


pkg/repository/git.go
```diff
@@ -122,53 +120,60 @@ func (r gitRepoController) getSHA() (string, error) {
...
 
-func getTreeDirFromRelPath(relPath string) (string, error) {
-
-	wdCtrl := filesystem.NewOsWorkingDirectoryCtrl()
-	wd, err := wdCtrl.Get()
+func (r gitRepoController) calculateLatestDirCommit(ctx context.Context, contextDir string) (string, error) {
+	repo, err := r.repoGetter.get(r.path)
 	if err != nil {
-		return "", err
+		return "", fmt.Errorf("failed to analyze git repo: %w", err)
```
andreafalzetti, line 173: shall we have a more specific error string here? I suggest this because we already have a few errors that say `failed to analyze git repo` and I am afraid it will become difficult to identify quickly from which of these errors the error is coming from.

pkg/repository/git.go
```diff
@@ -201,8 +210,8 @@ func (ogr oktetoGitRepository) Head() (*plumbing.Reference, error) {
 	return ogr.repo.Head()
 }
 
-func (ogr oktetoGitRepository) CommitObject(h plumbing.Hash) (gitCommitInterface, error) {
-	return ogr.repo.CommitObject(h)
+func (ogr oktetoGitRepository) Log(o *git.LogOptions) (object.CommitIter, error) {
+	return ogr.repo.Log(o)
```
andreafalzetti, line 214: what does this do?

pkg/repository/git.go
```diff
@@ -122,53 +120,60 @@ func (r gitRepoController) getSHA() (string, error) {
...
-		return "", fmt.Errorf("failed to get tree from commit: %w", err)
-	}
+	timeoutCh := make(chan struct{})
+	ch := make(chan commitResponse)
 
-	if context == "." {
-		return tree.Hash.String(), nil
-	}
+	timeoutErr := errors.New("timeout exceeded")
+	timeout := 3 * time.Second
```
andreafalzetti, line 138: What's the reason behind the `3 seconds`? In `isClean` we wait for only `1 second`. What makes this different?

pkg/repository/git.go
```diff
@@ -122,53 +120,60 @@ func (r gitRepoController) getSHA() (string, error) {
...
+		case ch <- commitResponse{
+			commit: commit,
+			err:    err,
+		}:
+		}
+	}()
+
+	s := <-ch
+
+	if s.err == timeoutErr {
```
andreafalzetti, line 163:
```suggestion
	if errors.Is(s.err, timeoutErr) {
```

pkg/repository/git.go
```diff
@@ -122,53 +120,60 @@ func (r gitRepoController) getSHA() (string, error) {
...
-	if err != nil {
-		return "", fmt.Errorf("failed to get tree from commit: %w", err)
-	}
+	timeoutCh := make(chan struct{})
+	ch := make(chan commitResponse)
 
-	if context == "." {
-		return tree.Hash.String(), nil
-	}
+	timeoutErr := errors.New("timeout exceeded")
```
andreafalzetti, line 137: considering we have `timeoutErr` in two funcs in this file, should we move it at the top level of thisfile?

andreafalzetti (approved):

cmd/build/v2/hasher.go
```diff
@@ -105,15 +102,13 @@ func (sh serviceHasher) hash(buildInfo *model.BuildInfo, commitType, commitHash
...
+	fmt.Fprintf(&b, "commit:%s;", commitHash)
 	fmt.Fprintf(&b, "target:%s;", buildInfo.Target)
 	fmt.Fprintf(&b, "build_args:%s;", argsText)
 	fmt.Fprintf(&b, "secrets:%s;", secretsText)
 	fmt.Fprintf(&b, "context:%s;", buildInfo.Context)
 	fmt.Fprintf(&b, "dockerfile:%s;", buildInfo.Dockerfile)
-	if commitType == buildContextCommitType {
-		fmt.Fprintf(&b, "dockerfile_content:%s;", getDockerfileContent(buildInfo.Dockerfile))
-	}
+	fmt.Fprintf(&b, "dockerfile_content:%s;", getDockerfileContent(buildInfo.Dockerfile))
```
andreafalzetti, line 111: while validating with `okteto/movies` if I run `okteto build -l=debug`, I've noticed that this tries to always open `Dockerfile` which retults in:

```
INFO[0001] error trying to read Dockerfile: %wopen Dockerfile: no such file or directory
```

Instead, if should try reading the content of the `worker/Dockerfile`, etc..

Can you have a look as well?

pkg/repository/git.go
jLopezbarb: Agree. Changing it to 1 second

pkg/repository/git.go
jLopezbarb: We need to get the latest commit for a folder. The dependency that we are using doesn't have a way to do it but has a wrapper for the `git log` command which has the commit of the folder. That's why I used this. 

If you mean when do we have to have this function is to make testable code

cmd/build/v2/hasher.go
jLopezbarb: Added on https://github.com/okteto/okteto/pull/4083/commits/8b112bbf068c337d9e55782a7996f6d873ddd5a5

pkg/repository/git.go
```diff
@@ -125,53 +124,59 @@ func (r gitRepoController) getSHA() (string, error) {
...
+type commitResponse struct {
+	err    error
+	commit string
+}
 
-	ref, err := repo.Head()
-	if err != nil {
-		return "", fmt.Errorf("failed to get HEAD from repo: %w", err)
-	}
+func (r gitRepoController) GeLatestDirCommit(contextDir string) (string, error) {
```
AdrianPedriza, line 132:
```suggestion
func (r gitRepoController) GetLatestDirCommit(contextDir string) (string, error) {
```

PR #4089, opened 2023-12-15
feat: use build context for isClean

jLopezbarb:
# Proposed changes
Fixes LAKE-124
## Calculate git diff on every build:
 - Calculates the changes vs the latest commit and adds this information to the hash generated
 - Calculates the untracked files and get the content of them. git diff doesn't calculate untracked files
## Move smart build to its own package:
- Moved all the code related to the smart build to its own package
## How to validate

1. Run `gh repo clone okteto/movies`
1. Run `cd movies`
1. Run `export OKTETO_SMART_BUILDS_USING_BUILD_CONTEXT=true`
1. Run `okteto build` twice 
1. Check that the commit that has been built is the latest commit of the dir.
1. Create a new file on one of the dirs and run `okteto build` again and check that it's being built again
1. Run okteto built again and check that it's not built
1. Modify a file and check that it's being built again
1. Check that building again uses the cache

## CLI Quality Reminders 🔧


cmd/build/v2/build.go
```diff
@@ -207,38 +207,43 @@ func (bc *OktetoBuilder) Build(ctx context.Context, options *types.BuildOptions)
...
 
 				meta.CacheHit = isBuilt
 				meta.CacheHitDuration = time.Since(cacheHitDurationStart)
 
 				if isBuilt {
-					bc.ioCtrl.Out().Infof("Skipping build of '%s' image because it's already built for commit %s", svcToBuild, bc.hasher.GetCommitHash(buildSvcInfo))
-					// if the built image belongs to global registry we clone it to the dev registry
-					// so that in can be used in dev containers (i.e. okteto up)
-					if bc.Registry.IsGlobalRegistry(imageWithDigest) {
-						bc.ioCtrl.Logger().Debugf("Copying image '%s' from global to personal registry", svcToBuild)
```
AdrianPedriza, line 233: why we have removed this condition?

cmd/build/v2/build.go
```diff
@@ -320,8 +325,15 @@ func (bc *OktetoBuilder) buildSvcFromDockerfile(ctx context.Context, manifest *m
 	bc.ioCtrl.Logger().Info(fmt.Sprintf("Building service '%s' from Dockerfile", svcName))
 	isStackManifest := manifest.Type == model.StackType
 	buildSvcInfo := bc.getBuildInfoWithoutVolumeMounts(manifest.Build[svcName], isStackManifest)
-	buildHash := bc.hasher.hashService(buildSvcInfo)
-	tagToBuild := newImageTagger(bc.Config).getServiceImageReference(manifest.Name, svcName, buildSvcInfo, buildHash)
+	var err error
+	var buildHash string
```
AdrianPedriza, line 329: if smart build is not enabled `buildHash` is empty. It is expected?

cmd/build/v2/build.go
```diff
@@ -354,9 +366,17 @@ func (bc *OktetoBuilder) addVolumeMounts(ctx context.Context, manifest *model.Ma
 
 	buildInfoCopy := manifest.Build[svcName].Copy()
 	buildInfoCopy.Image = ""
-	buildHash := bc.hasher.hashService(buildInfoCopy)
 
-	tagToBuild := newImageWithVolumesTagger(bc.Config).getServiceImageReference(manifest.Name, svcName, buildInfoCopy, buildHash)
+	var err error
+	var buildHash string
```
AdrianPedriza, line 371: same, when smart build is not enabled this value is empty. Should we use another approach to get the `buildhash` ?

cmd/build/v2/services.go
```diff
@@ -94,8 +94,17 @@ func (bc *OktetoBuilder) checkServiceToBuild(service string, manifest *model.Man
...
+
+	var err error
+	var buildHash string
+	if bc.smartBuildCtrl.IsEnabled() {
+		buildHash, err = bc.smartBuildCtrl.GetBuildHash(buildInfo)
+		if err != nil {
+			bc.ioCtrl.Logger().Infof("error getting build hash: %s", err)
+		}
+	}
+
```
AdrianPedriza, line 106: this block is repeated. Should we encapsulate it in a func?

cmd/build/v2/smartbuild/hasher.go
```diff
@@ -11,58 +11,67 @@
...
 	}
 }
 
 // hashProjectCommit returns the hash of the repository's commit
-func (sh *serviceHasher) hashProjectCommit(buildInfo *model.BuildInfo) string {
-	if sh.projectCommit == "" {
+func (sh *serviceHasher) hashProjectCommit(buildInfo *model.BuildInfo) (string, error) {
+	sh.lock.Lock()
+	projectCommit := sh.projectCommit
+	sh.lock.Unlock()
```
AdrianPedriza, line 59: why we need a lick here? when could be used at the same time?

pkg/repository/local_git.go
```diff
@@ -189,3 +194,29 @@ func (lg *LocalGit) GetLatestCommit(ctx context.Context, gitPath, dirPath string
 	}
 	return string(output), nil
 }
+
+// GetLatestCommit returns the latest commit of the repository at the given path
+func (lg *LocalGit) Diff(ctx context.Context, gitPath, dirPath string, fixAttempt int) (string, error) {
```
AdrianPedriza, line 199: can we add tests for this func?

cmd/build/v2/smartbuild/hasher.go
jLopezbarb: We need it when we are checking if the `GetServicesToBuild`. In order to get the services that are already built. We check that the services are built in parallel. If we don't have the lock there is a race condition and we would be computing the project commit at least two times

cmd/build/v2/build.go
jLopezbarb: We moved into the smart build package. We agreed that everything related to the smart build should be in its own package

cmd/build/v2/build.go
jLopezbarb: Yes, We were already checking checking in the tagger if the hash is empty to create it with the hash or using the okteto tag. Am I missing something?

cmd/build/v2/services.go
jLopezbarb: I think it's ok to have it explicit in order to now that is calling the smartbuild. I can move it into a new function but I think it's more readable like this

pkg/repository/local_git.go
jLopezbarb: Yes, this PR needs to revisit all the test. I wanted to have all the merge conflicts before start creating the rest of the tests

AdrianPedriza:
if I run `okteto build` with dirty changes en `api` then the image is build, ok. But running again it hits cache having the service dirty? it is because the diff hash is the same? before with dirty changes we were building image alwahs. Also shown a message `Skipping build of 'api' image because it's already built for commit 9228bde26a07b6b4410017ab9c9c0b1fc5a20048` is confusing because the service is dirty so IDK if use the git commit is good in this case

cmd/build/v2/smartbuild/smartbuild.go
```diff
@@ -0,0 +1,136 @@
...
+// GetServiceHash returns the hash of the service
+func (s *SmartBuildCtrl) GetServiceHash(buildInfo *model.BuildInfo) (string, error) {
+	s.ioCtrl.Logger().Debugf("getting service hash")
+	return s.hasher.hashBuildContext(buildInfo)
+}
+
+// GetBuildHash returns the hash of the build based on the env vars
+func (s *SmartBuildCtrl) GetBuildHash(buildInfo *model.BuildInfo) (string, error) {
+	s.ioCtrl.Logger().Debugf("getting hash based on the buildContext env var")
+	if s.isUsingBuildContext {
```
AdrianPedriza, line 100: it may be worthwhile to log the method used according to the env variable

cmd/build/v2/build.go
AdrianPedriza: I find it a bit confusing. I think the best thing to do would be at the time of checking if it is enabled or not to put the appropriate tag if possible. This way it is difficult to know at this point that the tagger has logic to control it.

cmd/build/v2/smartbuild/hasher.go
```diff
@@ -21,50 +21,60 @@ import (
...
+	// lock is a mutex to provide thread safety
+	lock sync.RWMutex
 }
 
-func newServiceHasher(gitRepoCtrl repositoryCommitRetriever) *serviceHasher {
+func newServiceHasher(gitRepoCtrl repositoryCommitRetriever, fs afero.Fs) *serviceHasher {
 	return &serviceHasher{
 		gitRepoCtrl:       gitRepoCtrl,
 		buildContextCache: map[string]string{},
+		diffHashCache:     map[string]string{},
```
AdrianPedriza, line 53: why we need this mem storage? I dont see use of it

cmd/build/v2/smartbuild/hasher.go
```diff
@@ -110,36 +117,22 @@ func (sh serviceHasher) hash(buildInfo *model.BuildInfo, commitHash string) stri
 	fmt.Fprintf(&b, "secrets:%s;", secretsText)
 	fmt.Fprintf(&b, "context:%s;", buildInfo.Context)
 	fmt.Fprintf(&b, "dockerfile:%s;", buildInfo.Dockerfile)
-	fmt.Fprintf(&b, "dockerfile_content:%s;", getDockerfileContent(buildInfo.Context, buildInfo.Dockerfile))
+	fmt.Fprintf(&b, "dockerfile_content:%s;", sh.getDockerfileContent(buildInfo.Context, buildInfo.Dockerfile))
+	fmt.Fprintf(&b, "diff:%s;", diff)
```
AdrianPedriza, line 121: why include always `diff` ?. I wouldn't include what is not needed in the hash. For project commit hash is always emtpy.

cmd/build/v2/smartbuild/hasher.go
```diff
@@ -149,3 +142,19 @@ func getDockerfileContent(dockerfileContext, dockerfilePath string) string {
 	encodedFile := sha256.Sum256(content)
 	return hex.EncodeToString(encodedFile[:])
 }
+
+func (sh *serviceHasher) getBuildContextHashInCache(buildContext string) string {
+	sh.lock.RLock()
```
AdrianPedriza, line 147: do we need a lock to read values?

cmd/build/v2/smartbuild/smartbuild.go
```diff
@@ -0,0 +1,136 @@
...
+	"github.com/okteto/okteto/pkg/log/io"
+	"github.com/okteto/okteto/pkg/model"
+	"github.com/spf13/afero"
+)
+
+const (
+	// OktetoEnableSmartBuildEnvVar represents whether the feature flag to enable smart builds is enabled or not
+	OktetoEnableSmartBuildEnvVar = "OKTETO_SMART_BUILDS_ENABLED"
+
+	// OktetoSmartBuildUsingContextEnvVar is the env var to enable smart builds using the build context instead of the project build
```
AdrianPedriza, line 29:
```suggestion
	// OktetoSmartBuildUsingContextEnvVar is the env var to enable smart builds using the build context instead of the project commit
```

pkg/repository/git.go
```diff
@@ -217,6 +323,28 @@ func (ogr oktetoGitRepository) Log(o *git.LogOptions) (object.CommitIter, error)
 	return ogr.repo.Log(o)
 }
 
+func (ogr oktetoGitRepository) calculateUntrackedFiles(ctx context.Context, contextDir string) ([]string, error) {
+	worktree, err := ogr.Worktree()
+	if err != nil {
+		return []string{}, fmt.Errorf("failed to infer the git repo's current branch: %w", err)
```
AdrianPedriza, line 329: is this log good? maybe something about `worktree` would be clearer

pkg/repository/local_git.go
```diff
@@ -77,11 +77,12 @@ func (*LocalExec) LookPath(file string) (string, error) {
 }
 
 type LocalGitInterface interface {
-	Status(ctx context.Context, dirPath string, fixAttempt int) (git.Status, error)
+	Status(ctx context.Context, reporoot, dirPath string, fixAttempt int) (git.Status, error)
```
AdrianPedriza, line 80:
```suggestion
	Status(ctx context.Context, repoRoot, dirPath string, fixAttempt int) (git.Status, error)
```

AdrianPedriza (approved):
I have tested several use cases and it works fine. I have a few comments but none blocking. Good work! 👍

cmd/build/v2/build.go
jLopezbarb: But I think that if we are not calculating the buildHash it should be empty. Otherwise the tagger should have logic for the smart build or viceversa

andreafalzetti (approved):

andreafalzetti:
@jLopezbarb Windows E2E tests are failing, can you have a look?

cmd/build/v2/smartbuild/hasher.go
jLopezbarb: It's empty but it simplifies the code by not needing to have two different strategies. It also doesn't take time to compute it.

cmd/build/v2/smartbuild/hasher.go
jLopezbarb: Yep, We are checking if the images are already built and given that several images access to the same map the lock is needed

PR #4090, opened 2023-12-18
Fix panic when parsing empty service sections (dev & build)

andreafalzetti:
# Proposed changes

Fixes LAKE-102, LAKE-131

This change fixes 2 panics occurring when the manifest has edge cases such as empty dev or build service sections.

Example:

```yaml
dev:
  frontend:
  backend:
    image: ...
```

or

```yaml
buikd:
  frontend:
  backend:
    context: .
```

In these two cases, the CLI would fail to execute any command because the manifest is not valid despite being valid the YAML. This is happening because we either try to access properties that are not there or because we initialize a service when the only info we have is the name.

The solution for `build` was to return a proper error and for `dev` handle gracefully the scenario which leads to a working `okteto up` session.

## How to validate

1. Clone the test repo `git clone git@github.com:andreafalzetti/movies.git`
1. Switch branch `git checkout broken-sections`
1. Verify that latest stable Okteto CLI has 2 panics with each of the broken manifests (there are 2).
1. Run `okteto build -f okteto-broken-build.yml`
1. Run `okteto build -f okteto-broken-dev.yml`
1. Now run the Okteto CLI built using this branch and verify that the panics are no longer occurring.
1. Notice how a meaninful error is returned instead 

## CLI Quality Reminders 🔧

### DIFF
--- a/cmd/build/command.go
+++ b/cmd/build/command.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"github.com/okteto/okteto/pkg/suggest"
 	"os"
 
 	"github.com/moby/buildkit/frontend/dockerfile/instructions"
@@ -152,6 +153,11 @@ func (bc *Command) getBuilder(options *types.BuildOptions, okCtx *okteto.OktetoC
 	manifest, err := bc.GetManifest(options.File)
 	if err != nil {
 		if options.File != "" && errors.Is(err, oktetoErrors.ErrInvalidManifest) && validateDockerfile(options.File) != nil {
+			return nil, err
+		}
+
+		var userFriendlyErr *suggest.UserFriendlyError
+		if errors.As(err, &userFriendlyErr) {
 			return nil, err
 		}
 
--- a/cmd/build/command.go
+++ b/cmd/build/command.go
@@ -18,7 +18,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"github.com/okteto/okteto/pkg/suggest"
 	"os"
 
 	"github.com/moby/buildkit/frontend/dockerfile/instructions"
@@ -36,6 +35,7 @@ import (
 	"github.com/okteto/okteto/pkg/model"
 	"github.com/okteto/okteto/pkg/okteto"
 	"github.com/okteto/okteto/pkg/registry"
+	"github.com/okteto/okteto/pkg/suggest"
 	"github.com/okteto/okteto/pkg/types"
 	"github.com/spf13/cobra"
 )
--- a/cmd/build/command.go
+++ b/cmd/build/command.go
@@ -35,7 +35,6 @@ import (
 	"github.com/okteto/okteto/pkg/model"
 	"github.com/okteto/okteto/pkg/okteto"
 	"github.com/okteto/okteto/pkg/registry"
-	"github.com/okteto/okteto/pkg/suggest"
 	"github.com/okteto/okteto/pkg/types"
 	"github.com/spf13/cobra"
 )
@@ -153,11 +152,6 @@ func (bc *Command) getBuilder(options *types.BuildOptions, okCtx *okteto.OktetoC
 	manifest, err := bc.GetManifest(options.File)
 	if err != nil {
 		if options.File != "" && errors.Is(err, oktetoErrors.ErrInvalidManifest) && validateDockerfile(options.File) != nil {
-			return nil, err
-		}
-
-		var userFriendlyErr *suggest.UserFriendlyError
-		if errors.As(err, &userFriendlyErr) {
 			return nil, err
 		}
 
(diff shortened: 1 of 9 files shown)


pkg/model/manifest.go
```diff
@@ -663,19 +663,21 @@ func Read(bytes []byte) (*Manifest, error) {
...
 	for _, dev := range manifest.Dev {
-		if (dev.Image.Context != "" || dev.Image.Dockerfile != "") && !hasShownWarning {
+		if dev.Image != nil && (dev.Image.Context != "" || dev.Image.Dockerfile != "") && !hasShownWarning {
 			hasShownWarning = true
 			oktetoLog.Yellow(`The 'image' extended syntax is deprecated and will be removed in a future version. Define the images you want to build in the 'build' section of your manifest. More info at https://www.okteto.com/docs/reference/manifest/#build"`)
 		}
+
 	}
 
-	if err := manifest.setDefaults(); err != nil {
```
andreafalzetti, line 672: we are setting default values before validating, this is not very useful because if the manifest is invalid, we should fail early and make the `setDefaults` code safer, because right now it accesses properties that might not be there, which are caught by the `validate`.

cmd/build/command.go
```diff
@@ -155,6 +156,11 @@ func (bc *Command) getBuilder(options *types.BuildOptions, okCtx *okteto.OktetoC
 			return nil, err
 		}
 
+		var userFriendlyErr *suggest.UserFriendlyError
```
andreafalzetti, line 159: I've added this check to avoid proceeding with a fallback to v1. Without this change we would return a misleading error: `Invalid Dockerfile: 'Dockerfile' is not a regular file`. Now we return `manifest validation failed: service 'frontend' build section not defined correctly`.

cmd/build/command_test.go
```diff
@@ -347,6 +348,19 @@ func TestBuilderIsProperlyGenerated(t *testing.T) {
 			expectedError:     false,
 			isBuildV2Expected: false,
 		},
+		{
+			name: "UserFriendlyError returned without fallback.",
```
andreafalzetti, line 352: Added test case for covering the [new logic above](https://github.com/okteto/okteto/pull/4090/files#r1429790892)

pkg/model/manifest.go
```diff
@@ -663,19 +663,21 @@ func Read(bytes []byte) (*Manifest, error) {
 
 	hasShownWarning := false
 	for _, dev := range manifest.Dev {
-		if (dev.Image.Context != "" || dev.Image.Dockerfile != "") && !hasShownWarning {
+		if dev.Image != nil && (dev.Image.Context != "" || dev.Image.Dockerfile != "") && !hasShownWarning {
```
andreafalzetti, line 666: `dev.Image` was `nil` in case of 

```yaml
dev:
  frontend:
  backend:
```

so this extra check makes it safer and removes the panic generated from this line when accessing `dev.Image.Context`.

pkg/model/manifest.go
```diff
@@ -689,6 +691,11 @@ func (m *Manifest) validate() error {
 }
 
 func (b *ManifestBuild) validate() error {
+	for k, v := range *b {
```
andreafalzetti, line 694: as part of the validation of the build section, we now check against `nil` services. I prefer this approach as it fails early the execution of the CLI. Another approach I've considered is to add a lot of checks against `nil` in the functions that iterate over the build section, but it was making the code messy and it's not as safe as this solution, because one case can always slip through and cause a panic.

pkg/model/manifest_test.go
```diff
@@ -634,7 +645,7 @@ func TestInferFromStack(t *testing.T) {
 								},
 							},
 						},
-						Mode: "sync",
+						Mode: constants.OktetoSyncModeFieldValue,
```
andreafalzetti, line 648: just replacing this hardcoded string with its equivalent const

pkg/model/manifest_test.go
```diff
@@ -1382,3 +1393,194 @@ func TestSecretValidate(t *testing.T) {
 		})
 	}
 }
+
+func TestRead(t *testing.T) {
```
andreafalzetti, line 1397: The func `Read` was not directly tested as far I could see. Becuase I've touched it (swapping order between `validate` and `setDefaults`), I've added some testing to it.

pkg/model/manifest.go
AdrianPedriza: I think some validation depends on the values configured in the `setDefaults` func for example `divert.driver`

andreafalzetti: @AdrianPedriza I had a look but I can't find a logic that could be an issue with this change. If you have any more specific suggestion, I am more than happy to look at it 🙏

jLopezbarb (approved):

pkg/model/manifest.go
AdrianPedriza: [here](https://github.com/okteto/okteto/blob/master/pkg/model/manifest.go#L789) we have a switch that check divert driver and that values can be comnfigured in `setDefault` [here](https://github.com/okteto/okteto/blob/master/pkg/model/manifest.go#L836) I think

andreafalzetti: Good catch @AdrianPedriza 👏 

I've reverted the order and added an additional `if` to skip when build is nil. Check it out in this commit: ead42ab9

AdrianPedriza (approved):

PR #4091, opened 2023-12-18
improve err response for expired tokens

AdrianPedriza:
# Proposed changes
- handle responses with `non-200 OK status code: 401 Unauthorized body: \"not-authorized: token is expired\\n\"` error message, meaning token is expired
- if token is provided by user and is expired, suggest a new token generation

## How to validate

1. using a cluster that return the reason when token is expired. 
2. run `okteto ctx use <cluster-url> --token <expired-token>`
3. check output is:
Captura de pantalla 2023-12-18 a las 11 15 42

## CLI Quality Reminders 🔧

### DIFF
--- a/cmd/context/create.go
+++ b/cmd/context/create.go
@@ -173,6 +173,7 @@ func (c *ContextCommand) UseContext(ctx context.Context, ctxOptions *ContextOpti
 	} else if ctxOptions.Token == "" {
 		// this is to avoid login with the browser again if we already have a valid token
 		ctxOptions.Token = okCtx.Token
+		ctxOptions.InferredToken = true
 		if ctxOptions.Builder == "" && okCtx.Builder != "" {
 			ctxOptions.Builder = okCtx.Builder
 		}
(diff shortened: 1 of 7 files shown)


andreafalzetti:
@AdrianPedriza can you fix the description of the PR? (remove please include summary etc) and if you could clarify the steps for configuring this: `using a cluster that return the reason when token is expired`?

cmd/context/options.go
```diff
@@ -86,6 +87,7 @@ func (o *ContextOptions) InitFromEnvVars() {
 			usedEnvVars = append(usedEnvVars, model.OktetoTokenEnvVar)
 		}
 		o.Token = envToken
+		o.InferredToken = true
```
AdrianPedriza, line 90: it is considered that having an expired token as an OKTETO_TOKEN variable must also be notified.

pkg/okteto/client.go
```diff
@@ -343,6 +343,8 @@ func translateAPIErr(err error) error {
 		return fmt.Errorf("server temporarily unavailable, please try again")
 	case "non-200 OK status code: 401 Unauthorized body: \"\"":
 		return fmt.Errorf("unauthorized. Please run 'okteto context url' and try again")
+	case "non-200 OK status code: 401 Unauthorized body: \"not-authorized: token is expired\\n\"":
```
jLopezbarb, line 346: Is it ok to have /n at the end of the error? checking the other graphql errors it doesn't follow the same as the other. Is it on purpose?

cmd/context/options.go
```diff
@@ -36,6 +36,7 @@ type ContextOptions struct {
 	IsOkteto              bool
 	raiseNotCtxError      bool
 	InsecureSkipTlsVerify bool
+	InferredToken         bool
```
jLopezbarb, line 39: Even though I don't think this is the best way to do it, I think this is the only way to do it with the actual code so I'm approving it but leaving this comment for future re-reviews

jLopezbarb (approved):

pkg/okteto/client.go
AdrianPedriza: when the answer comes with a body, the content has a body. this is how the error comes in the cli

cmd/context/options.go
AdrianPedriza: a refactoring would be necessary to avoid the approach followed. Currently, the context options are overwritten and it is difficult to know if the value is generated or given by the user

cmd/context/create.go
```diff
@@ -402,6 +411,11 @@ func (c ContextCommand) getUserContext(ctx context.Context, ctxName, ns, token s
 		userContext, err := client.User().GetContext(ctx, ns)
 
 		if err != nil {
+
```
andreafalzetti, line 414:
```suggestion
```

andreafalzetti (approved):

cmd/context/create.go
AdrianPedriza: 679abc9d40d055dace8663fc5247bab8207db2ef

PR #4098, opened 2023-12-21
refactor: move build model to pkg/build

AdrianPedriza:
## Motivation
- model pkg is a very large dependency which can be splited according to the command. Nowadays, the commands when they make use of the model import types and functions that are not required and with the possible problems that this can imply. With this PR the goal is to isolate the necessary model in the build command and reduce the dependency that this command currently imports. 

## Proposed changes
- move serialization of build section model into its own pkg `pkg/build`
- moves serialization tests
- creation of new pkg `pkg/utils` to avoid import cycles

## How to validate it
- running e2e test should be enough to know if the refactor is suitable
- all serialization logic needed only for `build` command is isolated in `pkg/build`


pkg/build/model.go
```diff
@@ -0,0 +1,419 @@
...
+	return nil
+}
+
+// MarshalYAML Implements the marshaler interface of the yaml pkg.
+func (v VolumeMounts) MarshalYAML() (interface{}, error) {
+	return v.RemotePath, nil
+}
+
+// UnmarshalYAML Implements the Unmarshaler interface of the yaml pkg.
+func (v *VolumeMounts) UnmarshalYAML(unmarshal func(interface{}) error) error {
```
andreafalzetti, line 267: shall we have `VolumeMounts` and its marshalling in a dedicated `volume.go` to make this file smaller?

pkg/build/model.go
```diff
@@ -0,0 +1,419 @@
...
+	RemotePath string
+}
+
+// BuildArg is an argument used on the build step.
+type BuildArg struct {
+	Name  string
+	Value string
+}
+
+func (v *BuildArg) String() string {
```
andreafalzetti, line 58: suggestion, why don't we have 1 type (struct + serialization) per file? for example instead of having `model.go` we could have `build_arg.go` + `build_arg_test.go` and so on, for `build_info.go` and other structs in this file

pkg/model/utils/utils.go
jLopezbarb, line 1: Why dio we need to move this to a `utils.go` file? Can we just extract the needed functions for the `pkg/build` or create a package that matches better the functions that are used? I don't think we should be creating a new utils package, because it will grow and it's going to become the new model package

pkg/build/model.go
```diff
@@ -0,0 +1,419 @@
...
+	"github.com/okteto/okteto/pkg/cache"
+	"github.com/okteto/okteto/pkg/env"
+	"github.com/okteto/okteto/pkg/filesystem"
+	oktetoLog "github.com/okteto/okteto/pkg/log"
+	"github.com/okteto/okteto/pkg/model/utils"
+	"github.com/spf13/afero"
+)
+
+// BuildInfo represents the build info to generate an image
+type BuildInfo struct {
```
jLopezbarb, line 33: NIT: Maybe we can remove the `Build` prefix (it's already on the prefix)

pkg/build/model.go
jLopezbarb, line 1: Should we split this file in different ones based on the struct? This way it's easier to check all functions related to each function. WDYT?

pkg/build/model.go
```diff
@@ -0,0 +1,419 @@
...
+	DependsOn        BuildDependsOn    `yaml:"depends_on,omitempty"`
+}
+
+type VolumeMounts struct {
+	LocalPath  string
+	RemotePath string
+}
+
+// BuildArg is an argument used on the build step.
+type BuildArg struct {
```
jLopezbarb, line 53:
```suggestion
type Arg struct {
```

pkg/build/model.go
```diff
@@ -0,0 +1,419 @@
...
+
+// ManifestBuild defines all the build section
+type ManifestBuild map[string]*BuildInfo
+
+// GetDockerfilePath returns the path to the Dockerfile
+func (b *BuildInfo) GetDockerfilePath() string {
+	if filepath.IsAbs(b.Dockerfile) {
+		return b.Dockerfile
+	}
+	fs := afero.NewOsFs()
```
jLopezbarb, line 83: Given that this is a refactor. Should we move the fs as parameter or extract it on a new method that already is initialised. This way this function could be tested properly. WDYT?

pkg/build/model.go
```diff
@@ -0,0 +1,419 @@
...
+type BuildArgs []BuildArg
+
+// BuildDependsOn represents the images that needs to be built before
+type BuildDependsOn []string
+
+// BuildSecrets represents the secrets to be injected to the build of the image
+type BuildSecrets map[string]string
+
+// ManifestBuild defines all the build section
+type ManifestBuild map[string]*BuildInfo
```
jLopezbarb, line 76: We are usually referring this as `ManifestBuildSection` Given that we are already on the build package. Should we rename it to `ManifestSection`

pkg/build/model.go
```diff
@@ -0,0 +1,419 @@
...
+// UnmarshalYAML Implements the Unmarshaler interface of the yaml pkg.
+func (buildInfo *BuildInfo) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	var rawString string
+	err := unmarshal(&rawString)
+	if err == nil {
+		buildInfo.Name = rawString
+		return nil
+	}
+
+	var rawBuildInfo buildInfoRaw
```
jLopezbarb, line 193: NIT: Maybe we can do something like what we do [here](https://github.com/okteto/okteto/blob/master/pkg/model/serializer.go#L1257-L1273). In case we add or remove a new field it will be added directly and we won't be failing. WDYT?

pkg/model/utils/utils.go
AdrianPedriza: I've moved utils to its own package because if not we would get an import cycle err. So I dont think that isolate `utils` in its own pkg is a bad idea. WDYT?

pkg/build/model.go
AdrianPedriza: renamed it to `manifesSection`

AdrianPedriza:
@andreafalzetti @jLopezbarb check latest commit, I've addressed all yours suggestions 👍

andreafalzetti (approved):

jLopezbarb:
I think there are test that are still missing like `volume_mounts.go` or the `Info Unmarshal`. Can we add tests for those? Also should we move the utils functions to the package that are being used?

AdrianPedriza:
> I think there are test that are still missing like `volume_mounts.go` or the `Info Unmarshal`. Can we add tests for those? Also should we move the utils functions to the package that are being used?

check latest test added in `pkg/build` pkg and a refactor removing old funcs and moving other to the package using them b838c4d9f3bdb26a230bd174691c293facb8e45f

jLopezbarb (approved):

PR #4099, opened 2023-12-22
fix: do not fail if error retrieving envs from image

AdrianPedriza:
# Proposed changes
- do not fail if the image does not exist
- add tests

## CLI Quality Reminders 🔧

### DIFF
--- a/cmd/up/exec.go
+++ b/cmd/up/exec.go
@@ -303,7 +303,7 @@ func (d *devContainerEnvGetter) getEnvsFromDevContainer(ctx context.Context, spe
 	return envs, nil
 }
 
-func (cmg *configMapGetter) getEnvsFromConfigMap(ctx context.Context, name string, namespace string, client kubernetes.Interface) ([]string, error) {
+func (cmg *configMapGetter) getEnvsFromConfigMap(ctx context.Context, name, namespace string, client kubernetes.Interface) ([]string, error) {
 	var envs []string
 
 	cmName := pipeline.TranslatePipelineName(name)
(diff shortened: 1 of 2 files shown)


cmd/up/exec.go
```diff
@@ -220,9 +220,10 @@ func (eg *envsGetter) getEnvs(ctx context.Context) ([]string, error) {
 		return nil, err
 	}
 
-	imageEnvs, err := eg.imageEnvsGetter.getEnvsFromImage(apps.GetDevContainer(app.PodSpec(), "").Image)
+	svcImage := apps.GetDevContainer(app.PodSpec(), "").Image
+	imageEnvs, err := eg.imageEnvsGetter.getEnvsFromImage(svcImage)
 	if err != nil {
-		return nil, err
+		oktetoLog.Warning("Could not to retrieve environment variables from the image '%s'", svcImage)
```
pchico83, line 226: Can we also log the original error in debug level?

pchico83 (approved):

andreafalzetti (approved):

cmd/up/exec.go
jLopezbarb, line 226: NIT: Can we add a comment why this shouldn't fail. Add an scenario where this is affected in case in the future we need to know why we are not returning an error in this case

jLopezbarb (approved):

cmd/up/exec.go
AdrianPedriza: 06ecfc7d7a4f4651111d0aee999fa24a8ef21c3a

AdrianPedriza: 06ecfc7d7a4f4651111d0aee999fa24a8ef21c3a

PR #4106, opened 2023-12-29
add depot.dev as a valid builder 

AdrianPedriza:
## Proposed changes
- add new `isDepotEnabled()` method to check if depot builder is enabled by checking needed env vars: `OKTETO_DEPOT_TOKEN` and `OKTETO_DEPOT_PROJECT_ID`. This builder will be used always in case both envs are set
- refactor `getSolveOpt` method to reuse code for `buildWithOkteto` and `buildWithDepot` builders.
- add integration test for buildWithDepot fn

## How to validate changes
- check that a build using okteto works as expected
- set `OKTETO_DEPOT_TOKEN` and `OKTETO_DEPOT_PROJECT_ID` envs:
- run `okteto build` and check image is build and push to the target registry.


pkg/cmd/build/build.go
```diff
@@ -76,16 +75,16 @@ func (ob *OktetoBuilder) GetBuilder() string {
...
-		if err := ob.buildWithDocker(ctx, buildOptions); err != nil {
-			return err
-		}
+		return ob.buildWithDocker(ctx, buildOptions)
 	} else {
-		if err := ob.buildWithOkteto(ctx, buildOptions, ioCtrl); err != nil {
-			return err
-		}
+		return ob.buildWithOkteto(ctx, buildOptions, ioCtrl)
 	}
```
ifbyol, line 87: Could we convert this in a switch? I think it would be easier to follow

pkg/cmd/build/buildkit.go
```diff
@@ -47,7 +50,35 @@ const (
 type buildWriter struct{}
 
 // getSolveOpt returns the buildkit solve options
-func getSolveOpt(buildOptions *types.BuildOptions, okctx OktetoContextInterface) (*client.SolveOpt, error) {
+func getSolveOpt(buildOptions *types.BuildOptions, okctx OktetoContextInterface, secretTempFolder string) (*client.SolveOpt, error) {
```
ifbyol, line 53: I think this function is also being used within `buildWithDepot` function, why it is within `buildkit.go` file? I think `build` is a better place as it is a common function for 2 different builders.

pkg/cmd/build/depot.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+
+	dp := &depotBuilder{}
+	dp.build, dp.err = build.NewBuild(ctx, req, token)
+	if dp.err != nil {
+		return nil, dp.err
+	}
+
+	// Acquire a buildkit machine.
+	dp.machine, dp.err = machine.Acquire(ctx, dp.build.ID, dp.build.Token, "amd64")
```
ifbyol, line 67: Shouldn't this have to take into account the `platform` specified as part of the build command instead of hardcoding always `amd64`?

pkg/cmd/build/depot.go
```diff
@@ -0,0 +1,89 @@
...
+		return nil, dp.err
+	}
+
+	return dp, nil
+}
+
+func (db *depotBuilder) release() {
+	db.build.Finish(db.err)
+	// ignore error releasing depot's machine
+	_ = db.machine.Release()
```
ifbyol, line 88: Would it be useful to log this as debug maybe?

ifbyol (commented):
Is there any test we can add?

jLopezbarb:
We need to modify the log message in order to inform the user that is building using depot instead of our instance in case we have the feature enabled

pkg/cmd/build/depot.go
```diff
@@ -0,0 +1,173 @@
...
+		}
+		defer func() {
+			err := db.fs.Remove(buildOptions.File)
+			if err != nil {
+				db.ioCtrl.Logger().Infof("failed to remove file after build process: %s", err)
+			}
+		}()
+	}
+
+	// create a temp folder - this will be remove once the build has finished
```
andreafalzetti, line 135:
```suggestion
	// create a temp folder - this will be removed once the build has finished
```

pkg/cmd/build/depot.go
```diff
@@ -0,0 +1,173 @@
...
+	}()
+
+	opt, err := getSolveOpt(buildOptions, db.okCtx, secretTempFolder, db.fs)
+	if err != nil {
+		return fmt.Errorf("failed to create build solver: %w", err)
+	}
+
+	db.ioCtrl.Logger().Info("Acquiring a depot's machine..")
+
+	defer db.release(build)
```
andreafalzetti, line 154: why do we have `defer` here?

pkg/cmd/build/depot.go
```diff
@@ -0,0 +1,173 @@
...
+	db.ioCtrl.Logger().Infof("building '%s' on depot's machine..", buildOptions.Tag)
+
+	return run(ctx, client, opt, buildOptions, db.okCtx, db.ioCtrl)
+}
+
+// getBuildkitClient returns a buildkit client connected to the depot's machine.
+func (db *depotBuilder) getBuildkitClient(ctx context.Context) (*buildkitClient.Client, error) {
+	// Check buildkitd readiness. When the buildkitd starts, it may take
+	// quite a while to be ready to accept connections when it loads a large boltdb.
```
andreafalzetti, line 163: what is "boltdb"?

pkg/cmd/build/depot.go
```diff
@@ -0,0 +1,173 @@
...
+	"github.com/moby/buildkit/client"
+	buildkitClient "github.com/moby/buildkit/client"
+	"github.com/okteto/okteto/pkg/log/io"
+	"github.com/okteto/okteto/pkg/types"
+	"github.com/spf13/afero"
+)
+
+const (
+	depotTokenEnvVar   = "DEPOT_TOKEN"
+	depotProjectEnvVar = "DEPOT_PROJECT"
```
andreafalzetti, line 33:
```suggestion
	depotProjectIdEnvVar = "DEPOT_PROJECT_ID"
```

I think it's much better if we specify that it must be the project ID, otherwise it can be confusing for the user, whether they need to specify the project Name or ID.

pkg/cmd/build/depot.go
```diff
@@ -0,0 +1,173 @@
...
+				Tags:    []string{buildOptions.Tag},
+				Push:    true,
+				Load:    true,
+			},
+		},
+	}
+
+	build, err := db.newDepotBuild(ctx, req, db.token)
+	if err != nil {
+		return err
```
andreafalzetti, line 106: What do you think if we wrap this error with the mention of `depot`? 

Because now the error for wrong credentials looks like this:

```
 x  Error building service 'frontend': unauthenticated: Invalid token
```

If I don't know that my okteto cli is using Depot, it may look like an invalid okteto token.

```suggestion
		return fmt.Errorf("depot build failed: %w", err)
```

pkg/cmd/build/depot.go
pchico83: The "depot" official variables are `DEPOT_PROJECT_ID` and `DEPOT_TOKEN`.
O would prepend `OKTETO_` to follow other envvars nomenclature, and to make it clear these variables are for Okteto, not for depot configuration:
- `OKTETO_DEPOT_PROJECT_ID`
- `OKTETO_DEPOT_TOKEN`

pkg/cmd/build/depot.go
AdrianPedriza: df88f28bd4282c983f8d649e9b9e8482279d0fc1

pkg/cmd/build/depot.go
AdrianPedriza: I moved it immediately after I had the resources available to release it. ceb4fd98c627c0f91f7c409437d555c9573db5af

pkg/cmd/build/depot.go
AdrianPedriza: 639120c2808fd060b13800d78e82742954bd027c

pkg/cmd/build/depot.go
```diff
@@ -0,0 +1,173 @@
...
+	"github.com/moby/buildkit/client"
+	buildkitClient "github.com/moby/buildkit/client"
+	"github.com/okteto/okteto/pkg/log/io"
+	"github.com/okteto/okteto/pkg/types"
+	"github.com/spf13/afero"
+)
+
+const (
+	depotTokenEnvVar   = "OKTETO_DEPOT_TOKEN"
+	depotProjectEnvVar = "OKTETO_DEPOT_PROJECT"
```
pchico83, line 33:
```suggestion
	depotProjectIdEnvVar = "OKTETO_DEPOT_PROJECT_ID"
```

AdrianPedriza: 14e5da3d69860cfaeb0e0024af98275219cfdda3

pkg/cmd/build/depot.go
AdrianPedriza: according to the depot go sdk documentation, the buildkit daemon can take a while to accept connections as it loads a key value database, this is a boltdb database

andreafalzetti: thanks for clarifying, I've never heard of it before :D

pkg/cmd/build/depot.go
```diff
@@ -0,0 +1,171 @@
...
+				Push:    true,
+				Load:    true,
+			},
+		},
+	}
+
+	build, err := db.newDepotBuild(ctx, req, db.token)
+	if err != nil {
+		return fmt.Errorf("depot build failed: %w", err)
+	}
```
andreafalzetti, line 107:
```suggestion
	}
	
	db.ioCtrl.Logger().Infof("[depot] build: %s", build.BuildURL)
	
```

Inspired to the actual `depot` CLI: 

Screenshot 2024-01-10 at 18 13 38

pkg/cmd/build/depot.go
```diff
@@ -0,0 +1,171 @@
...
+	if buildOptions.Platform == "" {
+		buildOptions.Platform = defaultPlatform
+	}
+
+	db.ioCtrl.Logger().Info("Acquiring a depot's machine..")
```
andreafalzetti, line 112:
```suggestion
	db.ioCtrl.Logger().Infof("[depot] connecting to %s machine", buildOptions.Platform)
```

Also, inspired by the output of Depot's CLI, I think it's valuable to output the platform used, and I would not use the word "acquire" which sounds odd personally (I know it's from their API)

andreafalzetti (approved):

pchico83:
@AdrianPedriza what's the default platform used by depot.dev if none is specified? does it depend on your laptop platform?

AdrianPedriza:
> @AdrianPedriza what's the default platform used by depot.dev if none is specified? does it depend on your laptop platform?

by default depot uses AMD64 as platform if none is given or does not match the expected possible ones https://github.com/depot/depot-go/blob/main/machine/machine.go#L53-L56. No inspection of the laptop platform is performed

pkg/cmd/build/depot.go
AdrianPedriza: 011dbfcb9e32cea3b849917383c8a8f4b4c616d9

PR #4108, opened 2024-01-03
Jlopezbarb/wrap errors

jLopezbarb:
# Proposed changes

Fixes LAKE-174

- Changes all the fmt.Errorf("%s/q/v, error.String())  to wrap them correctly
- Following with [effective go](https://go.dev/doc/effective_go#errors) best practices remove errors first letter capitalisation and remove the trailing dot from them.

## CLI Quality Reminders 🔧

### DIFF
--- a/cmd/build/remote/build.go
+++ b/cmd/build/remote/build.go
@@ -78,7 +78,7 @@ func (bc *OktetoBuilder) Build(ctx context.Context, options *types.BuildOptions)
 	}
 
 	if err := utils.CheckIfDirectory(path); err != nil {
-		return fmt.Errorf("invalid build context: %s", err.Error())
+		return fmt.Errorf("invalid build context: %w", err)
 	}
 	options.Path = path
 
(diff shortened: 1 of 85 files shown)


andreafalzetti:
@jLopezbarb can you check the linter and run the `e2e`? (Also PR title needs updating)

AdrianPedriza (commented):
in general I think it is a good initiative but I find it odd not to see that any error handler has changed. We modified the way the error is formed but no `strings.Contains` check on the error has been modified to `errors.Is`. Have we checked that we would not be failing to catch some errors after this change?

jLopezbarb:
> in general I think it is a good initiative but I find it odd not to see that any error handler has changed. We modified the way the error is formed but no `strings.Contains` check on the error has been modified to `errors.Is`. Have we checked that we would not be failing to catch some errors after this change?

If we use the contains within an error it will fail. The case is that we are already transforming the errors to String by calling `.Error()` so there wouldn't be any problem right now. In case we want to move to a better error assertion we should be checking all the errors in the CLI and not just the ones wrapped. Probably that could be another initiative. WDYT?

AdrianPedriza (approved):

andreafalzetti (approved):

PR #4117, opened 2024-01-10
Backport: fix: missing trailing slash in IsGlobalRegistry

andreafalzetti:
# Proposed changes

Backporting to `2.23` a fix which was done in `master` within a bigger PR (see https://github.com/okteto/okteto/pull/4051/files#diff-cb61d8f79451b9541de4a8cc0811523a68d15452b2f5971c7618ea5b423cf4ecR111)

The issue is that by not having the trailing slash, the global namespace can be confused with the prefix of a dev namespace (if they start with he same chars).

## CLI Quality Reminders 🔧

### DIFF
--- a/pkg/registry/registry.go
+++ b/pkg/registry/registry.go
@@ -107,7 +107,7 @@ func (or OktetoRegistry) IsOktetoRegistry(image string) bool {
 
 func (or OktetoRegistry) IsGlobalRegistry(image string) bool {
 	expandedImage := or.imageCtrl.expandImageRegistries(image)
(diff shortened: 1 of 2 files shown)


jLopezbarb (approved):

ifbyol (commented):
Could we add some unit test for this function?

andreafalzetti:
> Could we add some unit test for this function? Or if it already has it, could we add a test case for the scenario where it is failing?

Thanks for reminding me, I was rushing into the change. Added in ceae5dc130ade59677290e3606c07a9d7d803ba9

ifbyol (approved):

AgustinRamiroDiaz (approved):

PR #4139, opened 2024-01-23
lint(revive): enable and fix rule EXC0014

andreafalzetti:
# Proposed changes

This PR introduces a new linter `revive` with only one rule enabled for now: `EXC0014`.

```
# EXC0014 revive: Annoying issue about not having a comment. The rare codebase has such comments
- comment on exported (.+) should be of the form "(.+)..."
```

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -95,7 +95,6 @@ linters:
 
 issues:
   include:
-    #- EXC0012  # EXC0012 revive: Annoying issue about not having a comment. The rare codebase has such comments
     - EXC0014  # EXC0014 revive: Annoying issue about not having a comment. The rare codebase has such comments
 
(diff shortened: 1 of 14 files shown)


.golangci.yml
```diff
@@ -58,6 +58,11 @@ linters-settings:
       - int
       - make
       - '\.WithTimeout'
+  revive:
+    rules:
+      - name: exported
+        arguments:
+          - disableStutteringCheck
```
andreafalzetti, line 65: keeping `disableStutteringCheck` to focus on functions only in this PR. In future PRs, we can consider removing this line and enable extra checks (for types, vars, etc)

jLopezbarb (approved):

PR #4140, opened 2024-01-23
fix: panic on race condition

jLopezbarb:
# Proposed changes

Fixes a panic that was caused by a race condition when the port forward SSH handshake fails. It was stopping the forward manager [here](https://github.com/okteto/okteto/blob/master/pkg/ssh/manager.go#L162) and we were accessing to a nil reference [here](https://github.com/okteto/okteto/blob/master/pkg/k8s/forward/manager.go#L149)

The new error will indicate to check the logs if the secrets didn't have permissions or a clear error if the SSH port forwarding fails.

Screenshot 2024-01-23 at 18 53 09

The error is caused because the user doesn't have permissions to write on the okteto folder. The user must match with the repo folder 

## Reproduction steps

1. Connect with another user to the terminal that has permissions on your user (`sudo su`)
2. Create a new repository belonging to the user (`gh repo clone okteto/movies`) or copy an existing one (`cp -r /pat/to/movies .`)
3. Return to your user
4. Enter in the new repo that belongs to another user (`cd movies`)
5. Run `okteto up`

## CLI Quality Reminders 🔧

EDIT: Add reproduction steps and an explanation for the error

### DIFF
--- a/pkg/k8s/forward/manager.go
+++ b/pkg/k8s/forward/manager.go
@@ -146,7 +146,9 @@ func (p *PortForwardManager) Start(devPod, namespace string) error {
 			if !errors.Is(err, portforward.ErrLostConnectionToPod) {
 				p.activeDev.closeReady()
 			}
-			p.activeDev.err = err
+			if p.activeDev != nil {
+				p.activeDev.err = err
+			}
 		}
 	}()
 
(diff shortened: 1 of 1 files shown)


pkg/k8s/forward/manager.go
```diff
@@ -146,7 +146,9 @@ func (p *PortForwardManager) Start(devPod, namespace string) error {
 			if !errors.Is(err, portforward.ErrLostConnectionToPod) {
 				p.activeDev.closeReady()
 			}
-			p.activeDev.err = err
+			if p.activeDev != nil {
```
ifbyol, line 149: Should this check also apply to line 147? We are using also `activeDev`. Is activeDev != nil always in that scenario?

jLopezbarb: It shouldn't be needed but adding it in case there is a corner case I'm not contemplating

ifbyol (approved):

andreafalzetti (approved):

PR #4141, opened 2024-01-24
[Linter] [Revive] Refactor exported vars that are prefixed with pkg name

andreafalzetti:
# Proposed changes

Using `revive` without excluding any rule, means that most of the exported variables should not be named with the prefix of the pkg name, this PR will address that


jLopezbarb (approved):

cmd/context/create.go
```diff
@@ -230,11 +230,11 @@ func (c *ContextCommand) UseContext(ctx context.Context, ctxOptions *ContextOpti
 	}
 
 	if created && ctxOptions.IsOkteto {
-		oktetoLog.Success("Context '%s' created", okteto.RemoveSchema(ctxOptions.Context))
+		oktetoLog.Success("GetContext '%s' created", okteto.RemoveSchema(ctxOptions.Context))
```
ifbyol, line 233:
```suggestion
		oktetoLog.Success("Context '%s' created", okteto.RemoveSchema(ctxOptions.Context))
```

cmd/up/up.go
```diff
@@ -1016,7 +1016,7 @@ func terminateProcess(pid int) error {
 }
 
 func printDisplayContext(up *upContext) {
-	oktetoLog.Println(fmt.Sprintf("    %s   %s", oktetoLog.BlueString("Context:"), okteto.RemoveSchema(up.Dev.Context)))
+	oktetoLog.Println(fmt.Sprintf("    %s   %s", oktetoLog.BlueString("GetContext:"), okteto.RemoveSchema(up.Dev.Context)))
```
ifbyol, line 1019:
```suggestion
	oktetoLog.Println(fmt.Sprintf("    %s   %s", oktetoLog.BlueString("Context:"), okteto.RemoveSchema(up.Dev.Context)))
```

pkg/analytics/track.go
```diff
@@ -57,8 +57,8 @@ const (
 	previewDestroyEvent      = "DestroyPreview"
 	execEvent                = "Exec"
 	signupEvent              = "Signup"
-	contextEvent             = "Context"
-	contextUseNamespaceEvent = "Context Use-namespace"
+	contextEvent             = "GetContext"
```
ifbyol, line 60:
```suggestion
	contextEvent             = "Context"
```

pkg/cmd/build/log_filter.go
```diff
@@ -29,17 +29,17 @@ type Rule struct {
 	transformer TransformerFunc
 }
 
-type BuildKitLogsFilter struct {
+type LogsFilter struct {
```
ifbyol, line 32: I have doubts about this one. `BuildKit` is not the same than `build`. I guess this is detecting it because `BuildKit` is in camelcase. If we use `BuildkitLogsFilter` it might pass the validation.

I have doubts because this seems specific to BuildKit and not to build. If we do builds with other product (like Depot), we might need to use other filter. In this case, the prefix could be relevant. WDYT?

pkg/okteto/client.go
```diff
@@ -282,7 +282,7 @@ func NewOktetoClientFromUrl(url string) (*OktetoClient, error) {
 	return newOktetoClientFromGraphqlClient(u, httpClient)
 }
 
-// contextWithOauth2HttpClient returns a context.Context with a value of type oauth2.HTTPClient so oauth2.NewClient() can be bootstrapped with a custom http.Client
+// contextWithOauth2HttpClient returns a context.GetContext with a value of type oauth2.HTTPClient so oauth2.NewClient() can be bootstrapped with a custom http.Client
```
ifbyol, line 285:
```suggestion
// contextWithOauth2HttpClient returns a context.Context with a value of type oauth2.HTTPClient so oauth2.NewClient() can be bootstrapped with a custom http.Client
```

ifbyol (commented):
My comment about the `BuildKitLogFilter` is not blocker at all, just some thoughts I had while reviewing it.

The rest of comments (code suggestions) take a look because I think we are changing user-facing error messages and tracking event names

pkg/analytics/track.go
jLopezbarb: This change scares me a lot . We are changing the event we are sending to mixpannel. Are we ok with that

ifbyol: No, that's why I'm doing the suggestion, because this PR changes it. All my suggestions are to get the original value

andreafalzetti: good catch guys, I was still pending a self-review but thanks for spotting it!

pkg/cmd/build/log_filter.go
andreafalzetti: good call, changed in c7ba65fb

ifbyol (approved):

PR #4153, opened 2024-01-25
fix cves

jLopezbarb:
# Proposed changes

- Updates `github.com/whilp/git-urls`: It was not maintained anymore and had a medium CVE. Replaced with `github.com/chainguard-dev/git-urls`
- Updates `github.com/hashicorp/go-getter` from 1.6.2 to 1.7.3
- Upgrade helm version from 3.13.3 to 3.14.0

## CLI Quality Reminders 🔧


andreafalzetti (approved):

Dockerfile
```diff
@@ -1,7 +1,7 @@
 # syntax = docker/dockerfile:experimental
 
 ARG KUBECTL_VERSION=1.28.4
-ARG HELM_VERSION=3.13.3
+ARG HELM_VERSION=3.14.0
```
ifbyol, line 4: This helm version leaves out of support Kubernetes `1.25` which is actively supported by Okteto: https://helm.sh/docs/topics/version_skew/#supported-version-skew

Isn't there available a patch for `3.13.X`?

jLopezbarb: No, 3.13.3 is the latest 3.13 release. Also I've been checking and they don't continue patching previous helm versions so I've decided to revert the change until we don't support k8s 1.25

PR #4156, opened 2024-01-26
Improve experience for overnight `okteto up` sessions, and syntching timeouts

andreafalzetti:
# Proposed changes

This change aims to improve the retry mechanism currently affecting some users. The problem is summarised as follows:

When leaving the `okteto up` running overnight, the OS will pause the okteto CLI and other processes, causing a disconnection from the pod. When this happens, syncthing remote is not reachable, because the k8s port forwarding is also unavailable, therefore users may experience a full disconnection when starting their working day in the morning because we do not consider transient the syncthing timeout error.

After discussing different approaches, we agreed that, making permanently transient the timeout error would be incorrect, because syntching can fail to start due to configuration issues (ie. permissions), but, if the `okteto up` sequence was successfully completed once, we can exclude that syntching has configuration issues to start again, so this proposed change, is making that specific error to be considered transient depening on the previous successful execution. 

Note: there is a scenario that we should consider in the future, to listen for deployments, and make sure that the dev container has not been updated outside of the `okteto up` sequence. It could be that a user successfully runs `okteto up`, then break syncthing with another command, then we would consider the syncthing's timeout error transient when it's not. I think it's quite an edge case that can be considered as follow-up from this PR.

## How to validate

To test this change, it's quite tricky because we need to make `okteto up` suceed once, then break syncthing and trigger a reconnection, making only the remote ping to fail, and verify that it's considered transient.

The way I tested is by making the ping path dynamic, so that based on time of when I was starting `okteto up`, and simulating a disconnection by killing the pod, when the CLI tries to reconnect, for remote, the ping path is a 404, so after a few retries, it would trigger the timeout.

1. Modify `syncthing.go` to use a variable for the ping path `rest/system/ping`
1. In the `Ping` func, based on time, change this value for remote:
```go
if !local {
	// change the following time based on the execution of your debug
	if time.Now().Hour() >= 16 && time.Now().Minute() >= 19 && time.Now().Minute() <= 24 {
		s.PingPath = "/wrong/path"
		// note: s.PingPath does not exist, and should be added to the Syncthing struct
	}
}
```

## CLI Quality Reminders 🔧

### DIFF
--- a/cmd/up/errors.go
+++ b/cmd/up/errors.go
@@ -17,6 +17,7 @@ import (
 	"strings"
 
 	oktetoErrors "github.com/okteto/okteto/pkg/errors"
+	oktetoLog "github.com/okteto/okteto/pkg/log"
 )
 
 // isTransient is an extension of the oktetoErrors.IsTransient, this variant is used to add transient errors dynamically
@@ -33,6 +34,7 @@ func (up *upContext) isTransient(err error) bool {
 			return true
 		}
 		if !isTransientErr && up.unhandledTransientRetryCount < up.unhandledTransientMaxRetries {
+			oktetoLog.Debugf("handling error as transient because okteto up was successfully running before, but it's now failing with: %v (%d of %d)", err, up.unhandledTransientRetryCount, up.unhandledTransientMaxRetries)
 			up.unhandledTransientRetryCount++
 			return true
 		}
--- a/cmd/up/errors.go
+++ b/cmd/up/errors.go
@@ -1,0 +1,36 @@
+// Copyright 2024 The Okteto Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package up
+
+import (
+	"strings"
+
+	oktetoErrors "github.com/okteto/okteto/pkg/errors"
+)
+
+// isTransient is an extension of the oktetoErrors.IsTransient, this variant is used to add transient errors dynamically
+// based on the state of the upContext, in particular, if the up session was successfully started once before any retry
+func (up *upContext) isTransient(err error) bool {
+	if err == nil {
+		return false
+	}
+
+	if up.success {
+		if strings.Contains(err.Error(), "syncthing local=false didn't respond after") {
+			return true
+		}
+	}
+
+	return oktetoErrors.IsTransient(err)
+}
--- a/cmd/up/errors.go
+++ b/cmd/up/errors.go
@@ -14,8 +14,9 @@
 package up
 
 import (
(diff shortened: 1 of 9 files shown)


cmd/up/errors.go
```diff
@@ -0,0 +1,36 @@
...
+
+// isTransient is an extension of the oktetoErrors.IsTransient, this variant is used to add transient errors dynamically
+// based on the state of the upContext, in particular, if the up session was successfully started once before any retry
+func (up *upContext) isTransient(err error) bool {
+	if err == nil {
+		return false
+	}
+
+	if up.success {
+		if strings.Contains(err.Error(), "syncthing local=false didn't respond after") {
```
pchico83, line 30: there are more errors that might appear here. what about retrying for X times if `up.success` for any kind of error?

cmd/up/up.go
```diff
@@ -253,19 +253,21 @@ func Up(at analyticsTrackerInterface, ioCtrl *io.IOController, k8sLogger *io.K8s
...
+				StartTime:                    time.Now(),
+				Registry:                     registry.NewOktetoRegistry(okteto.Config{}),
+				Options:                      upOptions,
+				Fs:                           afero.NewOsFs(),
+				analyticsTracker:             at,
+				analyticsMeta:                upMeta,
+				K8sClientProvider:            okteto.NewK8sClientProviderWithLogger(k8sLogger),
+				tokenUpdater:                 newTokenUpdaterController(),
+				builder:                      buildv2.NewBuilderFromScratch(at, ioCtrl),
+				unhandledTransientMaxRetries: 100,
```
andreafalzetti, line 269: this `100` is very arbitrary, I think that `100` could be a good value to start testing with because we need a value that is a bit high, in case the retries need to run for hours overnight.

cmd/up/errors.go
andreafalzetti: @pchico83 thanks for the feedback, I've updated the PR

cmd/up/errors.go
```diff
@@ -0,0 +1,42 @@
...
+		return false
+	}
+
+	isTransientErr := oktetoErrors.IsTransient(err)
+
+	if up.success {
+		if strings.Contains(err.Error(), "syncthing local=false didn't respond after") {
+			return true
+		}
+		if !isTransientErr && up.unhandledTransientRetryCount < up.unhandledTransientMaxRetries {
```
jLopezbarb, line 35: Can we add a log here in order to track while debugging?

cmd/up/up.go
jLopezbarb: How are we going to know if it's high or low? do we have any metric that we could add in order to know what's the maximum value?

jLopezbarb (approved):
Minor comments. LGTM

cmd/up/errors.go
andreafalzetti: I was planning to add a log, but then realized that we can infer by looking at the logs, if a non transient error is triggering a new sequence, it means it was treated as transient. But now, explaining this, it makes me realize that a log is just easier :) Will add

andreafalzetti: log added in c488d90b

cc @jLopezbarb feel free to suggestion a different log message

cmd/up/up.go
andreafalzetti: @jLopezbarb with the new debug log added with c488d90b, we should be able to see the total max retries and current count, when looking at the logs. Do you have more ideas?

ifbyol (approved):

PR #4162, opened 2024-01-31
Revert "Improve experience for overnight `okteto up` sessions, and sy…

andreafalzetti:
This reverts commit 6b0313611dba7498ef8da14678a13c3fa8a13f6e.

The E2E tests are failing so I am reverting this commit to keep `master` working and work on the tests on a new branch.

## CLI Quality Reminders 🔧

For both authors and reviewers:

- Scrutinize for potential regressions
- Ensure key automated tests are in place
- Build the CLI and test using the validation steps
- Assess Developer Experience impact (log messages, performances, etc)
- If too broad, consider breaking into smaller PRs
- Adhere to our [code style](https://github.com/okteto/okteto/blob/master/docs/code-style.md) and [code review](https://github.com/okteto/okteto/blob/master/docs/code-review.md) guidelines

PR #4172, opened 2024-02-06
Improve experience for overnight `okteto up` sessions, and syntching timeouts

andreafalzetti:
# Proposed changes

This PR was already approved and merged: https://github.com/okteto/okteto/pull/4156

However it was reverted because E2E tests were failing: https://github.com/okteto/okteto/pull/4162

The tests were failing because the new behavior of `okteto up` is to retry 100 times before exiting and the tests were not ready for it.

Now I've added a new logic that checks if errors are non recoverable, for example by deleting the namespace in which the dev container is deployed.

## CLI Quality Reminders 🔧

### DIFF
--- a/cmd/up/errors.go
+++ b/cmd/up/errors.go
@@ -17,7 +17,6 @@ import (
 	"strings"
 
 	oktetoErrors "github.com/okteto/okteto/pkg/errors"
-	oktetoLog "github.com/okteto/okteto/pkg/log"
 )
 
 // isTransient is an extension of the oktetoErrors.IsTransient, this variant is used to add transient errors dynamically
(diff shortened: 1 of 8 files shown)


cmd/up/errors.go
```diff
@@ -0,0 +1,79 @@
...
+
+// isFatalAfterSuccessError checks if the error is non-recoverable, and it's used to stop the retry process
+func isFatalAfterSuccessError(err error) bool {
+	if err == nil {
+		return false
+	}
+
+	switch {
+
+	case
```
ifbyol, line 71: What is the effect of having a fatal error not considered here? Is it going to have the okteto binary stuck on a loop until it reaches the maximum number of retries?

andreafalzetti: @ifbyol yes, it's an interesting approach that was suggested from Pablo in the initial PR: https://github.com/okteto/okteto/pull/4156#discussion_r1467842204

jLopezbarb:
LGTM, but my only concern is that

jLopezbarb (approved):

andreafalzetti:
> LGTM, but my only concern is if the user wakes up, opens the laptop and sees 10 warning messages (100 retries and every 10 retries we show messages) indicating that he is reconnecting and then the same error message he had before this fix.
> 
> We are improving the experience for those who want to keep their session active but we will be showing warning messages that could be misleading

It's an important point to review, I was trying what happens now with okteto cli `2.25.2` and if I go offline (ie. turn-off the wi-fi), in a few minutes I get quite a few logs like:

```
Connection lost to your development container, reconnecting...
Connection lost to your development container, reconnecting...
Connection lost to your development container, reconnecting...
Connection lost to your development container, reconnecting...
Connection lost to your development container, reconnecting...
Connection lost to your development container, reconnecting...
```

I will try the experience of leaving `okteto up` running with the laptop sleeping overnight and see what happens.

cmd/up/errors.go
andreafalzetti: after futher considerations we have decided to remove the logic that retries on all errors

andreafalzetti:
@jLopezbarb @ifbyol I've updated the PR as discussed offline, removing the retry on-all errors logic to avoid the risk of affecting the developer experience of users while trying to improve it for others

ifbyol (approved):

jLopezbarb (approved):

PR #4174, opened 2024-02-06
fix: context insecure is maintained until deleted the context

jLopezbarb:
# Proposed changes

Fixes DEV-26

If the context was set using the `--insecure-skip-tls-verify` flag other commands like kubeconfig or namespace where not using the property set before and was creating an invalid kubeconfig which causes failures.

## How to validate

1. Run `okteto context URL --insecure-skip-tls-verify`
1. Run cat `$HOME/.okteto/context/config.json` and check that `isInsecure` is set
1. Run `okteto ns`
1. Run cat `$HOME/.okteto/context/config.json` and check that `isInsecure` is set
1. Run `okteto kubeconfig`  
1. Run cat `$HOME/.okteto/context/config.json` and check that `isInsecure` is set

## CLI Quality Reminders 🔧

### DIFF
--- a/cmd/context/create.go
+++ b/cmd/context/create.go
@@ -355,6 +355,7 @@ func getLoggedUserContext(ctx context.Context, c *Command, ctxOptions *Options) 
 	ctxOptions.Token = user.Token
 
 	okteto.GetContext().Token = user.Token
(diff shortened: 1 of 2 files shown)


andreafalzetti (approved):

I agree on using `sync.Once` despite we might be able to fix it with a refactor, but it would not be worth it as of now.

ifbyol (approved):

PR #4175, opened 2024-02-06
fix: hybrid interruption

jLopezbarb:
# Proposed changes

Fixes DEV-21

This is a first approach in order to show some information to the user. We can't reconnect to the tty because we are always getting the ttyou signal which causes that we can't log into the terminal no more. However I added the code in the wait for the command execution completion just in case we could find a way to unblock the tty.

## How to validate

1. Run an hybrid application
1. Delete the pod you are connected
1. See a proper error

## CLI Quality Reminders 🔧

### DIFF
--- a/cmd/up/exec.go
+++ b/cmd/up/exec.go
@@ -59,16 +59,16 @@ type HybridExecCtx struct {
 }
 
 // GetCommandToExec returns the command to exec into the hybrid mode
-func (he *hybridExecutor) GetCommandToExec(cmd []string) (*exec.Cmd, error) {
+func (he *hybridExecutor) GetCommandToExec(ctx context.Context, cmd []string) (*exec.Cmd, error) {
 	var c *exec.Cmd
 	if runtime.GOOS != "windows" {
-		c = exec.Command(cmd[0], cmd[1:]...)
+		c = exec.CommandContext(ctx, cmd[0], cmd[1:]...)
 	} else {
 		binary, err := expandExecutableInCurrentDirectory(cmd[0], he.workdir)
 		if err != nil {
 			return nil, err
 		}
-		c = exec.Command(binary, cmd[1:]...)
+		c = exec.CommandContext(ctx, binary, cmd[1:]...)
 	}
 
 	c.Env = he.envs
@@ -428,7 +428,7 @@ func (up *upContext) RunCommand(ctx context.Context, cmd []string) error {
 				return err
 			}
 
-			cmd, err := executor.GetCommandToExec(cmd)
+			cmd, err := executor.GetCommandToExec(ctx, cmd)
 			if err != nil {
 				return err
 			}
--- a/cmd/up/up.go
+++ b/cmd/up/up.go
@@ -762,6 +762,8 @@ func (up *upContext) activateLoop() {
 
 // waitUntilExitOrInterruptOrApply blocks execution until a stop signal is sent, a disconnect event or an error or the app is modify
 func (up *upContext) waitUntilExitOrInterruptOrApply(ctx context.Context) error {
+	ttySignals := make(chan os.Signal, 1)
+	signal.Notify(ttySignals, syscall.SIGTTIN, syscall.SIGTTOU)
 	for {
 		select {
 		case err := <-up.CommandResult:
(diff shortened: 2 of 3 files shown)


cmd/up/up.go
```diff
@@ -789,7 +791,12 @@ func (up *upContext) waitUntilExitOrInterruptOrApply(ctx context.Context) error
 		case err := <-up.GlobalForwarderStatus:
 			oktetoLog.Infof("exiting by error in global forward checker: %v", err)
 			return err
-
+		case <-ttySignals:
+			oktetoLog.Infof("SIGTTOU/SIGTTIN received, starting shutdown sequence preventing it to background the process")
+			return oktetoErrors.UserError{
+				E:    fmt.Errorf("terminal disconnect detected"),
```
jLopezbarb, line 797: I think I need some help here deciding what we show to the user

cmd/up/exec.go
```diff
@@ -59,16 +59,16 @@ type HybridExecCtx struct {
 }
 
 // GetCommandToExec returns the command to exec into the hybrid mode
-func (he *hybridExecutor) GetCommandToExec(cmd []string) (*exec.Cmd, error) {
+func (he *hybridExecutor) GetCommandToExec(ctx context.Context, cmd []string) (*exec.Cmd, error) {
 	var c *exec.Cmd
 	if runtime.GOOS != "windows" {
-		c = exec.Command(cmd[0], cmd[1:]...)
+		c = exec.CommandContext(ctx, cmd[0], cmd[1:]...)
```
jLopezbarb, line 65: Adding context so when we cancel the command we don't leave a zombie process

cmd/up/up.go
```diff
@@ -963,6 +970,9 @@ func (up *upContext) shutdownHybridMode() {
 		return
 	}
 
+	if err := up.hybridCommand.Cancel(); err != nil {
```
jLopezbarb, line 973: This is equal to running the code below of this, but this one doesn't leave the command zombie. We could remove the other command cleanup code later. WDYT?

andreafalzetti: @jLopezbarb do you mean using `up.hybridCommand.Cancel()` instead of `terminateProcess(up.hybridCommand.Process.Pid)`? I think it makes sense, but I would test it a few times to make sure.

In regards to killing child processes, I think we still need it

cmd/up/up.go
andreafalzetti, line 797:
```suggestion
				E:    fmt.Errorf("connection lost to your development container"),
```

We could reuse the current error message that we display when the connection is lost.

cmd/up/up.go
```diff
@@ -789,7 +791,12 @@ func (up *upContext) waitUntilExitOrInterruptOrApply(ctx context.Context) error
 		case err := <-up.GlobalForwarderStatus:
 			oktetoLog.Infof("exiting by error in global forward checker: %v", err)
 			return err
-
+		case <-ttySignals:
+			oktetoLog.Infof("SIGTTOU/SIGTTIN received, starting shutdown sequence preventing it to background the process")
+			return oktetoErrors.UserError{
+				E:    fmt.Errorf("terminal disconnect detected"),
+				Hint: "Try re-running 'okteto up' to reconnect",
```
andreafalzetti, line 798: Here's my suggestion, when the CLI disconnects we display a warning: `Connection lost to your development container, reconnecting...` in this case we cannot reconnect, so it's an error, but with a similar message.

```suggestion
				E:    fmt.Errorf("connection lost to your development container"),
				Hint: "Use 'okteto up' to reconnect",
```

andreafalzetti:
@jLopezbarb I have tried the changes and work as advertised. My question however is, shouldn't we try to recover the state instead of terminating the session?

If I use hybrid mode with any command like `npm start`, when it reconnects, it's able to restart the session and launch `npm start` again, if I use `bash`, with this change it will stop the `up` session.

Considering that we are intercepting the two signals now, can we just ignore them for example instead of exiting?

jLopezbarb:
> @jLopezbarb I have tried the changes and work as advertised. My question however is, shouldn't we try to recover the state instead of terminating the session?
> 
> If I use hybrid mode with any command like `npm start`, when it reconnects, it's able to restart the session and launch `npm start` again, if I use `bash`, with this change it will stop the `up` session.
> 
> Considering that we are intercepting the two signals now, can we just ignore them for example instead of exiting?

We are capturing the signal, but right now we can't recover the terminal state that was prior to the bash execution. If we ignore it we will be in an infinite loop where we are always capturing the signal and not showing anything to the user. So I guess we will need a second iteration where we try to recover from that state but the first thing is to not send the process into background and show a clear message to the user

PR #4177, opened 2024-02-12
Fix panic in `okteto up` when using remote deploy

andreafalzetti:
# Proposed changes

Fixes: DEV-147

`okteto up` would panic if the user attempted to remote deploy their service. To fix it, it was enough to pass the `ioCtrl` which was missing.

I've added a new integration test to cover this scenario. It's a copy of the existing test based on local deploy but I think it's worth having because they use very different deploy approaches.

Additionally, in the first commit, I've added a lock to prevent an additional panic that was occurring, about concurrent map writes happening in different goroutines.

## How to validate

To reproduce the bug follow these steps:

1. Using okteto `2.25.1`
1. Clone https://github.com/okteto/movies-frontend
1. Modify the `okteto.yml` to use remote deploy:
```
deploy:
  remote: true
  commands:
    - name: Deploy Frontend
    - kubectl apply -f code.yml
```
1. Run `okteto up`
1. Notice the panic
1. Repeat the tests with the CLI built from this branch and notice how the `okteto up` succeeds.

## CLI Quality Reminders 🔧

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -289,6 +289,7 @@ jobs:
           command: |
             $env:OKTETO_PATH="$($HOME)\project\artifacts\bin\okteto.exe"
             $env:Path+=";$($HOME)\project\artifacts\bin"
+            $env:SSH_AUTH_SOCK = (Get-Command ssh-agent).Definition -replace 'ssh-agent.exe','ssh-agent.sock'
             go test github.com/okteto/okteto/integration/up -tags="integration" --count=1 -v -timeout 45m
       - store_artifacts:
           path: C:\Users\circleci\.okteto
--- a/cmd/build/v2/smartbuild/hasher.go
+++ b/cmd/build/v2/smartbuild/hasher.go
@@ -88,7 +88,9 @@ func (sh *serviceHasher) hashBuildContext(buildInfo *build.Info) (string, error)
 			return "", fmt.Errorf("could not get build context diff sha: %w", err)
 		}
 
+		sh.lock.Lock()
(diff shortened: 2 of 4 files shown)


.circleci/config.yml
```diff
@@ -289,6 +289,7 @@ jobs:
           command: |
             $env:OKTETO_PATH="$($HOME)\project\artifacts\bin\okteto.exe"
             $env:Path+=";$($HOME)\project\artifacts\bin"
+            $env:SSH_AUTH_SOCK = (Get-Command ssh-agent).Definition -replace 'ssh-agent.exe','ssh-agent.sock'
```
ifbyol, line 292: Why is needed this now? For the new tests we added?

ifbyol (approved):
Minor comment to get more context about the change in the circleCI config but it's not a blocker

jLopezbarb (approved):

.circleci/config.yml
andreafalzetti: @ifbyol this is needed to make remote deploy work from CI. It's already set at line [281](https://github.com/okteto/okteto/pull/4177/files#diff-78a8a19706dbd2a4425dd72bdab0502ed7a2cef16365ab7030a5a0588927bf47R281) for the `deploy` tests. Now I am adding a test for `okteto up` + remote deploy so we also need it here.

Without this, Windows Integration tests fail in CI with an error related to `SSH_AUTH_SOCK`

ifbyol: Nice! Thanks for the context!

PR #4185, opened 2024-02-19
[Hybrid] Improve process management 

andreafalzetti:
# Proposed changes

Fixes DEV-21

This PR fixes 2 things:

1. Currently on Windows hybrid mode keeps child processes running.
2. Currently, on Linux/Mac the restart cause the CLI to end in the background (only for processes such as `bash`, `sh`, etc).

The new logic is:

1. On Windows, do not use `SIGTERM` because it's not supported, but always use `SIGKILL`.
2. On Unix, attempt to gracefully shut down the process, in case the process does not terminate itself in 3 seconds, Okteto CLI will terminate it with `SIGKILL`.

Notes:
- the `3 seconds` is arbitrary, it's unpredictable how long a process could take to gracefully terminate itself (for example: something like `npm` is very quick but a database might take longer)

## How to validate

On both Mac and Windows:

1. `git clone git@github.com:andreafalzetti/hybrid-test.git`
1. `cd hybridapp && npm install` 
1. `okteto up`
1. Kill the pod: `k delete pod -n <NS> $(kubectl get pod -n <NS> --no-headers -o custom-columns=":metadata.name" | grep hybridapp-okteto-)`
1. Watch the service restart on the same port (`8080`)

Change the `command` in `okteto.yml` to be `bash` and repeat the test above.

To test the timeout logic, I've written and used this program as the `command` of the manifest of the repo example:

Once executed, this program will ignore `SIGTERM` (try doing `kil -SIGTERM <PID>` and see). It can only be terminated with `CTRL+C` or `SIGKILL`

```go
package main

import (
	"os"
	"os/signal"
	"syscall"
)

func main() {
	// Ignore SIGTERM signal
	signal.Ignore(syscall.SIGTERM)

	// Create a channel to receive signals
	sigChan := make(chan os.Signal, 1)
	// Register SIGKILL signal to channel (Note: SIGKILL cannot be caught or ignored)
	signal.Notify(sigChan, syscall.SIGKILL)

	// Block and wait for signals
	<-sigChan
}



```

If you look at the `okteto.log` you should see something like:

```
time="2024-02-19T17:07:10+01:00" level=debug msg="terminating process: 9482" action=ac1bfac3-9e65-433c-a3c4-877c8e107a7f version=437a0962
time="2024-02-19T17:07:13+01:00" level=debug msg="graceful termination timed out, killing process: 9482" action=ac1bfac3-9e65-433c-a3c4-877c8e107a7f version=437a0962
time="2024-02-19T17:07:13+01:00" level=debug msg="process terminated successfully: 9482" action=ac1bfac3-9e65-433c-a3c4-877c8e107a7f version=437a0962
```

## CLI Quality Reminders 🔧

### DIFF
--- a/cmd/up/up.go
+++ b/cmd/up/up.go
@@ -50,6 +50,7 @@ import (
 	"github.com/okteto/okteto/pkg/model"
 	"github.com/okteto/okteto/pkg/okteto"
 	oktetoPath "github.com/okteto/okteto/pkg/path"
+	"github.com/okteto/okteto/pkg/process"
 	"github.com/okteto/okteto/pkg/registry"
 	"github.com/okteto/okteto/pkg/ssh"
 	"github.com/okteto/okteto/pkg/syncthing"
@@ -763,6 +764,8 @@ func (up *upContext) activateLoop() {
 
 // waitUntilExitOrInterruptOrApply blocks execution until a stop signal is sent, a disconnect event or an error or the app is modify
 func (up *upContext) waitUntilExitOrInterruptOrApply(ctx context.Context) error {
+	// only for unix because Windows does not support SIGTTIN and SIGTTOU
+	goToBackground := getSendToBackgroundSignals()
 	for {
 		select {
 		case err := <-up.CommandResult:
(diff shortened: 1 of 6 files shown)


andreafalzetti:
@viddo considering your knowledge and experience using hybrid, would you mind giving this change a try? 🙏

ifbyol (approved):

PR #4186, opened 2024-02-20
Fix `okteto endpoints` namespace detection from `okteto.yml` 

andreafalzetti:
# Proposed changes

`okteto endpoints` ignores the property `namespace` of the `okteto.yml`.

## How to validate

1. Create a new namespace where you have nothing deployed to (from now referred as NS1)
1. Select NS1 as your current default
1. Using any project (ie. https://github.com/okteto/movies-frontend)
1. Modify the `okteto.yml` and add a top-level `namespace: <NS2>` (replace `<NS2>` with a namespace different from NS1)
1. Run `okteto endpoints` and observe how the CLI outputs `Using <NS1> @ <Context> as context`
1. It's worth noticing that the endpoints are printed correctly, it's just the namespace name that is wrong (The CLI uses the correct one, but it displayes the wrong one)

Build the CLI from this PR and repeat the steps:

1. Observe how `okteto endpoints` print the correct namespace
3. Deploy the app and observe the right endpoints printer for each NS

Extra:
Currently if you run `okteto endpoints -f not-exists.yml` the endoints cmd falls back on `okteto.yml`. This behavior is inconsistent with other cmds so I've changed it to return an error (like `okteto deploy`).

### Test Scenarios

- `okteto endpoints`
- `okteto endpoints -n <NS>`
- `okteto endpoints` with `namespace: <NS>`

## CLI Quality Reminders 🔧


cmd/context/utils.go
```diff
@@ -224,7 +224,7 @@ func LoadStackWithContext(ctx context.Context, name, namespace string, stackPath
 }
 
 // LoadContextFromPath initializes the okteto context taking into account command flags and manifest namespace/context fields
-func LoadContextFromPath(ctx context.Context, namespace, k8sContext, path string) error {
+func LoadContextFromPath(ctx context.Context, namespace, k8sContext, path string, defaultCtxOpts *Options) error {
```
ifbyol, line 227: Why do we need a pointer for `defaultCtxOpts`? I see we always pass a value. We could just pass the value

ifbyol (commented):
Testing it I see that with the code in this PR, the context used is not printed in the CLI output.

You can see that when I use `ok` (the binary built with the changes in this branch), the context is not displayed and it directly shows the endpoints

If I use `okteto` (latest stable CLI version), it shows the context used (the wrong one)

Screenshot 2024-02-21 at 09 20 40

cmd/context/utils.go
andreafalzetti: no need to be pointer actually, the code is much better with a value type, changed in e0b7d6b40317993841a2b6226fdf6125a18d56af

thanks!

cmd/deploy/endpoints.go
```diff
@@ -93,32 +94,20 @@ func Endpoints(ctx context.Context, k8sLogger *io.K8sLogger) *cobra.Command {
...
-				Context:   ctxResource.Context,
-				Namespace: ctxResource.Namespace,
-			}
-			if options.Output == "" {
-				ctxOptions.Show = true
-			}
-			if err := contextCMD.NewContextCommand().Run(ctx, ctxOptions); err != nil {
-				return err
+			// Loads, updates and uses the context from path. If not found, it creates and uses a new context
+			if err := contextCMD.LoadContextFromPath(ctx, options.Namespace, options.K8sContext, options.ManifestPath, contextCMD.Options{Show: options.Output == ""}); err != nil {
```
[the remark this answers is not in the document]
andreafalzetti: the `okteto endpoints` command can be used as follows:

```
okteto endpoints --output=json | jq XYZ
```

or

```
okteto endpoints --output=md > README.md
```

In those scenarios we cannot output the plaintext header or it would break those scenarios. For `build` and `deploy` we don't have such use-cases, or at least I am not aware or them, so the `Show` value - only for `endpoints` depend on its `--output` flag.

Worth mentioning that for "plaintext" in most commands of the CLI we use empty string. So by default an empty `--output` flag means "plaintext". So the check is a bit odd. I hope that with 3.0 we can add a third allowed value for `--output plaintext` or `--output text` to make this more explicit.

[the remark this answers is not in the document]
ifbyol: For `deploy` and `destroy` we have the global flag `--log-output` which could be also `json` (I was confusing the flag output with that global flag). I was testing deploy and destroy, and when the `--log-output` flag is set, we are not displaying the context information even if we set `Show: true` in the context.

I see it working fine but I think this is confusing for the user. If I execute `okteto endpoints --log-output=json` it doesn't print anything in the output

ifbyol:
I was testing it and If I specify the namespace through the flag `-n`, we don't show the context information, but prior this change, we were displaying it

Screenshot 2024-02-22 at 09 27 23

andreafalzetti:
> I was testing it and If I specify the namespace through the flag `-n`, we don't show the context information, but prior this change, we were displaying it
> Screenshot 2024-02-22 at 09 27 23

Good catch, I think the scenario you reproduced is because when you specify the `namespace` in both places (manifest + flag), `contextCMD.LoadContextFromPath` fails (due to error `oktetoErrors.ErrNamespaceNotMatching` from the func `UpdateNamespace`), calling the fallback logic `contextCMD.NewContextCommand().Run`, which had `Show: false` hardcoded.

After looking at why it had `Show: false` hardcoded, it seems to me a safeguard to avoid showing the header twice. In fact, in case of error occurring after the display of the header in the first func, the second func might display it again. Considering that the display is done just before a `return nil`, I don't think it's risky to pass a dynamic `Show` in both functions.

However, keeping `false` will never make the header appear, so it's a conflicting logic, so I've pushed a new commit with this change, fixing this

cmd/deploy/endpoints.go
```diff
@@ -93,32 +94,22 @@ func Endpoints(ctx context.Context, k8sLogger *io.K8sLogger) *cobra.Command {
...
-				Context:   ctxResource.Context,
-				Namespace: ctxResource.Namespace,
-			}
-			if options.Output == "" {
-				ctxOptions.Show = true
-			}
-			if err := contextCMD.NewContextCommand().Run(ctx, ctxOptions); err != nil {
-				return err
+			// false for 'json' and 'md' to avoid breaking their syntax
+			showCtxHeader := options.Output == ""
```
ifbyol, line 107: 👏 Thanks for the comment clarifying the case

ifbyol (approved):

PR #4128, opened 2024-01-18
fix: remote with parent folders

jLopezbarb:
# Proposed changes

Fixes LAKE-191

We were using the path that the user was introducing in the --file. We should be using ManifestPath relative to the root project instead. This PR fixes that

## How to validate

1. Run cd api
1. Run okteto deploy -f ../okteto.yml  --remote

## CLI Quality Reminders 🔧

For both authors and reviewers:

- Scrutinize for potential regressions
- Ensure key automated tests are in place
- Build the CLI and test using the validation steps
- Assess Developer Experience impact (log messages, performances, etc)
- If too broad, consider breaking into smaller PRs
- Adhere to our [code style](https://github.com/okteto/okteto/blob/master/docs/code-style.md) and [code review](https://github.com/okteto/okteto/blob/master/docs/code-review.md) guidelines

PR #4189, opened 2024-02-23
Build command first refactor

ifbyol:
This is a first PR on the work to refactor the build part of the CLI. In this PR there are a few changes:
* Removed the builder that was under the `cmd/build/remote`. This was only used for the deploy and destroy remote but the logic in the builder wasn't really be needed for the remote scenario as we are the ones setting all the build options accordingly. In this way, we also stop tracking remote deploy and destroy as builds, as the logic for tracking was in this builder.
* Deploy and destroy remote have a dependency now with a remoteRunner instead of a builder. The idea is to make agnostic the commands about where the commands really run. It is not completely agnostic because the function params use `BuildOptions`, but it is a first step.
* All the logic related to build that was on v1 builder is now extracted to a "basic" builder (we can look for a better name). The idea is that the basic builder has the minimum logic to build an image and then, can be used by more complex builders to extend the functionality
* v1 builder is a wrapper of the basic builder which basically runs the build but also have the `IsV1` function (used in other parts of the code).
* v2 builder is also a wrapper of the basic builder adding all the logic about smart builds and git calculation. In this way, we break the dependency between v1 and v2 builders. They both use a basic builder
* Also, this PR fixes a bug in the remote destroy when the command is executed from a parent folder of the okteto manifest. The deploy was fixed in this [PR](https://github.com/okteto/okteto/pull/4128) but we didn't include that logic in the destroy
* Removed some unused dependencies in the v1 builder (`OktetoBuilderInterface` and `oktetoRegistryInterface`)

I created a couple of simple diagrams to understand the changes. Before the changes in this PR, the dependency/relationship between builders was the following:

Screenshot 2024-02-29 at 14 54 18

After the changes in this PR, the relationships are:

Screenshot 2024-02-29 at 14 54 29


cmd/build/v2/build_test.go
```diff
@@ -158,15 +159,14 @@ func (a *fakeAnalyticsTracker) TrackImageBuild(meta ...*analytics.ImageBuildMeta
 	a.metaPayload = meta
 }
 
-func NewFakeBuilder(builder OktetoBuilderInterface, registry oktetoRegistryInterface, cfg oktetoBuilderConfigInterface, analyticsTracker analyticsTrackerInterface) *OktetoBuilder {
+func NewFakeBuilder(builder buildCmd.OktetoBuilderInterface, registry oktetoRegistryInterface, cfg oktetoBuilderConfigInterface, analyticsTracker analyticsTrackerInterface) *OktetoBuilder {
 	return &OktetoBuilder{
-		Registry:          registry,
-		Builder:           builder,
+		Registry: registry,
+		// Builder:           builder,
```
andreafalzetti, line 165:
```suggestion
```

Should we remvoe this?

andreafalzetti (approved):
Great change 👏

go.mod
```diff
@@ -237,6 +237,7 @@ require (
 	github.com/samber/lo v1.38.1 // indirect
 	github.com/samber/slog-common v0.11.0 // indirect
 	github.com/skeema/knownhosts v1.2.1 // indirect
+	github.com/stretchr/objx v0.5.0 // indirect
```
jLopezbarb, line 240: Why do we need this change? I don't see any use on the code and we didn't update any dependency

cmd/destroy/remote.go
```diff
@@ -117,19 +123,24 @@ type remoteDestroyCommand struct {
 
 func newRemoteDestroyer(manifest *model.Manifest, ioCtrl *io.Controller) *remoteDestroyCommand {
 	fs := afero.NewOsFs()
-	builder := remoteBuild.NewBuilderFromScratch(ioCtrl)
+	runner := &buildCmd.OktetoBuilder{
```
jLopezbarb, line 126: NIT: Could we move this into a init function? I think right now is not a problem but if we ever end up adding more fields to buildCmd.OktetoBuilder, we can forget to add that field on all the initializations. WDYT?

go.mod
ifbyol: I'm not sure, I just run go mod tidy from the vscode and it was added. I will delete it and executed again to see if it pops up again

cmd/build/v2/build_test.go
ifbyol: Yes, good catch!

cmd/destroy/remote.go
ifbyol: I created the init function but that doesn't prevent you from that error. You can always create the OktetoBuilder outside the init function for whatever reason

cmd/build/v2/build_test.go
ifbyol: Removed: https://github.com/okteto/okteto/pull/4189/commits/ffb513fbeb752fe766c03aea2cd3e0090764ac94

andreafalzetti (approved):

jLopezbarb (approved):

PR #4198, opened 2024-03-08
DEV-186: Refactor deploy remote command to execute the minimum in the remote

ifbyol:
# Proposed changes

Fixes DEV-186

Refactor remote deploy command to avoid to execute duplicated logic and reduce the amount of things deployed in remote.

Changes in this PR:
* The Dockerfile generated to execute the remote logic doesn't call to `okteto deploy` anymore. Instead, it calls to a new command `remote-run`
* This new `remote-run` command is the one in charge of deploying: custom commands defined by the users, divert and external resources defined in the Okteto manifest.
* This new `remote-run` command just receives the minimum information needed to execute the remote logic and it doesn't need to read any manifest or similar.
* Docker-compose and endpoints are always deployed locally, even when the deploy command is run with the `remote` flag.
* Proxy runs now only when deploying custom commands, divert or external resources. This implied to do also the following changes:
  * As part of the docker compose translations we do, we now add the deployed-by label in all the resources created
  * As part of the docker compose translations we do, we now add the divert information (if any) to the pod spec of deployments, statefulsets and jobs (resources created by us in the compose translation)
* Created a `noop` divert driver to minimize the checks in the compose logic and inject that one when no divert information is provided in the okteto manifest 
* In the Dockerfile generated in the remote command, the Okteto command is now executed from the CLI added by us from the CLI image. This is to prevent any breaking change if someone is using a custom image with a pinned Okteto version. So, the command is executed with the full path added by us
* A common struct with the logic to deploy commands, divert and external resources was created and used y local and remote runners. So they logic is shared but not coupled.
* Changed the circleci configuration to run the integration tests. Now, we build the CLI image before running them. This is because the integrations tests for the remote command was using an old version of CLI (`latest`) to run the remote logic, instead of running the one in the branch you are modifying. This might slow down the integration tests, but I think it is a good tradeoff until we set a better way for this
* Fixed dependency variables available in the deploy section. There were 2 errors:
  * We were not setting the right name for the variables defined in the dependencies. There were 2 errors related with the naming:
    * We were using the configmap name instead of the dependency name, so we were generating an env var like: `OKTETO_DEPENDENCY_OKTETO-GIT-DATABASE_VARIABLE_USERNAME`. `OKTETO-GIT-` should not be included in the env var name as the pattern for those variables is: `OKTETO_DEPENDENCY_<NAME>_VARIABLE_<VARIABLE_NAME>` and `okteto-git-` is not part of the dependency name
    * On dependencies with `-` we were not converting them to `_`, but variables in bash cannot contain `-`
  * Those variables were not being available in the remote execution of commands. Dependency variables were being set as env vars after they are deployed, but then, they were not being passed to the remote execution, so if any command is using that variable is not working properly. Now, they are passed to the remote in the same way than the build variables generated on build time

In order to simplify the review, I'm attaching some diagrams on how the local and remote execution was working (flow diagram) vs how it is now. The main difference happens in the remote one (pay attention o how duplication of logic is now gone)

Old local deploy:

old-local

New local deploy:

local

Old remote deploy:

old-remote

New remote deploy:

remote

## How to validate

Do an intense testing of deploy in both versions (loca and remote). Testing different scenarios (manifest with commands, external resources, compose, endpoints), just compose files, route to folders (instead of directly to manifest), using vars

## CLI Quality Reminders 🔧

### DIFF
--- a/cmd/deploy/remote.go
+++ b/cmd/deploy/remote.go
@@ -32,6 +32,7 @@ import (
 	buildCmd "github.com/okteto/okteto/pkg/cmd/build"
 	"github.com/okteto/okteto/pkg/config"
 	"github.com/okteto/okteto/pkg/constants"
+	"github.com/okteto/okteto/pkg/deployable"
 	oktetoErrors "github.com/okteto/okteto/pkg/errors"
 	"github.com/okteto/okteto/pkg/filesystem"
 	oktetoLog "github.com/okteto/okteto/pkg/log"
@@ -41,6 +42,7 @@ import (
 	"github.com/okteto/okteto/pkg/remote"
 	"github.com/okteto/okteto/pkg/types"
 	"github.com/spf13/afero"
+	"gopkg.in/yaml.v2"
 )
 
 const (
@@ -62,6 +64,10 @@ ARG {{ .TokenArgName }}
 ARG {{ .ActionNameArgName }}
 ARG {{ .TlsCertBase64ArgName }}
 ARG {{ .InternalServerName }}
+ARG {{ .OktetoDeployable }}
+ARG {{ .GitHubRepositoryArgName }}
+ARG {{ .BuildKitHostArgName }}
+ARG {{ .OktetoRegistryURLArgName }}
 RUN mkdir -p /etc/ssl/certs/
 RUN echo "${{ .TlsCertBase64ArgName }}" | base64 -d > /etc/ssl/certs/okteto.crt
 
@@ -80,7 +86,7 @@ RUN okteto registrytoken install --force --log-output=json
 
 RUN --mount=type=secret,id=known_hosts --mount=id=remote,type=ssh \
   mkdir -p $HOME/.ssh && echo "UserKnownHostsFile=/run/secrets/known_hosts" >> $HOME/.ssh/config && \
-  okteto deploy --log-output=json --server-name="${{ .InternalServerName }}" {{ .DeployFlags }}
+  /okteto/bin/okteto remote-run --log-output=json --server-name="${{ .InternalServerName }}" {{ .CommandFlags }}
 `
 )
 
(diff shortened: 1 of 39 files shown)


cmd/deploy/remote.go
```diff
@@ -62,6 +64,10 @@ ARG {{ .TokenArgName }}
 ARG {{ .ActionNameArgName }}
 ARG {{ .TlsCertBase64ArgName }}
 ARG {{ .InternalServerName }}
+ARG {{ .OktetoDeployable }}
+ARG {{ .GitHubRepositoryArgName }}
+ARG {{ .BuildKitHostArgName }}
+ARG {{ .OktetoRegistryURLArgName }}
```
ifbyol, line 70: These were added because in the previous way, the local execution of the command within BuildKit these were available there for some internal logic we have. Now, the runner doesn't execute that logic, so this would need to be in the environment to prevent a breaking change on someone relying on these env vars

cmd/deploy/remote.go
```diff
@@ -80,7 +86,7 @@ RUN okteto registrytoken install --force --log-output=json
 
 RUN --mount=type=secret,id=known_hosts --mount=id=remote,type=ssh \
   mkdir -p $HOME/.ssh && echo "UserKnownHostsFile=/run/secrets/known_hosts" >> $HOME/.ssh/config && \
-  okteto deploy --log-output=json --server-name="${{ .InternalServerName }}" {{ .DeployFlags }}
+  /okteto/bin/okteto remote-run --log-output=json --server-name="${{ .InternalServerName }}" {{ .CommandFlags }}
```
ifbyol, line 89: This change is the one I mentioned in the description. I'm changing the command to use the whole path of the binary we get from our image to prevent a breaking change for those who might be using a custom image with a pinned CLI version (we are setting the CLI with the least priority in the PATH).

pkg/cmd/stack/deploy.go
```diff
@@ -66,12 +66,18 @@ type analyticsTrackerInterface interface {
...
+	UpdatePod(spec apiv1.PodSpec) apiv1.PodSpec
+}
+
 // Stack is the executor of stack commands
 type Stack struct {
 	K8sClient        kubernetes.Interface
 	Config           *rest.Config
 	AnalyticsTracker analyticsTrackerInterface
 	IoCtrl           *io.Controller
+	Divert           Divert
```
ifbyol, line 80: Added this to apply the divert translation as part of the compose logic as compose aren't deployed through the proxy anymore

pkg/cmd/stack/translate.go
```diff
@@ -108,7 +108,8 @@ func translateConfigMap(s *model.Stack) *apiv1.ConfigMap {
 		ObjectMeta: metav1.ObjectMeta{
 			Name: model.GetStackConfigMapName(s.Name),
 			Labels: map[string]string{
-				model.StackLabel: "true",
+				model.StackLabel:      "true",
+				model.DeployedByLabel: format.ResourceK8sMetaString(s.Name),
```
ifbyol, line 112: As the compose is not deployed through the proxy anymore, we add the deployed-by label in the compose translation

pkg/deployable/runner.go
```diff
@@ -0,0 +1,414 @@
+// Copyright 2024 The Okteto Authors
```
ifbyol, line 1: Most logic of this file was moved from local.go

pkg/divert/driver.go
```diff
@@ -36,19 +37,25 @@ type Driver interface {
 	UpdateVirtualService(vs *istioNetworkingV1beta1.VirtualService)
 }
 
-func New(m *model.Manifest, c kubernetes.Interface) (Driver, error) {
+func New(divert *model.DivertDeploy, name, namespace string, c kubernetes.Interface) (Driver, error) {
```
ifbyol, line 40: This doesn't depend from the manifest anymore, just from the minimum information needed for it to work

pkg/divert/driver.go
```diff
@@ -36,19 +37,25 @@ type Driver interface {
...
 		return nil, fmt.Errorf("error creating istio client: %w", err)
 	}
 
-	return istio.New(m, c, ic), nil
+	return istio.New(divert, name, namespace, c, ic), nil
+}
+
+// NewNoop returns a new noop driver for divert. Useful for cases in which divert is not defined
+// in the manifest but the flow needs a driver
+func NewNoop() Driver {
```
ifbyol, line 59: Created this to simplify the logic in the compose translation and prevent the `if divertDriver != nil`

pkg/externalresource/serializer.go
```diff
@@ -75,3 +75,7 @@ func (er *ExternalResource) UnmarshalYAML(unmarshal func(interface{}) error) err
 
 	return nil
 }
+
+func (notes *Notes) MarshalYAML() (interface{}, error) {
```
ifbyol, line 79: This is needed to convert back the external resources Notes from `model.Manifest` to its representation in yaml (a single string)

maroshii (approved):

PR #4199, opened 2024-03-11
[Compose] Allow setting K8S labels from `docker-compose.yml`

andreafalzetti:
# Proposed changes

Fixes DEV-215

This PR changes enables users to set k8s labels when using docker compose manifests.

At the moment both `labels` and `annotations` are translated into k8s annotations, with this change compose annotations will be k8s labels, and viceversa.

|         | `master` | `af/compose-annotations-dev-215` |
|---------|--------|-----|
| `service` |    Screenshot 2024-03-18 at 10 32 16    |   Screenshot 2024-03-18 at 10 34 26  |
| `volume`  |    Screenshot 2024-03-18 at 10 30 53    |    Screenshot 2024-03-18 at 10 23 24 |

## How to validate

2. Override the `docker-compose.yml` with [this one](https://gist.github.com/andreafalzetti/30311d9dba40b46422e65671bdbf3550)
Highlight of changes from my test `docker-compose.yml` are:
a. Adding `services.api.labels` and `services.api.annotations`
b. Adding `volumes.data.labels`

3. Run `okteto deploy` and observe annotations and labels for the api service and the volume. Compose annotations should become k8s labels, and vice versa

Additional checks:
1. Try making a very long label/annotation and observe the CLI failing with:
```
 x  Error updating kubernetes service: Service "api" is invalid: [metadata.labels: Invalid value: "dev.okteto.com.dev.okteto.com.dev.okteto.com.dev.okteto.com.dev.okteto.com.dev.okteto.com.dev.okteto.com.dev.okteto.com.dev.okteto.com.dev.okteto.com.dev.okteto.com.": name part must be no more than 63 characters, metadata.labels: Invalid value: "dev.okteto.com.dev.okteto.com.dev.okteto.com.dev.okteto.com.dev.okteto.com.dev.okteto.com.dev.okteto.com.dev.okteto.com.dev.okteto.com.dev.okteto.com.dev.okteto.com.": name part must consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyName',  or 'my.name',  or '123-abc', regex used for validation is '([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9]')]
```

## CLI Quality Reminders 🔧

### DIFF
--- a/integration/k8s-client.go
+++ b/integration/k8s-client.go
@@ -109,10 +109,19 @@ func GetStatefulsetList(ctx context.Context, ns string, client kubernetes.Interf
 	return sfsList.Items, err
 }
 
-// WaitForDeployment waits until a deployment is rollout correctly
+// GetVolume returns a volume given a namespace and name
+func GetVolume(ctx context.Context, ns, name string, client kubernetes.Interface) (*corev1.PersistentVolumeClaim, error) {
+	return client.CoreV1().PersistentVolumeClaims(ns).Get(ctx, name, metav1.GetOptions{})
+}
+
+// WaitForDeployment waits until a deployment is rolled out correctly
 func WaitForDeployment(kubectlBinary string, kubectlOpts *commands.KubectlOptions, revision int, timeout time.Duration) error {
 	ticker := time.NewTicker(1 * time.Second)
+	defer ticker.Stop()
+
 	to := time.NewTicker(timeout)
+	defer to.Stop()
+
 	retry := 0
 	for {
 		select {
@@ -121,6 +130,7 @@ func WaitForDeployment(kubectlBinary string, kubectlOpts *commands.KubectlOption
 		case <-ticker.C:
 			output, err := commands.RunKubectlRolloutDeployment(kubectlBinary, kubectlOpts, revision)
 			if err != nil {
+				retry++
 				continue
 			}
 
@@ -140,6 +150,7 @@ func WaitForDeployment(kubectlBinary string, kubectlOpts *commands.KubectlOption
 						return nil
 					}
 				}
+				retry++
 				continue
 			}
 			if strings.Contains(output, "successfully rolled out") {
--- a/pkg/model/stack_serializer.go
+++ b/pkg/model/stack_serializer.go
@@ -332,6 +332,10 @@ func unmarshalVolume(volume *VolumeTopLevel) (*VolumeSpec, error) {
 		}
 
 		for key, value := range volume.Labels {
+			result.Labels[key] = value
+		}
+
+		for key, value := range volume.Annotations {
 			result.Annotations[key] = value
 		}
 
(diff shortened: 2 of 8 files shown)


pkg/model/stack_serializer.go
```diff
@@ -405,17 +405,19 @@ func (serviceRaw *ServiceRaw) ToService(svcName string, stack *Stack) (*Service,
...
+	}
+
 	if stack.IsCompose {
 		if len(serviceRaw.Args.Values) > 0 {
 			return nil, fmt.Errorf("unsupported field for services.%s: 'args'", svcName)
 		}
 		svc.Entrypoint.Values = serviceRaw.Entrypoint.Values
 		svc.Command.Values = serviceRaw.Command.Values
-
-		for key, value := range unmarshalLabels(serviceRaw.Labels, serviceRaw.Deploy) {
```
andreafalzetti, line 415: in the serializer we were converting `labels` into `annotations` - I think the model should reflect the manifest, so labels -> labels, annotations -> annotations.

Then, in the deployer to k8s, we can translate them into each other.

integration/k8s-client.go
```diff
@@ -121,6 +130,7 @@ func WaitForDeployment(kubectlBinary string, kubectlOpts *commands.KubectlOption
 		case <-ticker.C:
 			output, err := commands.RunKubectlRolloutDeployment(kubectlBinary, kubectlOpts, revision)
 			if err != nil {
+				retry++
```
andreafalzetti, line 133: nit: the condition below `retry%10 == 0` is always `true` because we never increment the counter

integration/k8s-client.go
```diff
@@ -109,10 +109,19 @@ func GetStatefulsetList(ctx context.Context, ns string, client kubernetes.Interf
...
-// WaitForDeployment waits until a deployment is rollout correctly
+// GetVolume returns a volume given a namespace and name
+func GetVolume(ctx context.Context, ns, name string, client kubernetes.Interface) (*corev1.PersistentVolumeClaim, error) {
+	return client.CoreV1().PersistentVolumeClaims(ns).Get(ctx, name, metav1.GetOptions{})
+}
+
+// WaitForDeployment waits until a deployment is rolled out correctly
 func WaitForDeployment(kubectlBinary string, kubectlOpts *commands.KubectlOptions, revision int, timeout time.Duration) error {
 	ticker := time.NewTicker(1 * time.Second)
+	defer ticker.Stop()
```
andreafalzetti, line 120: nit: added for best practice

pkg/model/stack_serializer.go
ifbyol: Hmmm, I think that can be confusing. I don't think the model should represent always what is present in the manifest. We have some other things we change in the serializer. That's the idea of the serializer. Read the manifest and adapt it to whatever model we need internally. Propagate the difference between labels and annotations in compose to another layer might generate confusion. Also, I think you are changing more things here. If we translate a stack (not a compose), the labels have to be labels and annotations have to be annotations. Take a look to the else branch starting in line 419. Doing this change you are doing, we are inverting the meaning of labels and annotations for stacks, and that is not right. Could you review it, please?

Independently of where we end up having the translation from `labels` to `annotations` and viceversa, let's add a comment in the code, please. It is easy to be confused in several months, specially if someone without context in the CLI passes by that code.

ifbyol (changes requested):
I think we have changed the meaning of labels and annotations for stacks instead of doing it only for compose. The logic in the serializer is shared for both cases and we have removed the logic that was in the if/else checking that scenario

ifbyol:
@pchico83 What should be the behavior when the label specified in the annotations field is not compliance with the k8s labels limitation? Should we fail or just skip that label? My main concern is that composes that right now are working because something is translated to annotations could be broken because the value is not a valid k8s label value.

Also, this is a breaking change, should we be careful somehow about previous behavior? For example, should we add the annotations to both (labels and annotations) to prevent issues in scenarios like adding annotations for ArgoCD?

pchico83:
> @pchico83 What should be the behavior when the label specified in the annotations field is not compliance with the k8s labels limitation? Should we fail or just skip that label? My main concern is that composes that right now are working because something is translated to annotations could be broken because the value is not a valid k8s label value.
> 
> Also, this is a breaking change, should we be careful somehow about previous behavior? For example, should we add the annotations to both (labels and annotations) to prevent issues in scenarios like adding annotations for ArgoCD?

I would fail if the label is not compliant and also, document the breaking change. The current behavior feels more like a bug

andreafalzetti:
Closing in favour of https://github.com/okteto/okteto/pull/4204

PR #4203, opened 2024-03-18
Smart Build fixes and improvements

ifbyol:
# Proposed changes

Fixes DEV-187

## Context

A bit of context on how smart builds have to work. This only applies when the images to build doesn't specify a destination image through the `image` field. If the `image` field is set, we just use that one as destiny.

The idea is to use the dev registry (`okteto.dev`) as the registry where we have the latest build done for any image. And use the global registry (`okteto.global`) as a cache for images already built. 

That latest images stored in the dev registry have the tag `okteto`.

### Build flow

When we build an image (`okteto build` or `okteto deploy --build`), we do the following:
* We get the build context hash, which is mainly:
  * The last commit of the build context (folder) specified for an image
  * A hash of the git diff of the build context (folder) specified for an image
* We calculate a SHA for the service image with the different information like the build context, image targer, build args, etc...
* With that SHA, we see if the image is already built in the global registry:
  * If it is built, we just copy that image to the dev registry using the tag `okteto`
  * If it is not built, we build the image and push it to both global (`okteto.gobal/<image-name>:<calculated-sha>`) and dev (`okteto.dev/<image-name>:okteto`) registries

### Deploy build

When a deploy is executed with a manifest, we do the following operations:
* For every image defined in the manifest, we check if the tag `okteto` exists in the dev registry (`okteto.dev`).
  * If the image exists, we don't build it and use that one
  * If the image doesn't exist in the dev registry, we build it, and the flow mentioned above will start (image might be finally built or not depending on if that variable is present in the global registry with the sha or not)

### Changes

This PR includes several fixes and changes:
* Fixed an issue in which the `globalNamespace` flag for the context wasn't being propagated correctly to build command and the images for the global registry were being pushed to the wrong repository making the smart builds not working in dogfooding environments
* Remove `Building image '/tmp/....` message displayed as part of remote deploy and destroy as it is misleading because the main operation is not a build
* We remove the support of volume mounts as part of compose building our own images with the volumes. This is a breaking change.
* Project commit implementation of smart builds is now deleted and we only have the build context one
* We don't have a different logic if the current repository/context is dirty or not. We always behave in the same way
* When there is an error getting the current commit or getting the git diff, we generate a random commit/diff sha, so we force a rebuild of the image
* If there is an error trying to see if an image is already present or not in the registry, we consider that service to be built instead of failing the operation
* Fixed an issue with the internal cache of the hasher as it was caching the image SHA using the folder name, so 2 images using different targets or different build args were ending up with the same SHA (and they shouldn't)

## How to validate

Do different builds and deploy operations, cases I was testing:
* When the `image` field is set, that registry is used to be pushed.
  * When deploying (without specifically build the image), if the image already exists, we use that one
  * When building (`okteto build` or `okteto deploy --build`), the image is always built as we don't know if it has the latest changes or not
* When the `image` field is not set:
  * For new images, they are built and pushed to both global and dev registry.
    * The global one should use a SHA as tag
    * The dev one should use the `okteto` tag
  * If I change to a new namespace and I build the same images without doing any change on the repo, images are not built but they are copied from the global registry to the dev one
  * If I try to build the images again, it should say they already exists and don't build anything
  * Run `okteto build --no-cache` will rebuild them again even if they exists
  * Now I do changes to my repository and push it
  * If I try to build them, they should be built as a new commit is detected and they have to be pushed to both global and dev registry
  * If I go to the context of an image (folder) and I do any change there without pushing it and I build the images, only the affected image should be built pushing to both registries
  * If I deploy the manifest, images should not be built and the `okteto` tag will be built
  * If I change to a new namespace and I deploy the manifest, it will detect that images are not in the dev registry and it will try to build them. When building the images, it will realize that the images are in the global registry and they will be copied to the dev registry one
  * Then I do any change in the repository and push it to have a new commit. Then, change to a new namespace and deploy the manifest. It will trigger the build of the images

## CLI Quality Reminders 🔧

### DIFF
--- a/cmd/build/v2/services.go
+++ b/cmd/build/v2/services.go
@@ -103,7 +103,7 @@ func (bc *OktetoBuilder) checkServiceToBuildDuringDeploy(service string, manifes
 		buildCh <- service
 		return nil
 	} else if err != nil {
-		bc.ioCtrl.Out().Warning("could not verify if image for service %s is already in the registry. Building image...", service)
+		bc.ioCtrl.Out().Warning("could not verify if image for service %q is already in the registry. Building image...", service)
 		// If there is an error trying to get the image from the registry, we just rebuild that image
 		bc.ioCtrl.Logger().Debugf("unexpected error checking if the images exist: %s", err)
 		buildCh <- service
--- a/cmd/build/v2/services.go
+++ b/cmd/build/v2/services.go
@@ -96,7 +96,7 @@ func (bc *OktetoBuilder) checkServiceToBuildDuringDeploy(service string, manifes
 		buildInfo.Image = ""
 	}
 
-	imageChecker := getImageChecker(buildInfo, bc.Config, bc.Registry, bc.smartBuildCtrl, bc.ioCtrl.Logger())
+	imageChecker := getImageChecker(bc.Config, bc.Registry, bc.smartBuildCtrl, bc.ioCtrl.Logger())
 	imageWithDigest, err := imageChecker.getImageDigestReferenceForServiceDeploy(manifest.Name, service, buildInfo)
 	if oktetoErrors.IsNotFound(err) {
 		bc.ioCtrl.Logger().Debug("image not found, building image")
--- a/cmd/build/v2/services.go
+++ b/cmd/build/v2/services.go
@@ -19,6 +19,7 @@ import (
 	"strings"
 
 	oktetoErrors "github.com/okteto/okteto/pkg/errors"
+	oktetoLog "github.com/okteto/okteto/pkg/log"
 	"github.com/okteto/okteto/pkg/model"
 	"golang.org/x/sync/errgroup"
 )
@@ -103,6 +104,7 @@ func (bc *OktetoBuilder) checkServiceToBuildDuringDeploy(service string, manifes
 		buildCh <- service
 		return nil
 	} else if err != nil {
+		oktetoLog.Warning("could not verify if image for service %s is already in the registry. Building image...", service)
 		// If there is an error trying to get the image from the registry, we just rebuild that image
 		bc.ioCtrl.Logger().Debugf("unexpected error checking if the images exist: %s", err)
 		buildCh <- service
--- a/cmd/build/v2/services.go
+++ b/cmd/build/v2/services.go
@@ -103,6 +103,7 @@ func (bc *OktetoBuilder) checkServiceToBuildDuringDeploy(service string, manifes
 		buildCh <- service
 		return nil
 	} else if err != nil {
+		// If there is an error trying to get the image from the registry, we just rebuild that image
 		bc.ioCtrl.Logger().Debugf("unexpected error checking if the images exist: %s", err)
 		buildCh <- service
 		return nil
--- a/cmd/build/v2/services.go
+++ b/cmd/build/v2/services.go
@@ -19,7 +19,6 @@ import (
 	"strings"
 
 	oktetoErrors "github.com/okteto/okteto/pkg/errors"
-	oktetoLog "github.com/okteto/okteto/pkg/log"
 	"github.com/okteto/okteto/pkg/model"
 	"golang.org/x/sync/errgroup"
 )
@@ -104,7 +103,7 @@ func (bc *OktetoBuilder) checkServiceToBuildDuringDeploy(service string, manifes
 		buildCh <- service
 		return nil
 	} else if err != nil {
-		oktetoLog.Warning("could not verify if image for service %s is already in the registry. Building image...", service)
+		bc.ioCtrl.Out().Warning("could not verify if image for service %s is already in the registry. Building image...", service)
 		// If there is an error trying to get the image from the registry, we just rebuild that image
 		bc.ioCtrl.Logger().Debugf("unexpected error checking if the images exist: %s", err)
 		buildCh <- service
--- a/cmd/build/v2/services.go
+++ b/cmd/build/v2/services.go
@@ -97,7 +97,7 @@ func (bc *OktetoBuilder) checkServiceToBuild(service string, manifest *model.Man
 	var err error
 	var buildHash string
 	if bc.smartBuildCtrl.IsEnabled() {
-		buildHash, err = bc.smartBuildCtrl.GetBuildHash(buildInfo)
+		buildHash, err = bc.smartBuildCtrl.GetBuildHash(buildInfo, service)
 		if err != nil {
 			bc.ioCtrl.Logger().Infof("error getting build hash: %s", err)
 		}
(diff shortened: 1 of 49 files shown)


pkg/cmd/build/constants.go
```diff
@@ -0,0 +1,21 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package build
+
+const (
+	// DeployOutputModeOnBuild Defines the output mode set to BuildOptions when running a remote deploy
+	DeployOutputModeOnBuild = "deploy"
+	// DestroyOutputModeOnBuild Defines the output mode set to BuildOptions when running a remote destroy
+	DestroyOutputModeOnBuild = "destroy"
```
andreafalzetti, line 20: nit: we could also move `OktetoDefaultImageTag` from the `model` pkg, to keep reduce the model's size and responsability / avoid the import/dependency but also because that const is very much in the context of the build/registry

pkg/cmd/build/buildkit.go
```diff
@@ -364,12 +364,12 @@ func solveBuild(ctx context.Context, c *client.Client, opt *client.SolveOpt, pro
 			// We need to wait until the tty channel is closed to avoid writing to stdout while the tty is being used
 			_, err := progressui.DisplaySolveStatus(context.TODO(), c, ioCtrl.Out(), ttyChannel)
 			return err
-		case "deploy":
```
andreafalzetti, line 367: small change but very appreciated :)

pkg/cmd/build/build.go
```diff
@@ -85,19 +85,22 @@ func (ob *OktetoBuilder) GetBuilder() string {
...
-	builder := ob.GetBuilder()
-	buildMsg := fmt.Sprintf("Building '%s'", buildOptions.File)
-	depotEnabled := IsDepotEnabled()
-	if depotEnabled {
-		ioCtrl.Out().Infof("%s on depot's machine...", buildMsg)
-	} else if builder == "" {
-		ioCtrl.Out().Infof("%s using your local docker daemon", buildMsg)
-	} else {
-		ioCtrl.Out().Infof("%s in %s...", buildMsg, builder)
+	if !isDeployOrDestroy {
```
andreafalzetti, line 93: nice fix, that log was terrible, thanks for fixing it

pkg/cmd/build/constants.go
ifbyol: It seems easy, but it generates cycle import, we should for sure look for a better place for that constant, but I won't do it as part of this change just because I'm not sure which would be the right place and this PR already has several changes

cmd/build/v2/tagger.go
```diff
@@ -103,80 +108,33 @@ func (it imageTagger) getImageReferencesForTag(manifestName, svcToBuildName, tag
...
+func (imageTagger) getImageReferencesForDeploy(manifestName, svcToBuildName string) []string {
+	var imageReferences []string
 	sanitizedName := format.ResourceK8sMetaString(manifestName)
-	tagsToCheck := []string{}
-	for _, targetRegistry := range getTargetRegistries(i.cfg.IsOkteto()) {
-		tagsToCheck = append(tagsToCheck, useReferenceTemplateWithVolumes(targetRegistry, sanitizedName, svcToBuildName, tag))
-	}
-	return tagsToCheck
-}
+	imageReferences = append(imageReferences, useReferenceTemplate(constants.DevRegistry, sanitizedName, svcToBuildName, model.OktetoDefaultImageTag))
```
jLopezbarb, line 133: NIT: For me it's easier to read it like this but feel free to leave it as it is
```suggestion
	sanitizedName := format.ResourceK8sMetaString(manifestName)
	imageReferences := []string{useReferenceTemplate(constants.DevRegistry, sanitizedName, svcToBuildName, model.OktetoDefaultImageTag)}
	
	return imageReferences
```

cmd/build/v2/services.go
```diff
@@ -86,31 +87,28 @@ func (bc *OktetoBuilder) GetServicesToBuild(ctx context.Context, manifest *model
...
+	imageChecker := getImageChecker(bc.Config, bc.Registry, bc.smartBuildCtrl, bc.ioCtrl.Logger())
+	imageWithDigest, err := imageChecker.getImageDigestReferenceForServiceDeploy(manifest.Name, service, buildInfo)
 	if oktetoErrors.IsNotFound(err) {
 		bc.ioCtrl.Logger().Debug("image not found, building image")
 		buildCh <- service
 		return nil
 	} else if err != nil {
-		return err
+		oktetoLog.Warning("could not verify if image for service %s is already in the registry. Building image...", service)
+		// If there is an error trying to get the image from the registry, we just rebuild that image
```
jLopezbarb, line 108: We should use the new logger already injected in the oktetoBuilder
```suggestion
		bc.ioCtrl.Out().Warning("could not verify if image for service %s is already in the registry. Building image...", service)
```

pkg/repository/git.go
```diff
@@ -194,7 +195,7 @@ func (r gitRepoController) GetDiffHash(contextDir string) (string, error) {
 	diffCh := make(chan diffResponse)
 	untrackedFilesCh := make(chan untrackedFilesResponse)
 
-	timeout := 1 * time.Second
+	timeout := 5 * time.Second
```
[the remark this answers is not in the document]
ifbyol: Done!

pkg/registry/registry.go
```diff
@@ -195,5 +197,12 @@ func (or OktetoRegistry) CloneGlobalImageToDev(imageWithDigest, tag string) (str
 		return "", err
 	}
 
-	return devImage, nil
+	// To return always the sha256 os the dev image
+	r, err := or.GetImageTagWithDigest(devImage)
+	if err != nil {
+		// If there is an error getting the tag with digest we just return the dev image
```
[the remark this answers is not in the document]
ifbyol: Done!

pkg/cmd/build/buildkit.go
```diff
@@ -364,12 +364,12 @@ func solveBuild(ctx context.Context, c *client.Client, opt *client.SolveOpt, pro
...
-			err := deployDisplayer(context.TODO(), plainChannel, &types.BuildOptions{OutputMode: "deploy"})
+		case DeployOutputModeOnBuild:
+			err := deployDisplayer(context.TODO(), plainChannel, &types.BuildOptions{OutputMode: DeployOutputModeOnBuild})
 			commandFailChannel <- err
 			return err
-		case "destroy":
-			err := deployDisplayer(context.TODO(), plainChannel, &types.BuildOptions{OutputMode: "destroy"})
```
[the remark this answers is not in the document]
ifbyol: Changed!

cmd/build/v2/tagger.go
ifbyol: Changed!

cmd/build/v2/services.go
ifbyol: Changed! I had to add the `Warning` function to the IOController but it wasn't yet

jLopezbarb (approved):

andreafalzetti (approved):

maroshii (approved):

PR #4204, opened 2024-03-19
[Compose] Allow setting K8S labels from `docker-compose.yml`

andreafalzetti:
# Proposed changes

Fixes DEV-215

Note: this is an alternative PR to https://github.com/okteto/okteto/pull/4199. The original PR approached the problem by changing how the serializer would create the model (labels into labels, annotations into annotations), however it impacted stack too. With this PR we only change compose, as intended, and also used an approach that requires less changes and may be less confusing, which is to keep the serialization as is, and make the requires changes using the current logic.

This PR changes enables users to set k8s labels when using docker compose manifests.

At the moment both `labels` and `annotations` are translated into k8s annotations, with this change compose annotations will be k8s labels, and viceversa.

|         | `master` | `af/compose-annotations-dev-215-2` |
|---------|--------|-----|
| `service` |    Screenshot 2024-03-18 at 10 32 16    |   Screenshot 2024-03-18 at 10 34 26  |
| `volume`  |    Screenshot 2024-03-18 at 10 30 53    |    Screenshot 2024-03-18 at 10 23 24 |

## How to validate

2. Override the `docker-compose.yml` with [this one](https://gist.github.com/andreafalzetti/30311d9dba40b46422e65671bdbf3550)
Highlight of changes from my test `docker-compose.yml` are:
a. Adding `services.api.labels` and `services.api.annotations`
b. Adding `volumes.data.labels`

3. Run `okteto deploy` and observe annotations and labels for the api service and the volume. Compose annotations should become k8s labels, and vice versa

Additional checks:
1. Try making a very long label/annotation and observe the CLI failing with:
```
 x  Error updating kubernetes service: Service "api" is invalid: [metadata.labels: Invalid value: "dev.okteto.com.dev.okteto.com.dev.okteto.com.dev.okteto.com.dev.okteto.com.dev.okteto.com.dev.okteto.com.dev.okteto.com.dev.okteto.com.dev.okteto.com.dev.okteto.com.": name part must be no more than 63 characters, metadata.labels: Invalid value: "dev.okteto.com.dev.okteto.com.dev.okteto.com.dev.okteto.com.dev.okteto.com.dev.okteto.com.dev.okteto.com.dev.okteto.com.dev.okteto.com.dev.okteto.com.dev.okteto.com.": name part must consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyName',  or 'my.name',  or '123-abc', regex used for validation is '([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9]')]
```

## CLI Quality Reminders 🔧

### DIFF
--- a/integration/k8s-client.go
+++ b/integration/k8s-client.go
@@ -109,16 +109,26 @@ func GetStatefulsetList(ctx context.Context, ns string, client kubernetes.Interf
 	return sfsList.Items, err
 }
 
+// GetVolume returns a volume given a namespace and name
+func GetVolume(ctx context.Context, ns, name string, client kubernetes.Interface) (*corev1.PersistentVolumeClaim, error) {
+	return client.CoreV1().PersistentVolumeClaims(ns).Get(ctx, name, metav1.GetOptions{})
+}
+
 // WaitForDeployment waits until a deployment is rollout correctly
 func WaitForDeployment(kubectlBinary string, kubectlOpts *commands.KubectlOptions, revision int, timeout time.Duration) error {
 	ticker := time.NewTicker(1 * time.Second)
+	defer ticker.Stop()
+
 	to := time.NewTicker(timeout)
+	defer to.Stop()
+
 	retry := 0
 	for {
 		select {
 		case <-to.C:
 			return fmt.Errorf("%s didn't rollout after %s", kubectlOpts.Name, timeout.String())
 		case <-ticker.C:
+			retry++
 			output, err := commands.RunKubectlRolloutDeployment(kubectlBinary, kubectlOpts, revision)
 			if err != nil {
 				continue
--- a/pkg/model/stack_serializer.go
+++ b/pkg/model/stack_serializer.go
@@ -399,11 +399,14 @@ func (serviceRaw *ServiceRaw) ToService(svcName string, stack *Stack) (*Service,
 		translateHealtcheckCurlToHTTP(svc.Healtcheck)
 	}
 
-	svc.Annotations = serviceRaw.Annotations
+	svc.NodeSelector = serviceRaw.NodeSelector
+
+	if svc.Labels == nil {
+		svc.Labels = make(Labels)
+	}
 	if svc.Annotations == nil {
 		svc.Annotations = make(Annotations)
 	}
-	svc.NodeSelector = serviceRaw.NodeSelector
 
 	if stack.IsCompose {
 		if len(serviceRaw.Args.Values) > 0 {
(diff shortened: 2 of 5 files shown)


integration/k8s-client.go
```diff
@@ -109,16 +109,26 @@ func GetStatefulsetList(ctx context.Context, ns string, client kubernetes.Interf
 	return sfsList.Items, err
 }
 
+// GetVolume returns a volume given a namespace and name
+func GetVolume(ctx context.Context, ns, name string, client kubernetes.Interface) (*corev1.PersistentVolumeClaim, error) {
```
maroshii, line 113: Is this being used?

andreafalzetti: @maroshii https://github.com/okteto/okteto/pull/4204/files#diff-57626eb367add33c287ac942e1a7a64fa1181b54d59f0aaf5647957c8b9b92b7R238 and another instance below

pkg/model/stack_serializer.go
```diff
@@ -317,22 +317,27 @@ func (s *Stack) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
-		}
+func unmarshalVolume(volume *VolumeTopLevel, isCompose bool) (*VolumeSpec, error) {
+	result := &VolumeSpec{
+		Labels:      make(Labels),
+		Annotations: make(Annotations),
+	}
 
-		for key, value := range volume.Labels {
-			result.Annotations[key] = value
+	if volume != nil {
```
jLopezbarb, line 326: NIT: Maybe we can do something like this in order tohave less indentation and make it more readable. WDYT?
```suggestion
	if volume == nil {
	    return result
	}
	if isCompose {
	...
```

jLopezbarb (approved):
Tested with several projects from https://github.com/docker/awesome-compose and it's working as expected with stacks and compose (changing the name of the file)

pkg/model/stack_serializer.go
andreafalzetti: I don't have a strong preference so I've updated it in commit: fcc5bc75aad81c54cf7b5d1500a6907ddba4e2ef

pkg/model/stack_serializer.go
```diff
@@ -317,45 +317,52 @@ func (s *Stack) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
-					if err != nil {
-						return nil, err
-					}
-					result.Size.Value = qK8s
-				}
-				if key == "class" {
-					result.Class = value
+		for key, value := range volume.Labels {
+			result.Annotations[key] = value
+		}
```
ifbyol, line 343: Shouldn't this go to labels?

andreafalzetti: At the moment, for `stack` volumes we always transform both labels and annotations into annotations, so I don't want to change the behavior, which would be out of scope for this PR

PR #4205, opened 2024-03-19
feat: add events for build using manifest

jLopezbarb:
# Proposed changes

DEV-203

This PR adds a k8s event whenever a build is triggered using a manifest v2 either if it's called from an `okteto build` command or a `okteto deploy` command.

The event follows the following schema:
```json
{
	"$schema": "http://json-schema.org/draft-07/schema#",
	"type": "object",
	"properties": {
	  "devenv_name": {
		"type": "string",
		"description": "Name of the development environment"
	  },
	  "image_name": {
		"type": "string",
		"format": "string",
		"description": "The resulting image name"
	  },
	  "repository": {
		"type": "string",
		"format": "uri",
		"description": "URL of the code repository"
	  },
	  "namespace": {
		"type": "string",
		"description": "Namespace where the deployment will be created"
	  },
	  "smartBuildHit": {
		"type": "bool",
                "description": "If the image has been found in the cache or not",
	  },
	  "success": {
		"type": "boolean",
		"description": "Whether the overall process was successful"
	  },
	  "schema_version": {
		"type": "string",
		"description": "Version of the schema"
	  }
	  "duration": {
		"type": "float",
		"description": "Duration of the build in seconds"
	  }
	},
  }
```

## How to validate

1. Run `okteto build` on [okteto/movies](https://github.com/okteto/movies/blob/main/okteto.yml)
2. Run `kubectl get events.events.k8s.io -w`
3. Check that for each of the images there is a event created 

## CLI Quality Reminders 🔧


maroshii (approved):

ifbyol (approved):
Some minor comment and I agree with Fran. We are abusing of creating things inline instead of using dependency injection properly. We should start doing it properly as much as we can. But it has a broader scope than this PR

cmd/build/v2/analytics.go
```diff
@@ -0,0 +1,119 @@
...
+func NewEventTracker(ioCtrl *io.Controller, okCtx buildCmd.OktetoContextInterface) *eventTracker {
+	k8sClientProvider := okteto.NewK8sClientProvider()
+	c, _, err := k8sClientProvider.Provide(okCtx.GetCurrentCfg())
+	if err != nil {
+		ioCtrl.Logger().Infof("could not get k8s client: %s", err)
+	}
+	return &eventTracker{
+		eventTypeConverter: map[bool]EventType{
+			true:  normalEventType,
+			false: warningEventType,
```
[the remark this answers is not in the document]
jLopezbarb: It's just another way to filter the events with kubectl (we can't filter by the data we are sending). We could always send the event as `Normal` or leave the field empty but @jpf-okteto and me agreed that it won't hurt

pkg/insights/insights.go
```diff
@@ -0,0 +1,79 @@
...
+		k8sClientProvider: k8sClientProvider,
+		ioCtrl:            ioCtrl,
+	}
+}
+
+// TrackEvent tracks an event in the cluster
+// namespace: the namespace where the event is happening
+// insightType: the type of the event (for example: build, deploy, etc.)
+// data: the data of the event as JSON string
+func (ip *InsightsPublisher) TrackEvent(ctx context.Context, namespace, insightType, data string) {
```
ifbyol, line 53: It seems that this even should/could be private, right? And the ones to be public could be only the ones for build and deploy

pkg/repository/git.go
```diff
@@ -74,6 +75,42 @@ func (r gitRepoController) calculateIsClean(ctx context.Context) (bool, error) {
...
+	remotes, err := repo.Remotes()
+	if err != nil {
+		return "", fmt.Errorf("failed to get git repo's remote information: %w", err)
+	}
+
+	if len(remotes) == 0 {
+		return "", fmt.Errorf("git repo doesn't have any remote")
+	}
+
+	return remotes[0].Config().URLs[0], nil
```
ifbyol, line 111: Nit: We have a function with this exact same content in `model/utils/utils.go`. I would try to reuse it as much as possible (we can even move it out from model to avoid to import `model`

ifbyol (approved):
Minor comments but it looks good

pkg/repository/git.go
jLopezbarb: The idea is to remove it from `model/utils/utils.go`, but I'd do it as a separate PR

PR #4206, opened 2024-03-19
feat: add deploy insights

jLopezbarb:
# Proposed changes

DEV-201

This PR adds a k8s event whenever a build is triggered using a manifest v2 either if it's called from an okteto build command or a okteto deploy command.

The event follows the following schema:
```json
{
	"$schema": "http://json-schema.org/draft-07/schema#",
	"type": "object",
	"properties": {
	  "devenv_name": {
		"type": "string",
		"description": "Name of the development environment"
	  },
	  "repository": {
		"type": "string",
		"format": "uri",
		"description": "URL of the code repository"
	  },
	  "namespace": {
		"type": "string",
		"description": "Namespace where the deployment will be created"
	  },
	  "phases": {
		"type": "array",
		"description": "An array of the phases to be executed",
		"items": {
		  "type": "object",
		  "properties": {
			"name": {
			  "type": "string",
			  "description": "Name of the phase or the index of the command in the array"
			},
			"duration": {
			  "type": "integer",
			  "description": "Execution time of the command in seconds"
			}
		  },
		  "required": ["name", "idx", "duration"]
		}
	  },
	  "success": {
		"type": "boolean",
		"description": "Whether the overall process was successful"
	  },
	  "schema_version": {
		"type": "string",
		"description": "Version of the schema"
	  
	  }
	},
  }
```

## How to validate

1. Run `okteto deploy (--remote) (--wait)` on [okteto/movies](https://github.com/okteto/movies/blob/main/okteto.yml)
2. Run `kubectl get events.events.k8s.io -w`
3. Check that  there is a event created after the deploy with all the times

## CLI Quality Reminders 🔧


cmd/deploy/analytics.go
```diff
@@ -0,0 +1,129 @@
+// Copyright 2024 The Okteto Authors
```
maroshii, line 1: This looks like a copy of `cmd/build/v2/analytics.go` from the other PR. I think there's a lot in common and we should avoid copy/pasting this. If at some point we want to start sending other events we would need to copy/paste this yet again.

In terms of design, I think there should be a distinction between the transport and the payload.

- transport: The mechanism by which the data is sent. The k8s event, event type etc. Shared by deploy/build and others.
- payload: The actual data that is sent by the transport. This is where build and deploy will differ. They will simply have different data

The data is all serialized into json and sent in the `Note` metadata field so there's **a lot** in common between build and deploy. The payload could simply be a `json.Marshaler`. If I understand correctly, all the rest is the same.

ifbyol: Agree with this. We even do things differently, for example, the k8s client provider here is passed as parameter in the `New...` function while for the build events we create it within it.

The track function has to be generic and the build and deploy operations should be the ones gathering the data and converting it to the common interface defined for the track operation

pkg/cmd/pipeline/translate.go
```diff
@@ -189,6 +195,30 @@ func UpdateEnvs(ctx context.Context, name, namespace string, envs []string, c ku
...
+	}
+	phases = append(phases, phaseJSON{
+		Name:     phase,
+		Duration: duration.Seconds(),
+	})
+	encodedPhases, err := json.Marshal(phases)
+	if err != nil {
+		return err
+	}
+	cmap.Data[phasesField] = string(encodedPhases)
```
ifbyol, line 218: Would it make sense to encode the data to base64 as we do with other data like variables?

pkg/deployable/runner.go
```diff
@@ -280,16 +282,22 @@ func (r *Runner) runCommandsSection(ctx context.Context, params DeployParameters
...
+	for idx, command := range params.Deployable.Commands {
 		oktetoLog.Information("Running '%s'", command.Name)
 		oktetoLog.SetStage(command.Name)
 		oktetoLog.AddToBuffer(oktetoLog.InfoLevel, "Executing command '%s'...", command.Name)
 
-		if err := r.Executor.Execute(command, params.Variables); err != nil {
+		startTime := time.Now()
+		err := r.Executor.Execute(command, params.Variables)
+		elapsedTime := time.Since(startTime)
+		if err := r.ConfigMapHandler.AddPhase(ctx, params.Name, params.Namespace, fmt.Sprint(idx), elapsedTime); err != nil {
```
ifbyol, line 293: Why do we use the index as phase name?

pkg/cmd/pipeline/translate.go
```diff
@@ -189,6 +195,30 @@ func UpdateEnvs(ctx context.Context, name, namespace string, envs []string, c ku
 	return nil
 }
 
+func AddPhase(ctx context.Context, name, namespace, phase string, duration time.Duration, c kubernetes.Interface) error {
```
ifbyol, line 198: Include a comment to this function. It is not easy to know what a `Phase` is

pkg/deployable/runner.go
```diff
@@ -19,6 +19,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"time"
```
ifbyol, line 22: Don't you need to include also the time for the divert and external resources deployment? Those are also happening as part of the runner

cmd/deploy/cfghandler.go
```diff
@@ -113,6 +115,14 @@ func (ch *defaultConfigMapHandler) UpdateEnvsFromCommands(ctx context.Context, n
 	return nil
 }
 
+func (ch *defaultConfigMapHandler) AddPhase(ctx context.Context, name, namespace, phase string, duration time.Duration) error {
```
ifbyol, line 118: For the future, out of scope of this PR. If we have 2 different implementations to satisfy an interface and some of the functions have duplicated the logic, it is error prone. We should extract those implementations to a common piece and the implementations should use that common piece through struct embedding or dependency

pkg/cmd/pipeline/translate.go
jLopezbarb: I don't think it's necessary given that this data is not going to grow too much in the future and from the k8s spec it should be human readable.

pkg/deployable/runner.go
jLopezbarb: On the initial spec it was only though to be on compose/commands/wait. Let's ping @jpf-okteto in order to check if we should also include those one

pkg/cmd/pipeline/translate.go
ifbyol: It is not about the length, it is about the format and possible issues with chars or similar. As we are controlling the format of that, there might not be issues, but just in case. Anyways, this is completely a nit, you can ignore it

ifbyol:
@jLopezbarb I think you need to merge master. There are changes not sure they should appear (apart from the conflicts)

pkg/cmd/pipeline/translate.go
jLopezbarb: Right now it's a json that we generate. We'll have to rethink it if we add the commands names but right now we are the ones that set all the naming in the json generated

pkg/insights/deploy.go
```diff
@@ -0,0 +1,92 @@
...
+
+	var phases []phaseJSON
+	if err := json.Unmarshal([]byte(val), &phases); err != nil {
+		ip.ioCtrl.Logger().Infof("could not unmarshal phases from cmap: %s", err)
+		return
+	}
+
+	deployEvent := &deployEventJSON{
+		DevenvName:    name,
+		Repository:    cmap.Data["repository"],
```
ifbyol, line 79: We have to obfuscate this to prevent to send username/password as part of the repository information. 

Also, should we standardize how we send the repo? I mean, `git@github.com:okteto/okteto.git`, `https://github.com/okteto/okteto`, `https://github.com/okteto/okteto.git` and `ssh://git@github.com/okteto/okteto` points to the same repository but it wouldn't be easy to be grouped by repository if we send them in a different way. I think we should unify them and send them always in the same way (always https://, always ssh:// or similar)

cmd/deploy/deploy.go
```diff
@@ -464,7 +486,8 @@ func (dc *Command) deploy(ctx context.Context, deployOptions *Options, cwd strin
 		stage := "Endpoints configuration"
 		oktetoLog.SetStage(stage)
 		oktetoLog.Information("Running stage '%s'", stage)
-		if err := dc.deployEndpoints(ctx, deployOptions); err != nil {
+		err := dc.deployEndpoints(ctx, deployOptions)
+		if err != nil {
```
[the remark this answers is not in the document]
jLopezbarb: The previous implementation was getting all the duration for this phase also. We discarded it until we know how to treat null values. Reverting

pkg/cmd/pipeline/translate.go
```diff
@@ -189,6 +195,43 @@ func UpdateEnvs(ctx context.Context, name, namespace string, envs []string, c ku
 	return nil
 }
 
+// AddPhaseDuration adds a new phase to the configmap with the duration in seconds
+func AddPhaseDuration(ctx context.Context, name, namespace, phase string, duration time.Duration, c kubernetes.Interface) error {
+	cmap, err := configmaps.Get(ctx, TranslatePipelineName(name), namespace, c)
+	if err != nil {
+		return err
+	}
+	val, ok := cmap.Data[phasesField]
```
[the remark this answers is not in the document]
jLopezbarb: Given that right now we are only adding the phases field in order to track events and they are calculated and created in K8s in the CLI execution, there is no scenario where we might lose the information. 

If we wanted to store the information to show it on the UI, there is one scenario that we have to take a look (deploying with a previous CLI version)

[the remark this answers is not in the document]
jLopezbarb: Given that right now we are only adding the phases field in order to track events and they are calculated and created in K8s in the CLI execution, there is no scenario where we might lose the information. 

If we wanted to store the information to show it on the UI, there is one scenario that we have to take a look (deploying with a previous CLI version)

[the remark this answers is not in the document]
ifbyol: Would there be an scenario in which a deploy within another deploy would use the same configmap, overriding the `phase` field for the main one?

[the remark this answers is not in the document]
jLopezbarb: Do you mean if they have the same name? It'll create two events with the same information besides the duration. What it will happen is:

1. Main deploy starts the timer:
2. The deploy from inside starts the timer
3. The deploy from inside updates the configmap
4. The deploy from inside sends the event with its duration
5. The main deploy updates the configmap
6. The main deploy sends the event with its duration

I think it's the way it must behave but open to discuss it

maroshii (approved):

pkg/cmd/pipeline/translate.go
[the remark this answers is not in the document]
ifbyol: In the moment in which the main deploy add a phase before the "deploy inside" adds its own, the inside one would overwrite the `phase` field of the main one. But not sure if the scenario is possible or not

pkg/repository/repository.go
```diff
@@ -45,10 +45,19 @@ type repositoryURL struct {
 	url.URL
 }
 
-// String is a custom implementation for the url where User is removed
+// String is a custom implementation for the url where User is removed and the schema is forced to https
 func (r repositoryURL) String() string {
 	repo := r.URL
 	repo.User = nil
+
+	switch repo.Scheme {
```
ifbyol, line 53: I don't think we should change this function or at least we should change the name of the function. We are assuming that in all the places where we want the string representation of a repository URL we want to use `https` protocol. In some cases we might just want the URL as specified by the user.

Among other things, this is also being used by the function `GetAnonymizedRepo`, that is being used in different places. Have we double checked that in all the places where that is being used doesn't matter to change the URL scheme?

jLopezbarb: I have double checked. We use `GetAnonymizedRepo` in:
- build command for sending analytics, so with this change we'll be sending always the same scheme for the repo and have better metrics
- pipeline command in order to show a list of the repositories, which I think we should be showing all with the same schema also.
- In our info/debug logs to hide user information
- The new insights package

The only one that I have doubts is with the logs. It's only used in the inferer, but given that the comparison between two repositories URL takes that into account I think we are good to go

ifbyol (approved):

PR #4207, opened 2024-03-19
Don't add extra hosts for remote deploy w/ depot

maroshii:
If the remote build runs in depot, don't add the extra hosts, the certificate, nor the server-name.

Still needs testing but opening early to get feedback.

cc @jpf-okteto to get another set of eyes since he is the most familiar with the certs parts

### DIFF
--- a/cmd/destroy/remote.go
+++ b/cmd/destroy/remote.go
@@ -119,6 +119,8 @@ type remoteDestroyCommand struct {
 
 	// knownHostsPath  is the default known_hosts file path. Provided mostly for testing
 	knownHostsPath string
+
+	useInternalNetwork bool
 }
 
 func newRemoteDestroyer(manifest *model.Manifest, ioCtrl *io.Controller) *remoteDestroyCommand {
@@ -141,6 +143,7 @@ func newRemoteDestroyer(manifest *model.Manifest, ioCtrl *io.Controller) *remote
 		manifest:             manifest,
 		clusterMetadata:      fetchClusterMetadata,
 		ioCtrl:               ioCtrl,
+		useInternalNetwork:   !buildCmd.IsDepotEnabled(),
 	}
 }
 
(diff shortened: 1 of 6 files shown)


pkg/cmd/build/depot_test.go
```diff
@@ -45,41 +45,6 @@ func (m *fakeDepotMachine) Connect(ctx context.Context) (*buildkitClient.Client,
 	return nil, m.err
 }
 
-func Test_isDepotEnabled(t *testing.T) {
```
andreafalzetti, line 48: why removing the test instead of updating it?

maroshii:
> overall lgtm, just take a look to my comment, attempt to test it with a default self signed instance and if I'm correct, restore the certificate injection.

What comment?

maroshii:
Tested that this fixes the issue. Running `okteto deploy|destroy --remote` with depot enabled hangs synchronizing the context. After this change deploy and destroy commands work correctly.

@jpf-okteto added back the injection of the cert regardless of `useInternalNetwork`

pkg/cmd/build/depot_test.go
maroshii: I don't think this test added any value the way it was tbh. Now `IsDepotEnabled` depends on two envvars and adding a test for that seems unnecessary.

ifbyol:
> Tested that this fixes the issue. Running `okteto deploy|destroy --remote` with depot enabled hangs synchronizing the context. After this change deploy and destroy commands work correctly.
> 
> @jpf-okteto added back the injection of the cert regardless of `useInternalNetwork`

But do we have to inject always or only when the CLI is running with `--insecure-skip-tls-verify` flag

pkg/cmd/build/depot_test.go
ifbyol: Nit: the test is not only to verify the current implementation (which I agree is trivial), but also to make sure that next changes in the function don't brake the functionality. 

I could add a change in the function that breaks it without realizing.

Anyways, no strong opinion on keeping or removing the function, as this is a very simple case. It was just a note

ifbyol:
> > But do we have to inject always or only when the CLI is running with --insecure-skip-tls-verify flag
> 
> Before we were injecting always if it was returned from the metadata endpoint. This doesn't change that behavior.

But this was because we always use internal communication. In depot, we will go through the public endpoint, so if the certificate is set at LB level, the certificate we inject here is not the valid one to validate the one in the LB. Is that scenario going to fail (is it a strict validation) or it will try to use it, but if it is not valid, it will try with other certs?

cmd/destroy/remote.go
```diff
@@ -203,22 +206,26 @@ func (rd *remoteDestroyCommand) destroy(ctx context.Context, opts *Options) erro
 		fmt.Sprintf("%s=%s", model.OktetoNamespaceEnvVar, okteto.GetContext().Namespace),
 		fmt.Sprintf("%s=%s", model.OktetoTokenEnvVar, okteto.GetContext().Token),
 		fmt.Sprintf("%s=%s", constants.OktetoTlsCertBase64EnvVar, base64.StdEncoding.EncodeToString(sc.Certificate)),
-		fmt.Sprintf("%s=%s", constants.OktetoInternalServerNameEnvVar, sc.ServerName),
```
ifbyol, line 206: I think having the extra hosts implies that we can get rid off the internal server name flag on deploy and destroy. I think it is out of the scope of this PR but something to tackle soon, mentioning in case this is still relevant and some of the reviewers comments about it

ifbyol (approved):
👏

PR #4211, opened 2024-03-21
DEV-188: Refactor destroy command to follow same pattern used for deploy

ifbyol:
# Proposed changes

Fixes DEV-188

This PR includes the changes to follow the same approach we did for deploy command in [this PR](https://github.com/okteto/okteto/pull/4198). Avoid to execute the same command in remote that is triggering the remote to prevent execute twice logic like manifest parsing and misbehavior due to different environment.

Now, when the destroy command has to be executed remotely, it is executing `remote-run destroy` passing the minimum information needed by the remote to do its job. The only logic which is executed in the remote is the list of custom command, the rest of operations (destroy dependencies, divert, helm releases, k8s resources by label and configmap) are executed locally.

The information passed to the remote are: list of commands to be executed and the information related to external resources. The external resources information is sent to set the default values for the external resources endpoints which were available before because during the manifest parsing (that was being executed in the remote) we were setting them.

Apart of that, I unified the duplicated logic we had in `destroy/remote.go` and `deploy/remote.go` into a single file as almost all the logic in those files was duplicated and it was error prone.

A summary of all the changes in this PR are:
* Created `deploy` and `destroy` subcommands for `remote-run` main command. Former `remote-run` was moved to `deploy` and now `destroy` executes the logic needed for the destroy operation.
* The Dockerfile built when the destroy operation is executed remotely now executes `remote-run destroy` instead of `okteto destroy`
* The variables used during the deployment of the dev environment is passed now as command flag in the `remote-run destroy` command. Previously, the remote was getting them directly from the configmap but I prefered to reduce the dependencies in the remote and avoid the access to the configmap from the remote
* Unified most part of the logic within `deploy/remote.go` and `destroy/remote.go` into a single file in `remote/remote.go`:
  * Now, `deploy/remote.go` has the minimum logic to convert the commands options/flags to the parameters needed for the remote runner and call it handling accordingly any error. 
  * `destroy/remote.go` has the minimum logic to convert the commands options/flags to the parameters needed for the remote runner and call it handling accordingly any error. 
  * `remote/remote.go` has the logic to generate the temporal Dockerfile with the build args and call to the builder to build the image
* Added the env var `OKTETO_IS_PREVIEW_ENVIRONMENT` to replace the former `OKTETO_CURRENT_DEPLOY_BELONGS_TO_PREVIEW` to detect if the commands are being executed within a preview or a dev env. This env var is also now populated to the remote. The backend part is done [here](https://github.com/okteto/app/pull/7904). I just included the changes in this PR as it was trivial after the refactor.

## How to validate

To simplify the review, I tried to do small commits be functionality and adding meaningful commit messages

Tests all the destroy operation of dev enrionments you can think of, paying attention also to the logs and playing with the `--force` flag as the behavior should be different:
* Destroy all
* Local destroy 
* Remote destroy
* Local destroy from an installer
* Remote destroy an installer

## CLI Quality Reminders 🔧


cmd/deploy/remote.go
```diff
@@ -15,111 +15,31 @@ package deploy
 
 import (
 	"context"
-	"crypto/rand"
```
ifbyol, line 18: Most logic of this file was moved to `remote/remote.go` to unify logic with `deploy/remote.go`

cmd/deploy/validate.go
```diff
@@ -14,33 +14,17 @@
...
 func validateAndSet(variables []string, setEnv func(key, value string) error) error {
-	envVars, err := parse(variables)
+	envVars, err := env.Parse(variables)
 	if err != nil {
 		return err
 	}
 	return setOptionVarsAsEnvs(envVars, setEnv)
 }
 
-func parse(variables []string) ([]env.Var, error) {
```
ifbyol, line 31: Extracted to a common place to use it also in the destroy operation

cmd/destroy/all.go
```diff
@@ -22,51 +22,35 @@ import (
...
 	oktetoErrors "github.com/okteto/okteto/pkg/errors"
-	"github.com/okteto/okteto/pkg/k8s/secrets"
 	oktetoLog "github.com/okteto/okteto/pkg/log"
 	"github.com/okteto/okteto/pkg/model"
 	"github.com/okteto/okteto/pkg/okteto"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 type localDestroyAllCommand struct {
-	ConfigMapHandler  configMapHandler
```
ifbyol, line 36: Most of these dependencies were not being used by the `localDestroyAllCommand`. Also, this struct was a dependency of `localDestroyCommand` and these dependencies were being used from there. So I removed the dependencies from here and removed the dependency between `localDestroyAllCommand` and `localDestroyCommand`

cmd/destroy/destroy.go
```diff
@@ -267,63 +282,320 @@ func shouldRunInRemote(opts *Options) bool {
...
+	}
+	destroyer := newLocalDestroyerAll(dc.k8sClientProvider, dc.oktetoClient)
+
+	oktetoLog.Info("Destroying all...")
+
+	return destroyer.destroy(ctx, opts)
+}
+
+// destroy runs the logic needed to destroy a dev environment
+func (dc *destroyCommand) destroy(ctx context.Context, opts *Options) error {
```
ifbyol, line 332: Moved some logic from `local.go` to here, as some of those things have to be executed locally are common to the local and remote operation

cmd/destroy/local.go
```diff
@@ -16,300 +16,75 @@ package destroy
 import (
 	"context"
 	"errors"
-	"fmt"
```
ifbyol, line 19: The logic removed from here was moved to 2 places:
* Most of them were moved to `destroy/destroy.go` as is logic to be executed always locally
* Other part was moved to the `deployable/destroy.go` as that is the logic that has to be executed locally or remotely and it is shared logic

cmd/destroy/remote.go
```diff
@@ -15,253 +15,105 @@ package destroy
 
 import (
 	"context"
-	"crypto/rand"
```
ifbyol, line 18: Most part of this logic was moved to `remote/remote.go`

cmd/destroy/remote.go
```diff
@@ -284,161 +136,26 @@ func (rd *remoteDestroyCommand) destroy(ctx context.Context, opts *Options) erro
...
-			oktetoLog.Infof("invalid okteto CLI version %s. Using latest", versionString)
+			return nil, err
 		}
-		remoteOktetoImage := os.Getenv(constants.OktetoDeployRemoteImage)
-		if remoteOktetoImage != "" {
-			version = remoteOktetoImage
-		} else {
-			version = fmt.Sprintf(constants.OktetoCLIImageForRemoteTemplate, "latest")
+		for _, v := range variables {
+			varsToAddForDeploy = append(varsToAddForDeploy, fmt.Sprintf("--var %s=%q", v.Name, v.Value))
```
ifbyol, line 155: I'm adding the `var` flag to the destroy operation as it is the way we pass the variables to the `remote-run destroy` command

cmd/remoterun/command.go
```diff
@@ -19,137 +19,29 @@ import (
...
-	  namespace: ${OKTETO_NAMESPACE}
-...
-
-It is important that this command does the minimum and must not do calculations that the deploy triggering it already does. For example, this command must not build any image to be used during the deployment
-`,
+		Use:          "remote-run",
+		Short:        "Remote run management commands. These are the commands to be run remotely",
 		Hidden:       true,
 		SilenceUsage: true,
-		RunE: func(cmd *cobra.Command, args []string) error {
```
ifbyol, line 87: All these logic was moved to the subcommand deploy

cmd/remoterun/deploy.go
```diff
@@ -0,0 +1,149 @@
+// Copyright 2024 The Okteto Authors
```
ifbyol, line 1: All this logic was before on `command.go`

cmd/up/up.go
```diff
@@ -613,14 +613,18 @@ func (up *upContext) deployApp(ctx context.Context, ioCtrl *io.Controller, k8slo
 		isRemote = up.Manifest.Deploy.Image != ""
 	}
 
+	// We keep DeprecatedOktetoCurrentDeployBelongsToPreview for backward compatibility in case an old version of the backend
+	// is being used
+	isPreview := os.Getenv(model.DeprecatedOktetoCurrentDeployBelongsToPreview) == "true" ||
+		os.Getenv(constants.OktetoIsPreviewEnvVar) == "true"
```
ifbyol, line 619: This is not related to the destroy refactor per se, it is related with [this ticket](https://okteto.atlassian.net/browse/DEV-245)

pkg/deployable/destroy.go
```diff
@@ -0,0 +1,67 @@
+// Copyright 2024 The Okteto Authors
```
ifbyol, line 1: Common logic used by the local destroy command and the  remote-run destroy to executed the custom commands of deploy operations

pkg/remote/remote.go
```diff
@@ -14,20 +14,415 @@
 package remote
 
 import (
+	"context"
```
ifbyol, line 17: This contains the logic of the former `deploy/remote.go` and `destroy/remote.go` which was duplicated for both scenarios. Now it is common in this place

pkg/remote/remote.go
```diff
@@ -14,20 +14,415 @@
...
 const (
+	// DeployCommand is the command to deploy a dev environment remotely
+	DeployCommand = "deploy"
+	// DestroyCommand is the command to destroy a dev environment remotely
+	DestroyCommand         = "destroy"
 	oktetoDockerignoreName = ".oktetodeployignore"
+	dockerfileTemplate     = `
+FROM {{ .OktetoCLIImage }} as okteto-cli
+
+FROM {{ .UserDeployImage }} as deploy
```
jLopezbarb, line 55:
```suggestion
FROM {{ .UserDeployImage }} as runner
```

jLopezbarb (approved):

jLopezbarb:
Code looks much cleaner and the tests I did were successful. I just have some minor comments

pkg/model/const.go
```diff
@@ -221,9 +221,9 @@ const (
 	// OktetoRescanIntervalEnvVar defines the time between scans for syncthing
 	OktetoRescanIntervalEnvVar = "OKTETO_RESCAN_INTERVAL"
 
-	// OktetoCurrentDeployBelongsToPreview if set the current okteto deploy belongs
+	// DeprecatedOktetoCurrentDeployBelongsToPreview if set the current okteto deploy belongs
 	// to a preview environment
-	OktetoCurrentDeployBelongsToPreview = "OKTETO_CURRENT_DEPLOY_BELONGS_TO_PREVIEW"
+	DeprecatedOktetoCurrentDeployBelongsToPreview = "OKTETO_CURRENT_DEPLOY_BELONGS_TO_PREVIEW"
```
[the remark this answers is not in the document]
ifbyol: You mean at the end of the golang var, right?

pkg/remote/remote.go
ifbyol: Changed on my last commit

pkg/model/const.go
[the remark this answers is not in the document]
ifbyol: Changed on my latest commit

andreafalzetti (approved):

PR #4214, opened 2024-04-01
fix: run remote on buildkit

jLopezbarb:
# Proposed changes

Fixes LAKE-250

This PR modifies the behavior of the Okteto builder in the `pkg/cmd/build/build.go` file. This will ensure that if it's a deploy/destroy and it's running on remote we can't use a depot instance. It'll go to the buildkit instance in the cluster

## How to validate

1. Clone okteto/movies
1. Run `okteto deploy --remote -l info` and check that depot is not mentioned
1. Run `okteto destroy --remote -l info` and check that depot is not mentioned

## CLI Quality Reminders 🔧

### DIFF
--- a/integration/deploy/manifest_test.go
+++ b/integration/deploy/manifest_test.go
@@ -28,6 +28,7 @@ import (
 
 	"github.com/okteto/okteto/integration"
 	"github.com/okteto/okteto/integration/commands"
+	"github.com/okteto/okteto/pkg/cmd/build"
 	"github.com/okteto/okteto/pkg/k8s/kubeconfig"
 	"github.com/okteto/okteto/pkg/okteto"
 	"github.com/okteto/okteto/pkg/registry"
@@ -348,8 +349,8 @@ func TestDeployOktetoManifestExportCache(t *testing.T) {
 
 // TestDeployRemoteOktetoManifest tests the following scenario:
 // - Deploying a okteto manifest in remote with a build locally
+// - Check that is not running on depot
 func TestDeployRemoteOktetoManifest(t *testing.T) {
-	t.Parallel()
 	oktetoPath, err := integration.GetOktetoPath()
 	require.NoError(t, err)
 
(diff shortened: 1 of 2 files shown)


ifbyol (commented):
Could we add an e2e tests for this? I mean, adding a test with the Depot environment variable set and running a deploy and destroy remote.

I'm not sure which check we can do to make sure it is running in BuildKit. One option could be to set fake credentials for Depot (that would make the build in Depot fail) and make sure the result of the deploy and destroy remote is successful

pkg/cmd/build/build.go
```diff
@@ -104,7 +104,7 @@ func (ob *OktetoBuilder) Run(ctx context.Context, buildOptions *types.BuildOptio
 	}
 
 	switch {
-	case IsDepotEnabled():
+	case IsDepotEnabled() && !isDeployOrDestroy:
```
ifbyol, line 107: Let's add a clear comment above this `switch` explaining why for remote operations we don't use Depot. We will forget about it in a few months

integration/deploy/manifest_test.go
```diff
@@ -348,8 +349,8 @@ func TestDeployOktetoManifestExportCache(t *testing.T) {
 
 // TestDeployRemoteOktetoManifest tests the following scenario:
 // - Deploying a okteto manifest in remote with a build locally
+// - Check that is not running on depot
 func TestDeployRemoteOktetoManifest(t *testing.T) {
-	t.Parallel()
```
jLopezbarb, line 352: We need to remove the parallel because the test need to set fake env vars and not affect the rest of the tests

integration/deploy/manifest_test.go
```diff
@@ -370,17 +371,27 @@ func TestDeployRemoteOktetoManifest(t *testing.T) {
 	require.NoError(t, createOktetoManifestWithDeployRemote(dir))
 	require.NoError(t, createAppDockerfileWithCache(dir))
 
-	deployOptions := &commands.DeployOptions{
+	buildOptions := &commands.BuildOptions{
```
jLopezbarb, line 374: Created a previous build in order to not build in the deploy step

pkg/cmd/build/build.go
```diff
@@ -104,6 +104,9 @@ func (ob *OktetoBuilder) Run(ctx context.Context, buildOptions *types.BuildOptio
 	}
 
 	switch {
+	// When depot is available we only go to depot if it's not a deploy or a destroy.
```
ifbyol, line 107: @maroshii be aware of this for the `okteto test`. It should not go to Depot either

ifbyol (approved):

andreafalzetti (approved):

PR #4216, opened 2024-04-03
Initialize test command

maroshii:
Initializes `okteto test` command following current conventions and existing building blocks. There are some fundamental changes I think we need to do but I kept it out of this PR. Examples of this are:

- catching kill signals in every command and not cancelling root context
- cross importing from `cmd/:package`

# How it works

Before running `okteto test` we potentially run an `okteto build` and an `okteto deploy`. We then do a remote run of okteto test.

### Included in this PR

- pre building
- pre deploying
- command initialization
- manifest `test` entry
- variable interpolation in manifest
- remote execution
- Dependency resolution based on depends_on

### Still pending

- It runs remotely as a deploy action and not a test (assuming we want a remote test command?)
- testing

### DIFF
--- a/cmd/deploy/deploy.go
+++ b/cmd/deploy/deploy.go
@@ -73,10 +73,8 @@ type Options struct {
 	Name             string
 	Namespace        string
 	K8sContext       string
-	Repository       string
-	Branch           string
 	Variables        []string
-	servicesToDeploy []string
+	ServicesToDeploy []string
 	Timeout          time.Duration
 	Build            bool
 	Dependencies     bool
@@ -95,7 +93,7 @@ type builderInterface interface {
 type getDeployerFunc func(
 	context.Context, *Options,
 	buildEnvVarsGetter,
-	configMapHandler,
+	ConfigMapHandler,
 	okteto.K8sClientProviderWithLogger,
 	*io.Controller,
 	*io.K8sLogger,
@@ -112,10 +110,10 @@ type Command struct {
 	GetDeployer       getDeployerFunc
 	EndpointGetter    func(k8sLogger *io.K8sLogger) (EndpointGetter, error)
 	DeployWaiter      Waiter
-	CfgMapHandler     configMapHandler
+	CfgMapHandler     ConfigMapHandler
 	Fs                afero.Fs
 	PipelineCMD       pipelineCMD.DeployerInterface
-	AnalyticsTracker  analyticsTrackerInterface
+	AnalyticsTracker  AnalyticsTrackerInterface
 	IoCtrl            *io.Controller
 	K8sLogger         *io.K8sLogger
 
@@ -125,11 +123,11 @@ type Command struct {
 	// This can probably be improved using context cancellation
 	onCleanUp []cleanUpFunc
 
-	isRemote           bool
-	runningInInstaller bool
+	IsRemote           bool
+	RunningInInstaller bool
 }
 
-type analyticsTrackerInterface interface {
+type AnalyticsTrackerInterface interface {
 	TrackDeploy(dm analytics.DeployMetadata)
 	buildTrackerInterface
 }
@@ -146,7 +144,7 @@ type Deployer interface {
 }
 
 // Deploy deploys the okteto manifest
-func Deploy(ctx context.Context, at analyticsTrackerInterface, insightsTracker buildTrackerInterface, ioCtrl *io.Controller, k8sLogger *io.K8sLogger) *cobra.Command {
+func Deploy(ctx context.Context, at AnalyticsTrackerInterface, insightsTracker buildTrackerInterface, ioCtrl *io.Controller, k8sLogger *io.K8sLogger) *cobra.Command {
 	options := &Options{}
 	cmd := &cobra.Command{
 		Use:   "deploy [service...]",
@@ -197,7 +195,7 @@ func Deploy(ctx context.Context, at analyticsTrackerInterface, insightsTracker b
 			}
 
 			options.ShowCTA = oktetoLog.IsInteractive()
-			options.servicesToDeploy = args
+			options.ServicesToDeploy = args
 
 			k8sClientProvider := okteto.NewK8sClientProviderWithLogger(k8sLogger)
 			pc, err := pipelineCMD.NewCommand()
@@ -218,11 +216,11 @@ func Deploy(ctx context.Context, at analyticsTrackerInterface, insightsTracker b
 				Builder:            buildv2.NewBuilderFromScratch(ioCtrl, onBuildFinish),
 				DeployWaiter:       NewDeployWaiter(k8sClientProvider, k8sLogger),
 				EndpointGetter:     NewEndpointGetter,
-				isRemote:           env.LoadBoolean(constants.OktetoDeployRemote),
+				IsRemote:           env.LoadBoolean(constants.OktetoDeployRemote),
 				CfgMapHandler:      NewConfigmapHandler(k8sClientProvider, k8sLogger),
 				Fs:                 afero.NewOsFs(),
 				PipelineCMD:        pc,
-				runningInInstaller: config.RunningInInstaller(),
+				RunningInInstaller: config.RunningInInstaller(),
 				AnalyticsTracker:   at,
 				IoCtrl:             ioCtrl,
 				K8sLogger:          k8sLogger,
@@ -288,7 +286,7 @@ func (dc *Command) Run(ctx context.Context, deployOptions *Options) error {
 		return oktetoErrors.ErrManifestFoundButNoDeployAndDependenciesCommands
 	}
 
-	if len(deployOptions.servicesToDeploy) > 0 && deployOptions.Manifest.Deploy != nil && deployOptions.Manifest.Deploy.ComposeSection == nil {
+	if len(deployOptions.ServicesToDeploy) > 0 && deployOptions.Manifest.Deploy != nil && deployOptions.Manifest.Deploy.ComposeSection == nil {
 		return oktetoErrors.ErrDeployCantDeploySvcsIfNotCompose
 	}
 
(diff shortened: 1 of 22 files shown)


maroshii:
To test, add this to the go-getting-started guide `okteto.yml` file:

```yaml
dependencies:
  python:
    repository: https://github.com/okteto/python-getting-started
    variables:
      MYVAR: yeah
    wait: true

build:
  hello-world:
    context: .

test:
  unit:
    image: ${OKTETO_BUILD_HELLO_WORLD_IMAGE}
    command: 
      - "echo 'Running tests using ${OKTETO_BUILD_HELLO_WORLD_IMAGE}'"
      - "echo 'depends_on python with var ${OKTETO_DEPENDENCY_PYTHON_VARIABLE_MYVAR}'"

```

cmd/test/cmd.go
```diff
@@ -0,0 +1,251 @@
...
+		TemplateName:        "dockerfile",
+		CommandFlags:        commandsFlags,
+		BuildEnvVars:        builder.GetBuildEnvVars(),
+		DependenciesEnvVars: deployCMD.GetDependencyEnvVars(os.Environ),
+		DockerfileName:      "Dockerfile.test",
+		Deployable: deployable.Entity{
+			Commands: commands,
+		},
+		Manifest: manifest,
+		Command:  remote.DeployCommand, // TODO: use test?
```
ifbyol, line 249: As discussed in the daily, maybe it worths to create a specific command for test. The `remote-run deploy` does more things than just execute the list of commands:
* It starts the local proxy to forward all kubernetes calls
* It deploys external resources
* It deploys divert information 

So, it might print misleading logs in that regard or have a side effect

In fact, `destroy` is much more similar to `test` than `deploy`. We might even change the name of the remote commands to reuse destroy and test. But given pst experiences, I would have them separated (having main common logic)

pkg/model/test.go
```diff
@@ -0,0 +1,40 @@
...
+	var err error
+	if len(test.Image) > 0 {
+		test.Image, err = env.ExpandEnv(test.Image)
+		if err != nil {
+			return err
+		}
+	}
+	if len(test.Command) > 0 {
+		for i, v := range test.Command {
+			test.Command[i], err = env.ExpandEnv(v)
```
ifbyol, line 32: I think we don't have to expand variables of the commands, right? That should be expanded on runtime when commands are executed, right?

ifbyol (commented):
Maybe not for this initialization of the command, but we should probably review which flags of the Okteto command have to be exposed in the test command. 

The same we are not exposing all the flags of the `deploy` command in the `up` (when executing `okteto up --deploy`) or all the flags of the `build` command for the `deploy` command (when running `okteto deploy --build`), we might want to be consistent and not expose all of them and rely on default values.

Other cases to consider:
* Should the dev env name be inferred as we do in other commands if it is not provided?

cc @pchico83 as I think these things are not clear in the spec

cmd/test/cmd.go
```diff
@@ -0,0 +1,251 @@
...
+	cmd.Flags().StringArrayVarP(&options.Variables, "var", "v", []string{}, "set a variable (can be set more than once)")
+	cmd.Flags().DurationVarP(&options.Timeout, "timeout", "t", getDefaultTimeout(), "the length of time to wait for completion, zero means never. Any other values should contain a corresponding time unit e.g. 1s, 2m, 3h ")
+	cmd.Flags().BoolVarP(&options.Wait, "wait", "w", false, "wait until the development environment is deployed (defaults to false)")
+	cmd.Flags().BoolVarP(&options.Dependencies, "dependencies", "", false, "deploy the dependencies from manifest")
+
+	cmd.Flags().StringVar(&options.DeployName, "deploy-name", "", "name of the development environment name to be deployed")
+	cmd.Flags().StringArrayVar(&options.DeployServices, "deploy-services", []string{}, "name of the services of the development environment that should be deployed")
+	cmd.Flags().BoolVarP(&options.DeployWithoutBash, "deploy-no-bash", "", false, "execute commands without bash")
+	cmd.Flags().BoolVarP(&options.DeployInRemote, "deploy-remote", "", false, "force run deploy commands in remote")
```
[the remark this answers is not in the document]
maroshii: In general I agree with the comments about the deploy flags for this command. My thinking was using this command command for deploys and apps that are already dependent on this and are already deployed using these flags and would fail otherwise. The more we can remove, the better

cmd/test/cmd.go
```diff
@@ -0,0 +1,251 @@
...
+	cmd.Flags().StringVarP(&options.ManifestPath, "file", "f", "", "path to the okteto manifest file")
+	cmd.Flags().StringVarP(&options.Namespace, "namespace", "n", "", "overwrites the namespace where the development environment is deployed")
+	cmd.Flags().StringVarP(&options.K8sContext, "context", "c", "", "context where the development environment is deployed")
+	cmd.Flags().BoolVarP(&options.Build, "build", "", false, "force build of images before running the tests")
+	cmd.Flags().StringArrayVarP(&options.Variables, "var", "v", []string{}, "set a variable (can be set more than once)")
+	cmd.Flags().DurationVarP(&options.Timeout, "timeout", "t", getDefaultTimeout(), "the length of time to wait for completion, zero means never. Any other values should contain a corresponding time unit e.g. 1s, 2m, 3h ")
+	cmd.Flags().BoolVarP(&options.Wait, "wait", "w", false, "wait until the development environment is deployed (defaults to false)")
+	cmd.Flags().BoolVarP(&options.Dependencies, "dependencies", "", false, "deploy the dependencies from manifest")
+
+	cmd.Flags().StringVar(&options.DeployName, "deploy-name", "", "name of the development environment name to be deployed")
```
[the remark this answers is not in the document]
maroshii: `name` would imply it is for the test not for the deploy. This is the name of the deploy (not the test). The name of the test is the key in the map

cmd/test/cmd.go
```diff
@@ -0,0 +1,251 @@
...
+	oktetoErrors "github.com/okteto/okteto/pkg/errors"
+	oktetoLog "github.com/okteto/okteto/pkg/log"
+	"github.com/okteto/okteto/pkg/log/io"
+	"github.com/okteto/okteto/pkg/model"
+	"github.com/okteto/okteto/pkg/okteto"
+	oktetoPath "github.com/okteto/okteto/pkg/path"
+	"github.com/okteto/okteto/pkg/remote"
+)
+
+type Options struct {
```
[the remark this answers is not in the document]
maroshii: Yeah, I'll leave this for a future iteration

pkg/model/test.go
maroshii: Removed

cmd/test/cmd.go
maroshii: I'll do this in a follow-up. To avoid this PR getting too big.

maroshii:
Addressed most of the feedback specially around flags for the command. We might need to follow up and discuss these little details.

### Changes added after review (included in this PR)

- dependency resolution based on depends_ok and running multiple tests
- extended command format for test
- Remove unnecessary flags

### Still pending (for follow up PRs)

- inferring `name` (of the deploy or of the test?)
- New test command in the remote runner
- `--no-cache` flag
- Define flags provided to the deploy action

cmd/test/cmd.go
```diff
@@ -0,0 +1,260 @@
...
+				return err
+			}
+
+		},
+	}
+
+	cmd.Flags().StringVarP(&options.ManifestPath, "file", "f", "", "path to the okteto manifest file")
+	cmd.Flags().StringVarP(&options.Namespace, "namespace", "n", "", "overwrites the namespace where the development environment is deployed")
+	cmd.Flags().StringVarP(&options.K8sContext, "context", "c", "", "context where the development environment is deployed")
+	cmd.Flags().BoolVarP(&options.Build, "build", "", false, "force build of images before running the tests")
```
pchico83, line 92: Let's not support `--build` for now

cmd/test/cmd.go
```diff
@@ -0,0 +1,260 @@
...
+	cmd.Flags().StringArrayVarP(&options.Variables, "var", "v", []string{}, "set a variable (can be set more than once)")
+	cmd.Flags().DurationVarP(&options.Timeout, "timeout", "t", getDefaultTimeout(), "the length of time to wait for completion, zero means never. Any other values should contain a corresponding time unit e.g. 1s, 2m, 3h ")
```
pchico83, line 94: This would be the length of the test execution, right?

cmd/test/cmd.go
```diff
@@ -0,0 +1,251 @@
...
+	cmd.Flags().BoolVarP(&options.Wait, "wait", "w", false, "wait until the development environment is deployed (defaults to false)")
```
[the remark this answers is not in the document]
pchico83: Let's always wait and remove this flag for now. It's easier to add flags in the future than removing things

cmd/test/cmd.go
```diff
@@ -0,0 +1,260 @@
...
+	cmd.Flags().BoolVarP(&options.Dependencies, "dependencies", "", false, "deploy the dependencies from manifest")
```
pchico83, line 96: Let's remove this for now. Only if the deployment doesn't exist, we deploy it (similar to current `okteto up` behavior

cmd/test/cmd.go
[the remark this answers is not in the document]
pchico83: I think it should be `--name` for consistency. This is the name of the development environment you want to test. The name of the test will be a positional arg in the `test` command

cmd/deploy/deploy.go
```diff
@@ -73,10 +73,8 @@ type Options struct {
 	Name             string
 	Namespace        string
 	K8sContext       string
-	Repository       string
-	Branch           string
 	Variables        []string
-	servicesToDeploy []string
+	ServicesToDeploy []string
```
ifbyol, line 77: I think we could revert this. It wouldn't be needed anymore as we don't include the flag in the test command anymore

ifbyol (approved):
Minor comment, but it looks good for a first initialization of the command. I approve the PR but review comments from Pablo

cmd/test/cmd.go
maroshii: Yes. Including a potential deploy

cmd/deploy/deploy.go
maroshii: This would make `servicesToDeploy` the only private member. Based on current feedback we could also make `Wait` and `Dependencies` also private but that would leave us with some private options and some public options driven by external requirements. Given that `Options` is public, I would keep every member public for simplicity.

PR #4219, opened 2024-04-04
DEV-258: Changed command used to get the latest commit for a folder

ifbyol:
# Proposed changes

Fixes DEV-258

As the installer is cloning repositories always using only one commit, the build context calculation doesn't work well because you have different behaviors from local and the installer, generating unnecessary builds. As cloning the whole history might impact performance, we are changing the CLI to not use the latest commit of the folder but instead, use a hash of the files tracked by git within that folder. 

The command to be used is `/usr/bin/git --no-optional-locks ls-files -s . | /usr/bin/git --no-optional-locks hash-object --stdin`

## How to validate

1. Verify general behavior of Smart Builds. Check that cache is being use correctly when it has to

## CLI Quality Reminders 🔧

## Tests done

* I tested that a simple commit in a repo without affecting to none of the build context don't trigger a build of any image in the installer if images were already build
* I tested that a change in a single build context only triggers the rebuild of that image within the installer
* I was testing that it works on windows machines
* I was testing also that locally, smart builds keeps working as expected

### DIFF
--- a/pkg/repository/local_git.go
+++ b/pkg/repository/local_git.go
@@ -229,7 +229,7 @@ func (lg *LocalGit) GetDirContentSHA(ctx context.Context, gitPath, dirPath strin
 	}
 
 	lsFilesCmdArgs := []string{"--no-optional-locks", "ls-files", "-s", dirPath}
-	hashObjectCmdArgs := []string{"--no-optional-locks", "hash-objectasd", "--stdin"}
+	hashObjectCmdArgs := []string{"--no-optional-locks", "hash-object", "--stdin"}
 
 	output, err := lg.exec.RunPipeCommands(ctx, gitPath, lg.gitPath, lsFilesCmdArgs, lg.gitPath, hashObjectCmdArgs)
 	if err != nil {
--- a/pkg/repository/local_git.go
+++ b/pkg/repository/local_git.go
@@ -14,8 +14,10 @@
 package repository
 
 import (
+	"bytes"
 	"context"
 	"errors"
+	"io"
 	"os"
 	"os/exec"
 	"runtime"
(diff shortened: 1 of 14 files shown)


pkg/repository/local_git.go
```diff
@@ -36,13 +38,60 @@ var (
...
+// RunPipeCommands runs two commands in a pipeline. cmd1 | cmd2. Example:
+// /usr/bin/git --no-optional-locks ls-files -s . | /usr/bin/git --no-optional-locks hash-object --stdin
+func (le *LocalExec) RunPipeCommands(ctx context.Context, dir string, cmd1 string, cmd1Args []string, cmd2 string, cmd2Args []string) ([]byte, error) {
+	c1 := le.createCommand(ctx, dir, cmd1, cmd1Args...)
+	c2 := le.createCommand(ctx, dir, cmd2, cmd2Args...)
+
+	// Create a pipe to connect the stdout of c1 to the stdin of c2
+	pipeReader, pipeWriter := io.Pipe()
+	c1.Stdout = pipeWriter
+	c2.Stdin = pipeReader
```
maroshii, line 66: It might be useful to set  `c1.Stderr = os.Stderr` (by default it goes to `/dev/null` I think)

maroshii (approved):

pkg/repository/local_git.go
ifbyol: I have changed it to get the stderr to a buffer and log it if there is any error on any of the commands. @maroshii WDYT?

maroshii: 👌

maroshii (approved):

andreafalzetti (approved):

PR #4220, opened 2024-04-04
Init dag package

maroshii:
Initializes a `pkg/dag` package responsible for resolving order of execution of a set of objects. This will be necessary for resolving `depends_on` field in `okteto test`.

For more info see and docs see [pkg/dag/doc.go](https://github.com/okteto/okteto/pull/4220/files#diff-4fd62b6ae63a25ff3c69c048e3c62d03faa7ab2886048e248dd4a4909c86a6ac) file below.


ifbyol (approved):
Really useful! At some point we could add this also in the backend as there are some places where we do something similar 👏

PR #4225, opened 2024-04-10
chore: deprecation warning for manifest context & namespace

andreafalzetti:
# Proposed changes

Related: https://okteto.atlassian.net/browse/DEV-278

If the user specifies `context` or `namespace` in the Okteto maniest, we show a deprecation warning and suggest an alternative approach.

Screenshot 2024-04-10 at 17 32 07

## How to validate

Set the following fields in the `okteto.yml` and run any command (i.e. `okteto deploy`)

```
context: <CONTEXT URL>
namespace: <NS>
```

### Scenario Compose

1. Using https://github.com/okteto/movies-with-compose
1. Run `okteto deploy` and check that no warning is displayed
1. Add the `context` property in the `docker-compose.yml` and run `okteto deploy` again
1. Observe now the warning. Repeat the same for `namespace` (both should trigger the warning)
1. Make sure the warning is only visible once

Note: the following scenario does not display any warning because `context` and `namespace` are ignored in this case:
1. Rename your `docker-compose.yml` to `test.yml` and create an `okteto.yml` with the following content:
    ```
    deploy:
      compose: test.yml
    ```
1. Run `okteto deploy` and observe the warning is NOT being displayed and the context hardcoded in the compose file, it's not used.

### Scenario Deploy

1. Using https://github.com/okteto/go-getting-started
1. Run `okteto deploy` and observe no warnings
1. Run `okteto deploy --context <CONTEXT> --namespace <NS>` and observe no warnings
1. Now edit the `okteto.yml` and specify a `context` and/or `namespace` (test both combinations together and alone)
1. Run `okteto deploy` and observe warnings
1. Now run `okteto deploy --context <CONTEXT> --namespace <NS>` and observe warnings
1. Make sure the warnings are only printed once

Additionally, check with `okteto deploy --remote` to make sure the warning appears once.

### Scenario Up + Deploy

1. Using https://github.com/okteto/go-getting-started
1. Run `okteto up` and observe no warnings
1. Run `okteto up --context <CONTEXT> --namespace <NS>` and observe no warnings
1. Now edit the `okteto.yml` and specify a `context` and/or `namespace` (test both combinations together and alone)
1. Run `okteto up` and observe warnings
1. Now run `okteto up --context <CONTEXT> --namespace <NS>` and observe warnings
1. Make sure the warnings are only printed once

### Scenario Destroy

1. Using https://github.com/okteto/go-getting-started
1. Run `okteto destroy` and observe no warnings
1. Run `okteto destroy --context <CONTEXT> --namespace <NS>` and observe no warnings
1. Now edit the `okteto.yml` and specify a `context` and/or `namespace` (test both combinations together and alone)
1. Run `okteto destroy` and observe warnings
1. Now run `okteto destroy --context <CONTEXT> --namespace <NS>` and observe warnings
1. Make sure the warnings are only printed once

## CLI Quality Reminders 🔧

### DIFF
--- a/cmd/deploy/deploy.go
+++ b/cmd/deploy/deploy.go
@@ -333,6 +333,12 @@ func (dc *Command) Run(ctx context.Context, deployOptions *Options) error {
 		return err
 	}
 
+	if manifest != nil {
+		if manifest.Context != "" || manifest.Namespace != "" {
+			oktetoLog.Warning("Compose fields: 'context' and 'namespace' are deprecated. Use 'OKTETO_CONTEXT' and 'OKTETO_NAMESPACE' env vars instead.")
+		}
+	}
+
 	if dc.IsRemote || dc.RunningInInstaller {
 		currentVars, err := dc.CfgMapHandler.GetConfigmapVariablesEncoded(ctx, deployOptions.Name, deployOptions.Manifest.Namespace)
 		if err != nil {
--- a/cmd/deploy/deploy.go
+++ b/cmd/deploy/deploy.go
@@ -333,12 +333,6 @@ func (dc *Command) Run(ctx context.Context, deployOptions *Options) error {
 		return err
 	}
 
-	if manifest != nil {
-		if manifest.Context != "" || manifest.Namespace != "" {
-			oktetoLog.Warning("Compose fields: 'context' and 'namespace' are deprecated. Use 'OKTETO_CONTEXT' and 'OKTETO_NAMESPACE' env vars instead.")
-		}
-	}
-
 	if dc.IsRemote || dc.RunningInInstaller {
 		currentVars, err := dc.CfgMapHandler.GetConfigmapVariablesEncoded(ctx, deployOptions.Name, deployOptions.Manifest.Namespace)
 		if err != nil {
--- a/pkg/model/contextresource.go
+++ b/pkg/model/contextresource.go
@@ -14,20 +14,41 @@
 package model
 
 import (
-	"os"
-
+	"fmt"
 	"github.com/okteto/okteto/pkg/discovery"
 	oktetoErrors "github.com/okteto/okteto/pkg/errors"
 	"github.com/okteto/okteto/pkg/filesystem"
 	oktetoLog "github.com/okteto/okteto/pkg/log"
 	"github.com/spf13/afero"
 	yaml "gopkg.in/yaml.v3"
+	"os"
 )
 
 // ContextResource provides the context and namespace to operate within a manifest
 type ContextResource struct {
 	Context   string `yaml:"context"`
 	Namespace string `yaml:"namespace"`
+}
+
+// TODO: remove this function once we remove the support for 'context' and 'namespace' in the okteto manifest and docker compose
+func displayDeprecationWarningForContextAndNamespace(context, namespace string) {
+	var s1, s2 string
+
+	if context == "" && namespace == "" {
+		return
+	} else if context != "" && namespace != "" {
+		s1 = "'context' and 'namespace'"
+		s2 = fmt.Sprintf("'%s' and '%s'", OktetoContextEnvVar, OktetoNamespaceEnvVar)
+	} else if context != "" {
+		s1 = "'context'"
+		s2 = fmt.Sprintf("'%s'", OktetoContextEnvVar)
+	} else if namespace != "" {
+		s1 = "'namespace'"
+		s2 = fmt.Sprintf("'%s'", OktetoNamespaceEnvVar)
+	}
+
+	msg := "Setting %s in the okteto manifest or docker compose is deprecated. We recommend using the env var %s instead."
+	oktetoLog.Warning(fmt.Sprintf(msg, s1, s2))
 }
 
 // GetContextResource returns a ContextResource object from a given file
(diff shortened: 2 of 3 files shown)


ifbyol:
Could we check we only display once this message in case we read more than once the manifest? I think we had issues in the past displaying twice the same message because of that

jLopezbarb:
Should we also deprecate it for the compose serialisation and the in the dev section?
 I'd also recommend the `okteto context` and `okteto namespace` commands instead of the environment variables because if you set them and forget, it's going to be difficult to know when you change the namespace with the command and still use the one in the variable. WDYT?

andreafalzetti:
> Should we also deprecate it for the compose serialisation and the in the dev section? I'd also recommend the `okteto context` and `okteto namespace` commands instead of the environment variables because if you set them and forget, it's going to be difficult to know when you change the namespace with the command and still use the one in the variable. WDYT?

I've added it for compose, good shout. In regards with the message, I think the use case here is for when someone wants to hardcode those 2 vars, and are currently doing it in the manifest, but the idea is to suggest the `.env` in the future. Or env var in general for such use case. Generally speaking, I agree with you that we should suggest using `okteto context` etc.

andreafalzetti:
> Could we check we only display once this message in case we read more than once the manifest? I think we had issues in the past displaying twice the same message because of that

I couldn't find a way to make it display twice

ifbyol (changes requested):
I have found 2 different "bugs".

* I was able to see the message twice. I have the context set in the manifest and when I execute `okteto deploy`, it appears twice:

Screenshot 2024-04-16 at 10 05 39

This is the content of the Okteto manifest:

```
context: https://okteto.xxxxxxx

build:
  hello-world:
    context: .

deploy:
  - kubectl apply -f k8s.yml

dev:
  hello-world:
    image: okteto/golang:1
    command: bash
    sync:
      - .:/usr/src/app
    remote: 2222
    volumes:
      - /go
      - /root/.cache
    securityContext:
      capabilities:
        add:
          - SYS_PTRACE
    forward:
      - 2345:2345
```

* I'm getting the warning just because I specified the namespace as command flag, but I don't have `context` nor `namespace` in the manifest:

Screenshot 2024-04-16 at 10 08 14

This is the content of the okteto manifest:

```
build:
  hello-world:
    context: .

deploy:
  - kubectl apply -f k8s.yml

dev:
  hello-world:
    image: okteto/golang:1
    command: bash
    sync:
      - .:/usr/src/app
    remote: 2222
    volumes:
      - /go
      - /root/.cache
    securityContext:
      capabilities:
        add:
          - SYS_PTRACE
    forward:
      - 2345:2345
```

Also, the warning about Compose fields is being displayed in both scenarios but this case doesn't have a compose file. It is the go-getting-started example

cmd/deploy/deploy.go
```diff
@@ -317,6 +317,12 @@ func (dc *Command) Run(ctx context.Context, deployOptions *Options) error {
 		return err
 	}
 
+	if manifest != nil {
+		if manifest.Context != "" || manifest.Namespace != "" {
+			oktetoLog.Warning("Compose fields: 'context' and 'namespace' are deprecated. Use 'OKTETO_CONTEXT' and 'OKTETO_NAMESPACE' env vars instead.")
```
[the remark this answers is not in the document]
andreafalzetti: I've updated the code and it should be better now

ifbyol:
@andreafalzetti  I was able to get the message twice when having defined the namespace within the docker-compose with the `movies-with-compose` repository. Also, it is misleading because I don't have an Okteto manifest, but I'm getting the `Manifest fields:` message

Screenshot 2024-04-18 at 13 54 34

pkg/model/contextresource.go
```diff
@@ -50,6 +51,10 @@ func GetContextResource(path string) (*ContextResource, error) {
 		return nil, newManifestFriendlyError(err)
 	}
 
+	if ctxResource.Context != "" || ctxResource.Namespace != "" {
```
jLopezbarb, line 54: NIT: It's weird to me that I set both of the fields written even when I just wrote one of them in the manifest.

jLopezbarb (approved):
Tested and working as expected. Only one NIT comment

pkg/model/contextresource.go
andreafalzetti: @jLopezbarb I initially preferred to keep it simple, but you are right, it can be confusing if someone receives a warning of something that they are NOT doing. I've changed the code in 59a654ea963fe0ab36e0d251cda48bb772ff120f

Please have another look 🙏

maroshii (approved):

andreafalzetti:
@ifbyol the request for changes is preventing me to merge, can you have a look please?

ifbyol (approved):
In the spec it says that we should show a warning to encourage them to move it to the `.env` file. Not sure if that changed in offline conversations

>  If context or namespace are set in the Manifest, we will show a warning to move them to the .env file. 

In terms of functionality looks good, I'm approving the PR but it would be good to review the message we want to display, but I don't think is a blocker to merge this

andreafalzetti:
> In the spec it says that we should show a warning to encourage them to move it to the `.env` file. Not sure if that changed in offline conversations
> 
> > If context or namespace are set in the Manifest, we will show a warning to move them to the .env file.
> 
> In terms of functionality looks good, I'm approving the PR but it would be good to review the message we want to display, but I don't think is a blocker to merge this

@ifbyol I think it would be great to simply link to docs instead of explaining the `.env` because it could lead to further questions the warning. I kept it generic on purpose for now, so that users can export env vars (either manually or using the `.env`). 

I am adding a follow-up item to DEV-282 so that when we have updated the docs, we can add the link here

ifbyol:
> > In the spec it says that we should show a warning to encourage them to move it to the `.env` file. Not sure if that changed in offline conversations
> > > If context or namespace are set in the Manifest, we will show a warning to move them to the .env file.
> > 
> > 
> > In terms of functionality looks good, I'm approving the PR but it would be good to review the message we want to display, but I don't think is a blocker to merge this
> 
> @ifbyol I think it would be great to simply link to docs instead of explaining the `.env` because it could lead to further questions the warning. I kept it generic on purpose for now, so that users can export env vars (either manually or using the `.env`).
> 
> I am adding a follow-up item to [DEV-282](https://okteto.atlassian.net/browse/DEV-282) so that when we have updated the docs, we can add the link here

I think this is something to be decided by product, with our feedback/opinion. But product defines the experience

PR #4227, opened 2024-04-10
Add dedicated test command for remote run

maroshii:
Adds a dedicated `test` command for the remote run command

Fixes https://okteto.atlassian.net/browse/DEV-272

### DIFF
--- a/cmd/test/cmd.go
+++ b/cmd/test/cmd.go
@@ -235,9 +235,12 @@ func doRun(ctx context.Context, options *Options, ioCtrl *io.Controller, k8sLogg
 			DockerfileName:      "Dockerfile.test",
 			Deployable: deployable.Entity{
 				Commands: commands,
+				// Added this for backward compatibility. Before the refactor we were having the env variables for the external
+				// resources in the environment, so including it to set the env vars in the remote-run
+				External: manifest.External,
 			},
 			Manifest: manifest,
-			Command:  remote.DeployCommand, // TODO: use test?
+			Command:  remote.TestCommand,
 		}
 
 		ioCtrl.Logger().Infof("Executing test for: %s", name)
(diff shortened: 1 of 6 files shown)


cmd/test/cmd.go
```diff
@@ -235,9 +235,12 @@ func doRun(ctx context.Context, options *Options, ioCtrl *io.Controller, k8sLogg
 			DockerfileName:      "Dockerfile.test",
 			Deployable: deployable.Entity{
 				Commands: commands,
+				// Added this for backward compatibility. Before the refactor we were having the env variables for the external
+				// resources in the environment, so including it to set the env vars in the remote-run
+				External: manifest.External,
```
ifbyol, line 240: We will need to see what is really needed here. In the case of `destroy`, I added it because those variables were available before, but they only have the default value. If during the deploy, some value was generated, they are not available, but should they?

Example of dynamic generation of external resources variables: https://github.com/ifbyol/node-compose-getting-started/blob/ifbyol/complete-manifest/okteto.yml#L48. Should they be available in test execution even if they were generated during deploy phase?

ifbyol (approved):
In general looks good. The only thing, we need to clarify what is really needed and how it should behave regarding external resources and the variables, but they can be addressed in a following PR. Things to clarify:
* Does the test remote have to deploy external resources to apply any potential change in endpoints or similar?
* Should the variables for external resources be available in the test command, including dynamic values, for deploy executed even outside of okteto test?

cmd/test/cmd.go
maroshii: That's a good point. +1 for removing since there's no need for backwards compatibility. @pchico83 thoughts?

maroshii: I'll merge to open a the next PR in the queue. I'll remove this is a follow up we if need be

PR #4231, opened 2024-04-11
fix: skip volume mount translation into okteto manifest build

jLopezbarb:
# Proposed changes

If you try to run okteto build on a compose file that only have volume mounts We try to build the services but given that we've removed the build support for the volume mounts, we are not building anything which results into a weird case where the user get that it's going to build if it's not a git repo or no other output:
Screenshot 2024-04-11 at 16 58 50

The correct behaviour must be to fallback to the v1 given that we don't have any service to build.

## How to validate

1. Create the following `docker-compose.yml`:
```yaml
version: '2'
services:
  postgres:
    image: bitnami/postgresql:11.14.0-debian-10-r28
    environment:
      POSTGRES_USER: test
      POSTGRES_PASSWORD: test
      POSTGRES_DB: test
    volumes:
      - ./schema.sql:/test/schema.sql
```
2. Run `okteto build`
3. Check that it's fullbacking to the build v1

## CLI Quality Reminders 🔧

### DIFF
--- a/pkg/model/manifest.go
+++ b/pkg/model/manifest.go
@@ -1004,29 +1004,14 @@ func (m *Manifest) InferFromStack(cwd string) (*Manifest, error) {
 			m.Dev[svcName] = d
 		}
 
-		if svcInfo.Build == nil && len(svcInfo.VolumeMounts) == 0 {
+		if svcInfo.Build == nil {
 			continue
 		}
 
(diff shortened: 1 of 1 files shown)


ifbyol (approved):
Some CI job has failed but I think it is some transient error with an external service. It looks good!

PR #4232, opened 2024-04-11
[okteto test] Don't deploy if app already exists

maroshii:
Skip deploy in `okteto test` if the deploy already exists. Fixes https://okteto.atlassian.net/browse/DEV-271

It required some minor refactor to be able to use the same logic than the deploy command

### DIFF
--- a/cmd/deploy/command_configuration.go
+++ b/cmd/deploy/command_configuration.go
@@ -75,7 +75,10 @@ func setDeployOptionsValuesFromManifest(ctx context.Context, deployOptions *Opti
 			ManifestName: deployOptions.Manifest.Name,
 			ManifestPath: deployOptions.ManifestPathFlag,
 		}
-		deployOptions.Name = n.ResolveName(ctx)
+		name := n.ResolveName(ctx)
+		deployOptions.Name = name
+		deployOptions.Manifest.Name = name
+
 	} else {
 		if deployOptions.Manifest != nil {
 			deployOptions.Manifest.Name = deployOptions.Name
--- a/cmd/deploy/command_configuration.go
+++ b/cmd/deploy/command_configuration.go
@@ -39,6 +39,21 @@ const (
 	httpsScheme = "https"
 )
 
+type Namer struct {
+	Workdir      string
+	KubeClient   kubernetes.Interface
+	ManifestName string
+	ManifestPath string
+}
+
+func (na Namer) ResolveName(ctx context.Context) string {
+	if na.ManifestName != "" {
+		return na.ManifestName
+	}
+	inferer := devenvironment.NewNameInferer(na.KubeClient)
+	return inferer.InferName(ctx, na.Workdir, okteto.GetContext().Namespace, na.ManifestPath)
+}
+
 func setDeployOptionsValuesFromManifest(ctx context.Context, deployOptions *Options, cwd string, c kubernetes.Interface, k8sLogger *ioCtrl.K8sLogger) error {
 
 	if deployOptions.Manifest.Context == "" {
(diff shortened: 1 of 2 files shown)


cmd/deploy/command_configuration.go
```diff
@@ -49,18 +64,18 @@ func setDeployOptionsValuesFromManifest(ctx context.Context, deployOptions *Opti
...
+			return err
 		}
 
+		n := Namer{
+			Workdir:      cwd,
+			KubeClient:   c,
+			ManifestName: deployOptions.Manifest.Name,
+			ManifestPath: deployOptions.ManifestPathFlag,
+		}
+		deployOptions.Name = n.ResolveName(ctx)
```
ifbyol, line 78: We should update also `deployOptions.Manifest.Name` as we were doing before in line 61.

cmd/deploy/command_configuration.go
```diff
@@ -39,6 +39,21 @@ const (
 	httpsScheme = "https"
 )
 
+type Namer struct {
```
ifbyol, line 42: Not for this PR, but maybe this can be moved to another common package. This can probably be used in other commands and we would simplify the dependency between test and deploy CMD. Just a though

cmd/test/cmd.go
```diff
@@ -156,42 +157,73 @@ func doRun(ctx context.Context, options *Options, ioCtrl *io.Controller, k8sLogg
...
+			ShowCTA:          false,
+		}); err != nil {
+			oktetoLog.Error("deploy failed: %s", err.Error())
+			return err
+		}
+	} else {
+		// The deploy operation expands environment variables in the manifest. If
+		// we don't deploy, make sure to expand the envvars
+		if err := manifest.ExpandEnvVars(); err != nil {
+			return fmt.Errorf("failed to expand manifest environment variables.: %w", err)
```
ifbyol, line 224: Nit
```suggestion
			return fmt.Errorf("failed to expand manifest environment variables: %w", err)
```

cmd/test/cmd.go
```diff
@@ -156,42 +157,73 @@ func doRun(ctx context.Context, options *Options, ioCtrl *io.Controller, k8sLogg
...
+		}
+		oktetoLog.Information("'%s' was already deployed. To redeploy run 'okteto deploy'", name)
```
ifbyol, line 226: Should we include also `okteto test --deploy` as we do with `okteto up` [here](https://github.com/okteto/okteto/blob/07bd7bb3505aeea89c6735e5f07a7d5552162d69/cmd/up/up.go#L307)?

cmd/test/cmd.go
```diff
@@ -156,42 +157,73 @@ func doRun(ctx context.Context, options *Options, ioCtrl *io.Controller, k8sLogg
...
-		oktetoLog.Error("deploy failed: %s", err.Error())
-		return err
+	kubeClient, _, err := k8sClientProvider.Provide(okteto.GetContext().Cfg)
+	if err != nil {
+		return fmt.Errorf("could not instantiate kuberentes client: %w", err)
+	}
+
+	namer := deployCMD.Namer{
+		KubeClient:   kubeClient,
+		Workdir:      initialCWD,
```
ifbyol, line 167: Not sure we should use `initialCWD`. That is calculated before the working directory might change in line `137` (internally, `UpdateCWDtoManifestPath` changes the current working directory). In the case of deploy, we get the current working directory after changing it

cmd/test/cmd.go
maroshii: For now that flag doesn't exist. We should probably add it but I'll leave that for a future PR

ifbyol (approved):

PR #4244, opened 2024-04-15
Use DeployCommand for remoterun test

maroshii:
Use `DeployCommand` internally for `okteto test`.

There's a lot of logic that's already handled by `DeployCommand` (like envvar expansion) that we need as part of the test command that we lost in https://github.com/okteto/okteto/pull/4227 when we created a separate runner.

In this PR we keep the extra command but we move the internal implementation to use `DeployCommand`.

Before this change `$MYENV` would be empty (fixed in this PR).

```yaml
test:
  unit:
    commands:
      - name: "set env"
        command: echo "MYENV=test-${OKTETO_NAMESPACE}" >> $OKTETO_ENV
      - name: "run tests"
        command: echo "$MYENV"
```


ifbyol (commented):
According to the spec, each "test container"/"kind of test" just have a single command to be executed, not a list of commands.

cmd/remoterun/test.go
```diff
@@ -72,65 +59,49 @@ commands:
...
+				Name:      options.Name,
+				Namespace: oktetoContext.GetCurrentNamespace(),
+				// For the remote command, the manifest path is the current directory
+				ManifestPath: ".",
+				Deployable:   dep,
+				Variables:    options.Variables,
 			}
 
-			c := &TestCommand{
+			c := &DeployCommand{
```
ifbyol, line 96: We have to be careful. DeployCommand uses internally the `deployable.DeployRunner`, which internally adds phases for each command execution and might be misleading for deploy metrics as it is using the same name for the phase. I guess we will add Okteto Insights metrics for `okteto test`, so we might make the phase name parametrized or something to not mix metrics

ifbyol, line 96: Also, this `DeployCommand` will redeploy the external resources. If we don't share the `OKTETO_ENV` between deploy and test, we might be losing endpoints or failing the operation: https://github.com/okteto/okteto/blob/10d1b4633a369e64462057f190e25f80eb7923ce/pkg/externalresource/external_resource.go#L104

ifbyol:
> According to the spec, each "test container"/"kind of test" just have a single command to be executed, not a list of commands.

Regarding this, clarified offline. We are going to support a list of command for each kind of test but the spec is not updated with that information

PR #4248, opened 2024-04-17
Persist okteto env between deploy and test

maroshii:
The `okteto test` command now read `OKTETO_ENV` and the environment defined in the deploy section. This includes any dinamic values. Also, upon saving a value in `OKTETO_ENV` this value will be stored in the configmap and accesible in the deploy section.

To abe able to accomplish this, we had to add an extra argument `devenv-name` to the deploy-run to be able to retrieve and update the envvars.

## Testing

```yaml
deploy:
  - name: "set env"
    command: echo "DEPLOY_ENV=deploy-envvar" >> $OKTETO_ENV
  - name: "get deploy env"
    command: echo "deploy($DEPLOY_ENV)"
  - name: "get env from test"
    command: echo "test($TEST_ENV)"
  - name: "apply"
    command: kubectl apply -f k8s.yml


test:
  all:
    commands:
      - name: "set env"
        command: echo "TEST_ENV=test-envvar" >> $OKTETO_ENV
      - name: "get env from deploy"
        command: echo "deploy(${DEPLOY_ENV})"
      - name: "get env from test"
        command: echo "test(${TEST_ENV})"
```

Both envvars are accesible:

```
$ ok test
 i  Using maroshii-okteto-admin @ okteto.maroshii.dev.okteto.net as context
 i  'go-getting-started' was already deployed. To redeploy run 'okteto deploy'
 i  Running stage 'all'
 i  Running 'set env'
 i  Running 'get env from deploy'
deploy(deploy-envvar)
 i  Running 'get env from test'
test(test-envvar)
```

After running the tests, they are persisted in the cm:
```
kubectl get cm/okteto-git-go-getting-started -o jsonpath={.data.dependencyEnvs} | base64 -d | jq
{
  "DEPLOY_ENV": "deploy-envvar",
  "KUBECONFIG": "/.../.okteto/kubeconfig-deploy-go-getting-started-1713359524310",
  "OKTETO_AUTODISCOVERY_RELEASE_NAME": "go-getting-started",
  "OKTETO_DISABLE_SPINNER": "true",
  "OKTETO_DOMAIN": "...",
  "OKTETO_NAMESPACE": "...",
  "OKTETO_SKIP_CONFIG_CREDENTIALS_UPDATE": "true",
  "OKTETO_WITHIN_DEPLOY_COMMAND_CONTEXT": "true",
  "TEST_ENV": "test-envvar"
}
```

### DIFF
--- a/cmd/remoterun/test.go
+++ b/cmd/remoterun/test.go
@@ -15,17 +15,22 @@ package remoterun
 
 import (
 	"context"
+	"encoding/base64"
+	"encoding/json"
 	"fmt"
 	"os"
 	"strings"
 
 	contextCMD "github.com/okteto/okteto/cmd/context"
 	"github.com/okteto/okteto/cmd/utils/executor"
+	"github.com/okteto/okteto/pkg/cmd/pipeline"
 	"github.com/okteto/okteto/pkg/constants"
 	"github.com/okteto/okteto/pkg/deployable"
 	"github.com/okteto/okteto/pkg/k8s/kubeconfig"
 	oktetoLog "github.com/okteto/okteto/pkg/log"
+	"github.com/okteto/okteto/pkg/log/io"
 	"github.com/okteto/okteto/pkg/okteto"
+	"github.com/spf13/afero"
 	"github.com/spf13/cobra"
 )
 
@@ -36,18 +41,14 @@ type testRunner interface {
 
 // TestOptions flags accepted by the remote-run test command
 type TestOptions struct {
-	Name      string
-	Variables []string
-}
-
-// TestCommand struct with the dependencies needed to run the test operation
-type TestCommand struct {
-	runner testRunner
+	Name       string
+	DevEnvName string
+	Variables  []string
 }
 
 // Test starts the test command remotely. This is the command executed in the
 // remote environment when running okteto test
-func Test(ctx context.Context) *cobra.Command {
+func Test(ctx context.Context, k8sLogger *io.K8sLogger) *cobra.Command {
 	options := &TestOptions{}
 	cmd := &cobra.Command{
 		Use:   "test",
@@ -86,6 +87,11 @@ commands:
 				return fmt.Errorf("could not read information for tests: %w", err)
 			}
 
+			kubeClient, _, err := okteto.NewK8sClientProviderWithLogger(k8sLogger).ProvideWithLogger(okteto.GetContext().Cfg, k8sLogger)
+			if err != nil {
+				return fmt.Errorf("could not create kubernetes client: %s", err.Error())
+			}
+
 			// Set the default values for the external resources environment variables (endpoints)
 			for name, external := range dep.External {
 				external.SetDefaults(name)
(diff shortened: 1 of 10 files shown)


pkg/deployable/env_stepper.go
```diff
@@ -0,0 +1,79 @@
...
+	"sync"
+
+	"github.com/compose-spec/godotenv"
+	"github.com/spf13/afero"
+)
+
+// envStepper is an environment reader utility that reads the environment from
+// a temporary .env file and keeps a reference to it. It is meant to be called
+// to read environment variables in a command list loop
+type envStepper struct {
```
ifbyol, line 27: I like it! 👏

cmd/remoterun/test.go
```diff
@@ -86,16 +82,35 @@ commands:
...
+					}
+					base64Json, err := pipeline.GetConfigmapDependencyEnv(ctx, devEnvName, namespace, kubeClient)
+					if err != nil {
+						return nil, err
+					}
+					return decodeBase64JSON(base64Json)
+				},
+				SetDevEnvEnviron: func(devEnvName, namespace string, vars []string) (err error) {
+					if devEnvName != "" {
+						err = pipeline.UpdateEnvs(ctx, devEnvName, namespace, vars, kubeClient)
```
ifbyol, line 110: I think we should use other function or at the end, a different (maybe a new one) within the config map, mainly for 2 reasons:
* `OKTETO_ENV` variables are different that dependency envs
* Variables added within `dependencyEnvs` key are, in other parts of the code, added as `OKTETO_DEPENDENCY_<SERVICENAME>_VARIABLE_<VARIABLE>`, but these shouldn't be included. They are not variables belonging to any dependency.

Using the same configmap field to store 2 different things might be misleading, I would create a new field called `oktetoEnv` or similar specifically for these vars. Almost all the logic would be equal, the only thing that would change would be the function to be called here and within `GetDevEnvEnviron`

cmd/test/cmd.go
```diff
@@ -88,6 +89,7 @@ func Test(ctx context.Context, ioCtrl *io.Controller, k8sLogger *io.K8sLogger, a
 	cmd.Flags().StringArrayVarP(&options.Variables, "var", "v", []string{}, "set a variable (can be set more than once)")
 	cmd.Flags().DurationVarP(&options.Timeout, "timeout", "t", getDefaultTimeout(), "the length of time to wait for completion, zero means never. Any other values should contain a corresponding time unit e.g. 1s, 2m, 3h ")
 	cmd.Flags().StringVar(&options.Name, "name", "", "name of the development environment name to be deployed")
+	cmd.Flags().BoolVar(&options.Deploy, "deploy", false, "Always deploy the dev environment. If it's already deployed it will be redeployed")
```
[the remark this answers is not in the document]
maroshii: After using the command I think we definitely need it. It would be a way to ensure the OKTETO_ENV is always up to date when running tests (for CI for eg).

cmd/remoterun/test.go
maroshii: This is not doing anything new. Right now in deploy if you write to `OKTETO_ENV` it will be added to `dependencyEnvs`. I agree it's misleading but I don't think it's something we need to change in this PR. I actually started reading `variables` but then realized they were written to the other key 🤷‍♂️

ifbyol: > This is not doing anything new. Right now in deploy if you write to OKTETO_ENV it will be added to dependencyEnvs

Boh, I have been doings tests that is not working fine:
* If the deploy is executed remotely, env vars (`dependencyEnvs`) are not being added to the configmap. That is added through the function `UpdateEnvsFromCommands` but the struct implementing that function for remote executions is `deployInsideDeployConfigMapHandler` and that implementation is empty.
* When redeploying things from the UI, the `dependencyEnvs` is lost between executions. The translations of the configmap happening in the backend is not taking into account the existence of `dependencyEnvs` key, so as soon as it redeploys the dev environment, previous value is lost. I don't know if that is how it should work or a redeploy should be able to read previous okteto envs.

Also, conceptually seems wrong that we use `dependencyEnvs` key to store things to be consumed by the dev environment itself. The idea of that is to be read by a dev environment that has a dependency defined to me. I think that okteto env should be implemented in another key, and both (`dependencyEnvs` and `oktetoEnv`) should be read by the service that has a dependency with the other.

All this is out of the scope of this PR, but we need to review it. It is not a consistent behavior and might lead to odd scenarios.

ifbyol (approved):
Approving the PR as it satisfies the requirement for okteto test, but I think we need to review all the behavior of OKTETO_ENV and dependencies variables. Right now it doesn't have a consistent behavior, but it is out of the scope of okteto test, it affects to other parts cc @pchico83

cmd/remoterun/test.go
maroshii: Cool, thanks for the testing. Lets discuss tomorrow in daily. I'll hold on this merging this for now

PR #4251, opened 2024-04-18
mask multiline variables

andreafalzetti:
# Proposed changes

Fixes: DEV-293

Currently, multiline variables are not masked by the CLI. This is not intended, it's just the current behaviour because logs are printed line by line, therefore are never matched with the multi-line words that are passed to `AddMaskedWord`.

This solution is not so clean because it adds `***` for each line. However, it's the quickest and less intruisive. Let's discuss if there are other approaches that make sense!

## How to validate

1. Create a multi line variable in the UI or locally
1. Modify a manifest to print such var
1. Run `okteto deploy`

## CLI Quality Reminders 🔧

### DIFF
--- a/pkg/log/log.go
+++ b/pkg/log/log.go
@@ -313,13 +313,18 @@ func IsInteractive() bool {
 
 // AddMaskedWord adds a new word to be redacted, only if longer than 5 chars
 func AddMaskedWord(word string) {
+	clean := strings.TrimSpace(word)
+
+	// only mask words longer than 5 chars (i.e. 'true' and 'false' will not be masked)
 	minCharsToMask := 5
-	lines := strings.Split(word, "\n")
-	for _, line := range lines {
-		if strings.TrimSpace(word) != "" && len(line) > minCharsToMask {
-			log.maskedWords = append(log.maskedWords, line)
-		}
+	if len(clean) < minCharsToMask {
+		return
 	}
+	log.maskedWords = append(log.maskedWords, clean)
+
+	// multi-line support
+	lines := strings.Split(clean, "\n")
+	log.maskedWords = append(log.maskedWords, lines...)
 }
 
 // EnableMasking starts redacting all variables
--- a/pkg/log/log.go
+++ b/pkg/log/log.go
@@ -314,8 +314,11 @@ func IsInteractive() bool {
 // AddMaskedWord adds a new word to be redacted, only if longer than 5 chars
 func AddMaskedWord(word string) {
 	minCharsToMask := 5
-	if strings.TrimSpace(word) != "" && len(word) > minCharsToMask {
-		log.maskedWords = append(log.maskedWords, word)
+	lines := strings.Split(word, "\n")
+	for _, line := range lines {
+		if strings.TrimSpace(word) != "" && len(line) > minCharsToMask {
+			log.maskedWords = append(log.maskedWords, line)
+		}
 	}
 }
 
--- a/pkg/log/log.go
+++ b/pkg/log/log.go
@@ -322,7 +322,8 @@ func AddMaskedWord(word string) {
 	}
 	log.maskedWords = append(log.maskedWords, clean)
 
-	lines := strings.Split(clean, "\n")
+	crossPlatformCleanWord := strings.ReplaceAll(clean, "\r\n", "\n")
+	lines := strings.Split(crossPlatformCleanWord, "\n")
 	for _, line := range lines {
 		cleanLine := strings.TrimSpace(line)
 		if len(cleanLine) <= minCharsToMask {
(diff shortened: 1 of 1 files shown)


jLopezbarb:
I wasn't able to make it work with the following steps:
1. Run the following command
```bash
export multiline="Line 1
Line 2
Line 3
```
2. Run an `okteto deploy` that has an step `echo $multiline`

Can you revise it pls?

andreafalzetti:
@jLopezbarb 

> I wasn't able to make it work with the following steps:
> 
>     1. Run the following command
> 
> 
> ```shell
> export multiline="Line 1
> Line 2
> Line 3
> ```
> 
>     2. Run an `okteto deploy` that has an step `echo $multiline`
> 
> 
> Can you revise it pls?

@jLopezbarb local env vars are not masked, you either need to use `.env` or a platform variable (user/cluster from the UI), or using the deploy `--var` flag

maroshii (approved):

ifbyol (commented):
This has a problem, but not sure if this would be a blocker or not.

If I have something like this:

```
This is the first line
this is the second line
github
This is the last line
```

On every place where `github` is mentioned, it will be masked, but it shouldn't.

If you have a manifest like this one 

```
deploy:
  - echo "This is a test to show github word"
  - echo "Cloning https://github.com/okteto/movies...."
  - echo "This is another log where github is mentioned"
 ```

You can see how in every place where `github` is mentioned, it is masked. You can see it here

Screenshot 2024-04-25 at 18 41 47


@codyjlandstrom @pchico83 Do you see it relevant enough to block this? I don't see it a big deal, usually multiline variables will be things like private keys or things like that and it is hard to find "common" word, but mentioning it just it case you find it relevant.

As mentioned in the description, this is because we print longs line by line, so the whole multiline variable would never be detected.

**Comment by pchico83** [comment]
> This has a problem, but not sure if this would be a blocker or not.
> 
> If I have something like this:
> 
> ```
> This is the first line
> this is the second line
> github
> This is the last line
> ```
> 
> On every place where `github` is mentioned, it will be masked, but it shouldn't.
> 
> If you have a manifest like this one
> 
> ```
> deploy:
>   - echo "This is a test to show github word"
>   - echo "Cloning https://github.com/okteto/movies...."
>   - echo "This is another log where github is mentioned"
> ```
> 
> You can see how in every place where `github` is mentioned, it is masked. You can see it here
> 
> Screenshot 2024-04-25 at 18 41 47
> @codyjlandstrom @pchico83 Do you see it relevant enough to block this? I don't see it a big deal, usually multiline variables will be things like private keys or things like that and it is hard to find "common" word, but mentioning it just it case you find it relevant.
> 
> As mentioned in the description, this is because we print longs line by line, so the whole multiline variable would never be detected.

I think it's fine. As long as we don't mask sublines with 5 chars or less.

**Comment by ifbyol** [comment]
> I think it's fine. As long as we don't mask sublines with 5 chars or less.

Ok, perfect!

@andreafalzetti That case ^ is not working fine. I have modified my previous example with the following multiline variable

```
This is the first line
this is the second line

gith
 
This is the last line
```

(The last blank line has an space)

This is the output

Screenshot 2024-04-25 at 19 01 14

**Review by ifbyol — CHANGES_REQUESTED** [review]
Requesting a change to fix the issue in which an empty line in the multiline secret breaks completely the output

**Comment by andreafalzetti** [comment]
@ifbyol I've updated the PR, can you have another look when you can? Thanks

File: pkg/log/log.go [hunk 1]
```diff
@@ -313,9 +313,23 @@ func IsInteractive() bool {
...
 	minCharsToMask := 5
-	if strings.TrimSpace(word) != "" && len(word) > minCharsToMask {
-		log.maskedWords = append(log.maskedWords, word)
+	if len(clean) < minCharsToMask {
+		return
+	}
+	log.maskedWords = append(log.maskedWords, clean)
+
+	// multi-line support
+	lines := strings.Split(clean, "\n")
```
ifbyol, line 326: @andreafalzetti  Could you check in a windows system? The character for new lines is not the same so not sure there might be some issue there

ifbyol (approved):
Now looks good. I left a comment about checking this with Windows systems as the character for new line are not the same. We had issues in the past with that. Not sure if in this case would be relevant or not

pkg/log/log.go
andreafalzetti: @ifbyol good point, I didn't remember about Win. I've tested now and it already works, I've done tests with PowerShell and cmd.exe but just in case I've added cross platform support with the last commit, to avoid any suprise in any edge case that I could not reproduce on Windows!

ifbyol (approved):

PR #4254, opened 2024-04-19
feat: deprecate compose file detection by naming

jLopezbarb:
# Proposed changes

Partly fixes DEV-248.

We are deprecating the detection of a compose or stack file based on the name of the file. Now the users will have to use `OKTETO_SUPPORT_STACKS_ENABLED=true` in order to have previous behaviour.

- Adds new env var `OKTETO_SUPPORT_STACKS_ENABLED` ([Documentation PR](https://github.com/okteto/docs/pull/706)/[Stack vs Compose community post](https://www.notion.so/okteto/Important-Update-Migrating-from-Okteto-Stacks-to-Docker-Compose-53292a7f72324beca9e0f6b9eddcc217?pvs=4))
- The new env var is the one that says if a file is a compose or not

## How to validate

1. From https://github.com/okteto/movies-with-compose
2. Rename `docker-compose.yml` to `stack.yml`
3. Run `okteto deploy`

## CLI Quality Reminders 🔧

### DIFF
--- a/pkg/model/stack.go
+++ b/pkg/model/stack.go
@@ -792,8 +792,8 @@ func isFileCompose(path string) bool {
 	isStackSupported := env.LoadBooleanOrDefault(stackSupportEnabledEnvVar, false)
 	if !isStackSupported {
 		oktetoLog.Infof("%s is set to false. File will be treated as compose", stackSupportEnabledEnvVar)
-		if isComposeFileName {
-			oktetoLog.Warning("The file %s will be deprecated as a default compose file name in a future version. Please consider renaming your compose file to 'okteto-stack.yml'", base)
+		if !isComposeFileName {
+			oktetoLog.Warning("The file %s will be deprecated as a stack file name in a future version. Please consider migrating your file to a compose", base)
 		}
 		return true
 	}
--- a/pkg/model/stack.go
+++ b/pkg/model/stack.go
@@ -40,6 +40,12 @@ import (
 	resource "k8s.io/apimachinery/pkg/api/resource"
 )
 
+const (
+	// stacksSupportEnabledEnvVar is the environment variable to know if we should use compose or stacks
+	// Default is false
+	stackSupportEnabledEnvVar = "OKTETO_SUPPORT_STACKS_ENABLED"
+)
+
 var (
 	errBadStackName     = "must consist of lower case alphanumeric characters or '-', and must start and end with an alphanumeric character"
 	deprecatedManifests = []string{"stack.yml", "stack.yaml"}
@@ -777,9 +783,21 @@ func (svcResources *ServiceResources) IsDefaultValue() bool {
 	return svcResources.CPU.Value.IsZero() && svcResources.Memory.Value.IsZero() && svcResources.Storage.Size.Value.IsZero() && svcResources.Storage.Class == ""
 }
 
-func isPathAComposeFile(path string) bool {
+// isFileCompose checks if the path is a compose file
+// if the env var OKTETO_SUPPORT_STACKS_ENABLED is set to true, it will return true for any file no matter the name
+// if the env var is not set, it will return true for files that start with "compose", "docker-compose" or "okteto-compose"
+func isFileCompose(path string) bool {
 	base := filepath.Base(path)
-	return strings.HasPrefix(base, "compose") || strings.HasPrefix(base, "docker-compose") || strings.HasPrefix(base, "okteto-compose")
+	isComposeFileName := strings.HasPrefix(base, "compose") || strings.HasPrefix(base, "docker-compose") || strings.HasPrefix(base, "okteto-compose")
+	if !env.LoadBooleanOrDefault(stackSupportEnabledEnvVar, false) {
+		oktetoLog.Infof("%s is set to false. File will be treated as compose", stackSupportEnabledEnvVar)
+		if isComposeFileName {
+			oktetoLog.Warning("The file %s will be deprecated as a default compose file name in a future version. Please consider renaming your compose file to 'okteto-stack.yml'", base)
+		}
+		return true
+	}
+	oktetoLog.Infof("%s is set to true. Detecting if file is compose by name", stackSupportEnabledEnvVar)
+	return isComposeFileName
 }
 
 // LoadStack loads an okteto stack manifest checking "yml" and "yaml"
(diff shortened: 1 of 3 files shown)


pkg/model/stack.go
```diff
@@ -777,9 +783,21 @@ func (svcResources *ServiceResources) IsDefaultValue() bool {
...
 
-func isPathAComposeFile(path string) bool {
+// isFileCompose checks if the path is a compose file
+// if the env var OKTETO_SUPPORT_STACKS_ENABLED is set to true, it will return true for any file no matter the name
+// if the env var is not set, it will return true for files that start with "compose", "docker-compose" or "okteto-compose"
+func isFileCompose(path string) bool {
 	base := filepath.Base(path)
-	return strings.HasPrefix(base, "compose") || strings.HasPrefix(base, "docker-compose") || strings.HasPrefix(base, "okteto-compose")
+	isComposeFileName := strings.HasPrefix(base, "compose") || strings.HasPrefix(base, "docker-compose") || strings.HasPrefix(base, "okteto-compose")
+	if !env.LoadBooleanOrDefault(stackSupportEnabledEnvVar, false) {
```
pchico83, line 792:
```suggestion
	if !env.LoadBooleanOrDefault(stackSupportEnabledEnvVar, false) {
```
Shouldn't it be without the negation?

jLopezbarb: No, We want to enter in the if when `env.LoadBooleanOrDefault(stackSupportEnabledEnvVar, false)` is `false` as we want to always return that is a compose if the env var is not enabled. That's why it's negated, but reading again the code I think it's a bit complex to understand directly with the function, so I'm setting a variable in order to make it more readable

pkg/model/stack.go
```diff
@@ -777,9 +783,22 @@ func (svcResources *ServiceResources) IsDefaultValue() bool {
...
+// if the env var OKTETO_SUPPORT_STACKS_ENABLED is set to true, it will return true for any file no matter the name
+// if the env var is not set, it will return true for files that start with "compose", "docker-compose" or "okteto-compose"
+func isFileCompose(path string) bool {
 	base := filepath.Base(path)
-	return strings.HasPrefix(base, "compose") || strings.HasPrefix(base, "docker-compose") || strings.HasPrefix(base, "okteto-compose")
+	isComposeFileName := strings.HasPrefix(base, "compose") || strings.HasPrefix(base, "docker-compose") || strings.HasPrefix(base, "okteto-compose")
+	isStackSupported := env.LoadBooleanOrDefault(stackSupportEnabledEnvVar, false)
+	if !isStackSupported {
+		oktetoLog.Infof("%s is set to false. File will be treated as compose", stackSupportEnabledEnvVar)
+		if isComposeFileName {
```
pchico83, line 795: @jLopezbarb then I don't understand the warning. I guess we should warn if OKTETO_SUPPORT_STACKS_ENABLED is true and the file is a stack file, right?
Maybe we should keep `OKTETO_SUPPORT_STACKS_ENABLED: true` by default but start to warn our users

pkg/model/stack.go
```diff
@@ -777,9 +783,22 @@ func (svcResources *ServiceResources) IsDefaultValue() bool {
...
 
-func isPathAComposeFile(path string) bool {
+// isFileCompose checks if the path is a compose file
+// if the env var OKTETO_SUPPORT_STACKS_ENABLED is set to true, it will return true for any file no matter the name
+// if the env var is not set, it will return true for files that start with "compose", "docker-compose" or "okteto-compose"
+func isFileCompose(path string) bool {
 	base := filepath.Base(path)
-	return strings.HasPrefix(base, "compose") || strings.HasPrefix(base, "docker-compose") || strings.HasPrefix(base, "okteto-compose")
+	isComposeFileName := strings.HasPrefix(base, "compose") || strings.HasPrefix(base, "docker-compose") || strings.HasPrefix(base, "okteto-compose")
+	isStackSupported := env.LoadBooleanOrDefault(stackSupportEnabledEnvVar, false)
```
pchico83, line 792: @jLopezbarb can we keep `true` as default? this way we can announce this change on the next release and give time to our users to adapt to this change. And we could change the default in CLI 3.0

pchico83: Also, could we track if a file is managed as an Okteto Stack or Docker Compose? This will help us to make the decision of removing Okteto Stacks, or defaulting to Docker Compose

jLopezbarb: > can we keep true as default? this way we can announce this change on the next release and give time to our users to adapt to this change. And we could change the default in CLI 3.0

Done in https://github.com/okteto/okteto/pull/4254/commits/11abd9133b11698e27db082700eb6696f9fbbc89

 > Also, could we track if a file is managed as an Okteto Stack or Docker Compose? This will help us to make the decision of removing Okteto Stacks, or defaulting to Docker Compose
 
@pchico83  We are already tracking it in mixpannel for okteto deploys, should we track it as a separate event?

pchico83: No, that fine. Do we need to adapt the logic of the metric to consider the envvar?

pkg/model/stack.go
```diff
@@ -777,9 +783,22 @@ func (svcResources *ServiceResources) IsDefaultValue() bool {
...
 	base := filepath.Base(path)
-	return strings.HasPrefix(base, "compose") || strings.HasPrefix(base, "docker-compose") || strings.HasPrefix(base, "okteto-compose")
+	isComposeFileName := strings.HasPrefix(base, "compose") || strings.HasPrefix(base, "docker-compose") || strings.HasPrefix(base, "okteto-compose")
+	isStackSupported := env.LoadBooleanOrDefault(stackSupportEnabledEnvVar, true)
+	if !isStackSupported {
+		oktetoLog.Infof("%s is set to false. File will be treated as compose", stackSupportEnabledEnvVar)
+		return true
+	}
+	if !isComposeFileName {
+		oktetoLog.Warning("okteto stack syntax is deprecated. Please consider migrating to compose syntax: https://community.okteto.com/t/important-update-migrating-from-okteto-stacks-to-docker-compose/1262")
```
pchico83, line 798:
```suggestion
		oktetoLog.Warning("Okteto Stack syntax is deprecated. Please consider migrating to Docker Compose syntax: https://community.okteto.com/t/important-update-migrating-from-okteto-stacks-to-docker-compose/1262")
```
Also, could you share a screenshot? Maybe it looks better with a new line after "Okteto Stack syntax is deprecated"

jLopezbarb: I think the first line is very tiny compared to the second one. But the rest of warning messages are this way so I'd go with splitting in two lines
Screenshot 2024-04-26 at 14 09 36

pchico83 (approved):

pkg/model/stack.go
jLopezbarb: No, we get if it's a compose or a stack based on this code so there is no need to change it elsewhere

ifbyol (approved):

PR #4255, opened 2024-04-19
Build automatically an image in a compose scenario where build is not specified, image is specified and there are volume mounts

ifbyol:
# WIP

Created PR to run the e2e tests

# Proposed changes

Fixes DEV-298

## How to validate

1.
1.
1.

## CLI Quality Reminders 🔧

### DIFF
--- a/pkg/build/file.go
+++ b/pkg/build/file.go
@@ -1,0 +1,65 @@
+// Copyright 2024 The Okteto Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package build
+
+import (
+	"bufio"
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/okteto/okteto/pkg/config"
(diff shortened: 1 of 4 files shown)


pkg/build/file.go
```diff
@@ -0,0 +1,65 @@
...
+	if err := fs.MkdirAll(dockerfileTmpFolder, filePerm); err != nil {
+		return build, fmt.Errorf("failed to create %s: %w", dockerfileTmpFolder, err)
+	}
+
+	tmpFile, err := afero.TempFile(fs, dockerfileTmpFolder, "buildkit-")
+	if err != nil {
+		return build, err
+	}
+
+	datawriter := bufio.NewWriter(tmpFile)
```
maroshii, line 47: Why do we need this extra writer? It looks to me that we can create a string in memory and simply write to `tmpFile` once

ifbyol: I just recovered the function that we were using before the big refactor included as part of `2.26`. I just change it to use a string in memory instead of create another writter or builder

pkg/build/file.go
```diff
@@ -0,0 +1,63 @@
...
+// CreateDockerfileWithVolumeMounts creates the Dockerfile with volume mounts and returns the BuildInfo
+func CreateDockerfileWithVolumeMounts(context, image string, volumes []VolumeMounts, fs afero.Fs) (*Info, error) {
+	build := &Info{}
+
+	ctx, err := filepath.Abs(context)
+	if err != nil {
+		return build, nil
+	}
+	build.Context = ctx
+	dockerfileTmpFolder := filepath.Join(config.GetOktetoHomeWithFilesystem(fs), ".dockerfile")
```
pchico83, line 34: Do we have an issue to ignore the dockerfile name from the smart builds logic?

pchico83 (approved):

pkg/build/file.go
ifbyol: Not yet, but I have a note to create it. I will do it now

PR #4256, opened 2024-04-22
Fix linting errors in master (samples, etc)

andreafalzetti:
# Proposed changes

This PR fixes different linting errors that we have now in `master`. To validate, run `make lint`.


.pre-commit-config.yaml
```diff
@@ -23,7 +23,7 @@ repos:
       - id: check-toml
       - id: check-vcs-permalinks
       - id: detect-private-key
-        exclude: ^pkg/k8s/secrets/certs\.go$|^pkg/syncthing/certs\.go$|^cmd/deploy/certs\.go$
```
andreafalzetti, line 26: this file has been moved now, but we had the old rule, which makes the linter complain

jLopezbarb (approved):

PR #4257, opened 2024-04-22
Move from concept of "Secret" to "Variable"

andreafalzetti:
# Proposed changes

Fixes: DEV-280

With this PR we align with the Okteto Platform UI, the concept of "Secret" is now called "Variable".

This change affects both the codebase (renaming identifiers) but also the user-experience (via logs)

## How to validate

Try interacting with User/Cluster variables

## CLI Quality Reminders 🔧


cmd/context/use.go
```diff
@@ -226,17 +226,17 @@ func getContext(ctxOptions *Options) (string, error) {
...
-	for _, secret := range secrets {
-		value, exists := os.LookupEnv(secret.Name)
+func exportPlatformVariablesToEnv(variables []env.Var) {
+	for _, v := range variables {
+		value, exists := os.LookupEnv(v.Name)
 		if exists {
-			if value != secret.Value {
-				oktetoLog.Warning("$%s secret is being overridden by a local environment variable by the same name.", secret.Name)
+			if value != v.Value {
+				oktetoLog.Warning("$%s Okteto Platform variable is being overridden by a local environment variable with the same name.", v.Name)
```
andreafalzetti, line 234: this is the only user-facing change

pchico83, line 234:
```suggestion
				oktetoLog.Warning("$%s Okteto variable is being overridden by a local environment variable with the same name.", v.Name)
```
We don't have a concept for "Okteto Platform variable"

andreafalzetti: @pchico83 we have been using this term internally in the dev team to indicate User plus Cluster variables. Do you have a better alternative in mind? Or how would you change the copy? cc @codyjlandstrom

maroshii (approved):

cmd/context/use.go
codyjlandstrom: "Admin variables" is the name that we've been using across the [documentation](https://www.okteto.com/docs/admin/dashboard/#variables).
This [doc](https://www.okteto.com/docs/admin/dashboard/#variables) helps to document the hierarchy of the variables if that's helpful.
Lets proceed with "Admin Variables"

codyjlandstrom: Seperately, we should rename the "Variables" section in the admin UI to be "Admin Variables" for clarity

pchico83: @rlamana could you create an issue to track the UI changes?

rlamana: Just to clarify, we are talking about changing "Variables" to "Admin Variables" in the Admin section of the UI, right?

rlamana: Issue created here: https://okteto.atlassian.net/browse/DEV-313

andreafalzetti: Does "Admin Variables" include the "User Variables"? Why do we call them "Admin" even if I am just a user and I can add my own vars? 🤔  @codyjlandstrom

rlamana: We call them admin because only admin can define those and they have a cluster wide scope. As a user, the only thing you need to now is that you can override Admin variables. We should inform about this through the docs and the UI.

rlamana: @codyjlandstrom @pchico83 @ifbyol could we list in the "user variables" the variables that are already defined and that the user can override? Would that be useful? A priori I think it could help a lot. Not sure if there are security implications, we could only show the name but not the values, as I think we already do today...

andreafalzetti: @codyjlandstrom in the docs we have the concept of "User Variables" and "Admin Variables" they get merged together at this point so I cannot differenciate and let the user know which one is it. That's wht I am trying to find a common name for those. Thoughts?

codyjlandstrom: Here is that document on variable preference: https://www.okteto.com/docs/core/okteto-variables/#user-defined-variables.
@andreafalzetti Since I'm just getting up to speed with this feature. The "Platform Variable" in reference can only be created by the admin?

pchico83: I think the closer term to Okteto Platform Variables is "Okteto Variables":
https://www.okteto.com/docs/core/okteto-variables/
I would go with:
```
Okteto Variable '%s' is overridden by a local environment variable with the same name
```

ifbyol: We are talking different things in the same thread and we might be confusing the original purpose of the comment.

For context, CLI doesn't know if they are Admin or User variables. CLI just gets the variables that have to be taken into account. For example, If a User variable overrides an Admin Variable, CLI will just know the User one.

> The "Platform Variable" in reference can only be created by the admin?

Replying to @codyjlandstrom. "Platform Variable" as @andreafalzetti  called it is any Variable defined through the UI (Admin or User). 

Taking into account that, I agree with @pchico83 's last comment about just call them `Okteto Variable`. Let's discuss in other thread/place the other changes mentioned here to avoid confusions

cmd/context/use.go
```diff
@@ -226,17 +226,17 @@ func getContext(ctxOptions *Options) (string, error) {
...
-	for _, secret := range secrets {
-		value, exists := os.LookupEnv(secret.Name)
+func exportPlatformVariablesToEnv(variables []env.Var) {
+	for _, v := range variables {
+		value, exists := os.LookupEnv(v.Name)
 		if exists {
-			if value != secret.Value {
-				oktetoLog.Warning("$%s secret is being overridden by a local environment variable by the same name.", secret.Name)
+			if value != v.Value {
+				oktetoLog.Warning("$%s User or Admin variable defined in the Okteto UI is being overridden by a local environment variable with the same name.", v.Name)
```
andreafalzetti, line 234: @pchico83 can you help me unblock this PR and agree on a msg? I've updated to `User or Admin variable` instead of `Platform Variables`. WDYT?

pchico83: @andreafalzetti it could also be a variable in `.env` right? I would just say:
```
Okteto Variable '%s' is overridden by a local environment variable with the same name
```

ifbyol (approved):

pchico83 (approved):

jLopezbarb (approved):

PR #4261, opened 2024-04-24
Detect syncthing readiness issue

andreafalzetti:
# Proposed changes

Fixes DEV-149

This PR aims to improve the user experience in regards with the `okteto up` command and in particular, when there is no enough space left on the disk, for the Okteto CLI to start syncthing successfully.

## How to validate

Internal demo video: https://www.loom.com/share/1ebd6590e0984138aa57b14b95e02a4f 

I recommended using GitHub Codespaces to avoid filling your personal laptop's disk.

1. Start a Codespace for `okteto/okteto` in this branch
1. Build the CLI
1. Setup your Okteto context (ie. I've used an handy `.env` file which I copied from local)
1. Clone a sample repo such as `git clone https://github.com/okteto/go-getting-started.git`
1. Check the available space (`df -h` and see the `overlay` mount as reference)
1. Fill the space as much as you can (use `fallocate -l <size> <file>` ie. `fallocate -l 10G blob`)
1. Go to `go-getting-started` (or your sample repo) and run `okteto up`
1. If you managed to fill up the disk entirely you should get the error as follows:

Screenshot 2024-04-24 at 15 01 05

## CLI Quality Reminders 🔧

### DIFF
--- a/pkg/syncthing/syncthing.go
+++ b/pkg/syncthing/syncthing.go
@@ -34,6 +34,7 @@ import (
 	"github.com/google/uuid"
 	"github.com/okteto/okteto/pkg/config"
 	oktetoErrors "github.com/okteto/okteto/pkg/errors"
+	"github.com/okteto/okteto/pkg/filesystem"
 	oktetoLog "github.com/okteto/okteto/pkg/log"
 	"github.com/okteto/okteto/pkg/model"
 	"github.com/shirou/gopsutil/process"
@@ -70,6 +71,11 @@ const (
 	GUIPort = 8384
 
 	maxRetries = 3
+
+	maxLogTailLinesToRead = 5
+
+	// one line of logs in UTF-8 is between 3-400 bytes
+	maxLogTailChunkByteSize int64 = 1024
 )
 
 var regexErrOpeningDatabase = regexp.MustCompile("Error opening database: mkdir .*: no space left on device")
@@ -386,13 +392,19 @@ func (s *Syncthing) IdentifyReadinessIssue(fs afero.Fs) error {
 
 // RegexMatchesLogs checks if a regex matches in the syncthing logs
 func (s *Syncthing) RegexMatchesLogs(fs afero.Fs, regx *regexp.Regexp) bool {
-	log, err := afero.ReadFile(fs, s.LogPath)
+	lines, err := filesystem.GetLastNLines(fs, s.LogPath, maxLogTailLinesToRead, maxLogTailChunkByteSize)
 	if err != nil {
 		oktetoLog.Infof("error reading syncthing log: %s", err)
 		return false
 	}
 
-	return regx.Match(log)
+	for _, log := range lines {
+		if regx.MatchString(log) {
+			return true
+		}
+	}
+
+	return false
 }
 
 // Ping checks if syncthing is available
--- a/pkg/filesystem/file.go
+++ b/pkg/filesystem/file.go
@@ -14,6 +14,7 @@
 package filesystem
 
 import (
+	"bufio"
 	"io"
 	"os"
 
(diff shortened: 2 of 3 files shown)


pkg/syncthing/syncthing.go
```diff
@@ -362,6 +369,28 @@ func (s *Syncthing) WaitForPing(ctx context.Context, local bool) error {
 	}
 }
 
+// IdentifyReadinessIssue attempts to identify the issue that is preventing syncthing from being ready
+func (s *Syncthing) IdentifyReadinessIssue(fs afero.Fs) error {
+	if s.RegexMatchesLogs(fs, regexp.MustCompile("Error opening database: mkdir .*: no space left on device")) {
+		return oktetoErrors.ErrInsufficientSpace
```
pchico83, line 375: I would be more specific that the lack of space is in the developer's machine, not in the remote container

cmd/up/syncthing.go
```diff
@@ -110,11 +110,11 @@ func (up *upContext) startSyncthing(ctx context.Context) error {
 		return err
 	}
 
-	if err := up.Sy.WaitForPing(ctx, true); err != nil {
+	if err := up.Sy.WaitForPing(ctx, true, up.Fs); err != nil {
```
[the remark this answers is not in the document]
ifbyol: Agree, I think we should add `Fs` to `Synthing` struct (and honestly, I think it makes sense as it works with files) and make sure we always pass one when creating it. I think `Syncthing` is only being created in a couple of cases and one of them is created within `up`, so it already has the filesystem

pkg/syncthing/syncthing.go
```diff
@@ -362,6 +369,28 @@ func (s *Syncthing) WaitForPing(ctx context.Context, local bool) error {
 	}
 }
 
+// IdentifyReadinessIssue attempts to identify the issue that is preventing syncthing from being ready
+func (s *Syncthing) IdentifyReadinessIssue(fs afero.Fs) error {
+	if s.RegexMatchesLogs(fs, regexp.MustCompile("Error opening database: mkdir .*: no space left on device")) {
```
ifbyol, line 374: 2 things to be aware:
* `MustCompile` will throw a panic if the expression cannot be compiled. As it is a static string I don't see a problem
* As the expression we pass to `MustCompile` is static you could create a global variable with the regex, so it is only compiled once.

I think given the use case of the CLI and this code both things are fine, but commenting them as you have more context about it and it case it might be relevant

pkg/syncthing/syncthing.go
andreafalzetti: agree, updated in 679f0d40 to be:

```
ErrInsufficientSpaceOnUserDisk = fmt.Errorf("there isn't enough disk space available on your machine to continue")
```

pkg/syncthing/syncthing.go
andreafalzetti: Thanks Nacho, I am aware of `MustCompile` and added test on-purpose to make sure we notice if it panics.  I've moved them at the top of the file (personally, I prefer the neglectable operation cost of compiling over having 2 vars defined ~200 lines of code above the line where they are used, considering they are not shared with any other function either, it makes the code less readable to me)

pkg/syncthing/syncthing.go
```diff
@@ -362,6 +369,28 @@ func (s *Syncthing) WaitForPing(ctx context.Context, local bool) error {
...
+	}
+	if s.RegexMatchesLogs(fs, regexp.MustCompile("insufficient space on disk for database")) {
+		return oktetoErrors.ErrInsufficientSpace
+	}
+	return nil
+}
+
+// RegexMatchesLogs checks if a regex matches in the syncthing logs
+func (s *Syncthing) RegexMatchesLogs(fs afero.Fs, regx *regexp.Regexp) bool {
+	log, err := afero.ReadFile(fs, s.LogPath)
```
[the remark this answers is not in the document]
andreafalzetti: @maroshii changes in 20f052f4271e6da53d2fbd97da2b2ec43eedbc0c

cmd/up/syncthing.go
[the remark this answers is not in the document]
andreafalzetti: To avoid blocking this PR, I've opened another one for that refactor: https://github.com/okteto/okteto/pull/4266

maroshii (approved):

PR #4269, opened 2024-04-29
add user defined caches to okteto test

maroshii:
# Proposed changes

Add `cache` section to test manifest section. These caches expand to buildkit's cache mounts: https://docs.docker.com/build/guide/mounts/#add-a-cache-mount


pkg/model/test.go
```diff
@@ -19,6 +19,7 @@ type Test struct {
 	Context   string        `yaml:"context,omitempty"`
 	Commands  []TestCommand `yaml:"commands,omitempty"`
 	DependsOn []string      `yaml:"depends_on,omitempty"`
+	Cache     []string      `yaml:"cache,omitempty"`
```
jLopezbarb, line 22: I don't see a custom unmarshall in order to have the following syntax:

```yaml
tests:
  cache: my-image
```

We already have that in the compose spec. Should we be doing it here?

pkg/remote/remote_test.go
```diff
@@ -374,7 +374,9 @@ ARG OKTETO_INVALIDATE_CACHE
 
 RUN okteto registrytoken install --force --log-output=json
 
-RUN --mount=type=secret,id=known_hosts --mount=id=remote,type=ssh \
+RUN \
```
jLopezbarb, line 377: Should we create a test with cache in order to test that it won't break?

andreafalzetti (approved):
I can only approve based on trust because there is no context on what this change should do, neither in the PR or the ticket

pkg/model/test.go
maroshii: Should we allow this? cc @pchico83

pchico83: For simplicity, I would go with a list of strings for now.
But I would rename the field to `caches`

PR #4270, opened 2024-04-29
fix: clone images on specified services

jLopezbarb:
# Proposed changes

Fixes DEV-232

- Fix multi tags with okteto registries. 
  - When we run a command like `okteto build -t okteto.dev/myapp:1.0.0,okteto.dev/myapp:latest`
  - It was preventing us to create a clone
- Create a new clone function that clones one image into another without calculating anything
  - Reuse that new function in the `cloneFromGlobalToDev`
- Fix: Smartbuild was not getting the correct interface 
  - The interface was the same but the methods were named different. In this PR we use the same naming
- Created `getGlobalTagFromDevIfNeccesary` in order to get the global tag from dev if we need to build into the global registry too

It has created a side effect where we show that the image has been successfully pushed into the two registries. I think it's better for the user so he knows what's happening. Are we ok with that
Screenshot 2024-05-22 at 12 49 15

## How to validate

1. Copy the following okteto.ylm
```yaml
build:
  my-service:
    image: okteto.dev/my-service:my-tag
    context: my-service
```
2. Create `my-service` folder with a simple `Dockerfile` (`FROM alpine`)
3. Run `okteto build`
4. Check that we are pushing to the dev registry to the image in the field and the global registry with the hash as the tag

## CLI Quality Reminders 🔧


cmd/build/v2/tagger.go
```diff
@@ -92,6 +94,19 @@ func (it imageTagger) getServiceImageReference(manifestName, svcName string, b *
 	return useReferenceTemplate(targetRegistry, sanitizedName, svcName, model.OktetoDefaultImageTag)
 }
 
+func (it imageTagger) getGlobalTagFromDevIfNeccesary(tags, namespace, registryURL, buildHash, manifestName, svcName string, ic registry.ImageCtrl) string {
+	tagList := strings.Split(tags, ",")
```
jLopezbarb, line 98: We need to expand this to build all the global images from the parsed tags

pkg/cmd/build/build.go
```diff
@@ -203,19 +203,23 @@ func (ob *OktetoBuilder) buildWithDocker(ctx context.Context, buildOptions *type
...
-	if (reg.IsOktetoRegistry(imageTag)) && strings.Count(imageTag, "/") != 1 {
-		prefix := constants.DevRegistry
-		if reg.IsGlobalRegistry(imageTag) {
-			prefix = constants.GlobalRegistry
-		}
-		return oktetoErrors.UserError{
-			E:    fmt.Errorf("'%s' isn't a valid image tag", imageTag),
-			Hint: fmt.Sprintf("The Okteto Registry syntax is: '%s/image_name'", prefix),
+	tags := strings.Split(imageTags, ",")
+	for _, tag := range tags {
```
jLopezbarb, line 213: Fix to pass several tags with` okteto.dev`

ifbyol:
> It has created a side effect where we show that the image has been successfully pushed into the two registries. I think it's better for the user so he knows what's happening. Are we ok with that

@jLopezbarb Why the double message? Before it was also happening but only on BuildKit logs, but not in the final success step

ifbyol (commented):
As we talked offline, we are mixing 2 ways of cloning images:
* Old one: If the user didn't specify an image in the build info, we pass one tag in `buildInfo.Tag` and then, in buildkit.go, we fill `buildOptions.DevTag` with the dev tag to push
* New one (added in this PR). If the user specify an image, and it is for dev registry, we pass 2 images in `buildOptions.Tag` and then, `DevTag` is not filled just because dev tag is the first in the list

We should unify both ways to just one to avoid problems in the future

cmd/build/v2/build.go
```diff
@@ -341,13 +343,24 @@ func (bc *OktetoBuilder) buildSvcFromDockerfile(ctx context.Context, manifest *m
...
-	tagToBuild := newImageTagger(bc.Config, bc.smartBuildCtrl).getServiceImageReference(manifest.Name, svcName, buildSvcInfo, buildHash)
-	buildSvcInfo.Image = tagToBuild
+	it := newImageTagger(bc.Config, bc.smartBuildCtrl)
+	tagsToBuild := it.getServiceImageReference(manifest.Name, svcName, buildSvcInfo, buildHash)
+	imageCtrl := registry.NewImageCtrl(bc.oktetoContext)
+	globalImage := it.getGlobalTagFromDevIfNeccesary(tagsToBuild, bc.oktetoContext.GetNamespace(), bc.oktetoContext.GetRegistryURL(), buildHash, manifest.Name, svcName, imageCtrl)
+	if globalImage != "" {
+		tagsToBuild = fmt.Sprintf("%s,%s", tagsToBuild, globalImage)
+	}
+	if buildSvcInfo.Image == "" {
```
ifbyol, line 357: How hard would be to generate always the dev tag when image is empty, so we don't need to account for this special case?

I mean, this has an implicit knowledge. When the image is not specified, we are building the image for the global registry (don't know exactly why we build that one), and then, we generate the dev one. It is hard to remember to change this if we change which image is built first in that case.

There shouldn't be special cases and consider everything the same as much as possible for simplicity and code maintainability.

If we don't change it, at least, let's add a comment before the if to specify why we need this if

cmd/build/v2/build.go
```diff
@@ -341,13 +343,24 @@ func (bc *OktetoBuilder) buildSvcFromDockerfile(ctx context.Context, manifest *m
...
+		devImage := it.getDevTagFromGlobalIfNeccesary(tagsToBuild, bc.oktetoContext.GetNamespace(), bc.oktetoContext.GetGlobalNamespace(), bc.oktetoContext.GetRegistryURL(), buildHash, manifest.Name, svcName, imageCtrl)
+		if devImage != "" {
+			tagsToBuild = fmt.Sprintf("%s,%s", tagsToBuild, devImage)
```
ifbyol, line 360: Doing this in that order, it will imply that [here](https://github.com/okteto/okteto/pull/4270/files#diff-2bcc58eee26848ca1af49b8495ad2e33224ada8c2be4a3439596d5f3bea42c6fR102) we will display the global image when the image is calculated by us, but we should display the dev one

Screenshot 2024-05-30 at 09 44 49

Check in this screenshot which image is being displayed

cmd/build/v2/smartbuild/smartbuild.go
```diff
@@ -93,18 +95,20 @@ func (s *Ctrl) GetBuildHash(buildInfo *build.Info, service string) string {
...
-// if the built image belongs to global registry we clone it to the dev registry
-// so that in can be used in dev containers (i.e. okteto up)
-func (s *Ctrl) CloneGlobalImageToDev(image string) (string, error) {
-	if s.registryController.IsGlobalRegistry(image) {
-		s.ioCtrl.Logger().Debugf("Copying image '%s' from global to personal registry", image)
-		devImage, err := s.registryController.CloneGlobalImageToDev(image)
-		if err != nil {
-			return "", fmt.Errorf("error cloning image '%s': %w", image, err)
+// Clone clones the image specified into the target
+func (s *Ctrl) Clone(from string, buildInfo *build.Info) (string, error) {
```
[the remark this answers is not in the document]
jLopezbarb: Previous code was to clone from global to dev, so the calculus is different.

cmd/build/v2/smartbuild/smartbuild_test.go
```diff
@@ -185,70 +188,6 @@ func TestGetBuildHash(t *testing.T) {
 	assert.Equal(t, "hash", out)
 }
 
-func TestCloneGlobalImageToDev(t *testing.T) {
```
ifbyol, line 188: I think you need to change also the tests. It is no longer `TestClone`

ifbyol:
Discussed offline, there is 1 case that doesn't work as expected. If I set my own image with a specific name, it is being cloned to global with other name instead of using the 1 I specified

teresaromero (approved):
Tested couple scenarios, LGTM

ifbyol:
I found another issue while validating the changes. It is related to my previous message: https://github.com/okteto/okteto/pull/4270#issuecomment-2158617664

When I specify a custom image in the `image` field pointing to an Okteto dev registry image, e.g. `okteto.dev/<image-name>:<my-custom-tag>`, Smart Builds is not checking the right image. Instead of using the image name specified in `image` field, it is calculating the image name based on the service name. It should also check using the name specified in the `image` field

ifbyol (approved):

PR #4272, opened 2024-04-30
tests(fix): remove defer on tests

jLopezbarb:
# Proposed changes

Fixes Make explicit call to remove namespace. We have a bug that the defer was not executed on the last tests, leaving a namespace without deleting.

This change has one side effect. Now the tests that fails will not remove the ns that fails in order to look into it

## CLI Quality Reminders 🔧


maroshii (approved):
> Now the tests that fails will not remove the ns that fails in order to look into it

I think this is acceptable

andreafalzetti:
@jLopezbarb I've added the label to run the e2e tests so we can try the change

andreafalzetti (approved):
Approving, just check the e2e tests if they pass before merging 🙏

PR #4275, opened 2024-04-30
fix: exec args retrieval

jLopezbarb:
# Proposed changes

Fixes DEV-327

- Modify the function to return a string instead of a list of strings. This makes the executor execute the command correctly
- Modify the test in order to get only the part that is needed

With this change both use cases works:
Screenshot 2024-04-30 at 16 22 13

## How to validate

1. Run okteto exec api -- echo "this is a test"
2. See that it's printing this is a test
3. Run okteto exec api -- "echo this is a test"
4. See that it's printing this is a test

## CLI Quality Reminders 🔧

### DIFF
--- a/cmd/exec.go
+++ b/cmd/exec.go
@@ -20,6 +20,7 @@ import (
 	"os"
 	"time"
 
+	"github.com/kballard/go-shellquote"
 	contextCMD "github.com/okteto/okteto/cmd/context"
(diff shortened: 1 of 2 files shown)


jLopezbarb:
Closing in favour of following the same approach as [kubernetes/kubectl](https://github.com/kubernetes/kubectl)

PR #4277, opened 2024-05-02
Initialize ignore package

maroshii:
Creates a new generic `pkg/ignore` for managing okteto ignore files. The syntax proposed for the new `.oktetoignore` file is the following:

```.gitignore
.git

[deploy]
integration

[destroy]
frontend/*
backend

[test]
chart

[test.frontend]
backend
```

AFAIK, this is a adhoc syntax and not something that can be parsed using solely external libraries like viper, or others.

## Technical details

To parse this file we simply group them by section in memory and consider each section a `.dockerignore` file. We can then use [docker internal libraries](https://github.com/moby/patternmatcher/blob/main/ignorefile/ignorefile.go) when necessary to parse the ignored files (not in this PR).

This will be mostly use in the context of remote execution. For this we would simply need to append the values of a section (deploy, destroy, test) to to the `.dockerignore` file we already send to buildkit


maroshii:
@okteto/backend this PR prepares everything to be able to use a combined `.oktetoignore` file for deploy/destroy/test in remote executions.

```.gitignore
[deploy]
integration/*
test/*
[test.unit]
integration/*
```

```golang
/* pseudo code */
oktetoIgnoreFile := ".oktetoignore" // <------------ how to resolve this
ignorer := ignore.NewFromFile(oktetoIgnoreFile) 
rules, _ := ignorer.Rules("deploy")

params := &remote.Params{
  //...
  rules: rules,
}

runner.Run(ctx, params)
```

The only thing missing is how to resolve the `.oktetoignore` file which is somewhat tricky. For eg, if we define it at the root of the repo tests may have a sub directory as the build context so the file rules won't match. If want to resolve it hierarchically we need to build it and it may be very error prone given our poor handling of `cwd`. Not sure if we already have something like this in place.  cc @pchico83

pchico83:
> @okteto/backend this PR prepares everything to be able to use a combined `.oktetoignore` file for deploy/destroy/test in remote executions.
> 
> ```gitignore
> [deploy]
> integration/*
> test/*
> [test.unit]
> integration/*
> ```
> 
> ```go
> /* pseudo code */
> oktetoIgnoreFile := ".oktetoignore" // <------------ how to resolve this
> ignorer := ignore.NewFromFile(oktetoIgnoreFile) 
> rules, _ := ignorer.Rules("deploy")
> 
> params := &remote.Params{
>   //...
>   rules: rules,
> }
> 
> runner.Run(ctx, params)
> ```
> 
> The only thing missing is how to resolve the `.oktetoignore` file which is somewhat tricky. For eg, if we define it at the root of the repo tests may have a sub directory as the build context so the file rules won't match. If want to resolve it hierarchically we need to build it and it may be very error prone given our poor handling of `cwd`. Not sure if we already have something like this in place. cc @pchico83

@maroshii Let's start as simple as possible. I would only support `.oktetoignore` in the same path as the okteto manifest. And all paths would be relative to this location.
If a `test` container has a different `context`, `.oktetoignore` still refers to the location of the okteto manifest.
For example, you could have a `.oktetoignore` like:
```
[test.buildkit]
buildkit/*.jpg
```
and a test container like:
```
test:
   buildkit:
      context: buildkit
```

maroshii:
@pchico83 I don't think that's the simplest approach.

That configuration you posted won't work because the build uses the buildkit folder as the root context, so the ignore path would need to be `*.jpg` not `buildkit/*.jpg`. 
We would need to re-write all the rules to match the context as `cwd` but it's not very straightforward since that would affect the pattern matcher and rewrite the regex. We need to account things like absolute paths, negations, conditionals and nested folder and star (for eg: `/buildkit/*.jpg`, `build???/*.jpg`, `buildkit/buildkit/*.jpg`, `build*/*.jpg`, `!buildkit/README.md`).

https://docs.docker.com/build/building/context/#dockerignore-files

pchico83:
> @pchico83 I don't think that's the simplest approach.
> 
> That configuration you posted won't work because the build uses the buildkit folder as the root context, so the ignore path would need to be `*.jpg` not `buildkit/*.jpg`. We would need to re-write all the rules to match the context as `cwd` but it's not very straightforward since that would affect the pattern matcher and rewrite the regex. We need to account things like absolute paths, negations, conditionals and nested folder and star (for eg: `/buildkit/*.jpg`, `build???/*.jpg`, `buildkit/buildkit/*.jpg`, `build*/*.jpg`, `!buildkit/README.md`).
> 
> https://docs.docker.com/build/building/context/#dockerignore-files

@maroshii what about supporting a `.oktetoignore` file for every context? I mean, each `.oktetoignore` file should be located on the same path than the `context` of a given command

jLopezbarb:
How should I create the file if I want to have the same ignore file for two test units:

```.gitignore
[test.buildkit]
!e2e/buildkit
[test.depot]
!e2e/buildkit
```
Should we create a pattern in order to indicate that two has the same config:
```.gitignore
[test.buildkit/test.depot]
!e2e/buildkit
```

maroshii:
> what about supporting a .oktetoignore file for every context? I mean, each .oktetoignore file should be located on the same path than the context of a given command

+1 I think this would be the best option

pchico83:
> > what about supporting a .oktetoignore file for every context? I mean, each .oktetoignore file should be located on the same path than the context of a given command
> 
> +1 I think this would be the best option

@maroshii let's go with it then

maroshii:
Currently with the latest changes, `okteto deploy` and `okteto destroy` **only** red the `.oktetodeployignore` file and `okteto test` reads **only** the `.oktetoignore` file

pchico83:
> Currently with the latest changes, `okteto deploy` and `okteto destroy` **only** red the `.oktetodeployignore` file and `okteto test` reads **only** the `.oktetoignore` file

@maroshii can we make `deploy` and `destroy` to read `.oktetoignore`, if it exists, before `.oktetodeployignore`? The plan is to deprecate `.oktetodeployignore` in favor of `.oktetoignore`

pchico83:
> > @maroshii can we make `deploy` and `destroy` to read `.oktetoignore`, if it exists, before `.oktetodeployignore`? The plan is to deprecate `.oktetodeployignore` in favor of `.oktetoignore`
> 
> I don't think we should include it in this PR. This change is scoped to `okteto test` which won't be been released in the next cycle. I'll create a follow up ticket to add it before the next release

@maroshii that works, but for the June release I would combine both expiriences

andreafalzetti:
@maroshii I am trying to ignore `node_modules` from the following setup, but I am not able to. Am I doing something wrong?

```bash
okteto.yml
setup/
├─ node_modules/
├─ package.json
├─ package-lock.json
.oktetoignore
```

```yaml
# okteto.yml
test:
  setup:
    context: setup
    commands:
      - echo 1

deploy:
  - kubectl apply -f k8s.yml

dev:
  hello-world:
    image: okteto/golang:1
    command: bash
    sync:
      - .:/usr/src/app
    volumes:
      - /go
      - /root/.cache
    securityContext:
      capabilities:
        add:
          - SYS_PTRACE
    forward:
      - 2345:2345
```

```console
# .oktetoignore
[test]
node_modules
```

```bash
du -hs setup/node_modules
 78M	setup/node_modules
```

```bash
okteto test
 i  Using andreafalzetti @ product.okteto.dev as context
 i  'go-getting-started' was already deployed. To redeploy run 'okteto deploy'
 i  Executing 'setup'
⠇ Synchronizing context: 12.20MB
[interrupted] 
```

pchico83:
> @maroshii I am trying to ignore `node_modules` from the following setup, but I am not able to. Am I doing something wrong?
> 
> ```shell
> okteto.yml
> setup/
> ├─ Dockerfile
> ├─ node_modules/
> ├─ package.json
> ├─ package-lock.json
> .oktetoignore
> ```
> 
> ```yaml
> # okteto.yml
> test:
>   setup:
>     context: setup
>     commands:
>       - echo 1
> 
> deploy:
>   - kubectl apply -f k8s.yml
> 
> dev:
>   hello-world:
>     image: okteto/golang:1
>     command: bash
>     sync:
>       - .:/usr/src/app
>     volumes:
>       - /go
>       - /root/.cache
>     securityContext:
>       capabilities:
>         add:
>           - SYS_PTRACE
>     forward:
>       - 2345:2345
> ```
> 
> ```yaml
> # .oktetoignore
> [test]
> node_modules
> ```
> 
> ```yaml
> # setup/Dockerfile
> FROM alpine
> ```
> 
> ```shell
> du -hs setup/node_modules
>  78M	setup/node_modules
> ```
> 
> ```shell
> okteto test
>  i  'go-getting-started' was already deployed. To redeploy run 'okteto deploy'
>  i  Executing 'setup'
> ⠇ Synchronizing context: 12.20MB
> [interrupted] 
> ```

@andreafalzetti the `.oktetoignore` file must be located in the `context` folder. In this case, `setup/.oktetoignore`

pchico83:
> @pchico83 I tried moving it into `setup` initially didn't skip `node_modules`. I had to change `[test]` to `[test.setup]`. Is it expected?

@andreafalzetti nope, [test] should apply to all test containers. and [test.setup] only to that particular test container

PR #4289, opened 2024-05-10
okteto exec to work the same as kubectl exec

jLopezbarb:
DEV-327

# Proposed changes

In this PR we've addressed that `okteto exec` was behaving different to `kubectl exec`. Also we didn't have any test for that command, which now we can have. 

- Use the same pre-processing as kubernetes for `kubectl exec` and `okteto exec`. We still maintain the selector in case the user has more than one dev environment and doesn´t add it to the arguments
- Add unit tests for the whole command

## How to validate

1. Run `okteto up` on any okteto configured repository 
2. Run `okteto exec name -- echo test` and check that is working
3. Run `okteto exec name -- "echo test"` and observe an error

## CLI Quality Reminders 🔧


andreafalzetti:
Great work 👏 Code changes LGTM. I've tested different scenarios and everything worked besides the following:

1. Using: https://github.com/andreafalzetti/hybrid-test
2. `okteto up`
3. once the up session is up & running in another terminal run: `okteto exec hybridapp -l debug -- time`
4. notice how the command "time" is not executed. I've tried with `touch file.txt` too and it seems like it's a problem of not executing the command instead of a problem of "output".
5.

andreafalzetti (commented):
Great work 👏 Code changes LGTM. I've tested different scenarios and everything worked for me besides the following:

1. Using: https://github.com/andreafalzetti/hybrid-test
2. `okteto up`
3. once the up session is up & running in another terminal run: `okteto exec hybridapp -l debug -- time`
4. notice how the command "time" is not executed. I've tried with `touch file.txt` too and it seems like it's a problem of not executing the command instead of a problem of "output".

**Update**

As we noticed offline, it only seems to be related to `time`. Other commands execute corretly.

andreafalzetti (approved):

cmd/exec/mixpannel.go
```diff
@@ -0,0 +1,36 @@
+// Copyright 2024 The Okteto Authors
```
ifbyol, line 1: Nit: Single `n` in the filename? mixpanel vs mixpannel

cmd/exec.go
```diff
@@ -20,6 +20,7 @@ import (
 	"os"
 	"time"
 
+	"github.com/kballard/go-shellquote"
```
ifbyol, line 23: I think this file is not being used anymore. Now everything goes through `exec` package, right?

cmd/exec/cmd.go
```diff
@@ -0,0 +1,157 @@
...
+	k8sClientProvider okteto.K8sClientProvider
+	mixpannelTracker  metadataTracker
+
+	getExecutorFunc func(dev *model.Dev, podName string) (executor, error)
+	onExecFinish    []onExecFinish
+}
+
+// NewExec creates a new exec command
+func NewExec(fs afero.Fs, ioCtrl *io.Controller) *Exec {
+	k8sProvider := okteto.NewK8sClientProvider()
```
ifbyol, line 58: Ideally, this should come from `main.go` too. That is something common to all commands that can be created just once

cmd/exec/cmd.go
```diff
@@ -0,0 +1,157 @@
...
+	mixpannelTracker := &mixpannelTrack{
```
ifbyol, line 59: mixpanel? (single `n` instead of double `n`)

cmd/exec/cmd.go
```diff
@@ -0,0 +1,157 @@
...
+		ioCtrl:            ioCtrl,
+		fs:                fs,
+		k8sClientProvider: k8sProvider,
+		appRetriever:      newAppRetriever(ioCtrl, k8sProvider),
+		mixpannelTracker:  mixpannelTracker,
+		onExecFinish: []onExecFinish{
+			mixpannelTracker.Track,
+		},
+	}
+	e.getExecutorFunc = e.getExecutor
```
ifbyol, line 72: Maybe I'm missing something, but I don't get why this? `getExecutor` is already a function of `e`, why do we need to set a field of the struct with that function?

cmd/exec/cmd.go
```diff
@@ -0,0 +1,157 @@
...
+This command allows you to run tools or perform operations directly on a dev environment that is already running.
+For more information on managing pods, refer to the okteto documentation: https://www.okteto.com/docs/`,
+		Example: `# Run the 'echo this is a test' command inside the pod named 'my-pod'
+okteto exec my-pod -- echo this is a test
+
+# Get an interactive shell session inside the pod named 'my-pod'
+okteto exec my-pod`,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			manifestOpts := contextCMD.ManifestOptions{Filename: execFlags.manifestPath, Namespace: execFlags.namespace, K8sContext: execFlags.k8sContext}
+			manifest, err := contextCMD.LoadManifestWithContext(ctx, manifestOpts, e.fs)
```
ifbyol, line 94: Does this take into consideration the env vars for context and namespace after the refactor done by andrea?

cmd/exec/options.go
```diff
@@ -0,0 +1,99 @@
...
+import (
+	"errors"
+	"fmt"
+
+	"github.com/okteto/okteto/cmd/utils"
+	"github.com/okteto/okteto/pkg/log/io"
+	"github.com/okteto/okteto/pkg/model"
+)
+
+type Options struct {
```
ifbyol, line 25: Nit: If this is not being used outside this package, we might not want to export it

cmd/exec/cmd.go
```diff
@@ -0,0 +1,157 @@
...
+	e.mixpannelTracker.SetMetadata(&analytics.TrackExecMetadata{
+		Mode:               dev.Mode,
+		FirstArgIsDev:      opts.firstArgIsDevName,
+		Success:            err == nil,
+		IsOktetoRepository: utils.IsOktetoRepo(),
+		IsInteractive:      dev.IsInteractive(),
+	})
+	for _, f := range e.onExecFinish {
+		f()
+	}
```
ifbyol, line 155: This is a bit odd. I mean, it feels odd that we call directly to mixpanelTracker to set some metadata to be sent, but then, we don't use it directly to track the information. 

2 alternatives to this:
* As we only have 1 tracker or just one operation to do when the command ends, I think we can remove the `onExecFinish` field in the struct and just call to the mixpanelTracker
* functions within `onExecFinish` field receive the "metadata" to be sent and remove the dependency with `mixpanelTracker` in the `ExecCommand`

With any of the cases we can remove one dependency from `Exec` struct

ifbyol (commented):
There is nothing blocker but some comments would be good to addres

cmd/exec/cmd.go
jLopezbarb: Yes, I just tested and its working correctly with the latest changes done

ifbyol:
I have been testing it and I found small things to improve or thing that have changed.

1. When you don't specify a dev env where to execute the command, before, we were displaying the selector with just the dev environments in dev mode. Now, we display all the devs (in that screen, the terminal in the bottom is current CLI version and the other one is the version of this branch). If you select a dev which is not in dev mode, it fails (which is fine), but I think we should display only the services that are in dev mode

Screenshot 2024-05-24 at 13 19 39

2. When a command fails the execution, we are displaying the error message (the red one) in the same line that previous errors, Could we just add a new line to show the red message in a new line? This is not new, but if it is something easy to fix

Screenshot 2024-05-24 at 13 25 08

3. When I execute `okteto exec` without specifying a service (and there are several services) and without specifying a command, I have to select the dev env first, to then, fail with an error indicating that a command is required. Could we do the validation first instead of making the user to select a dev to then, fail?

ifbyol (commented):
Do we have some e2e test for the `exec` command?

jLopezbarb:
> Do we have some e2e test for the `exec` command?

Not right now, I'll create a simple one for this

ifbyol:
The error message when I try to execute a command for a non-existing dev container has changed with a less meaningful error and we are missing the hint. Is that expected? 

With your changes

Screenshot 2024-05-27 at 10 55 49

Without your changes

Screenshot 2024-05-27 at 10 56 46

We should not change that unless it is strictly necessary. As it was before it is giving better DX and giving hints to the user on why it failed and how it can be solved

ifbyol:
Something similar happens when you don't specify a dev, but there are no devs on dev mode

With these changes

Screenshot 2024-05-27 at 11 00 24

Without these changes

Screenshot 2024-05-27 at 10 59 48

Specially, this error `Failed to set dev from manifest` is not meaningful and can be misleading for users

PR #4295, opened 2024-05-13
Add ignore root section to .oktetoignore

maroshii:
Include root section of the `.oktetoignore` file for `okteto test`

Prior to this change, in this example the `.git` folder was not being included in the rules:

```
.git
[test]
manifests
[test.frontend]
backend
```


andreafalzetti (approved):

```
test:
  unit:
    context: .
    commands:
      - echo ".git dir size ="
      - du -hs .git
```


```
du: cannot access '.git': No such file or directory
 x  Error on stage unit:  x  Exit status 1
```

while with `2.27.1`:

```
 i  Running 'echo ".git dir size ="'
.git dir size =
 i  Running 'du -hs .git'
936K	.git
```

PR #4298, opened 2024-05-20
Include all manifest properties to "friendly err" util + Add okteto test alias

andreafalzetti:
Fixes DEV-349

While validating `okteto test` I've noticed that the CLI would not improve the unmarshalling error messages like it does for other sections. After reviewing, I've noticed two things:

1. the `okteto test` unmarshall uses an `alias` to avoid recursion, this is a struct not known in `model.Manifest` so it's not handled
2. the property `Commands` of Test (alias) was missing, and after reviewing more in depth, I've noticed that all properties that had child props, would be missing because of a wrong condition.

This PR fixes the 2 scenarios above.

## How to validate

Following, some possible real-case scenarios that user might experience:

### Incorrect sub-property of `test.<name>` (command -> commands)

```yaml
test:
  unit:
    context: unit
    command:
      - name: Run unit tests
        command: echo 1
```

**2.27.1**
Screenshot 2024-05-20 at 11 10 21

**NEW**
Screenshot 2024-05-20 at 11 08 58

### Incorrect sub-property of `test.<name>.commands` (commands -> command)

```yaml
test:
  unit:
    context: unit
    commands:
      - name: Run unit tests
        commands: echo 1
```

**2.27.1**

Screenshot 2024-05-20 at 11 20 33

**NEW**

Screenshot 2024-05-20 at 11 25 41

## CLI Quality Reminders 🔧


pkg/model/schema.go
```diff
@@ -109,15 +109,16 @@ func getStructKeys(t interface{}) map[string][]string {
...
 			}
-		} else {
-			yamlTag := field.Tag.Get("yaml")
-			if yamlTag != "" && yamlTag != "-" {
-				parts := strings.Split(yamlTag, ",")
-				if len(parts) > 0 {
-					result[structFullName] = mergeAndSortUnique(result[structFullName], []string{parts[0]})
-				}
+		}
+
```
ifbyol, line 113: Why removing the else? Does it have to be applied always now?

andreafalzetti: this loop iterates over each property, then based on the type, it may generate a recursion, but it was only adding child properties, basically ignoring itself, resulting in missing parent properties

pkg/model/test.go
```diff
@@ -34,8 +34,8 @@ func (test *Test) expandEnvVars() error {
 }
 
 func (t *Test) UnmarshalYAML(unmarshal func(interface{}) error) error {
-	type alias Test
-	var tt alias
+	type testAlias Test
```
andreafalzetti, line 37: I renamed the two different `alias` types into `testAlias` and `testAliasCommand` so that we can differenciate between them, otherwise we always get `model.alias` in the unmarshalling error

ifbyol (approved):

jLopezbarb (approved):

PR #4302, opened 2024-05-20
Partial implementation of artifacts for test command

maroshii:
## Proposed changes

Adds `artifacts` to test command that allows to export files to the local file system after buildkit execution:

```yaml
test:
  echo:
    artifacts:
      - a
      - b
      - c
      - d
    commands:
      - mkdir a b c d
      - echo "A" > a/1.txt
      - echo "B" > b/2.txt
      - echo "C" > c/3.txt
      - echo "D" > d/4.txt
      - ls -l
```

The above manifest will export the following the local filesystem:

```
./.okteto-output/
├── a
│   └── 1.txt
├── b
│   └── 2.txt
├── c
│   └── 3.txt
└── d
    └── 4.txt
```

### Open Questions

- Do we want to write to `.okteto-output` or do we rather just right relative to the cwd?

---

# :warning: CAVEAT :warning:

This was marked as a _partial_ implementation because it only works for successful builds. If the build fails nothing will get written. I'm not sure if buildkit has a way to export regardless of failure but this is still a big unknown for this feature. [The docs for local export](https://github.com/moby/buildkit?tab=readme-ov-file#local-directory) has no info on this and assumes that the build succeeds. I'll keep investigating cc @pchico83

### DIFF
--- a/pkg/remote/remote_test.go
+++ b/pkg/remote/remote_test.go
@@ -373,6 +373,7 @@ ARG OKTETO_GIT_COMMIT
 ARG OKTETO_GIT_BRANCH
 ARG OKTETO_INVALIDATE_CACHE
 
+RUN echo "$OKTETO_INVALIDATE_CACHE" > /var/lib/.oktetocachekey
 RUN okteto registrytoken install --force --log-output=json
 
 RUN \
@@ -380,6 +381,10 @@ RUN \
   --mount=type=secret,id=known_hosts --mount=id=remote,type=ssh \
   mkdir -p $HOME/.ssh && echo "UserKnownHostsFile=/run/secrets/known_hosts" >> $HOME/.ssh/config && \
   /okteto/bin/okteto remote-run deploy --log-output=json --server-name="$INTERNAL_SERVER_NAME" --name "test"
+
+FROM scratch
+COPY --from=runner /var/lib/.oktetocachekey .oktetocachekey
+
 `,
 				buildEnvVars:      map[string]string{"OKTETO_BUIL_SVC_IMAGE": "ONE_VALUE", "OKTETO_BUIL_SVC2_IMAGE": "TWO_VALUE"},
 				dependencyEnvVars: map[string]string{"OKTETO_DEPENDENCY_DATABASE_VARIABLE_PASSWORD": "dependency_pass", "OKTETO_DEPENDENCY_DATABASE_VARIABLE_USERNAME": "dependency_user"},


andreafalzetti:
Code changes LGTM. About the failing build I have no suggestions yet, but I've done some initial tests, so far so good, only one odd behaviour perhaps with the following scenario, I think due to cache, when I export `/` it also gives me artifacts from previous builds. Sharing in case it's a behaviour that we don't want? Maybe it's confusing?

```yaml
test:
  unit:
    context: unit
    artifacts:
      - /
    commands:
      - name: Run unit tests
        command: mkdir report && echo hola > report/hola.txt && echo ciao > report/ciao.txt
```

```bash
okteto test

# i  Using ******@ ***** as context
# i  'go-getting-started' was already deployed. To redeploy run 'okteto deploy'
# i  Executing 'unit'


ls -la .okteto-output
drwx------   4 andrea  staff  128 20 May 22:31 .
drwxr-xr-x  19 andrea  staff  608 20 May 22:31 ..
-rw-r--r--   1 andrea  staff    5 20 May 09:26 hola.txt
drwxr-xr-x   4 andrea  staff  128 20 May 22:28 report

ls -la .okteto-output/report
drwxr-xr-x  4 andrea  staff  128 20 May 22:28 .
drwx------  4 andrea  staff  128 20 May 22:31 ..
-rw-r--r--  1 andrea  staff    5 20 May 22:28 ciao.txt
-rw-r--r--  1 andrea  staff    5 20 May 22:28 hola.txt
```

I've tried removing the `.okteto-output` but I still get the `hola.txt` (look at the different time as well).

pchico83:
I think exporting only on success is okay for this iteration. I don't think buildkit will help, we would need to always return success at the buildkit level and manage the error on our side.
About `.okteto-output,` I think the experience is better if we `write relative to the cwd`. That helps, for example, if you do things like `go mod tidy`. In general, adopting the Okteto Test requires fewer changes if we use the current path. And we can always make this path optional if it becomes an issue

jLopezbarb:
Should we support something like what CircleCI does to export it to a custom local path:
```
- store_artifacts:
          path: /tmp/artifact-1
          destination: artifact-file
```

I think it's ok to have a default destination but I think the user could want to export to a specific file that is already in the `.oktetoignore`. WDYT?

jLopezbarb:
Tested and working correctly. Also tested removing a file and checked if the cache would regenerate the file. So despite of the question that I asked I think the code and behaviour is ok

pchico83:
> > I don't think buildkit will help, we would need to always return success at the buildkit level and manage the error on our side.
> 
> Yeah, agree that's the way to go. It will have an impact on buildkit metrics tho since once we change this all builds will succeed

@maroshii we don't rely on buildkit for metrics, the build status is reported by the Okteto CLI. We would also have control on the metrics generated by failed builds

pchico83:
> * Extended syntax for path and destination in artifact section is supported

@maroshii could you explain the new syntax with a few samples if needed?

andreafalzetti:
as discussed offline, with the latest changes in this PR, the logs of the test commands are not visibile. Example

```
test:
  load:
    image: artilleryio/artillery:latest
    context: .
    commands:
      - name: "Say Hi"
        command: echo hiii && exit 1
```

With `2.27.21` I get:

```
 i  Using ** @ ** as context
 i  'tmp.RMs0oBYF6M' was already deployed. To redeploy run 'okteto deploy'
 i  Executing 'load'
 i  Running stage 'load'
 i  Running 'Say Hi'
hiii
 x  Error on stage load:  x  Exit status 1
```

With this build:

```
 i  Using ** @ ** as context
 i  'tmp.RMs0oBYF6M' was already deployed. To redeploy run 'okteto deploy'
 i  Executing 'load'
```

jLopezbarb (approved):

PR #4304, opened 2024-05-21
fix: warning displayed even when the file was invalid

jLopezbarb:
# Proposed changes

Fixes DEV-347

This PR only shows the warning if the stack file is a valid stack. We're losing the warning on non valid compose but we are reducing the noise for other file types

## How to validate

1. Create the following `okteto.yml`:
```yaml
deploy
  - kubectl apply -f k8s.yml

dev:
  hello-world:
    image: okteto/golang:1
    command: bash
    sync:
      - .:/usr/src/app
    volumes:
      - /go
      - /root/.cache
    securityContext:
      capabilities:
        add:
          - SYS_PTRACE
    forward:
      - 2345:2345
```
1. Run `okteto deploy`
1. See that the error is not shown

## CLI Quality Reminders 🔧

### DIFF
--- a/pkg/model/stack.go
+++ b/pkg/model/stack.go
@@ -798,14 +798,20 @@ func isFileCompose(path string) bool {
 		return true
 	}
 
-	stackDeprecationWarningOnce.Do(func() {
-		if !isComposeFileName {
-			oktetoLog.Warning(`Okteto Stack syntax is deprecated.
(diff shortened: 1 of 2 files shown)


pkg/model/stack.go
```diff
@@ -842,6 +848,9 @@ func LoadStack(name string, stackPaths []string, validate bool, fs afero.Fs) (*S
 			return nil, err
 		}
 	}
+	for _, path := range resultStack.Paths {
```
andreafalzetti, line 851: `resultStack` can potentially be `nil`

andreafalzetti (approved):

PR #4307, opened 2024-05-21
DEV-335 Repository is not being calculated correctly for Smart Builds when the manifest is within a subdirectory of a repository

ifbyol:
# Proposed changes

Fixes DEV-335

When calculating the git repository for Smart Builds we were not checking parent folders to get git metadata and extract the remote URL. This was generating that for example, if you have the following structure:

```
root/
  |- subdirA/
         |- okteto.yml
```

And you execute `okteto deploy` from `subdirA`, image were being rebuilt every time because we don't extract git metadata correctly. 

Also, the filename stored in the configmap wasn't right because we were storing empty or `okteto.yml` instead of storing the file relative to the root of the repo, so redeploy or destroy from the UI are not working correctly. **Currently, we just store the filename if it was specified by the user using `-f` flag.**

In other part of the code, we were calculating the repository checking parent folders, but in a different way, so I modified both places to use the same logic. In this case, we use the property `DetectDotGit` and pass it to `git.PlainOpenWithOptions`.

Starting to calculate the git repository generates that the dev environment name inferred might be different, as the git repository has preference over the folder name.

## How to validate

Execute deploys and build operations in different scenarios:
* From the root of a repo having the manifest in the root without specifying it with `-f` (no filename is stored in the configmap)
* From the root of a repo having the manifest in the root specifying it with `-f`
* From the root of a repo having the manifest in a subfolder specifying it with `-f`
* From the root of a repo having a compose in the root without specifying it with `-f` (no filename is stored in the configmap)
* From the root of a repo having a compose in the root specifying it with `-f`
* From the root of a repo having a compose in a subfolder specifying it with `-f`
* From a subfolder of a repo having the manifest in the same folder without specifying it with `-f` (no filename is stored in the configmap)
* From a subfolder of a repo having the manifest in the same folder specifying it with `-f`
* From a subfolder of a repo having the manifest in a subfolder specifying it with `-f`
* From a subfolder of a repo having a compose in the same folder without specifying it with `-f` (no filename is stored in the configmap)
* From a subfolder of a repo having a compose in the same folder specifying it with `-f`
* From a subfolder of a repo having a compose in a subfolder specifying it with `-f`
* From a subfolder of a repo having a manifest in the root specifying it with `-f`
* From a subfolder of a repo having a compose in the root specifying it with `-f`
* From a folder with no repository having a manifest
* From a folder with no repository having a compose

For all the cases with repositories, make sure that redeploy works fine. Also, check that the filename stored in the configmap is the correct one when using the `-f` option (otherwise, the filename is not stored)

## CLI Quality Reminders 🔧


pkg/cmd/stack/translate_test.go
```diff
@@ -120,8 +120,9 @@ func Test_translateDeployment(t *testing.T) {
 	}
 	assert.Equal(t, result.Labels, labels)
 	annotations := map[string]string{
-		"annotation1": "value1",
-		"annotation2": "value2",
+		"annotation1":           "value1",
+		"annotation2":           "value2",
+		"dev.okteto.com/sample": "true",
```
ifbyol, line 125: This (and the rest of changes in this file) is needed because, as the tests run within `okteto/okteto` repository, it is being detected and added the annotation because it is considered an Okteto sample. I tried to change it, but there is no easy way to do it as it uses current working directory internally

pkg/model/stack_test.go
```diff
@@ -1153,8 +1157,10 @@ func Test_getStackName(t *testing.T) {
 
 	for _, tt := range tests {
 		t.Run(tt.testName, func(t *testing.T) {
+			dir := t.TempDir()
```
ifbyol, line 1160: This is needed to prevent the test to detect `okteto/okteto` repository

andreafalzetti (approved):

PR #4309, opened 2024-05-22
Validate Test section to avoid panic if test section is not populated properly

andreafalzetti:
# Proposed changes

Fixes DEV-348

In this PR we're adding a validation layer for the Test section of the Okteto Manifest. Currently, passing empty section, or empty test throws panics.

Important Note: Curently, if no tests are configured, the Okteto CLI will just deploy the application. But there is no logs or anything that informs the user that "nothing" has been done test-wise. With this change we make sure they are informed if they run `test` without configuring them. It also exit non-zero, so if this was intended to always exit 0 (for CI reasons for example), I can make it a warning.

## How to validate

Using the following manifests, and run `okteto test`

**Missing `test` section**
 
```
dev:
  hello-world:
    image: okteto/golang:1
    command: bash
    sync:
      - .:/usr/src/app
    volumes:
      - /go
      - /root/.cache
    securityContext:
      capabilities:
        add:
          - SYS_PTRACE
    forward:
      - 2345:2345
```

**Empty `test` section**
 
```
test:

dev:
  hello-world:
    image: okteto/golang:1
    command: bash
    sync:
      - .:/usr/src/app
    volumes:
      - /go
      - /root/.cache
    securityContext:
      capabilities:
        add:
          - SYS_PTRACE
    forward:
      - 2345:2345
```

**Empty test**
 
```
test:
  unit:

dev:
  hello-world:
    image: okteto/golang:1
    command: bash
    sync:
      - .:/usr/src/app
    volumes:
      - /go
      - /root/.cache
    securityContext:
      capabilities:
        add:
          - SYS_PTRACE
    forward:
      - 2345:2345
```

## CLI Quality Reminders 🔧

### DIFF
--- a/pkg/model/test.go
+++ b/pkg/model/test.go
@@ -10,9 +10,14 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
+
 package model
 
-import "github.com/okteto/okteto/pkg/env"
+import (
+	"fmt"
+
+	"github.com/okteto/okteto/pkg/env"
+)
 
 type Test struct {
 	Image     string        `yaml:"image,omitempty"`
@@ -20,6 +25,31 @@ type Test struct {
 	Commands  []TestCommand `yaml:"commands,omitempty"`
 	DependsOn []string      `yaml:"depends_on,omitempty"`
 	Caches    []string      `yaml:"caches,omitempty"`
+}
+
+var (
+	// TODO: add link to docs when available
+	errNoTestsDefined = fmt.Errorf("there are no tests configured in your Okteto Manifest")
+)
+
+func (test ManifestTests) Validate() error {
+	if test == nil {
+		return errNoTestsDefined
+	}
+
+	hasAtLeastOne := false
+	for _, t := range test {
+		if t != nil {
+			hasAtLeastOne = true
+			break
+		}
+	}
+
+	if !hasAtLeastOne {
+		return errNoTestsDefined
+	}
+
+	return nil
 }
 
 func (test *Test) expandEnvVars() error {
--- a/cmd/test/cmd.go
+++ b/cmd/test/cmd.go
@@ -10,6 +10,7 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
+
 package test
 
 import (
@@ -160,6 +161,10 @@ func doRun(ctx context.Context, servicesToTest []string, options *Options, ioCtr
 
 	manifest, err := model.GetManifestV2(options.ManifestPath, fs)
 	if err != nil {
+		return analytics.TestMetadata{}, err
+	}
+
+	if err := manifest.Test.Validate(); err != nil {
 		return analytics.TestMetadata{}, err
 	}
 


ifbyol:
> Important Note: Curently, if no tests are configured, the Okteto CLI will just deploy the application. But there is no logs or anything that informs the user that "nothing" has been done test-wise. With this change we make sure they are informed if they run test without configuring them. It also exit non-zero, so if this was intended to always exit 0 (for CI reasons for example), I can make it a warning.

What other tools or languages behaves when no tests are defined? I mean, for example image I execute `go test` in a directory without tests. Does it fails or just print no tests were run and that's it?

andreafalzetti:
> What other tools or languages behaves when no tests are defined? I mean, for example image I execute `go test` in a directory without tests. Does it fails or just print no tests were run and that's it?

I you run `go test` in an empty directory, it fails with: `exit 1`: `go: go.mod file not found in current directory or any parent directory; see 'go help modules'`.

Once you init the project (`go mod init..`) it fails `exit 1`: `no Go files in /var/folders/nt/ggnw9z4d1712sftc8606lj7r0000gn/T/tmp.j2E7UpOEcD`

pkg/model/test.go
```diff
@@ -22,6 +27,31 @@ type Test struct {
...
+
+func (test ManifestTests) Validate() error {
+	if test == nil {
+		return errNoTestsDefined
+	}
+
+	hasAtLeastOne := false
+	for _, t := range test {
+		if t != nil {
+			hasAtLeastOne = true
```
jLopezbarb, line 43: I think this code could be simplified by removing the bool and returning nil as soon as we detect there is one without the break. WDYT?
```suggestion
	for _, t := range test {
		if t != nil {
			return nil
		}
	}
```

andreafalzetti: @jLopezbarb I personally find more readable and easier to maintain the current code, because having `return nil` means we cannot add additional validations after this check. If we wanrt to add one, we need to refactor and add the bool. Your suggestion uses less memory but in our context, I think the benefits are minimal compared to keeping the code easy to change. WDYT?

jLopezbarb:
@andreafalzetti I know that is out of scope of this PR but I also found one more panic that maybe we could fix it on this PR.
Steps to reproduce:
1. Using the following `okteto.yml`:
```yaml
build:
  backend:
    context: .
test:
  backend:
    image: okteto/golang:1
    context: backend
    commands:
      - name: command1
        command: echo "this is command 1"
      - name: command2
        command: echo "this is command 2"
```
2. Run `okteto test`

andreafalzetti:
> ```yaml
> build:
>   backend:
>     context: .
> test:
>   backend:
>     image: okteto/golang:1
>     context: backend
>     commands:
>       - name: command1
>         command: echo "this is command 1"
>       - name: command2
>         command: echo "this is command 2"
> ```

Are you sure? Can you double check your okteto version and share the stack trace? With that manifest I cannot reproduce it.

andreafalzetti:
> @andreafalzetti I know that is out of scope of this PR but I also found one more panic that maybe we could fix it on this PR. Steps to reproduce:
> 
>     1. Using the following `okteto.yml`:
> 
> 
> ```yaml
> build:
>   backend:
>     context: .
> test:
>   backend:
>     image: okteto/golang:1
>     context: backend
>     commands:
>       - name: command1
>         command: echo "this is command 1"
>       - name: command2
>         command: echo "this is command 2"
> ```
> 
>     2. Run `okteto test`

Good catch! As discussed offline, we'll tackle that case in DEV-359

PR #4311, opened 2024-05-22
remove unused code (build)

andreafalzetti:
Removing unused code:

- basic builder `NewBuilderFromScratch` is never called
- `getServiceShaInCache` is never called (only in tests)
- `getImageReferencesForTagWithDefaults`  is never called (only in tests)

### DIFF
--- a/cmd/build/basic/build.go
+++ b/cmd/build/basic/build.go
@@ -21,12 +21,10 @@ import (
 
 	"github.com/okteto/okteto/cmd/utils"
 	"github.com/okteto/okteto/pkg/analytics"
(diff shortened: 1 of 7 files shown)


andreafalzetti:
@ifbyol did you leave the basic builder `NewBuilderFromScratch` on purpose for some reasons? Or shall we remove it given it's not used?

ifbyol:
> @ifbyol did you leave the basic builder `NewBuilderFromScratch` on purpose for some reasons? Or shall we remove it given it's not used?

I think it was being used in the first refactor but probably usages were removed in other changes. It can be removed

ifbyol (approved):

PR #4312, opened 2024-05-23
fix panic when inferring stack with empty image

andreafalzetti:
# Proposed changes

Fixes DEV-368

This PR fixes a panic that is caused when using Docker Compose and Okteto Manifests together and the `Image` field of the service is empty. For example, when running `okteto down` or `okteto destroy` the image is not built, which may cause the image field to be empty, in case the compose rely on built-in Okteto Variables such as `${OKTETO_BUILD_<SERVICE>_IMAGE}`

This is a panic that was introduced in `2.27.x` because of recent changes.

## How to validate

1. Using https://github.com/okteto/movies-with-compose
1. Change `services.app.image` to `${OKTETO_BUILD_APP_IMAGE}`
3. Create the following `okteto.yml`
```
build:
  app:
    context: api
deploy:
  compose: docker-compose.yml
```
4. Run `okteto build` and observe the panic
5. Now repeat with the new build and validate that the panic does not occurr
6. Now test that `deploy` + `up` + `down` and `destroy` work as expected and cause no panics.

## CLI Quality Reminders 🔧

### DIFF
--- a/pkg/model/manifest.go
+++ b/pkg/model/manifest.go
@@ -35,7 +35,7 @@ import (
 	oktetoLog "github.com/okteto/okteto/pkg/log"
 	"github.com/okteto/okteto/pkg/model/forward"
 	"github.com/spf13/afero"
-	yaml "gopkg.in/yaml.v2"
+	"gopkg.in/yaml.v2"
 	yaml3 "gopkg.in/yaml.v3"
 )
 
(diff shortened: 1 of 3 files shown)


pkg/model/manifest.go
```diff
@@ -35,7 +35,7 @@ import (
 	oktetoLog "github.com/okteto/okteto/pkg/log"
 	"github.com/okteto/okteto/pkg/model/forward"
 	"github.com/spf13/afero"
-	yaml "gopkg.in/yaml.v2"
```
andreafalzetti, line 38: goland insists in removing this, which is redundant, happy to revert it if we prefer so.

ifbyol (approved):

PR #4315, opened 2024-05-23
Make sure /var/lib exists in remote runner image

maroshii:
When using a custom image for deploy, destroy or test the directory `/var/lib` may not exist. We need to make sure it exists or the redirection fails. This was caught by integration tests

### DIFF
--- a/pkg/remote/remote.go
+++ b/pkg/remote/remote.go
@@ -91,7 +91,7 @@ ARG {{ .GitCommitArgName }}
 ARG {{ .GitBranchArgName }}
 ARG {{ .InvalidateCacheArgName }}
 
-RUN mkdir -p /var/lib && echo "${{ .InvalidateCacheArgName }}" > /var/lib/.oktetocachekey
+RUN echo "${{ .InvalidateCacheArgName }}" > /etc/.oktetocachekey
 RUN okteto registrytoken install --force --log-output=json
 
 RUN \
@@ -106,7 +106,7 @@ FROM scratch
 COPY --from=runner /okteto/src/{{$artifact.Path}} {{$artifact.Destination}}
 {{ end }}
 {{ else -}}
-COPY --from=runner /var/lib/.oktetocachekey .oktetocachekey
+COPY --from=runner /etc/.oktetocachekey .oktetocachekey
 {{ end }}
 `
 )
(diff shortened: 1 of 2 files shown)


pkg/remote/remote.go
```diff
@@ -91,7 +91,7 @@ ARG {{ .GitCommitArgName }}
 ARG {{ .GitBranchArgName }}
 ARG {{ .InvalidateCacheArgName }}
 
-RUN echo "${{ .InvalidateCacheArgName }}" > /var/lib/.oktetocachekey
+RUN mkdir -p /var/lib && echo "${{ .InvalidateCacheArgName }}" > /var/lib/.oktetocachekey
```
pchico83, line 94: why do we need to write `. InvalidateCacheArgName` to `/var/lib/.oktetocachekey`?

pkg/remote/remote_test.go
```diff
@@ -373,7 +373,7 @@ ARG OKTETO_GIT_COMMIT
 ARG OKTETO_GIT_BRANCH
 ARG OKTETO_INVALIDATE_CACHE
 
-RUN echo "$OKTETO_INVALIDATE_CACHE" > /var/lib/.oktetocachekey
+RUN mkdir -p /var/lib && echo "$OKTETO_INVALIDATE_CACHE" > /var/lib/.oktetocachekey
```
pchico83, line 376: `mkdir -p /var/lib` might fail, especially in not root images. I would avoid writing to `/var/lib` if possible, but if there is no better option, we should give a proper error if this command fails

jLopezbarb (approved):

pkg/remote/remote_test.go
pchico83: But that one is in the `$HOME`, which the user usually have access to it

pkg/remote/remote.go
pchico83: Why is it relevant to if the resulting image is scratch or not? we never push it
I think setting the `target` of the build request if there are no artifacts would be a better approach than writing to `/var/lib/.oktetocachekey`

pkg/remote/remote_test.go
pchico83: Since we already depend on $HOME, I would say `$HOME/oktetocachekey`

pchico83 (approved):

PR #4316, opened 2024-05-23
Use .oktetoignore in deploy and destroy remote

maroshii:
We now inject rules for destroy and deploy remote a we do for test. This is merged with the contents of `.oktetodeployignore` file.

Updated docs as part of this: https://github.com/okteto/docs/pull/743

# Validation

Create the following folders:
```bash
mkdir report-never && echo never > report-never/never
mkdir report-deploy && echo deploy > report-deploy/deploy
mkdir report-destroy && echo destroy > report-destroy/destroy
```

Using the following `.oktetoignore` file:

```
report-never

[deploy]
report-destroy

[destroy]
report-deploy
```

with an `ls` in the deploy and destroy sections:
```
deploy:
  - ls

destroy:
  - ls
```

For deploy we should only get `report-deploy` and for destroy we should only get `report-destroy`

### DIFF
--- a/pkg/remote/remote.go
+++ b/pkg/remote/remote.go
@@ -473,12 +473,12 @@ func createDockerignoreFileWithFilesystem(cwd, tmpDir string, rules []string, us
 			if !errors.Is(err, os.ErrNotExist) {
 				return err
 			}
-			oktetoLog.Warning("Ignoring files through %s is deprecated and will be removed in future versions. Please use .oktetoignore. More info here: https://www.okteto.com/docs/core/remote-execution/#ignoring-files", oktetoDockerignoreName)
 		} else {
 			dockerignoreContent, err = afero.ReadFile(fs, dockerignoreFilePath)
 			if err != nil {
 				return err
 			}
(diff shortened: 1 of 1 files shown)


ifbyol (commented):
If both files are defined, we are merging both, which I think it is misleading. If the new file is defined, it should mean that users are migrated to new file, so I don't think we should keep taking into account `.oktetodeployignore`. 

WDYT @pchico83 @codyjlandstrom ?

pchico83:
> If both files are defined, we are merging both, which I think it is misleading. If the new file is defined, it should mean that users are migrated to new file, so I don't think we should keep taking into account `.oktetodeployignore`.
> 
> WDYT @pchico83 @codyjlandstrom ?

Both options sound good to me. But for simplicity, I would also go with one file or the other, without merging tham

andreafalzetti (approved):
Code changes LGTM. I've also tested the scenario described and it works both on Mac and Windows.

As already mentioned the warning for `.oktetodeployignore` appears when I don't have it, and it should be fixed before merging.

About the "merge" vs "override" I think both options are good, I defer to Product the final call, however I would go for now with the current implementation as it's good, and we can just drop the support for the old format in the future.

pchico83:
> I think what we have is ok (merging) until we fully deprecate the old file. User could be using `.oktetoignore` only for `okteto test` and not for deploy in which case no rules would be included.

@maroshii SGTM

ifbyol (approved):

PR #4317, opened 2024-05-24
okteto test: return info log instead of failing when no tests defined

andreafalzetti:
# Proposed changes

As discussed offline, we want to return an info log instead of an error in case `okteto test` is executed with an Okteto Manifest that have no tests defined within it.

**Before**
Screenshot 2024-05-24 at 14 26 47

**After**
Screenshot 2024-05-24 at 14 27 42

## CLI Quality Reminders 🔧

### DIFF
--- a/cmd/test/cmd.go
+++ b/cmd/test/cmd.go
@@ -15,6 +15,7 @@ package test
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"os"
 	"os/signal"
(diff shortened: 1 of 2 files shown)


pchico83:
Could we add a link to https://www.okteto.com/docs/core/okteto-manifest/#test?
The section isn't available yet, but it will be available for the next release

ifbyol (approved):
Approved, but be aware of the documentation link to include. I think we can already include i

andreafalzetti:
Link added, amazing. Thanks!

PR #4320, opened 2024-05-27
[okteto test] only deploy when using `--deploy` flag

andreafalzetti:
# Proposed changes

Fixes DEV-358 & DEV-359

This PR changes `okteto test` to only deploy when requested by the user using the `--deploy` flag. Additionally, it makes sure that if the `deploy` section is missing, the CLI does not panic.

With the new change, when using `--deploy`, it will deploy or redeploy.

## How to validate

1. Using https://github.com/okteto/go-getting-started
2. Create a new namespace or make sure the sample is not deployed to the current ns
3. Add the following section to the `okteto.yml`
```yaml
test:
  unit:
    context: .
    commands:
      - name: "Run unit tests"
        command: echo 1
```
4. Run `okteto test` and observe that the application doesn't get deployed by default
5. Now run `okteto test --deploy` and observe the app being deployed

## CLI Quality Reminders 🔧

### DIFF
--- a/cmd/test/cmd.go
+++ b/cmd/test/cmd.go
@@ -31,7 +31,6 @@ import (
 	"github.com/okteto/okteto/cmd/utils"
 	"github.com/okteto/okteto/pkg/analytics"
 	buildCMD "github.com/okteto/okteto/pkg/cmd/build"
-	"github.com/okteto/okteto/pkg/cmd/pipeline"
 	"github.com/okteto/okteto/pkg/config"
 	"github.com/okteto/okteto/pkg/constants"
 	"github.com/okteto/okteto/pkg/dag"
@@ -187,11 +186,6 @@ func doRun(ctx context.Context, servicesToTest []string, options *Options, ioCtr
 		tracker.TrackImageBuild,
 	})
 
-	kubeClient, _, err := k8sClientProvider.Provide(okteto.GetContext().Cfg)
-	if err != nil {
-		return analytics.TestMetadata{}, fmt.Errorf("could not instantiate kubernetes client: %w", err)
-	}
-
 	cwd, err := os.Getwd()
 	if err != nil {
 		return analytics.TestMetadata{}, fmt.Errorf("failed to get the current working directory to resolve name: %w", err)
@@ -240,25 +234,6 @@ func doRun(ctx context.Context, servicesToTest []string, options *Options, ioCtr
 			return analytics.TestMetadata{}, err
 		}
 	}
-
-	namer := deployCMD.Namer{
-		KubeClient:   kubeClient,
-		Workdir:      cwd,
-		ManifestPath: options.ManifestPathFlag,
-		ManifestName: options.Name,
-	}
-
-	devenvName := options.Name
-	if devenvName == "" {
-		devenvName = namer.ResolveName(ctx)
-	}
-
-	namespace := manifest.Namespace
-	if namespace == "" {
-		namespace = okteto.GetContext().Namespace
-	}
-
-	wasDeployed := pipeline.IsDeployed(ctx, devenvName, namespace, kubeClient)
 
 	if options.Deploy {
 		c := deployCMD.Command{
@@ -312,7 +287,7 @@ func doRun(ctx context.Context, servicesToTest []string, options *Options, ioCtr
 
 	metadata := analytics.TestMetadata{
 		StagesCount: len(testServices),
-		WasDeployed: wasDeployed,
+		Deployed:    options.Deploy,
 		WasBuilt:    wasBuilt,
 	}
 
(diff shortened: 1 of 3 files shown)


cmd/test/cmd.go
```diff
@@ -258,9 +258,9 @@ func doRun(ctx context.Context, servicesToTest []string, options *Options, ioCtr
 		namespace = okteto.GetContext().Namespace
 	}
 
-	needsDeploy := options.Deploy || !pipeline.IsDeployed(ctx, devenvName, namespace, kubeClient)
+	wasDeployed := pipeline.IsDeployed(ctx, devenvName, namespace, kubeClient)
 
-	if needsDeploy {
+	if options.Deploy {
```
pchico83, line 263: when deploying, could we also set `build: true`?

cmd/test/cmd.go
```diff
@@ -308,12 +308,11 @@ func doRun(ctx context.Context, servicesToTest []string, options *Options, ioCtr
 		if err := manifest.ExpandEnvVars(); err != nil {
 			return analytics.TestMetadata{}, fmt.Errorf("failed to expand manifest environment variables: %w", err)
 		}
-		oktetoLog.Information("'%s' was already deployed. To redeploy run 'okteto deploy'", devenvName)
 	}
 
 	metadata := analytics.TestMetadata{
 		StagesCount: len(testServices),
-		WasDeployed: needsDeploy,
+		WasDeployed: wasDeployed,
```
pchico83, line 315: I would track if the `--deploy` flag is used and remove the computation of `wasDeployed`

cmd/test/cmd.go
```diff
@@ -308,12 +308,11 @@ func doRun(ctx context.Context, servicesToTest []string, options *Options, ioCtr
 		if err := manifest.ExpandEnvVars(); err != nil {
 			return analytics.TestMetadata{}, fmt.Errorf("failed to expand manifest environment variables: %w", err)
 		}
-		oktetoLog.Information("'%s' was already deployed. To redeploy run 'okteto deploy'", devenvName)
```
andreafalzetti, line 311: I've removed this log now because when not passing the `--deploy` flag, we are not deploying, but only expanding vars. I find confusing this log. Maybe we can add another one for when a user uses `--deploy` on an application that was already deployed and we inform that it was already deployed. Not sure tho. Thoughts @pchico83 @ifbyol ?

pchico83: I think it's fine to remove it

cmd/test/cmd.go
andreafalzetti: @pchico83 any objection in renaming it to `Deployed`, then? so that we can keep the name consisent and leave `WasDeployed` available in the future, in case we want to track "re-deploys".

pchico83: Yes, `deployed` sounds good

andreafalzetti: Changed in 34441f69

andreafalzetti: Changed in 27cc3316

cmd/test/cmd.go
```diff
@@ -288,18 +264,18 @@ func doRun(ctx context.Context, servicesToTest []string, options *Options, ioCtr
...
+			RunInRemote:      runInRemote,
 			Wait:             true,
 			ShowCTA:          false,
 		})
-		c.TrackDeploy(manifest, manifest.Deploy.Remote, deployStartTime, err)
+		c.TrackDeploy(manifest, runInRemote, deployStartTime, err)
 
 		if err != nil {
-			oktetoLog.Error("deploy failed: %s", err.Error())
+			oktetoLog.Errorf("deploy failed: %s", err.Error())
```
andreafalzetti, line 278: this was displaying broken:

Screenshot 2024-05-27 at 19 34 03

it's also redundant, so we may consider remove it. Thoughts?

cmd/test/cmd.go
```diff
@@ -241,26 +235,14 @@ func doRun(ctx context.Context, servicesToTest []string, options *Options, ioCtr
...
-	if devenvName == "" {
-		devenvName = namer.ResolveName(ctx)
-	}
+	shouldDeploy := options.Deploy
 
-	namespace := manifest.Namespace
-	if namespace == "" {
-		namespace = okteto.GetContext().Namespace
+	if shouldDeploy && manifest.Deploy == nil {
+		oktetoLog.Warning("Nothing to deploy. The 'deploy' section of your Okteto Manifest is empty. For more information, check the docs: https://okteto.com/docs/core/okteto-manifest/#deploy")
```
andreafalzetti, line 241: @pchico83 I am adding a warning if the user set `--deploy` but then has an empty deploy section. Let me know if you would change anything.

andreafalzetti: This also preents the panic that we have now, while accessing `manifest.Deploy.Remote`.

ifbyol (approved):

cmd/test/cmd.go
ifbyol: `Nothing to deploy` sounds a bit odd but don't know why. I don't have a better alternative. I think we can merge it as it is and change it later if needed

jLopezbarb (approved):

PR #4321, opened 2024-05-27
remove unused code

andreafalzetti:
# Proposed changes

Removing unreachable code and unused vars.

## CLI Quality Reminders 🔧


pkg/filesystem/workdir.go
```diff
@@ -1,34 +0,0 @@
...
-// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-// See the License for the specific language governing permissions and
-// limitations under the License.
-
-package filesystem
-
-import "path/filepath"
-
-// GetWorkdirFromManifestPath sets the path
-func GetWorkdirFromManifestPath(manifestPath string) string {
```
ifbyol, line 19: I'm wondering if we should be using this one instead of the one within `model` package. It seems the implementation is the same and it would help us to reduce the footprint of model package along the codebase

pkg/filesystem/workdir.go
```diff
@@ -1,34 +0,0 @@
...
-	dir := filepath.Dir(manifestPath)
-	if filepath.Base(dir) == ".okteto" {
-		dir = filepath.Dir(dir)
-	}
-	return dir
-}
-
-// GetManifestPathFromWorkdir returns the path from a workdir
-func GetManifestPathFromWorkdir(manifestPath, workdir string) string {
```
ifbyol, line 28: The same about this function. It is also located in `model` package and they seem to be the same implementation

ifbyol (commented):
In general looks good but there is a couple of functions that we might want to delete them from `model` package in order to reduce the footprint of that package in the codebase

pkg/filesystem/workdir.go
andreafalzetti: good idea, updated ✅

pkg/filesystem/workdir.go
andreafalzetti: good idea, updated ✅

ifbyol (approved):

PR #4324, opened 2024-05-28
Add OKTETO_DOMAIN to remote test runner

maroshii:
# Proposed changes

The remote test runner doesn't inject the known okteto values. This was carryover from when we decided not to share OKTETO_ENV from deploy section.

Still TBD whether we want to add all the other envvars from the remote deploy command: https://github.com/okteto/okteto/blob/ad8c37c3002a2b123134af95fd865b94b1101c0d/pkg/deployable/deploy.go#L244-L267

# Validation

```yaml
test:
  env:
    commands:
      - printenv | grep OKTETO_DOMAIN
```

Before this change, this was returning and error. After this change this works and the full env is printed to stdout

### DIFF
--- a/cmd/remoterun/test.go
+++ b/cmd/remoterun/test.go
@@ -25,6 +25,7 @@ import (
 	"github.com/okteto/okteto/pkg/deployable"
 	"github.com/okteto/okteto/pkg/k8s/kubeconfig"
 	oktetoLog "github.com/okteto/okteto/pkg/log"
+	"github.com/okteto/okteto/pkg/model"
 	"github.com/okteto/okteto/pkg/okteto"
 	"github.com/spf13/afero"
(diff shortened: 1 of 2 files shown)


ifbyol:
> # Proposed changes
> The remote test runner doesn't inject the known okteto values. This was carryover from when we decided not to share OKTETO_ENV from deploy section.
> 
> Still TBD whether we want to add all the other envvars from the remote deploy command:
> 
> https://github.com/okteto/okteto/blob/ad8c37c3002a2b123134af95fd865b94b1101c0d/pkg/deployable/deploy.go#L244-L267
> 

Some of them don't make sense anymore. They were needed when we had inception in the command (deploy was calling another deploy), but probably now they don't make sense

PR #4326, opened 2024-06-03
fix: use the correct image on okteteo test

jLopezbarb:

# Proposed changes

Fixes DEV-379

In this change we have fixed the way of calculating the images that eneds to be built for the okteto test command. We were checking if the service was already built once per service instead of doing it for all the services at once, expanding the manifest several times, which lead to inconsistency in the manifest.test.image field. This change aims to calculate if the images are being built once, setting the correct env vars for running the tests.

## How to validate

1. Use https://github.com/okteto/getting-started/blob/1a12b279b3dd7aabbc69f01d5a4388b0db564efb/okteto.yml
1. Run okteto test -l info
1. Check on the log message that the image being used for the test is the correct one

## CLI Quality Reminders 🔧


cmd/build/v2/services.go
```diff
@@ -24,6 +25,14 @@ import (
 	"golang.org/x/sync/errgroup"
 )
 
+var (
+	// ErrimageIsNotAOktetoBuildSyntax is returned when the image is not an okteto build syntax
+	ErrimageIsNotAOktetoBuildSyntax = errors.New("image is not an okteto build syntax")
```
andreafalzetti, line 30:
```suggestion
	ErrImageIsNotAOktetoBuildSyntax = errors.New("image is not an okteto build syntax")
```

cmd/build/v2/services.go
```diff
@@ -24,6 +25,14 @@ import (
 	"golang.org/x/sync/errgroup"
 )
 
+var (
+	// ErrimageIsNotAOktetoBuildSyntax is returned when the image is not an okteto build syntax
```
andreafalzetti, line 29:
```suggestion
	// ErrImageIsNotAOktetoBuildSyntax is returned when the image is not an okteto build syntax
```

cmd/test/cmd.go
```diff
@@ -210,30 +215,9 @@ func doRun(ctx context.Context, servicesToTest []string, options *Options, ioCtr
...
-		if err := builder.Build(ctx, &types.BuildOptions{
-			EnableStages: true,
-			Manifest:     manifest,
-			CommandArgs:  svc,
-		}); err != nil {
-			return analytics.TestMetadata{}, err
-		}
+	wasBuilt, err := doBuild(ctx, manifest, testServices, builder, ioCtrl)
+	if err != nil {
+		return analytics.TestMetadata{}, fmt.Errorf("failed to build services: %w", err)
```
andreafalzetti, line 220: I think users will have an hard time if they get this error, to understand this error message. I would say 'image' rather than 'services'. WDYT?

Currently hitting this error results in:

```
 i  Using andreafalzetti @ <context> as context
 x  Failed to build services: the image '${OKTETO_BUILD_NO_IMAGE}' is not in the build section of the manifest
```

I would change it to something like:
Screenshot 2024-06-03 at 22 55 17

```suggestion
		return analytics.TestMetadata{}, fmt.Errorf("okteto test needs to build the images defined but failed: %w", err)
```

which results in:

Screenshot 2024-06-03 at 23 06 54

cmd/test/cmd.go
```diff
@@ -379,3 +363,41 @@ func shouldRunInRemote(manifestDeploy *model.DeployInfo) bool {
...
+func doBuild(ctx context.Context, manifest *model.Manifest, svcs []string, builder builder, ioCtrl *io.Controller) (bool, error) {
+	// make sure the images used for the tests exist. If they don't build them
+	svcsToBuild := []string{}
+	for _, name := range svcs {
+		imgName := manifest.Test[name].Image
+		getImg := func(manifest *model.Manifest) string { return imgName }
+		svc, err := builder.GetSvcToBuildFromRegex(manifest, getImg)
+		if err != nil {
+			if errors.Is(err, buildv2.ErrOktetBuildSyntaxImageIsNotInBuildSection) {
+				return false, fmt.Errorf("the image '%s' is not in the build section of the manifest", imgName)
```
andreafalzetti, line 376: I would suggest to make this error clearer, together with the suggestion above, this error can become very clear and point to the docs.

Example:

Screenshot 2024-06-03 at 23 14 15


```suggestion
				return false, fmt.Errorf("test '%s' needs image '%s' but it's not defined in the build section of the Okteto Manifest. See: https://www.okteto.com/docs/core/okteto-variables/#built-in-environment-variables-for-images-in-okteto-registry", name, imgName)
```

cmd/build/v2/services.go
```diff
@@ -24,6 +25,14 @@ import (
...
+	ErrimageIsNotAOktetoBuildSyntax = errors.New("image is not an okteto build syntax")
+
+	// ErrOktetBuildSyntaxImageIsNotInBuildSection is returned when the image is not in the build section
+	ErrOktetBuildSyntaxImageIsNotInBuildSection = errors.New("image is not in the build section")
```
ifbyol, line 33: We remark the build section, but what would happen if the image is defined in a compose file? I mean, imagine that my dev environment is deployed with a docker-compose but I have a `test` section in the manifest referring to one of the images built through the compose? Would in this case the image being built too? Or will it just happen only if they are defined in the build section?

cmd/build/v2/services.go
```diff
@@ -122,13 +131,13 @@ func (bc *OktetoBuilder) checkServiceToBuildDuringDeploy(service string, manifes
...
 
-func (bc *OktetoBuilder) GetServicesToBuildForImage(ctx context.Context, manifest *model.Manifest, imgFinder model.ImageFromManifest) ([]string, error) {
+func (bc *OktetoBuilder) GetSvcToBuildFromRegex(manifest *model.Manifest, imgFinder model.ImageFromManifest) (string, error) {
 	img := imgFinder(manifest)
 	reg := regexp.MustCompile(`OKTETO_BUILD_(\w+)_`)
 	matches := reg.FindStringSubmatch(img)
 	foundMatches := 2
 	if len(matches) == 0 {
-		return nil, nil
+		return "", ErrimageIsNotAOktetoBuildSyntax
```
ifbyol, line 140: Should this really fail? What would happen if they specify an env var they have defined in their env? I mean, if this is another kind of env var, e.g. `API_IMAGE`

cmd/destroy/build_control.go
```diff
@@ -42,27 +42,38 @@ func newBuildCtrl(name string, analyticsTracker, insights buildTrackerInterface,
...
+	}
+
+	if len(svcsToBuild) == 0 {
+		oktetoLog.Debug("destroy.image is already built")
 		return nil
 	}
 	buildOptions := &types.BuildOptions{
 		EnableStages: true,
 		Manifest:     manifest,
 		CommandArgs:  svcsToBuild,
```
ifbyol, line 74: Should then specify here `svcToBuild`? Or should we just update the info logger below to use `svcsToBuild` instead of `svcToBuild`?

pkg/build/manifest_section.go
```diff
@@ -73,7 +73,7 @@ func getListDiff(l1, l2 []string) []string {
 	}
 	added := []string{}
 	for _, svcName := range longerList {
-		if _, ok := shorterListSet[svcName]; ok {
+		if _, ok := shorterListSet[svcName]; !ok {
```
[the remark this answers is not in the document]
jLopezbarb: We were getting the common elements from the list. I noticed on the logs that I was not getting the correct one on the logs and then after checking the tests I could verify it. Given that the function name is `getListDiff` I can say it's correct

cmd/test/cmd.go
```diff
@@ -377,3 +361,41 @@ func shouldRunInRemote(manifestDeploy *model.DeployInfo) bool {
...
+	svcsToBuild := []string{}
+	for _, name := range svcs {
+		imgName := manifest.Test[name].Image
+		getImg := func(manifest *model.Manifest) string { return imgName }
+		svc, err := builder.GetSvcToBuildFromRegex(ctx, manifest, getImg)
+		if err != nil {
+			if errors.Is(err, buildv2.ErrOktetBuildSyntaxImageIsNotInBuildSection) {
+				return false, fmt.Errorf("the image '%s' is not in the build section of the manifest", imgName)
+			}
+			ioCtrl.Logger().Debugf("error getting services to build for image '%s': %s", imgName, err)
```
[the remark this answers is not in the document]
jLopezbarb: Yes, all the `test.image` goes through here. 
If we return an error here we'll be erroring for images like `golang` that are valid but are not in our format. 
If we continue without adding it to the `svcsToBuild` we could be missing build sections like the following one:
```yaml
build:
  test-image:
    context: .
    image: okteto/ci:latest
```

cmd/build/v2/services.go
jLopezbarb: When we load the compose we translate compose build sections into okteto build section, so if the test is referring to a build defined in a compose manifest it will be built.

cmd/build/v2/services.go
jLopezbarb: It's failing the function because we couldn't get the image from the regex but after it in the caller function we are capturing the error and managing it

pkg/build/manifest_section.go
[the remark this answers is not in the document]
ifbyol: Well, this is then changing somewhere the behavior. Could we break some scenario?

Also, could we update the unit test to reflect this behavior? I see there is a test but it wasn't needed to be changed, probably there are some use cases missed

cmd/build/v2/services.go
ifbyol: But we are not handling this error specifically in the caller. We only handle `ErrOktetBuildSyntaxImageIsNotInBuildSection`. Is that expected?

jLopezbarb: We don't handle because if it's not the other error we should continue with the execution to check if the image is already built. I'll add a comment to make it clearer

PR #4329, opened 2024-06-04
fix: only use requirements from yaml

jLopezbarb:
# Proposed changes

Fixes DEV-224

This PR aims to use only the resources requests and limits that the user has set in the yaml for the dev container and remove the ones that the original container could have. 

I also had to modify the names because while reading it several times was hard for me to follow what was the meaning of the them.

## How to validate
Via compose:
1. Copy the following compose:
```yaml
version: '3.8'

services:
  nginx:
    image: nginx
    ports:
      - "1222:80"
    deploy:
      resources:
        limits:
          cpus: '0.5'
          memory: 512M
        reservations:
          cpus: '0.2'
          memory: 256M
    volumes:
    - .:/usr/share/nginx/html
```
2. Run `okteto deploy` and check that the resources have been set correctly
3. Run `okteto up` and check that there is no resources set on the dev container

Via deployment/sfs 
1. Create a deployment with resources:
```yaml
# nginx-deployment.yaml

apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx-deployment
spec:
  replicas: 1
  selector:
    matchLabels:
      app: nginx
  template:
    metadata:
      labels:
        app: nginx
    spec:
      containers:
      - name: nginx
        image: nginx
        resources:
          requests:
            memory: "64Mi"
            cpu: "250m"
          limits:
            memory: "128Mi"
            cpu: "500m"
```
2. Create the following `okteto.yml`:
```yaml
deploy:
- kubectl apply -f k8s.yml

dev:
  nginx-deployment:
    image: nginx
    command: ["nginx", "-g", "daemon off;"]
    sync:
    - .:/usr/share/nginx/html
    forward:
    - 1234:80
```
3. Run `okteto deploy` and check that the resources have been set correctly
4. Run `okteto up` and check that there is no resources set on the dev container

## CLI Quality Reminders 🔧

Edit: formatting

### DIFF
--- a/pkg/k8s/apps/translate.go
+++ b/pkg/k8s/apps/translate.go
@@ -249,31 +249,37 @@ func TranslateLifecycle(c *apiv1.Container, l *model.Lifecycle) {
 }
 
 // TranslateResources translates the resources attached to a container
-func TranslateResources(c *apiv1.Container, rr model.ResourceRequirements) {
+func TranslateResources(container *apiv1.Container, ruleResource model.ResourceRequirements) {
+	ruleResourceList := []model.ResourceList{ruleResource.Requests, ruleResource.Limits}
+	containerResourceList := []*apiv1.ResourceList{&container.Resources.Requests, &container.Resources.Limits}
+	resourceTypes := []apiv1.ResourceName{apiv1.ResourceMemory, apiv1.ResourceCPU}
 
-	rrl := []model.ResourceList{rr.Requests, rr.Limits}
-	cr := []*apiv1.ResourceList{&c.Resources.Requests, &c.Resources.Limits}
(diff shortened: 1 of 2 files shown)


ifbyol:
> I've tested with older CLI and this one, and I was able to verify the correct behaviour in both cases (compose/k8s). The only thing which I want to highlight is that while my checks have been done with `kubectl` and it was all good, in the Okteto UI the 'YAML' tab showing the service that is 'In Development' is not updated, so I still see the resources of the deployment.
> 
> Approving becuase I think that is out of scope, but let's discuss if any follow-up ticket is needed. cc @ifbyol

That is more or less expected, as we display the original deployment. Don't think it is a problem

ifbyol (approved):

PR #4331, opened 2024-06-07
Reserved variable names

teresaromero:
Fixes https://okteto.atlassian.net/browse/DEV-408

We are restricting certain variable names when the user uses `--var` flag input.
Commands affected are:
- okteto deploy
- okteto pipeline deploy
- okteto preview deploy
- okteto test
- okteto deploy --dependencies (manifest might have variables for the dependencies)

The PR adds a check when the cmd is run in order to validate the name. As the validation is shared across commands a new package validator has been created with this function.

Tests:
- Running CLI commands affected with variables
- Used CLI bult-in at installer and check compatibility

Screenshots from testing

Screenshot 2024-06-10 at 08 54 44

Screenshot 2024-06-10 at 08 54 56

Screenshot 2024-06-10 at 09 00 00

Using custom okteto yaml with dependencies with variable https://github.com/teresaromero/go-getting-started/blob/tere/installer-dependencies/okteto.yml 

Screenshot 2024-06-10 at 08 56 53

### DIFF
--- a/cmd/deploy/deploy.go
+++ b/cmd/deploy/deploy.go
@@ -643,7 +643,7 @@ func (dc *Command) deployDependencies(ctx context.Context, deployOptions *Option
 		oktetoLog.Information("Deploying dependency  '%s'", depName)
 		oktetoLog.SetStage(fmt.Sprintf("Deploying dependency %s", depName))
 
-		if err := validator.CheckForbiddenEnvName(dep.Variables); err != nil {
+		if err := validator.CheckReservedEnvName(dep.Variables); err != nil {
 			return err
 		}
 
--- a/cmd/deploy/deploy.go
+++ b/cmd/deploy/deploy.go
@@ -643,7 +643,7 @@ func (dc *Command) deployDependencies(ctx context.Context, deployOptions *Option
 		oktetoLog.Information("Deploying dependency  '%s'", depName)
 		oktetoLog.SetStage(fmt.Sprintf("Deploying dependency %s", depName))
 
-		if err := validator.CheckReservedEnvName(dep.Variables); err != nil {
+		if err := validator.CheckReservedVarName(dep.Variables); err != nil {
 			return err
 		}
 
--- a/cmd/deploy/deploy.go
+++ b/cmd/deploy/deploy.go
@@ -48,6 +48,7 @@ import (
 	oktetoPath "github.com/okteto/okteto/pkg/path"
 	"github.com/okteto/okteto/pkg/repository"
 	"github.com/okteto/okteto/pkg/types"
+	"github.com/okteto/okteto/pkg/validator"
 	"github.com/spf13/afero"
 	"github.com/spf13/cobra"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -641,6 +642,11 @@ func (dc *Command) deployDependencies(ctx context.Context, deployOptions *Option
 	for depName, dep := range deployOptions.Manifest.Dependencies {
 		oktetoLog.Information("Deploying dependency  '%s'", depName)
 		oktetoLog.SetStage(fmt.Sprintf("Deploying dependency %s", depName))
+
+		if err := validator.CheckForbiddenEnvName(dep.Variables); err != nil {
+			return err
+		}
+
 		dep.Variables = append(dep.Variables, env.Var{
 			Name:  "OKTETO_ORIGIN",
 			Value: "okteto-deploy",
--- a/cmd/deploy/validate.go
+++ b/cmd/deploy/validate.go
@@ -21,7 +21,7 @@ import (
 // validateAndSet returns error when variables dont have expected format NAME=VALUE or NAME is not allowed
 // when variable is valid, it sets its value as env variable
 func validateAndSet(variables []string, setEnv func(key, value string) error) error {
-	if err := validator.CheckForbiddenVariablesNameOption(variables); err != nil {
+	if err := validator.CheckReservedVariablesNameOption(variables); err != nil {
 		return err
 	}
 
--- a/cmd/deploy/validate.go
+++ b/cmd/deploy/validate.go
@@ -15,9 +15,16 @@ package deploy
 
 import (
 	"github.com/okteto/okteto/pkg/env"
+	"github.com/okteto/okteto/pkg/validator"
 )
 
+// validateAndSet returns error when variables dont have expected format NAME=VALUE or NAME is not allowed
+// when variable is valid, it sets its value as env variable
 func validateAndSet(variables []string, setEnv func(key, value string) error) error {
+	if err := validator.CheckVariablesFlag(variables); err != nil {
+		return err
+	}
+
 	envVars, err := env.Parse(variables)
 	if err != nil {
 		return err
--- a/cmd/deploy/validate.go
+++ b/cmd/deploy/validate.go
@@ -21,7 +21,7 @@ import (
 // validateAndSet returns error when variables dont have expected format NAME=VALUE or NAME is not allowed
 // when variable is valid, it sets its value as env variable
 func validateAndSet(variables []string, setEnv func(key, value string) error) error {
-	if err := validator.CheckVariablesFlag(variables); err != nil {
+	if err := validator.CheckForbiddenVariablesNameOption(variables); err != nil {
 		return err
 	}
 
--- a/pkg/validator/validator.go
+++ b/pkg/validator/validator.go
@@ -19,17 +19,21 @@ import (
 	"strings"
 
 	"github.com/okteto/okteto/pkg/env"
+	oktetoErrors "github.com/okteto/okteto/pkg/errors"
 )
 
 // ErrReservedVariableName is raised when a variable from cmd option has invalid name
-var ErrReservedVariableName = errors.New("reserved variable name. See documentation for more info: https://www.okteto.com/docs/core/credentials/environment-variables/")
+var ErrReservedVariableName = errors.New("reserved variable name")
 
 // CheckReservedVariablesNameOption returns an error when any of the variable names from command flags are not allowed as input
 func CheckReservedVariablesNameOption(variables []string) error {
 	for _, v := range variables {
 		name, _, ok := strings.Cut(v, "=")
 		if ok && isReservedVariableName(name) {
-			return fmt.Errorf("%s is %w", name, ErrReservedVariableName)
+			return oktetoErrors.UserError{
+				E:    fmt.Errorf("%s is %w.", name, ErrReservedVariableName),
+				Hint: "See documentation for more info: https://www.okteto.com/docs/core/credentials/environment-variables/",
+			}
 		}
 	}
 	return nil
--- a/pkg/validator/validator.go
+++ b/pkg/validator/validator.go
@@ -21,8 +21,8 @@ import (
 // errForbiddenVariableName is raised when a variable from cmd option has invalid name
 var errForbiddenVariableName = errors.New("variable name not allowed: <placeholder>")
 
-// CheckVariablesFlag returns an error when any of the variable names from command flags are not allowed as input
-func CheckVariablesFlag(variables []string) error {
+// CheckForbiddenVariablesNameOption returns an error when any of the variable names from command flags are not allowed as input
+func CheckForbiddenVariablesNameOption(variables []string) error {
 	for _, v := range variables {
 		name, _, ok := strings.Cut(v, "=")
 		if ok && isForbiddenVariableName(name) {
--- a/pkg/validator/validator.go
+++ b/pkg/validator/validator.go
@@ -18,15 +18,15 @@ import (
 	"strings"
 )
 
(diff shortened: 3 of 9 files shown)


pkg/validator/validator.go
```diff
@@ -0,0 +1,57 @@
...
+
+import (
+	"errors"
+	"strings"
+
+	"github.com/okteto/okteto/pkg/env"
+)
+
+// ErrForbiddenVariableName is raised when a variable from cmd option has invalid name
+var ErrForbiddenVariableName = errors.New(`Error: Forbidden variable names
```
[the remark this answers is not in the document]
andreafalzetti: also my preference is for adding the actual vars used like we do for the warnings (`OKTETO_CONTEXT` and `OKTETO_NAMESPACE`)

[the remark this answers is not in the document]
andreafalzetti: nit 2. I think a better word wold be "Rervesed" instead of "Forbidden", but happy with Forbidden if product is happy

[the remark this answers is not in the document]
teresaromero: I guess you mean "Reserved" ? ;)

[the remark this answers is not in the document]
teresaromero: I see, i did not found this page of doc, and i thought we were talking about the built-in variables. 
Then i am looking into the wrong concept of variables.... i will also take this into account for the app PR.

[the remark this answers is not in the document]
codyjlandstrom: That docs link is great.
Even though it feels a bit strong, I think using the "Forbidden" term is better here just so the user is very clear on the error.

cmd/deploy/deploy.go
```diff
@@ -641,6 +642,11 @@ func (dc *Command) deployDependencies(ctx context.Context, deployOptions *Option
 	for depName, dep := range deployOptions.Manifest.Dependencies {
 		oktetoLog.Information("Deploying dependency  '%s'", depName)
 		oktetoLog.SetStage(fmt.Sprintf("Deploying dependency %s", depName))
+
+		if err := validator.CheckReservedEnvName(dep.Variables); err != nil {
```
andreafalzetti, line 646:
```suggestion
		if err := validator.CheckReservedVarName(dep.Variables); err != nil {
```

andreafalzetti: nit. I find misleading `EnvName`,, maybe `VarName` is more explicit?

pkg/validator/validator.go
```diff
@@ -0,0 +1,57 @@
...
+
+// ErrReservedVariableName is raised when a variable from cmd option has invalid name
+var ErrReservedVariableName = errors.New("Reserved variable name. See documentation for more info: https://www.okteto.com/docs/core/credentials/environment-variables/")
+
+// CheckReservedVariablesNameOption returns an error when any of the variable names from command flags are not allowed as input
+func CheckReservedVariablesNameOption(variables []string) error {
+	for _, v := range variables {
+		name, _, ok := strings.Cut(v, "=")
+		if ok && isReservedVariableName(name) {
+			return ErrReservedVariableName
```
andreafalzetti, line 31: In different places of the CLI we always name the var causing the warning, for example when authenticating via env vars, we mention them. I imagine a complex command, where the user might not notice which var is reserved, it may be tricky or require further checks (ie. open the docs) to understand which one is reserved. Why not making it simpler by printing like:

```
 x  OKTETO_CONTEXT is a reserved variable name. See documentation for more info: https://www.okteto.com/docs/core/credentials/environment-variables/
 ```


```suggestion
			return fmt.Errorf("%s is a %w", name, ErrReservedVariableName)
```

pkg/validator/validator.go
```diff
@@ -0,0 +1,57 @@
...
+
+import (
+	"errors"
+	"strings"
+
+	"github.com/okteto/okteto/pkg/env"
+)
+
+// ErrReservedVariableName is raised when a variable from cmd option has invalid name
+var ErrReservedVariableName = errors.New("Reserved variable name. See documentation for more info: https://www.okteto.com/docs/core/credentials/environment-variables/")
```
andreafalzetti, line 24:
```suggestion
var ErrReservedVariableName = errors.New("reserved variable name. See documentation for more info: https://www.okteto.com/docs/core/credentials/environment-variables/")
```

(required if we add the var name)

pkg/validator/validator.go
```diff
@@ -0,0 +1,57 @@
...
+		}
+	}
+	return nil
+}
+
+// CheckReservedEnvName returns an error when any of the variable names from dependency manifest
+func CheckReservedEnvName(variables []env.Var) error {
+	for _, v := range variables {
+		if isReservedVariableName(v.Name) {
+			return ErrReservedVariableName
```
andreafalzetti, line 41:
```suggestion
			return fmt.Errorf("%s is a %w", v.Name, ErrReservedVariableName)
```

(required if we add the var name)

cmd/deploy/deploy.go
teresaromero: i used this because of the type.... in this case is not a list of strings, but an env.Env 🤷🏻‍♀️

andreafalzetti: @teresaromero the type is `env.Var`

teresaromero: nothing, i was wrong. its already changed

pkg/validator/validator.go
```diff
@@ -0,0 +1,58 @@
...
+
+// ErrReservedVariableName is raised when a variable from cmd option has invalid name
+var ErrReservedVariableName = errors.New("reserved variable name. See documentation for more info: https://www.okteto.com/docs/core/credentials/environment-variables/")
+
+// CheckReservedVariablesNameOption returns an error when any of the variable names from command flags are not allowed as input
+func CheckReservedVariablesNameOption(variables []string) error {
+	for _, v := range variables {
+		name, _, ok := strings.Cut(v, "=")
+		if ok && isReservedVariableName(name) {
+			return fmt.Errorf("%s is %w", name, ErrReservedVariableName)
```
jLopezbarb, line 32: Could we use a `UserError` here? I think it's more readable if we had the `See documentation for more info: https://www.okteto.com/docs/core/credentials/environment-variables/` message in a new line

pkg/validator/validator.go
```diff
@@ -0,0 +1,58 @@
...
+		}
+	}
+	return nil
+}
+
+// CheckReservedVarName returns an error when any of the variable names from dependency manifest
+func CheckReservedVarName(variables []env.Var) error {
```
jLopezbarb, line 39: `CheckReservedVariablesNameOption` and `CheckReservedVarName` is mostly the same. Could we transform `variables []string` in to  `variables []env.Var` in order to have just one function. Otherwise when we change one function we could forget to update the other one.

cmd/deploy/validate.go
```diff
@@ -15,9 +15,16 @@ package deploy
 
 import (
 	"github.com/okteto/okteto/pkg/env"
+	"github.com/okteto/okteto/pkg/validator"
 )
 
+// validateAndSet returns error when variables dont have expected format NAME=VALUE or NAME is not allowed
+// when variable is valid, it sets its value as env variable
 func validateAndSet(variables []string, setEnv func(key, value string) error) error {
+	if err := validator.CheckReservedVariablesNameOption(variables); err != nil {
```
jLopezbarb, line 24: How it should behave if I add a reserved variable and run the command on remote? Maybe I want to deploy on a different namespace, What should be the best way to do it?

teresaromero: validation is intended to be done when the cmd is starting, not getting any deeper in the code. when running the remote, as far as i know is using the deploy command as well with the flag remote?

pkg/validator/validator.go
teresaromero: i did not wanted to make any redundant transformations in order to validate from this point. If you think that would be worth it i can do the transformation for then

andreafalzetti: I agree with Tere it's so simple it's not worth in my opinion to change it

andreafalzetti (approved):
Nice one 👍

pkg/validator/validator.go
teresaromero: I've updated this, it shows like this 

Screenshot 2024-06-12 at 12 15 21

jLopezbarb (approved):

audiolion:
👋  hey we ran into this error message today (using @latest tag on GH actions). Looked at the docs link but I am still unclear on what I should be doing? We are using OKTETO_NAMESPACE today to set the namespace to the PR name, e.g. `okteto preview deploy pr-4732 --var OKTETO_ENV=dev --var OKTETO_NAMESPACE=pr-4732`. Do we just remove the usage of OKTETO_NAMESPACE? What impact does that have?

codyjlandstrom:
Hi @audiolion 👋 
Sorry the error message didn't come through. Mind sharing the error that you're seeing again?
Your usage of `OKTETO_NAMESPACE` looks correct.

audiolion:
> Hi @audiolion 👋 Sorry the error message didn't come through. Mind sharing the error that you're seeing again? Your usage of `OKTETO_NAMESPACE` looks correct.

Hey @codyjlandstrom thanks for the fast response. Here is the entire log:

```
Run okteto/deploy-preview@latest
  with:
    name: pr-47[2](https://github.com/SironaMedical/services-frontend/actions/runs/9844359889/job/27178091344#step:6:2)0
    scope: global
    timeout: 45m
    variables: ENV=local, PUBLIC_DEPLOY_ENV=preview, NODE_ENV=production, OKTETO_ENV=dev, OKTETO_NAMESPACE=pr-4720, DD_API_KEY=***,
  env:
    GITHUB_TOKEN: ***
    SIRONA_GITHUB_PACKAGES_READ_TOKEN: ***
/usr/bin/docker run --name d[3](https://github.com/SironaMedical/services-frontend/actions/runs/9844359889/job/27178091344#step:6:3)6752c11ab7d8f3b[4](https://github.com/SironaMedical/services-frontend/actions/runs/9844359889/job/27178091344#step:6:4)c53b14cbf54b2b4f[5](https://github.com/SironaMedical/services-frontend/actions/runs/9844359889/job/27178091344#step:6:5)59_5090f5 --label 9d3[6](https://github.com/SironaMedical/services-frontend/actions/runs/9844359889/job/27178091344#step:6:6)75 --workdir /github/workspace --rm -e "GITHUB_TOKEN" -e "SIRONA_GITHUB_PACKAGES_READ_TOKEN" -e "INPUT_NAME" -e "INPUT_SCOPE" -e "INPUT_TIMEOUT" -e "INPUT_VARIABLES" -e "INPUT_FILE" -e "INPUT_BRANCH" -e "INPUT_LOG-LEVEL" -e "HOME" -e "GITHUB_JOB" -e "GITHUB_REF" -e "GITHUB_SHA" -e "GITHUB_REPOSITORY" -e "GITHUB_REPOSITORY_OWNER" -e "GITHUB_REPOSITORY_OWNER_ID" -e "GITHUB_RUN_ID" -e "GITHUB_RUN_NUMBER" -e "GITHUB_RETENTION_DAYS" -e "GITHUB_RUN_ATTEMPT" -e "GITHUB_REPOSITORY_ID" -e "GITHUB_ACTOR_ID" -e "GITHUB_ACTOR" -e "GITHUB_TRIGGERING_ACTOR" -e "GITHUB_WORKFLOW" -e "GITHUB_HEAD_REF" -e "GITHUB_BASE_REF" -e "GITHUB_EVENT_NAME" -e "GITHUB_SERVER_URL" -e "GITHUB_API_URL" -e "GITHUB_GRAPHQL_URL" -e "GITHUB_REF_NAME" -e "GITHUB_REF_PROTECTED" -e "GITHUB_REF_TYPE" -e "GITHUB_WORKFLOW_REF" -e "GITHUB_WORKFLOW_SHA" -e "GITHUB_WORKSPACE" -e "GITHUB_ACTION" -e "GITHUB_EVENT_PATH" -e "GITHUB_ACTION_REPOSITORY" -e "GITHUB_ACT
running: okteto preview deploy pr-4[7](https://github.com/SironaMedical/services-frontend/actions/runs/9844359889/job/27178091344#step:6:7)20 --scope global --branch=feat/rp-2210-field-pane-set-any-default-picklist --repository=https://github.com/SironaMedical/services-frontend --sourceUrl=https://github.com/SironaMedical/services-frontend/pull/4720 --timeout=45m --var ENV=local --var PUBLIC_DEPLOY_ENV=preview --var NODE_ENV=production --var OKTETO_ENV=dev --var OKTETO_NAMESPACE=pr-4720 --var DD_API_KEY=*** --wait
 x  OKTETO_NAMESPACE is reserved variable name.
    See documentation for more info: https://www.okteto.com/docs/core/credentials/environment-variables/
```

audiolion:
Update based on my discussion with Cody Landstrom @ Okteto:

I removed the usage of the reserved variable and the live preview is working. I don't think we should have been using namespaces for live previews anyways because they are intended for development environments to provide isolation / collaboration.

codyjlandstrom:
Thanks @audiolion!
Don't hesitate to reach out if anything else comes up

PR #4334, opened 2024-06-11
Generate artifacts in `okteto test` including when tests fail

andreafalzetti:
# Proposed changes

Fixes: DEV-382

This change allows exporting artifacts from `okteto test` including when tests fail. Previously, because the docker build failed, the artifacts were not exported. With this change, the build will never fail due to failing tests, so artifacts can always be retrieved.

## Other approaches considered

The initial attempt at making this change was moving the `remote-run` inside a `sh` script with `set +e` but that didn't work, especially the `set +e`, despite having a reproduced the same scenario outside of the remote-run in a vanilla Dockerfile it would work. Another way to swallow the failure was wrapping the cmd inside an `if` - that worked well and we could also differenciate between success / failure, however we don't need to handle such signals because the current implementation already makes the test container failure message appear accordingly, because the logs of the remote-run are still parsed (the cmd exit code is swallowed but everything else stays as is). So I've decided to simplify and go with the `|| true` approach. In [this commit](https://github.com/okteto/okteto/pull/4334/commits/fa18546ad6e14a460f884be6c1759d0e0af1ff2d) you can see the other implementation, keeping it for futurew references.

## How to validate

Important: Make sure to use `--no-cache` and check the exit code of the Okteto CLI using `echo $?` after each test. You can use `make clean` between tests to remove the generated artifacts.

1. Check out the branch `af/okteto-test-dev-382` of https://github.com/okteto/go-getting-started
1. Run `okteto test --no-cache` with the latest stable `2.28.0` and observe all artifacts being exported because tests are passing. 
1. Modify the tests (unit/e2e) and break them
1. Re-run with latest stable and observe that no artifacts were exported.
1. Now repest the validation with the CLI from branch and observe artifacts always being exported
1. Check with `echo $?` that the CLI exits non-zero when the tests fail

## CLI Quality Reminders 🔧

### DIFF
--- a/pkg/remote/remote.go
+++ b/pkg/remote/remote.go
@@ -98,13 +98,18 @@ RUN \
   {{range $key, $path := .Caches }}--mount=type=cache,target={{$path}} {{end}}\
   --mount=type=secret,id=known_hosts --mount=id=remote,type=ssh \
   mkdir -p $HOME/.ssh && echo "UserKnownHostsFile=/run/secrets/known_hosts" >> $HOME/.ssh/config && \
-  /okteto/bin/okteto remote-run {{ .Command }} --log-output=json --server-name="${{ .InternalServerName }}" {{ .CommandFlags }} || true
+  /okteto/bin/okteto remote-run {{ .Command }} --log-output=json --server-name="${{ .InternalServerName }}" {{ .CommandFlags }}{{ if eq .Command "test" }} || true{{ end }}
+
+{{range $key, $artifact := .Artifacts }}
+RUN if [ -f /okteto/src/{{$artifact.Path}} ]; then \
+    mkdir -p $(dirname /okteto/artifacts/{{$artifact.Destination}}) && \
+    cp /okteto/src/{{$artifact.Path}} /okteto/artifacts/{{$artifact.Destination}}; \
+  fi
+{{end}}
 
 FROM scratch
 {{ if gt (len .Artifacts) 0 -}}
-{{ range $key, $artifact := .Artifacts -}}
-COPY --from=runner /okteto/src/{{$artifact.Path}} {{$artifact.Destination}}
-{{ end }}
+COPY --from=runner /okteto/artifacts/ /
 {{ else -}}
 COPY --from=runner /etc/.oktetocachekey .oktetocachekey
 {{ end }}
--- a/pkg/remote/remote.go
+++ b/pkg/remote/remote.go
@@ -94,20 +94,11 @@ ARG {{ .InvalidateCacheArgName }}
 RUN echo "${{ .InvalidateCacheArgName }}" > /etc/.oktetocachekey
 RUN okteto registrytoken install --force --log-output=json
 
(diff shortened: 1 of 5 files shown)


pkg/remote/remote.go
```diff
@@ -98,7 +98,7 @@ RUN \
   {{range $key, $path := .Caches }}--mount=type=cache,target={{$path}} {{end}}\
   --mount=type=secret,id=known_hosts --mount=id=remote,type=ssh \
   mkdir -p $HOME/.ssh && echo "UserKnownHostsFile=/run/secrets/known_hosts" >> $HOME/.ssh/config && \
-  /okteto/bin/okteto remote-run {{ .Command }} --log-output=json --server-name="${{ .InternalServerName }}" {{ .CommandFlags }}
+  /okteto/bin/okteto remote-run {{ .Command }} --log-output=json --server-name="${{ .InternalServerName }}" {{ .CommandFlags }} || true
```
pchico83, line 101: @andreafalzetti with this, we won't return a failure to the user, right? we have to keep that behavior.
Maybe something like `|| echo "something to identify command failed" can be used to populate the error to the user

pchico83: > @pchico83 the change makes the docker run step NOT to fail, but the Okteto CLI will continue to fail because `remote-run` continues to log and the build logger parses the output like is today, so the experience is the following:
> 
> Screenshot 2024-06-12 at 15 05 34
> Have a try at the branch if you can. Also the integration test showcase this behaviour

@andreafalzetti oh perfect, thanks for the update!

jLopezbarb (approved):

ifbyol (commented):
@andreafalzetti This change affects also to deploy and destroy operations in remote. Have we checked to do some of those operations from a UI deploy? AS we rely on logs to determine the success of the operation, and the logs is a tricky part when interacting with a UI deployment, it would be interesting to test everything keeps working fine and logs in that scenario are fine too

pkg/remote/remote_test.go
```diff
@@ -325,7 +325,7 @@ func TestCreateDockerfile(t *testing.T) {
 				params: &Params{
 					BaseImage:           "test-image",
 					Manifest:            fakeManifest,
-					BuildEnvVars:        map[string]string{"OKTETO_BUIL_SVC_IMAGE": "ONE_VALUE", "OKTETO_BUIL_SVC2_IMAGE": "TWO_VALUE"},
+					BuildEnvVars:        map[string]string{"OKTETO_BUIL_SVC_IMAGE": "ONE_VALUE", "OKTETO_BUILD_SVC2_IMAGE": "TWO_VALUE"},
```
jLopezbarb, line 328: I think you forgot a D here:
```suggestion
					BuildEnvVars:        map[string]string{"OKTETO_BUILD_SVC_IMAGE": "ONE_VALUE", "OKTETO_BUILD_SVC2_IMAGE": "TWO_VALUE"},
```

andreafalzetti: I copied the typo from the existing test but it's already fixed in 39438f2469503d77697788d1ab0c478bc8a1f1c8

ifbyol (approved):

PR #4336, opened 2024-06-13
fix: okteto test to use always root user

jLopezbarb:
# Proposed changes

Fixes DEV-423

If you set an image on a test that doesn't have root access on okteto test command, now okteto is able to run the command without an error

## How to validate

1. Create the following `Dockerfile`:
``` dockerfile
FROM ubuntu:latest

WORKDIR /app
COPY . .

RUN chown -R 1000:1000 /app
USER 1000
```
2. Create the following okteto manifest
```yaml
build:
  tests:
    context: .
test:
  hello:
    image: $OKTETO_BUILD_TESTS_IMAGE
    commands:
    - echo hello
```
3. Run `okteto test`

## CLI Quality Reminders 🔧


jLopezbarb:
> Is see we are only setting it to `true` in the go tests not in the command, am I missing something?

yup, when I divided the PR in two I didn't add the change to make it work on test 🤦

andreafalzetti (approved):

ifbyol (approved):
Could we add an integration tests for this scenario? It doesn't have to be in this PR

PR #4354, opened 2024-06-19
Admins to be able to deploy into namespaces or previews

teresaromero:
# Proposed changes

Resolves https://okteto.atlassian.net/browse/DEV-247

In order to deploy dev environments into any namespace or preview being an admin, some changes needed to be done.
When checking the user access to a namespace, query namespace and preview to check with API in there is access.

Tests where fixed for the changes introduced on the function.

## How to validate

This change needs current changes for backend https://github.com/okteto/app/pull/8231 
- using the preview environment from backend PR
- create additional user not admin and try to run `okteto deploy` or `okteto pipeline deploy` with the --namespace flag pointing to the user namespace
- the command deploy will fail (pending more changes regarding serviceaccounts)
- the command pipeline deploy will succeed but the pipeline will fail due to the pending changes
- run `okteto namespace use` with previews and namespaces

### DIFF
--- a/pkg/okteto/preview.go
+++ b/pkg/okteto/preview.go
@@ -414,7 +414,7 @@ func (c *previewClient) Get(ctx context.Context, previewName string) (*types.Pre
 	queryStruct := getPreviewQuery{}
 
 	variables := map[string]interface{}{
-		"id": previewName,
+		"id": graphql.String(previewName),
 	}
 	err := query(ctx, &queryStruct, variables, c.client)
 	if err != nil {
--- a/pkg/okteto/preview.go
+++ b/pkg/okteto/preview.go
@@ -406,6 +406,9 @@ func (*previewClient) translateErr(err error, name string) error {
 		return oktetoErrors.UserError{E: ErrUnauthorizedGlobalCreation,
 			Hint: "Please log in with an administrator account or use a personal preview environment"}
 	}
+	if err.Error() == "namespace-not-found" {
+		return oktetoErrors.ErrNamespaceNotFound
+	}
 	return err
 }
 
--- a/pkg/okteto/preview.go
+++ b/pkg/okteto/preview.go
@@ -406,9 +406,6 @@ func (*previewClient) translateErr(err error, name string) error {
 		return oktetoErrors.UserError{E: ErrUnauthorizedGlobalCreation,
 			Hint: "Please log in with an administrator account or use a personal preview environment"}
 	}
-	if err.Error() == "namespace-not-found" {
-		return oktetoErrors.ErrNamespaceNotFound
-	}
 	return err
 }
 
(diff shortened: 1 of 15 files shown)


pkg/okteto/preview.go
```diff
@@ -402,5 +406,25 @@ func (*previewClient) translateErr(err error, name string) error {
...
+// Get gets the given preview environment, returns its ID if found
+func (c *previewClient) Get(ctx context.Context, previewName string) (*types.Preview, error) {
+	queryStruct := getPreviewQuery{}
+
+	variables := map[string]interface{}{
+		"id": graphql.String(previewName),
+	}
+	err := query(ctx, &queryStruct, variables, c.client)
+	if err != nil {
+		return nil, err
```
ifbyol, line 424: I think you need to call here to `translateErr`, if not, the change you did in that function won't be applied in this call, where I believe you need it

ifbyol (commented):
It seems to be working fine for namespaces but not for previews. If I try to execute `okteto ns use <preview>` it fails with `Not found` error even when I'm the owner of the preview and I should be able to do it

Screenshot 2024-06-20 at 16 22 11

`okteto` is the latest stable version and `ok` is the compiled version of this branch. As you can see, executing `okteto ctx use <preview>` works with latest stable but it doesn't for the same preview for this CLI

pkg/okteto/preview.go
teresaromero: `translateErr` is called prior exiting query or mutation functions, so the error is transformed before.

I was investigating this and the cause was the error thrown by the API when a namespace is a preview and we query `space` is "not-found" which does not fall into `namespace-not-found` so it was not looking for the preview.

I added some changes and tests to check this.

I've been doing some smoke test and i fix it. At the capture, the namespace creating gives the following error at API

`ERRO[3566] failed to create space teresaromero-caracolaaa  error="failed to create rolebindings for teresaromero-caracolaaa: error creating kubernetes role binding: RoleBinding.rbac.authorization.k8s.io \"teresaromero-caracolaaa\" is invalid: roleRef.name: Required value" http.request.id=e9f2ba6c-7ee9-4485-af8e-8e3aed188c05 instance.id=teresaromero-okteto-api-okteto-6946d9fcf6-9dqh2 user.id=okteto-admin`

Let me know if you find something else. Thanks!
Screenshot 2024-06-21 at 08 43 26

ifbyol (approved):

pkg/okteto/preview.go
ifbyol: The error about the namespace creation should not happen. Was it that in your dev env? Was it updated with master?

teresaromero: i re-tested, it was puntual error from my dev

PR #4356, opened 2024-06-19
Reuse exec args syntax on up

jLopezbarb:
# Proposed changes

Fixes DEV-438

- Move the arg parser to its own package (previously on exec). This will allow us to reuse it on okteto up and exec and in a future for other commands.
- Use the new logic on exec and up
- Adds new examples for `okteto up --help`

## How to validate

1. Using movies with compose
1. Run `okteto up api -- echo hello` and check that everything is ok
1. Run `okteto up`  and check that the selector is shown
1. Run `okteto up -- echo hello` and check that the selector is shown and the command is the right one

## CLI Quality Reminders 🔧


andreafalzetti:
Testing normal `okteto up` without the cmd, it fails. Is it expected?

Screenshot 2024-06-19 at 16 32 32

andreafalzetti (approved):

andreafalzetti (commented):
Check the linting errors before merging please

ifbyol (commented):
Prior to this change, when I executed `okteto up` and I only had 1 service defined in `dev` section, it was starting automatically the sequence for okteto up, but with these changes, it shows you the selector. Is that expected? 

IMHO, I think we should stick to the prior behavior. If only 1 service is specified, showing the selector only generates friction

ifbyol:
When I execute `okteto exec` without command, it first ask for the dev container where to run the exec, and when you select one, it fails because the command is not specified. Before these changes, we fail before listing the dev containers. I think we should keep failing early. We know that command should be mandatory for exec, so if it is not sepcified, we should fail before requesting the user to perform any other action

teresaromero (approved):

ifbyol (approved):

PR #4357, opened 2024-06-19
remove manifest `context` and `namespace`

andreafalzetti:
# Proposed changes

Fixes: DEV-462
Depends: https://github.com/okteto/okteto/pull/4359 ✅ 

This PR includes removal of from the Okteto Manifest of the following properties:
- root-level `context` and `namespace`
- Dev `context` and `namespace`
- Dependency `context` and `namespace`

## How to validate

This change impacts all commands of the Okteto CLI. Review carefully 🙏 

## CLI Quality Reminders 🔧


cmd/build/command.go
```diff
@@ -215,31 +214,6 @@ func getOktetoContext(ctx context.Context, options *types.BuildOptions) (*okteto
 		Show:      true,
 	}
 
-	// before calling the context command, there is need to retrieve the context and
-	// namespace through the given manifest. If the manifest is a Dockerfile, this
-	// information cannot be extracted so call to GetContextResource is skipped.
-	if err := validateDockerfile(options.File); err != nil {
```
andreafalzetti, line 221: we don't need to detect dockerfile anymore because there is no need to read `context` and `namespace` from the manifest anymore

cmd/deploy/deploy.go
```diff
@@ -392,7 +385,7 @@ func (dc *Command) Run(ctx context.Context, deployOptions *Options) error {
 	dc.IoCtrl.Logger().Debugf("manifest path to store in metadata: %q", manifestPathForConfigMap)
 	data := &pipeline.CfgData{
 		Name:       deployOptions.Name,
-		Namespace:  deployOptions.Manifest.Namespace,
+		Namespace:  okteto.GetContext().Namespace,
```
jLopezbarb, line 388: I think we should use the same context during the function. On line 352 we are using `deployOptions.Namespace`. Is there any reason for using `okteto.Context` vs `deployOptions`?

cmd/remoterun/test.go
```diff
@@ -59,7 +59,7 @@ commands:
 				return fmt.Errorf("--name is required")
 			}
 
-			oktetoContext, err := contextCMD.NewContextCommand().RunStateless(ctx, &contextCMD.Options{})
+			oktetoContext, err := contextCMD.NewContextCommand().RunStateless(ctx, &contextCMD.Options{Show: true})
```
jLopezbarb, line 62: Why do we need to show the context the user is using in the remote? I checked and as it's not staged it's not showing but maybe I'm missing something

cmd/test/cmd.go
```diff
@@ -121,14 +121,8 @@ func Test(ctx context.Context, ioCtrl *io.Controller, k8sLogger *io.K8sLogger, a
...
 
-	// Loads, updates and uses the context from path. If not found, it creates and uses a new context
-	if err := contextCMD.LoadContextFromPath(ctx, options.Namespace, options.K8sContext, options.ManifestPath, contextCMD.Options{Show: true}); err != nil {
-		if err.Error() == fmt.Errorf(oktetoErrors.ErrNotLogged, okteto.GetContext().Name).Error() {
-			return analytics.TestMetadata{}, err
-		}
-		if err := contextCMD.NewContextCommand().Run(ctx, &contextCMD.Options{Namespace: options.Namespace}); err != nil {
-			return analytics.TestMetadata{}, err
-		}
+	if err := contextCMD.NewContextCommand().Run(ctx, &contextCMD.Options{Namespace: options.Namespace}); err != nil {
```
jLopezbarb, line 124: I think we've lost the showing message due to this change

cmd/up/activate.go
```diff
@@ -65,7 +65,7 @@ func (up *upContext) activate() error {
 	if err != nil {
 		return err
 	}
-	app, create, err := utils.GetApp(ctx, up.Dev, k8sClient, up.isRetry)
+	app, create, err := utils.GetApp(ctx, up.Dev, okteto.GetContext().Namespace, k8sClient, up.isRetry)
```
jLopezbarb, line 68: Should we maintain the same way to get the namespace during the function? We already got it as `up.Namespace`. Is this one different?

cmd/up/exec.go
```diff
@@ -440,7 +440,7 @@ func (up *upContext) RunCommand(ctx context.Context, cmd []string) error {
 			hybridCtx := &HybridExecCtx{
 				Dev:       up.Dev,
 				Name:      up.Manifest.Name,
-				Namespace: up.Manifest.Namespace,
+				Namespace: okteto.GetContext().Namespace,
```
jLopezbarb, line 443: Same as before. We have in line 429 `up.Namespace`

cmd/up/up.go
```diff
@@ -236,14 +245,14 @@ okteto up my-svc -- echo this is a test
 				return fmt.Errorf("failed to load k8s client: %w", err)
 			}
 
-			if upOptions.Deploy || !pipeline.IsDeployed(ctx, up.Manifest.Name, up.Manifest.Namespace, k8sClient) {
+			if upOptions.Deploy || !pipeline.IsDeployed(ctx, up.Manifest.Name, okteto.GetContext().Namespace, k8sClient) {
```
jLopezbarb, line 248: As we already have it in the upContext. Should we start using `up.Namespace` instead?

pkg/model/dev.go
```diff
@@ -832,7 +800,7 @@ func (dev *Dev) LabelsSelector() string {
 }
 
 // ToTranslationRule translates a dev struct into a translation rule
-func (dev *Dev) ToTranslationRule(main *Dev, reset bool) *TranslationRule {
+func (dev *Dev) ToTranslationRule(namespace string, main *Dev, reset bool) *TranslationRule {
```
jLopezbarb, line 803: NIT: In Go, variables are ordered by importance, so I think it's weird to have the namespace preceding the dev. During the code we always set `*Dev` first, but this one is just a NIT

jLopezbarb (commented):
Tested and worked correctly with different setups. The only bug that I found was the one in `okteto test` but other than that seems ok to me

pkg/model/dev.go
andreafalzetti: good catch, I've changed all the others to have dev first but not sure what happened here. Fixed in bc31d56ec5ec18dfe0f6ab88d34bbfb044847eb4

cmd/remoterun/test.go
andreafalzetti: mistake addressed here: 8f74637c130af204cb71b1e5abdca09428edb7d2

cmd/test/cmd.go
andreafalzetti: thanks for noticing it, I've fixed it here: 30d6709ffa4c1c5cdeb67d12931a364ad06cde19

cmd/deploy/deploy.go
andreafalzetti: good catch, addressed in b03660f7567bb946bdfb878350840e2623ff0542

cmd/up/exec.go
andreafalzetti: fixed in b262d22a36a539a60335bee8182f6a2c348d7642

cmd/up/activate.go
andreafalzetti: fixed in https://github.com/okteto/okteto/commit/b262d22a36a539a60335bee8182f6a2c348d7642

jLopezbarb (approved):

PR #4359, opened 2024-06-24
remove manifest v1 support

jLopezbarb:
# Proposed changes

Fixes DEV-447

- Remove manifest v1 serialisation
- Remove isV2 from the manifest and analytics
- Remove build when isV1
- Remove syntax for build.Image (only supported by manifest v1)
- Create the new tests and adapt the older ones to use manifest v2

## CLI Quality Reminders 🔧

For both authors and reviewers:

- Scrutinize for potential regressions
- Ensure key automated tests are in place
- Build the CLI and test using the validation steps
- Assess Developer Experience impact (log messages, performances, etc)
- If too broad, consider breaking into smaller PRs
- Adhere to our [code style](https://github.com/okteto/okteto/blob/master/docs/code-style.md) and [code review](https://github.com/okteto/okteto/blob/master/docs/code-review.md) guidelines

PR #4360, opened 2024-06-25
refactor: add test to manifestFields at isManifestFieldNotFound

teresaromero:
# Proposed changes

Fixes https://okteto.atlassian.net/browse/DEV-401

Marshalling of the manifest was failing as `test` was not being recognized as manifest key

## How to validate

Using `go-getting-started`m with some test, leave the manifest just with the test section, then run `okteto test`
Screenshot 2024-06-25 at 12 38 29


ifbyol (commented):
should we include a test? We might include a test that parses a manifest with a single `test` section and make sure the manifest is generated properly with the test section specified

teresaromero:
This would be solve with https://github.com/okteto/okteto/pull/4359

PR #4361, opened 2024-06-25
fix: smart build stuck on untracked files

jLopezbarb:
# Issue

Fixes a bug when calculating the untracked file content didn't timeout correctly. In order to reproduce follow the following steps:

1. On a large JavaScript repo run `yarn install`
2. Don't add `node_modules` to the `.gitignore` or remove the line that adds it
3. Run `okteto build` and check that the 5 second timeout is not working

As part of this PR we have also checked that [exec.CommandContext](https://pkg.go.dev/os/exec#CommandContext) fails correctly after the timeout. If you want to try this, you should change the git command with `"/bin/sleep", "9999")`. 

## Proposed changes

Fixes DEV-479

- Fixes the error by checking if the context has already been cancelled
- Add a log to know when it's running the `git status` command
- Add a log to know the number of files we have to calculate the content (untracked files). If this number is high we should probably need to add some files/folders to the `.dockerignore` to speed up the builds

## How to validate

1. On a large JavaScript repo run `yarn install`
2. Don't add `node_modules` to the `.dockerignore` or remove the line that adds it
3.  Run `okteto build` and check that the 5 second timeout is not working

## CLI Quality Reminders 🔧

### DIFF
--- a/pkg/repository/git.go
+++ b/pkg/repository/git.go
@@ -270,6 +270,7 @@ func (r gitRepoController) GetDiffHash(contextDir string) (string, error) {
 		diff, err := repo.GetDiff(ctx, contextDir, NewLocalGit("git", &LocalExec{}))
 		select {
 		case <-timeoutCh:
+			oktetoLog.Debug("Timeout exceeded calculating git diff: assuming dirty commit")
 		case diffCh <- diffResponse{
 			diff: diff,
 			err:  err,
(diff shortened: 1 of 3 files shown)


andreafalzetti (approved):
tested and works as expected

teresaromero (approved):

PR #4370, opened 2024-06-28
Remove remote check when running build inside deploy remote

teresaromero:
# Proposed changes

Fixes https://okteto.atlassian.net/browse/DEV-213

When running `okteto deploy` in remote, and the manifest had some deploy command that was `okteto build`, this build was being skipped as the builder was running inside the remote container.

- Removing the check for remote environment at the Builder
- Remove unused repository remote controller

### Testing

- Tested manually building the image and using it as OKTETO_REMOTE_CLI_IMAGE
- Tested at the installer


pkg/repository/repository.go
```diff
@@ -75,11 +73,6 @@ func newGitURL(path string) repositoryURL {
 // NewRepository creates a repository controller
 func NewRepository(path string) Repository {
 	var controller repositoryInterface = newGitRepoController(path)
-	// check if we are inside a remote deploy
-	if v := os.Getenv(constants.OktetoDeployRemote); v != "" {
```
andreafalzetti, line 79: could you please help me understand why we can also remove this? I assume other cmds go through this logic in remote 🤔

ifbyol: Because if not, the images built within the remote won't take advantage of Smart Builds, as functions like `GetDiffHash` or `GetLatestDirSHA` returns a `not-implemented` error.

Most of these checks (checking if we are in remote) were added to prevent duplicated behaviors (and sometimes, getting different results) when `okteto deploy --remote` was executing a `okteto deploy` in BuildKit. Instead of splitting the code into 2 different commands, one the deploy triggered and another one with the logic to only execute the needed logic in remote as we have now, we just "blocked" some execution of logic adding those checks. Some examples are: 
* Check in `build.go` was done to prevent that any image could be built as part of the `okteto deploy` executed in BuildKit when they were already built as part of the local execution of `okteto deploy --remote`
* Remote repository controller was done to prevent any "wrong" calculation as part of the `okteto depoy` executed in BuildKit that could lead to indicate wrong images in `OKTETO_BUILD_XXXX` env vars
* `deployInsideDeployConfigMapHandler` was done to prevent that the `okteto deploy` executed in BuildKit modified the same configmap that was being handled by the local `okteto deploy --remote`.

And probably there are more. 

We need to review carefully where this is being used and review possible side effects, but any `okteto` command executed as part of the list of custom commands of an Okteto manifest should have to calculate properly the repository, as it doesn't have be always the one that was triggered with `okteto deploy --remote`. I could clone a repository as part of the commands and execute an okteto command for that repository

jLopezbarb:
Code looks good and testing it I haven't found any weird scenario. The only thing I'd add is if we could get rid off the remote [dockerfile args](https://github.com/okteto/okteto/blob/master/pkg/remote/remote.go#L93) the variable that we were using in the code deleted.

ifbyol:
> Code looks good and testing it I haven't found any weird scenario. The only thing I'd add is if we could get rid off the remote [dockerfile args](https://github.com/okteto/okteto/blob/master/pkg/remote/remote.go#L93) the variable that we were using in the code deleted.

I'm not sure we can remove it. If some custom logic of the customer relies on `OKTETO_COMMIT` would start to fail. I'm not sure if it is a well known variable, but I think we told some customers to use it

jLopezbarb (approved):
I tested everything I could check and it's working properly but I feel like I'm missing an scenario

PR #4375, opened 2024-07-01
feat: remove support for deploy certain services

jLopezbarb:
# Proposed changes

Fixes DEV-445

This PR removes the support for deploying several stack services from an `okteto deploy` command. 

## How to validate

1. Run `okteto deploy api` and check that an error is thrown
1. Using a compose file with dependant services
1. Run just one of the services that depends on another

Example: 
```yaml
version: '3'
services:
  a:
    image: nginx
    container_name: a

  b:
    image: nginx
    container_name: b
    depends_on:
      - a

  c:
    image: nginx
    container_name: c
```

``` yaml
deploy:
  compose:
    - file: docker-compose.yml
      services:
        - b
```
## CLI Quality Reminders 🔧

### DIFF
--- a/cmd/deploy/deploy.go
+++ b/cmd/deploy/deploy.go
@@ -162,6 +162,7 @@ func Deploy(ctx context.Context, at AnalyticsTrackerInterface, insightsTracker b
 	cmd := &cobra.Command{
 		Use:   "deploy",
 		Short: "Execute the list of commands specified in the 'deploy' section of your okteto manifest",
+		Args:  utils.NoArgsAccepted(""),
 		RunE: func(cmd *cobra.Command, _ []string) error {
 			// validate cmd options
 			if options.Dependencies && !okteto.IsOkteto() {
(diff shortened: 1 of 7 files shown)


cmd/deploy/deploy.go
```diff
@@ -72,19 +72,19 @@ type Options struct {
...
-	RunWithoutBash   bool
-	RunInRemote      bool
-	Wait             bool
-	ShowCTA          bool
+	ManifestPath          string
+	Name                  string
+	Namespace             string
+	K8sContext            string
+	Variables             []string
+	StackServicesToDeploy []string
```
andreafalzetti, line 80: why did you rename this to `StackServicesToDeploy`?

andreafalzetti (approved):

cmd/deploy/deploy.go
jLopezbarb: `ServicesToDeploy` was used through different parts of the command and was confusing (at least for me) while validating the user input so I decided to make it more clear. Let me know if we should rename back to its name

PR #4379, opened 2024-07-03
Allow remote deploy inside remote deploy

teresaromero:
# Proposed changes

Fixes https://okteto.atlassian.net/browse/DEV-214

When running an `okteto deploy` with a deploy remote command as part of the deploy manifest, the deploy inside deploy was not being run remotly, but locally based on the runner of the "parent" deploy.

This was caused as we where skippping a remote deploy if we already where inside a remote deploy, this PR removes this check.

Other related changes:
- Remove the remote check from NewConfigmapHandler, as this was causing the child deploy to not create a configmap
- Small fix on log, the log was appearing when deploy and its being reused for testing, used this PR to change this.

## How to validate

- Build the image with the changes and locally use the env OKTETO_REMOTE_CLI_IMAGE to use it on remote runners
- Using 2 manifests

```
# okteto.yml
deploy:
  remote: true
  commands:
    - git clone https://github.com/okteto/movies.git
    - cd movies && okteto deploy --remote
    - echo "hello from okteto.yml"
    - okteto deploy -f other-okteto.yml --remote
```

```
# other-okteto.yml
deploy:
  image: aquasec/trivy:latest
  commands:
    - echo "hello from other-okteto.yml"
    - trivy image python:3.4-alpine
```

run `okteto deploy` will deploy the parent manifest,
-  a configmap for movies should be created and all resources should be under the movies dev environment
- trivy command from child manifest should be run without error

## Tests
integration test && Manual testing running locally and inside the pipeline installer

### DIFF
--- a/integration/deploy/deploy_remote_test.go
+++ b/integration/deploy/deploy_remote_test.go
@@ -53,7 +53,7 @@ func TestDeployInDeployRemote(t *testing.T) {
 	require.NoError(t, createOktetoManifestWithName(dir, parentManifestContent, "okteto.yml"))
 	require.NoError(t, createOktetoManifestWithName(dir, childManifestContent, "other-okteto.yml"))
 
-	testNamespace := "teresaromero-depindep"
+	testNamespace := integration.GetTestNamespace("DeployInDeployRemote", user)
 	namespaceOpts := &commands.NamespaceOptions{
 		Namespace:  testNamespace,
 		Token:      token,
--- a/integration/deploy/deploy_remote_test.go
+++ b/integration/deploy/deploy_remote_test.go
@@ -1,7 +1,7 @@
 //go:build integration
 // +build integration
 
-// Copyright 2023 The Okteto Authors
+// Copyright 2024 The Okteto Authors
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
--- a/integration/deploy/deploy_remote_test.go
+++ b/integration/deploy/deploy_remote_test.go
@@ -28,18 +28,14 @@ var (
 	parentManifestContent = `
 deploy:
   remote: true
-  image: golang:1.22.5-bookworm
   commands:
-    - go version
-    - version=$(go version); if [ $version -ne "go version go1.22.5 linux/amd64" ]; then exit 1; fi
     - okteto deploy -f other-okteto.yml --remote`
 
 	childManifestContent = `
 deploy:
-  image: golang:1.21.12-bookworm
+  image: aquasec/trivy:latest
   commands:
-    - go version
-    - version=$(go version); if [ $version -ne "go version go1.21.12 linux/amd64" ]; then exit 1; fi`
+    - trivy -q image alpine:3.14`
 )
 
 // TestDeployInDeployRemote test the scenario where an okteto deploy is run inside an okteto deploy in remote
--- a/integration/deploy/deploy_remote_test.go
+++ b/integration/deploy/deploy_remote_test.go
@@ -1,0 +1,81 @@
+//go:build integration
+// +build integration
+
+// Copyright 2023 The Okteto Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package deploy
+
+import (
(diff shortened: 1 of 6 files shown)


integration/okteto.go
```diff
@@ -87,7 +87,7 @@ func GetTestNamespace(prefix, user string) string {
 	} else {
 		os = os[:3]
 	}
-	namespace := fmt.Sprintf("%s-%s-%d-%s", prefix, os, time.Now().Unix(), user)
+	namespace := fmt.Sprintf("%s-%s-%s-%d", user, prefix, os, time.Now().Unix())
```
teresaromero, line 90: @jLopezbarb @andreafalzetti i propose to change the test namespace string so it starts with the user, as i move accross environments and in my devcluster the namespace has to start with my user.... do you think this can break ? let me know your thoughts! thanks

integration/deploy/deploy_remote_test.go
```diff
@@ -0,0 +1,77 @@
+//go:build integration
+// +build integration
+
+// Copyright 2023 The Okteto Authors
```
andreafalzetti, line 4:
```suggestion
// Copyright 2024 The Okteto Authors
```

integration/deploy/deploy_remote_test.go
```diff
@@ -0,0 +1,77 @@
...
+	childManifestContent = `
+deploy:
+  image: aquasec/trivy:latest
+  commands:
+    - trivy -q image alpine:3.14`
+)
+
+// TestDeployInDeployRemote test the scenario where an okteto deploy is run inside an okteto deploy in remote
+// image base for the child deploy should be the specified at the child manifest and not the parent
+func TestDeployInDeployRemote(t *testing.T) {
```
andreafalzetti, line 43: when I execute this test locally, it fails with `sh: 1: trivy: not found`.

Does it pass for you, @teresaromero?

teresaromero: locally are you using the custom CLI image? if not it will fail becaue the changes are on the remote runner

andreafalzetti: no I wasn't, thanks!

integration/deploy/manifest_test.go
```diff
@@ -487,7 +487,7 @@ func TestDeployOktetoManifestWithinRepository(t *testing.T) {
 
 	require.NoError(t, createOktetoManifest(dir, simpleOktetoManifestContent))
 
-	testNamespace := integration.GetTestNamespace("ifbyol-DeployManifestWithinRepo", user)
+	testNamespace := integration.GetTestNamespace("DeployManifestWithinRepo", user)
```
teresaromero, line 490: @ifbyol i removed this, was any reason to use your user here?

pkg/remote/remote.go
```diff
@@ -371,7 +371,7 @@ func (r *Runner) Run(ctx context.Context, params *Params) error {
 	if len(params.Artifacts) > 0 {
 		buildOptions.LocalOutputPath = buildCtx
 	}
-	r.ioCtrl.Logger().Infof("Executing test with the following image: %s", params.BaseImage)
+	r.ioCtrl.Logger().Infof("Executing remote with the following base image: %s", params.BaseImage)
```
teresaromero, line 374: @jLopezbarb changed this as is being used by deploy and test cmd

integration/okteto.go
jLopezbarb: Can we create a function and based on an environment  change the behaviour? On staging the user has to be at the end and given that we passed the tests on staging everything will fail

teresaromero: i removed the commit, we can think about this when we improve integration. thanks

integration/deploy/manifest_test.go
ifbyol: No, I tested it against my dev environment and `ifbyol` prefix was required. I forgot to remove it

ifbyol (commented):
I think this would have a side effect that we need to double check. Doing this, we start to pass a `defaultConfigMapHandler` to remote runner when initializing with `NewDeployRunnerForRemote`. https://github.com/okteto/okteto/blob/ca964abd95556b276a46d936db5c311655b96303/cmd/remoterun/deploy.go#L106

That config map handler is used within `runCommandsSection` having an impact on the metrics stored for Okteto Insights (https://github.com/okteto/okteto/blob/ca964abd95556b276a46d936db5c311655b96303/pkg/deployable/deploy.go#L295 and https://github.com/okteto/okteto/blob/ca964abd95556b276a46d936db5c311655b96303/pkg/deployable/deploy.go#L317) and also adding environment variables generated in `OKTETO_ENV` as `dependencyEnvs` (https://github.com/okteto/okteto/blob/ca964abd95556b276a46d936db5c311655b96303/pkg/deployable/deploy.go#L321)

Have we checked the impact on those cases? Before, they were not doing anything and now, they would be applying some changes to the config map. Is that what we want? 

Pinging @jLopezbarb for the Okteto Insights case.

@teresaromero Have you checked if this `a configmap for movies should be created and all resources should be under the movies dev environment` is the same result when executed locally instead of remotely?

PR #4384, opened 2024-07-08
Merge build.name and build.image in Manifest

jLopezbarb:
# Proposed changes

Fixes DEV-451

Remove the `name` field in the `buildInfo` struct and use the `buildInfo.Image` instead

## How to validate

1. Check that in the extended notation the user cannot set `buildInfo.Name` (as it was a field only in okteto `manifest.Dev`)
1. Check that stack continue working as expected
1. Check that build works as expected

## CLI Quality Reminders 🔧


andreafalzetti:
@jLopezbarb check the integration tests, they are failing

pkg/cmd/build/build_test.go
```diff
@@ -486,38 +484,38 @@ func TestExtractFromContextAndDockerfile(t *testing.T) {
 		expectedError      string
 		dockerfilesCreated []string
 	}{
-		{
```
andreafalzetti, line 489: @jLopezbarb what are we going to do with the commented test?

jLopezbarb: I forgot to remove the comment. Tanks for sporting it

pkg/build/info.go
```diff
@@ -119,7 +117,7 @@ func (i *Info) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	var rawString string
 	err := unmarshal(&rawString)
 	if err == nil {
-		i.Name = rawString
+		i.Context = rawString
```
teresaromero, line 120: should this be i.Image ?

pkg/build/info_test.go
```diff
@@ -298,13 +297,12 @@ func TestUnmarshalInfo(t *testing.T) {
 			name:  "unmarshal string",
 			input: "an string value",
 			expected: &Info{
-				Name: "an string value",
+				Context: "an string value",
```
teresaromero, line 300: Same, why are we replacing Name by Context ? should it be Image?

pkg/cmd/build/build.go
```diff
@@ -251,7 +251,7 @@ func OptsFromBuildInfo(manifestName, svcName string, b *build.Info, o *types.Bui
 
 	file := b.Dockerfile
 	if b.Context != "" && b.Dockerfile != "" {
-		file = extractFromContextAndDockerfile(b.Context, b.Dockerfile, svcName)
+		file = extractFromContextAndDockerfile(b.Context, b.Dockerfile, svcName, os.Getwd)
```
teresaromero, line 254: why are we using `os.Getwd` directly, instead of using afero? 🤔

pkg/cmd/build/build_test.go
```diff
@@ -538,6 +536,10 @@ func TestExtractFromContextAndDockerfile(t *testing.T) {
 		},
 	}
 
+	getwd := func() (string, error) {
```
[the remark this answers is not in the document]
jLopezbarb: It's inside the test function so It won't be used anywhere outside the test.

pkg/cmd/build/build.go
jLopezbarb: afero doesn't have `os.Getwd` and I had to mock the call to the system so I added as param so I could test the function

pkg/build/info.go
jLopezbarb: No, this is for the build unmarshaller on this scenario:
```yaml
services:
  api:
    build: api
```
We are passing the context. We were using name to later set the real context

pkg/build/info_test.go
jLopezbarb: This is the test for the previous scenario:
```yaml
services:
  api:
    build: api
```
We are passing the context. We were using name to later set the real context

pkg/cmd/build/build_test.go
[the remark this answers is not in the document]
teresaromero: that is right, i wasnt seeing the whole code in the pr, my bad :D

teresaromero:
tested 👍🏻  post-merged and looks good

andreafalzetti (commented):
LGTM

PR #4390, opened 2024-07-11
deploy waits by default

andreafalzetti:
# Proposed changes

Fixes: DEV-443

Deploy will wait by default

## How to validate

1. Deploy a repo containing several services and endpoints
1. Observe the CLI waiting to deploying all services
1. Observe the endpoints being visible at the end

## CLI Quality Reminders 🔧

### DIFF
--- a/integration/commands/deploy.go
+++ b/integration/commands/deploy.go
@@ -186,9 +186,8 @@ func getDeployCmd(oktetoPath string, deployOptions *DeployOptions) *exec.Cmd {
 	if deployOptions.IsRemote {
 		cmd.Args = append(cmd.Args, "--remote")
 	}
-	if deployOptions.Wait {
-		cmd.Args = append(cmd.Args, "--wait")
-	}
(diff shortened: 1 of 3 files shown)


integration/commands/deploy.go
```diff
@@ -186,9 +186,8 @@ func getDeployCmd(oktetoPath string, deployOptions *DeployOptions) *exec.Cmd {
 	if deployOptions.IsRemote {
 		cmd.Args = append(cmd.Args, "--remote")
 	}
-	if deployOptions.Wait {
```
andreafalzetti, line 189: this only works for passing `Wait: true`

integration/deploy/compose_test.go
```diff
@@ -69,7 +69,9 @@ const (
       retries: 5
       start_period: 30s
   db:
-    image: alpine
```
andreafalzetti, line 72: with `alpine` the init job would exit with an error, we never noticed because the deploy didn't wait

teresaromero (approved):
👍🏻 
i've notice codecov is covering `integration`  this folder should not be tracked?

PR #4401, opened 2024-07-17
Bump kubectl and other binaries. Fix CVEs

maroshii:
Bump kubectl version to fix [CVE-2024-24790](https://github.com/advisories/GHSA-49gw-vxvf-fc2g). This requires a backport to 2.29 and bumping the cli image in the backend to fix the pipeline installer critical CVE

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,8 @@
 # syntax = docker/dockerfile:experimental
 
 ARG KUBECTL_VERSION=1.29.7
-ARG HELM_VERSION=3.14.4
-ARG KUSTOMIZE_VERSION=5.4.1
+ARG HELM_VERSION=3.15.3
+ARG KUSTOMIZE_VERSION=5.4.2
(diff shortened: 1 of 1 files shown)


pchico83 (approved):

ifbyol (approved):

Dockerfile
```diff
@@ -1,8 +1,8 @@
 # syntax = docker/dockerfile:experimental
 
-ARG KUBECTL_VERSION=1.29.4
-ARG HELM_VERSION=3.14.4
-ARG KUSTOMIZE_VERSION=5.4.1
+ARG KUBECTL_VERSION=1.29.7
+ARG HELM_VERSION=3.15.3
```
jLopezbarb, line 4: Just wanted to raise concern that this Helm version support kubernetes 1.30 to 1.27 which differs from what[ we support ](https://www.okteto.com/docs/release-notes/#1220). I don't think we should have any problem but leaving it here in case we face a problem

PR #4402, opened 2024-07-17
feat: validate manifest before executing a command

jLopezbarb:

# Proposed changes

Fixes DEV-466

- Adds a new function to check if the manifest is valid for okteto CLI only
- Use the function on the commands that are available for CLI only instances

## How to validate

1. Use ´okteto context´ to switch to a non okteto instance
1. Run okteto up with any of the non supported fields
1. See error

## CLI Quality Reminders 🔧

### DIFF
--- a/pkg/model/manifest.go
+++ b/pkg/model/manifest.go
@@ -1265,7 +1265,7 @@ func (m *Manifest) ValidateForCLIOnly() error {
 	}
 	if len(invalidFields) > 0 {
 		return oktetoErrors.UserError{
-			E:    fmt.Errorf("the following fields are not allowed in this context: %s", strings.Join(invalidFields, ", ")),
+			E:    fmt.Errorf("these features are only supported if you install Okteto in your cluster: %s", strings.Join(invalidFields, ", ")),
 			Hint: "To install Okteto platform on your cluster check: https://www.okteto.com/docs/get-started/install/",
 		}
 	}
--- a/pkg/model/manifest.go
+++ b/pkg/model/manifest.go
@@ -1265,7 +1265,7 @@ func (m *Manifest) ValidateForCLIOnly() error {
 	}
 	if len(invalidFields) > 0 {
 		return oktetoErrors.UserError{
-			E:    fmt.Errorf("these features are only supported if you install Okteto in your cluster: %s", strings.Join(invalidFields, ", ")),
+			E:    fmt.Errorf("these features are only supported if you install Okteto in your context: %s", strings.Join(invalidFields, ", ")),
 			Hint: "To install Okteto platform on your cluster check: https://www.okteto.com/docs/get-started/install/",
 		}
 	}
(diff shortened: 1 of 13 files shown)


pkg/model/manifest.go
```diff
@@ -1206,3 +1206,68 @@ func getBuildContextForComposeWithVolumeMounts(m *Manifest) (string, error) {
...
+	}
+	if m.Icon != "" {
+		invalidFields = append(invalidFields, "icon")
+	}
+	if m.Name != "" {
+		invalidFields = append(invalidFields, "name")
+	}
+	if len(invalidFields) > 0 {
+		return oktetoErrors.UserError{
+			E:    fmt.Errorf("the following fields are not allowed in this context: %s", strings.Join(invalidFields, ", ")),
```
andreafalzetti, line 1268: I think the error message should not be saying "invalid" fields, but we should say something like:

"These features are only supported if you install Okteto in your cluster" or something along those lines

cc @codyjlandstrom can you help out here?

pkg/model/manifest.go
```diff
@@ -1206,3 +1206,68 @@ func getBuildContextForComposeWithVolumeMounts(m *Manifest) (string, error) {
...
+	}
+	return true
+}
+
+func (mt ManifestTests) IsEmpty() bool {
+	return len(mt) == 0
+}
+
+func (m *Manifest) ValidateForCLIOnly() error {
+	invalidFields := []string{}
```
andreafalzetti, line 1241: I am concerned this function will be missed to be updated when we introduced new top-level properties.

According to the ticket `dev` is the only property supported, if that is true, couldn't we check of any additional property different from `dev` instead of explicitly checking them?

pkg/model/manifest.go
codyjlandstrom: I agree with your suggestion @andreafalzetti - then the user knows what the context is.

"The following features are only supported if you install Okteto in your cluster:"

pkg/model/manifest.go
jLopezbarb: The problem with that is that we initialise some of them because otherwise we have nil pointers exceptions. Changing it to just check dev could be addressed but that will require a lot of changes and possible nil pointers that we can forget updating

pkg/model/manifest.go
andreafalzetti: @codyjlandstrom FYI in various of places of the CLI we use the word "Context" instead of "Cluster". This is something product requested to change some time ago. To keep this consistency I suggest we keep "context" in this change and if we want to change the naming in the future, do it in all places with another PR.

@jLopezbarb WDYT?

andreafalzetti:
@jLopezbarb e2e failing, if you rebase from the 3.0 branch, it's also easier to re-run failed e2e

andreafalzetti (approved):
Approving to unblock, but I would rename "cluster" to "context" and check the e2e that are failing (I think flaky)

PR #4404, opened 2024-07-17
add arg `CacheByName` to remote builds to allow parallel executions

andreafalzetti:
# Proposed changes

Fixes: DEV-522

This change is required to allow `okteto test` to run in parallel. At the moment if a user runs multiple tests in parallel, they are running in sequence, waiting for each to complete. After investigating, the root cause seems to be related to:

> Docker uses a build cache to speed up builds. If both builds are trying to access or modify the same cache layers, Docker might serialize the builds to prevent conflicts.

We're adding a new arg, to allow running each test in parallel.

Disclaimer: this change also affect `deploy` and `destroy` and it will invalidate the cache of existing projects built, even if they are re-built for the same commit etc. The other approach that I've considered is to add this argument only for `okteto test`, we can easily do that, but I think it goes against our current approach of keeping everything aligned as much as possible, otherwise the template starts to become a hell of conditionals. 

## How to validate

1. Use https://github.com/okteto/go-getting-started
2. Add the following to the `okteto.yml`
```yaml
test:
  sleep1:
    commands:
      - echo starting...
      - sleep 60
  sleep2:
    commands:
      - echo starting...
      - sleep 60
  sleep3:
    commands:
      - echo starting...
      - sleep 60
```
3. Run `okteto test sleep1`, `okteto test sleep2` and `okteto test sleep3` in 3 terminals at the same time
4. Observe how the test start without having the others to finish
5. Repeat the same with the latest stable Okteto CLI `2.29.1` and observe that in this case, each test will wait until another one has finished before starting.

## CLI Quality Reminders 🔧

### DIFF
--- a/pkg/remote/remote.go
+++ b/pkg/remote/remote.go
@@ -93,7 +93,7 @@ ENV {{$key}} {{$val}}
 ARG {{ .GitCommitArgName }}
 ARG {{ .GitBranchArgName }}
 ARG {{ .InvalidateCacheArgName }}
-ARG {{ .CacheNamespaceArgName }}
+ARG {{ .CacheByNameArgName }}
 
 RUN echo "${{ .InvalidateCacheArgName }}" > /etc/.oktetocachekey
 RUN okteto registrytoken install --force --log-output=json
@@ -154,8 +154,8 @@ type Params struct {
 	// to a random value which essentially means no-cache. Setting this to a
 	// static or known value will reuse the build cache
 	CacheInvalidationKey string
-	// CacheNamespace is used to allow multiple builds to run concurrently without interfering with each other cache
-	CacheNamespace string
+	// CacheByName is used to allow multiple builds to run concurrently without interfering with each other cache
+	CacheByName    string
 	TemplateName   string
 	DockerfileName string
 	KnownHostsPath string
@@ -200,7 +200,7 @@ type dockerfileTemplateProperties struct {
 	GitCommitArgName         string
 	GitBranchArgName         string
 	InvalidateCacheArgName   string
-	CacheNamespaceArgName    string
+	CacheByNameArgName       string
 	CommandFlags             string
 	OktetoDeployable         string
 	GitHubRepositoryArgName  string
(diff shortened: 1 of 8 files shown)


pkg/remote/remote.go
```diff
@@ -153,10 +154,12 @@ type Params struct {
 	// to a random value which essentially means no-cache. Setting this to a
 	// static or known value will reuse the build cache
 	CacheInvalidationKey string
-	TemplateName         string
-	DockerfileName       string
-	KnownHostsPath       string
-	BaseImage            string
+	// CacheNamespace is used to allow multiple builds to run concurrently without interfering with each other cache
+	CacheNamespace string
```
pchico83, line 158: Why `CacheNamespace`? should it be something like `CachebyName`?

pchico83 (commented):
Not related to this PR, by I am curious about the following ARGs:
```
ARG {{ .NamespaceArgName }}
ARG {{ .GitCommitArgName }}
ARG {{ .GitBranchArgName }}
{{range $key, $val := .OktetoBuildEnvVars }}
ENV {{$key}} {{$val}}
{{end}}
```
Those are going to seriously limiting the efficiency of the test cache

pkg/remote/remote.go
andreafalzetti: I don't mind `CacheByName`, it's very similar to `CacheNamespace` semantically, initially I was going for `Name` but it wasn't specific enough in the end we're adding it for grouping cache builds, so namespace made sense. Does anyone have other names they want to suggest? @jLopezbarb @maroshii

andreafalzetti: I don't want the `namespace` part of `CacheNamespace` get confused with the `OKTETO_NAMESPACE` for instance, so I willl rename it to `CacheByName` for now

andreafalzetti: changed in 3a982e1167924fb2b9e92d7f3c78c17710bbf8d0

andreafalzetti: changed in ffca6730909cdb276e7a08e4cecbb5f7b9d3fa43

andreafalzetti:
> Not related to this PR, by I am curious about the following ARGs:
> 
> ```
> ARG {{ .NamespaceArgName }}
> ARG {{ .GitCommitArgName }}
> ARG {{ .GitBranchArgName }}
> {{range $key, $val := .OktetoBuildEnvVars }}
> ENV {{$key}} {{$val}}
> {{end}}
> ```
> 
> Those are going to seriously limiting the efficiency of the test cache

As you know this file is shared with `deploy` and `destroy`, but we could consider simpifying the caching for `test` if we think it makes sense.

pchico83:
> > Not related to this PR, by I am curious about the following ARGs:
> > ```
> > ARG {{ .NamespaceArgName }}
> > ARG {{ .GitCommitArgName }}
> > ARG {{ .GitBranchArgName }}
> > {{range $key, $val := .OktetoBuildEnvVars }}
> > ENV {{$key}} {{$val}}
> > {{end}}
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Those are going to seriously limiting the efficiency of the test cache
> 
> As you know this file is shared with `deploy` and `destroy`, but we could consider simpifying the caching for `test` if we think it makes sense.

@andreafalzetti I think it's fine for now. If Okteto Test get's traction, and people complain about cache not being optimal, we can always revisit it

andreafalzetti:
I am moving this PR to draft to prevent merging as we noticed the use-case reported might not be fixed with this change

andreafalzetti:
I am closing this PR as it seems an unnecessary fix as of now. cc @jLopezbarb

PR #4418, opened 2024-08-05
Update windows executors to pinned version

teresaromero:
# Proposed changes
https://okteto.atlassian.net/browse/DEV-533
Update CircleCI windows orb settings based on docs.

CircleCI notified that the current orb we are using is using an image which will be deprecated in the following weeks.
According to the docs the "default" image is not availble yet... So this PR pins the version of the executor used, instead of "default" which uses "serever-2022", we pin this version, and also the version of the image inside the orb, instead of "current" use the latests tag available "2024.4.1"

Update guide: https://support.circleci.com/hc/en-us/articles/23614965074459-2024-Image-Deprecations-and-Brownouts-Android-Linux-VM-Remote-Docker-Windows
(tried to just add version: default to image but pipeline would not work)

Docs on orb: https://circleci.com/developer/orbs/orb/circleci/windows#usage-run_windows_2022
Docs on windowns images: https://circleci.com/developer/machine/image/windows-server-2022-gui

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -112,7 +112,9 @@ jobs:
           destination: coverage.txt
 
   run-windows-unit-test:
-    executor: win/default
+    executor:
+      name: win/default
+      version: default
     environment:
       OKTETO_USER: cindylopez
     steps:
@@ -289,7 +291,9 @@ jobs:
             okteto build --platform "linux/arm64,linux/amd64" --build-arg VERSION_STRING=$CIRCLE_SHA1 -f Dockerfile .
 
   test-e2e-setup-windows:
-    executor: win/default
+    executor:
+      name: win/default
+      version: default
     environment:
       OKTETO_CONTEXT: https://staging.okteto.dev
       OKTETO_USER: cindylopez
@@ -316,7 +320,9 @@ jobs:
             - C:\Users\circleci\go\pkg
             - C:\Go\pkg
   e2e-deprecated-windows:
-    executor: win/default
+    executor:
+      name: win/default
+      version: default
     environment:
       OKTETO_CONTEXT: https://staging.okteto.dev
       OKTETO_USER: cindylopez
@@ -347,7 +353,9 @@ jobs:
             go test github.com/okteto/okteto/integration/deprecated/stack -tags="integration" --count=1 -v -timeout 15m
 
   e2e-build-windows:
-    executor: win/default
+    executor:
+      name: win/default
+      version: default
     environment:
       OKTETO_CONTEXT: https://staging.okteto.dev
       OKTETO_USER: cindylopez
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -112,7 +114,10 @@ jobs:
           destination: coverage.txt
 
   run-windows-unit-test:
-    executor: win/default
+    executor:
```
jLopezbarb, line 117: I don´t know if it's possible, but I see that the same executor config is defined across all the different jobs. Could we add the executor definition into top `executors` (line 39) and just [reference it on the rest](https://circleci.com/docs/configuration-reference/#executors). This way when we change it (version or whatever) we just need to change it on one part

.circleci/config.yml
```diff
@@ -316,7 +323,9 @@ jobs:
             - C:\Users\circleci\go\pkg
             - C:\Go\pkg
   e2e-deprecated-windows:
-    executor: win/default
+    executor:
+      name: win/server-2022
+      version: current
```
jLopezbarb, line 328: Why is this version different to the others? Could we use the same version for all the jobs?

teresaromero: my bad

jLopezbarb (approved):

.circleci/config.yml
```diff
@@ -112,7 +114,10 @@ jobs:
           destination: coverage.txt
 
   run-windows-unit-test:
-    executor: win/default
+    executor:
+      # https://circleci.com/developer/machine/image/windows-server-2022-gui
+      name: win/server-2022
+      version: 2024.04.1
```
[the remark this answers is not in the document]
teresaromero: The update guide said... https://support.circleci.com/hc/en-us/articles/23614965074459-2024-Image-Deprecations-and-Brownouts-Android-Linux-VM-Remote-Docker-Windows

> You may need to make additional changes to your config if you are using an Orb that selects a machine image for you. If you're using old and outdated versions CircleCI Orbs that use deprecated images by default you'll need to either:

> - Upgrade the orb to a more recent version (all most recent versions of the CircleCI orbs that use images use non-deprecated images by default). This is the recommended course of action.

> - Ensure that you are explicitly passing in, via the appropriate Orb parameter, the name of a non deprecated image to use.

Using the orb updated (5.0.0) and using the default setting (server-2022) we got the message that it was deprecated... by using "current" as default level. I opted by option 2 and pin the versions, as "default" tag does not exist yet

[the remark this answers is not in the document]
ifbyol: hmmm, ok, I was reading this: https://circleci.com/developer/machine/image/windows-server-2022-gui. In theory, `version: current` should point to `the latest, production ready Windows image`, which I would expect to be the one you were using or at least, the previous one without the need of pinning it, but it doesn't seem to be the case. Thanks for checking it!

PR #4422, opened 2024-08-08
Return success when destroy/delete preview/namespace and err is not found

teresaromero:
# Proposed changes

Fixes https://okteto.atlassian.net/browse/DEV-516

Proposed changes are to return a ~success~ information message when the err from delete client is `not-found`

Screenshot
Screenshot 2024-08-19 at 07 55 59

### DIFF
--- a/cmd/namespace/delete.go
+++ b/cmd/namespace/delete.go
@@ -74,7 +74,7 @@ func (nc *Command) ExecuteDeleteNamespace(ctx context.Context, namespace string,
 	// trigger namespace deletion
 	if err := nc.okClient.Namespaces().Delete(ctx, namespace); err != nil {
 		if oktetoErrors.IsNotFound(err) {
-			oktetoLog.Success("Namespace '%s' not found", namespace)
+			oktetoLog.Information("Namespace '%s' not found", namespace)
 			return nil
 		}
 		return fmt.Errorf("%w: %w", errFailedDeleteNamespace, err)
--- a/cmd/namespace/delete.go
+++ b/cmd/namespace/delete.go
@@ -73,6 +73,10 @@ func (nc *Command) ExecuteDeleteNamespace(ctx context.Context, namespace string,
 
 	// trigger namespace deletion
 	if err := nc.okClient.Namespaces().Delete(ctx, namespace); err != nil {
+		if oktetoErrors.IsNotFound(err) {
+			oktetoLog.Success("Namespace '%s' not found", namespace)
+			return nil
+		}
 		return fmt.Errorf("%w: %w", errFailedDeleteNamespace, err)
 	}
 
--- a/cmd/namespace/delete_test.go
+++ b/cmd/namespace/delete_test.go
@@ -88,7 +88,7 @@ func Test_deleteNamespace(t *testing.T) {
 			toDeleteNs:                      "test-non-existing",
 			initialNamespacesAtOktetoClient: initNamespaces,
 			finalNs:                         currentNamespace,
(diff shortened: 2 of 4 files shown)


andreafalzetti (approved):

cmd/namespace/delete.go
```diff
@@ -73,6 +73,10 @@ func (nc *Command) ExecuteDeleteNamespace(ctx context.Context, namespace string,
 
 	// trigger namespace deletion
 	if err := nc.okClient.Namespaces().Delete(ctx, namespace); err != nil {
+		if oktetoErrors.IsNotFound(err) {
+			oktetoLog.Success("Namespace '%s' not found", namespace)
```
teresaromero, line 77: @codyjlandstrom please when you have the chance review the copy for the messages for this case.
When a namespace or preview(below) is not found when trying to remove it, we are shifting to return no error and say to the user the namespace or preview was not found.

When a user does not have access, we also return not-found. Shall we change the error to other type or we keep this behavour also? 

cc @ifbyol @jLopezbarb

codyjlandstrom: This new copy sounds great.

I think we should keep the current behavior to return a not-found when a user does not have access to a namespace. Similar to how github returns a generic 404 when you try to load a repo that you don't have access to.

codyjlandstrom: Outlining our options for how to handle message when a namespace is not found:

- Option 1: Keep it as is. Success, Namespace not found
- Option 2: Notify the user in an info message, but don't exit. `i` Namespace not found `oktetoLog.Infof("Namespace '%s' not found", namespace)`
- Option 3: Notify the user in a warning message, but don't exit. `Warning` Namespace not found `oktetoLog.Warning("Namespace '%s' not found", namespace)`
- Option 4: Notify the user in a warning message and exit. `Warning` Namespace not found `oktetoLog.Warning("Namespace '%s' not found", exit)`

I'm inclined to suggest Option3, that we warn the user in case they were expecting a namespace to be found in deleted. I don't think we should exit in case a user is running this through an automation or CI pipeline. Curious what everyone else thinks though!

CC: @ifbyol

teresaromero: not sure if with "exit" you mean error exit or just exiting the process? 
i think exiting, meaning, closing the process with any feedback is needed to stop the command, if not, are we waiting for input from the used with correct namespace? that will get the command stuck at ci...

regarding the message i feel more about an information symbol and showing that namespace is not found.

lets see if you have some more feedback! thanks

codyjlandstrom: Ahh to clarify what I meant. I think we should not exit the process and just show the info message. If the process is complete, we can exit.
The use case I was thinking about is when an Okteto admin writes a script to delete many namespaces: if one namespace name doesn't match, we should let them know but not prevent the automation from finishing.

Let me know if that makes sense

teresaromero: Ok, understood the exit thing.

If Warning makes more sense from a product perspective, then i will change to that and let the ppl review

jLopezbarb (approved):

ifbyol:
> # Proposed changes
> Fixes https://okteto.atlassian.net/browse/DEV-516
> 
> Proposed changes are to return a ~success~ information message when the err from delete client is `not-found`
> 
> Screenshot Screenshot 2024-08-19 at 07 55 59

@codyjlandstrom What do you think?

codyjlandstrom:
> @codyjlandstrom What do you think?

That update sounds great!

PR #4423, opened 2024-08-09
Prevent logs to be written from registry credential

teresaromero:
# Proposed changes

https://okteto.atlassian.net/browse/DEV-286

- Create a Silent Formatter for logs to silence the logs, this prevents the JSON formatter to throw the error of empty stange
- Use silent format for oktetoLog implementation, this prevents the logs of info to be shown

Tested:

- Added my DockerHub credentials to my dev, tun the command and see the output
Docs on this https://www.okteto.com/docs/admin/registry-credentials/dockerhub/

```
❯ echo https://index.docker.io/v1/ | ok registrytoken get
{"ServerURL":"https://index.docker.io/v1/","Username":"<username>","Secret":"<token>"}
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -94,12 +94,8 @@ func main() {
 	ioController.Logger().SetLevel(io.WarnLevel)
 	oktetoLog.Init(logrus.WarnLevel) // TODO: Remove when we fully move to ioController
 	if registrytoken.IsRegistryCredentialHelperCommand(os.Args) {
-		oktetoLog.SetOutput(os.Stderr)          // TODO: Remove when we fully move to ioController
-		oktetoLog.SetLevel(oktetoLog.InfoLevel) // TODO: Remove when we fully move to ioController
-		oktetoLog.SetOutputFormat("silent")
-
-		ioController.Logger().SetLevel(io.InfoLevel)
-		ioController.SetOutputFormat(io.JSONFormat)
+		oktetoLog.SetOutputFormat(io.SilentFormat)
+		ioController.SetOutputFormat(io.SilentFormat)
 	}
 
 	var logLevel string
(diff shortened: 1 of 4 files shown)


main.go
```diff
@@ -94,9 +94,9 @@ func main() {
 	ioController.Logger().SetLevel(io.WarnLevel)
 	oktetoLog.Init(logrus.WarnLevel) // TODO: Remove when we fully move to ioController
 	if registrytoken.IsRegistryCredentialHelperCommand(os.Args) {
-		oktetoLog.SetOutput(os.Stderr)                  // TODO: Remove when we fully move to ioController
-		oktetoLog.SetLevel(oktetoLog.InfoLevel)         // TODO: Remove when we fully move to ioController
-		oktetoLog.SetOutputFormat(oktetoLog.JSONFormat) // TODO: Remove when we fully move to ioController
+		oktetoLog.SetOutput(os.Stderr)          // TODO: Remove when we fully move to ioController
+		oktetoLog.SetLevel(oktetoLog.InfoLevel) // TODO: Remove when we fully move to ioController
+		oktetoLog.SetOutputFormat("silent")
```
teresaromero, line 99: I've got to not-show the logs when using this format, as in the `kubetoken` cmd

but

the stage log still shows as this is using the JSONFormat from the ioController, i believe we could have a "silent" format implemented for the ioController ?? @jLopezbarb was there a reason for not implementing this on the new controller? just to understand and perhaps create it to solve this.

@ifbyol @maroshii just for my understanding, is there a reason we need to have them json or could they be silent as in kubetoken?

jLopezbarb: There was no reason. We didn't need it back at the time so we choose to only implement the minimal things. I'd add it if it's necessary to fix the issue without hesitation

teresaromero:
> To be on the safe side, I would test that deploying an dev env with a private image works in the installer works

Tested the following:
- Updated installer with this branch build
- Created token for my dockerhub
- Created private repository at dockerhub "teresaromero/private"
- Used `compose-getting-started` as repository, replaced the `vote` service with the image `teresaromero/private:vote`
- Run `okteto build vote` && `okteto deploy` 

All seems to be working fine, it would be great if someone else could have a chance to test this just to be sure, as i had some setting up problems and this was my second try it worked.

ifbyol:
I was validating it and it is working fine

PR #4426, opened 2024-08-13
DEV-575: Fix UpdateEnvs function to allow variable values with '='

ifbyol:
# Proposed changes

Fixes DEV-575

We had a regression in the last release when deploying something remotely and a variable with a `=` in the value is stored in `OKTETO_ENV`

The simplest reproduction scenario is having a manifest like this one:

```
deploy:
  remote: true
  commands:
    - echo "URL=https://okteto.docs?okteto=rocks" >> $OKTETO_ENV
```

If you execute `okteto deploy`, it succeeds in `2.29.x` version, but it fails on `2.30.x`. The error returned is `Error building image '': build failed: failed to solve: process "/bin/sh -c mkdir -p $HOME/.ssh && echo \"UserKnownHostsFile=/run/secrets/known_hosts\" >> $HOME/.ssh/config &&   /okteto/bin/okteto remote-run deploy --log-output=json --server-name=\"$INTERNAL_SERVER_NAME\" --name \"test-okteto-env\"" did not complete successfully: exit code: 1` which is not really helpful.

The reason is because since `2.30.x`, remote executions also update the configmap to store the `OKTETO_ENV` and enters into the function `UpdateEnvs`, which is the one failing.

Apart from fixing that scenario, this PR also address:
* Fixing also the local execution when the variable had `=`. This was like that several versions. In this case, the error returned was something more useful, but now, it doesn't fail
* Added some code to return a more clear error if there is some issue updating the `OKTETO_ENV` at the end of the execution. In that way, we return a clear error why it has failed.

Screenshot 2024-08-13 at 10 40 28

## How to validate

Check the manifest described in the description and:

1. Execute `okteto deploy` on `2.30`. You would see how it fails with the error mentioned above.
2. Use this branch and execute it again. You will see how it doesn't fail anymore and the variables are properly stored in the configmap under the key `dependencyEnvs`

## CLI Quality Reminders 🔧

### DIFF
--- a/pkg/cmd/pipeline/translate.go
+++ b/pkg/cmd/pipeline/translate.go
@@ -167,9 +167,9 @@ func UpdateEnvs(ctx context.Context, name, namespace string, envs []string, c ku
 
 	if cmap != nil {
 		envsToSet := make(map[string]string, len(envs))
+		envFormatParts := 2
 		for _, env := range envs {
-			result := strings.Split(env, "=")
-			envFormatParts := 2
+			result := strings.SplitN(env, "=", envFormatParts)
 			if len(result) != envFormatParts {
 				return fmt.Errorf("invalid env format: '%s'", env)
 			}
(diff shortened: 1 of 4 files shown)


teresaromero (approved):

jLopezbarb (approved):
Tested and works as expected

PR #4429, opened 2024-08-21
fix: syncthing restarts with relay enabled

jLopezbarb:

# Proposed changes

Fixes #DEV-395

- Adds a new volume as a source of truth for the syncthing config which is separated from the syncthing data
- Bump okteto binary to use these new volumes

Depends on:
- https://github.com/okteto/supervisor/pull/15
- https://github.com/okteto/bin-image/pull/12

## How to validate

1. Run okteto up 
1. Remove /var/syncthing/config.xml
1. Kill syncthing process 
1. Check that the /var/syncthing/config.xml is regenerated with the relay enabled to false

## CLI Quality Reminders 🔧

### DIFF
--- a/pkg/k8s/apps/translate.go
+++ b/pkg/k8s/apps/translate.go
@@ -37,8 +37,7 @@ const (
 
 	// syncthing
 	oktetoSyncSecretVolume = "okteto-sync-secret" // skipcq GSC-G101  not a secret
-	oktetoSyncDataVolume   = "okteto-sync-data"
-	oktetoDevSecretVolume  = "okteto-dev-secret" // skipcq GSC-G101  not a secret
+	oktetoDevSecretVolume  = "okteto-dev-secret"  // skipcq GSC-G101  not a secret
 	oktetoSecretTemplate   = "okteto-%s"
 )
 
@@ -332,10 +331,6 @@ func TranslateVolumeMounts(c *apiv1.Container, rule *model.TranslationRule) {
 			MountPath: "/var/syncthing/secret/",
 			ReadOnly:  true,
 		},
-		apiv1.VolumeMount{
-			Name:      oktetoSyncDataVolume,
-			MountPath: "/var/syncthing/data/",
-		},
 	)
 
 	if len(rule.Secrets) > 0 {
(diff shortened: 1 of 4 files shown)


pkg/model/dev.go
```diff
@@ -43,7 +43,7 @@ import (
 
 var (
 	// OktetoBinImageTag image tag with okteto internal binaries
-	OktetoBinImageTag = "okteto/bin:1.5.0"
+	OktetoBinImageTag = "jlopezbarb/bin-image:1.6.0"
```
andreafalzetti, line 46:
```suggestion
	OktetoBinImageTag = "okteto/bin:1.6.0"
```

andreafalzetti: reminder to change this before merging

pkg/k8s/apps/translate.go
```diff
@@ -561,44 +568,71 @@ func TranslateOktetoInitFromImageContainer(spec *apiv1.PodSpec, rule *model.Tran
...
-						Mode: &mode,
-					},
-					{
-						Key:  "key.pem",
-						Path: "key.pem",
-						Mode: &mode,
+// TranslateOktetoSyncthingVolumes translates the syncthing secret container of a pod
+func TranslateOktetoSyncthingVolumes(spec *apiv1.PodSpec, name string) {
+	if spec.Volumes == nil {
+		spec.Volumes = []apiv1.Volume{}
```
andreafalzetti, line 592: @jLopezbarb the unit test `Test_translateWithVolumes` needs to be updated reflecting these changes

andreafalzetti (approved):
I was able to reproduce the bug and validate the fix, nice work.

Approving with comments:

1. fix the tests
2. add PR label
3. change `OktetoBinImageTag`

pkg/model/dev.go
```diff
@@ -43,7 +43,7 @@ import (
 
 var (
 	// OktetoBinImageTag image tag with okteto internal binaries
-	OktetoBinImageTag = "okteto/bin:1.5.0"
+	OktetoBinImageTag = "jlopezbarb/bin:1.6.0"
```
ifbyol, line 46: Remember to change it to the good image name

ifbyol (commented):
If someone was disconnected from okteto up, and executes okteto up, will this work fine?

jLopezbarb:
> If someone was disconnected from okteto up, and executes okteto up, will this work fine?

What do you mean by disconnected? Does an exit works? Everytime that we start an `okteto up` session we copy the files that could be removed so it'll work for that case

ifbyol:
> > If someone was disconnected from okteto up, and executes okteto up, will this work fine?
> 
> What do you mean by disconnected? Does an exit works? Everytime that we start an `okteto up` session we copy the files that could be removed so it'll work for that case

I mean any disconnection from the pod which doesn't imply to delete the dev container. `exit` would be the same case I guess

andreafalzetti (approved):

ifbyol (approved):

PR #4435, opened 2024-08-27
Add EC encryption to `okteto up`

### LINKED ISSUE #4399: Add configurable SSH server key type

Okteto generates an RSA key when creating the SSH server for `okteto up`.

https://github.com/okteto/okteto/blob/29b67e00c07a33576f845f212012614c067f6332/pkg/ssh/key.go#L87

Visual Studio 2022 V17.10 no longer supports RSA key for remote debugging.

https://learn.microsoft.com/en-us/cpp/linux/connect-to-your-remote-linux-computer?view=msvc-170#set-up-the-remote-connection

> Versions of Visual Studio before 17.10 support EC, RSA, and DSA keys for remote connections. Because of security concerns, RSA and DSA keys are no longer supported in VS 17.10 and later. Only EC keys are currently supported. To create a key pair compatible with the connection manager use the command: ssh-keygen -m pem -t ecdsa -f <key-name>

The ability to configure what type of SSH key is generated to support backward and forward compatibility.

The ability to tell Okteto what key file to use for it's ssh server.

jLopezbarb:
# Proposed changes

Fixes #4399 DEV-603

- Change the RSA key generated by okteto to a EC key
- Changes the remote SSH server to start using EC keys instead of RSA (Done in https://github.com/okteto/remote)

Depends on:
- https://github.com/okteto/remote/pull/33
- https://github.com/okteto/bin-image/pull/14

## How to validate

Use the variable `OKTETO_BIN=jlopezbarb/bin:ecdsa` by running `export OKTETO_BIN=jlopezbarb/bin:ecdsa`

Test using https://github.com/okteto/go-getting-started but it works with any repo

1. Run `okteto up` and check that you get the message  `✓  Client certificates generated`
2. On another terminal run ssh `hello-world.okteto` and check that the warning mentions `ECDSA` (`Warning: Permanently added '[localhost]:59218' (ECDSA) to the list of known hosts`)
3. Check on `.ssh/config` that your last entry (or the one for `hello-world.okteto`) has an `ecdsa` encryption
4. From the `.ssh/config`, get the `IdentityFile`, read it and check that the file is an elliptic curve key file

## CLI Quality Reminders 🔧


pkg/ssh/key.go
```diff
@@ -29,9 +29,12 @@ import (
...
 const (
-	privateKeyFile = "id_rsa_okteto"
-	publicKeyFile  = "id_rsa_okteto.pub"
-	bitSize        = 4096
+	privateKeyFile = "id_ecdsa_okteto"
+	publicKeyFile  = "id_ecdsa_okteto.pub"
+)
+
+var (
+	curve = elliptic.P256() // You can use P256(), P384(), or P521()
```
ifbyol, line 37: Have we checked this is a valid one for the main scenario we want to fix?

jLopezbarb: Yup, This is the list of available algos:
Screenshot 2024-08-28 at 13 19 30

teresaromero (approved):

pkg/ssh/key.go
teresaromero: would it be interesting to allow the user to choose this? just a thought, in case some algorithm is not supported we could have different algos for the user to choose ?

jLopezbarb: I think it's not worth adding that because the maintainability vs the use is not worth. 
It's something doable but would require a new field in the manifest which is something we don't want right now. Then we should match with the algo in the remote container and on the start select the algo and create the key programatically adding more time to the start, but it's something we could reevaluate in the future.

teresaromero: yup, not for this PR, just mentioning as a "future" thingy! thanks

ifbyol (approved):
I was testing it and it looks good. I testes remote connection from Goland and VSCode.

We would need to test it in Visual Studio, but it looks good

andreafalzetti:
Tested in GoLang ✅

PR #4436, opened 2024-08-27
feat: Add env vars wit prefix OKTETO_ENV 

jLopezbarb:
# Proposed changes

Fixes DEV-580

Adds all the possible feature flags to the deploy on remote

## How to validate

1. Using the following `okteto.yml`:
```yaml
deploy:
  - echo $OKTETO_FEATURE_FLAG
```
2. Run `export OKTETO_FEATURE_FLAG=true`
3. Run `okteto deploy --remote` and check that the value is the correct one

## CLI Quality Reminders 🔧

### DIFF
--- a/pkg/remote/remote.go
+++ b/pkg/remote/remote.go
@@ -585,6 +585,9 @@ func getOktetoPrefixEnvVars(environ []string) map[string]string {
 	envFormatParts := 2
 	for _, v := range environ {
 		result := strings.SplitN(v, "=", envFormatParts)
+		if len(result) != envFormatParts {
+			continue
+		}
 		key := result[0]
 		if strings.HasPrefix(key, "OKTETO_") {
 			value := result[1]
--- a/pkg/remote/remote.go
+++ b/pkg/remote/remote.go
@@ -63,7 +63,9 @@ USER 0
 ENV PATH="${PATH}:/okteto/bin"
 COPY --from=okteto-cli /usr/local/bin/* /okteto/bin/
 
-
+{{range $key, $val := .OktetoPrefixEnvVars }}
+ENV {{$key}} {{$val}}
+{{end}}
 ENV {{ .RemoteDeployEnvVar }} true
 ARG {{ .NamespaceArgName }}
 ARG {{ .ContextArgName }}
@@ -141,6 +143,7 @@ type Runner struct {
 	// sshAuthSockEnvvar is the default for SSH_AUTH_SOCK. Provided mostly for testing
 	sshAuthSockEnvvar  string
 	useInternalNetwork bool
+	getEnviron         func() []string
 }
 
 // Params struct to pass the necessary parameters to create the Dockerfile
@@ -187,6 +190,7 @@ type dockerfileTemplateProperties struct {
 	RemoteDeployEnvVar       string
 	OktetoBuildEnvVars       map[string]string
 	OktetoDependencyEnvVars  map[string]string
+	OktetoPrefixEnvVars      map[string]string
 	ContextArgName           string
 	NamespaceArgName         string
 	TokenArgName             string
@@ -219,6 +223,7 @@ func NewRunner(ioCtrl *io.Controller, builder Builder) *Runner {
 		ioCtrl:               ioCtrl,
 		builder:              builder,
 		oktetoClientProvider: okteto.NewOktetoClientProvider(),
+		getEnviron:           os.Environ,
 	}
 }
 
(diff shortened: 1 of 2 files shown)


pkg/remote/remote_test.go
```diff
@@ -817,3 +835,24 @@ func TestCreateDockerignoreFileWithFilesystem(t *testing.T) {
...
+	environ := []string{
+		"OKTETO_ENV_VAR_1=value1",
+		"OKTETO_ENV_VAR_2=value2",
+		"NON_OKTETO_ENV_VAR=value3",
+	}
+
+	prefixEnvVars := getOktetoPrefixEnvVars(environ)
+
+	for key, value := range expectedEnvVars {
+		if prefixEnvVars[key] != value {
```
andreafalzetti, line 854: I would use `assert.Equal`.

pkg/remote/remote.go
```diff
@@ -573,3 +579,13 @@ func GetOriginalCWD(workingDirectoryCtrl filesystem.WorkingDirectoryInterface, m
 	manifestPathDir := filepath.Dir(filepath.Clean(fmt.Sprintf("/%s", manifestPath)))
 	return strings.TrimSuffix(cwd, manifestPathDir), nil
 }
+
+func getOktetoPrefixEnvVars(environ []string) map[string]string {
+	prefixEnvVars := make(map[string]string)
+	for _, v := range environ {
+		if strings.HasPrefix(v, "OKTETO_") {
```
andreafalzetti, line 586: it's unlikely to happen if we use `os.Environ()` as the source, but you never know. This func will panic for values such as `"OKTETO_ENV_VAR_2`, with missing `=`.

We had a similar regression lately (see https://github.com/okteto/okteto/pull/4426/files), I would write defensive code, ensuring the edge-cases are covered, just in case.

pkg/remote/remote_test.go
jLopezbarb: We are getting the `os.Environ` so any variable that is locally will be injected. I want to confirm that only the one ones I added in the test are the correct ones.

andreafalzetti: you can still use 

```go
assert.Equal(t, value, prefixEnvVars[key])
```

and remove the if, no?

pkg/remote/remote_test.go
```diff
@@ -817,3 +835,24 @@ func TestCreateDockerignoreFileWithFilesystem(t *testing.T) {
...
+			t.Errorf("Expected value for environment variable %s is %s, but got %s", key, value, prefixEnvVars[key])
+		}
```
andreafalzetti, line 856:
```suggestion
		assert.Equal(t, value, prefixEnvVars[key])
```

pkg/remote/remote.go
```diff
@@ -573,3 +579,17 @@ func GetOriginalCWD(workingDirectoryCtrl filesystem.WorkingDirectoryInterface, m
 	manifestPathDir := filepath.Dir(filepath.Clean(fmt.Sprintf("/%s", manifestPath)))
 	return strings.TrimSuffix(cwd, manifestPathDir), nil
 }
+
+func getOktetoPrefixEnvVars(environ []string) map[string]string {
+	prefixEnvVars := make(map[string]string)
+	envFormatParts := 2
+	for _, v := range environ {
+		result := strings.SplitN(v, "=", envFormatParts)
```
andreafalzetti, line 587: This will still panic if a variable has no `=`. I would add the following

```
if len(result) != envFormatParts {
  continue
}
```

pkg/remote/remote_test.go
```diff
@@ -817,3 +835,25 @@ func TestCreateDockerignoreFileWithFilesystem(t *testing.T) {
 		})
 	}
 }
+
+func TestGetOktetoPrefixEnvVars(t *testing.T) {
```
teresaromero, line 839: nit: i've tested with lowecase env... and its not selected. We enforce uppercase only or shall we consider `okteto_env` also a possible env? 🤔

teresaromero (approved):

andreafalzetti (approved):

pkg/remote/remote_test.go
andreafalzetti: good point, I think it could be an edge case that if someone cannot work around we might consider adding

jLopezbarb: I don't think it's worth adding them. This feature was for the feature flag where we require them to be uppercase

pkg/remote/remote.go
```diff
@@ -63,7 +63,9 @@ USER 0
 ENV PATH="${PATH}:/okteto/bin"
 COPY --from=okteto-cli /usr/local/bin/* /okteto/bin/
 
-
+{{range $key, $val := .OktetoPrefixEnvVars }}
+ENV {{$key}} {{$val}}
+{{end}}
```
ifbyol, line 68: We are setting several environment variables here, but then, we set some ARG (probably some of them are duplicated). Could there be some side effect? I guess the `ARG` would override the `EVN` defined previously?

jLopezbarb: Yup, I tested and put them at the beginning because of that. Only if the arg is empty it's not overwritten which is something we are not doing.

PR #4448, opened 2024-09-06
feat: Add extractImageTagFromPullAccessDeniedError function

jLopezbarb:
# Proposed changes

Fixes DEV-407

If there is no tagImage we get the name of the image that is failing from the error message. We need to use a regex because the grpc error is not exported and can't be used as an error. This has the problem that if something change we'd need to change but it's the only way to do it right now

This commit adds a new function `extractImageTagFromPullAccessDeniedError` to the `errors.go` file. This function extracts the image tag from the error message when a pull access is denied. It uses a regular expression to find the image tag and returns it. This function is used in the `getErrorMessage` function to provide a more informative error message when pulling an inaccessible or non-existent image.

| **Before**                                   | **After**                |
|----------------------------------------------|--------------------------|
| Screenshot 2024-09-06 at 12 35 06 | Screenshot 2024-09-06 at 12 35 25 |

## How to validate

1. On any repo with a deploy section, add deploy.image and set it to a non existent image
1. See error

## CLI Quality Reminders 🔧

### DIFF
--- a/cmd/deploy/remote.go
+++ b/cmd/deploy/remote.go
@@ -143,6 +143,7 @@ func (rd *remoteDeployer) Deploy(ctx context.Context, deployOptions *Options) er
 		oktetoLog.SetStage("remote deploy")
 		var userErr oktetoErrors.UserError
 		if errors.As(err, &userErr) {
+			oktetoLog.AddToBuffer(oktetoLog.ErrorLevel, "error deploying application: %s", userErr.Error())
 			return userErr
 		}
 		return fmt.Errorf("error deploying application: %w", err)
--- a/pkg/cmd/build/errors.go
+++ b/pkg/cmd/build/errors.go
@@ -35,12 +35,12 @@ func getErrorMessage(err error, tag string) error {
 	switch {
 	case isLoggedIntoRegistryButDontHavePermissions(err):
 		err = oktetoErrors.UserError{
-			E:    fmt.Errorf("error building image '%s': You are not authorized to push image '%s'", tag, imageTag),
+			E:    fmt.Errorf("failed to push image '%s': You are not authorized to push image '%s'", tag, imageTag),
 			Hint: fmt.Sprintf("Please log in into the registry '%s' with a user with push permissions to '%s' or use another image.", imageRegistry, imageTag),
 		}
 	case isNotLoggedIntoRegistry(err):
 		err = oktetoErrors.UserError{
-			E:    fmt.Errorf("error building image '%s': You are not authorized to push image '%s'", tag, imageTag),
+			E:    fmt.Errorf("failed to push image '%s': You are not authorized to push image '%s'", tag, imageTag),
 			Hint: fmt.Sprintf("Log in into the registry '%s' and verify that you have permissions to push the image '%s'.", imageRegistry, imageTag),
 		}
 	case isBuildkitServiceUnavailable(err):
@@ -53,7 +53,7 @@ func getErrorMessage(err error, tag string) error {
 			imageTag = extractImageTagFromPullAccessDeniedError(err)
 		}
 		err = oktetoErrors.UserError{
-			E:    fmt.Errorf("error building image: failed to pull image '%s'. The repository is not accessible or it does not exist", imageTag),
+			E:    fmt.Errorf("failed to pull image '%s'. The repository is not accessible or it does not exist", imageTag),
 			Hint: fmt.Sprintf("Please verify the name of the image '%s' to make sure it exists.", imageTag),
 		}
 	default:
--- a/pkg/cmd/build/errors.go
+++ b/pkg/cmd/build/errors.go
@@ -16,6 +16,7 @@ package build
 import (
 	"errors"
 	"fmt"
+	"regexp"
 	"strings"
 
 	oktetoErrors "github.com/okteto/okteto/pkg/errors"
@@ -48,6 +49,9 @@ func getErrorMessage(err error, tag string) error {
 			Hint: "Please try again later.",
 		}
 	case isPullAccessDenied(err):
+		if imageTag == "" {
+			imageTag = extractImageTagFromPullAccessDeniedError(err)
+		}
 		err = oktetoErrors.UserError{
 			E:    fmt.Errorf("error building image: failed to pull image '%s'. The repository is not accessible or it does not exist", imageTag),
 			Hint: fmt.Sprintf("Please verify the name of the image '%s' to make sure it exists.", imageTag),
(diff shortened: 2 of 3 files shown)


ifbyol:
@jLopezbarb How does this look like in the UI when this happens? The original report is that we only show an `exit code 1` in the UI. Does that have changed?

jLopezbarb:
> @jLopezbarb How does this look like in the UI when this happens? The original report is that we only show an `exit code 1` in the UI. Does that have changed?

I misunderstood the original issue. What would be the best stage to set it? Should it be something like `Remote setup`?

ifbyol:
> > @jLopezbarb How does this look like in the UI when this happens? The original report is that we only show an `exit code 1` in the UI. Does that have changed?
> 
> I misunderstood the original issue. What would be the best stage to set it? Should it be something like `Remote setup`?

@codyjlandstrom @pchico83 How do we want to show it in the UI? A specific stage or a generic one with the error?

andreafalzetti (approved):

pchico83:
> > > @jLopezbarb How does this look like in the UI when this happens? The original report is that we only show an `exit code 1` in the UI. Does that have changed?
> > 
> > 
> > I misunderstood the original issue. What would be the best stage to set it? Should it be something like `Remote setup`?
> 
> @codyjlandstrom @pchico83 How do we want to show it in the UI? A specific stage or a generic one with specific error message? (For context, with stage we refer to the title we use to distinguish different phases of the operation and that can be collapsed)

Could it be part of the "Load Manifest"? or at what moment do we catch the error?

jLopezbarb:
> > > > @jLopezbarb How does this look like in the UI when this happens? The original report is that we only show an `exit code 1` in the UI. Does that have changed?
> > > 
> > > 
> > > I misunderstood the original issue. What would be the best stage to set it? Should it be something like `Remote setup`?
> > 
> > 
> > @codyjlandstrom @pchico83 How do we want to show it in the UI? A specific stage or a generic one with specific error message? (For context, with stage we refer to the title we use to distinguish different phases of the operation and that can be collapsed)
> 
> Could it be part of the "Load Manifest"? or at what moment do we catch the error?

We catch it during build (deploy) time. Checking the images at the beginning when loading the manifest is something doable (at least the one set by the user).

pchico83:
> > > > > @jLopezbarb How does this look like in the UI when this happens? The original report is that we only show an `exit code 1` in the UI. Does that have changed?
> > > > 
> > > > 
> > > > I misunderstood the original issue. What would be the best stage to set it? Should it be something like `Remote setup`?
> > > 
> > > 
> > > @codyjlandstrom @pchico83 How do we want to show it in the UI? A specific stage or a generic one with specific error message? (For context, with stage we refer to the title we use to distinguish different phases of the operation and that can be collapsed)
> > 
> > 
> > Could it be part of the "Load Manifest"? or at what moment do we catch the error?
> 
> We catch it during build (deploy) time. Checking the images at the beginning when loading the manifest is something doable (at least the one set by the user).

What's the easiest solution? To show it as part of the deploy step? If yes, I think that's a good balance

jLopezbarb:
> > > > > > @jLopezbarb How does this look like in the UI when this happens? The original report is that we only show an `exit code 1` in the UI. Does that have changed?
> > > > > 
> > > > > 
> > > > > I misunderstood the original issue. What would be the best stage to set it? Should it be something like `Remote setup`?
> > > > 
> > > > 
> > > > @codyjlandstrom @pchico83 How do we want to show it in the UI? A specific stage or a generic one with specific error message? (For context, with stage we refer to the title we use to distinguish different phases of the operation and that can be collapsed)
> > > 
> > > 
> > > Could it be part of the "Load Manifest"? or at what moment do we catch the error?
> > 
> > 
> > We catch it during build (deploy) time. Checking the images at the beginning when loading the manifest is something doable (at least the one set by the user).
> 
> What's the easiest solution? To show it as part of the deploy step? If yes, I think that's a good balance

That's the easiest one but we need to add a new Stage or use the Internal server error

pchico83:
> > > > > > > @jLopezbarb How does this look like in the UI when this happens? The original report is that we only show an `exit code 1` in the UI. Does that have changed?
> > > > > > 
> > > > > > 
> > > > > > I misunderstood the original issue. What would be the best stage to set it? Should it be something like `Remote setup`?
> > > > > 
> > > > > 
> > > > > @codyjlandstrom @pchico83 How do we want to show it in the UI? A specific stage or a generic one with specific error message? (For context, with stage we refer to the title we use to distinguish different phases of the operation and that can be collapsed)
> > > > 
> > > > 
> > > > Could it be part of the "Load Manifest"? or at what moment do we catch the error?
> > > 
> > > 
> > > We catch it during build (deploy) time. Checking the images at the beginning when loading the manifest is something doable (at least the one set by the user).
> > 
> > 
> > What's the easiest solution? To show it as part of the deploy step? If yes, I think that's a good balance
> 
> That's the easiest one but we need to add a new Stage or use the Internal server error

Why? can we use the "deploy" stage?

jLopezbarb:
The change affects all the uncontrolled errors that happens during the remote deploy. We already had a stage but weren't adding the errors to the configmap (probably we missed that part). Other errors that will be shown are buildkit not available and permission errors.
Here is how it looks like in the UI:
Screenshot 2024-09-23 at 18 36 08
And here how it looks like in the Terminal
Screenshot 2024-09-23 at 18 38 20
cc: @pchico83

pchico83:
> The change affects all the uncontrolled errors that happens during the remote deploy. We already had a stage but weren't adding the errors to the configmap (probably we missed that part). Other errors that will be shown are buildkit not available and permission errors. Here is how it looks like in the UI: Screenshot 2024-09-23 at 18 36 08 And here how it looks like in the Terminal Screenshot 2024-09-23 at 18 38 20 cc: @pchico83

@jLopezbarb I'm not sure if it's out of the scope of this PR, but the part "error building image" is misleading, isn't it?

codyjlandstrom:
> > The change affects all the uncontrolled errors that happens during the remote deploy. We already had a stage but weren't adding the errors to the configmap (probably we missed that part). Other errors that will be shown are buildkit not available and permission errors. Here is how it looks like in the UI: Screenshot 2024-09-23 at 18 36 08 And here how it looks like in the Terminal Screenshot 2024-09-23 at 18 38 20 cc: @pchico83
> 
> @jLopezbarb I'm not sure if it's out of the scope of this PR, but the part "error building image" is misleading, isn't it?

I think just saying `Failed to pull image 'x'` would be more straightforward for the user

jLopezbarb:
> > > The change affects all the uncontrolled errors that happens during the remote deploy. We already had a stage but weren't adding the errors to the configmap (probably we missed that part). Other errors that will be shown are buildkit not available and permission errors. Here is how it looks like in the UI: Screenshot 2024-09-23 at 18 36 08 And here how it looks like in the Terminal Screenshot 2024-09-23 at 18 38 20 cc: @pchico83
> > 
> > 
> > @jLopezbarb I'm not sure if it's out of the scope of this PR, but the part "error building image" is misleading, isn't it?
> 
> I think just saying `Failed to pull image 'x'` would be more straightforward for the user

I think it's out of scope, but I addressed on https://github.com/okteto/okteto/pull/4448/commits/be7c86a3287a94b9c5156aec838a2c979164dd3e

PR #4452, opened 2024-09-11
fix(remote): :bug: multiline preparation on dockerfile

jLopezbarb:

# Proposed changes

Fixes # (issue)

- Use the specific syntax that docker uses for allowing multiline env vars or args.
- Add to all the e2e tests an env var that is multiline in order to test that it doesn't break

## How to validate

1. With the following okteto.yml
```yaml
deploy:
  - echo $OKTETO_MULTILINE
```
2. Run `export OKTETO_MULTILINE="This is splitted in
two lines"`
3. Run `okteto deploy` and check that is working

## CLI Quality Reminders 🔧

### DIFF
--- a/cmd/registrytoken/install.go
+++ b/cmd/registrytoken/install.go
@@ -24,6 +24,7 @@ import (
 
 type InstallOptions struct {
 	Overwrite bool
+	filename  string
 }
 
 func Install() *cobra.Command {
@@ -49,6 +50,7 @@ func Install() *cobra.Command {
 			}
 
 			conf.CredentialsStore = "okteto"
+			conf.Filename = options.filename
 
 			if err := conf.Save(); err != nil {
 				return errors.Wrapf(err, "couldn't save docker config file at %q", confDir)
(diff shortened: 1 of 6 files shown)


pkg/remote/remote.go
```diff
@@ -85,11 +85,11 @@ COPY . /okteto/src
 WORKDIR /okteto/src
 
 {{range $key, $val := .OktetoBuildEnvVars }}
-ENV {{$key}} {{$val}}
+ENV {{$key}}={{$val}}
```
ifbyol, line 88: surrounding the value between `"` is not enough to specify the multi-line value?

jLopezbarb: No, If you surround with "" but don't set the equal everything is in just one line

ifbyol: I'm meaning leaving the equal and using the quotes. Do you still need to scape the new line char within `formatEnvVarValueForDocker`?

jLopezbarb: Yes, there is no way of doing it in separate lines in the dockerfile

cmd/registrytoken/install.go
```diff
@@ -34,6 +34,7 @@ func Install() *cobra.Command {
 		RunE: func(cmd *cobra.Command, args []string) error {
 			confDir := config.Dir()
 			conf, err := config.Load(confDir)
+			oktetoLog.SetStage("install registry credential helper")
```
ifbyol, line 37: Why this? Wouldn't this make that there is a new stage in the UI logs?

ifbyol (approved):

PR #4454, opened 2024-09-12
Fix typos and remove dead code

andreafalzetti:
Small PR to fix some typos and remove some dead code


pkg/analytics/track.go
```diff
@@ -115,11 +108,6 @@ func TrackPreviewDestroy(success bool) {
 	track(previewDestroyEvent, success, nil)
 }
 
-// TrackManifestHasChanged sends a tracking event to mixpanel when the okteto up command fails because manifest has changed
-func TrackManifestHasChanged(success bool) {
```
andreafalzetti, line 119: I find it odd that we have this unused funcs, but I've checked old versions and it's the same

ifbyol (approved):

PR #4458, opened 2024-09-17
Support okteto volume access mode, volume mode, labels and annotations

pchico83:
# Proposed changes

By adding support to labels, annotations, access mode and volume node, users have more flexibility to customize the creation of the okteto persistent volume.
For example, this allows the creation of EFS volumes to be shared across nodes when using [services](https://www.okteto.com/docs/reference/okteto-manifest/#services-object-optional).

### DIFF
--- a/pkg/k8s/apps/translate.go
+++ b/pkg/k8s/apps/translate.go
@@ -24,7 +24,6 @@ import (
 	"github.com/okteto/okteto/pkg/model"
 	apiv1 "k8s.io/api/core/v1"
 	resource "k8s.io/apimachinery/pkg/api/resource"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/utils/pointer"
 )
 
@@ -100,7 +99,6 @@ func (tr *Translation) translate() error {
 		}
 
 		tr.DevApp.TemplateObjectMeta().Labels[model.DetachedDevLabel] = tr.getDevName()
-		TranslatePodAffinity(tr.DevApp.PodSpec(), tr.MainDev.Name)
 	}
 
 	tr.DevApp.PodSpec().TerminationGracePeriodSeconds = pointer.Int64(0)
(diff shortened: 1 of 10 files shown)


pkg/k8s/apps/translate.go
```diff
@@ -160,30 +158,6 @@ func TranslateDevTolerations(spec *apiv1.PodSpec, tolerations []apiv1.Toleration
 	spec.Tolerations = append(spec.Tolerations, tolerations...)
 }
 
-// TranslatePodAffinity translates the affinity of pod to be all on the same node
-func TranslatePodAffinity(spec *apiv1.PodSpec, name string) {
```
pchico83, line 164: I moved this logic to `ToTranslationRule`. Translations rules already supports affinities. We can inject the affinities directly in the translation rule. This way we have more context to make this decision (for example, we have access to all the PersistentVolume fields).

pkg/k8s/volumes/crud.go
```diff
@@ -155,6 +155,25 @@ func checkPVCValues(pvc *apiv1.PersistentVolumeClaim, dev *model.Dev, devPath st
 			)
 		}
 	}
+
+	if len(pvc.Spec.AccessModes) == 1 && dev.PersistentVolumeAccessMode() != pvc.Spec.AccessModes[0] {
```
pchico83, line 159: This force a failure if the user modifies the access modes or volume mode of an existing PVC

pkg/model/dev.go
```diff
@@ -892,11 +921,17 @@ func (dev *Dev) ToTranslationRule(main *Dev, namespace, username string, reset b
 			}
 			rule.Args = append(rule.Args, "-s", fmt.Sprintf("%s:%s", filename, s.RemotePath))
 		}
-	} else if len(dev.Args.Values) > 0 {
-		rule.Args = dev.Args.Values
-	} else if len(dev.Command.Values) > 0 {
-		rule.Command = dev.Command.Values
-		rule.Args = []string{}
+	} else {
+		if main.PersistentVolumeAccessMode() != apiv1.ReadWriteMany {
```
pchico83, line 925: If the access mode supports `ReadWriteMany`, we don't need to force required affinities to deploy all pods in the same node

PR #4461, opened 2024-09-18
Add CI env var to okteto test

jLopezbarb:
# Proposed changes

Fixes DEV-510

This PR passes the env vars as part of the comand instead of passing just a string. This will allow to add more env vars without having to create new params for the commands.
This PR adds a new CI variable to okteto test command in order to make it work the same way as in any other CI provider

## How to validate

1. With the following okteto manifest:
```yaml
deploy:
  - printenv
test:
  unit:
    - printenv
```
1. Run okteto test
1. See that there is a CI environment set to true

## CLI Quality Reminders 🔧


teresaromero (approved):

cmd/test/cmd.go
```diff
@@ -348,6 +348,15 @@ func doRun(ctx context.Context, servicesToTest []string, options *Options, ioCtr
 			DependenciesEnvVars: deployCMD.GetDependencyEnvVars(os.Environ),
 			ExecutionEnvVars:    deployable.GetPlatformEnvironment(ctx),
 			DockerfileName:      "Dockerfile.test",
+			OktetoCommandSpecificEnvVars: map[string]string{
+				constants.OktetoIsPreviewEnvVar: os.Getenv(constants.OktetoIsPreviewEnvVar),
+				constants.CIEnvVar: func() string {
+					if val, ok := os.LookupEnv(constants.CIEnvVar); ok {
+						return val
+					}
+					return "true"
```
ifbyol, line 357: I'm not sure we should be defaulting to `true` if it is not present. We would be mainly changing the purpose of the `CI` env var.

@viddo as the ticket reported, what would be the benefit of defaulting to true?

I don't have strong opinion, but I find it odd that we consider always that `okteto test` is CI

viddo: My bad, yeah agree with you @ifbyol! It makes more sense that it's set to _false_ by default when missing.
```suggestion
					return "false"
```

andreafalzetti: I also think it should be default to `false`, it would be really odd to have `CI` true on something I run locally.

andreafalzetti (approved):

cmd/test/cmd.go
jLopezbarb: It depends. I think if we want to have the same behaviour always it should be true so we can have the same variables in local (with okteto test) CI (with and without okteto test)

pchico83: I think defaulting to `true` is better too. `okteto test` value proposition is that tests run exactly the same on local and on CI. What are the drawbacks of defaulting to tue?

ifbyol: > we can have the same variables in local (with okteto test) CI (with and without okteto test)

> run exactly the same on local and on CI

As I see it, the runtime environment of the tests is composed by: 
* Underlying OS + tooling (Linux/windows, yarn/go/node/junit, etc...)
* Environment variables

Both local and CI can never be the same in terms of environment variables, unless they are forced by the user. Each CI provider inject their own env variables that might be needed for the tests. Each `GITHUB_` or `CIRCLE_` (without taking into account project or organization env vars) env var would be also missing when running those tests from local.

As I see it, the real value is in having the same tooling and versions when running from both places.

> What are the drawbacks of defaulting to tue?

IMHO, Injecting just `CI` environment might be misleading, because we might be missing other env vars injected during those workflows, generating some odd behavior or changing a bit the execution. For me, adding a subset of env vars trying to simulate an scenario that it is not actually happening, is worst, because it would be kind of a mix.

If I'm running something outside a CI flow, I wouldn't expected to have something set as CI. I would expect to have to set it myself manually. It is related to what Nicklas mentioned in the issue related to the [Principle of least astonishment](https://en.wikipedia.org/wiki/Principle_of_least_astonishment).

Anyways, I'm not sure 100% sure and that is why I started the conversation

pchico83: @ifbyol could you clarify `Both local and CI can never be the same in terms of environment variables`? My understanding is that this is not true, we don't automatically inject CI environment variables in test containers.

ifbyol: We don't do it automatically, but I think they can be injected using `.env` file for example if your tests need them for whatever reason (for example, to know the PR number to compose the endpoints to be used by the e2e tests), right?

If you tests depend on those env vars, and you have configured the repo to make it work in circleci + okteto. When you run them locally, you will never have those env vars unless you explicitly set them. `CI` would be the exception.

There is the exception of GitHub action of Okteto test, where the action will inject automatically those env vars (`GITHUB_`), so running the GitHub Action of Okteto test wouldn't be the same than running just `okteto test`. Why? Because the environment variables provided from local or GitHub actions change

pchico83: We should go in the direction of eliminating those differences between local and remote, but I see your point. It's very hard to do this right while we depend on external CI providers.
We only have one user data point at the moment, and they were forcing `CI=true` on local to simulate CI better. They were very confused about it and lost a lot of time until they realized it was because the `CI` envvar.
I know we aren't there, but I think the fewer differences we can have between `okteto test` running locally or on remote the better. `okteto test` is not to run tests locally, it's to test CI locally. I want to re-inforce this idea in the team.

ifbyol: I still have doubts about it, but lets set the default to `true` then

viddo: [Type 2 decision](https://www.theuncertaintyproject.org/tools/decision-types), isn't it? At least before v3.0 is finally released, I figure.

viddo: BTW/aside, @pchico83 
> Running okteto test locally is not to run tests locally, it's to test CI locally. I want to re-inforce this idea in the team.

IMHO, that's an important nuance, ought to be stated & emphasized up & center on https://www.okteto.com/docs/testing/ shouldn't it?

pchico83: @viddo 100%. In general, we need to give some love to the Okteto Test docs. We have low capacity at the moment, but once we onboard our first users, we have to give a general pass to the docs

PR #4462, opened 2024-09-18
Add X-Okteto-CLI-Version version to requests

maroshii:
There seem to be a lot of different http clients that are being used. This seems to cover all the cases but it needs review


ifbyol (approved):
👏

andreafalzetti (approved):

PR #4470, opened 2024-09-23
Refactor authProvider to prioritize local registry credentials

jLopezbarb:

# Proposed changes

Fixes DEV-680

- Rename variable from OKTETO_LOCAL_REGISTRY_STORE_ENABLED to OKTETO_LOCAL_REGISTRY_STORE_PRIORITY_ENABLED
- Now we are going to retrieve always local and okteto credentials but will prioritise okteto one if the env var is unset or false

## How to validate

1. Add a registry credentials that you already have in local from the UI
1. Run `okteto build -t registry/repo/image:latest` and check that you are using the one from okteto
1. Set OKTETO_LOCAL_REGISTRY_STORE_PRIORITY_ENABLED to true
1. Run again `okteto build -t registry/repo/image:latest` and check that you are using the one from local

## CLI Quality Reminders 🔧

### DIFF
--- a/pkg/cmd/build/authprovider.go
+++ b/pkg/cmd/build/authprovider.go
@@ -148,22 +148,31 @@ func (ap *authProvider) Credentials(ctx context.Context, req *auth.CredentialsRe
 		return nil, err
 	}
 
-	retrieveFromLocalFirst := env.LoadBooleanOrDefault(oktetoLocalRegistryStorePriorityEnabledEnvVarKey, false)
 	oktetoCredentials := ap.getOktetoCredentials(originalHost, c)
+	// Check if we need to retrieve local credentials
+	retrieveFromLocalFirst := env.LoadBooleanOrDefault(oktetoLocalRegistryStorePriorityEnabledEnvVarKey, false)
+
+	// If Okteto credentials are valid and we're not prioritizing local, return Okteto credentials
+	if ap.areCredentialsValid(oktetoCredentials) && !retrieveFromLocalFirst {
+		return oktetoCredentials, nil
+	}
+
+	// Fetch local credentials only if Okteto credentials are invalid or we are prioritizing local
 	localCredentials, err := ap.getLocalCredentials(req)
 	if err != nil {
 		return nil, err
 	}
 
-	areCredentialsValid := func(creds *auth.CredentialsResponse) bool {
-		return creds != nil && creds.Username != "" && creds.Secret != ""
-	}
-
-	// Return local credentials if prioritizing local and valid, otherwise Okteto credentials if valid, fallback to local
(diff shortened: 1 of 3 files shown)


pkg/registry/client.go
```diff
@@ -209,7 +209,7 @@ func (c client) getAuthentication(ref name.Reference) remote.Option {
 		authn.DefaultKeychain,
 		authn.NewKeychainFromHelper(inlineHelper(c.config.GetExternalRegistryCredentials)),
 	)
-	if !env.LoadBooleanOrDefault(oktetoLocalRegistryStoreEnabledEnvVarKey, false) {
+	if !env.LoadBooleanOrDefault(oktetoLocalRegistryStorePriorityEnabledEnvVarKey, false) {
```
ifbyol, line 212: Wasn't this flag supposed to not take into account local credentials? Did you have a bug in latest versions of CLI?

pkg/cmd/build/authprovider.go
```diff
@@ -148,36 +148,47 @@ func (ap *authProvider) Credentials(ctx context.Context, req *auth.CredentialsRe
...
 
-	retrieveFromLocal := env.LoadBooleanOrDefault(oktetoLocalRegistryStoreEnabledEnvVarKey, false)
-	if !retrieveFromLocal {
-		return credentials, nil
+	areCredentialsValid := func(creds *auth.CredentialsResponse) bool {
+		return creds != nil && creds.Username != "" && creds.Secret != ""
 	}
 
+	// Return local credentials if prioritizing local and valid, otherwise Okteto credentials if valid, fallback to local
+	if (retrieveFromLocalFirst && areCredentialsValid(localCredentials)) || !areCredentialsValid(oktetoCredentials) {
```
ifbyol, line 163: Why do we check now if `oktetoCredentials` are valid? Before we were not checking it and we were defaulting to the Okteto ones

jLopezbarb:
> Are we restoring the current behavior as of today that was removed in 3.0? Asking because if we are not this is a breaking change

It's not the same behavior, it's a breaking change, we were not even contemplating the local credentials even if they existed. With this change we get them but we only give more priority to the okteto one

pkg/registry/client.go
jLopezbarb: We implemented it to not take in count the local credentials but during testing we decided to contemplate the local env vars and only change the priority based on the env var

pkg/cmd/build/authprovider.go
jLopezbarb: We were not checking it because we were checking if the localCredentials were valid and if not we returned the oktetoCredentials. We now have to give priority and need to check both of them

PR #4473, opened 2024-09-23
Add specific error for compose on cli only 

jLopezbarb:
# Proposed changes

Fixes DEV-684

Adds an specific error when using a compose file with a CLI-only cluster

## How to validate
1. Using minikube or any other CLI-only cluster
1. Run okteto up with a compose file
1. Run okteto build with a compose file

## CLI Quality Reminders 🔧

### DIFF
--- a/cmd/build/command.go
+++ b/cmd/build/command.go
@@ -181,9 +181,9 @@ func (bc *Command) getBuilder(options *types.BuildOptions, okCtx *okteto.Context
 				bc.analyticsTracker.TrackImageBuild,
 				bc.insights.TrackImageBuild,
 			}
-			if manifest.Type == model.StackType {
+			if !okCtx.IsOktetoCluster() && manifest.Type == model.StackType {
 				return nil, oktetoErrors.UserError{
-					E: fmt.Errorf("Docker Compose format is only available using the Okteto Platform"),
+					E: fmt.Errorf("docker Compose format is only available using the Okteto Platform"),
 					Hint: `Follow this link to install the Okteto Platform in your Kubernetes cluster:
     https://www.okteto.com/docs/get-started/install`,
 				}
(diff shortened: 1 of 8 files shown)


cmd/exec/cmd.go
```diff
@@ -120,6 +121,13 @@ okteto exec api -- bash`,
 			}
 
 			if !okteto.IsOkteto() {
+				if manifest.Type == model.StackType {
+					return oktetoErrors.UserError{
+						E: fmt.Errorf("docker Compose format is only available using the Okteto Platform"),
+						Hint: `Follow this link to install the Okteto Platform in your Kubernetes cluster:
+    https://www.okteto.com/docs/get-started/install`,
+					}
+				}
```
ifbyol, line 130: Wouldn't this check fit better within `ValidateForCLIOnly` function? I mean, as far as I know, docker-compose is not supported on CLI-only scenarios, independently of the command, so I think it makes sense to include it as part of that validation instead of spreading around the code different checks for the same case.

Also, in all the places where we call the `ValidateForCLIOnly`, we are adding this new check, and the docker-compose file is a manifest, so it fits well to be added there

cmd/build/command.go
```diff
@@ -181,6 +181,13 @@ func (bc *Command) getBuilder(options *types.BuildOptions, okCtx *okteto.Context
 				bc.analyticsTracker.TrackImageBuild,
 				bc.insights.TrackImageBuild,
 			}
+			if !okCtx.IsOktetoCluster() && manifest.Type == model.StackType {
+				return nil, oktetoErrors.UserError{
+					E: fmt.Errorf("docker Compose format is only available using the Okteto Platform"),
+					Hint: `Follow this link to install the Okteto Platform in your Kubernetes cluster:
+    https://www.okteto.com/docs/get-started/install`,
+				}
+			}
```
ifbyol, line 190: Build command is not supported for non-okteto contexts, so we don't arrive to this point, so we don't show this message. This is what I get with this CLI version

```
ok build -f docker-compose.yml
 i  Using okteto @ xxxxxxxx as context
 x  This command is only available installing Okteto in your cluster
    Visit our docs to learn more about the Okteto Platform:
    https://www.okteto.com/docs
```

PR #4478, opened 2024-09-24
Update func comment `HasDeployedSomething`

andreafalzetti:
Small fix to adjust the func comment because `Jobs` are not part of the logic.


jLopezbarb (approved):

jLopezbarb:
Should we check if the pipeline has deployed a job instead?

ifbyol (approved):

andreafalzetti:
> Should we check if the pipeline has deployed a job instead?

I think it's unpredictable how long the Job can take, and it would need product discussion, merging this improvement for now

PR #4487, opened 2024-09-26
Refactor getStackServicesToDeploy function to handle multiple Compose…

jLopezbarb:
…Info objects

The getStackServicesToDeploy function in command_configuration.go has been refactored to handle multiple ComposeInfo objects. Previously, it only processed the first ComposeInfo object and ignored the rest. Now, it correctly appends the services to deploy from all ComposeInfo objects and validates the defined services accordingly.

This change ensures that all services specified in the ComposeInfo objects are included in the deployment process.

Fixes DEV-692

# Proposed changes

When we removed the services to deploy from flags we updated the services toDeploy from the compose file but were not checking one of the conditions if run from a okteto manifest

## How to validate

1. With the following `okteto.yml`
```yaml
deploy:
  compose: compose.yml
```
2. Deploy using `okteto deploy`
3. Check that there is no error

## CLI Quality Reminders 🔧

### DIFF
--- a/cmd/deploy/command_configuration.go
+++ b/cmd/deploy/command_configuration.go
@@ -93,10 +93,18 @@ func setDeployOptionsValuesFromManifest(ctx context.Context, deployOptions *Opti
 func getStackServicesToDeploy(ctx context.Context, composeSectionInfo *model.ComposeSectionInfo, c kubernetes.Interface) ([]string, error) {
 	svcs := []string{}
 
-	for _, composeInfo := range composeSectionInfo.ComposesInfo {
-		svcs = append(svcs, composeInfo.ServicesToDeploy...)
-	}
 	if len(composeSectionInfo.ComposesInfo) > 0 {
+		servicesToDeploy := 0
+		for _, composeInfo := range composeSectionInfo.ComposesInfo {
+			svcs = append(svcs, composeInfo.ServicesToDeploy...)
+			servicesToDeploy += len(composeInfo.ServicesToDeploy)
+		}
+		if servicesToDeploy == 0 {
+			for service := range composeSectionInfo.Stack.Services {
+				svcs = append(svcs, service)
+			}
+			return svcs, nil
+		}
 		if err := stack.ValidateDefinedServices(composeSectionInfo.Stack, svcs); err != nil {
 			return []string{}, err
 		}
--- a/cmd/deploy/command_configuration.go
+++ b/cmd/deploy/command_configuration.go
@@ -15,7 +15,6 @@ package deploy
 
 import (
 	"context"
-	"fmt"
 	"net/url"
 	"os"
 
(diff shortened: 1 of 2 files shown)


cmd/deploy/command_configuration.go
```diff
@@ -93,10 +93,18 @@ func setDeployOptionsValuesFromManifest(ctx context.Context, deployOptions *Opti
 func getStackServicesToDeploy(ctx context.Context, composeSectionInfo *model.ComposeSectionInfo, c kubernetes.Interface) ([]string, error) {
 	svcs := []string{}
 
-	for _, composeInfo := range composeSectionInfo.ComposesInfo {
-		svcs = append(svcs, composeInfo.ServicesToDeploy...)
-	}
 	if len(composeSectionInfo.ComposesInfo) > 0 {
```
teresaromero, line 96: why dont we loop `composeSectionInfo.ComposesInfo` ?

cmd/deploy/command_configuration.go
```diff
@@ -93,10 +93,18 @@ func setDeployOptionsValuesFromManifest(ctx context.Context, deployOptions *Opti
...
+		servicesToDeploy := 0
+		for _, composeInfo := range composeSectionInfo.ComposesInfo {
+			svcs = append(svcs, composeInfo.ServicesToDeploy...)
+			servicesToDeploy += len(composeInfo.ServicesToDeploy)
+		}
+		if servicesToDeploy == 0 {
+			for service := range composeSectionInfo.Stack.Services {
```
teresaromero, line 103: can Stack be nil?

cmd/deploy/command_configuration_test.go
```diff
@@ -207,6 +207,18 @@ func Test_getStackServicesToDeploy(t *testing.T) {
 			},
 			expected: []string{},
 		},
+		{
+			name: "MultipleComposeInfo",
```
ifbyol, line 211: There are 3 test cases with the same title. Could we rename them to make them more specific? What is the difference between each of them?

ifbyol (commented):
@jLopezbarb What do we mean with `Previously, it only processed the first ComposeInfo object and ignored the rest.`?

Before it wasn't even considering a single compose file/service

cmd/deploy/command_configuration.go
jLopezbarb: Because if it has the section [with services](https://www.okteto.com/docs/reference/okteto-manifest/#deploy-with-compose) we need to only deploy those, but if it's empty we need to deploy all (previous used by args)

cmd/deploy/command_configuration.go
jLopezbarb: Not here but I'll add the check

jLopezbarb:
> @jLopezbarb What do we mean with `Previously, it only processed the first ComposeInfo object and ignored the rest.`?
> 
> Before it wasn't even considering a single compose file/service

I mean that we were checking the services to Deploy from the first composeInfo but if those have several services to deploy we were not doing it right. In fact I'm thinking there is still an error when there is a compose Info with and without services. I'll make the changes and come back to you

jLopezbarb:
> > @jLopezbarb What do we mean with `Previously, it only processed the first ComposeInfo object and ignored the rest.`?
> > Before it wasn't even considering a single compose file/service
> 
> I mean that we were checking the services to Deploy from the first composeInfo but if those have several services to deploy we were not doing it right. In fact I'm thinking there is still an error when there is a compose Info with and without services. I'll make the changes and come back to you

Checked and it's now working as expected

ifbyol (approved):

PR #4494, opened 2024-09-27
DEV-690: Use right build context path to get the right diff and sha in git operations for Smart Builds

ifbyol:
# Proposed changes

Fixes DEV-690

When images specify a parent folder as context, or if the command is executed from a parent folder and the build context a is grand-child folder, the calculation of the Smart Builds context wasn't being done correctly, because the path specified to git commands wasn't right.

Apart from that, in some cases, the build context is a relative path (when the build information is specified in an Okteto manifest), and other times it is an absolute path (when the build information comes from a compose file), so it was confusing the path passed to git commands.

With these changes, we now send always to git a local absolute path to the folder containing the image build context:
* If the build context is already an absolute path, we just pass that value to the git command, as it was already calculated properly
* If the build context is a relative path, we just join the current working directory + the build context. As the build context was already calculated based on the working directory, joining both of them, we have the right absolute path, so git always works with absolute routes to folders.

I added also a few debug commands, so we can see in the traces which commands are executed with git to calculate the build context.

## How to validate

Using the repository https://github.com/ifbyol/test-okteto and the branch `ifbyol/test-smart-build-contexts`, you could test all the different scenarios (each folder is an scenario to test). Without my changes, you can see how some of those cases don't work as expected, and after my changes, smart builds works as expected.

To validate it properly, it should be executed from the root of the repo, but also from other paths, to make sure we always specify the right path to git to do the Smart Build calculations

## CLI Quality Reminders 🔧

### DIFF
--- a/cmd/build/v2/smartbuild/hasher_test.go
+++ b/cmd/build/v2/smartbuild/hasher_test.go
@@ -14,228 +14,228 @@
 package smartbuild
 
 import (
-    "errors"
-    "path/filepath"
-    "testing"
+	"errors"
+	"path/filepath"
+	"testing"
 
-    "github.com/okteto/okteto/pkg/build"
-    "github.com/spf13/afero"
-    "github.com/stretchr/testify/assert"
-    "github.com/stretchr/testify/mock"
+	"github.com/okteto/okteto/pkg/build"
+	"github.com/spf13/afero"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/mock"
 )
 
 type fakeWorkingDirGetter struct {
-    err        error
-    workingDir string
+	err        error
+	workingDir string
 }
 
 func (f fakeWorkingDirGetter) Get() (string, error) {
-    if f.err != nil {
-        return "", f.err
-    }
-    return f.workingDir, nil
+	if f.err != nil {
+		return "", f.err
+	}
+	return f.workingDir, nil
 }
 
 type mockConfigRepo struct {
-    mock.Mock
+	mock.Mock
 }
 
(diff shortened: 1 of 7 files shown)


cmd/build/v2/smartbuild/hasher_test.go
```diff
@@ -20,8 +20,38 @@ import (
...
+func (f fakeWorkingDirGetter) Get() (string, error) {
+	if f.err != nil {
+		return "", f.err
+	}
+	return f.workingDir, nil
+}
+
+type mockConfigRepo struct {
+	mock.Mock
+}
```
ifbyol, line 40: Added this mock to be able to do assertions based on the build context passed to the repoCtrl, and verify we are passing the expected values

pkg/repository/local_git.go
```diff
@@ -47,6 +47,7 @@ type LocalExec struct{}
 
 func (le *LocalExec) RunCommand(ctx context.Context, dir string, name string, arg ...string) ([]byte, error) {
 	c := le.createCommand(ctx, dir, name, arg...)
+	oktetoLog.Debugf("executing command: %s", strings.Join(c.Args, " "))
```
ifbyol, line 50: As mention in the description, these are some debug logs that might help us to troubleshoot issue

cmd/build/v2/smartbuild/hasher_test.go
teresaromero: 😮 i did not knew thisss

teresaromero (approved):

jLopezbarb (approved):

PR #4500, opened 2024-09-30
Fix okteto status: namespace load for synthing

teresaromero:
# Proposed changes

Fixes https://okteto.atlassian.net/browse/DEV-707

Discovered while testing `okteto status` was not reporting devmode status.

- Removed unused Namespace and Context from manifest Options (keys from struct and assignations) The values are not read
- Load of synthing info was using `namespace` which was only referencing the flag value, but not the context value.

Tested:

overriden flag value with non running
❯ ok status --namespace teresaromero-patata
 i  Using teresaromero-patata @ - as context
 x  Development mode isn't enabled on your deployment
    Run 'okteto up' to enable it and try again

overriden flag value with current namespace
❯ ok status --namespace teresaromero
 i  Using teresaromero @ - as context
 ✓  Synchronization status: 100.00%
 
current namespace
 ❯ ok status
 i  Using teresaromero @ - as context
 ✓  Synchronization status: 100.00%

## CLI Quality Reminders 🔧

### DIFF
--- a/cmd/status.go
+++ b/cmd/status.go
@@ -73,7 +73,7 @@ func Status(fs afero.Fs) *cobra.Command {
 				return err
 			}
 
-			manifestOpts := contextCMD.ManifestOptions{Filename: devPath, Namespace: namespace, K8sContext: k8sContext}
+			manifestOpts := contextCMD.ManifestOptions{Filename: devPath}
 			manifest, err := model.GetManifestV2(manifestOpts.Filename, afero.NewOsFs())
 			if err != nil {
 				return err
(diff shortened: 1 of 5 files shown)


cmd/status.go
```diff
@@ -73,7 +73,7 @@ func Status(fs afero.Fs) *cobra.Command {
 				return err
 			}
 
-			manifestOpts := contextCMD.ManifestOptions{Filename: devPath, Namespace: namespace, K8sContext: k8sContext}
+			manifestOpts := contextCMD.ManifestOptions{Filename: devPath}
```
jLopezbarb, line 76: If we don't use it anymore. Should we remove it completely from the `ManifestOptions`?

teresaromero: As talked offline, removed from ManifestOptions Namespace and K8sContext as they are not used

jLopezbarb (approved):

PR #4511, opened 2024-10-01
move name inferrer inside destroyCommand.destroy

teresaromero:
# Proposed changes

Fixes https://okteto.atlassian.net/browse/DEV-706

Debugged the issue and found that the `destroy` command was not taking into account the manifest.name option when is run inside a deploy manifest.

The name inferrer was setting the name as the repo name when the configmap was not found, as a fallback. After this was set, when the manifest was read, the name inside the manifest was not taken into account.

Proposed changes:
- create a buildControlInterface to be able to use the correct name when providing the build controller
- decide the ops.Name in the same place once we have the manifest to have all params taken into account

## How to validate

Build the binary and deploy a dev instance. Reproduce the steps as mentioned on the ticket. 
Fork with changes without the installation of 2.31 https://github.com/teresaromero/okteto-test 

## CLI Quality Reminders 🔧

### DIFF
--- a/cmd/destroy/destroy.go
+++ b/cmd/destroy/destroy.go
@@ -114,6 +114,10 @@ type buildTrackerInterface interface {
 	TrackImageBuild(context.Context, *analytics.ImageBuildMetadata)
 }
 
+type buildControlProviderInterface interface {
+	provide(name string) buildCtrl
+}
+
 type destroyCommand struct {
 	executor             executor.ManifestExecutor
 	nsDestroyer          destroyer
@@ -126,7 +130,7 @@ type destroyCommand struct {
 	ioCtrl               *io.Controller
 	getDivertDriver      divertProvider
 	getPipelineDestroyer pipelineDestroyerProvider
-	buildCtrl            buildCtrl
+	buildCtrlProvider    buildControlProviderInterface
 }
 
 // Destroy destroys the dev application defined by the manifest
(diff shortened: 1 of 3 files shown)


ifbyol (approved):
For the future, we will need to unify all this logic for all the commands. It should be identical in most of the cases, but we have duplicated a lot of this code and it generates this kind oof issues (a command doesn't work as the others)

I was testing it and it works fine. My comment inline is more about double-checking that moving the logic to calculate the proper name doesn't affect to things that are in the middle of where it was before and it is now

cmd/destroy/destroy.go
```diff
@@ -364,15 +353,36 @@ func (dc *destroyCommand) destroy(ctx context.Context, opts *Options) error {
 	}
 
 	opts.Manifest = manifest
+
+	// name for cfg map being destroyed
+	// name set by flag has priority over manifest name
+	// if no name is set, the name of the manifest is used
+	// if no name is set in the manifest, the name is inferred
+	if opts.Name == "" && opts.Manifest.Name != "" {
```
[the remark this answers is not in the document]
teresaromero: you are right. i did not see the kubeconfig read of the name... the vision i have is to unify where the name and manifest.name are set so the read operations are done after...

i've updated with a commit to move this before reading the name for anything... this involves reading the manifest earlier in the destroy cmd. I've been testing it and i dont see any issue now. With the previous change the name could be not inferred.

cmd/destroy/destroy.go
```diff
@@ -189,23 +193,6 @@ If you need to destroy external resources (like s3 buckets or other Cloud resour
 				return oktetoErrors.ErrContextIsNotOktetoCluster
 			}
 
-			cwd, err := os.Getwd()
-			if err != nil {
-				return fmt.Errorf("failed to get the current working directory: %w", err)
-			}
```
ifbyol, line 195: You don't have to remove this. In line 172, there might be a change in the working directory of the command, so `initialWd` might not be the correct one you have to pass to `setOptionsNameAndManifestName`

andreafalzetti (approved):

cmd/destroy/destroy.go
teresaromero: reverted this change

PR #4513, opened 2024-10-01
Support for "priorityClassName" in dev containers

pchico83:
Priority classes are very useful when using [services](https://www.okteto.com/docs/reference/okteto-manifest/#services-object-optional). By giving more priority to the pods created by `okteto up`, users can enforce that all pods will go to the same node, even if it's required to evict other pods

### DIFF
--- a/pkg/model/dev.go
+++ b/pkg/model/dev.go
@@ -68,6 +68,7 @@ type Dev struct {
 	Name                 string                `json:"name,omitempty" yaml:"name,omitempty"`
 	Container            string                `json:"container,omitempty" yaml:"container,omitempty"`
 	ServiceAccount       string                `json:"serviceAccount,omitempty" yaml:"serviceAccount,omitempty"`
+	PriorityClassName    string                `json:"priorityClassName,omitempty" yaml:"priorityClassName,omitempty"`
 	parentSyncFolder     string
 	Interface            string           `json:"interface,omitempty" yaml:"interface,omitempty"`
 	Mode                 string           `json:"mode,omitempty" yaml:"mode,omitempty"`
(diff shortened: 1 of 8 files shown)


pkg/model/dev.go
```diff
@@ -816,21 +817,22 @@ func TranslatePodAffinity(tr *TranslationRule, name string) {
...
+		Container:         dev.Container,
+		ImagePullPolicy:   dev.ImagePullPolicy,
+		Environment:       dev.Environment,
+		Secrets:           dev.Secrets,
+		WorkDir:           dev.Workdir,
+		PersistentVolume:  main.PersistentVolumeEnabled(),
+		Volumes:           []VolumeMount{},
+		SecurityContext:   dev.SecurityContext,
+		ServiceAccount:    dev.ServiceAccount,
+		PriorityClassName: main.PriorityClassName,
```
ifbyol, line 829: Why this would be the only property to be read from `main` instead of doing it from `dev`? What is the difference?

ifbyol (approved):
A comment asking a doubt, but in general looks good!

pkg/model/dev.go
pchico83: The idea is to define the `PriorityClassName` only for the main container, instead of adding it to every pod in the `services` section. We could instruct the user to define it on every `service`, but I think it adds more friction to the user. Both would work anyway. 
If we go with defining it for every service, something that we are considering is support for okteto manifest templates, and that might help too

ifbyol: ah, I see, that's the difference. I think it makes sense

jLopezbarb (approved):

PR #4528, opened 2024-10-08
feat(build): if buildkit is not available wait for it

jLopezbarb:
# Proposed changes

Fixes partly DEV-679

- Adds metadata from buildkit: Right now we are only sending the buildkit waiting for available time but the idea is that the buildTime that we are getting right now we get it from buildkit itself instead of calculating a rough estimation.
- Refactor buildkit client creation: We had that code in a folder that we don't want to continue having there, therefore I moved the buildkit client creation to its own package under `pkg/build/buildkit`. The idea is to continue moving code there adding tests to it during the refactors.
- Create a new function to wait for buildkit. If buildkit is not available we create a client and check that the information is available. If it's not available after x time we return a proper error to the user. Probably we need to update the copy of the message. Right now it looks like this:
Screenshot 2024-10-08 at 18 48 57

- Added the environment variables `OKTETO_BUILDKIT_WAIT_TIMEOUT` and `OKTETO_RETRY_BUILDKIT_TIME` in order to configure the parameters. (Needs documentation when we decide the naming convention)

### Check that the spinner works

1. Run `kubectl scale statefulset yourBuildkitStatefulset --replicas=0`
1. Run `okteto build --no-cache` on any repo with build available
1. See that the spinner is running:
Screenshot 2024-10-08 at 18 30 53

### Check that the environment variables work

1. Run `kubectl scale statefulset yourBuildkitStatefulset --replicas=0`
1. Run `export  OKTETO_BUILDKIT_WAIT_TIMEOUT=15m`
1. Run `export OKTETO_RETRY_BUILDKIT_TIME=50s`
1. Run `okteto build --no-cache --log-level=debug` on any repo with build available
1. Check that the buildkit waiting time stops when reached 30s and there are updates every second

### Check that it recovers correctly
1. Run `kubectl scale statefulset yourBuildkitStatefulset --replicas=0`
1. Run `okteto build --no-cache` on any repo with build available
1. See that the spinner is running
1. Run `kubectl scale statefulset yourBuildkitStatefulset --replicas=1 ` on another terminal
1. See that the image starts building 

## CLI Quality Reminders 🔧


pkg/build/buildkit/wait.go
```diff
@@ -0,0 +1,135 @@
...
+	sp := bw.Logger.Out().Spinner("Waiting for BuildKit server to be ready...")
+	sp.Start()
+	defer sp.Stop()
+
+	startWaitingTime := time.Now()
+	for {
+		select {
+		case <-ctx.Done():
+			bw.waitingTime = bw.MaxWaitTime
+			return fmt.Errorf("timeout reached: BuildKit server is not responding after %v", bw.MaxWaitTime)
```
jLopezbarb, line 112:
```suggestion
			return fmt.Errorf("buildkit not available for %v", bw.MaxWaitTime)
```

pkg/build/buildkit/wait.go
```diff
@@ -0,0 +1,135 @@
...
+	startWaitingTime := time.Now()
+	for {
+		select {
+		case <-ctx.Done():
+			bw.waitingTime = bw.MaxWaitTime
+			return fmt.Errorf("buildkit not available for %v", bw.MaxWaitTime)
+		default:
+			c, err := bw.buildkitClientFactory.GetBuildkitClient(ctx)
+			if err != nil {
+				bw.Logger.Infof("Failed to connect to BuildKit server: %v\n", err)
```
pchico83, line 116:
```suggestion
				bw.Logger.Infof("Failed to connect to Okteto Build service: %v\n", err)
```

pkg/build/buildkit/wait.go
```diff
@@ -0,0 +1,135 @@
...
+				bw.Logger.Infof("Retrying in %v...\n", bw.RetryInterval)
+				bw.Sleeper.Sleep(bw.RetryInterval)
+				continue
+			}
+			_, err = c.Info(ctx)
+			if err != nil {
+				bw.Logger.Infof("Failed to get BuildKit server info: %v\n", err)
```
pchico83, line 123:
```suggestion
				bw.Logger.Infof("Failed to get Okteto Build service info: %v\n", err)
```

pkg/build/buildkit/wait.go
```diff
@@ -0,0 +1,135 @@
...
+				bw.Logger.Infof("Retrying in %v...\n", bw.RetryInterval)
+				bw.Sleeper.Sleep(bw.RetryInterval)
+				continue
+			}
+			bw.waitingTime = time.Since(startWaitingTime)
+			bw.Logger.Infof("Connected to BuildKit server.")
```
pchico83, line 129:
```suggestion
			bw.Logger.Infof("Connected to Okteto Build service.")
```

pkg/build/buildkit/wait.go
```diff
@@ -0,0 +1,135 @@
...
+func (bw *buildkitWaiter) GetWaitingTime() time.Duration {
+	return bw.waitingTime
+}
+
+// WaitForBuildKit waits for the BuildKit server to become available
+func (bw *buildkitWaiter) WaitUntilIsUp(ctx context.Context) error {
+	ctx, cancel := context.WithTimeout(ctx, bw.MaxWaitTime)
+	defer cancel()
+
+	sp := bw.Logger.Out().Spinner("Waiting for BuildKit server to be ready...")
```
pchico83, line 103:
```suggestion
	sp := bw.Logger.Out().Spinner("Waiting for Okteto Build service to be available...")
```

pchico83 (approved):
Let's use consistent names for buildkit. I would always use "buildkit service", instead of just "buildkit" or "BuildKit server"

pkg/build/buildkit/client.go
```diff
@@ -0,0 +1,166 @@
...
+// osFileWriter implements the fileWriter interface
+type osFileWriter struct{}
+
+// WriteFile writes data to a file
+func (o *osFileWriter) WriteFile(filename string, data []byte, perm os.FileMode) error {
+	return os.WriteFile(filename, data, perm)
+}
+
+// buildkitClientCreator defines an interface for creating buildkit clients
+type buildkitClientCreator interface {
```
ifbyol, line 73:
```suggestion
type clientCreator interface {
```

Nit. This is already within a package called `buildkit`. Would it make sense to just call it `clientCreator`?

pkg/build/buildkit/wait.go
```diff
@@ -0,0 +1,135 @@
...
+type buildkitClientFactoryWaiter interface {
+	GetBuildkitClient(ctx context.Context) (clientInfoRetriever, error)
+}
+
+type clientInfoRetriever interface {
+	Info(ctx context.Context) (*client.Info, error)
+}
+
+// buildkitWaiter encapsulates the logic to check if the BuildKit server is up and running
+type buildkitWaiter struct {
```
ifbyol, line 62: Why do we have this type private to the package, but some of its properties are public? This sounds odd. Specially not using an interface in the consumer of this struct.

We are mixing approaches here. Why `waitingTime` is private, but we expose a function to retrieve its value, but for others, we do the opposite (make it public). 

Also, it sounds odd that `Logger` is exported. It should be internal and only log wait operations. I think it is even being used only internal.

Could you review this and follow the same pattern? I think this struct should also be public if it is going to be used directly outside this package

pkg/build/buildkit/client_test.go
```diff
@@ -0,0 +1,263 @@
...
+	"net/url"
+	"os"
+	"testing"
+
+	"github.com/moby/buildkit/client"
+	"github.com/okteto/okteto/pkg/log/io"
+	"github.com/stretchr/testify/assert"
+)
+
+// urlParser defines an interface for parsing URLs
```
andreafalzetti, line 28:
```suggestion
// fakeURLParser defines an interface for parsing URLs
```

pkg/build/buildkit/client_test.go
```diff
@@ -0,0 +1,263 @@
...
+	url *url.URL
+	err error
+}
+
+// Parse parses a URL
+func (p *fakeURLParser) Parse(rawurl string) (*url.URL, error) {
+	return p.url, p.err
+}
+
+// certDecoder defines an interface for decoding certificates
```
andreafalzetti, line 39:
```suggestion
// fakeCertDecoder defines an interface for decoding certificates
```

andreafalzetti: or remove it

pkg/build/buildkit/client_test.go
andreafalzetti: or remove it

pkg/build/buildkit/wait.go
```diff
@@ -0,0 +1,135 @@
...
+	maxWaitTime = 10 * time.Minute
+
+	// retryTime is the time to wait between retries
+	retryTime = 5 * time.Second
+
+	// maxBuildkitWaitTimeEnvVar is the environment variable to set the max wait time for buildkit
+	maxBuildkitWaitTimeEnvVar = "OKTETO_BUILDKIT_WAIT_TIMEOUT"
+
+	// retryBuildkitTimeEnvVar is the environment variable to set the retry time for buildkit
+	retryBuildkitTimeEnvVar = "OKTETO_RETRY_BUILDKIT_TIME"
```
andreafalzetti, line 37: I would use the word Interval instead of Time

pkg/build/buildkit/wait.go
```diff
@@ -0,0 +1,135 @@
...
+		buildkitClientFactory: buildkitClientFactoryToWaitFactory(factory),
+		Logger:                logger,
+	}
+}
+
+func (bw *buildkitWaiter) GetWaitingTime() time.Duration {
+	return bw.waitingTime
+}
+
+// WaitForBuildKit waits for the BuildKit server to become available
```
andreafalzetti, line 98:
```suggestion
// WaitUntilIsUp waits for the Buildkit server to become available
```

andreafalzetti:
When I run `okteto build --progress plain --log-level debug --no-cache` and scaled to `0`, I got this:

```
INFO[0039] Error building service 'worker': error building image 'registry.andreafalzetti.dev.okteto.net/andreafalzetti-okteto-admin/movies-worker:okteto,registry.andreafalzetti.dev.okteto.net/andreafalzetti/movies-worker:adeda6b211c4152c1a2a0e249b77e9776d7962bde8eb2ff3ef664fdd1bae167c': build failed: failed to receive status: rpc error: code = Internal desc = stream terminated by RST_STREAM with error code: INTERNAL_ERROR
 x  Error building service 'worker': error building image 'registry.andreafalzetti.dev.okteto.net/andreafalzetti-okteto-admin/movies-worker:okteto,registry.andreafalzetti.dev.okteto.net/andreafalzetti/movies-worker:adeda6b211c4152c1a2a0e249b77e9776d7962bde8eb2ff3ef664fdd1bae167c': build failed: failed to receive status: rpc error: code = Internal desc = stream terminated by RST_STREAM with error code: INTERNAL_ERROR
 ```

Only when I re-run the command, I got the spinner: `⠧ Waiting for BuildKit service to be ready...`.

Is it expected?

File: pkg/build/buildkit/wait.go [hunk 12]
```diff
@@ -0,0 +1,135 @@
...
+	sp := bw.logger.Out().Spinner("Waiting for BuildKit service to be ready...")
+	sp.Start()
+	defer sp.Stop()
+
+	startWaitingTime := time.Now()
+	for {
+		select {
+		case <-ctx.Done():
+			bw.waitingTime = bw.maxWaitTime
+			return fmt.Errorf("buildkit service not available for %v", bw.maxWaitTime)
```
andreafalzetti, line 112: nit: the following scenarios can be a _misconfiguration_ but  when `bw.retryInterval` is grather than `bw.maxWaitTime`, the total time should be `bw.retryInterval`.

Example:

```
OKTETO_RETRY_BUILDKIT_INTERVAL=20s OKTETO_BUILDKIT_WAIT_TIMEOUT=1s okd build
 x  Error building service 'frontend': buildkit service not available for 1s
```

andreafalzetti (approved):
Left some comments. The only one we should be careful is about the error. 

# 👏

jLopezbarb:
> When I run `okteto build --progress plain --log-level debug --no-cache` and scaled to `0`, I got this:
> 
> ```
> INFO[0039] Error building service 'worker': error building image 'registry.andreafalzetti.dev.okteto.net/andreafalzetti-okteto-admin/movies-worker:okteto,registry.andreafalzetti.dev.okteto.net/andreafalzetti/movies-worker:adeda6b211c4152c1a2a0e249b77e9776d7962bde8eb2ff3ef664fdd1bae167c': build failed: failed to receive status: rpc error: code = Internal desc = stream terminated by RST_STREAM with error code: INTERNAL_ERROR
>  x  Error building service 'worker': error building image 'registry.andreafalzetti.dev.okteto.net/andreafalzetti-okteto-admin/movies-worker:okteto,registry.andreafalzetti.dev.okteto.net/andreafalzetti/movies-worker:adeda6b211c4152c1a2a0e249b77e9776d7962bde8eb2ff3ef664fdd1bae167c': build failed: failed to receive status: rpc error: code = Internal desc = stream terminated by RST_STREAM with error code: INTERNAL_ERROR
> ```
> 
> Only when I re-run the command, I got the spinner: `⠧ Waiting for BuildKit service to be ready...`.
> 
> Then, I deleted the statefulset and got this error after the last pod was terminated:
> 
> ```
>  x  Error building service 'rent': error building image 'registry.andreafalzetti.dev.okteto.net/andreafalzetti-okteto-admin/movies-rent:okteto,registry.andreafalzetti.dev.okteto.net/andreafalzetti/movies-rent:d96f6cabb4fac9b0d75f790efd051e1862212efe19827a398157d4dd278c3d12': build failed: failed to receive status: rpc error: code = Internal desc = stream terminated by RST_STREAM with error code: INTERNAL_ERROR
> ```
> 
> Is it expected?

Yes, that's expected at least in this PR. I did the issue in two different PRs, in this PR it only waits if buildkit is not available at the begining but the following PR does the retries

pkg/build/buildkit/wait.go
jLopezbarb: I agree but it's weird that the max waiting time that a user has set is lower than the error we show

pkg/build/buildkit/client.go
teresaromero, line 1: out of curiosity, why under `build` and not its own pkg as a third party dependency?

pkg/build/buildkit/wait.go
```diff
@@ -0,0 +1,135 @@
...
+	retryBuildkitIntervalEnvVar = "OKTETO_RETRY_BUILDKIT_INTERVAL"
+)
+
+// sleeper defines an interface for sleeping
+type sleeper interface {
+	Sleep(duration time.Duration)
+}
+
+// RealSleeper implements the Sleeper interface using time.Sleep
+type RealSleeper struct{}
```
teresaromero, line 46: what does it mean "real" ?

teresaromero (approved):

PR #4529, opened 2024-10-10
feat: retry if buildkit not available

jLopezbarb:
# Proposed changes

Fixes partly DEV-679
Depends on: #4528 

- Refactor buildkit solve build: In order to be able to retry a certain amount of types we needed to refactor it. 
- Move buildkit errors to its own package.
- Check buildkit grpc erros. All transient errors from grpc that must be retried were code 13 and 14 ([GRPC Errors](https://grpc.io/docs/guides/status-codes/#the-full-list-of-status-codes))
- Add unit tests checking all the possible scenarios. See pkg/build/buildkit/runner_test.go for the scenarios
- Added the environment variable `OKTETO_MAX_RETRIES_FOR_BUILDKIT_TRANSIENT_ERRORS` in order to configure how many times a build will be attempted. (Needs documentation when we decide the naming convention)

## Scenarios tested

Tested all the scenarios using the following Dockerfile

```Dockerfile
FROM alpine

RUN sleep 10000000000000
```

- [ ] Scale down buildkit statefulset while building (Code 13)
- [ ] kubectl delete pod (Code 13)
- [ ] kubectl exec removing buildkit (Code 13) pod restarts
- [ ] kubectl exec removing build session (error non transient)
- [ ] kubectl exec removing command inside session (error non transient)
- [ ] kubectl drain node (Code 13)
- [ ] kubectl delete node (Code 14)

## CLI Quality Reminders 🔧


pchico83:
@jLopezbarb let's refer to buildkit as "Okteto Build service". Could you share screenshots of the errors, or info messages shown to the user?

pkg/cmd/build/build.go
```diff
@@ -175,41 +174,14 @@ func (ob *OktetoBuilder) buildWithOkteto(ctx context.Context, buildOptions *type
...
-		return err
-	}
-
-	var tag string
-	if buildOptions != nil {
-		tag = buildOptions.Tag
-		if buildOptions.Manifest != nil && buildOptions.Manifest.Deploy != nil {
-			tag = buildOptions.Manifest.Deploy.Image
-		}
-	}
```
ifbyol, line 210: Where is this logic now? Isn't it needed?

ifbyol (commented):
Have you checked how logs in the UI and CLI looks like when there is a retry during a remote deploy, destroy o test (in this case the UI doesn't apply) operation? Could you share an screenshot on how it looks like, please?

pkg/build/buildkit/runner.go
```diff
@@ -0,0 +1,157 @@
...
+	WaitUntilIsUp(ctx context.Context) error
+}
+
+type buildkitClientFactory interface {
+	GetBuildkitClient(ctx context.Context) (*client.Client, error)
+}
+
+type runnerMetadata struct {
+	attempts int
+	// TODO: use this to track the time it takes to solve a build instead of the full build time
```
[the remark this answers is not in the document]
jLopezbarb: There is no ticket but checking the actual (code)[https://github.com/okteto/okteto/blob/master/cmd/build/v2/build.go#L293-L301] we are also adding the time waiting for the actual build and retries, instead of just the build proccess. With this change and if the retries augments this problem will be more common

pkg/cmd/build/build.go
jLopezbarb: I checked that with deploy remote we were already capturing the correct error. We changed from checking if the buildOptions was empty infer that is a deploy/destroy/test operation to capture the error directly from the error itself (changed in 3.0)

pkg/build/buildkit/errors.go
```diff
@@ -22,10 +22,35 @@ import (
...
+	"google.golang.org/grpc/status"
 )
 
-// getErrorMessage returns the parsed error message
-func getErrorMessage(err error, tag string) error {
+var (
+	// ErrBuildConnecionFailed is returned when the connection to buildkit fails
+	ErrBuildConnecionFailed = errors.New("build connection failure")
+
+	gpcRetryableCodes = map[uint32]interface{}{
```
andreafalzetti, line 32: grpc?

pkg/build/buildkit/runner.go
```diff
@@ -0,0 +1,167 @@
...
+	var solveTime time.Duration
+	defer func() {
+		r.metadata.attempts = attempts
+		r.metadata.solveTime = solveTime
+	}()
+	for {
+		attempts++
+		if attempts > 1 {
+			r.logger.Logger().Infof("retrying build, attempt %d", attempts)
+			r.logger.Out().Warning("Buildkit connection failure. Retrying...")
```
andreafalzetti, line 94: In the [other PR](https://github.com/okteto/okteto/pull/4528#pullrequestreview-2360853208), we mentioned using always the same name. Shall we change this to adhere?

```suggestion
			r.logger.Out().Warning("BuildKit service connection failure. Retrying...")
```

pkg/build/buildkit/runner.go
```diff
@@ -0,0 +1,167 @@
...
+		}
+		// if buildkit is not available for 10 minutes, we should fail
+		if err := r.waiter.WaitUntilIsUp(ctx); err != nil {
+			r.logger.Logger().Infof("failed to wait for buildkit to be available: %s", err)
```
andreafalzetti, line 98:
```suggestion
			r.logger.Logger().Infof("failed to wait for the BuildKit service to be available: %s", err)
```

andreafalzetti (approved):

jLopezbarb:
> Have you checked how logs in the UI and CLI looks like when there is a retry during a remote deploy, destroy o test (in this case the UI doesn't apply) operation? Could you share an screenshot on how it looks like, please?

In the UI We have the logs for all the retries but only one stage is shown:
Screenshot 2024-10-17 at 18 30 04
Screenshot 2024-10-17 at 18 31 28

In the CLI  we show a message when we do a retry:
Screenshot 2024-10-17 at 18 32 01

pkg/build/buildkit/errors.go
```diff
@@ -22,10 +22,35 @@ import (
...
+	ErrBuildConnecionFailed = errors.New("build connection failure")
+
+	grpcRetryableCodes = map[uint32]interface{}{
+		13: nil, // transport is closing (e.g. client buildkit pod is being deleted)
+		14: nil, // unavailable (e.g. no buildkit service available)
+		15: nil, // Data loss: request did not complete
+	}
+)
+
+type OktetoCommandErr struct {
```
teresaromero, line 39: i am confused by the naming, does this mean is a command from the okteto manifest or is a command of buildkit running? :/ 
if is the second, should we remove okteto from the type?

jLopezbarb: It's the name of the command inside buildkit. I'll change it if its clearer

teresaromero (approved):

PR #4530, opened 2024-10-11
update stack destroy logic to use service name label

jLopezbarb:
# Proposed changes

Fixes DEV-634

There was a bug that we were removing the deployment of the service that was running as the dev container leaving the original without scaling. This PR checks if the deployment is part of the stack or not by using their service name label instead of the name. This way the current dev container is not removed and the okteto up sesisons are not forced to retry

## How to validate

2. Run `okteto deploy -f docker-compose.yml`
3. Create the following `okteto.yml`
```yaml
deploy:
  compose:
  - file: docker-compose.yml
    services:
      - frontend
dev:
  testing:
    image: okteto/movies-with-compose:api
    selector:
      stack.okteto.com/service: api
    forward:
    - 9090:frontend:8080
```
4. Run `okteto up testing`
5. Run `okteto deploy` 

## CLI Quality Reminders 🔧

### DIFF
--- a/pkg/cmd/pipeline/translate.go
+++ b/pkg/cmd/pipeline/translate.go
@@ -382,6 +382,13 @@ func AddDevAnnotations(ctx context.Context, manifest *model.Manifest, c kubernet
 			oktetoLog.Infof("could not add %s dev annotations due to: %s", devName, err.Error())
 			continue
 		}
+		if repo != "" && app.ObjectMeta().Annotations[model.OktetoRepositoryAnnotation] == repo {
+			continue
+		}
+		if app.ObjectMeta().Annotations[model.OktetoDevNameAnnotation] != "" {
+			continue
+		}
+
 		if repo != "" {
 			app.ObjectMeta().Annotations[model.OktetoRepositoryAnnotation] = removeSensitiveDataFromGitURL(repo)
 		}
--- a/pkg/cmd/pipeline/translate.go
+++ b/pkg/cmd/pipeline/translate.go
@@ -382,10 +382,14 @@ func AddDevAnnotations(ctx context.Context, manifest *model.Manifest, c kubernet
 			oktetoLog.Infof("could not add %s dev annotations due to: %s", devName, err.Error())
 			continue
 		}
-		if repo != "" && app.ObjectMeta().Annotations[model.OktetoRepositoryAnnotation] == repo {
+		sanitisedRepo := removeSensitiveDataFromGitURL(repo)
+		repositoryAnnotation := app.ObjectMeta().Annotations[model.OktetoRepositoryAnnotation]
+		devNameAnnotation := app.ObjectMeta().Annotations[model.OktetoDevNameAnnotation]
+
+		if repo != "" && repositoryAnnotation == sanitisedRepo {
 			continue
 		}
-		if app.ObjectMeta().Annotations[model.OktetoDevNameAnnotation] != "" {
+		if devNameAnnotation != "" {
 			continue
 		}
 
--- a/pkg/cmd/pipeline/translate.go
+++ b/pkg/cmd/pipeline/translate.go
@@ -394,7 +394,7 @@ func AddDevAnnotations(ctx context.Context, manifest *model.Manifest, c kubernet
 		}
 
 		if repo != "" {
-			app.ObjectMeta().Annotations[model.OktetoRepositoryAnnotation] = removeSensitiveDataFromGitURL(repo)
+			app.ObjectMeta().Annotations[model.OktetoRepositoryAnnotation] = sanitisedRepo
 		}
 		app.ObjectMeta().Annotations[model.OktetoDevNameAnnotation] = devName
 		if err := app.PatchAnnotations(ctx, c); err != nil {
--- a/pkg/cmd/stack/destroy.go
+++ b/pkg/cmd/stack/destroy.go
@@ -55,7 +55,7 @@ func destroyDeployments(ctx context.Context, s *model.Stack, c kubernetes.Interf
 	}
 	for i := range dList {
 		svcName := dList[i].ObjectMeta.Labels[model.StackServiceNameLabel]
-		if _, ok := s.Services[svcName]; ok && s.Services[svcName].IsDeployment() {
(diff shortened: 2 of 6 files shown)


ifbyol:
Should the `okteto up` session try to reconnect again? I was testing and the deployment is not deleted, but the up session is stopped

jLopezbarb:
I think it's something it shouldn't happen, but it's something is already happening so I think it should be treated as another issue. WDYT?

ifbyol:
> I think it's something it shouldn't happen, but it's something is already happening so I think it should be treated as another issue. WDYT?

It is already happening because we were deleting the deployment with the dev mode. The main issue would still be that api is not running because we leave it as a dev mode, but with an active session, so I'd say it is not fixed

@pchico83 What should be the expected behavior?

pchico83:
> > I think it's something it shouldn't happen, but it's something is already happening so I think it should be treated as another issue. WDYT?
> 
> It is already happening because we were deleting the deployment with the dev mode. The main issue would still be that api is not running because we leave it as a dev mode, but with an active session, so I'd say it is not fixed
> 
> @pchico83 What should be the expected behavior?



> > I think it's something it shouldn't happen, but it's something is already happening so I think it should be treated as another issue. WDYT?
> 
> It is already happening because we were deleting the deployment with the dev mode. The main issue would still be that api is not running because we leave it as a dev mode, but with an active session, so I'd say it is not fixed
> 
> @pchico83 What should be the expected behavior?

As part of this issue, the behavior should be:
- if I redeploy and my service has no changes, my dev container is not restarted and the `okteto up` session keeps connected
- if I redeploy and my service *has changes* that updates the original deployment, my dev container is recreated and `okteto up` reconnects automatically

jLopezbarb:
The problem was that we were adding the [devAnnotations](https://github.com/okteto/okteto/blob/1c8e5e280f8ecf039c4456ddd33e7dd71ea997d1/pkg/cmd/pipeline/translate.go#L373). In this specific case there is a problem because we have two different dev in the manifest pointing to the same app. We fixed it by only replacing when the information is not already there

pkg/cmd/pipeline/translate.go
```diff
@@ -382,6 +382,13 @@ func AddDevAnnotations(ctx context.Context, manifest *model.Manifest, c kubernet
 			oktetoLog.Infof("could not add %s dev annotations due to: %s", devName, err.Error())
 			continue
 		}
+		if repo != "" && app.ObjectMeta().Annotations[model.OktetoRepositoryAnnotation] == repo {
+			continue
+		}
+		if app.ObjectMeta().Annotations[model.OktetoDevNameAnnotation] != "" {
+			continue
+		}
```
teresaromero, line 390:
```suggestion
                 annotationRepo := app.ObjectMeta().Annotations[model.OktetoRepositoryAnnotation]
		if  annotationRepo  == repo && repo != "" {
			continue
		}
```

both have to be the same and not empty ? could we reduce conditions like that?

pkg/cmd/stack/destroy.go
```diff
@@ -54,7 +54,8 @@ func destroyDeployments(ctx context.Context, s *model.Stack, c kubernetes.Interf
 		return err
 	}
 	for i := range dList {
-		if _, ok := s.Services[dList[i].Name]; ok && s.Services[dList[i].Name].IsDeployment() {
+		svcName := dList[i].ObjectMeta.Labels[model.StackServiceNameLabel]
+		if _, ok := s.Services[svcName]; ok && s.Services[svcName].IsDeployment() {
```
teresaromero, line 58:
```suggestion
		if service, ok := s.Services[svcName]; ok && service.IsDeployment() {
```

teresaromero (approved):

pkg/cmd/pipeline/translate.go
ifbyol: I think you are removing the condition to check `model.OktetoDevNameAnnotation` annotation. Maybe I'm seeing wrong the diff you provide, but each if is checking a different annotation

pkg/cmd/pipeline/translate.go
```diff
@@ -382,6 +382,13 @@ func AddDevAnnotations(ctx context.Context, manifest *model.Manifest, c kubernet
 			oktetoLog.Infof("could not add %s dev annotations due to: %s", devName, err.Error())
 			continue
 		}
+		if repo != "" && app.ObjectMeta().Annotations[model.OktetoRepositoryAnnotation] == repo {
```
ifbyol, line 385: Should we check into account if the repo is sanitized? In line 393 we are setting the value of the annotation with the sanitized repo 

`app.ObjectMeta().Annotations[model.OktetoRepositoryAnnotation] = removeSensitiveDataFromGitURL(repo)`

so I assume we should check also that value here instead of just `repo`, right?

pkg/cmd/pipeline/translate.go
teresaromero: omg, you are right. my mistake

teresaromero (approved):

pkg/cmd/pipeline/translate.go
```diff
@@ -382,6 +382,17 @@ func AddDevAnnotations(ctx context.Context, manifest *model.Manifest, c kubernet
...
+
+		if repo != "" && repositoryAnnotation == sanitisedRepo {
+			continue
+		}
+		if devNameAnnotation != "" {
+			continue
+		}
+
 		if repo != "" {
 			app.ObjectMeta().Annotations[model.OktetoRepositoryAnnotation] = removeSensitiveDataFromGitURL(repo)
```
ifbyol, line 397:
```suggestion
			app.ObjectMeta().Annotations[model.OktetoRepositoryAnnotation] = sanitisedRepo
```

Nit

ifbyol (commented):
I was testing it, and I think this part `if I redeploy and my service has changes that updates the original deployment, my dev container is recreated and okteto up reconnects automatically` is not working as defined

If I execute `okteto up api` and then, modify the example you provided to deploy also `api` service, the current up session is killed and it doesn't reconnect

ifbyol (approved):

PR #4532, opened 2024-10-17
Deploy/Destroy context

maroshii:
Add `context` property to deploy and destroy commands.

`context` will be considered in resolving the cwd in deploy and destroy commands remote AND local.

Given the following project:

```
.
├── file.txt
└── nested
    └── okteto.yml
```

And the following manifest:

```yaml
deploy:
  context: ../
  commands:
    - cat file.txt

destroy:
  context: ../
  commands:
    - cat file.txt
```

This should work in either remote or local:

```bash
okteto deploy -f nested/okteto.yml [--remote]
okteto destroy -f nested/okteto.yml [--remote]
(cd nested; okteto deploy [--remote])
(cd nested; okteto destroy [--remote])

```

### DIFF
--- a/cmd/deploy/remote_test.go
+++ b/cmd/deploy/remote_test.go
@@ -243,7 +243,7 @@ func TestDeployRemoteWithCtx(t *testing.T) {
 	runner.AssertExpectations(t)
 }
 func TestDeployRemote(t *testing.T) {
-	workdirCtrl := fakefs.NewFakeWorkingDirectoryCtrl("/")
+	workdirCtrl := fakefs.NewFakeWorkingDirectoryCtrl("/path/to/manifest")
 	manifest := &model.Manifest{
 		Deploy: &model.DeployInfo{
 			Image: "test-image",
@@ -313,7 +313,7 @@ func TestDeployRemote(t *testing.T) {
 }
 
 func TestDeployRemoteWithError(t *testing.T) {
-	workdirCtrl := fakefs.NewFakeWorkingDirectoryCtrl("/")
+	workdirCtrl := fakefs.NewFakeWorkingDirectoryCtrl("/path/to/manifest")
 	manifest := &model.Manifest{
 		Deploy: &model.DeployInfo{
 			Image: "test-image",
--- a/cmd/destroy/remote.go
+++ b/cmd/destroy/remote.go
@@ -117,7 +117,7 @@ func (rd *remoteDestroyCommand) Destroy(ctx context.Context, opts *Options) erro
 
 	ctxPath := cwd
 	if opts.Manifest.Destroy != nil {
-		ctxPath = path.Clean(path.Join(cwd, opts.Manifest.Deploy.Context))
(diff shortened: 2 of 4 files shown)


teresaromero (commented):
can we run e2e and perhaps add an integration test with this case?

cmd/deploy/remote.go
```diff
@@ -116,6 +124,12 @@ func (rd *remoteDeployer) Deploy(ctx context.Context, deployOptions *Options) er
 		return fmt.Errorf("failed to create ignore rules for remote deploy: %w", err)
 	}
 
+	var ctxPath string
+	if deployOptions.Manifest.Deploy != nil {
+		manifestPathDir := filepath.Dir(deployOptions.ManifestPathFlag)
+		ctxPath = path.Clean(path.Join(cwd, manifestPathDir, deployOptions.Manifest.Deploy.Context))
```
jLopezbarb, line 130: What if `deployOptions.Manifest.Deploy.Context` is absolute?

ifbyol, line 130: If I'm not wrong, `cwd` here should already be pointing to the dir where the manifest is located, right @jLopezbarb ?

ifbyol (commented):
I found an scenario where it is not working fine.

Having the following filesystem structure:

```
- app/
       | - okteto.yaml
       | - subdir/
              | - subdir-file.txt 
```

And having the Okteto Manifest with the following content:

```
deploy:
  context: subdir
  commands:
    - ls -la
    - cat subdir-file.txt
    - cd .. && ls -la
    - pwd

destroy:
  context: subdir
  commands:
    - ls -la
    - cat subdir-file.txt
    - cd .. && ls -la
    - pwd
```

When I'm inside the `app/subdir` folder and I execute `okteto deploy -f ../okteto.yaml --remote` or `okteto destroy -f ../okteto.yaml --remote`, I'm having the following error:

`Error building image '': build failed: resolve : lstat /private/tmp/test-context-subdir/subdir: no such file or directory`

In the path specified there, `app` is missing between `test-context-subdir` and `subdir`.

Also, unit tests seem to be failing

cmd/deploy/remote.go
jLopezbarb: Tested and ctxPath is set to a wrong path

jLopezbarb: Why? cwd it's retrieved using the function `GetOriginalCWD` which returns the cwd that the user has used

ifbyol: > Using the vanilla cwd fixes the issue given that at some point I assume we are changing the workdir to where the manifest is located.

Yes, deploy, destroy and tests commands call to this function `UpdateCWDtoManifestPath` which updated the CWD to the folder where the Okteto Manifest is located

ifbyol (approved):
Code and functionality looks good! I've been testing it. It seems some unit tests are failing, but seems more related to adapt them to new changes that something wrong

jLopezbarb (approved):

teresaromero (approved):

PR #4533, opened 2024-10-22
remove validation from context store

jLopezbarb:
# Proposed changes

Fixes DEV-694

This was a problem introduced by DEV-339. This PR reverts the validation from that PR and fixes the context store if it's not completed. 

## How to validate

1. Modify your context/config.json and remove all the contexts
1. Run okteto deploy/context and check that there is no panic

## CLI Quality Reminders 🔧

### DIFF
--- a/pkg/okteto/context.go
+++ b/pkg/okteto/context.go
@@ -267,19 +267,6 @@ func GetContextStoreFromStorePath() *ContextStore {
 	return ctxStore
 }
 
-func ValidateContextStore(ctxStore *ContextStore) error {
-	if ctxStore.Contexts == nil {
-		return fmt.Errorf("contexts cannot be nil")
-	}
-	if ctxStore.CurrentContext == "" {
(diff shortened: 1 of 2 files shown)


pkg/okteto/context.go
```diff
@@ -236,6 +236,9 @@ func GetContextStore() *ContextStore {
 
 	if ContextExists() {
```
teresaromero, line 237: how come if "Context exists" we check if is nil inside? 🤔

ifbyol (approved):

pkg/okteto/context.go
jLopezbarb: It can exist the file but be an empty json `{}` which led to panics later on the code

teresaromero: nit: it that case i would rather call the `ContextExists` `ContextFileExists`

teresaromero (approved):

PR #4536, opened 2024-10-23
Deploy Execution default and flags outcome

teresaromero:
# Proposed changes

Resolves https://okteto.atlassian.net/browse/DEV-759

- Add new internal Option to Deploy and Destroy cmd to capture wheather the user has used the flag "remote" or is the default value for boolean (false)
- Change Manifest Deploy/Destroy option "Remote" to a boolean pointer to be able to detect if the value is set by the user at the manifest or not

Change the shouldRunRemote functions logic to adapt to the following:

- If user uses flag, this value is used
- If user uses manifest option, this value is used. 
- If local execution and manifest image is set, we keep running in remote
- Default is now read from cluster settings

## CLI Quality Reminders 🔧


cmd/deploy/deploy.go
```diff
@@ -573,15 +576,15 @@ func getDefaultTimeout() time.Duration {
...
+	if opts.RunInRemoteSet {
+		return opts.RunInRemote
 	}
 
-	// remote option set in the manifest via a remote deployer image or the remote option enabled
+	// remote option set in the manifest via the remote option enabled
 	if opts.Manifest != nil && opts.Manifest.Deploy != nil {
-		if opts.Manifest.Deploy.Image != "" || opts.Manifest.Deploy.Remote {
-			return true
+		if opts.Manifest.Deploy.Remote != nil {
```
andreafalzetti, line 586: with this change, if I have `image:` defined, it doesn't necessarily remote deploy, correct?

If that's the case, it sounds a breaking change somehow, no? I couldn't find reference to it in the specs 🤔

teresaromero: This is part of the migration path, there is a block in the spec for this. i might have missunderstood smthg. i am reviewing this. thanks!

ifbyol: We need to double check this, because there was a comment from Pablo in the spec a bit confusing

teresaromero: added a comment on the spec to get some answers to this. thanks

teresaromero: Updated PR and open for reviews

cmd/test/cmd.go
```diff
@@ -380,9 +380,12 @@ func doRun(ctx context.Context, servicesToTest []string, options *Options, ioCtr
 
 func shouldRunInRemote(manifestDeploy *model.DeployInfo) bool {
 	if manifestDeploy != nil {
-		if manifestDeploy.Image != "" || manifestDeploy.Remote {
+		if manifestDeploy.Image != "" {
```
ifbyol, line 383: We are not checking the value of the env var here, It doesn't seem correct but I might be missing something

ifbyol: Ok, this works because this is changing the `--remote` flag for the deploy command, but we are not changing the value `RunInRemoteSet`, so it is not considering as the value to apply. Which seems odd.

I don't know if we should calculate here if the deploy has to be executed remotely or not. I'd leave it empty and rely on the deploy command to do the calculation, as we do for `okteto up`. If not, I would do the full calculation here setting also `RunInRemoteSet` to be consistent. As it is now, it is a bit misleading imo

ifbyol:
In terms of functionality, it works fine, but I would review this comment: https://github.com/okteto/okteto/pull/4536#discussion_r1820420284

cmd/test/cmd.go
teresaromero: i am looking into this. The test cmd is requiring this information to set the deploy.Options (as if the flag was being used)

The logic that sets if is deploy remote or not is being used when GetDeployer, so i think this might not be needed as the calc to run remote is done there. I will check it and get back. thanks!

teresaromero: after looking into the code, updated the PR using in test the same criteria as in deploy cmd @andreafalzetti please take a look as initial implementor to check the changes make sense. thanks!

ifbyol (approved):

jLopezbarb (approved):

PR #4537, opened 2024-10-23
DEV-758: Implement ssh-forwarder to contact with ssh-agent running in the cluster when executing remote operations (deploy, destroy, test)

ifbyol:
# Proposed changes

Fixes DEV-758

Implement ssh-forwarder process to use ssh-agent running in the Okteto cluster when some command defined in an Okteto Manifest needs ssh, and that Okteto Manifest is executed remotely (okteto deploy --remote, okteto destroy --remote, okteto test).

Main changes included in the PR are:
* As part of any subcommand of `remote-run` command, start a go routine which listens in a local UNIX socket and forward everything received there to the ssh-agent running in the Okteto cluster.
* Modify the Dockerfile generated for remote operations (deploy, destroy and test) to define `SSH_AUTH_SOCK` env var if the ssh-agent host name is defined
* Don't mount anymore the local agent running in the machine
* Retrieve from the cluster metadata query the information related to the ssh-agent (hostname, internal IP and port), and pass it to the `remote-run` command through env variables defined in the generated Dockerfile
* Mount always the known_hosts file if exists locally

## How to validate

1. Update your dev environment with the latest changes of master
2. Build the image from this branch using your environment context.
  * `okteto ctx use https://...`
  * `okteto build -f Dockerfile -t okteto.dev/cli:test-pr` 
5. Take the expanded URL of the image build, and define an admin variable named `OKTETO_REMOTE_CLI_IMAGE` with the URL as value
6. Build the go binary from this branch
7. Take the public SSH of your cluster and add it to your user in GitHub (https://www.okteto.com/docs/admin/private-repositories/ssh-key/)
8. Create an Okteto Manifest like the following one:

```
deploy:
  remote: true
  commands:
    - ssh-add -L
    - git clone <private-repo>
```
9. Execute `okteto deploy` with the binary created from this branch. You should see how it lists one identity, and it should match with the SSH key showed in the admin console of your environment. You should also be able to clone the private repository.
10. You can also try with `destroy` and `test` sections of the Okteto Manifest, as they should work too

## CLI Quality Reminders 🔧

### DIFF
--- a/cmd/remoterun/deploy.go
+++ b/cmd/remoterun/deploy.go
@@ -141,12 +141,21 @@ It is important that this command does the minimum and must not do calculations 
 			sshAgentPort := os.Getenv(constants.OktetoSshAgentPortEnvVar)
 			sshSocket := os.Getenv(constants.OktetoSshAgentSocketEnvVar)
 
-			if sshAgentHostname != "" {
+			if sshAgentHostname != "" && sshAgentPort != "" && sshSocket != "" {
 				forwarder := newSSHForwarder()
 				go func() {
 					err := forwarder.startSshForwarder(ctx, sshAgentHostname, sshAgentPort, sshSocket, oktetoContext.GetCurrentToken())
 					oktetoLog.Infof("error starting ssh forwarder %v", err)
 				}()
+			} else {
+				message := fmt.Sprintf("ssh forwarded not started because some mandatory configuration parameters are missing. %s=%s, %s=%s and %s=%s",
+					constants.OktetoSshAgentHostnameEnvVar,
+					sshAgentHostname,
+					constants.OktetoSshAgentPortEnvVar,
+					sshAgentPort,
+					constants.OktetoSshAgentSocketEnvVar,
+					sshSocket)
+				oktetoLog.Info(message)
 			}
 
 			return c.Run(ctx, params)
--- a/cmd/remoterun/deploy.go
+++ b/cmd/remoterun/deploy.go
@@ -143,7 +143,10 @@ It is important that this command does the minimum and must not do calculations 
 
 			if sshAgentHostname != "" {
 				forwarder := newSSHForwarder()
-				go forwarder.startSshForwarder(sshAgentHostname, sshAgentPort, sshSocket, oktetoContext.GetCurrentToken())
+				go func() {
+					err := forwarder.startSshForwarder(sshAgentHostname, sshAgentPort, sshSocket, oktetoContext.GetCurrentToken())
+					oktetoLog.Infof("error starting ssh forwarder %v", err)
+				}()
 			}
 
 			return c.Run(ctx, params)
--- a/cmd/remoterun/deploy.go
+++ b/cmd/remoterun/deploy.go
@@ -16,9 +16,11 @@ package remoterun
 import (
 	"context"
 	"fmt"
+	"os"
 
 	contextCMD "github.com/okteto/okteto/cmd/context"
 	deployCMD "github.com/okteto/okteto/cmd/deploy"
+	"github.com/okteto/okteto/pkg/constants"
 	"github.com/okteto/okteto/pkg/deployable"
 	oktetoErrors "github.com/okteto/okteto/pkg/errors"
 	oktetoLog "github.com/okteto/okteto/pkg/log"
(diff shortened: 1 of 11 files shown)


cmd/remoterun/ssh_test.go
```diff
@@ -0,0 +1,189 @@
...
+	"net"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/okteto/okteto/pkg/model"
+	"github.com/stretchr/testify/require"
+)
+
+func TestSSHForwarder(t *testing.T) {
```
ifbyol, line 36: This test is kind of an integration one instead of a unit one, but it doesn't take too much time, and I think it is better to run it as a unit one, but let me know if you prefer to move it to an e2e one

andreafalzetti (approved):

cmd/remoterun/ssh.go
```diff
@@ -0,0 +1,148 @@
...
+// This process just listens a UNIX socket, and forward the messages to the SSH Agent exposed in the hostname
+// and port received as parameters
+func (s *sshForwarder) startSshForwarder(sshAgentHostname, sshAgentPort, userToken string) {
+	// Remove existing socket if it exists
+	os.Remove(remote.SshAgentLocalSocket)
+
+	// Listen on the UNIX domain socket
+	localListener, err := net.Listen("unix", remote.SshAgentLocalSocket)
+	if err != nil {
+		oktetoLog.Fatalf("Failed to listen on UNIX socket %s: %v", remote.SshAgentLocalSocket, err)
```
jLopezbarb, line 57: Do we need to exit the full program? (this also affects the main execution. Here is a playground [showing this](https://go.dev/play/p/bQ9qDsXDiIp)). Can we consider returning a proper error to the user with our format?

cmd/remoterun/deploy.go
```diff
@@ -135,6 +137,14 @@ It is important that this command does the minimum and must not do calculations
 				runner: runner,
 			}
 
+			sshAgentHostname := os.Getenv(constants.OktetoSshAgentHostnameEnvVar)
+			sshAgentPort := os.Getenv(constants.OktetoSshAgentPortEnvVar)
```
jLopezbarb, line 141: Should we validate that when one is set the other one is also set?

ifbyol: I don't think there could be possible that case. If one is defined, the others are too, but added them to the check just incase

cmd/remoterun/ssh.go
ifbyol: Good catch! I did it in that way to catch faster issue while implementing it gut forgot to change it. It returns an error now, and it doesn't panic

cmd/remoterun/ssh.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+	defer localListener.Close()
```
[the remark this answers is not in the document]
ifbyol: The panic in the error was a mistake. Changed. I also did code changes to handle context cancellation and close the connection

cmd/remoterun/ssh.go
```diff
@@ -0,0 +1,148 @@
...
+	go func() {
+		_, err := io.Copy(localConn, remoteConn)
+		if err != nil && !isClosedNetworkError(err) {
+			oktetoLog.Errorf("Error copying from remote to local: %v", err)
+		}
+		errChan <- err
+	}()
+
+	// Wait for both directions to finish
+	<-errChan
```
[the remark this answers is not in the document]
ifbyol: Yeah, makes sense, changed to use errgroup, thanks!

pkg/types/cluster_metadata.go
```diff
@@ -20,6 +20,9 @@ type ClusterMetadata struct {
 	BuildKitInternalIP  string
 	PublicDomain        string
 	CompanyName         string
+	SshAgentInternalIP  string
```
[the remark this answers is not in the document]
ifbyol: Changed this and the rest of variables, thanks!

cmd/remoterun/ssh.go
```diff
@@ -0,0 +1,148 @@
...
+
+	// Connections will be closed by deferred calls
+}
+
+func isClosedNetworkError(err error) bool {
+	if err == nil {
+		return false
+	}
+
+	return errors.Is(err, net.ErrClosed)
```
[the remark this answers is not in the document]
ifbyol: Yeah, makes sense, removed this function and used directly `errors.Is(net.ErrClosed)` check

ifbyol:
> lgtm 👏
> 
> notes:
> 
> * can we randomize the ssh agent sock path?

Changed to use a different sock path each time

> * can we use timeouts for the protocol answers? this also applies to the backend side

Added changes to include timeout in write and response operations

PR #4544, opened 2024-10-28
DEV-765: Allow to custom known_hosts from runner image

ifbyol:
# Proposed changes

Fixes DEV-765

Make available a known_hosts file defined in `$HOME/.ssh/known_hosts`, so customers might define their own file in the runner base image, and it would be taken into consideration in the remote operations to perform ssh operations.

In that way, If I use an enterprise GitLab under my org, I could create a runner base image including the known_hosts for that gitlab enterprise domain, so users don't need to have it in their local known_hosts file.

Note: [UserKnownHostsFile](https://man.openbsd.org/ssh_config#UserKnownHostsFile) accepts several files separated by whitespaces. I didn't find documentation about which order applies, but after some testing, I think it just tries them and uses the one that works. I tested having a file with wrong keys defined and another one with the good one, and I had the same result independently of the file I was specifying first.

## How to validate

With the same setup described in [this other PR](https://github.com/okteto/okteto/pull/4537):
* Rename your local known_hosts file (usually located in `$HOME/.ssh/known_hosts`) to something else, for example, `known_hosts_copy`
* If you try to clone a private repository from GitHub, you should get an error validating the host.
* Now, build a custom image like this one changing `<route-to-known-hosts-copy>` by the path of the cope made two steps before. Make sure that the image is available to be used (it should be public or for the proper Okteto registry).
```
FROM okteto/pipeline-runner:1.0.5
COPY --chmod=0600 <route-to-known-hosts-copy> /root/.ssh/known_hosts
```
* Change your dev environment runner image (in helm) to use the one built in the previous step
* Try to run again some example cloning the private repository, it should work as the known hosts is now using the one from the base image

## CLI Quality Reminders 🔧

### DIFF
--- a/pkg/remote/remote.go
+++ b/pkg/remote/remote.go
@@ -110,7 +110,7 @@ ENV {{$key}}={{$val}}
 RUN \
   {{range $key, $path := .Caches }}--mount=type=cache,target={{$path}},sharing=private {{end}}\
   --mount=type=secret,id=known_hosts \
-  mkdir -p $HOME/.ssh && echo "UserKnownHostsFile=/run/secrets/known_hosts" >> $HOME/.ssh/config && \
+  mkdir -p $HOME/.ssh && echo "UserKnownHostsFile=/run/secrets/known_hosts $HOME/.ssh/known_hosts" >> $HOME/.ssh/config && \
 {{ if .SSHAgentHostname }}  export SSH_AUTH_SOCK={{ .SSHAgentSocket }} && \{{ end }}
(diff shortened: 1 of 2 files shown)


andreafalzetti (approved):
Validated and it works as expected ✅

jLopezbarb (approved):

ifbyol:
> Have we considered setting known_hosts at runtime through the metadata query instead of requiring a new pipeline runner image with the embedded file? These are not exclusive by the way, so even if we add it to the metadata query we can keep this

Good point! Yes, we have talked about it, and it is in the spec as a future evolution of the feature. The same than adding the ssh agent usage to the installer jobs. This would also be beneficial for the installer jobs as they have the same issue if you try to clone a repository from other host different to the default one (the repo specified in the pipeline o preview deployment).

I have pending to create the tickets for those evolutions, as we think it would be a good approach to simplify this

PR #4545, opened 2024-10-29
DEV-510: Fix availability of CI env var on remote executions

ifbyol:
# Proposed changes

Fixes DEV-510

Nicklas reported some time ago that the `CI` env var wasn't being propagated correctly to the test execution in remote. I was taking a look at it today as it also affected to env var `OKTETO_IS_PREVIEW_ENVIRONMENT`. After some testing, I saw that is related to using `ARG` instead of `ENV`. I'm not sure why, but using ARG with default value is not making that env var available in `remote-run` command. The only way I had to make it available was:
* Passing the build arg with the value [here](https://github.com/okteto/okteto/blob/master/pkg/remote/remote.go#L336)
* Using `ENV` instead of `ARG`

As there is not final image built, which is the main difference between `ENV` and  `ARG` (the availability of that var in the running container or not), I opted by switching it to use `ENV`

## How to validate

* Build the image from this branch
* Export in your terminal the variables `CI` and `OKTETO_IS_PREVIEW_ENVIRONMENT`. `export CI=true` and `export OKTETO_IS_PREVIEW_ENVIRONMENT=true`
* Then, prepare an Okteto manifest like the following one:
```
test:
  unit:
    commands:
      - echo $CI
      - echo $OKTETO_IS_PREVIEW_ENVIRONMENT
```
* If you run `okteto test unit` with the CLI 3.0, you will see how nothing is printed
* If you run it with the version of this branch, you will see how both values are printed

## CLI Quality Reminders 🔧


teresaromero (approved):
was curious and read about this here, in case helps someonelse undertand the issue https://vsupalov.com/docker-arg-env-variable-guide/

jLopezbarb (approved):

pchico83 (approved):

PR #4555, opened 2024-11-05
Deploy watch refactor

teresaromero:
# Proposed changes

This PR is open to discuss the possible refactor of the watch mechanism on the deploy cmd
Following up this PR https://github.com/okteto/okteto/pull/4547

Additionally, i spotted a wierd behavour when certain timeouts where use, i added two additional commits to discuss a possible improvement.

Currently, the wait implementation is done with a ticker every 5 seconds that does a polling of deployments and statefulsets, looking if their replicas are running and skipping when all al running.

Whenever the timeout is set lower than 5 seconds, the process exits with error, but there havent been a call to check the state of the deployment. 

- OPTION1 To solve this my proposal on this commit https://github.com/okteto/okteto/commit/90f22f0c6daf6d04faf6730b4ea6099967a48b77 is to use context with Timeout and wait after the first iteration if there are resources not running, maintaining the 5 second polling. Once the timeout reaches the deadline its canceled and if not finished err is returned.

Also, the mechanism is calling twice to k8s api, one for retrieving deployments and statfulsets list, other to retrieve the deployment or statefulset.

- OPTION 2 watch the deployments and sfs. With this approach, the context used on the k8s pod watch has a timeout deadline so it returns err if not finish before that. While the pods change from its initial state (resourceVersion) the cli will listen for all to be running.

## How to validate

- Compile the binary and run deployments with timeouts/default timeouts
- Check the status of the deployment is correctly updated when running okteto deploy

## CLI Quality Reminders 🔧

### DIFF
--- a/cmd/deploy/wait.go
+++ b/cmd/deploy/wait.go
@@ -57,7 +57,7 @@ func (dw *Waiter) wait(ctx context.Context, opts *Options, namespace string) err
 	}()
 	select {
 	case <-ctx.Done():
-		return fmt.Errorf("%s resources where not healthy after %s", opts.Name, opts.Timeout)
+		return fmt.Errorf("%s resources were not healthy after %s", opts.Name, opts.Timeout)
 	case <-stop:
 		oktetoLog.Infof("CTRL+C received, starting shutdown sequence")
 		return oktetoErrors.ErrIntSig
--- a/cmd/deploy/wait.go
+++ b/cmd/deploy/wait.go
@@ -18,7 +18,6 @@ import (
 	"fmt"
 	"os"
 	"os/signal"
-	"time"
 
 	"github.com/okteto/okteto/pkg/cmd/pipeline"
 	oktetoErrors "github.com/okteto/okteto/pkg/errors"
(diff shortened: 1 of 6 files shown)


cmd/deploy/wait.go
```diff
@@ -66,48 +73,22 @@ func (dw *Waiter) wait(ctx context.Context, opts *Options) error {
...
-			}
-			if !areAllRunning {
-				continue
-			}
-			return nil
-		}
-	}
+	labelSelector := fmt.Sprintf("%s=%s", model.DeployedByLabel, format.ResourceK8sMetaString(opts.Manifest.Name))
+
+	errgroup, ctx := errgroup.WithContext(ctx)
```
teresaromero, line 84: @ifbyol following up your comment of the initial PR https://github.com/okteto/okteto/pull/4547#pullrequestreview-2413953969 

I've updated the aproach creating 2 errgroups to check the deployments and staefulsets instead, but watching instead of polling.

teresaromero:
after offline conversation, closing this PR

PR #4557, opened 2024-11-06
Validate version against cluster supported versions

maroshii:
Check whether the current version is compatible with the cluster version. If it isn't, stop execution and return an error requesting the user to update its cli version.

A few important aspects:

- This can be skipped by simply setting `OKTETO_SKIP_CLUSTER_CLI_VERSION=1`
- If nothing is returned from the metadata query we don't fail and let it through (backwards compatibility)
- We only assert against MAJOR.MINOR versions

### DIFF
--- a/cmd/context/create.go
+++ b/cmd/context/create.go
@@ -332,7 +332,7 @@ func checkCLIVersion(currentVersion, recommendedVersion, minMajorMinor string) e
 	}
 
 	if version.LessThan(recVersion) {
-		oktetoLog.Warning(fmt.Sprintf("Your okteto version %s is older than the recommended version of your okteto instance: %s. Upgrade to the recommended version or set OKTETO_SKIP_CLUSTER_CLI_VERSION=true to supress this message.", currentMajorMinor, recommendedMajorMinor))
+		oktetoLog.Warning(fmt.Sprintf("Your Okteto CLI version %s is older than the recommended version of your Okteto instance: %s. Upgrade to the recommended version or set OKTETO_SKIP_CLUSTER_CLI_VERSION=true to suppress this message.", currentMajorMinor, recommendedMajorMinor))
 	}
 
 	if version.GreaterThan(recVersion) {


pchico83:
@maroshii could you add an screenshot of the error to see how it looks like?
Also, would the code to check for new available cli versions based on the chart-supported versions go on a different PR?
For example, I am in 3.1, cluster supports 3.2 and 3.1, and last official CLI is 3.3, I would get a warning to upgrade to CLI 3.2. But if I am on 3.2, I don't get a warning to upgrade because my cluster doesn't support CLI 3.3

pchico83:
> Screenshot 2024-11-07 at 10 11 29 AM
> cc @pchico83

What about something like this?
```
x Unsupported Okteto CLI version: 3.0.0.
i  Your Okteto instance supports Okteto CLI versions: xxx, yyy.
   Please update your Okteto CLI or contact your Okteto administrator.
```
The first one is error, and the second one is info message.
In general, it's important that messages don't exceed 80 chars and they are properly indented after new lines

pkg/types/cluster_metadata.go
```diff
@@ -24,5 +24,6 @@ type ClusterMetadata struct {
 	SSHAgentHostname    string
 	SSHAgentPort        string
 	Certificate         []byte
+	CliVersionRange     []string
```
jLopezbarb, line 27: What happens if a backend doesn't return the CliversionRange? I think the CLI will fail in that case

cmd/context/create.go
```diff
@@ -282,6 +283,33 @@ func (c *Command) initOktetoContext(ctx context.Context, ctxOptions *Options) er
 	okteto.GetContext().IsTrial = clusterMetadata.IsTrialLicense
 	okteto.GetContext().CompanyName = clusterMetadata.CompanyName
 
+	if len(clusterMetadata.CliVersionRange) > 0 {
+		skipCheck, ok := os.LookupEnv("OKTETO_SKIP_CLUSTER_CLI_VERSION")
```
jLopezbarb, line 287: Could we use `env.LoadBoolean()`? It covers all the possible true values

cmd/context/create.go
```diff
@@ -282,6 +283,33 @@ func (c *Command) initOktetoContext(ctx context.Context, ctxOptions *Options) er
 	okteto.GetContext().IsTrial = clusterMetadata.IsTrialLicense
 	okteto.GetContext().CompanyName = clusterMetadata.CompanyName
 
+	if len(clusterMetadata.CliVersionRange) > 0 {
```
jLopezbarb, line 286: Can we move this into a new function to make it testable?

cmd/context/create.go
andreafalzetti: +1 also for consistency with other feature toggles we use booleans instead of integers

cmd/context/create.go
andreafalzetti, line 286: I was testing with a local build of this branch, and with the cluster updated with this feature where I've set that I can only use the latest CLI. Should I get some warning in that case considering that the version is the commit?

I was adding some logs to see if the version that I set in the backend matches the verison that the CLI receives, simply adding `fmt.Println("CliVersionRange", clusterMetadata.CliVersionRange)`, but I always get an empty array. I've checked the query and looks good. Can you help me spot why it's logging an empty array? What am I missing?

ifbyol (approved):
I have been testing it and it looks great! I'm approving it from a functionality point of view, but I'd wait for a confirmation on a couple of comments I added related to some messaging

cmd/context/create.go
```diff
@@ -282,13 +283,42 @@ func (c *Command) initOktetoContext(ctx context.Context, ctxOptions *Options) er
...
 
 	os.Setenv(model.OktetoUserNameEnvVar, okteto.GetContext().Username)
 
 	return nil
 }
 
+func checkCLIMinVersion(currentVersion, minVersion string) error {
+	version, err := semver.NewVersion(currentVersion)
+	if err != nil {
+		oktetoLog.Warning("You are using a non-standard okteto version (%s) that may be incompatible with your okteto cluster. Set OKTETO_SKIP_CLUSTER_CLI_VERSION=1 to supress this message.", currentVersion)
```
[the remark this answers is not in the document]
pchico83: Right, we use booleans in other envvars.
By the way, with the new functionality (only limit min version), should we rename the envvar to `OKTETO_SKIP_CLUSTER_MINIMUM_CLI_VERSION`?

cmd/context/create.go
```diff
@@ -282,13 +283,42 @@ func (c *Command) initOktetoContext(ctx context.Context, ctxOptions *Options) er
...
+		return nil
+	}
+	min, err := semver.NewVersion(minVersion)
+	if err != nil {
+		return fmt.Errorf("failed to parse cluster min version: %v", err)
+	}
+	if version.LessThan(min) {
+		return oktetoErrors.UserError{
+			E:    fmt.Errorf("unsupported okteto CLI version: %s", currentVersion),
+			Hint: fmt.Sprintf("Your Okteto instance supports a minimum Okteto CLI version of %s.\n    Please update your Okteto CLI to the latest version or contact your Okteto administrator.", minVersion),
```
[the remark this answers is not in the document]
pchico83: Yes, I would say the CLI version package with the cluster.
Same thing for the warning to upgrade, let's show it based in the CLI version package with the cluster

cmd/context/create.go
```diff
@@ -282,13 +283,68 @@ func (c *Command) initOktetoContext(ctx context.Context, ctxOptions *Options) er
...
+		}
+	}
+
+	if version.LessThan(recVersion) {
+		oktetoLog.Warning(fmt.Sprintf("Your okteto version %s is older than the recommended version of your okteto instance: %s. Upgrade to the recommended version or set OKTETO_SKIP_CLUSTER_CLI_VERSION=true to supress this message.", currentMajorMinor, recommendedMajorMinor))
+	}
+
+	if version.GreaterThan(recVersion) {
+		oktetoLog.Warning(fmt.Sprintf("Your okteto version %s is newer than the recommended version of your okteto instance: %s. Downgrade to the recommended version or set OKTETO_SKIP_CLUSTER_CLI_VERSION=true to supress this message.", currentMajorMinor, recommendedMajorMinor))
+		return nil
```
ifbyol, line 340: @codyjlandstrom @pchico83 Do we want to show a warning when a user is using a higher version of the CLI to ask them to downgrade?

cmd/context/create.go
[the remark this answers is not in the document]
ifbyol: @codyjlandstrom  @pchico83 should we do it even for newer versions asking to downgrade? We talked about it yesterday in the daly and we had some doubts about that specific case.

jLopezbarb (approved):

cmd/context/create.go
```diff
@@ -282,13 +283,68 @@ func (c *Command) initOktetoContext(ctx context.Context, ctxOptions *Options) er
...
+	}
+	if version.LessThan(min) {
+		return oktetoErrors.UserError{
+			E:    fmt.Errorf("unsupported okteto CLI version: %s", currentVersion),
+			Hint: fmt.Sprintf("Your Okteto instance has a recommended Okteto CLI version of %s and supports a minimum version of %s.\n    Please update your Okteto CLI or contact your Okteto administrator.", recommendedMajorMinor, minMajorMinor),
+		}
+	}
+
+	if version.LessThan(recVersion) {
+		oktetoLog.Warning(fmt.Sprintf("Your okteto version %s is older than the recommended version of your okteto instance: %s. Upgrade to the recommended version or set OKTETO_SKIP_CLUSTER_CLI_VERSION=true to supress this message.", currentMajorMinor, recommendedMajorMinor))
```
codyjlandstrom, line 335:
```suggestion
		oktetoLog.Warning(fmt.Sprintf("Your Okteto CLI version %s is older than the recommended version of your Okteto instance: %s. Upgrade to the recommended version or set OKTETO_SKIP_CLUSTER_CLI_VERSION=true to suppress this message.", currentMajorMinor, recommendedMajorMinor))
```

cmd/context/create.go
```diff
@@ -282,13 +283,68 @@ func (c *Command) initOktetoContext(ctx context.Context, ctxOptions *Options) er
...
+	}
+
+	if version.GreaterThan(recVersion) {
+		oktetoLog.Warning(fmt.Sprintf("Your okteto version %s is newer than the recommended version of your okteto instance: %s. Downgrade to the recommended version or set OKTETO_SKIP_CLUSTER_CLI_VERSION=true to supress this message.", currentMajorMinor, recommendedMajorMinor))
```
codyjlandstrom, line 339:
```suggestion
		oktetoLog.Warning(fmt.Sprintf("Your Okteto CLI version %s is newer than the recommended version of your Okteto instance: %s. Downgrade to the recommended version or set OKTETO_SKIP_CLUSTER_CLI_VERSION=true to suppress this message.", currentMajorMinor, recommendedMajorMinor))
```

PR #4561, opened 2024-11-11
fix: error unmarshalling

jLopezbarb:
# Proposed changes

Fixes DEV-776

Umarhsal doesn't handle the strict field in the subsequent unmarshalls and we need to do the check for the extra fields in the custom serialisation

Also detected a tests that was considering the bug as a future

## How to validate

1. gh repo clone okteto/python-docker-compose
2. cd python-docker-compose/
3. git checkout a08b4418ca5781214486291948f31c01be5b2987
4. okteto deploy

## CLI Quality Reminders 🔧

### DIFF
--- a/pkg/model/serializer.go
+++ b/pkg/model/serializer.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"reflect"
 	"runtime"
 	"sort"
 	"strconv"
@@ -33,6 +34,7 @@ import (
 	"github.com/okteto/okteto/pkg/externalresource"
 	oktetoLog "github.com/okteto/okteto/pkg/log"
 	"github.com/okteto/okteto/pkg/model/forward"
+	"gopkg.in/yaml.v3"
 	apiv1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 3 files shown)


ifbyol (commented):
From a functionality perspective, looks good- @codyjlandstrom @pchico83 What do you think about the error message?

Screenshot 2024-11-12 at 09 57 15

Maybe we could add a link to the Okteto manifest docs? 

One important thing to bear in mind is that, with these changes, we not longer show the former validation message, which I think it looked better (see following screenshot):

Screenshot 2024-11-12 at 10 07 05

There we specify the line where the unknown field is located and we include a link to the documentation. @jLopezbarb How hard would be to have the same error message, but taking into account the `forward` scenario?

ifbyol:
@jLopezbarb Also, several e2e are broken due to these changes. I honestly don't think all of them use "bad" manifests, but it could be. So, or we have broken something in the logic, or we need to update those tests to use good manifests

pkg/model/serializer.go
```diff
@@ -822,26 +828,52 @@ func (m *Manifest) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
+	}
+
+	if len(unknownFields) > 0 {
+		return fmt.Errorf("unknown fields in manifest: %v", unknownFields)
 	}
-	*d = DeployCommand(extendedCommand)
 	return nil
 }
 
+func getStructFields(iface interface{}) map[string]struct{} {
```
andreafalzetti, line 856: an implementation of this, already exists in `model` (`getStructKeys`)

andreafalzetti: https://github.com/okteto/okteto/blob/master/pkg/model/schema.go#L43

PR #4563, opened 2024-11-14
Divert using services with external names

maroshii:
Instead of pseuso cloning the service and tinkering with the endpoints, swap the ingress services for services with `type: ExternalName` and point to the parent service.

This change also allows to remove all the webhook mutation logic for endpoints in the okteto backend


pchico83:
@maroshii this is great!! much simpler to maintain and to explain to our users :raised_hands::raised_hands::raised_hands:
I have been testing it and I face a few issues:
- When following this sample https://github.com/okteto/movies-catalog, and I am going to deploy https://github.com/okteto/movies-rentals, I get an error:
```
 x  Error diverting ingress 'pchico83/catalog' service 'catalog': admission webhook "service.webhook.okteto.com" denied the request: service pchico83/catalog doesn't have published ports
```
We probably need to change the validation in our webhook to skip ExternalNames
- All the ingresses in the developer namespace should add an annotation to inject the divert header
- All ingresses/services created by divert, should have a label `dev.okteto.com/host: true`. We will use this label to show them in the UI in a specific way to visualize divert

jLopezbarb (approved):

pchico83:
> > @maroshii this is great!! much simpler to maintain and to explain to our users 🙌🙌🙌 I have been testing it and I face a few issues:
> > 
> > * When following this sample https://github.com/okteto/movies-catalog, and I am going to deploy https://github.com/okteto/movies-rentals, I get an error:
> > 
> > ```
> >  x  Error diverting ingress 'pchico83/catalog' service 'catalog': admission webhook "service.webhook.okteto.com" denied the request: service pchico83/catalog doesn't have published ports
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > We probably need to change the validation in our webhook to skip ExternalNames
> > 
> > * All the ingresses in the developer namespace should add an annotation to inject the divert header
> > * All ingresses/services created by divert, should have a label `dev.okteto.com/host: true`. We will use this label to show them in the UI in a specific way to visualize divert
> 
> @pchico83 there are two chart settings that are relevant for this change:
> 
> ```yaml
> externalNames:
>   enabled: true
> 
> convertLoadBalancedServices:
>   enabled: false
> ```
> 
> The first one, is requires for this to work. The second, is required or services are forced to `type=ClusterIP` iirc. TBD what we want to do with them but that's how it is right now and we need to review.

@maroshii as part of this iteration, I would do the following:
- Default `externalNames.enabled` to true. I think we disabled them by default to address a vulnerability in Okteto Cloud (cc: @jpf-okteto)
- Update the `convertLoadBalancedServices` logic to ignore service ExternalNames and leave them as they are.
WDYT?

pkg/divert/nginx/services.go
```diff
@@ -81,20 +62,16 @@ func translateService(name, namespace string, s *apiv1.Service) *apiv1.Service {
...
 	labels.SetInMetadata(&result.ObjectMeta, model.DeployedByLabel, format.ResourceK8sMetaString(name))
-	// create a headless service pointing to an endpoints object that resolves to service cluster ip in the diverted namespace
-	result.Spec.ClusterIP = apiv1.ClusterIPNone
-	result.Spec.ClusterIPs = nil
-	result.Spec.Selector = nil
 	if result.Annotations == nil {
 		result.Annotations = map[string]string{}
 	}
 	result.Annotations[model.OktetoAutoCreateAnnotation] = "true"
+	result.Annotations[model.OktetoDivertedSourceAnnotation] = s.Namespace
```
ifbyol, line 75: Just curiosity, why is it needed in the service? I see that in the backend we are only checking the annotation for the ingress.

ifbyol (commented):
Overall looks good, but I'd wait for Pablo's approval as he is more familiar with the feature

pchico83:
@maroshii are you adding the header injection in this PR? or a separate one? is it ready for final review?

pkg/divert/nginx/services.go
ifbyol: 👍  Ok, thanks for the clarification. No problem on adding it, just curiosity

pchico83:
> @pchico83 in the last commit I added new changes so that we always update the ingresses and remove the equality check as well as the auto-create annotation. The ingress deployed in the child namespace was not being updated correctly

The cache is very useful in large environments, like Katana (100 ingresses), where updating 100 ingresses can take several minutes everytime you run `okteto deploy`, while only the first time there are changes to do. This is a common scenario because divert is meant for the case of having many services.
Can we investigate why it's not updated properly?
The component deployed in the okteto manifest (rentals in this case), should create its ingress/service as part of the okteto manifest commands. The only thing we should add is the header injection in the "rentals" ingress. All other ingresses/services are "copied" from the parent namespace

pkg/model/const.go
```diff
@@ -126,6 +126,11 @@ const (
 	RemoteSubPath = "okteto-remote"
 	// OktetoAutoCreateAnnotation indicates if the deployment was auto generated by okteto up
 	OktetoAutoCreateAnnotation = "dev.okteto.com/auto-create"
+	// OktetoDivertedSourceAnnotation indicates the namespace source of the diverted resource (usually a service or an ingress)
+	// The value of this annotation is the namespace of the original resource
+	OktetoDivertedSourceAnnotation = "divert.okteto.com/source-namespace"
```
pchico83, line 131: To keep it shorter and consistent with the okteto manifest divert section, I would rename this one to `divert.okteto.com/namespace`

pchico83 (approved):

PR #4565, opened 2024-11-15
Configure e2e tests to run in product staging

pchico83:
The main changes are in the first commit; the second commit just applies a refactor in the `GetTestNamespace` function.

The refactor is to generate shorter namespace names. Otherwise, we hit the 63 limit for application endpoints. As part of it, I am not adding the okteto username, I don't think it provides value and adds 18 extra chars.

Tests run 30% on average, and it feels like flakiness improves a lot, too.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -41,9 +41,10 @@ executors:
     docker:
       - image: okteto/golang-ci:2.7.0
     environment:
-      OKTETO_CONTEXT: https://staging.okteto.dev
-      OKTETO_USER: cindylopez
-      OKTETO_APPS_SUBDOMAIN: staging.okteto.dev
+      OKTETO_CONTEXT: https://okteto.staging.dev.okteto.net
+      OKTETO_USER: staging-okteto-bot
+      OKTETO_APPS_SUBDOMAIN: staging.dev.okteto.net
+      OKTETO_NAMESPACE_PREFIX: staging
 
 jobs:
   golangci-lint:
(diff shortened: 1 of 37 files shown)


.circleci/config.yml
```diff
@@ -41,9 +41,10 @@ executors:
     docker:
       - image: okteto/golang-ci:2.7.0
     environment:
-      OKTETO_CONTEXT: https://staging.okteto.dev
-      OKTETO_USER: cindylopez
-      OKTETO_APPS_SUBDOMAIN: staging.okteto.dev
+      OKTETO_CONTEXT: https://okteto.staging.dev.okteto.net
```
pchico83, line 44: When migration is finished, this will be https://staging.okteto.dev again. But we need this change while both environments are running

.circleci/config.yml
```diff
@@ -41,9 +41,10 @@ executors:
...
+      OKTETO_USER: staging-okteto-bot
+      OKTETO_APPS_SUBDOMAIN: staging.dev.okteto.net
+      OKTETO_NAMESPACE_PREFIX: staging
```
pchico83, line 47: This is because all namespaces in product staging are prefixed with `staging-`

.circleci/config.yml
```diff
@@ -41,9 +41,10 @@ executors:
     docker:
       - image: okteto/golang-ci:2.7.0
     environment:
-      OKTETO_CONTEXT: https://staging.okteto.dev
-      OKTETO_USER: cindylopez
-      OKTETO_APPS_SUBDOMAIN: staging.okteto.dev
+      OKTETO_CONTEXT: https://okteto.staging.dev.okteto.net
+      OKTETO_USER: staging-okteto-bot
```
ifbyol, line 45: The user would be `okteto-bot`, we are not prefixing the okteto bot user with the release name.

Also, in staging we had configure `cindylopez` to allow 300 namespaces. Would it be needed for this user too?

ifbyol (approved):

.circleci/config.yml
pchico83: The user is `staging-okteto-bot`, I have been testing it. It could be because of the `namespacePrefix` properly.
Since `okteto-bot` doesn't have a limit for namespaces, we don't need to configure extra quotas

pchico83: Oh, that makes sense, let me try to remove it!

jLopezbarb (approved):

.circleci/config.yml
pchico83: Looking good, I love deleting code :-) 
Thanks for the tip!

ifbyol (approved):

PR #4566, opened 2024-11-15
Add command to generate Okteto Manifest JSON Schema

andreafalzetti:
This PR introduces a new command `generate-schema` that can output the Okteto Manifest JSON Schema.

In a follow-up PR I will add the generate schema in the repo.

## Considered approaches

1. **Manual writing the JSON Schema**
a. Manually writing JSON code is very error prone and the DevX is terrible. The schema file is about 1500 lines.
2. **Generate the JSON Schema from the Manifest struct with the `jsonschema` annotation**
a. This was the approach that I initially tried and it was working well, until I faced limitations with several of our manifest properties that have conditionals and different ways to declare things, this approach is too limited for our complex use case. 
3. **Create `pkg/schema` and create a command to generate the JSON schema** ✅
a. Despite this approach is more verbose, it's a guided and typed approach, making it safe, testable and easy to scale and keep up to date, writing unit tests, etc.

### How to validate

1. Build this branch
2. Run
```bash
okteto generate-schema -o schema.json
```
3. In VS Code, ensure you have the [YAML extension](https://marketplace.visualstudio.com/items?itemName=redhat.vscode-yaml) installed
4. Then, go into settings and add this:
```json
{
    "yaml.schemas": {
        "<path to okteto json schema>": [
            "okteto.yml",
            "okteto.yaml",
        ]
    }
}
```
5. Edit an okteto manifest and observe auto-completion and error highlighting for invalid properties


pkg/schema/schema_test.go
```diff
@@ -0,0 +1,554 @@
...
+
+	err = schema.Validate(obj)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func Test_Schema(t *testing.T) {
```
andreafalzetti, line 62: This test mentions other repos for sake of using real manifests, but I it could be using other manifests with different test names. The purpose is solely to have real manifests to be tested as whole, instead of testing each section invididualy.

It's not intended as integration test, but should be kept as unit and be used to validate changes in the schema pkg.

cmd/generate-schema.go
```diff
@@ -0,0 +1,63 @@
...
+				fmt.Print(string(json))
+				return nil
+			}
+			err = toFile(json, outputFilePath)
+
+			return err
+		},
+	}
+
+	cmd.Flags().StringVarP(&outputFilePath, "output-file", "o", "", "Path to the file where the json schema will be stored")
```
teresaromero, line 61: should this be cwd?

pkg/model/schema.go
```diff
@@ -39,8 +39,8 @@ func mergeAndSortUnique(slice1, slice2 []string) []string {
 	return result
 }
 
-// getStructKeys recursively goes through a given struct and returns a map of struct names to their fields
-func getStructKeys(t interface{}) map[string][]string {
+// GetStructKeys recursively goes through a given struct and returns a map of struct names to their fields
+func GetStructKeys(t interface{}) map[string][]string {
```
teresaromero, line 43: As we created a ´schema´ pkg, i believe this one now belongs to the schema pkg? i see the logic of extracting the keys as part of it... also might move the manifest friendy error logic too and exported from the schema pkg... wdyt?

pkg/externalresource/external_resource.go
```diff
@@ -33,9 +33,9 @@ type Section map[string]*ExternalResource
 
 // ExternalResource represents information on an external resource
 type ExternalResource struct {
-	Icon      string
-	Notes     *Notes
-	Endpoints []*ExternalEndpoint
+	Icon      string              `json:"icon,omitempty" yaml:"icon,omitempty"`
```
teresaromero, line 36: why i this necesary? i think the exported model is declared at the serializer ? https://github.com/okteto/okteto/blob/3f323291cb3dc189175c92cf38d6299a25ab2721/pkg/externalresource/serializer.go#L22

andreafalzetti: as all the other structs of the manifest, they are annotated for the marshallers/unmarshallers using the yaml/json annotations, as they were not annotated, they were ingored by GetStructKeys

cmd/generate-schema.go
andreafalzetti: what do you mean? 🤔

pkg/model/schema.go
andreafalzetti: @teresaromero I tried but as they are very related, it creates a cycle import between `pkg/model` and `pkg/schema`

cmd/generate-schema.go
andreafalzetti: changed to `-`

andreafalzetti:
As discussed with the team offline I am trying to wrap up this work, so I will be merging for now and if any additional feedback is left on this PR, I will make sure to include it in follow-up PRs. cc @teresaromero

PR #4567, opened 2024-11-18
Simplify "make integration-actions" to reduce their execution time

pchico83:
As part of [this PR in the okteto/app repo](https://github.com/okteto/app/pull/8885), we will start running `make integration-actions` on every PR. In order to not slow down PR test executions, I have tried to speed up the execution of these tests.

Changes include:
- Deploy go-getting-started instead of movies. We only want to test the action works.
- Remove "apply" test, this action is deprecated
- Remove "TestPipelineActionsWithCompose". I don't think this tests provides value over "TestPipelineActions". We are only testing the action interface

Pending changes on follow up PRs could be:
- Run action tests in parallel
- Implement a test for the "okteto/test" GH action


andreafalzetti:
I assume you run the tests locally. FYI we can also use labels `run-e2e` to validate the changes and have logs of passing test, I am adding the label to the PR

PR #4572, opened 2024-11-19
Update error when image does not exist on `okteto deploy`

teresaromero:
# Proposed changes

Fixes https://okteto.atlassian.net/browse/DEV-711
Fixes https://okteto.atlassian.net/browse/PROD-229

When running `okteto deploy` with a given manifest that includes an image field on the deploy section, if the image was not found, we were throwing a generic error.

This PR addresses not-found images, as well as the change of the error message when the image is pull access denied.
Included under the same PR the changes relative to include the docs for private registry credentials. Testing this changes came across a bug where the userError was not being displayed correctly, as the output was being overriden by an empty hint user error.

EDIT: 
- added a refactor regarding the regex extracting the image from the message and unified also cases for errors with same output to the user.
- added different hints in case image used is the official cli image or a fork of it

## How to validate

Build binary and run `okteto deploy` with different combinations of images at the deploy section of the manifest

Output 

```
❯ ok deploy
 i  Using - @ - as context
 x  Error deploying application: the image 'docker.io/okteto/okteto:xxxx' is not accessible or it does not exist
    Please verify the name of the image 'docker.io/okteto/okteto:xxxx' to make sure it exists.
    When using private registries, make sure Okteto Registry Credentials are correctly configured.
    See more at: https://www.okteto.com/docs/admin/registry-credentials/
```

## CLI Quality Reminders 🔧

### DIFF
--- a/cmd/deploy/deploy.go
+++ b/cmd/deploy/deploy.go
@@ -450,8 +450,11 @@ func (dc *Command) Run(ctx context.Context, deployOptions *Options) error {
 	oktetoLog.AddToBuffer(oktetoLog.InfoLevel, "EOF")
 
 	if err != nil {
-		if err == oktetoErrors.ErrIntSig {
+		if errors.Is(err, oktetoErrors.ErrIntSig) {
 			return nil
+		}
+		if errors.As(err, &oktetoErrors.UserError{}) {
+			return err
 		}
 		err = oktetoErrors.UserError{E: err}
 		data.Status = pipeline.ErrorStatus
--- a/cmd/deploy/deploy.go
+++ b/cmd/deploy/deploy.go
@@ -453,10 +453,10 @@ func (dc *Command) Run(ctx context.Context, deployOptions *Options) error {
 		if errors.Is(err, oktetoErrors.ErrIntSig) {
 			return nil
 		}
-		if errors.As(err, &oktetoErrors.UserError{}) {
-			return err
+		// transform internal errors to user errors
+		if !errors.As(err, &oktetoErrors.UserError{}) {
+			err = oktetoErrors.UserError{E: err}
 		}
-		err = oktetoErrors.UserError{E: err}
 		data.Status = pipeline.ErrorStatus
 	} else {
 		// This has to be set only when the command succeeds for the case in which the deploy is executed within an
--- a/pkg/build/buildkit/errors.go
+++ b/pkg/build/buildkit/errors.go
@@ -20,8 +20,6 @@ import (
 	"strings"
 
 	oktetoErrors "github.com/okteto/okteto/pkg/errors"
-	"github.com/okteto/okteto/pkg/okteto"
-	"github.com/okteto/okteto/pkg/registry"
 	"google.golang.org/grpc/status"
 )
 
@@ -50,40 +48,28 @@ func (e CommandErr) Error() string {
 }
 
 // GetErrorMessage returns the parsed error message
-func GetErrorMessage(err error, tag string) error {
+func GetSolveErrorMessage(err error) error {
 	if err == nil {
 		return nil
 	}
 
-	imageCtrl := registry.NewImageCtrl(okteto.Config{})
-	imageRegistry, imageTag := imageCtrl.GetRegistryAndRepo(tag)
+	imageFromError := extractImageFromError(err)
 	switch {
-	case isLoggedIntoRegistryButDontHavePermissions(err):
-		err = oktetoErrors.UserError{
-			E:    fmt.Errorf("failed to push image '%s': You are not authorized to push image '%s'", tag, imageTag),
-			Hint: fmt.Sprintf("Please log in into the registry '%s' with a user with push permissions to '%s' or use another image.", imageRegistry, imageTag),
-		}
-	case isNotLoggedIntoRegistry(err):
-		err = oktetoErrors.UserError{
-			E:    fmt.Errorf("failed to push image '%s': You are not authorized to push image '%s'", tag, imageTag),
-			Hint: fmt.Sprintf("Log in into the registry '%s' and verify that you have permissions to push the image '%s'.", imageRegistry, imageTag),
-		}
 	case isBuildkitServiceUnavailable(err):
 		err = oktetoErrors.UserError{
 			E:    fmt.Errorf("buildkit service is not available at the moment"),
 			Hint: "Please try again later.",
 		}
-	case isPullAccessDenied(err) || isNotFound(err):
-		if imageTag == "" && isPullAccessDenied(err) {
-			imageTag = extractImageTagFromPullAccessDeniedError(err)
-		} else if imageTag == "" && isNotFound(err) {
-			imageTag = extractImageTagFromNotFoundError(err)
-		}
+	case isLoggedIntoRegistryButDontHavePermissions(err) ||
+		isNotLoggedIntoRegistry(err) ||
+		isPullAccessDenied(err) ||
+		isNotFound(err) ||
+		isHostNotFound(err):
 		err = oktetoErrors.UserError{
-			E: fmt.Errorf("the image '%s' is not accessible or it does not exist", imageTag),
-			Hint: fmt.Sprintf(`Please verify the name of the image '%s' to make sure it exists.
+			E: fmt.Errorf("the image '%s' is not accessible or it does not exist", imageFromError),
+			Hint: `Please verify the name of the image to make sure it exists.
     When using private registries, make sure Okteto Registry Credentials are correctly configured.
-    See more at: https://www.okteto.com/docs/admin/registry-credentials/`, imageTag),
+    See more at: https://www.okteto.com/docs/admin/registry-credentials/`,
 		}
 	default:
 		var cmdErr CommandErr
@@ -91,30 +77,22 @@ func GetErrorMessage(err error, tag string) error {
 			return cmdErr
 		}
 		err = oktetoErrors.UserError{
-			E: fmt.Errorf("error building image '%s': %w", tag, err),
+			E: err,
 		}
 	}
 	return err
 }
 
 const (
-	regexForImageTag = `([a-zA-Z0-9\.\/_-]+(:[a-zA-Z0-9-]+)?)`
+	// regexForImageFromFailedToSolveErr is the regex to extract the image from an error message
+	// buildkit solve errors provide the image name between :
+	regexForImageFromFailedToSolveErr = `: ([a-zA-Z0-9\.\/_-]+(:[a-zA-Z0-9-]+)?):`
 )
 
-func extractImageTagFromPullAccessDeniedError(err error) string {
-	re := regexp.MustCompile(regexForImageTag + `: pull access denied`)
-	matches := re.FindStringSubmatch(err.Error())
-	if len(matches) > 1 {
-		return matches[1]
-	}
-	return ""
-}
-
-func extractImageTagFromNotFoundError(err error) string {
-	re := regexp.MustCompile(regexForImageTag + `: not found`)
-	matches := re.FindStringSubmatch(err.Error())
-	if len(matches) > 1 {
-		return matches[1]
+func extractImageFromError(err error) string {
+	re := regexp.MustCompile(regexForImageFromFailedToSolveErr)
+	if matchs := re.FindStringSubmatch(err.Error()); len(matchs) > 1 {
+		return matchs[1]
 	}
 	return ""
 }
@@ -183,4 +161,9 @@ func isPullAccessDenied(err error) bool {
 // IsNotFound returns true when the error is because the resource is not found
 func isNotFound(err error) bool {
 	return strings.Contains(err.Error(), "not found")
+}
+
+// isHostNotFound returns true when the error is because the host is not found
+func isHostNotFound(err error) bool {
+	return strings.Contains(err.Error(), "failed to do request") && strings.Contains(err.Error(), "no such host")
 }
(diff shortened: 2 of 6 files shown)


pkg/build/buildkit/errors.go
```diff
@@ -73,13 +73,15 @@ func GetErrorMessage(err error, tag string) error {
...
+		if imageTag == "" && isPullAccessDenied(err) {
 			imageTag = extractImageTagFromPullAccessDeniedError(err)
+		} else if imageTag == "" && isNotFound(err) {
+			imageTag = extractImageTagFromNotFoundError(err)
 		}
 		err = oktetoErrors.UserError{
-			E:    fmt.Errorf("failed to pull image '%s'. The repository is not accessible or it does not exist", imageTag),
-			Hint: fmt.Sprintf("Please verify the name of the image '%s' to make sure it exists.", imageTag),
+			E:    fmt.Errorf("the image '%s' is not accessible or it does not exist", imageTag),
+			Hint: fmt.Sprintf("Please verify the name of the image '%s' to make sure it exists.\nWhen using private registries, Okteto Registry Credentials are correctly configured https://www.okteto.com/docs/admin/registry-credentials/", imageTag),
```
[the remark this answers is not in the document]
ifbyol: In other places we are just using golang multiline strings with the character "`". Look for other samples in the code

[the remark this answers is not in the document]
ifbyol: @codyjlandstrom Could you review the message, please? It sounds a bit odd to me. I would say something like: `When using private registries, make sure Okteto Registry Credentials are correctly configured https://www.okteto.com/docs/admin/registry-credentials/` (notice the "make sure" addition in the message)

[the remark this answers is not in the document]
teresaromero: i tried with `` and tabs / spaces but also was not getting the exact point. i will take a look.

regarding the copy i think i eat it with my coffee because clearly there is something missing :D thanks

pkg/build/buildkit/errors.go
```diff
@@ -94,7 +96,16 @@ func GetErrorMessage(err error, tag string) error {
...
+	re := regexp.MustCompile(`([a-zA-Z0-9\.\/_-]+(:[a-zA-Z0-9]+)?): pull access denied`)
+	matches := re.FindStringSubmatch(err.Error())
+	if len(matches) > 1 {
+		return matches[1]
+	}
+	return ""
+}
+
+func extractImageTagFromNotFoundError(err error) string {
+	re := regexp.MustCompile(`([a-zA-Z0-9\.\/_-]+(:[a-zA-Z0-9]+)?): not found`)
```
ifbyol, line 108: I think it is not getting properly the image name.

With the following `image` field in an Okteto manifest `ifbyol/pipeline-runner:go-getting-started`, it is printing the following error message

Screenshot 2024-11-20 at 09 39 06

So it seems it is only getting the tag of the image?

pkg/build/buildkit/errors.go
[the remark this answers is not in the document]
teresaromero: Updated as in description, I added a third line with the link i saw it too long for output, let me know any other copy change!

pkg/build/buildkit/errors.go
teresaromero: yup, the expression for capturing the tag was not correct, fixed and added tests.

ifbyol:
I'm testing different scenarios and we are getting different errors that I believe we should also change.

------
One case failing. I set the environment variable `OKTETO_REMOTE_CLI_IMAGE` to a non-existing host (it happens the same if I use that image in the `image` field of the Okteto Manifest), for example to `non-existing.okteto.dev/ifbyol/cli:debug-4` and I execute a remote deploy, I get the following error:

```
 x  Error deploying application: error building image '': build failed: failed to solve: non-existing.okteto.dev/ifbyol/cli:debug-4: failed to do request: Head "https://non-existing.okteto.dev/v2/ifbyol/cli/manifests/debug-4": dial tcp: lookup non-existing.okteto.dev on xx.xxxx.x.xx:xx: no such host
```

We shouldn't say that is failing building the image. It should say that it doesn't exist or cannot be found

-----

Another case failing, I was just able to reproduce it indicating an Okteto Registry of another Okteto instance, so it doesn't have the right credentials to pull the image, so I guess it is due to some string matching or something, but the error we get in that scenario is:

```
 x  Error deploying application: failed to push image '': You are not authorized to push image ''
    Log in into the registry 'docker.io' and verify that you have permissions to push the image ''.
```

I don't know why the message say is trying to push an image because the flow doesn't trigger any image push.

teresaromero:
> I'm testing different scenarios and we are getting different errors that I believe we should also change.
> 
> One case failing. I set the environment variable `OKTETO_REMOTE_CLI_IMAGE` to a non-existing host (it happens the same if I use that image in the `image` field of the Okteto Manifest), for example to `non-existing.okteto.dev/ifbyol/cli:debug-4` and I execute a remote deploy, I get the following error:
> 
> ```
>  x  Error deploying application: error building image '': build failed: failed to solve: non-existing.okteto.dev/ifbyol/cli:debug-4: failed to do request: Head "https://non-existing.okteto.dev/v2/ifbyol/cli/manifests/debug-4": dial tcp: lookup non-existing.okteto.dev on xx.xxxx.x.xx:xx: no such host
> ```
> 
> We shouldn't say that is failing building the image. It should say that it doesn't exist or cannot be found
> 
> Another case failing, I was just able to reproduce it indicating an Okteto Registry of another Okteto instance, so it doesn't have the right credentials to pull the image, so I guess it is due to some string matching or something, but the error we get in that scenario is:
> 
> ```
>  x  Error deploying application: failed to push image '': You are not authorized to push image ''
>     Log in into the registry 'docker.io' and verify that you have permissions to push the image ''.
> ```
> 
> I don't know why the message say is trying to push an image because the flow doesn't trigger any image push.

i will have a look to those and check, probably we have not captured correctly the errors

ifbyol (commented):
@codyjlandstrom @pchico83 In general, this is an improvement, but there is a case in which I still think it is really hard for the user to know what is happening.

If I execute an `okteto deploy --remote` and the CLI image is not found, we are indicating that the image is not found or it doesn't exist, but it is hard to know why that image is needed and what I need to do to unblock myself. An example of error we return (I defined the env var `OKTETO_REMOTE_CLI_IMAGE` for the CLI image)

```
ok deploy --remote
 i  Using ifbyol-okteto-admin @ xxxxxx.okteto.net as context
 x  Error deploying application: the image 'ifyol/pipeline-runner:go-getting-started' is not accessible or it does not exist
    Please verify the name of the image to make sure it exists.
    When using private registries, make sure Okteto Registry Credentials are correctly configured.
    See more at: https://www.okteto.com/docs/admin/registry-credentials/
```

How might a user know what they would need to do?

pchico83:
> @codyjlandstrom @pchico83 In general, this is an improvement, but there is a case in which I still think it is really hard for the user to know what is happening.
> 
> If I execute an `okteto deploy --remote` and the CLI image is not found, we are indicating that the image is not found or it doesn't exist, but it is hard to know why that image is needed and what I need to do to unblock myself. An example of error we return (I defined the env var `OKTETO_REMOTE_CLI_IMAGE` for the CLI image)
> 
> ```
> ok deploy --remote
>  i  Using ifbyol-okteto-admin @ xxxxxx.okteto.net as context
>  x  Error deploying application: the image 'ifyol/pipeline-runner:go-getting-started' is not accessible or it does not exist
>     Please verify the name of the image to make sure it exists.
>     When using private registries, make sure Okteto Registry Credentials are correctly configured.
>     See more at: https://www.okteto.com/docs/admin/registry-credentials/
> ```
> 
> How might a user know what they would need to do?

+1 for improving the message. If `OKTETO_REMOTE_CLI_IMAGE` is our official image, we should tell the user to check if they have access to dockerhub and redirect to our airgap docs. If it's our official image but hosted in the user registry, we should also link to the airgap docs, but mention if they migrated this particular version to their registry.

pkg/build/buildkit/errors.go
```diff
@@ -49,54 +47,51 @@ func (e CommandErr) Error() string {
...
-	matches := re.FindStringSubmatch(err.Error())
-	if len(matches) > 1 {
+const (
+	// regexForImageFromFailedToSolveErr is the regex to extract the image from an error message
+	// buildkit solve errors provide the image name between :
+	regexForImageFromFailedToSolveErr = `: ([a-zA-Z0-9\.\/_-]+(:[a-zA-Z0-9-]+)?):`
+)
+
+func extractImageFromError(err error) string {
+	re := regexp.MustCompile(regexForImageFromFailedToSolveErr)
```
jLopezbarb, line 93: Should we move this to the var section? That way its already precomputed

pkg/build/buildkit/errors.go
```diff
@@ -49,54 +47,51 @@ func (e CommandErr) Error() string {
...
-		}
+	case isLoggedIntoRegistryButDontHavePermissions(err) ||
+		isNotLoggedIntoRegistry(err) ||
+		isPullAccessDenied(err) ||
+		isNotFound(err) ||
+		isHostNotFound(err):
 		err = oktetoErrors.UserError{
-			E:    fmt.Errorf("failed to pull image '%s'. The repository is not accessible or it does not exist", imageTag),
-			Hint: fmt.Sprintf("Please verify the name of the image '%s' to make sure it exists.", imageTag),
+			E: fmt.Errorf("the image '%s' is not accessible or it does not exist", imageFromError),
```
jLopezbarb, line 69: We were distinguishing if the error was a permission problem vs not logged into the registry. Do we want to change that?

teresaromero: That is the case, yes

pkg/build/buildkit/errors.go
```diff
@@ -49,54 +47,83 @@ func (e CommandErr) Error() string {
...
+			Hint: `Please verify the name of the image to make sure it exists.
+    When using private registries, make sure Okteto Registry Credentials are correctly configured.
+    See more at: https://www.okteto.com/docs/admin/registry-credentials/`,
+		}
+
+		if isOktetoRemoteImage(imageFromError) {
+			err = oktetoErrors.UserError{
+				E: fmt.Errorf("the image '%s' is not accessible or it does not exist", imageFromError),
+				Hint: `Please verify you have access to dockerhub.
+    If you are using airgapped environment check our docs on how to use Okteto Remote in airgapped environments: [TBD]`,
```
teresaromero, line 96: probably link to docs of airgapped woulld be https://github.com/okteto/docs/pull/885, let me know what link should we use for now instead? @ifbyol @pchico83

pkg/build/buildkit/errors.go
```diff
@@ -49,54 +47,83 @@ func (e CommandErr) Error() string {
...
+			E: fmt.Errorf("the image '%s' is not accessible or it does not exist", imageFromError),
+			Hint: `Please verify the name of the image to make sure it exists.
+    When using private registries, make sure Okteto Registry Credentials are correctly configured.
+    See more at: https://www.okteto.com/docs/admin/registry-credentials/`,
+		}
+
+		if isOktetoRemoteImage(imageFromError) {
+			err = oktetoErrors.UserError{
+				E: fmt.Errorf("the image '%s' is not accessible or it does not exist", imageFromError),
+				Hint: `Please verify you have access to dockerhub.
```
pchico83, line 95: dockerhub -> Docker Hub

pkg/build/buildkit/errors.go
pchico83: Link will be available at https://www.okteto.com/docs/self-hosted/manage/air-gapped/

pkg/build/buildkit/errors.go
```diff
@@ -49,54 +47,85 @@ func (e CommandErr) Error() string {
...
-	case isPullAccessDenied(err):
-		if imageTag == "" {
-			imageTag = extractImageTagFromPullAccessDeniedError(err)
-		}
+	case isImageIsNotAccessibleErr(err):
 		err = oktetoErrors.UserError{
-			E:    fmt.Errorf("failed to pull image '%s'. The repository is not accessible or it does not exist", imageTag),
-			Hint: fmt.Sprintf("Please verify the name of the image '%s' to make sure it exists.", imageTag),
+			E: fmt.Errorf("the image '%s' is not accessible or it does not exist", imageFromError),
+			Hint: `Please verify the name of the image to make sure it exists.
```
teresaromero, line 87: @codyjlandstrom when possible please review this new errors and hints. thanks!

ifbyol:
@teresaromero It seems that when it fails the deploy operation due to some of these reasons, the dev environment gets in `progressing` status forever and its status is not updated to `Error` as I would expect, and as it happens in the latest stable version. Not sure if related to these changes.

How to reproduce:
* Clone `okteto/go-getting-started`
* Modify `okteto.yml` and change the deploy section to be
```yaml
deploy:
  image: okteto/okteto:no-exist
  commands:
  - kubectl apply -f k8s.yml
```
* If you execute `okteto deploy` with the compiled version of this branch, you can see how the dev environment is stuck in `progressing`. That doesn't happen if you run it with the latest stable version

cmd/deploy/deploy.go
```diff
@@ -450,10 +450,13 @@ func (dc *Command) Run(ctx context.Context, deployOptions *Options) error {
 	oktetoLog.AddToBuffer(oktetoLog.InfoLevel, "EOF")
 
 	if err != nil {
-		if err == oktetoErrors.ErrIntSig {
+		if errors.Is(err, oktetoErrors.ErrIntSig) {
 			return nil
 		}
-		err = oktetoErrors.UserError{E: err}
+		// transform internal errors to user errors
```
teresaromero, line 456: @ifbyol the bug was introduced here, i fixed now. I refactored this as we where transforming the err to UserError missing the hint, when i change that i returned the err instead of just overwrite it. with this last change is all working as expected

ifbyol (approved):

jLopezbarb (approved):

PR #4580, opened 2024-11-21
Implement use of OKTETO_DEV_PERSISTENT_VOLUME_SIZE

teresaromero:
# Proposed changes

Resolves https://okteto.atlassian.net/browse/PROD-253

When assigning a volume size to the pvc running `okteto up` the default value is overriden by the env value if present.

Default value is used when there is no definition of size at the manifest.

## How to validate

Build binary, add the variable to admin variables at the cluster with a value.
Run okteto up and check the value is the one set by the env, different than the 5Gi default one
Run okteto up with a manifest that has a defined size, the pvc should be the one on the manifest.

Added unit test for the new cases.

### DIFF
--- a/pkg/model/volumes.go
+++ b/pkg/model/volumes.go
@@ -15,6 +15,7 @@ package model
 
 import (
 	"fmt"
+	"os"
 	"path/filepath"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


pkg/model/volumes.go
```diff
@@ -25,6 +26,9 @@ import (
 
 const (
 	defaultVolumeSize = "5Gi"
+
+	// devPersistentVolumeEnabledEnvVar is the name of the environment variable to change the defaultVolumeSize value
+	devPersistentVolumeSizeEnvVar = "OKTETO_DEV_PERSISTENT_VOLUME_SIZE"
```
ifbyol, line 31: Could you send a PR to document it [here](https://www.okteto.com/docs/reference/feature-flags/), please?

ifbyol (approved):

PR #4582, opened 2024-11-21
Added runAsNonRoot field to Json Schema for dev.securityContext property

ifbyol:
# Proposed changes

Added fields `runAsNonRoot` and `AllowPrivilegeEscalation` for the `dev.securityContext` field in the Okteto Manifest. I realized that the schema wasn't supporting them, but the `okteto up` command is supporting them, as it does the translation correctly. But I see that in the docs we don't indicate those fields as supported so not fully sure. Could we confirm @pchico83 @jLopezbarb @codyjlandstrom ?


pchico83 (approved):
They are supported, but we forgot to document them

PR #4584, opened 2024-11-26
Remove endpoints from deprecated function resources

viddo:
"Functions" has been deprecated for some time and returns empty list, but removing usage here so we can clean up API eventually.

## How to validate

```shell
make build
OKTETO_CONTEXT=<SOME-CTX> ./bin/okteto endpoints
```

### DIFF
--- a/pkg/okteto/endpoint.go
+++ b/pkg/okteto/endpoint.go
@@ -43,7 +43,6 @@ type Component struct {
 type Space struct {
 	Deployments  []Component
 	Statefulsets []Component
-	Functions    []Component
 	Externals    []Component
 }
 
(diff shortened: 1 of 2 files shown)


teresaromero (approved):
Depending on https://github.com/okteto/app/pull/8908#pullrequestreview-2461140265, how would be the release and compatibility?

viddo:
> Depending on [okteto/app#8908 (review)](https://github.com/okteto/app/pull/8908#pullrequestreview-2461140265), how would be the release and compatibility?

Note that this change/PR is not _depending_ on that PR, or vice versa. They are _related_ in that this enables cleaning up the last remnants on the API that that PR touches on. But it cannot be finalized until this change has been merged and prev CLI versions have been deprecated (i.e. <= 3.1.0, assuming this would goes out in a 3.1.x/3.2.x). So I don't think there's nothing more to it in terms of release or compatibility here?

PR #4585, opened 2024-11-26
Fix liveness probe with depends_on relation

teresaromero:
# Proposed changes

Fixes https://okteto.atlassian.net/browse/DEV-260

When using `depends_on` conditions under a compose manifest, we are not translating this condition correctly into liveness probe.

Liveness probe is currently only assigned if the user uses a custom field under the healthcheck of a service. If other service depends on this one under the condition of healthy, the liveness probe is not used.

This changes propose that when a compose is using the healthy condition, this translate into a liveness condition of the kubernetes pod.

## How to validate

build the binary and run a deployment with compose services that depend on each other, for example the one on the issue

### DIFF
--- a/pkg/model/stack_serializer.go
+++ b/pkg/model/stack_serializer.go
@@ -371,6 +371,21 @@ func getAccessiblePorts(ports []PortRaw) []PortRaw {
 	return accessiblePorts
 }
 
+// setLivenessFromDependsOnServiceHealthy sets the liveness probe to true if the service depends on another service being healthy
+func setLivenessFromDependsOnServiceHealthy(svc *Service, svcName string, s *Stack) {
+	// loop for each service in the stack to find if some of it is dependant of svcName for being healthy
+	for _, depService := range s.Services {
+		// if dependant service has a service_healty dependency for svcName, then set liveness probe to true
+		if condition, ok := depService.DependsOn[svcName]; ok && condition.Condition == DependsOnServiceHealthy {
+			if svc.Healtcheck == nil {
+				svc.Healtcheck = &HealthCheck{}
+			}
+			svc.Healtcheck.Liveness = true
+			return
+		}
+	}
+}
+
 func (serviceRaw *ServiceRaw) ToService(svcName string, stack *Stack) (*Service, error) {
 	svc := &Service{}
 	var err error
@@ -397,6 +412,7 @@ func (serviceRaw *ServiceRaw) ToService(svcName string, stack *Stack) (*Service,
 	if serviceRaw.Healthcheck != nil && !serviceRaw.Healthcheck.Disable {
 		svc.Healtcheck = serviceRaw.Healthcheck
 		translateHealtcheckCurlToHTTP(svc.Healtcheck)
+		setLivenessFromDependsOnServiceHealthy(svc, svcName, stack)
 	}
 
 	svc.NodeSelector = serviceRaw.NodeSelector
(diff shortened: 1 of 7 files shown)


pkg/model/stack_serializer.go
```diff
@@ -371,6 +371,21 @@ func getAccessiblePorts(ports []PortRaw) []PortRaw {
 	return accessiblePorts
 }
 
+// setLivenessFromDependsOnServiceHealthy sets the liveness probe to true if the service depends on another service being healthy
+func setLivenessFromDependsOnServiceHealthy(svc *Service, svcName string, s *Stack) {
+	// loop for each service in the stack to find if some of it is dependent of svcName for being healthy
+	for _, depService := range s.Services {
+		// if dependent service has a service_healty dependency for svcName, then set liveness probe to true
+		if condition, ok := depService.DependsOn[svcName]; ok && condition.Condition == DependsOnServiceHealthy {
+			if svc.Healtcheck == nil {
```
jLopezbarb, line 380: NIT: I just noticed that there is a typo on `svc.Healtcheck`. Should we change it to `svc.Healthcheck`? I see the yaml is correctly spelled

pkg/model/stack_serializer.go
```diff
@@ -371,6 +371,21 @@ func getAccessiblePorts(ports []PortRaw) []PortRaw {
...
+				svc.Healtcheck = &HealthCheck{}
+			}
+			svc.Healtcheck.Liveness = true
```
jLopezbarb, line 383: Should we check here if `svc.Healthcheck.Disable != true`

pkg/model/stack_serializer_test.go
```diff
@@ -2286,3 +2286,67 @@ func Test_TranslateOktetoStackPortsToComposePorts(t *testing.T) {
...
+			name: "app depends_on service_healthy db",
+			args: args{
+				svc:     &Service{},
+				svcName: "db",
+				s: &Stack{
+					Services: map[string]*Service{
+						"app": {
+							DependsOn: DependsOn{
+								"db": {
+									Condition: "service_healthy",
```
jLopezbarb, line 2311: NIT: Should we use the constant here? If the name changes we don't need to change it in the tests

jLopezbarb (commented):
Code does what is expected

teresaromero:
> The only concern is that we are adding some magic in the function and if the user adds `readiness` we are transforming it into `liveness`

The proposed change is to translate healthy status from compose to liveness probe of kubernetes. If we want to wait for the service to be healthy, this is translated to liveness, there is no direct translation between compose states and kubernetes ones...

in any case is not readiness for livenesss... can you elaborate on how this changes makes readiness equivalent to liveness? i think i dont see it, sorry

ifbyol:
> > The only concern is that we are adding some magic in the function and if the user adds `readiness` we are transforming it into `liveness`
> 
> The proposed change is to translate healthy status from compose to liveness probe of kubernetes. If we want to wait for the service to be healthy, this is translated to liveness, there is no direct translation between compose states and kubernetes ones... perhaps this should be reviewed by product @ifbyol @pchico83 @codyjlandstrom
> 
> in any case is not readiness for liveness... @jLopezbarb can you elaborate on how this changes makes readiness equivalent to liveness? i think i dont see it, sorry

I don't think the solution is the right one for the problem. We should not change the type of check to do (readiness for liveness). What we should do is to change how we check if a resource is not starting due to a readinessProbe failure. That is the real issue. We were assuming that a readinessProbe failure meant that some container would have some restart, but that is not the case and our assumption is not right, so we need to change. If we need to do different checks depending on the health check, it is fine, they are different.

We should review what happens on Kubernetes when a readinessProbe fails, and adapt the logic to it

teresaromero:
closing this PR in favor of a new approach, currently investigating

PR #4586, opened 2024-11-26
fixes for `make lint`

teresaromero:
# Proposed changes

Running locally `make lint` shows couple errors, this PR fixes them.

I propose to either remove the local command if this is not used by anyone or add it to CI so we dont have to be fixing the static from time to time

@andreafalzetti  there is an inconsistency between make lint and the schema generator. we should take a look the formatting of the schema if we go forward with make lint.


ifbyol:
> @andreafalzetti there is an inconsistency between make lint and the schema generator. we should take a look the formatting of the schema if we go forward with make lint.

@teresaromero There is probably a way of telling the pre-commit to ignore the schema file, and it will probably the easiest one. If not, every time we run the make lint it would apply the change.

We are already ignoring some files in some checks, for example: https://github.com/okteto/okteto/blob/master/.pre-commit-config.yaml#L26

PR #4592, opened 2024-11-28
feat: support rootless runners (PLAT-745)

jpf-okteto:
# Proposed changes

Resolves PLAT-745

Modify Deploy Remote Dockerfile template to support rootless runner container images.

- Use `WORKDIR` to create (if it doesn't exist) a folder owned by the `USER` inherited from the runner container image.
- Add a `RUN` prior to `WORKDIR` to validate that `/okteto` folder has the right user/group ownership. If it doesn't, `WORKDIR` would not alter the ownership which could cause other issues below, that's why it fails the validation.
- Change the way CA is injected by piping it to a rootless writable file path and using `SSL_CERT_DIR` environment variable to tell openssl compatible libraries ([including Go](https://pkg.go.dev/crypto/x509#SystemCertPool)) to consume it.
- Change the filepath where `.oktetocachekey` is stored to a rootless writable file path.
- Use again `WORKDIR` to create `/okteto/src` folder with the right ownership and move it before `COPY`.
- Use `--chmod=0777` to make /okteto/src content readable, writable and executable by world, which includes any rootless user defined in the image. Using `0777` in this scenario is not a security concern because builds are contained.

Challenges:
- Can't set values from `RUN` (e.g. `id -u`) to `ENV` to be used as `USER $var`.
- Can't interpolate values dinamically at `RUN --chown=$var` or `RUN --mount`.
- `VOLUME` doesn't create folders in buildkit.

Other alternatives evaluated and discarded:
- Using multi stage to read the user id via `id -u` and do `chown` in a stage by forcefully setting `USER 0`. Discarded because it would use a rootful stage at some point.
- Splitting Deploy Remote in two phases to obtain the user id of the runner container image and set it directly in the template later. Discarded due to complexity.

## How to validate

```yaml
deploy:
  image: okteto/pipeline-runner:1.0.6-rootless
  commands:
    # write to new file
    - echo foo > new_file
    # write to existing file
    - echo bar > okteto.yml
```

### DIFF
--- a/pkg/remote/remote.go
+++ b/pkg/remote/remote.go
@@ -61,6 +61,8 @@ FROM {{ .UserRunnerImage }} as runner
 USER 0
 {{ end -}}
 ENV PATH="${PATH}:/okteto/bin"
+RUN if [ -d /okteto ]; then echo "/okteto folder is reserved for internal use"; exit 1; fi
+WORKDIR /okteto
 COPY --from=okteto-cli /usr/local/bin/* /okteto/bin/
 
 
@@ -75,11 +77,11 @@ ARG {{ .OktetoDeployable }}
 ARG {{ .GitHubRepositoryArgName }}
 ARG {{ .BuildKitHostArgName }}
 ARG {{ .OktetoRegistryURLArgName }}
-RUN mkdir -p /etc/ssl/certs/
-RUN echo "${{ .TlsCertBase64ArgName }}" | base64 -d > /etc/ssl/certs/okteto.crt
+RUN mkdir -p /okteto/.ssl/certs && echo "${{ .TlsCertBase64ArgName }}" | base64 -d > /okteto/.ssl/certs/okteto.crt
+ENV SSL_CERT_DIR=/etc/ssl/certs:/okteto/.ssl/certs
 
-COPY . /okteto/src
 WORKDIR /okteto/src
+COPY --chmod=0777 . /okteto/src
 
 {{range $key, $val := .OktetoBuildEnvVars }}
 ENV {{$key}}={{$val}}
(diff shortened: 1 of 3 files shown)


pkg/remote/remote.go
```diff
@@ -61,6 +61,8 @@ FROM {{ .UserRunnerImage }} as runner
 USER 0
 {{ end -}}
 ENV PATH="${PATH}:/okteto/bin"
+RUN if [ -d /okteto ]; then echo "/okteto folder is reserved for internal use"; exit 1; fi
```
pchico83, line 64: Creating `/okteto` in the image is a valid use case. For example, if you need to install node modules for your test container, you have to do it in your test image at the `/okteto/src` location

pkg/remote/remote.go
```diff
@@ -75,11 +77,11 @@ ARG {{ .OktetoDeployable }}
...
 ARG {{ .BuildKitHostArgName }}
 ARG {{ .OktetoRegistryURLArgName }}
-RUN mkdir -p /etc/ssl/certs/
-RUN echo "${{ .TlsCertBase64ArgName }}" | base64 -d > /etc/ssl/certs/okteto.crt
+RUN mkdir -p /okteto/.ssl/certs && echo "${{ .TlsCertBase64ArgName }}" | base64 -d > /okteto/.ssl/certs/okteto.crt
+ENV SSL_CERT_DIR=/etc/ssl/certs:/okteto/.ssl/certs
 
-COPY . /okteto/src
 WORKDIR /okteto/src
+COPY --chmod=0777 . /okteto/src
```
pchico83, line 84: why do we need `--chmod=0777` now?

pchico83 (changes requested):

pkg/remote/remote.go
pchico83: Also, can we remove the `{{ if .UseRootUser -}}` logic now?

pchico83:
@jpf-okteto you need to pull from master to run the e2e tests vs staging.okteto.dev, which is now running in the product cluster https://github.com/okteto/okteto/pull/4581

PR #4595, opened 2024-12-02
Enhance healthcheck failure handling and logging

teresaromero:
# Proposed changes

Fixes https://okteto.atlassian.net/browse/DEV-260

This PR proposes a change to show an information log whenever a readiness probe is failing and a service is waiting for being deployed.
It keeps the behaviour of returning an error when a liveness probe fails.

Refactors:
 Events: detect the readiness and the liveness probe fail message and return new `Failure` so the code can identify the origin.
Pod probes detection: when readiness fails, log a message while waiting for being deployed, when livenes return error.

Screenshot 2024-12-02 at 14 15 49
Screenshot 2024-12-02 at 14 16 35

## How to validate

Build the binary and configure different probes and healthchecks

### DIFF
--- a/pkg/cmd/stack/deploy.go
+++ b/pkg/cmd/stack/deploy.go
@@ -370,13 +370,13 @@ func deployServices(ctx context.Context, stack *model.Stack, k8sClient kubernete
 					}
 					if failedServices := getServicesWithFailedProbes(ctx, stack, svcName, k8sClient); len(failedServices) > 0 {
 						for service, err := range failedServices {
-							message := fmt.Sprintf("Service '%s' cannot be deployed because dependent service '%s' is failing its healthcheck probes: %s", svcName, service, err)
+							errMessage := fmt.Errorf("Service '%s' cannot be deployed because dependent service '%s' is failing its healthcheck probes: %s", svcName, service, err)
 							if oktetoErrors.IsLivenessProbeFailed(err) {
-								return fmt.Errorf(message)
+								return errMessage
 							} else if oktetoErrors.IsReadinessProbeFailed(err) {
 								if _, ok := serviceWarnings[service]; !ok {
-									oktetoLog.Information(message)
-									serviceWarnings[service] = message
+									oktetoLog.Information(errMessage.Error())
+									serviceWarnings[service] = errMessage.Error()
 								}
 							}
 						}
--- a/pkg/cmd/stack/deploy.go
+++ b/pkg/cmd/stack/deploy.go
@@ -370,11 +370,11 @@ func deployServices(ctx context.Context, stack *model.Stack, k8sClient kubernete
 					}
 					if failedServices := getServicesWithFailedProbes(ctx, stack, svcName, k8sClient); len(failedServices) > 0 {
 						for service, err := range failedServices {
+							message := fmt.Sprintf("Service '%s' cannot be deployed because dependant service '%s' is failing its healthcheck probes: %s", svcName, service, err)
 							if oktetoErrors.IsLivenessProbeFailed(err) {
-								return fmt.Errorf("service '%s' has failed its healthcheck probes: %w", service, err)
+								return fmt.Errorf(message)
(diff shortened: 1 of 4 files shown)


pkg/k8s/events/crud.go
```diff
@@ -37,23 +37,41 @@ func List(ctx context.Context, namespace, podName string, c kubernetes.Interface
...
-		return ""
+		return nil
 	}
-	previousKilling := false
 	for i := len(events) - 1; i >= 0; i-- {
-		if previousKilling {
-			if events[i].Reason == "Unhealthy" && strings.Contains(events[i].Message, "probe failed") {
-				return events[i].Message
-			}
+		event := events[i]
```
ifbyol, line 62: Why do we remove the check about the killing pod event?

pkg/cmd/stack/deploy.go
```diff
@@ -368,8 +369,16 @@ func deployServices(ctx context.Context, stack *model.Stack, k8sClient kubernete
...
 					}
 					if failedServices := getServicesWithFailedProbes(ctx, stack, svcName, k8sClient); len(failedServices) > 0 {
-						for key, value := range failedServices {
-							return fmt.Errorf("service '%s' has failed his healthcheck probes: %s", key, value)
+						for service, err := range failedServices {
+							if oktetoErrors.IsLivenessProbeFailed(err) {
+								return fmt.Errorf("service '%s' has failed its healthcheck probes: %w", service, err)
+							} else if oktetoErrors.IsReadinessProbeFailed(err) {
+								if _, ok := serviceWarnings[service]; !ok {
+									message := fmt.Sprintf("service '%s' is failing its healthcheck probes: %s", service, err)
```
ifbyol, line 377: Could we add the name of the service that is depending on this? I mean, something in the line of: `Service 'x' cannot be deployed because dependant service ' y'  is failing its healthcheck probes: ....`

pkg/k8s/events/crud.go
teresaromero: the event we want to check is the liveness probe being failed. After looking into the logic, i dont see how checking if "previous" was a killing event adds value to the output of the probe having failed ?

teresaromero: updated the code to include this condition. i assumed the killing event was not need as a condition to check the events what that the restart counts is greater than 0, but these do not reflect exactly that it has been killed.

pkg/k8s/events/crud.go
```diff
@@ -37,23 +38,46 @@ func List(ctx context.Context, namespace, podName string, c kubernetes.Interface
...
+		if killedPod && livenessProbeFailed(event) {
+			return errors.ErrLivenessProbeFailed
 		}
-		if events[i].Reason == "Killing" && (strings.Contains(events[i].Message, "failed liveness probe") || strings.Contains(events[i].Message, "failed readiness probe")) {
-			previousKilling = true
+		if readinessProbeFailed(event) {
```
teresaromero, line 72: killed pod is not condition for readiness, as if the readiness probe fails it does not provoke a restart

pkg/errors/errors.go
```diff
@@ -285,3 +288,11 @@ func IsClosedNetwork(err error) bool {
 func IsErrGitHubNotVerifiedEmail(err error) bool {
 	return err.Error() == ErrGitHubNotVerifiedEmail.Error()
 }
+
+func IsReadinessProbeFailed(err error) bool {
+	return err == ErrReadinessProbeFailed
```
ifbyol, line 293: We should use `errors.Is()` instead of `==`: https://go.dev/wiki/ErrorValueFAQ#how-should-i-change-my-error-handling-code-to-work-with-the-new-features

Same for `IsLivenessProbeFailed`

ifbyol (approved):
Added a comment about how we check the error, but apart from that, everything else looks good, and it is working as expected

pkg/errors/errors.go
teresaromero: i've used directly error.Is and remove the custom function 👍🏻

PR #4612, opened 2024-12-09
Add skipIfNoCache to tests

maroshii:
Restore functionality to skip tests if no cache is set. Depends on:
- #4608 

By defaults, cache will be disabled. To enabled cache for tests, users should add `skipIfNoFileChanges=true` to the/each section in the manifest:

```yaml
test:
  unit:
    image: okteto/golang:1
    skipIfNoFileChanges: true
    commands:
      - "go test . -v"
```

### DIFF
--- a/schema.json
+++ b/schema.json
@@ -1264,6 +1264,11 @@
               "description": "Base image used to run your test.",
               "title": "image",
               "type": "string"
+            },
+            "skipIfNoFileChanges": {
+              "description": "Skip the test execution if no files have changed since the last test run. This is useful to avoid running tests when the code hasn't changed.",
+              "title": "skipIfNoFileChanges",
+              "type": "boolean"
             }
           },
           "required": [


pkg/model/test.go
```diff
@@ -21,13 +21,14 @@ import (
...
-	Artifacts []Artifact    `yaml:"artifacts,omitempty"`
-	Hosts     []Host        `yaml:"hosts,omitempty"`
+	Image               string        `yaml:"image,omitempty"`
+	Context             string        `yaml:"context,omitempty"`
+	Commands            []TestCommand `yaml:"commands,omitempty"`
+	DependsOn           []string      `yaml:"depends_on,omitempty"`
+	Caches              []string      `yaml:"caches,omitempty"`
+	Artifacts           []Artifact    `yaml:"artifacts,omitempty"`
+	Hosts               []Host        `yaml:"hosts,omitempty"`
+	SkipIfNoFileChanges bool          `yaml:"skipIfNoFileChanges,omitempty"`
```
ifbyol, line 31: We should include this field in the JSON schema. Ask @andreafalzetti if you need help to know how to do it

ifbyol:
A couple of questions:
* Could you include an screenshot on how it looks like the output when the test is skipped?
* Related to Nickla's questions, How does the `--no-cache` parameter affects to it?

ifbyol:
I think you need to run the `generate-schema` command to update the local `schema.json` file. @andreafalzetti Should we document in the readme the process of updating the schema when changing something in the Okteto Manifest?

pkg/remote/remote.go
```diff
@@ -320,7 +326,7 @@ func (r *Runner) Run(ctx context.Context, params *Params) error {
 		outputMode = buildCmd.DeployOutputModeOnBuild
 	}
 
-	buildOptions := buildCmd.OptsFromBuildInfoForRemoteDeploy(buildInfo, &types.BuildOptions{OutputMode: outputMode, NoCache: params.NoCache})
+	buildOptions := buildCmd.OptsFromBuildInfoForRemoteDeploy(buildInfo, &types.BuildOptions{OutputMode: outputMode})
```
teresaromero, line 329: the cache option was not passed to the remote, was this intentionally? we do not use cache for remote?

teresaromero:
should we update integration test to cover this change? can we run integration with the current ones to see if affects?

ifbyol:
I was testing it and it looks great! 

@codyjlandstrom @pchico83 I have a doubt on how the cache should work when there is a `depend_on`. I will try to explain the scenario I refer to. Imagine I have the following manifest:

```
test:
  unit:
    context: unit/
    commands:
      - name: "Run unit tests"
        command: go test....

  integration:
    skipIfNoFileChanges: true
    context: integration/
    commands:
      - name: "Run integration tests"
         command: go test .....
```

If I execute `okteto test`, and `unit` is executed because it is not using cache or there was a change in its context, should `integration` run too even if there wasn't a change in the integration context because it depends on `unit`?

pchico83:
> I was testing it and it looks great!
> 
> @codyjlandstrom @pchico83 I have a doubt on how the cache should work when there is a `depend_on`. I will try to explain the scenario I refer to. Imagine I have the following manifest:
> 
> ```
> test:
>   unit:
>     context: unit/
>     commands:
>       - name: "Run unit tests"
>         command: go test....
> 
>   integration:
>     skipIfNoFileChanges: true
>     context: integration/
>     commands:
>       - name: "Run integration tests"
>          command: go test .....
> ```
> 
> If I execute `okteto test`, and `unit` is executed because it is not using cache or there was a change in its context, should `integration` run too even if there wasn't a change in the integration context because it depends on `unit`?

I wouldn't tight `skipIfNoFileChanges` with dependencies until we have user feedback and clear use cases. It's always easier to add functionality than to remove it. In general, I expect `skipIfNoFileChanges` to be used on unit and deterministic tests without dependencies

ifbyol (approved):

PR #4619, opened 2024-12-10
[JSON Schema] Add links to docs

andreafalzetti:
In this PR we introduce links to our docs in the JSON Schema. It couples the links and the JSON Schema but the result is that from the IDE I can jump directly to the Okteto docs ✨ 

Screenshot 2024-12-10 at 15 52 35

The PR also contains minor improvements/fixes (eg. the icon field)

### DIFF
--- a/pkg/schema/external.go
+++ b/pkg/schema/external.go
@@ -40,13 +40,15 @@ func (external) JSONSchema() *jsonschema.Schema {
 	})
 	externalProps.Set("icon", &jsonschema.Schema{
 		Type:        &jsonschema.Type{Types: []string{"string"}},
+		Enum:        []any{"default", "container", "dashboard", "database", "function", "graph", "storage", "launchdarkly", "mongodb", "gcp", "aws", "okteto"},
 		Title:       "icon",
-		Description: "Icon to represent the external resource",
+		Description: withManifestRefDocLink("Icon to represent the external resource", "icon-string-optional"),
+		Default:     "default",
 	})
 	externalProps.Set("endpoints", &jsonschema.Schema{
 		Type:        &jsonschema.Type{Types: []string{"array"}},
 		Title:       "endpoints",
-		Description: "List of endpoints to access the external resource",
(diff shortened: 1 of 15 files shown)


pkg/schema/external.go
```diff
@@ -57,8 +59,6 @@ func (external) JSONSchema() *jsonschema.Schema {
 
 	return &jsonschema.Schema{
 		Type:                 &jsonschema.Type{Types: []string{"object"}},
-		Title:                "external",
-		Description:          "A list of external resources that are part of your development environment",
```
ifbyol, line 61: Just curiosity, why removing the title and description of this item?

pkg/schema/icon.go
```diff
@@ -13,21 +13,14 @@
...
+import (
+	"github.com/kubeark/jsonschema"
+)
 
 type icon struct{}
 
 func (icon) JSONSchema() *jsonschema.Schema {
 	return &jsonschema.Schema{
 		Type: &jsonschema.Type{Types: []string{"string"}},
-		AnyOf: []*jsonschema.Schema{
```
ifbyol, line 23: Why removing this enum? Doesn't it apply anymore?

ifbyol (approved):
Approving it as this is an improvement, but this is very brittle and hard to maintain. It is really easy that some link is broken and we are not aware. 

We should centralize the links and have a test or step in CI that validates that links are still valid. Also, we have the problem with the versions (of docs and the schema). We will need to think on how to maintain this in a long term

pkg/schema/icon.go
andreafalzetti: The enum should only apply to `external.*.icon`, not the root-level `icon` which is only a string (URL)

pkg/schema/external.go
andreafalzetti: The title for root-level properties are defined in the struct with the `jsonschema` annotation, these were getting ignored

teresaromero (approved):

PR #4622, opened 2024-12-11
add validation for empty commands in manifest tests

teresaromero:
# Proposed changes

Fixes DEV-807

Add validation for the manifest for test.
As we want to enforce the field to not be empty, an error is returned.

## How to validate

Create a test manifest and run `okteto test <name>`

For example

```
test:
  unit:
    context: .
```

Output:

```
❯ ok test unit
 i  Using - @ - as context
 x  Commands cannot be empty
```

### DIFF
--- a/pkg/model/test.go
+++ b/pkg/model/test.go
@@ -41,6 +41,7 @@ var (
 	ErrHostMalformed   = fmt.Errorf("host is malformed")
 	ErrInvalidHostName = fmt.Errorf("invalid hostname")
 	ErrInvalidIp       = fmt.Errorf("invalid ip")
+	ErrEmptyCommands   = fmt.Errorf("commands cannot be empty")
 )
 
 func (test ManifestTests) Validate() error {
@@ -52,6 +53,9 @@ func (test ManifestTests) Validate() error {
 	for _, t := range test {
 		if t != nil {
 			hasAtLeastOne = true
+			if len(t.Commands) == 0 {
+				return ErrEmptyCommands
+			}
 			break
 		}
 	}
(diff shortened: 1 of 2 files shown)


andreafalzetti (approved):

pkg/model/test.go
```diff
@@ -52,6 +53,9 @@ func (test ManifestTests) Validate() error {
 	for _, t := range test {
 		if t != nil {
 			hasAtLeastOne = true
+			if len(t.Commands) == 0 {
+				return ErrEmptyCommands
+			}
```
ifbyol, line 58: I think this would only check that the first non-nil test has commands, there is a break in line 59 that would prevent the rest of tests.

For example, with the following manifest, I'm not getting any error:

```
test:
  unit:
    commands:
      - echo "hola"
  integration:
    image: alpine
```

ifbyol (changes requested):

pkg/model/test.go
```diff
@@ -41,6 +41,7 @@ var (
 	ErrHostMalformed   = fmt.Errorf("host is malformed")
 	ErrInvalidHostName = fmt.Errorf("invalid hostname")
 	ErrInvalidIp       = fmt.Errorf("invalid ip")
+	ErrEmptyCommands   = fmt.Errorf("commands cannot be empty")
```
ifbyol, line 44: Could we include the test name in the error? I think it would help to users to know exactly where the problem might be

ifbyol:
> I guess I duplicated this PR: #4628

To give some context to @teresaromero . While validating this, I found a panic error not related to these changes, so I asked Fran to take a look, but the solution covers both scenarios, so I think we could close this one in favor of the other one, as it fixes both errors.

Sorry for the mess, I though they were in different parts of the code

PR #4626, opened 2024-12-12
Add internal flag to buildkit solveOpt

maroshii:
I've been searching in docs and general searches and couldn't find much about this flag. This is the ref in code: https://github.com/moby/buildkit/blob/master/api/services/control/control.proto#L74

We should test this before making it the default/GA.

### DIFF
--- a/pkg/build/buildkit/opt.go
+++ b/pkg/build/buildkit/opt.go
@@ -225,6 +225,11 @@ func (b *SolveOptBuilder) Build(ctx context.Context, buildOptions *types.BuildOp
 		Session:       attachable,
 		CacheImports:  []client.CacheOptionsEntry{},
 		CacheExports:  []client.CacheOptionsEntry{},
+		// Setting the internal flag marks the build as internal which prevents the
+		// build from appearing in build history and improves performance.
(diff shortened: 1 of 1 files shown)


ifbyol:
@pchico83 @jLopezbarb What do you think about it? Right now we are not using actively history in BuildKit, but it was useful in some incidents to see what was going on.

I think we can merge it, and play a bit in our side with it to see if there is some difference. The default value would keep being the same as now.

pchico83:
> @pchico83 @jLopezbarb What do you think about it? Right now we are not using actively history in BuildKit, but it was useful in some incidents to see what was going on.
> 
> I think we can merge it, and play a bit in our side with it to see if there is some difference. The default value would keep being the same as now.

Sounds good to me, we can evaluate performance on the product cluster first for example

ifbyol (approved):

jLopezbarb:
> @pchico83 @jLopezbarb What do you think about it? Right now we are not using actively history in BuildKit, but it was useful in some incidents to see what was going on.
> 
> I think we can merge it, and play a bit in our side with it to see if there is some difference. The default value would keep being the same as now.

If the performance is the same and it helps us I think its a good addition

PR #4628, opened 2024-12-12
Add validation to test section

maroshii:
Fix panic for empty test section:

```yaml
test:
  unit:
    image: okteto/golang:1
    commands:
      - "echo ok"
  integration:
```

Looks like this:

Screenshot 2024-12-12 at 2 39 56 PM

### DIFF
--- a/pkg/model/test_test.go
+++ b/pkg/model/test_test.go
@@ -23,9 +23,9 @@ import (
 
 func TestValidate(t *testing.T) {
 	tests := []struct {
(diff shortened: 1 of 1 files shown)


ifbyol (approved):
Nit: Would it make sense to add a unit test to the function even if it is something simple?

andreafalzetti (approved):

teresaromero (approved):
+1 to adding a small unit test

PR #4629, opened 2024-12-13
DEV-834: Fixed scenario where an okteto deploy executed within a manifest with remote flag wasn't getting correctly the variables

ifbyol:
# Proposed changes

Fixes DEV-834

When an Okteto manifest was executing another `okteto deploy` within its deploy commands passing some variables, the inner deploy command wasn't receiving them correctly.

So, imagine you have 2 manifest like the following ones:

`okteto.yaml`
```yaml
deploy:
  remote: true
  commands:
    - okteto deploy -f other-okteto.yml --var INNER_VAR="${VAR1}"
```

`other-okteto.yml`
```yaml
deploy:
  remote: true
  commands:
    - name: echo variable
      command: echo "inner var ${INNER_VAR}"
```

The `INNER_VAR` variable was empty, so only `inner var` was being printed in the output. After the changes in this PR, it prints `inner var ***`. Asterisks are because the variable is masked, but that means that the inner command is reading it correctly.

This check was added in this PR https://github.com/okteto/okteto/pull/3688, when the deploy command was triggering another `okteto deploy` command as part of the Dockerfile executing remote operations, because originally variables weren't being passed as `--var` to the remote command, so it was reading them from the configmap. As the deploy command doesn't trigger another deploy command anymore (it uses `remote-run`), and variables are already passed with `--var`, this check doesn't apply anymore.

The fun fact is that, it wasn't actually reading the variables from the configmap, because as it was a remote operation, we were using `deployInsideDeployConfigMapHandler`, which was returning empty string. It was working, because the deploy command reads the variables and set them as environment variables, being available for the execution of the commands. 

This was broken when we stop calling okteto deploy on remote, because `remote-run` doesn't read the configmap. As any command executed within a remote Okteto Manifest has the env var `OKTETO_DEPLOY_REMOTE` set, it was entering in that if, ignoring vars passed in the command, and getting them from an empty configmap.

Modified an e2e test to cover this case

## How to validate

With the 2 manifest described above, do the following one:
* With latest stable CLI, execute `okteto deploy --var VAR1=test-value`
* You will see how in the output you only see `inner var`
* Now, build the binary of this branch
* Build the docker image of this branch, as you need to specify it for the remote operation, to do that:
  * To build the docker image: `okteto build -f Dockerfile . -t okteto.dev/cli:validation`
  * `export OKTETO_REMOTE_CLI_IMAGE=<generated-image>`
* Execute `okteto deploy --var VAR1=test-value`
* You should see now `inner var ***` (asterisks are expected because variables are masked)


cmd/deploy/deploy.go
```diff
@@ -363,7 +363,7 @@ func (dc *Command) Run(ctx context.Context, deployOptions *Options) error {
 		return err
 	}
 
-	if dc.IsRemote || dc.RunningInInstaller {
+	if dc.RunningInInstaller {
```
ifbyol, line 366: The whole if can be probably be removed, as the installer is now passing variables to the command using the `--var` flag, but left it here because it might imply other changes, and in the case of the installer is not a big deal, because when the deployment is executed from the UI (or through a preview environment) the variables are stored in the configmap. But when the deploy is executed directly from the CLI, variables are not yet stored in the configmap.

So, leaving that for future evaluation

jLopezbarb (approved):
Tested also adding special characters such as \" or \' and it's working correctly

jLopezbarb (approved):

ifbyol:
I was testing the following scenarios:
* Nested `okteto deploy` execution with `remote`
* Nested `okteto deploy` execution without `remote`
* Simple (no nested) `okteto deploy` execution with `remote`
* Simple (no nested) `okteto deploy` execution without `remote`
* The four scenarios described above within an installer job (deploying the scenarios from the UI)

All tests were good, so merging it

PR #4634, opened 2024-12-19
Fran/smart builds ignore

maroshii:
# Proposed changes

Consider `.dockerignore` and `.oktetoignore` for calculating smart builds.

It filters the lines of ls-files, untrack files and the diff targets based on okteto and docker ignore files.

## Testing

```bash
echo "README.md" > .oktetoignore

okteto build
Building '/.../go-getting-started/Dockerfile' in tcp://buildkit...

echo "beep boop" >> README.md

# although README.md changed, it is in the ignore file so hash matches
okteto build
 i  Okteto Smart Builds is skipping build of 'hello-world' because it's already built from cache.

echo "" > .oktetoignore

okteto build
Building '/.../go-getting-started/Dockerfile' in tcp://buildkit...

```

### DIFF
--- a/pkg/repository/git.go
+++ b/pkg/repository/git.go
@@ -47,24 +47,25 @@ var (
 type gitRepoController struct {
 	repoGetter repositoryGetterInterface
 	fs         afero.Fs
-	ignorer    ignore.Ignorer
+	ignorer    func(subpath string) ignore.Ignorer
 	path       string
 }
 
 func newGitRepoController(path string) gitRepoController {
-	var ignorer ignore.Ignorer = ignore.Never
-
-	if env.LoadBoolean("OKTETO_SMART_BUILDS_IGNORE_FILES_ENABLED") {
-		ignorer = ignore.NewMultiIgnorer(
-			ignore.NewDockerIgnorer(".dockerignore"),
-			ignore.NewOktetoIgnorer(".oktetoignore").BuildOnly(),
-		)
-	}
 	return gitRepoController{
 		repoGetter: gitRepositoryGetter{},
 		path:       path,
 		fs:         afero.NewOsFs(),
-		ignorer:    ignorer,
+		ignorer: func(subpath string) ignore.Ignorer {
+			if !env.LoadBoolean("OKTETO_SMART_BUILDS_IGNORE_FILES_ENABLED") {
+				return ignore.Never
+			}
+			return ignore.NewMultiIgnorer(
+				ignore.NewDockerIgnorer(filepath.Join(subpath, ".dockerignore")),
+				ignore.NewOktetoIgnorer(filepath.Join(subpath, ".oktetoignore")).BuildOnly(),
+			)
+
+		},
 	}
 }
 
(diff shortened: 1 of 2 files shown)


jLopezbarb (approved):
Tested the change for both commands and it seems to be working correctly

pkg/repository/local_git.go
```diff
@@ -295,7 +300,90 @@ func (lg *LocalGit) GetDirContentSHA(ctx context.Context, gitPath, dirPath strin
...
+			parts := strings.Fields(line)
+
+			if len(parts) < lsFilesColumnCount {
+				// if we can't parse a line, add the file to the list.
+				// It's better to over build than to over cache
+				//nolint:errcheck
+				pw2.Write([]byte(line + "\n"))
+			} else {
+				filename := parts[3]
+				shouldIgnore, err := lg.ignorer.Ignore(filename)
```
ifbyol, line 353: Adding this within `GetDirContentSHA` might provoke that all images are rebuilt, as the sha would be different even if there are no changes. 

I think that changing it only for `Diff` function and `ListUntrackedFiles` would be enough. Do you recall which case we wanted to cover adding this?

The relevant part would be not to trigger builds when there is a change in an ignored file, but this function I think it is not being used for  that.

pkg/repository/git.go
```diff
@@ -53,6 +55,10 @@ func newGitRepoController(path string) gitRepoController {
 		repoGetter: gitRepositoryGetter{},
 		path:       path,
 		fs:         afero.NewOsFs(),
+		ignorer: ignore.NewMultiIgnorer(
+			ignore.NewDockerIgnorer(".dockerignore"),
+			ignore.NewOktetoIgnorer(".oktetoignore").BuildOnly(),
+		),
```
ifbyol, line 61: Would it make sense to add an environment variable to disable this behavior in case something bad happens and builds are not re-triggered when needed? By default it would have this behavior, but if some customer is affected by something unexpected we can tell them to set that env and don't do any ignore (it would be just to create here a never ignorer or a multi ignore). WDYT?

ifbyol (commented):
We talked about some restrictions for this to work due to paths and the location of the files. What are those restrictions? My understanding is that:
* The ignore file should be in the root of the repo
* The command has to be executed from the root of the repo

Is that correct?

pkg/repository/local_git.go
ifbyol: Talked offline about this. This is needed because the output of `git ls` command would be different if there is a change in a file, so it would trigger a new rebuild. So, even if the first build is retriggered, it is fine.

We have talked also about possible performance implications or error. That's why we will add an environment variable to enable this new behavior only when it is enabled, so we can test and tell customers to enable it

pkg/repository/git.go
ifbyol: As discussed offline, we will add a feature flag to enable this behavior. By default, it will be disabled. 

I was thinking about the env var name. What do you think about any of these:
* `OKTETO_SMART_BUILDS_IGNORE_FILES_ENABLED`
* `OKTETO_SMART_BUILDS_USE_IGNORE_FILES`
* `OKTETO_SMART_BUILDS_READ_IGNORE_FILES`

ifbyol (approved):
I have been testing all the different scenarios, and it seems to be working fine in all the cases. 

I used this repository https://github.com/ifbyol/test-okteto with the following branch `ifbyol/test-smart-build-contexts` to represent the main common scenarios, and all of them work correctly.

Would it make sense to add a log to indicate the git path and the dir path on the git operations to help us to troubleshoot any possible issue? This is not blocker at all

PR #4636, opened 2025-01-03
fix cves

maroshii:


Dockerfile
```diff
@@ -1,6 +1,6 @@
-ARG KUBECTL_VERSION=1.30.7
-ARG HELM_VERSION=3.16.3
-ARG SYNCTHING_VERSION=1.28.0
+ARG KUBECTL_VERSION=1.30.8
+ARG HELM_VERSION=3.16.4
+ARG SYNCTHING_VERSION=1.28.1
```
jLopezbarb, line 3: We also need to modify it here: https://github.com/okteto/okteto/blob/9f4a35b6afb69e88d34b1bfac5bbafd83bdb048a/pkg/syncthing/install.go#L33

ifbyol (approved):

PR #4649, opened 2025-01-13
deps: update Syncthing version to 1.29.2

jLopezbarb:
Updates syncthing to 1.29.2

https://github.com/syncthing/syncthing/releases/tag/v1.29.2


ifbyol (approved):
We need to updated the docs too: https://github.com/okteto/docs/blob/main/src/content/variables.json#L6

codyjlandstrom:
> We need to updated the docs too: https://github.com/okteto/docs/blob/main/src/content/variables.json#L6

I added this as part of the docs release process. We can update when we create docs version 1.29

PR #4651, opened 2025-01-14
fix divert with docker compose

maroshii:
Fixes divert with docker compose. It also fixes an issue where compose was not respecting the previous labels and annotations, so the autocreate annotations and the ones previously set were being overriden.

### DIFF
--- a/pkg/k8s/services/crud.go
+++ b/pkg/k8s/services/crud.go
@@ -49,27 +49,17 @@ func Deploy(ctx context.Context, s *apiv1.Service, c kubernetes.Interface) error
 	} else {
 		oktetoLog.Infof("updating service '%s'", s.Name)
 
-		for k, v := range s.Annotations {
-			if old.Annotations == nil {
-				old.Annotations = make(map[string]string)
-			}
-			old.Annotations[k] = v
-		}
+		isDivertDeploy := old.Spec.Type == apiv1.ServiceTypeExternalName && old.Annotations[model.OktetoAutoCreateAnnotation] == "true"
 
-		for k, v := range s.Labels {
-			if old.Labels == nil {
-				old.Labels = make(map[string]string)
-			}
-			old.Annotations[k] = v
-		}
-
+		old.Annotations = s.Annotations
+		old.Labels = s.Labels
 		old.Spec.Ports = s.Spec.Ports
 		old.Spec.Selector = s.Spec.Selector
 
 		if s.Spec.Type == apiv1.ServiceTypeClusterIP {
-			if old.Spec.Type == apiv1.ServiceTypeExternalName && old.Annotations[model.OktetoAutoCreateAnnotation] == "true" {
-				old.Spec.Type = apiv1.ServiceTypeClusterIP
-				old.Spec.ExternalName = ""
+			if isDivertDeploy {
+				old.Spec.Type = s.Spec.Type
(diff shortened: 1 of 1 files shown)


jLopezbarb (approved):

pkg/k8s/services/crud.go
```diff
@@ -48,10 +48,31 @@ func Deploy(ctx context.Context, s *apiv1.Service, c kubernetes.Interface) error
...
+		}
+
 		old.Spec.Ports = s.Spec.Ports
 		old.Spec.Selector = s.Spec.Selector
+
+		if s.Spec.Type == apiv1.ServiceTypeClusterIP {
+			if old.Spec.Type == apiv1.ServiceTypeExternalName && old.Annotations[model.OktetoAutoCreateAnnotation] == "true" {
+				old.Spec.Type = apiv1.ServiceTypeClusterIP
+				old.Spec.ExternalName = ""
+			}
```
ifbyol, line 73: Would it make sense to copy these values from `s` as the rest of fields?

pkg/divert/nginx/services.go
```diff
@@ -40,11 +40,6 @@ func (d *Driver) divertService(ctx context.Context, name string) error {
 			if !k8sErrors.IsAlreadyExists(err) {
 				return err
 			}
-			// the service was created, refresh the cache
-			newS, err = d.client.CoreV1().Services(d.namespace).Get(ctx, newS.Name, metav1.GetOptions{})
-			if err != nil {
-				return nil
-			}
```
ifbyol, line 47: Why we need this? Don't we need to refresh the cache (base on the inline comment)

pkg/k8s/services/crud.go
```diff
@@ -48,10 +48,31 @@ func Deploy(ctx context.Context, s *apiv1.Service, c kubernetes.Interface) error
...
+				old.Annotations = make(map[string]string)
+			}
+			old.Annotations[k] = v
+		}
+
+		for k, v := range s.Labels {
+			if old.Labels == nil {
+				old.Labels = make(map[string]string)
+			}
+			old.Annotations[k] = v
```
pchico83, line 63: This should be `Labels`, no?

pkg/k8s/services/crud.go
```diff
@@ -48,10 +48,31 @@ func Deploy(ctx context.Context, s *apiv1.Service, c kubernetes.Interface) error
 		oktetoLog.Infof("created service '%s'", s.Name)
 	} else {
 		oktetoLog.Infof("updating service '%s'", s.Name)
-		old.Annotations = s.Annotations
-		old.Labels = s.Labels
+
+		for k, v := range s.Annotations {
```
pchico83, line 52: Adding annotations and/or labels will break compose deploy logic. For example, if I drop labels/annotations in my docker compose yaml file, `okteto deploy` will keep the previous ones.

pchico83: For this fix, I wouldn't mind if we lose the divert annotations, and we do a proper fix later, for example, by running the divert logic after the docker compose logic

pchico83 (approved):

pkg/divert/nginx/services.go
ifbyol: 👍  perfect, thanks for the explanation!

ifbyol (approved):

PR #4655, opened 2025-01-16
Resolve platform env for test in remote-run

maroshii:
Adds platform environment to `remote-run`. We were already doing is in the deploys but for some reason not in destroy and test. After adding `skipIfNoFileChanges` to test, if cloud credentials was configured cache was always busted regardless of the flag becase cloud credentials are generared each time.

### DIFF
--- a/cmd/deploy/deploy.go
+++ b/cmd/deploy/deploy.go
@@ -102,7 +102,6 @@ type getDeployerFunc func(
 	*io.Controller,
 	*io.K8sLogger,
 	dependencyEnvVarsGetter,
-	executionEnvVarsGetter,
 ) (Deployer, error)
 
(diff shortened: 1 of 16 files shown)


ifbyol:
I have sent you a video privately where this is not working for me. Maybe I'm doing something wrong, or something with my user (I'm directly using the product cluster). Could you check it, please?

ifbyol (approved):
I was testing it and it works correctly.

During the tests, I found another issue that makes the usage of the cache for tests hard to use.

As this is an improvement because it unifies how the variables are injected for deploy, destroy and test, I think we can merge it (I was testing it and seems to be working fine), but if we prefer to discard it to prevent issues in cloud credentials behavior, I'm ok too

jLopezbarb (approved):

PR #4656, opened 2025-01-23
Don't show warning for newer CLI versions

maroshii:
Although the message is annoying, it is useful for debugging purposes to have this info for troubleshooting executions so I moved it to debug logging. If this is still not desirable I can remove


ifbyol (approved):
Good point about moving it to debug! Definitely, it will be useful!

We also have to change the one about the version being lower to the recommended one: https://github.com/okteto/okteto/blob/master/cmd/context/create.go#L341.

Approving it assuming the change will be done!

jLopezbarb (approved):

PR #4657, opened 2025-01-27
feat: add test event insights

jLopezbarb:
# Proposed changes

Fixes DEV-306

Send an event everytime a test is run so it gets propagated into the cluster

## How to validate

1. Using the right backend
1. Run `okteto test` on a repo that has them configured 
1. Run `curl -H 'Authorization: Bearer ${INSIGHT_TOKEN}' ${CLUSTER_URL}/metrics' | grep -E 'okteto_usage_test'`

## CLI Quality Reminders 🔧

### DIFF
--- a/pkg/analytics/test.go
+++ b/pkg/analytics/test.go
@@ -42,6 +42,27 @@ type TestMetadata struct {
 	StagesCount int
 }
 
+// SingleTestMetadata contains the metadata of a single test execution
+type SingleTestMetadata struct {
+	// DevenvName is the name of the development environment where the test was run
+	DevenvName string `json:"devenv_name"`
+
+	// Namespace is the namespace where the test was run
+	Namespace string `json:"namespace"`
+
+	// TestName is the name of the test that was run
+	TestName string `json:"test_name"`
+
+	// Repository is the repository where the test was run
+	Repository string `json:"repository"`
+
+	// Duration is the duration of the test execution
+	Duration time.Duration `json:"duration"`
+
+	// Success is whether the test succeeded or not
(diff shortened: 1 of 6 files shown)


pkg/analytics/test.go
```diff
@@ -42,6 +42,27 @@ type TestMetadata struct {
 	StagesCount int
 }
 
+// SingleTestMetadata contains the metadata of a single test execution
+type SingleTestMetadata struct {
+	// DevenvName is the name of the development environment where the test was run
+	DevenvName string `json:"devenv_name"`
```
ifbyol, line 48: Same open question I left in the backend implementation. Does the dev env name provide some value in the metric? It might be misleading as we might have a name, but without having a dev environment deployed in the namespace

ifbyol: Discussed offline about it. Dev env name is the only way we have to identify okteto manifests for the different operations and it is how we do it also for other operations as build, so it has to be specified

pkg/insights/test.go
```diff
@@ -0,0 +1,65 @@
+// Copyright 2023 The Okteto Authors
```
ifbyol, line 1:
```suggestion
// Copyright 2025 The Okteto Authors
```

pkg/insights/test_test.go
```diff
@@ -0,0 +1,99 @@
+// Copyright 2023 The Okteto Authors
```
ifbyol, line 1:
```suggestion
// Copyright 2025 The Okteto Authors
```

ifbyol (approved):

PR #4663, opened 2025-02-06
fix: cli was not waiting if buildkit not available

jLopezbarb:
# Proposed changes

Bug introduced in #4577 

We needed to get some buildkit info to create the buildkit options. We introduced the bug checking if we should or not add the frontend versus the buildkit server information. In order to do that we need to wait for the buildkit server to be up and running  

## How to validate

1. Scale buildkit sts replicas to 0
1. Run `okteto build` on a manifest with build section

## CLI Quality Reminders 🔧

### DIFF
--- a/pkg/cmd/build/build.go
+++ b/pkg/cmd/build/build.go
@@ -156,6 +156,10 @@ func (ob *OktetoBuilder) buildWithOkteto(ctx context.Context, buildOptions *type
 
 	reg := registry.NewOktetoRegistry(GetRegistryConfigFromOktetoConfig(ob.OktetoContext))
 
+	if err := buildkitWaiter.WaitUntilIsUp(ctx); err != nil {
+		return err
+	}
+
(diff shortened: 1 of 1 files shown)


pkg/cmd/build/build.go
```diff
@@ -156,6 +156,10 @@ func (ob *OktetoBuilder) buildWithOkteto(ctx context.Context, buildOptions *type
 
 	reg := registry.NewOktetoRegistry(GetRegistryConfigFromOktetoConfig(ob.OktetoContext))
 
+	if err := buildkitWaiter.WaitUntilIsUp(ctx); err != nil {
```
ifbyol, line 159: If we are calling it, and waiting here, would it make sense to remove the call we do within `buildSolver.Run`? or should we keep both in case we call to `Run` from another place?

jLopezbarb: I'd keep it in case there is a restart in the middle of both events but we can remove it also.

ifbyol (approved):

PR #4664, opened 2025-02-06
DEV-859: Updated kubernetes dependencies to 0.31.x

ifbyol:
# Proposed changes

Fixes DEV-859

Updated kubernetes dependencies to `0.31.x` to match our current kubernetes supported versions. This update also solves the issue we detected in our dogfooding environments where the field `appArmorProfile` was being lost in the BuildKit deployment due to the proxy translations.

The issue was that the `PodSecurityContext` struct didn't include yet the `appArmorProfile` field, so in the translation, it was being lost.

I have tested that after these changes, the filed is not lost anymore

Main changes included:
* Change the usage of `k8s.io/utils/pointer` in favor of `k8s.io/utils/ptr`
* Type for resource requirements for PVCs specs have changed from `apiv1.ResourceRequirements` to `apiv1.VolumeResourceRequirements`
* I had to remove a check in a test because the fake client now returns, along with the proper error, an empty object instead of nil when a resource is not found in the cluster. I added an inline comment about it
* A couple of minor changes to accommodate the code to new function's signature

## How to validate

Mainly verify everything keeps working fine after the upgrade. Deploys, compose translations, okteto up, etc...


pkg/k8s/deployments/crud_test.go
```diff
@@ -208,9 +208,6 @@ func TestGet(t *testing.T) {
 			if tt.expectedFoundCount > 0 && d == nil {
 				t.Fatalf("expected %d deployments, instead found none", tt.expectedFoundCount)
 			}
-			if tt.expectedFoundCount == 0 && d != nil {
```
ifbyol, line 211: I removed this because it the new versions of client-go, the fake client returns empty resources instead of nil when the resource doesn't exit (along with the error). https://github.com/kubernetes/client-go/commit/3d4f98db0e40e6bf0a0d6e48d4ef2e37365f6ba3#diff-5f29a60c4fc9f826637dd6d5496770ec8a20b960dc6b23433fc7c312a0351f0bR121

So, as we are already checking the error, and verifying is a not found when a deployment is not present, I think this can be removed without affecting the test

ifbyol:
I was testing everything that came to my mind:
* Docker compose translation
* Deploys
* Destroys
* Okteto up and exec
* Builds
* Registry credentials resolution
* Remote deploy
* Executed as part of the installer job

jLopezbarb (approved):
From what i tested the cobra change didn't affect and everything seems to be working correctly

PR #4668, opened 2025-02-11
feat: allow proxy protobuff integration

jLopezbarb:
# Proposed changes

Fixes DEV-851

- Move all the proxy translation to its own file depending.
- Created an interface to translate the k8s object depending into the contentype is json or protobuff.
- Only recreated the object meta translation given that this change only affects the k8s core API (Pod, PodTemplate, ReplicationController, Service, Endpoints, Namespace, Node, Secret, ConfigMap, PersistentVolume, PersistentVolumeClaim, ServiceAccount, Event, ComponentStatus, LimitRange, ResourceQuota, Binding)

## How to validate

1. Using the [movies](https://github.com/okteto/movies) repo
1. Build the CLI and export it with `export OKTETO_REMOTE_CLI_IMAGE=yourcliimage`
1. Run `okteto deploy`

## CLI Quality Reminders 🔧


pkg/deployable/proxy.go
```diff
@@ -246,8 +241,13 @@ func (ph *proxyHandler) getProxyHandler(token string, clusterConfig *rest.Config
...
 				return
 			}
-
-			b, err = ph.translateBody(b)
+			var translator translator
+			if r.Header.Get("Content-Type") == "application/json" {
+				translator = NewJSONTranslator(b, ph.Name, ph.DivertDriver)
+			} else {
+				translator = NewProtobufTranslator(b, ph.Name, ph.DivertDriver)
+			}
```
ifbyol, line 249: Instead of creating the translators inline, we can have a map [string]translator, where the key is the content-type, so we only created them once, but used as many times as needed

pkg/deployable/protobuf_translator.go
```diff
@@ -0,0 +1,225 @@
...
+	if err := p.translateMetadata(obj); err != nil {
+		return nil, err
+	}
+
+	var buf bytes.Buffer
+	if err := p.serializer.Encode(obj, &buf); err != nil {
+		return nil, fmt.Errorf("could not encode resource: %w", err)
+	}
+
+	switch obj.GetObjectKind().GroupVersionKind().Kind {
```
ifbyol, line 67: What about VirtualService translation?

ifbyol (commented):
In general looks good, but I would check the virtual services translations, and the experience when we return unsupported type. What do we show to the customer?

pkg/deployable/proxy.go
```diff
@@ -246,8 +249,15 @@ func (ph *proxyHandler) getProxyHandler(token string, clusterConfig *rest.Config
 				reverseProxy.ServeHTTP(rw, r)
 				return
 			}
-
-			b, err = ph.translateBody(b)
+			var translator translator
+			if v, ok := ph.translators[r.Header.Get("Content-Type")]; ok {
+				translator = v
+			} else {
+				oktetoLog.Infof("unsupported content type: %s", r.Header.Get("Content-Type"))
```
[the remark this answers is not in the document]
jLopezbarb: They won't notice any error if they don't use log-level flag. The resources won't be grouped together and the destroy of the devenv won't destroy the resources.

pkg/deployable/protobuf_translator.go
jLopezbarb: VirtualService is not implementing `runtime.Object` which prevents me to get the spec from the protobuf object.

pkg/deployable/proxy.go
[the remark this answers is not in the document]
ifbyol: I added the comment in the wrong line. We are returning an unsupportedMediaType error in the response, which I guess would return an error to the client, not finishing the deploy operation, right?

pkg/analytics/proxy.go
```diff
@@ -0,0 +1,26 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package analytics
+
+const (
+	unsupportedContentTypeEvent = "Unsupported Content Type"
```
ifbyol, line 17: Nit: Should we specify that is for the deploy proxy or something similar. It is easy we forget about what this event mean in the future

pkg/deployable/protobuf_translator.go
ifbyol: ok, but it is something we should keep an eye on. If not, divert with virtual services might stop working. But for now, it seems that client won't change the content type used

ifbyol (approved):
Minor comment, but overall looks good

PR #4669, opened 2025-02-11
fix smart build tests

maroshii:
The `git diff HEAD` failed because there are no commits in the repo meaning that HEAD doesn't exists. This fixes the issue by simply creating an empty commit.

```
$ mkdir empty; cd $_
$ git init
Initialized empty Git repository in /Users/maroshii/projects/okteto/empty/.git/
$ touch "FROM alpine > Dockerfile"
$ git diff HEAD
fatal: ambiguous argument 'HEAD': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'

```

### DIFF
--- a/integration/git.go
+++ b/integration/git.go
@@ -65,7 +65,7 @@ func GitInit(path string, withEmptycommit bool) error {
 		return fmt.Errorf("git init failed: %s - %w", string(o), err)
 	}
 	if withEmptycommit {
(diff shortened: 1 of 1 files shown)


ifbyol (approved):
Nice catch! 👏

PR #4675, opened 2025-02-21
Add analytics to track how long it takes the dev container to start

rberrelleza:
# Proposed changes
Add analytics to track the following:
- how long it takes the dev container to go from created to started
- if the up sequence waited for the deployment to awaken 
- how long the developer waits for the deployment to awaken
- if the developer waited for the image to be pulled

This change aims to have more analytics information to help understand the impact of sleep/awake policies at scale.

This is what it looks like with the new properties
image

## How to validate

1. enable analytics
1. deploy the environment and put it to sleep
1. run "okteto up" 
1. exit "okteto up"
1. check on mixpanel for the new properties of the event

## CLI Quality Reminders 🔧


cmd/up/activate.go
```diff
@@ -547,6 +555,7 @@ func (up *upContext) waitUntilAppIsAwaken(ctx context.Context, app apps.App) err
 
 			// If the app is not sleeping anymore, we are done
 			if _, ok := appToCheck.ObjectMeta().Annotations[model.StateBeforeSleepingAnnontation]; !ok {
+				up.analyticsMeta.WaitedForAwakenDuration(time.Since(startWaitingForAwaken))
```
rberrelleza, line 558: This isn't super exact since sometimes the deployment might awaken by the time we hit this part of the code, but I think it's good enough to measure the user experience (did they have to wait for the deployment to wake up?).  Appreciate advise if there's a better place to put this.

pkg/analytics/up.go
```diff
@@ -191,6 +203,16 @@ func (u *UpMetricsMetadata) ExecDuration(duration time.Duration) {
 	u.execDuration = duration
 }
 
+func (u *UpMetricsMetadata) WaitedForAwakenDuration(duration time.Duration) {
+	u.waitedForAwaken = true
+	u.waitedForAwakenDuration = duration
```
rberrelleza, line 208: I added both a boolean and a duration because, on occasions, the wait could be True but the duration 0. But if this is too cumbersome, I'm happy to remove the boolean.

ifbyol (approved):
Looks good to me, but I would like that @jLopezbarb could take a look at it too

jLopezbarb (approved):

jLopezbarb:
The only thing that I can be a little e bit more skeptical is that the pulled image will be almost always true and I think measuring the time could be better.

PR #4682, opened 2025-03-06
add action e2e to okteto test

jLopezbarb:
# Proposed changes

Add e2e test for the okteto test action 

## CLI Quality Reminders 🔧

### DIFF
--- a/integration/actions/test_test.go
+++ b/integration/actions/test_test.go
@@ -71,7 +71,7 @@ func executeOktetoTestAction(namespace, oktetoManifest string) error {
 		return fmt.Errorf("filepath.Abs: %s", err)
 	}
 
-	command := exec.Command(testBinary, "", namespace, "", "", "", "", "", "", "")
+	command := exec.Command(testBinary, "", namespace, "", "", "", "VAR1=Hello", "", "", "")
 	command.Env = os.Environ()
 	command.Dir = filepath.Dir(oktetoManifest)
 	o, err = command.CombinedOutput()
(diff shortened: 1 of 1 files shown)


integration/actions/test_test.go
```diff
@@ -0,0 +1,118 @@
...
+	if err := integration.CloneGitRepoWithBranch(actionRepo, oktetoVersion); err != nil {
+		if err := integration.CloneGitRepo(actionRepo); err != nil {
+			return err
+		}
+		log.Printf("cloned repo %s main branch\n", actionRepo)
+	}
+	log.Printf("cloned repo %s \n", actionRepo)
+	defer integration.DeleteGitRepo(actionFolder)
+
+	builder := exec.Command("go", "build", "-o", "test", "./cmd/main.go")
```
ifbyol, line 61: I remember you mentioned something about you prefer to do this instead of just a `run main.go`, what was it?

integration/actions/test_test.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+
+	command := exec.Command(testBinary, "", namespace, "", "", "", "", "", "", "")
+	command.Env = os.Environ()
+	command.Dir = filepath.Dir(oktetoManifest)
+	o, err = command.CombinedOutput()
+	if err != nil {
+		return fmt.Errorf("test: %s", string(o))
+	}
+
```
ifbyol, line 81: Could we verify that the artifacts specified in the `integration` test was generated? Could we also pass some variables to test and verify that are being used as part of the commands?

Could we also add another case in which we execute just a subset of tests?

integration/actions/test_test.go
```diff
@@ -0,0 +1,118 @@
...
+	log.Printf("destroy preview output: \n%s\n", string(o))
```
ifbyol, line 82:
```suggestion
	log.Printf("okteto test output: \n%s\n", string(o))
```

integration/actions/test_test.go
jLopezbarb: The dockerfile builds the action and after that executes the binary. I wanted to have it as similar as I can so if we break that step, we can have a test detecting it although we already have an specific test for it in the action repo

ifbyol (approved):

PR #4673, opened 2025-02-18
Add back smart builds ignore

maroshii:
- Original PR #4634
- Revert #4665

This change puts back the original change. Only conflicts were in the `go.sum` file

PR #4689, opened 2025-03-24
DEV-877: Split ignorer logic for Smart Build into 2 different functions according to the flag, and use a temporary file for the SHA calculation

ifbyol:
# Proposed changes

After the changes introduced to ignore defined in `.dockerignore` files in the Smart Build logics [here](https://github.com/okteto/okteto/pull/4673), it turns out that for big repos with a lot of files, the command executed to calculate the SHA was failing, provoking that the image is built every time.

Even if there was a feature flag for the feature, the bug comes in a common place, so any could be affected independently of enabling the feature or not.

In this PR we are doing 2 things:
* Only change the logic to calculate things when the feature flag is enabled. If disabled, old logic would be executed (restored functions as they were before the changes in the Smart Builds logic). The idea of this is to prevent users which don't have the feature enabled to have some impact due to the new functionality
* Fix the way dir SHA is calculated to prevent that big repos can calculate it properly. We are creating a temporary file with the content of the `git ls-objects` command, to then, pass it to the `git hash-object` function, instead of doing a pipe as we were doing before

## How to validate

There are a couple of ways for validating it.

You need to play with the environment variable `OKTETO_SMART_BUILDS_IGNORE_FILES_ENABLED` to enable and disable.

When the environment variable `OKTETO_SMART_BUILDS_IGNORE_FILES_ENABLED` set to false, if you run a build using the log-level `debug` you will see how this command is executed: `/usr/bin/git --no-optional-locks ls-files -s <path-to-some-folder> | /usr/bin/git --no-optional-locks hash-object --stdin`. This is to calculate the SHA of the directory. 

Prior to this change, you can see how the following commands are executed:
* `/usr/bin/git --no-optional-locks ls-files -s`. To list all the files in the directory
* `echo <filtered output of previous command> | /usr/bin/git --no-optional-locks hash-object --stdin` To get the SHA

When the list of files was too long, it was generating an error, so image was being built always. With this PR, if the flag is disabled, we restore the original way of calculating the SHA, as no function to ignore files has to be applied, so no need to change the commands to execute.

Then, if you change the value of the environment variable `OKTETO_SMART_BUILDS_IGNORE_FILES_ENABLED` to true, and you see the logs, you will see now how the command to be used is different, fixing the issue when the repository has too many files.

You can see how now the command `/usr/bin/git --no-optional-locks ls-files -s` is executed first to get the list of files, and then, `/usr/bin/git --no-optional-locks hash-object <base-path>/.okteto/git-untracked-backend-1742893596755404000` to calculate the SHA of the directory. It is using a temporary file to store the filtered output of the `git ls-files` command to prevent the issue of passing a huge argument to the exec command

### DIFF
--- a/pkg/repository/git.go
+++ b/pkg/repository/git.go
@@ -35,7 +35,8 @@ import (
 )
 
 const (
-	gitOperationTimeout = 5 * time.Second
+	gitOperationTimeout    = 5 * time.Second
+	limitDiffSizeToLogInMB = 1024 * 1024
 )
 
 var (
@@ -364,7 +365,13 @@ func (r gitRepoController) GetDiffHash(contextDir string) (string, error) {
 	}
 
 	hashFrom := fmt.Sprintf("%s-%s", diffResponse.diff, untrackedFilesResponse.untrackedFilesDiff)
-	oktetoLog.Infof("hashing diff: %s", hashFrom)
+	logDiff := hashFrom
+	if len(hashFrom) > limitDiffSizeToLogInMB {
+		logDiff = hashFrom[:limitDiffSizeToLogInMB]
+	}
+
+	// As the diff might be huge, setting an arbitrary limit in the chars we print in the logs, if not, it could cause performance issues
+	oktetoLog.Infof("hashing diff: %v", logDiff)
 	diffHash := sha256.Sum256([]byte(hashFrom))
 	return hex.EncodeToString(diffHash[:]), nil
 }
--- a/pkg/repository/git.go
+++ b/pkg/repository/git.go
@@ -14,510 +14,517 @@
 package repository
 
 import (
-	"context"
-	"crypto/sha256"
-	"encoding/hex"
-	"errors"
-	"fmt"
-	"path/filepath"
-	"sort"
-	"strings"
-	"time"
+    "context"
+    "crypto/sha256"
+    "encoding/hex"
+    "errors"
+    "fmt"
+    "path/filepath"
+    "sort"
+    "strings"
+    "time"
 
-	giturls "github.com/chainguard-dev/git-urls"
-	"github.com/go-git/go-git/v5"
-	"github.com/go-git/go-git/v5/plumbing"
-	"github.com/go-git/go-git/v5/plumbing/object"
-	"github.com/okteto/okteto/pkg/env"
-	"github.com/okteto/okteto/pkg/ignore"
-	oktetoLog "github.com/okteto/okteto/pkg/log"
-	"github.com/spf13/afero"
+    giturls "github.com/chainguard-dev/git-urls"
+    "github.com/go-git/go-git/v5"
+    "github.com/go-git/go-git/v5/plumbing"
+    "github.com/go-git/go-git/v5/plumbing/object"
+    "github.com/okteto/okteto/pkg/env"
+    "github.com/okteto/okteto/pkg/ignore"
+    oktetoLog "github.com/okteto/okteto/pkg/log"
+    "github.com/spf13/afero"
 )
 
 const (
-	gitOperationTimeout = 5 * time.Second
+    gitOperationTimeout = 5 * time.Second
 )
 
 var (
-	errNotCleanRepo    = errors.New("repository is not clean")
-	errTimeoutExceeded = errors.New("timeout exceeded")
-	errFindingRepo     = errors.New("top level git repo directory cannot be found")
+    errNotCleanRepo    = errors.New("repository is not clean")
+    errTimeoutExceeded = errors.New("timeout exceeded")
+    errFindingRepo     = errors.New("top level git repo directory cannot be found")
 )
 
 type gitRepoController struct {
-	repoGetter repositoryGetterInterface
-	fs         afero.Fs
-	ignorer    func(subpath string) ignore.Ignorer
-	path       string
+    repoGetter        repositoryGetterInterface
+    fs                afero.Fs
+    ignorer           func(subpath string) ignore.Ignorer
+    path              string
+    shouldIgnoreFiles bool
 }
 
 func newGitRepoController(path string) gitRepoController {
-	return gitRepoController{
-		repoGetter: gitRepositoryGetter{},
-		path:       path,
-		fs:         afero.NewOsFs(),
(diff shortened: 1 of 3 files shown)


pkg/repository/git.go
```diff
@@ -320,22 +319,31 @@ func (r gitRepoController) GetDiffHash(contextDir string) (string, error) {
...
+					}
+					if !shouldIgnore {
+						filteredUntrackedFiles = append(filteredUntrackedFiles, f)
+					} else {
+						oktetoLog.Debugf("skipping %v file change in GetDiffHash based on known ignore files", f)
+					}
+
+				}
+
+				untrackedFiles = filteredUntrackedFiles
```
ifbyol, line 342: This is the new path. It is only considered when the feature flag is enabled

pkg/repository/local_git.go
```diff
@@ -345,24 +363,104 @@ func (lg *LocalGit) GetDirContentSHA(ctx context.Context, gitPath, dirPath strin
...
+		}
+	}()
+
+	_, err = f.WriteString(filteredLS)
+	if err != nil {
+		oktetoLog.Debugf("failed to write untracked file: %v", err)
+		return "", err
+	}
+
+	hashObjectCmdArgs := []string{"--no-optional-locks", "hash-object", f.Name()}
```
ifbyol, line 386: This is the new logic to create a temporary file with the output of the ls-files command, instead of doing a pipe (the problem with big repos is that we were exceeding the maximum length of an argument

pkg/repository/local_git.go
```diff
@@ -345,24 +363,104 @@ func (lg *LocalGit) GetDirContentSHA(ctx context.Context, gitPath, dirPath strin
...
 
 	if err != nil {
+		oktetoLog.Debugf("failed to calculate hash of directory %s: %v", dirPath, err)
 		return handleError(err)
 	}
 	return string(output), nil
 }
 
+// getDirContentSHAWithoutIgnore calculates the SHA of the content of the given directory using git ls-files and git hash-object
+func (lg *LocalGit) getDirContentSHAWithoutIgnore(ctx context.Context, gitPath, dirPath string, fixAttempt int) (string, error) {
```
ifbyol, line 397: It is the content of the function `GitDirContent` before the change. It can be found [here](https://github.com/okteto/okteto/blob/c33b3258595857747075f99ebf6508569a8665de/pkg/repository/local_git.go#L273)

pkg/repository/local_git.go
```diff
@@ -345,24 +363,104 @@ func (lg *LocalGit) GetDirContentSHA(ctx context.Context, gitPath, dirPath strin
...
 
+	if lg.shouldIgnoreFiles {
+		return lg.diffWithIgnore(ctx, gitPath, dirPath, fixAttempt)
+	}
+
+	return lg.diffWithoutIgnore(ctx, gitPath, dirPath, fixAttempt)
+}
+
+// diffWithoutIgnore calculates the diff of the repository at the given path
+func (lg *LocalGit) diffWithoutIgnore(ctx context.Context, gitPath, dirPath string, fixAttempt int) (string, error) {
```
ifbyol, line 436: It is the implementation of the function `Diff` as it was before the problematic changes. It can be found [here](https://github.com/okteto/okteto/blob/c33b3258595857747075f99ebf6508569a8665de/pkg/repository/local_git.go#L302)

pkg/repository/local_git_test.go
```diff
@@ -20,9 +20,39 @@ import (
...
 	"time"
 
+	"github.com/okteto/okteto/pkg/ignore"
+	"github.com/spf13/afero"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/mock"
+	"github.com/stretchr/testify/require"
 )
 
+type mockLocalGit struct {
```
ifbyol, line 30: Even if we already have a `mockLocalExec`, I wasn't be able to use it as I had to mock 2 calls to `RunCommand` in the same test, and it wasn't allowing it. So, instead of modifying all the tests, I created another implementation using testify mocks for simplicity

pkg/repository/local_git.go
```diff
@@ -345,24 +363,104 @@ func (lg *LocalGit) GetDirContentSHA(ctx context.Context, gitPath, dirPath strin
...
+	f, err := lg.fs.Create(filepath.Join(config.GetOktetoHome(), fmt.Sprintf("git-untracked-%s-%d", filepath.Base(dirPath), time.Now().UnixNano())))
+	if err != nil {
+		oktetoLog.Debugf("failed to create untracked file: %v", err)
+		return "", err
+	}
+
+	defer func() {
+		err := lg.fs.Remove(f.Name())
+		if err != nil {
+			oktetoLog.Debugf("failed to close untracked file: %v", err)
```
jLopezbarb, line 376:
```suggestion
			oktetoLog.Debugf("failed to remove untracked file: %v", err)
```

pkg/repository/local_git.go
```diff
@@ -345,24 +363,104 @@ func (lg *LocalGit) GetDirContentSHA(ctx context.Context, gitPath, dirPath strin
 
 	filteredLS := strings.Join(filteredLines, "\n")
 
-	hashObjectCmdArgs := []string{"--no-optional-locks", "hash-object", "--stdin"}
-	output, err := lg.exec.RunPipeCommands(ctx, gitPath, "echo", []string{filteredLS}, lg.gitPath, hashObjectCmdArgs)
+	// As the list of files might be too long, we write it to a file and calculate the hash of the file
+	f, err := lg.fs.Create(filepath.Join(config.GetOktetoHome(), fmt.Sprintf("git-untracked-%s-%d", filepath.Base(dirPath), time.Now().UnixNano())))
```
jLopezbarb, line 367: I don't really know if I like this. I think it should be preferable to write into a temple instead of okteto home. 

If we have a problem removing it if we do it in a temp file when the computer restarts it will be deleted, while if we do this in the okteto home we could be increasing the folder

ifbyol: We have this problem in other parts of the CLI like the custom kubeconfig for the deploys. We should unify and standardize how we do it, but I don't know if we should do it as part of this PR or not

jLopezbarb (approved):

PR #4699, opened 2025-04-10
PROD-344: Add predefined Okteto environment variables to diverted pods, so it can be known at runtime time if they are in a diverted namespace or not

ifbyol:
PROD-344

## Description

This PR adds two environment variables to the divert drivers (nginx and istio) to provide information about the shared and diverted environments:

* OKTETO_SHARED_ENVIRONMENT: Contains the namespace of the shared environment (d.divert.Namespace)
* OKTETO_DIVERTED_ENVIRONMENT: Contains the namespace of the diverted environment (d.namespace)

### Changes made:

1. Added two new constants in pkg/constants/constants.go:
   - OktetoSharedEnvironmentEnvVar
   - OktetoDivertedEnvironmentEnvVar

2. Modified the UpdatePod function in both divert drivers:
   - pkg/divert/nginx/divert.go
   - pkg/divert/istio/divert.go

3. Added a helper function updateEnvVar to both drivers to handle adding or updating environment variables.

4. Added comprehensive unit tests for both drivers:
   - Test_updateEnvVar: Tests the helper function
   - Test_UpdatePod: Tests the main functionality with various scenarios

### Testing:
- Added unit tests for all new functionality
- All existing tests continue to pass

This change will allow applications running in diverted environments to be aware of both the shared and diverted namespaces, which can be useful for various integration scenarios.

### DIFF
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -14,136 +14,142 @@
 package constants
 
 const (
-	// OktetoNameEnvVar defines if the command is running inside okteot
-	OktetoNameEnvVar = "OKTETO_NAME"
+        // OktetoNameEnvVar defines if the command is running inside okteot
+        OktetoNameEnvVar = "OKTETO_NAME"
 
-	// DevLabel indicates the deployment is in dev mode
-	DevLabel = "dev.okteto.com"
+        // DevLabel indicates the deployment is in dev mode
+        DevLabel = "dev.okteto.com"
 
-	// OktetoURLAnnotation indicates the okteto cluster public url
-	OktetoURLAnnotation = "dev.okteto.com/url"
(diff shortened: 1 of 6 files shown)


jLopezbarb:
Tested and it's working but I have one question.

If I do a `okteto up` should the init containers (the ones generated by okteto) have the env vars?

ifbyol:
> Tested and it's working but I have one question.
> 
> If I do a `okteto up` should the init containers (the ones generated by okteto) have the env vars?

I don't think so. The idea of having those environment variables is that user's application might used them in some scenarios, like to decide to which queue should be connected based on a message header. In our init containers, at least for now, we don't need those environment variables, as our logic doesn't need them. But for the ones specified by users, might be necessary, for example, if they want to pre-load some data in a database

jLopezbarb (approved):

PR #4700, opened 2025-04-10
Add unit tests to test section in okteto manifest

oktetobot:
This PR adds a dedicated test section to the Okteto manifest to improve the testing workflow. The changes include:

1. Modified the `.oktetoignore` file to optimize file synchronization for testing
2.. Added setup commands to create machine ID files needed for the analytics tests to pass in container environments

These changes make it easier to run specific tests in isolation and ensure that tests that depend on machine IDs work correctly in containerized environments. The test section in the Okteto manifest provides a standardized way to run tests that is consistent with Okteto documentation and best practices.

The PR addresses the issue where the `Test_generatedMachineID` test was failing in container environments due to the inability to generate a proper machine ID. By creating the necessary machine ID files before running the tests, we ensure that the tests pass consistently regardless of the environment they are running in.

### DIFF
--- a/okteto.yml
+++ b/okteto.yml
@@ -30,6 +30,8 @@ test:
       - /go/pkg/
       - /root/.cache/go-build/
     commands:
+      # We need to set the machine-id in order to run the tests that require dbus
       - mkdir -p /var/lib/dbus && echo "11111111111111111111111111111111" | tee /var/lib/dbus/machine-id && chmod 644 /var/lib/dbus/machine-id
+      # We need to initialize the git repository in order to check that the annotation dev.okteto.com/sample is set
       - git init && git remote add origin https://github.com/okteto/okteto.git
       - make test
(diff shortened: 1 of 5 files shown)


okteto.yml
```diff
@@ -20,3 +20,16 @@ dev:
...
+    image: okteto/golang:1
+    artifacts:
+      - coverage.txt
+      - coverage.html
+    caches:
+      - /go/pkg/
+      - /root/.cache/go-build/
+    commands:
+      - mkdir -p /var/lib/dbus && echo "11111111111111111111111111111111" | tee /var/lib/dbus/machine-id && chmod 644 /var/lib/dbus/machine-id
+      - git init && git remote add origin https://github.com/okteto/okteto.git
```
ifbyol, line 34: Why are these 2 lines needed? I assume is because unit test need them for some reason. 

If this is needed, it is probably because our unit tests are not so "unit" as we believe, so we might need to take a loot at it at some point (not as part of this PR). But could we at least add a comment on each line indicating why is needed? In that way, we don't forget about it or take it into account when working with the tests

ifbyol (approved):

PR #4701, opened 2025-04-11
PROD-343: Change when the divert deploy logic is called to avoid problems when endpoints are defined in the manifest

ifbyol:
# Proposed changes

PROD-343

If you had a manifest with Divert, and then, you specify also the endpoints section, the update of the ingresses when the endpoints are deployed was making that the divert annotations were missing from the ingresses, so the header wasn't being propagated.

So, this started to happen when we re-organized the remote vs local deploy, as the divert deploy started to happen first. What I'm doing in this PR is to change the order, so, in the deployable phase (the one that can be happened locally or remotely, and the one running the proxy), we only apply the divert logic related to the proxy, but we don't call the `Deploy` function of the driver. That is started to be called at the end of the main flow (things that always happen locally), as the last stage of the deploy, so we ensure that the compose deployment (we already had issues in the past), and the endpoints deployment don't affect the divert logic. 

I have been checking, and there is no need of executing the divert deploy in the proxy.

I was testing all the output from the UI and CLI, and the output looks good too.

## How to validate

This repository can be used to validate the changes: https://github.com/ifbyol/test-okteto-divert

Follow the next steps:
* First, use the current CLI you have installer (without these changes)
* Deploy the main Okteto Manifest of the repo in a namespace
* Then, change the `.env` file to et the namespace where you deployed the whole application
* Then, deploy specifying `okteto.serviceb.yaml` as manifest in a new namespace

If you access to the endpoint `https://test-okteto-divert-{rest-of-domain}/servicea/chain`, you will see how three services are reported from the common namespace (which is wrong), and if you check the ingress generated in the last namespace, you will see how the divert annotations are missing. Also, you can see how the divert stage is executed before the "deploy compose"

Now, build the CLI from this branch and follow the same steps described above (recommendation is to deploy it from scratch). 

In this case, you will see that the endpoint works fine. If you check the ingress, you will see how the divert annotation is applied, and you can see how the divert stage is running **after** the "deploy compose" one. So, any endpoint generated from the deploy compose phase is also updated accordingly when the divert phase is executed

### DIFF
--- a/pkg/deployable/deploy.go
+++ b/pkg/deployable/deploy.go
@@ -40,11 +40,6 @@ const (
 	deployCommandsPhaseName = "commands"
 )
 
-// DivertDeployer defines the operations to deploy the divert section of a deployable
-type DivertDeployer interface {
-	Deploy(ctx context.Context) error
-}
-
 // ProxyInterface defines the different operations to work with the proxy run by Okteto
 type ProxyInterface interface {
 	Start()
@@ -87,7 +82,6 @@ type DeployRunner struct {
 	Executor           executor.ManifestExecutor
 	K8sClientProvider  okteto.K8sClientProviderWithLogger
 	Fs                 afero.Fs
-	DivertDeployer     DivertDeployer
 	GetExternalControl func(cfg *rest.Config) ExternalResourceInterface
 	k8sLogger          *io.K8sLogger
 	TempKubeconfigFile string
@@ -223,7 +217,6 @@ func (r *DeployRunner) RunDeploy(ctx context.Context, params DeployParameters) e
 			return err
 		}
 		r.Proxy.SetDivert(driver)
-		r.DivertDeployer = driver
 	}
 	r.Proxy.InitTranslator()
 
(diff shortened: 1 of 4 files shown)


pkg/deployable/deploy.go
```diff
@@ -334,16 +327,6 @@ func (r *DeployRunner) runCommandsSection(ctx context.Context, params DeployPara
...
-	// deploy divert if any
-	if params.Deployable.Divert != nil && params.Deployable.Divert.Namespace != params.Namespace {
-		oktetoLog.SetStage("Deploy Divert")
-		if err := r.DivertDeployer.Deploy(ctx); err != nil {
-			oktetoLog.AddToBuffer(oktetoLog.ErrorLevel, "error creating divert: %s", err.Error())
-			return err
-		}
-		oktetoLog.SetStage("")
-	}
-
```
ifbyol, line 346: The change is to move this from there to `cmd/deploy/deploy.go`

jLopezbarb (approved):

PR #4704, opened 2025-04-21
Add git binary for the dockerfile image

jLopezbarb:

# Proposed changes

Fixes DEV-893

Add static git to the final image so it can be used by the okteto cli for smartbuilds

## CLI Quality Reminders 🔧


ifbyol (commented):
I think we need to install something else in the final image, as I'm not able to clone via SHH nor HTTPS

jLopezbarb:
> I think we need to install something else in the final image, as I'm not able to clone via SHH nor HTTPS

Spoke offline. Given that we don't need git clone for build we are not going to do it for this first iteration. We will review it in the future if needed

PR #4705, opened 2025-04-21
Option 2: add spinner forchecking services

jLopezbarb:
# Proposed changes

Fixes DEV-861

This PR has been created with cindy.

Adds a new spinner when is checking if a image has been built

## How to validate

1. Run okteto build with a bad connection
2. Run okteto build with a repo that has several build

## CLI Quality Reminders 🔧


ifbyol (commented):
I think we will need to move the info message `Okteto Smart Builds is skipping build of 'xxxx' because it's already built from cache` to a bit later, just after we copy the image to the dev registry. If not, there is a moment where we say that the `api` image was skipped, but the spinner still shows that is checking if the `api` image was built.

ifbyol (approved):

PR #4716, opened 2025-05-13
allow to get dependency builds env vars

jLopezbarb:

# Proposed changes

Fixes DEV-940

When building and using smart build store the build env vars in the configmap so it can be used by any dependant service. The name of the variable is `fmt.Sprintf("OKTETO_DEPENDENCY_%s_BUILD_%s_%s", dependencyName, build, buildEnvVar)`

## How to validate

1. Check the following repos: 
- https://github.com/jLopezbarb/hello-world-nochart
- https://github.com/jLopezbarb/hello-world-chart
- https://github.com/jLopezbarb/hola-mundo-nochart
2. Using this CLI as the CLI for the backend deploy the https://github.com/jLopezbarb/hello-world-chart and check that is working correctly

## CLI Quality Reminders 🔧


pkg/cmd/pipeline/translate.go
```diff
@@ -43,19 +43,20 @@ import (
...
+	outputField      = "output"
+	repoField        = "repository"
+	branchField      = "branch"
+	filenameField    = "filename"
+	yamlField        = "yaml"
+	iconField        = "icon"
+	actionLockField  = "actionLock"
+	actionNameField  = "actionName"
+	variablesField   = "variables"
+	buildEnvVarField = "build-envs"
```
ifbyol, line 57: Have you checked if some change is needed in the backend to avoid to lose this new fields in redeploys or similar?

ifbyol, line 57: Also, we are using usually camelcase names for the fields. I would stick to it. `dev-branch` is the only exception, and we should have followed the same pattern

cmd/deploy/deploy.go
```diff
@@ -735,6 +735,17 @@ func (dc *Command) deployDependencies(ctx context.Context, deployOptions *Option
 		if err := dc.PipelineCMD.ExecuteDeployPipeline(ctx, pipOpts); err != nil {
 			return err
 		}
+		if dep.Wait {
+			depBuildEnvs, err := dc.CfgMapHandler.GetDependencyBuildEnvVars(ctx, depName, deployOptions.Namespace)
+			if err != nil {
+				return fmt.Errorf("could not get dependency build env vars: %w", err)
+			}
+			for key, val := range depBuildEnvs {
+				if err := os.Setenv(key, val); err != nil {
```
ifbyol, line 744: Have you checked if this works for remote deployments?

ifbyol (commented):
I would consider to add an e2e for this case, as this might be a tricky case, but it has to be done once the CLI is integrated in the backend

pkg/cmd/pipeline/translate.go
jLopezbarb: I tested and it seems to be working, also I couldn't find any reference where we might lose this

jLopezbarb:
> I would consider to add an e2e for this case, as this might be a tricky case, but it has to be done once the CLI is integrated in the backend

Created a ticket for the next sprint

cmd/deploy/deploy.go
jLopezbarb: They are passed as part of the Okteto dependency variables. We inject all the variables with the prefix `OKTETO_DEPENDENCY_`. I already tested that is working :)

ifbyol (approved):

PR #4718, opened 2025-05-16
feat: add destroy with dependencies on CLI

jLopezbarb:

# Proposed changes

Fixes DEV-941

Adds the ability to run `okteto pipeline destroy --dependencies` from the CLI and call the correct graphql endpoint

## How to validate

With a cluster that has the correct configuration and CLI

1. Run okteto pipeline deploy on a repo that has dependencies in the manifest
1. Run okteto pipeline destroy --dependencies
1. Check that the devenvs have been eliminated

## CLI Quality Reminders 🔧

### DIFF
--- a/cmd/pipeline/destroy.go
+++ b/cmd/pipeline/destroy.go
@@ -34,21 +34,23 @@ import (
 
 // destroyFlags represents the user input for a pipeline destroy command
 type destroyFlags struct {
-	name           string
-	k8sContext     string
-	namespace      string
-	wait           bool
-	destroyVolumes bool
-	timeout        time.Duration
+	name                string
+	k8sContext          string
+	namespace           string
+	wait                bool
+	destroyVolumes      bool
+	timeout             time.Duration
+	destroyDependencies bool
 }
 
 // DestroyOptions options to destroy pipeline command
 type DestroyOptions struct {
-	Name           string
-	Namespace      string
-	Wait           bool
-	DestroyVolumes bool
-	Timeout        time.Duration
+	Name                string
+	Namespace           string
+	Wait                bool
+	DestroyVolumes      bool
+	Timeout             time.Duration
+	DestroyDependencies bool
 }
 
 func destroy(ctx context.Context) *cobra.Command {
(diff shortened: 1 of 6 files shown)


cmd/pipeline/destroy.go
```diff
@@ -89,6 +91,7 @@ okteto pipeline destroy --wait=false`,
 	cmd.Flags().BoolVarP(&flags.wait, "wait", "w", true, "wait until the Development Environment is destroyed")
 	cmd.Flags().BoolVarP(&flags.destroyVolumes, "volumes", "v", false, "destroy persistent volumes created by the Development Environment")
 	cmd.Flags().DurationVarP(&flags.timeout, "timeout", "t", fiveMinutes, "the duration to wait for the Development Environment to be destroyed. Any value should contain a corresponding time unit e.g. 1s, 2m, 3h")
+	cmd.Flags().BoolVar(&flags.destroyDependencies, "dependencies", false, "destroy repositories in the 'dependencies' section")
```
ifbyol, line 94: Should we use the same description in all the `--dependencies` flag of destroy commands? The one in `okteto destroy` has another description

pkg/okteto/pipeline.go
```diff
@@ -240,52 +248,79 @@ func (c *pipelineClient) GetByName(ctx context.Context, name, namespace string)
...
-			Repository: string(mutation.Response.GitDeploy.Repository),
-			Status:     string(mutation.Response.GitDeploy.Status),
+			ID:         string(response.GitDeploy.Id),
+			Name:       string(response.GitDeploy.Name),
+			Repository: string(response.GitDeploy.Repository),
+			Status:     string(response.GitDeploy.Status),
 		}
 	} else {
-		var mutation destroyPipelineWithoutVolumesMutation
+		var mutation destroyPipelineWithoutVolumesMutationAndDependencies
```
ifbyol, line 291: Minor thing, but why don't we unify both branch and just send `destroyVolumes: false` when the destroy flag is not set? It would simplify the code. Not needed to tackle, but wondering why it is done like that?

jLopezbarb: We need to handle the case where the graphql doesn't support the flag, that's why I have to create 2 new structs

ifbyol (approved):
I tested it and it is working fine. I'm approving it as the 2 comments I added are minor things, so they don't block the PR

PR #4720, opened 2025-05-20
Ensure kubetoken output includes kind

codyjlandstrom:
## Summary
- ensure kubetoken serializer sets `kind` and `apiVersion`
- add test for serializer

## Testing
- `go test ./...` *(fails: Test_translateService etc.)*

### DIFF
--- a/cmd/kubetoken/kubetoken.go
+++ b/cmd/kubetoken/kubetoken.go
@@ -25,6 +25,7 @@ import (
 	"github.com/okteto/okteto/pkg/types"
 	"github.com/spf13/cobra"
 	"k8s.io/client-go/kubernetes"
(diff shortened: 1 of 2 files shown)


ifbyol:
@codyjlandstrom I think for now, we could decline this. The change is a bit bigger and it might imply a breaking change given that we are not returning the right format. The output doesn't satisfy the expected spec, and we would need to review more carefully

jLopezbarb:
Hey @codyjlandstrom, Should we close it and prioritise it accordingly?

codyjlandstrom:
> Hey @codyjlandstrom, Should we close it and prioritise it accordingly?

Yes, lets go ahead and close this for now

PR #4725, opened 2025-05-28
feat: enhance deploy and destroy pipelines with dependency handling

jLopezbarb:
- Added support for managing dependencies in both deploy and destroy commands.
- Introduced flags to control redeployment and destruction of dependencies.
- Updated pipeline mutations to include an `isDependency` argument for better context handling.
- Enhanced logging and error handling for dependency operations.

This update improves the flexibility and control over dependency management during pipeline operations.

# Proposed changes

Fixes DEV-961

This PR fixes the infinite redeployment when the setting is activated in the backend 

## How to validate

1. Using a cluster set the redeploy dependencies to true and try running `okteto pipeline deploy/destroy` with dependencies and without them 

## CLI Quality Reminders 🔧

### DIFF
--- a/cmd/deploy/deploy.go
+++ b/cmd/deploy/deploy.go
@@ -706,7 +706,6 @@ func (dc *Command) deployDependencies(ctx context.Context, deployOptions *Option
 	for depName, dep := range deployOptions.Manifest.Dependencies {
 		oktetoLog.Information("Deploying dependency  '%s'", depName)
 		oktetoLog.SetStage(fmt.Sprintf("Deploying dependency %s", depName))
-
 		if err := validator.CheckReservedVarName(dep.Variables); err != nil {
 			return err
 		}
@@ -730,6 +729,7 @@ func (dc *Command) deployDependencies(ctx context.Context, deployOptions *Option
 			Timeout:      dep.GetTimeout(deployOptions.Timeout),
 			SkipIfExists: !deployOptions.Dependencies,
 			Namespace:    okteto.GetContext().Namespace,
+			IsDependency: true,
 		}
 
 		if err := dc.PipelineCMD.ExecuteDeployPipeline(ctx, pipOpts); err != nil {
@@ -746,6 +746,7 @@ func (dc *Command) deployDependencies(ctx context.Context, deployOptions *Option
 				}
 			}
 		}
+		oktetoLog.SetStage("")
 	}
 	oktetoLog.SetStage("")
 	return nil
--- a/cmd/pipeline/deploy.go
+++ b/cmd/pipeline/deploy.go
@@ -30,6 +30,7 @@ import (
 	"github.com/okteto/okteto/pkg/cmd/pipeline"
 	"github.com/okteto/okteto/pkg/constants"
 	"github.com/okteto/okteto/pkg/devenvironment"
+	"github.com/okteto/okteto/pkg/env"
 	oktetoErrors "github.com/okteto/okteto/pkg/errors"
 	"github.com/okteto/okteto/pkg/k8s/configmaps"
 	oktetoLog "github.com/okteto/okteto/pkg/log"
@@ -66,6 +67,7 @@ type deployFlags struct {
 	skipIfExists         bool
 	reuseParams          bool
 	redeployDependencies bool
+	dependenciesIsSet    bool
 }
 
 // DeployOptions represents options for deploy pipeline command
@@ -82,6 +84,8 @@ type DeployOptions struct {
 	SkipIfExists         bool
 	ReuseParams          bool
 	RedeployDependencies bool
+	IsDependency         bool
+	DependenciesIsSet    bool
 }
 
 func deploy(ctx context.Context) *cobra.Command {
@@ -93,7 +97,7 @@ func deploy(ctx context.Context) *cobra.Command {
 		Example: `To run the deploy without the Okteto CLI wait for its completion, use the '--wait=false' flag:
 okteto pipeline deploy --wait=false`,
 		RunE: func(cmd *cobra.Command, args []string) error {
-
+			flags.dependenciesIsSet = cmd.Flags().Changed("dependencies")
 			if err := validator.CheckReservedVariablesNameOption(flags.variables); err != nil {
 				return err
 			}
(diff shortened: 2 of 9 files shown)


cmd/pipeline/destroy.go
```diff
@@ -124,7 +130,23 @@ func (pc *Command) ExecuteDestroyPipeline(ctx context.Context, opts *DestroyOpti
 	return nil
 }
 
-func (pc *Command) destroyPipeline(ctx context.Context, name, namespace string, destroyVolumes, destroyDependencies bool) (*types.GitDeployResponse, error) {
+func shouldDestroyDependencies(opts *DestroyOptions) bool {
+	isForceRedeployDependenciesSetInOktetoInstance := env.LoadBoolean(constants.OktetoForceDestroyDependencies)
```
ifbyol, line 134: We don't have yet the code in the CLI to set this `OktetoForceDestroyDependencies` based on the server information, right?

cmd/pipeline/destroy.go
```diff
@@ -124,7 +130,23 @@ func (pc *Command) ExecuteDestroyPipeline(ctx context.Context, opts *DestroyOpti
...
+	if !isInsideDependency && isForceRedeployDependenciesSetInOktetoInstance {
+		// the user forces --dependencies=false
+		if opts.DependenciesIsSet && !opts.DestroyDependencies {
+			return false
+		}
+		return true
+	}
+	if isInsideDependency {
+		return false
+	}
```
ifbyol, line 145: if you move this check before the line 136, we get an early return, and the condition in the if in the line `136` would be simpler

cmd/pipeline/destroy.go
ifbyol, line 134:
```suggestion
	isForceDestroyDependenciesSetInOktetoInstance := env.LoadBoolean(constants.OktetoForceDestroyDependencies)
```

cmd/pipeline/deploy.go
```diff
@@ -249,6 +254,26 @@ func (pc *Command) ExecuteDeployPipeline(ctx context.Context, opts *DeployOption
...
+	isInsideDependency := env.LoadBoolean(constants.OktetoIsDependencyEnvVar)
+	if !isInsideDependency && isForceRedeployDependenciesSetInOktetoInstance {
+		// the user forces --dependencies=false
+		if opts.DependenciesIsSet && !opts.RedeployDependencies {
+			return false
+		}
+		return true
+	}
+
+	if isInsideDependency {
```
ifbyol, line 270: Same about moving this condition first in the function, so condition in line `262` would be simpler

cmd/deploy/deploy.go
```diff
@@ -746,6 +746,7 @@ func (dc *Command) deployDependencies(ctx context.Context, deployOptions *Option
 				}
 			}
 		}
+		oktetoLog.SetStage("")
```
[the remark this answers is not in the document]
jLopezbarb: I could have done it as part of another PR but it's needed because when redeploying we introduce one of the information message from the following dependency on the same stage as the previous one. I opted to include it here since it was a small fix 
Screenshot 2025-05-29 at 12 49 52

cmd/pipeline/destroy.go
jLopezbarb: This variable comes from the backend in the same way as `OKTETO_FORCE_REMOTE` does for the remote flag

cmd/pipeline/deploy.go
```diff
@@ -66,6 +67,7 @@ type deployFlags struct {
 	skipIfExists         bool
 	reuseParams          bool
 	redeployDependencies bool
+	dependenciesIsSet    bool
```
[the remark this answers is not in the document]
jLopezbarb: I wanted to follow the same approach as we have in the deploy command, but I agree that we should completely remove it and only use it on the options

cmd/pipeline/deploy.go
```diff
@@ -249,6 +253,27 @@ func (pc *Command) ExecuteDeployPipeline(ctx context.Context, opts *DeployOption
...
+// default behavior is set by cluster config, but can be overridden by the user using the flag --dependencies
+func shouldRedeployDependencies(opts *DeployOptions) bool {
+	isForceRedeployDependenciesSetInOktetoInstance := env.LoadBoolean(constants.OktetoForceRedeployDependencies)
+	isInsideDependency := env.LoadBoolean(constants.OktetoIsDependencyEnvVar)
+
+	if isInsideDependency {
+		return false
+	}
+
+	if !isInsideDependency && isForceRedeployDependenciesSetInOktetoInstance {
```
ifbyol, line 266:
```suggestion
	if isForceRedeployDependenciesSetInOktetoInstance {
```

This is not needed as here, we can say it is not inside a dependency because if not, it would have returned in line 263

In fact, I think this can be simplified to avoid different indentation levels:

```golang
func shouldRedeployDependencies(opts *DeployOptions) bool {
	isForceRedeployDependenciesSetInOktetoInstance := env.LoadBoolean(constants.OktetoForceRedeployDependencies)
	isInsideDependency := env.LoadBoolean(constants.OktetoIsDependencyEnvVar)

	if isInsideDependency {
		return false
	}

	if !isForceRedeployDependenciesSetInOktetoInstance {
		return opts.RedeployDependencies
	}

	// the user forces --dependencies=false
	if opts.DependenciesIsSet && !opts.RedeployDependencies {
		return false
	}
	return true
}
```

cmd/pipeline/destroy.go
```diff
@@ -124,7 +130,24 @@ func (pc *Command) ExecuteDestroyPipeline(ctx context.Context, opts *DestroyOpti
...
 }
 
-func (pc *Command) destroyPipeline(ctx context.Context, name, namespace string, destroyVolumes, destroyDependencies bool) (*types.GitDeployResponse, error) {
+func shouldDestroyDependencies(opts *DestroyOptions) bool {
+	isForceDestroyDependenciesSetInOktetoInstance := env.LoadBoolean(constants.OktetoForceDestroyDependencies)
+	isInsideDependency := env.LoadBoolean(constants.OktetoIsDependencyEnvVar)
+	if isInsideDependency {
+		return false
+	}
+	if !isInsideDependency && isForceDestroyDependenciesSetInOktetoInstance {
```
ifbyol, line 139:
```suggestion
	if isForceDestroyDependenciesSetInOktetoInstance {
```

Same here, this is not needed anymore after moving the other check, right? This can be simplified as the other function similar to this one

ifbyol (approved):
Approved as it is looking good, but please, review the 2 comments to make more readable (in my opinion) the functions that check if the redeploy should apply or not

PR #4731, opened 2025-06-12
fix: wait until sync completes to run services

jLopezbarb:

# Proposed changes

Fixes PROD-293

**Problem:**
We were not waiting for the sync to complete on the `dev[svcname].services`. When syncing large files or multiple files, this caused the service commands to execute before synchronization was finished, leading to failures and undesired behaviors.

**Solution:**
This PR fixes the issue by splitting the deployment of the dev environments into two separate steps: mainDev and services. We deploy the mainDev, synchronise the files and then start the rest of the services. This ensures that synchronization completes before the service commands run.

## How to validate

1. Using https://github.com/jLopezbarb/test-services
1. Create a big file, for example using dd: `dd if=/dev/zero of=bigfile.txt bs=1024k count=1024`
1. Run `okteto up`
1. Check that the web service could not start
1. Run again using this branch

## CLI Quality Reminders 🔧

### DIFF
--- a/cmd/up/devdeployer.go
+++ b/cmd/up/devdeployer.go
@@ -62,24 +62,11 @@ func newDevDeployer(translations map[string]*apps.Translation, k8sClient kuberne
 
 // deployMainDev deploys the main dev
 func (dd *devDeployer) deployMainDev(ctx context.Context) error {
-
-	err := dd.deploy(ctx, dd.mainTranslation)
-	if err != nil {
-		return err
-	}
-
-	if !dd.servicesUpWait {
-		return dd.deployDevServices(ctx)
-	}
-	return nil
+	return dd.deploy(ctx, dd.mainTranslation)
 }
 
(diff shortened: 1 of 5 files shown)


cmd/up/devdeployer.go
```diff
@@ -0,0 +1,107 @@
...
+		return dd.deployDevServices(ctx)
+	}
+	return nil
+}
+
+// deployDevServices deploys the dev services
+func (dd *devDeployer) deployDevServices(ctx context.Context) error {
+	if !dd.servicesUpWait {
+		return nil
+	}
```
ifbyol: Could we add a comment here mentioning that if the flag is false, the `deployMainDev` function would deploy them? It is not obvious seeing the variable name, and it is not everything self-contained in the same function. I mean, it is not obvious to remember that in a few months

ifbyol (approved):

PR #4759, opened 2025-08-22
Allow to override security context ReadOnlyRootFilesystem property on dev containers

ifbyol:
# Proposed changes

PROD-400

This PR includes changes to support the definition of `ReadOnlyRootFilesystem` within `securityContext` field in the Okteto manifest for dev containers. In this way, you can override the value of that property when setting a container in dev mode. It also modifies the schema.json to support that new property.

## How to validate

You can use the [movies repository](https://github.com/okteto/movies).
* Go to the api deployment template and change the template to add a `securityContext` with the property `ReadOnlyRootFilesystem` to true
* Deploy the application
* Then, go to the Okteto Manifest and for the api entry in the dev section, do the following changes:
  * Remove the line with `workdir`
  * Set the mode to hybrid
  * Remove the capabilities within the `securityContext`
* Execute `okteto up api` command with an official CLI. You will see an error indicating that there is no permissions to write in the pod
* Then, go to the Okteto Manifest again and change add  `ReadOnlyRootFilesystem` to `false` within `securityContext` to the api entry in the dev section
* Execute `okteto up api` again, this time it should start correctly


pkg/k8s/apps/translate.go
```diff
@@ -487,6 +487,10 @@ func TranslateContainerSecurityContext(c *apiv1.Container, s *model.SecurityCont
 		c.SecurityContext.AllowPrivilegeEscalation = s.AllowPrivilegeEscalation
 	}
 
+	if s.ReadOnlyRootFilesystem != nil {
+		c.SecurityContext.ReadOnlyRootFilesystem = s.ReadOnlyRootFilesystem
+	}
+
```
ifbyol, line 493: @pchico83 @jLopezbarb In the line 501 of this file, you can see how the `ReadOnlyRootFilesystem` property is being set to nil in the dev container if the capabilities are set in the Okteto Manifest. Is that expected? I left it to avoid a breaking change

pkg/schema/dev.go
```diff
@@ -408,6 +408,10 @@ func (dev) JSONSchema() *jsonschema.Schema {
 		Type:  &jsonschema.Type{Types: []string{"boolean"}},
 		Title: "allowPrivilegeEscalation",
 	})
+	securityContextProps.Set("readOnlyRootFilesystem", &jsonschema.Schema{
```
pchico83, line 411: Reminder to check if we need to do something else to update the schema everywhere

ifbyol: No need to do some extra step. The schema published in schemastore is available directly from github to the mast version. It is not versioned. https://github.com/SchemaStore/schemastore/blob/master/src/api/json/catalog.json#L4106

PR #4760, opened 2025-08-27
fix: wait for sse

jLopezbarb:
# Proposed changes

Fixes DEV-1029

Created a new file in order to wait for the sse in the okteto preview deploy/destroy in order to connect to the right job
Screenshot 2025-08-27 at 09 30 39

## How to validate

1. Run several okteto preview deploys until you get the message
1. Wait until it connects
1.
1.

## CLI Quality Reminders 🔧

### DIFF
--- a/cmd/preview/deploy.go
+++ b/cmd/preview/deploy.go
@@ -160,8 +160,8 @@ func (pw *Command) waitUntilRunning(ctx context.Context, name, namespace string,
 	wg.Add(1)
 	go func(wg *sync.WaitGroup) {
 		defer wg.Done()
-		err := pw.okClient.Stream().PipelineLogs(waitCtx, name, namespace, a.Name)
-		if err != nil {
+		err := streamPreviewLogsWithTimeout(waitCtx, pw.okClient, name, namespace, a.Name, timeout)
+		if err != nil && !errors.Is(err, context.Canceled) && !errors.Is(err, context.DeadlineExceeded) {
 			oktetoLog.Warning("preview logs cannot be streamed due to connectivity issues")
 			oktetoLog.Infof("preview logs cannot be streamed due to connectivity issues: %v", err)
 		}
--- a/cmd/preview/streaming.go
+++ b/cmd/preview/streaming.go
@@ -1,0 +1,117 @@
+// Copyright 2023 The Okteto Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package preview
+
+import (
+	"context"
+	"errors"
+	"math"
+	"time"
+
+	oktetoLog "github.com/okteto/okteto/pkg/log"
+	"github.com/okteto/okteto/pkg/types"
+)
+
+// streamFn represents a function that attempts to stream logs and returns an error
+type streamFn func(ctx context.Context) error
+
+// calculateExponentialBackoff calculates the next retry delay using exponential backoff
+func calculateExponentialBackoff(attempt int) time.Duration {
+	// Prevent overflow for very large attempts by capping at a reasonable value
+	// After attempt 6, we hit the 30s cap anyway, so no need to calculate large powers
+	if attempt > 10 {
+		attempt = 10
+	}
+	
+	// Use the same formula as the original stream package: (2^attempt - 1) * 0.5 seconds
+	baseDelaySeconds := (math.Pow(2, float64(attempt)) - 1) * 0.5
+	baseDelay := time.Duration(baseDelaySeconds * float64(time.Second))
+	
+	// Cap the maximum delay at 30 seconds 
+	maxDelay := 30 * time.Second
+	if baseDelay > maxDelay {
+		baseDelay = maxDelay
+	}
+	
+	// Ensure minimum delay of 500ms
+	minDelay := 500 * time.Millisecond
+	if baseDelay < minDelay {
+		baseDelay = minDelay
+	}
+	
+	return baseDelay
+}
+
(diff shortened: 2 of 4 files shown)


cmd/preview/streaming.go
```diff
@@ -0,0 +1,120 @@
...
+		baseDelay = minDelay
+	}
+
+	return baseDelay
+}
+
+// streamWithExponentialBackoff attempts to stream logs with exponential backoff until the context timeout
+func streamWithExponentialBackoff(ctx context.Context, timeout time.Duration, streamFunc streamFn, operationName string) error {
+	// Create a context with the user's timeout
+	timeoutCtx, cancel := context.WithTimeout(ctx, timeout)
```
ifbyol: This context should already have the wait timeout, right? In cmd/preview/deploy.go line 163, we pass `waitCtx`, which I assume is a context with the wait timeout

cmd/preview/streaming.go
```diff
@@ -0,0 +1,120 @@
...
+			// Log the retry attempt
+			oktetoLog.Warning("preview stream client not reachable, waiting to reconnect...")
+			oktetoLog.Infof("%s streaming error: %v, retrying in %v (attempt %d)", operationName, err, retryDelay, attempt)
+
+			// Wait before retrying, but respect context cancellation
+			select {
+			case <-timeoutCtx.Done():
+				return timeoutCtx.Err()
+			case <-time.After(retryDelay):
+				// Continue to next attempt
```
ifbyol: Would it make sense to do it only for a specific error like not-found (which means that the job couldn't be found) I don't recall which other errors might return the SSE endpoint, but there might be some legit case where it should fail instead of just retry

ifbyol:
I was reviewing this, and I see that the underlying streaming API defined in `stream.go` already has some retry mechanism. This is doing 2 retries per retry you added. Why don't we add a backoff in the current retry and increase the existing number of retries or something like that? It seems a bit too much to have this double retry mechanism. 

Also, makes sense to add it there because this is a problem that could also happen when deploying pipelines, and If I'm not wrong, they use the same functions

ifbyol:
> > I was reviewing this, and I see that the underlying streaming API defined in `stream.go` already has some retry mechanism. This is doing 2 retries per retry you added. Why don't we add a backoff in the current retry and increase the existing number of retries or something like that? It seems a bit too much to have this double retry mechanism.
> > Also, makes sense to add it there because this is a problem that could also happen when deploying pipelines, and If I'm not wrong, they use the same functions
> 
> I didn't want to change that because it's used on more sites rather than this like pipeline

But precisely, pipeline has also this same problem if the job takes more time to be scheduled. With this solution, we would be only solving it for previews.

Where else are we using that apart from pipeline and preview?

cmd/destroy/all.go
```diff
@@ -70,12 +73,14 @@ func (lda *localDestroyAllCommand) destroy(ctx context.Context, opts *Options) e
...
 		defer wg.Done()
 		exit <- lda.waitForNamespaceDestroyAllToComplete(waitCtx, opts.Namespace)
+		logsCtxCancel()
 	}(&wg)
 
 	wg.Add(1)
 	go func(wg *sync.WaitGroup) {
 		defer wg.Done()
-		err := lda.oktetoClient.Stream().DestroyAllLogs(waitCtx, opts.Namespace)
+		reconnectionTime := 5 * time.Minute
```
ifbyol:
```suggestion
		connectionTimeout := 5 * time.Minute
```

Maybe this is more accurate? at least the timeout

cmd/namespace/delete.go
```diff
@@ -118,12 +121,14 @@ func (nc *Command) watchDelete(ctx context.Context, namespace string, k8sLogger
...
 		defer wg.Done()
 		exit <- nc.waitForNamespaceDeleted(waitCtx, namespace, k8sLogger)
+		logsCtxCancel()
 	}(&wg)
 
 	wg.Add(1)
 	go func(wg *sync.WaitGroup) {
 		defer wg.Done()
-		err := nc.okClient.Stream().DestroyAllLogs(waitCtx, namespace)
+		reconnectionTime := 5 * time.Minute
```
ifbyol:
```suggestion
		connectionTimeout := 5 * time.Minute
```

Same here

cmd/namespace/delete.go
```diff
@@ -118,12 +121,14 @@ func (nc *Command) watchDelete(ctx context.Context, namespace string, k8sLogger
 	go func(wg *sync.WaitGroup) {
 		defer wg.Done()
 		exit <- nc.waitForNamespaceDeleted(waitCtx, namespace, k8sLogger)
+		logsCtxCancel()
```
ifbyol, line 124: Sorry for not being clear when we talk about it offline, but this should be done in all the operations that might stream, right? I mean for deploy operations too.

The scenario we talked about could also happen on deploys (deploy finish but the connection could not be stablished yet)

pkg/stream/stream.go
```diff
@@ -45,30 +41,52 @@ func request(c *http.Client, url string) (*http.Response, error) {
...
 
-		if attempts >= maxRetryAttempts {
-			return nil, fmt.Errorf("server disconnected, maxRetries reached")
-		}
+			// Calculate exponential backoff delay
+			attempts++
+			retryDelay := calculateExponentialBackoff(attempts)
+
+			// Log the retry attempt
+			oktetoLog.Warning("preview stream client not reachable, waiting to reconnect...")
```
ifbyol: This message is specific to preview, but this function is being used for the rest of operations. What about `unable to connect to stream  logs, waiting to reconnect....`

cmd/preview/deploy.go
```diff
@@ -160,8 +160,8 @@ func (pw *Command) waitUntilRunning(ctx context.Context, name, namespace string,
 	wg.Add(1)
 	go func(wg *sync.WaitGroup) {
 		defer wg.Done()
-		err := pw.okClient.Stream().PipelineLogs(waitCtx, name, namespace, a.Name)
-		if err != nil {
+		err := pw.okClient.Stream().PipelineLogs(waitCtx, name, namespace, a.Name, timeout)
+		if err != nil && !errors.Is(err, context.Canceled) && !errors.Is(err, context.DeadlineExceeded) {
```
ifbyol, line 166: Don't we have an issue in the goroutine in line `171`? We are not calling to `defer wg.Done()`, but we are adding 1 to the waitgroup just before. Am I missing something?

ifbyol (approved):

PR #4785, opened 2025-09-22
Fix Cache tests

jLopezbarb:
# Proposed changes

Fixes DEV-1069

Add a new id based on repositoryHash/manifestname/testname/testidx in order to generate a unique id and fix the cache in the remote 
Fix the tests accordingly

## How to validate

Run a okteto test with two tests contianers sharing the same cache like for example one caching node modules and then using it in the next one

## CLI Quality Reminders 🔧

### DIFF
--- a/pkg/remote/remote.go
+++ b/pkg/remote/remote.go
@@ -100,7 +100,7 @@ RUN echo "${{ .InvalidateCacheArgName }}" > /etc/.oktetocachekey
 RUN okteto registrytoken install --force --log-output=json
 
 RUN \
-  {{range $key, $path := .Caches }}--mount=type=cache,target={{$path}},sharing=private {{end}}\
+  {{range $key, $path := .Caches }}--mount=type=cache,id={{$.ManifestName}}-{{$.TestName}}-{{$key}},target={{$path}},sharing=private {{end}}\
   --mount=type=secret,id=known_hosts \
   --mount=type=secret,id={{ .SSHAgentSocketArgName }},env=SSH_AUTH_SOCK \
   mkdir -p $HOME/.ssh && echo "UserKnownHostsFile=/run/secrets/known_hosts $HOME/.ssh/known_hosts" >> $HOME/.ssh/config && \
@@ -177,6 +177,8 @@ type Params struct {
 	CommandFlags []string
 	Caches       []string
 	Hosts        []model.Host
+	// TestName is the name of the test container, used for cache isolation
+	TestName string
 	// IgnoreRules are the ignoring rules added to this build execution.
 	// Rules follow the .dockerignore syntax as defined in:
 	// https://docs.docker.com/build/building/context/#syntax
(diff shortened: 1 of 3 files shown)


pkg/remote/remote.go
```diff
@@ -100,7 +103,7 @@ RUN echo "${{ .InvalidateCacheArgName }}" > /etc/.oktetocachekey
 RUN okteto registrytoken install --force --log-output=json
 
 RUN \
-  {{range $key, $path := .Caches }}--mount=type=cache,target={{$path}},sharing=private {{end}}\
+  {{range $key, $path := .Caches }}--mount=type=cache,id={{$.RepoHash}}-{{$.ManifestName}}-{{$.TestName}}-{{$key}},target={{$path}},sharing=private {{end}}\
```
ifbyol: Why do we use the `$key`? If the order of caches changes, the cache won't be used, I think we should include the `$path` instead of the `$key`.

Also, is there a max length for the id values?

pkg/remote/remote.go
```diff
@@ -393,6 +401,17 @@ func (r *Runner) createDockerfile(tmpDir string, params *Params) (string, error)
 		return "", err
 	}
 
+	repoHash := ""
+	if params.Manifest != nil && params.Manifest.ManifestPath != "" {
+		repo := repository.NewRepository(params.Manifest.ManifestPath)
+		repoURL := repo.GetAnonymizedRepo()
```
ifbyol: Does this function always return a "stable" value of the repo? I mean, in case of the same repo being cloned using different URLs, would it return the same URL?

pkg/remote/remote.go
pchico83: +1 for using the cache path instead of "key".
About manifest name, what would be better, manifest name or manifest path?
Also, I would generate a hash of the final cache-id to avoid format issues

pkg/remote/remote.go
pchico83: Also, what's the value when the folder doesn't belong to a git repo?

pkg/remote/remote.go
jLopezbarb: > Also, is there a max length for the id values?

Not documented, I did several tests with different values and all seemed to work

> About manifest name, what would be better, manifest name or manifest path?

I think manifest name it's better because if we work with git worktrees you can have the same manifest but different path

pkg/remote/remote.go
jLopezbarb: > Does this function always return a "stable" value of the repo? I mean, in case of the same repo being cloned using different URLs, would it return the same URL?

Yes. It's done [here](https://github.com/okteto/okteto/blob/master/pkg/repository/repository.go#L48)

> Also, what's the value when the folder doesn't belong to a git repo?

It was empty string(`""`), now I have changed to the manifest path

pchico83 (approved):

PR #4787, opened 2025-09-29
fix: cves

jLopezbarb:
Fixes the CVES for the new version of the CLI


ifbyol:
Syncthing (golang dependency) and kustomize seems to have also some vulnerability, can they be fixed too?

PR #4789, opened 2025-10-01
Jlo/add cache id in run cmds

jLopezbarb:
# Proposed changes

Fixes DEV-1150

Refactor the translation of the Dockerfile so it can be tested.
Adds a new translation for the cache mounting:
If `mount type=cache && not contains id` we set it based on repo, folder and target

## How to validate

1. Use the following Dockerfiles:
``` Dockerile
# frontend
FROM okteto/node:20 as runner

USER 0

ARG AAA

COPY . /okteto/src
WORKDIR /okteto/src

ENV OKTETO_SMART_BUILDS_ENABLED="true"

ENV CI=true

RUN --mount=type=cache,target=./.eslintcache --mount=type=cache,target=./.yarn/cache,sharing=private --mount=type=cache,target=./node_modules,sharing=private yarn install --immutable
```
```Dockerfile
# backend
FROM node:24-slim as runner

USER 0

ARG AAA

COPY . /okteto/src
WORKDIR /okteto/src

ENV OKTETO_SMART_BUILDS_ENABLED="true"

ENV CI=true

RUN --mount=type=cache,target=./node_modules npm install
```
2. Build frontend using `okteto build` on its folder 
3. Build backend using `okteto build` on its folder
4. Create a new change on the frontend folder
5. Build frontend using `okteto build` on its folder 
6. See no error

## CLI Quality Reminders 🔧

### DIFF
--- a/pkg/cmd/build/file.go
+++ b/pkg/cmd/build/file.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strings"
 
 	"github.com/okteto/okteto/pkg/config"
@@ -32,8 +33,8 @@ const (
 )
 
 // GetDockerfile returns the dockerfile with the cache and registry translations
-func GetDockerfile(dockerFile string, okCtx OktetoContextInterface) (string, error) {
-	file, err := getTranslatedDockerFile(dockerFile, okCtx)
+func GetDockerfile(dockerFile string, okCtx OktetoContextInterface, projectHash string) (string, error) {
+	file, err := getTranslatedDockerFile(dockerFile, okCtx, projectHash)
 	if err != nil {
 		return "", errors.Wrap(err, "failed to create temporary build folder")
 	}
@@ -41,7 +42,7 @@ func GetDockerfile(dockerFile string, okCtx OktetoContextInterface) (string, err
 	return file, nil
 }
 
-func getTranslatedDockerFile(filename string, okCtx OktetoContextInterface) (string, error) {
+func getTranslatedDockerFile(filename string, okCtx OktetoContextInterface, projectHash string) (string, error) {
 	file, err := os.Open(filename)
 	if err != nil {
 		return "", err
@@ -70,6 +71,7 @@ func getTranslatedDockerFile(filename string, okCtx OktetoContextInterface) (str
 	for scanner.Scan() {
 		line := scanner.Text()
 		translatedLine := translateOktetoRegistryImage(line, okCtx)
+		translatedLine = translateCacheHandler(translatedLine, projectHash)
 		_, err = datawriter.WriteString(translatedLine + "\n")
 		if err != nil {
 			return "", fmt.Errorf("failed to write dockerfile: %w", err)
@@ -153,4 +155,31 @@ func copyFile(orig, dest string) error {
 		return err
 	}
 	return nil
+}
+
+func translateCacheHandler(input, projectHash string) string {
+	// Check if this RUN command has a cache mount
+	hasCacheMount, err := regexp.MatchString(`^RUN.*--mount=.*type=cache`, input)
+	if err != nil || !hasCacheMount {
+		return input
+	}
+
+	// If an id is already defined, leave unchanged
+	hasID, err := regexp.MatchString(`^RUN.*--mount=[^ ]*id=`, input)
+	if err == nil && hasID {
+		return input
+	}
+
+	target := ""
+	re := regexp.MustCompile(`--mount=[^ ]*target=([^, ]+)`)
+	if matches := re.FindStringSubmatch(input); len(matches) > 1 {
+		target = matches[1]
+	}
+
+	id := projectHash
+	if target != "" {
+		id = fmt.Sprintf("%s-%s", projectHash, target)
+	}
+	// Otherwise, insert id=<projectHash> into the mount
+	return strings.ReplaceAll(input, "--mount=", fmt.Sprintf("--mount=id=%s,", id))
 }
--- a/pkg/cmd/build/file.go
+++ b/pkg/cmd/build/file.go
@@ -14,14 +14,12 @@
 package build
 
 import (
-	"bufio"
 	"fmt"
 	"os"
 	"path/filepath"
 	"regexp"
 	"strings"
 
-	"github.com/okteto/okteto/pkg/config"
 	"github.com/okteto/okteto/pkg/constants"
 	oktetoLog "github.com/okteto/okteto/pkg/log"
 	"github.com/okteto/okteto/pkg/registry"
@@ -33,8 +31,8 @@ const (
 )
 
 // GetDockerfile returns the dockerfile with the cache and registry translations
-func GetDockerfile(dockerFile string, okCtx OktetoContextInterface, projectHash string) (string, error) {
-	file, err := getTranslatedDockerFile(dockerFile, okCtx, projectHash)
+func GetDockerfile(dockerFile string, okCtx OktetoContextInterface, repoURL, dockerfilePath, target string) (string, error) {
+	file, err := getTranslatedDockerFile(dockerFile, okCtx, repoURL, dockerFile, target)
 	if err != nil {
 		return "", errors.Wrap(err, "failed to create temporary build folder")
 	}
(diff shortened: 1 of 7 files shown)


pkg/cmd/build/translator.go
```diff
@@ -0,0 +1,258 @@
...
+	// Generate SHA256 hash
+	hasher := sha256.New()
+	hasher.Write([]byte(input))
+	hash := hasher.Sum(nil)
+
+	// Return first 12 characters of hex hash for readability
+	return hex.EncodeToString(hash)[:12]
+}
+
+func (cmt cacheMountTranslator) translate(line string) string {
```
jLopezbarb, line 202: This is the main change. The other things is moving code to allow having the id here and refactoring

pkg/cmd/build/file.go
```diff
@@ -154,3 +125,30 @@ func copyFile(orig, dest string) error {
...
+
+	target := ""
+	re := regexp.MustCompile(`--mount=[^ ]*target=([^, ]+)`)
+	if matches := re.FindStringSubmatch(input); len(matches) > 1 {
+		target = matches[1]
+	}
+
+	id := projectHash
+	if target != "" {
+		id = fmt.Sprintf("%s-%s", projectHash, target)
```
ifbyol: Shouldn't be everything hashed as we do in the test caches?

pkg/cmd/build/file.go
```diff
@@ -154,3 +125,30 @@ func copyFile(orig, dest string) error {
 	}
 	return nil
 }
+
+func translateCacheHandler(input, projectHash string) string {
```
ifbyol: I'm confused. Where is this being used? I see also that `translator.go` has also logic to do this translation. I assume the other one is the one that has to be used, and this one deleted?

pkg/cmd/build/translator.go
pchico83: Could you provide some values for repo, folder and target?
I guess repo is the repo name, like `app` in https://github.com/okteto/app
Folder is the folder of the dockerfile? I would use the relative path of the dockerfile respect to the repo root folder.
Why do we need `target`? what is it?
Fianlly, what's the value of `repo` if the dockerfile doesn't belong to a git repo?

pkg/cmd/build/translator.go
```diff
@@ -0,0 +1,258 @@
...
+	mountRegex         *regexp.Regexp
+	targetRegex        *regexp.Regexp
+}
+
+func newCacheMountTranslator(repo, dockerfilePath, target string) cacheMountTranslator {
+	cacheMountRegex := regexp.MustCompile(`^RUN.*--mount=.*type=cache`)
+	hasIDRegex := regexp.MustCompile(`^RUN.*--mount=[^ ]*id=`)
+	targetExtractRegex := regexp.MustCompile(`--mount=[^ ]*target=([^, ]+)`)
+	mountRegex := regexp.MustCompile(`--mount=([^[:space:]]+)`)
+	targetRegex := regexp.MustCompile(`target=([^,\s]+)`)
```
ifbyol: Every time we create a translator we compile the same regex again. Let's extract them to internal variables in the package. Also, I don't think that passing them as dependency of `cacheMountTranslator` provides any value here

pkg/cmd/build/translator.go
```diff
@@ -0,0 +1,258 @@
...
+	// Find all --mount= occurrences and process them individually
+	result := line
+
+	// Find all mount definitions
+	matches := cmt.mountRegex.FindAllStringSubmatch(result, -1)
+	for _, match := range matches {
+
+		if len(match) < 2 {
+			continue
+		}
```
ifbyol, line 222: Why this check? Could we add an inline comment, please?

pkg/cmd/build/translator.go
```diff
@@ -0,0 +1,258 @@
...
+
+		// Check if this is a cache mount
+		if !strings.Contains(mountParams, "type=cache") {
+			continue
+		}
+
+		// Check if it already has an id
+		if strings.Contains(mountParams, "id=") {
+			continue
+		}
```
ifbyol, line 240: Wasn't this already checked with a regex in the line 212?

pkg/cmd/build/translator.go
```diff
@@ -0,0 +1,258 @@
...
+		// Extract target from this specific mount
+		target := ""
+		if targetMatch := cmt.targetRegex.FindStringSubmatch(mountParams); len(targetMatch) > 1 {
+			target = targetMatch[1]
+		}
+
+		// Generate unique ID for this mount
+		id := cmt.projectHash
+		if target != "" {
+			id = fmt.Sprintf("%s-%s", cmt.projectHash, target)
```
ifbyol: Shouldn't the `target` being part of the hash as it happens with test caches? (maybe I'm wrong, but I'd swear that in the test case, this was part of the hash instead of being independent

pkg/cmd/build/translator.go
jLopezbarb: I removed the regex in order to fix a bug when several mount caches and at least one had id set

pkg/cmd/build/translator.go
jLopezbarb: we need to check that we had the fullmount and the params. This will allow us to change it later to add the id

pkg/cmd/build/translator.go
jLopezbarb: > Could you provide some values for repo, folder and target?

Repo is the repo URL: https://github.com/okteto/okteto
dockerfilePath is the path for the Dockerfile from the buildInfo: Dockerfile or frontend/Dockerfile or /usr/app/Dockerfile
buildTarget is the target for the build, it can be empty but it helps us to differentiate: dev/prod/""
cacheTarget is the path where the cache will be mounted: ./node_modules

pchico83: Great, thanks! A few more questions:
- why do we need `buildTarget`?
- What's `repo` when we are not in a git repo?
- is it repo the same if the repo is cloned via ssh and https?

I am thinking if we should use manifest name instead of repo to simplify the logic and make it more similar to the case of test caches. We would replace dockerfile path by test container name, but the rest would be the same

jLopezbarb: > why do we need buildTarget

I added it in case there is a multistage with different mount cache but pointing to the same folder

> What's repo when we are not in a git repo?

It's the manifestPath, the path to the okteto manifest, I opted to go for the manifest in case there is more than one.

> is it repo the same if the repo is cloned via ssh and https?

Yes, we are using the same translation that we do when using remote, so we will always use the same URL.

pchico83: Then it SGMT, thanks @jLopezbarb !

pkg/cmd/build/translator.go
```diff
@@ -0,0 +1,267 @@
...
+	"github.com/okteto/okteto/pkg/registry"
+)
+
+const (
+	tmpFilePrefix = "buildkit-"
+)
+
+var (
+	cacheMountRegex    = regexp.MustCompile(`^RUN.*--mount=.*type=cache`)
+	hasIDRegex         = regexp.MustCompile(`^RUN.*--mount=[^ ]*id=`)
```
ifbyol: I think this regex is not used anymore

pkg/cmd/build/translator.go
ifbyol: > I am thinking if we should use manifest name instead of repo to simplify the logic and make it more similar to the case of test caches.

Just a question. @pchico83 mentions this, but @jLopezbarb mentions that we only use the name when the repo is not defined.

Should we use always name instead of repo? Or should we include the manifest name as part of the hash apart of the repo (as we do in `okteto test`)? 

Just to be aligned

pkg/cmd/build/translator.go
```diff
@@ -0,0 +1,267 @@
...
+		// Check if this is a cache mount
+		// Example: type=cache,target=... => true
+		// Example: type=secret => false
+		if !strings.Contains(mountParams, "type=cache") {
+			continue
+		}
+
+		// Check if it already has an id
+		// Example: id=test,type=cache,target=... => true
+		// Example: type=cache,target=... => false
```
ifbyol, line 236: 👏 Thanks for these comments!

pkg/cmd/build/translator.go
```diff
@@ -0,0 +1,267 @@
...
+	return line
+}
+
+type cacheMountTranslator struct {
+	repo               string
+	dockerfilePath     string
+	buildTarget        string
+	hash               func(repositoryURL, dockerfilePath, buildTarget, cacheTarget string) string
+	cacheMountRegex    *regexp.Regexp
+	hasIDRegex         *regexp.Regexp
```
ifbyol: I think this property is not being used

pkg/cmd/build/translator_test.go
```diff
@@ -0,0 +1,784 @@
...
+			require.Len(t, dt.translators, tt.expectedTranslatorCount)
+			require.NotEmpty(t, dt.tmpFolder)
+		})
+	}
+}
+
+func TestDockerfileTranslator_Translate(t *testing.T) {
+	tests := []struct {
+		name            string
+		inputContent    string
```
ifbyol: I think `inputContent` is not being used, and it is redundant with the definition of `setupTranslator` function.

Or we remove this property, or we initialize `setupTranslator` in the test function (insted of here) using the value of `tt.inputContent`

pkg/cmd/build/translator_test.go
```diff
@@ -0,0 +1,784 @@
...
+		expectedOutput  string
+		setupTranslator func() *DockerfileTranslator
+	}{
+		{
```
ifbyol, line 170: I think it is important to have a case to have more than one translator to test that all of them are used. It doesn't have to be the real translator, it can be a mocked one.

pkg/cmd/build/translator_test.go
```diff
@@ -0,0 +1,784 @@
...
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			result := rt.translate(tt.input)
+			assert.Equal(t, tt.expected, result)
+		})
+	}
+}
+
+func TestNewCacheMountTranslator(t *testing.T) {
```
ifbyol: Nit: I don't think this test provides some value. It is not testing business logic. Same for the other new... tests. Anyways, as they are written, I'm ok leaving them

pkg/cmd/build/translator.go
```diff
@@ -0,0 +1,267 @@
...
+		// Example: type=cache,target=... => true
+		// Example: type=secret => false
+		if !strings.Contains(mountParams, "type=cache") {
+			continue
+		}
+
+		// Check if it already has an id
+		// Example: id=test,type=cache,target=... => true
+		// Example: type=cache,target=... => false
+		if strings.Contains(mountParams, "id=") {
```
ifbyol: This check is wrong. if the mount instruction uses `uid=` or `guid=`, this would be true and it would not add the id. Example of instruction: 

```
RUN --mount=type=cache,target=/home/okteto/.npm,uid=1001,gid=1001 \
    npm ci --cache /home/okteto/.npm
```

ifbyol:
> > > I am thinking if we should use manifest name instead of repo to simplify the logic and make it more similar to the case of test caches.
> > 
> > 
> > Just a question. @pchico83 mentions this, but @jLopezbarb mentions that we only use the name when the repo is not defined.
> > Should we use always name instead of repo? Or should we include the manifest name as part of the hash apart of the repo (as we do in `okteto test`)?
> > Just to be aligned
> 
> I wouldn't use the name because we will have a problem with the build that we do directly on Dockerfiles. With the path I think we covered both cases. WDYT?

@jLopezbarb Which problem you mean with build of standalone Dockerfiles?

jLopezbarb:
> > > > I am thinking if we should use manifest name instead of repo to simplify the logic and make it more similar to the case of test caches.
> > > 
> > > 
> > > Just a question. @pchico83 mentions this, but @jLopezbarb mentions that we only use the name when the repo is not defined.
> > > Should we use always name instead of repo? Or should we include the manifest name as part of the hash apart of the repo (as we do in `okteto test`)?
> > > Just to be aligned
> > 
> > 
> > I wouldn't use the name because we will have a problem with the build that we do directly on Dockerfiles. With the path I think we covered both cases. WDYT?
> 
> @jLopezbarb Which problem you mean with build of standalone Dockerfiles?

It's not a problem, it's that the manifest name will be empty so I preferred this method because it will be the same regarding of if it's through a manifest or a dockerfile

PR #4794, opened 2025-10-06
refactor: enhance namespace destruction logic in localDestroyAllCommand

jLopezbarb:
# Proposed changes

Fixes DEV-1046

- Introduced a new constant for the ticker duration during the destruction process.
- Improved the waitForNamespaceDestroyAllToComplete method to ensure comprehensive checks for resource destruction.
- Added a new checkAllResourcesDestroyed method to verify that all resources, including configmaps, are properly removed after the destruction operation.

## How to validate

1. Run `okteto destroy --all` several times and check that it always finishes

## CLI Quality Reminders 🔧

### DIFF
--- a/cmd/destroy/all.go
+++ b/cmd/destroy/all.go
@@ -182,7 +182,7 @@ func (lda *localDestroyAllCommand) checkAllResourcesDestroyed(ctx context.Contex
 	for _, cfg := range cfgList.Items {
 		for l := range cfg.GetLabels() {
 			if _, ok := resourcesLabels[l]; ok {
-				return fmt.Errorf("some resources where not destroyed")
+				return fmt.Errorf("some resources were not destroyed")
 			}
 		}
 	}


cmd/destroy/all.go
```diff
@@ -164,3 +160,31 @@ func (lda *localDestroyAllCommand) waitForNamespaceDestroyAllToComplete(ctx cont
...
+	cfgList, err := k8s.CoreV1().ConfigMaps(namespace).List(ctx, metav1.ListOptions{})
+	if err != nil {
+		return err
+	}
+
+	// no configmap for resources of the given namespace should exist
+	for _, cfg := range cfgList.Items {
+		for l := range cfg.GetLabels() {
+			if _, ok := resourcesLabels[l]; ok {
+				return fmt.Errorf("namespace destroy all failed: some resources where not destroyed")
```
ifbyol: I think we should change the logs then. Before, this was running when the destroy operation in theory finished. So, when we found something, that would mean that something wasn't deleted. Now, this can be executed while the job is still being executed, or at least, if an error happens, we keeps waiting in the main loop

cmd/destroy/all.go
```diff
@@ -129,32 +138,19 @@ func (lda *localDestroyAllCommand) waitForNamespaceDestroyAllToComplete(ctx cont
...
+						continue
 					}
+					hasBeenDestroyingAll = true
+				}
 
-					// exit the waiting loop when status is active again
-					return nil
+				if err := lda.checkAllResourcesDestroyed(ctx, namespace, c); err != nil {
+					oktetoLog.Infof("namespace destroy all failed: %v", err)
+					continue
```
ifbyol: Why do we keep within the loop in case of error?

jLopezbarb: The case that after 10s the action has not been created but there are still resources destroying

cmd/destroy/all.go
jLopezbarb: Updated the logs

ifbyol:
```suggestion
				return fmt.Errorf("namespace destroy all failed: some resources were not destroyed")
```

cmd/destroy/all.go
```diff
@@ -164,3 +162,31 @@ func (lda *localDestroyAllCommand) waitForNamespaceDestroyAllToComplete(ctx cont
...
+	cfgList, err := k8s.CoreV1().ConfigMaps(namespace).List(ctx, metav1.ListOptions{})
+	if err != nil {
+		return err
+	}
+
+	// no configmap for resources of the given namespace should exist
+	for _, cfg := range cfgList.Items {
+		for l := range cfg.GetLabels() {
+			if _, ok := resourcesLabels[l]; ok {
+				return fmt.Errorf("some resources where not destroyed")
```
ifbyol:
```suggestion
				return fmt.Errorf("some resources were not destroyed")
```

PR #4801, opened 2025-10-14
fix: improve error handling in log streaming during resource destruction

jLopezbarb:
# Proposed changes

Fixes DEV-1167

- Enhanced the error handling in the destroy method of localDestroyAllCommand to avoid logging warnings for context cancellation and deadline exceeded errors. This change ensures that only relevant connectivity issues are reported, improving the clarity of log outputs during the destruction process.

Follows the same approach that we already have for pipeline deploy and destroy

## How to validate

1. Run okteto destroy --all and check that the log is not appearing

## CLI Quality Reminders 🔧

### DIFF
--- a/cmd/namespace/delete.go
+++ b/cmd/namespace/delete.go
@@ -81,7 +81,7 @@ func (nc *Command) ExecuteDeleteNamespace(ctx context.Context, namespace string,
 	}
 
 	if err := nc.watchDelete(ctx, namespace, k8sLogger); err != nil {
-		return fmt.Errorf("watching namespace deletion error: %w", err)
+		return fmt.Errorf("watching namespace deletion stopped: %w", err)
 	}
 
 	oktetoLog.Success("Namespace '%s' deleted", namespace)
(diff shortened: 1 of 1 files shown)


ifbyol:
I got a panic when stopping a `destroy --all` operation 

```
panic: send on closed channel

goroutine 22 [running]:
github.com/okteto/okteto/cmd/destroy.(*localDestroyAllCommand).destroy.func1(0x0?)
	github.com/okteto/okteto/cmd/destroy/all.go:80 +0x7c
created by github.com/okteto/okteto/cmd/destroy.(*localDestroyAllCommand).destroy in goroutine 1
	github.com/okteto/okteto/cmd/destroy/all.go:78 +0x2a8
```

I deployed a repository with a destroy section, and then, I executed `okteto destroy --all` and in the middle of the operation, I got that panic. Could you take a look at it?

I cannot reproduce it always, so I would say there is some race condition

cmd/namespace/delete.go
```diff
@@ -138,7 +137,9 @@ func (nc *Command) watchDelete(ctx context.Context, namespace string, k8sLogger
 	select {
 	case <-stop:
 		ctxCancel()
+		logsCtxCancel()
 		oktetoLog.Infof("CTRL+C received, exit")
+		oktetoLog.Information("CTRL+C received, cancelling context and logs context but operation will continue in background")
```
ifbyol:
```suggestion
		oktetoLog.Information("CTRL+C received, cancelling wait and logs streaming but operation will continue in background")
```

ifbyol: Also, checking how it looks like



I would change the message error by something like: `Watching namespace deletion stopped: interrupt signa received`

ifbyol: Screenshot 2025-10-21 at 14 04 17

This is the message I refer on my previous comment ^. For some reason, I couldn't add it to it

PR #4803, opened 2025-10-23
Refactor build steps 

jLopezbarb:
# Proposed changes

- Refactor build in order to future improvements (parallel builds, parallel checks) and speed up things
- New steps: Calculate metadata, calculate hashes, calculate if is build, clone cached images, build images

## How to validate

1. Run okteto build with several flags
1. Run okteto deploy and check that the smart build is working there

### DIFF
--- a/cmd/build/v2/smartbuild/sequential.go
+++ b/cmd/build/v2/smartbuild/sequential.go
@@ -156,15 +156,18 @@ func (s *SequentialCheckStrategy) cloneGlobalImageToDev(cachedImage, manifestNam
 // (directly or indirectly) are also marked as not cached without checking cache
 func (s *SequentialCheckStrategy) addDependentsToNotCached(svc *buildTypes.BuildInfo, dependantMap map[string][]string, processed map[string]bool, notCachedSvcs []*buildTypes.BuildInfo, buildInfoByName map[string]*buildTypes.BuildInfo) []*buildTypes.BuildInfo {
 	for _, dependent := range dependantMap[svc.Name()] {
-		if !processed[dependent] {
-			processed[dependent] = true
-			bi, ok := buildInfoByName[dependent]
-			if ok {
-				notCachedSvcs = append(notCachedSvcs, bi)
-			}
-			// Recursively add dependents of this dependent
-			notCachedSvcs = s.addDependentsToNotCached(bi, dependantMap, processed, notCachedSvcs, buildInfoByName)
+		if processed[dependent] {
+			continue
 		}
+		processed[dependent] = true
+		bi, ok := buildInfoByName[dependent]
+		if !ok {
+			continue
+		}
+
+		notCachedSvcs = append(notCachedSvcs, bi)
+		// Recursively add dependents of this dependent
+		notCachedSvcs = s.addDependentsToNotCached(bi, dependantMap, processed, notCachedSvcs, buildInfoByName)
 	}
 
 	return notCachedSvcs


cmd/build/v2/checker/checker.go
```diff
@@ -0,0 +1,92 @@
...
+	"github.com/okteto/okteto/cmd/build/v2/environment"
+	"github.com/okteto/okteto/cmd/build/v2/metadata"
+	"github.com/okteto/okteto/pkg/build"
+	"github.com/okteto/okteto/pkg/env"
+	"github.com/okteto/okteto/pkg/log/io"
+	"github.com/okteto/okteto/pkg/registry"
+)
+
+const (
+	parallelCheckStrategyEnvVar = "OKTETO_BUILD_CHECK_STRATEGY_PARALLEL"
```
ifbyol: Which will be the default value for this env var?

cmd/build/v2/checker/checker.go
```diff
@@ -0,0 +1,92 @@
...
+	return i.checkStrategy.GetImageDigestReferenceForServiceDeploy(manifestName, service, buildInfo)
+}
+
+// CloneGlobalImagesToDev clones multiple global images to the development environment.
+// This is useful for ensuring development environments have access to the latest global images.
+func (i *ImageCacheChecker) CloneGlobalImagesToDev(manifestName string, buildManifest build.ManifestBuild, images []string) error {
+	if err := i.checkStrategy.CloneGlobalImagesToDev(manifestName, buildManifest, images); err != nil {
+		return err
+	}
+	return nil
```
ifbyol:
```suggestion
	return i.checkStrategy.CloneGlobalImagesToDev(manifestName, buildManifest, images)
```

cmd/build/v2/checker/checker.go
```diff
@@ -0,0 +1,92 @@
...
+// SmartBuildController defines the interface for smart build operations.
+type SmartBuildController interface {
+	GetBuildHash(buildInfo *build.Info, service string) string
+	CloneGlobalImageToDev(globalImage, svcImage string) (string, error)
+}
+
+// CheckStrategy defines the interface for different cache checking strategies.
+type CheckStrategy interface {
+	CheckServicesCache(ctx context.Context, manifestName string, buildManifest build.ManifestBuild, svcsToBuild []string) (cachedServices []string, notCachedServices []string, err error)
+	CloneGlobalImagesToDev(manifestName string, buildManifest build.ManifestBuild, images []string) error
```
ifbyol: Why don't we encapsulate this under the same operation to outside of this? I mean, this could return only the list of non-cached images, and make sure the rest of them are copied, optimizing the process, instead of having 2 different "parallel" implementations (one for the check, and another one for the clone)

cmd/build/v2/checker/checker.go
```diff
@@ -0,0 +1,92 @@
+// Copyright 2023 The Okteto Authors
```
ifbyol:
```suggestion
// Copyright 2025 The Okteto Authors
```

cmd/build/v2/build.go
```diff
@@ -143,25 +175,49 @@ func NewBuilderFromScratch(ioCtrl *io.Controller, onBuildFinish []OnBuildFinish)
...
+		metadataCollector:    metadataCollector,
+		onBuildFinish:        onBuildFinish,
+		serviceEnvVarsSetter: serviceEnvVarsSetter,
+		tagger:               tagger,
 	}
 }
 
+// GetBuildEnvVars returns the build env vars
+func (ob *OktetoBuilder) GetBuildEnvVars() map[string]string {
+	return ob.serviceEnvVarsSetter.GetBuildEnvVars()
```
ifbyol: Nit: It sounds odd to have something that is a setter, but then, you call a get. If it handles both get and set, you can call it hendler or manager

cmd/build/v2/build.go
```diff
@@ -218,125 +270,80 @@ func (ob *OktetoBuilder) Build(ctx context.Context, options *types.BuildOptions)
...
+		return err
+	}
+	tree, err = tree.Subtree(toBuildSvcs...)
+	if err != nil {
+		return err
+	}
+	toBuildSvcs = tree.Ordered()
 
-			repoHashDurationStart := time.Now()
+	ob.ioCtrl.Logger().Infof("Images to build: [%s]", strings.Join(toBuildSvcs, ", "))
```
ifbyol: Nit: This might be a bit misleading, as we don't know yet which images are going to be built. These are the potential images to check

cmd/build/v2/build.go
```diff
@@ -218,125 +270,80 @@ func (ob *OktetoBuilder) Build(ctx context.Context, options *types.BuildOptions)
...
+		if err != nil {
+			return fmt.Errorf("error checking images: %w", err)
+		}
+		ob.ioCtrl.Logger().Infof("Images cached: [%s]", strings.Join(cachedServices, ", "))
+		ob.ioCtrl.Logger().Infof("Images not cached: [%s]", strings.Join(notCachedServices, ", "))
 
-			if !ob.oktetoContext.IsOktetoCluster() && buildSvcInfo.Image == "" {
-				return fmt.Errorf("'build.%s.image' is required if your context doesn't have Okteto installed", svcToBuild)
-			}
+		sp = ob.ioCtrl.Out().Spinner("Cloning global images to dev...")
```
ifbyol: I don't think this is something meaningful to the final user. They probably don't know what do we mean with global images

cmd/build/v2/build.go
```diff
@@ -218,125 +270,80 @@ func (ob *OktetoBuilder) Build(ctx context.Context, options *types.BuildOptions)
...
+		sp.Start()
+		if err := ob.imageChecker.CloneGlobalImagesToDev(options.Manifest.Name, options.Manifest.Build, cachedServices); err != nil {
```
ifbyol: I don't see the need of extracting this outside of the image checker for mainly 2 reasons:
* To this main function, we only care the images to be built, which are the ones we have to explicitly build. For the rest, we just need the final image sha
* I wouldn't expect something called imageChecker to have a function to Clone images.

For me, it is not an image checker. It is an image cache handler, which tells you which images are or not already built, and make sure that the cached images are where they should be. The fact of having to copy an image or not it is not responsibility of the builder, builder's responsibility is to build non-cached images.

This also forces us to have a stage that I truly believe it is not meaningful for the final user. The fact that an image has to be cloned within the registry is not relevant for 90% of the users probably.

cmd/build/v2/build.go
```diff
@@ -218,125 +270,80 @@ func (ob *OktetoBuilder) Build(ctx context.Context, options *types.BuildOptions)
...
-					continue
-				}
-			}
+	sp := ob.ioCtrl.Out().Spinner("Collecting metadata for the images")
+	sp.Start()
+	if err := ob.metadataCollector.CollectMetadata(ctx, options.Manifest.Name, options.Manifest.Build, toBuildSvcs); err != nil {
+		return err
+	}
+	sp.Stop()
+	metaMap := ob.metadataCollector.GetMetadataMap()
```
ifbyol: Why do we need to calculate this in advance? Which is the benefit of it? Can't we calculate it within the checker?

It forces us to have a stage that most of the times is really fast, and it complicates things as we generate several goroutines

cmd/build/v2/build.go
```diff
@@ -218,125 +270,80 @@ func (ob *OktetoBuilder) Build(ctx context.Context, options *types.BuildOptions)
...
+
+	notCachedServices := toBuildSvcs
+	var cachedServices []string
+	if !options.NoCache && ob.smartBuildCtrl.IsEnabled() {
+		sp := ob.ioCtrl.Out().Spinner("Checking if the images are already built from cache...")
+		sp.Start()
+		defer sp.Stop()
+		cachedServices, notCachedServices, err = ob.imageChecker.CheckImages(ctx, options.Manifest.Name, options.Manifest.Build, toBuildSvcs)
```
ifbyol: Shouldn't this happen through the smart builds controller? I mean, at the end, smart builds is the cache enabler, but it sounds odd that the logic remains in other place. I think we should be coherent about it

cmd/build/v2/build.go
```diff
@@ -455,7 +462,7 @@ func validateServices(buildSection build.ManifestBuild, svcsToBuild []string) er
 	return nil
 }
 
-func getImageChecker(cfg oktetoBuilderConfigInterface, registry registryImageCheckerInterface, sbc smartBuildController, logger loggerInfo) imageCheckerInterface {
-	tagger := newImageTagger(cfg, sbc)
-	return newImageChecker(cfg, registry, tagger, logger)
-}
+// func getImageChecker(cfg oktetoBuilderConfigInterface, registry registryImageCheckerInterface, sbc smartBuildController, logger loggerInfo) imageChecker {
```
ifbyol: Can this be removed?

cmd/build/v2/graph.go
```diff
@@ -0,0 +1,56 @@
+// Copyright 2023 The Okteto Authors
```
ifbyol:
```suggestion
// Copyright 2025 The Okteto Authors
```

cmd/build/v2/checker/checker_test.go
```diff
@@ -0,0 +1,378 @@
+// Copyright 2023 The Okteto Authors
```
ifbyol:
```suggestion
// Copyright 2025 The Okteto Authors
```

cmd/build/v2/checker/image_cache.go
```diff
@@ -0,0 +1,109 @@
+// Copyright 2023 The Okteto Authors
```
ifbyol:
```suggestion
// Copyright 2025 The Okteto Authors
```

cmd/build/v2/checker/image_cache_test.go
```diff
@@ -0,0 +1,519 @@
+// Copyright 2023 The Okteto Authors
```
ifbyol:
```suggestion
// Copyright 2025 The Okteto Authors
```

cmd/build/v2/checker/sequential.go
```diff
@@ -0,0 +1,185 @@
+// Copyright 2023 The Okteto Authors
```
ifbyol:
```suggestion
// Copyright 2025 The Okteto Authors
```

cmd/build/v2/checker/sequential_test.go
```diff
@@ -0,0 +1,521 @@
+// Copyright 2023 The Okteto Authors
```
ifbyol:
```suggestion
// Copyright 2025 The Okteto Authors
```

cmd/build/v2/environment/environment.go
```diff
@@ -0,0 +1,108 @@
+// Copyright 2023 The Okteto Authors
```
ifbyol:
```suggestion
// Copyright 2025 The Okteto Authors
```

cmd/build/v2/metadata/metadata.go
```diff
@@ -0,0 +1,162 @@
+// Copyright 2023 The Okteto Authors
```
ifbyol:
```suggestion
// Copyright 2025 The Okteto Authors
```

cmd/build/v2/metadata/metadata_test.go
```diff
@@ -0,0 +1,556 @@
+// Copyright 2023 The Okteto Authors
```
ifbyol:
```suggestion
// Copyright 2025 The Okteto Authors
```

cmd/build/v2/checker/checker.go
```diff
@@ -0,0 +1,92 @@
...
+type CheckStrategy interface {
+	CheckServicesCache(ctx context.Context, manifestName string, buildManifest build.ManifestBuild, svcsToBuild []string) (cachedServices []string, notCachedServices []string, err error)
+	CloneGlobalImagesToDev(manifestName string, buildManifest build.ManifestBuild, images []string) error
+	GetImageDigestReferenceForServiceDeploy(manifestName, service string, buildInfo *build.Info) (string, error)
+}
+
+// ImageCacheChecker is responsible for checking image cache status and managing image operations.
+type ImageCacheChecker struct {
+	checkStrategy CheckStrategy
+}
```
ifbyol: You don't need this wrapper. You just need the interface, the 2 implementations, and a factory to create one or another based on the env variable, but this wrapper doesn't provide value. It just makes things more complex adding one more indirection level.

I would say that this file could be completely removed

cmd/build/v2/checker/image_cache.go
```diff
@@ -0,0 +1,109 @@
...
+	registryURL string
+	imageCtrl   registry.ImageCtrl
+	registry    DigestResolver
+	logger      Logger
+
+	mu    sync.RWMutex
+	cache map[string]string
+}
+
+func NewImageChecker(tagger ImageTagger, namespace string, registryURL string, imageCtrl registry.ImageCtrl, registry DigestResolver, logger Logger) *RegistryCacheProbe {
```
ifbyol: This is called `NewImageChecker` but it is returning `RegistryCacheProbe`, I'd update the name of the function to match what it really returns, specially having another struct that is called image checker

cmd/build/v2/checker/sequential.go
```diff
@@ -0,0 +1,185 @@
...
+				processed[dependant] = true
+				notCachedSvcs = append(notCachedSvcs, dependant)
+				// Set metadata for dependant services
+				dependantMeta := s.metadataCollector.GetMetadata(dependant)
+				dependantMeta.CacheHit = false
+				dependantMeta.CacheHitDuration = 0 // No cache check performed
+
+				// Recursively add dependants of this dependant
+				notCachedSvcs = s.addDependantsToNotCached(dependant, dependantMap, processed, notCachedSvcs)
+			}
```
ifbyol: Nit: it is easier to read and follow if we do an early break and we leave the main logic without one less level of indentation, specially taking into account that it would be the 3rd level.
```suggestion
			if processed[dependant] {
			    continue
			}
			processed[dependant] = true
			notCachedSvcs = append(notCachedSvcs, dependant)
			// Set metadata for dependant services
			dependantMeta := s.metadataCollector.GetMetadata(dependant)
			dependantMeta.CacheHit = false
			dependantMeta.CacheHitDuration = 0 // No cache check performed

			// Recursively add dependants of this dependant
			notCachedSvcs = s.addDependantsToNotCached(dependant, dependantMap, processed, notCachedSvcs)
			
```

cmd/build/v2/checker/sequential.go
```diff
@@ -0,0 +1,185 @@
...
+		}
+	}
+	return cachedSvcs, notCachedSvcs, nil
+}
+
+// addDependantsToNotCached recursively adds all dependant services to notCached
+// This ensures that when a service is not cached, all services that depend on it
+// (directly or indirectly) are also marked as not cached without checking cache
+func (s *SequentialCheckStrategy) addDependantsToNotCached(svc string, dependantMap map[string][]string, processed map[string]bool, notCachedSvcs []string) []string {
+	if dependants, exists := dependantMap[svc]; exists {
```
ifbyol: Nit: same about readability. We can do an early return if it doesn't exit, making the function more readable
```suggestion
     dependants, exists := dependantMap[svc]
	if !exists  {
	    return notCachedSvcs
	}
	
	for _, dependant := range dependants {
	    ....
	}
```

cmd/build/v2/build.go
```diff
@@ -218,125 +270,80 @@ func (ob *OktetoBuilder) Build(ctx context.Context, options *types.BuildOptions)
...
+	notCachedServices := toBuildSvcs
+	var cachedServices []string
+	if !options.NoCache && ob.smartBuildCtrl.IsEnabled() {
+		sp := ob.ioCtrl.Out().Spinner("Checking if the images are already built from cache...")
+		sp.Start()
+		defer sp.Stop()
+		cachedServices, notCachedServices, err = ob.imageChecker.CheckImages(ctx, options.Manifest.Name, options.Manifest.Build, toBuildSvcs)
+		if err != nil {
+			return fmt.Errorf("error checking images: %w", err)
+		}
```
ifbyol, line 262: I would say this is a braking change. Before, if some error happened while checking the cache, we were just rebuilding that image. Check that the former function `checkIfBuildHashIsBuilt` does not even return an error

cmd/build/v2/graph.go
```diff
@@ -0,0 +1,56 @@
...
+}
+
+func (b *buildGraph) GetGraph() (*dag.Tree, error) {
+	return dag.From(b.nodes...)
+}
+
+func newBuildGraph(buildManifest build.ManifestBuild, services []string) *buildGraph {
+	graph := &buildGraph{}
+	for _, svc := range services {
+		graph.AddNode(&buildNode{id: svc, dependsOn: buildManifest[svc].DependsOn})
```
ifbyol, line 53: What would happen with cycles?

cmd/build/v2/build.go
```diff
@@ -218,125 +270,80 @@ func (ob *OktetoBuilder) Build(ctx context.Context, options *types.BuildOptions)
...
+	}
+	sp.Stop()
+	metaMap := ob.metadataCollector.GetMetadataMap()
+
+	notCachedServices := toBuildSvcs
+	var cachedServices []string
+	if !options.NoCache && ob.smartBuildCtrl.IsEnabled() {
+		sp := ob.ioCtrl.Out().Spinner("Checking if the images are already built from cache...")
+		sp.Start()
+		defer sp.Stop()
```
ifbyol, line 258: Shouldn't this have to be on line 304 instead of using a defer?

cmd/build/v2/checker/checker_test.go
```diff
@@ -0,0 +1,378 @@
...
+			digest, err := checker.GetImageDigestReferenceForServiceDeploy(tt.manifestName, tt.service, tt.buildInfo)
+
+			// Verify results
+			if tt.expectedError {
+				assert.Error(t, err)
+				assert.Equal(t, tt.mockError, err)
+			} else {
+				assert.NoError(t, err)
+				assert.Equal(t, tt.expectedDigest, digest)
+			}
```
ifbyol: We agreed on avoiding this kind of checks in test in favor of having more test functions (apply to other tests)

cmd/build/v2/checker/image_cache.go
```diff
@@ -0,0 +1,109 @@
...
+		registryURL: registryURL,
+		imageCtrl:   imageCtrl,
+		registry:    registry,
+		mu:          sync.RWMutex{},
+		cache:       make(map[string]string),
+		logger:      logger,
+	}
+}
+
+func (c *RegistryCacheProbe) IsCached(ctx context.Context, manifestName, image, buildHash, svcToBuild string) (bool, string, error) {
```
ifbyol: `ctx` parameter is not being used within the function, it can be removed

cmd/build/v2/graph.go
jLopezbarb: cycles are not allowed in this part. We have already shown an error when we reach this code

cmd/build/v2/smartbuild/image_cache.go
```diff
@@ -0,0 +1,108 @@
...
+		}
+	} else {
+		// [name]:[tag] being the tag the buildHash
+		referencesToCheck = c.tagger.GetImageReferencesForTag(manifestName, svcToBuild, buildHash)
+	}
+
+	for _, ref := range referencesToCheck {
+		c.mu.RLock()
+		_, ok := c.cache[ref]
+		c.mu.RUnlock()
```
ifbyol: Why do we need this cache? Is there any scenario where we can have to build the same image twice to use this local cache?

cmd/build/v2/smartbuild/image_cache_test.go
```diff
@@ -0,0 +1,516 @@
...
+	mockConfig.On("IsOktetoCluster").Return(false)
+	mockConfig.On("GetGlobalNamespace").Return("test-global-namespace")
+	mockConfig.On("GetNamespace").Return("test-namespace")
+	mockConfig.On("GetRegistryURL").Return("test-registry.com")
+
+	return registry.NewImageCtrl(mockConfig)
+}
+
+func TestNewRegistryCacheProbe(t *testing.T) {
+	t.Run("creates registry cache probe with all dependencies", func(t *testing.T) {
```
ifbyol: Nit: For a single test, why we use this instead of directly the test?

cmd/build/v2/smartbuild/image_cache_test.go
```diff
@@ -0,0 +1,516 @@
...
+	mockConfig := &MockImageConfig{}
+	mockConfig.On("IsOktetoCluster").Return(false)
+	mockConfig.On("GetGlobalNamespace").Return("test-global-namespace")
+	mockConfig.On("GetNamespace").Return("test-namespace")
+	mockConfig.On("GetRegistryURL").Return("test-registry.com")
+
+	return registry.NewImageCtrl(mockConfig)
+}
+
+func TestNewRegistryCacheProbe(t *testing.T) {
```
ifbyol: Nit: NewRegistryCacheProbe doesn't have any specific logic more than initializing a struct, I don't think it worths to have a test for it

cmd/build/v2/smartbuild/image_cache_test.go
```diff
@@ -0,0 +1,516 @@
...
+		t.Run(tt.name, func(t *testing.T) {
+			// Create mocks
+			mockTagger := &MockImageTaggerForCache{}
+			mockDigestResolver := &MockDigestResolverForCache{}
+			mockLogger := &MockLoggerForCache{}
+			imageCtrl := createTestImageCtrl()
+
+			// Set up mock expectations
+			if tt.buildHash != "" {
+				if tt.image != "" {
```
ifbyol: We agreed to avoid tests with conditionals inside as they make the tests harder to follow. We have 3 indentation levels to set expectations based on then, and also we have some assertions with conditionals too.

Please, review all the tests to change it

cmd/build/v2/smartbuild/image_cache_test.go
```diff
@@ -0,0 +1,516 @@
...
+				mockDigestResolver,
+				mockLogger,
+			)
+
+			// Execute
+			digest, err := probe.LookupReferenceWithDigest(tt.reference)
+
+			// Verify results
+			assert.Equal(t, tt.expectedDigest, digest)
+			if tt.expectedError {
```
ifbyol: Same about conditionals in the tests

cmd/build/v2/smartbuild/image_cache_test.go
```diff
@@ -0,0 +1,516 @@
...
+				cached, _, err := probe.IsCached("test-manifest", "test-image", "hash123", "test-service")
+				assert.True(t, cached)
+				assert.NoError(t, err)
+			}()
+		}
+
+		// Wait for all goroutines to complete
+		for i := 0; i < 10; i++ {
+			<-done
+		}
```
ifbyol: This test is a bit scary. I would use a waitGroup instead of a channel as it is easier to follow an understand, and I think it would accomplish the same

cmd/build/v2/smartbuild/sequential.go
```diff
@@ -0,0 +1,198 @@
...
+	}
+
+	for _, svc := range svcsToBuild {
+		// Skip if already processed
+		name := svc.Name()
+		if processed[name] {
+			continue
+		}
+
+		s.ioCtrl.SetStage(fmt.Sprintf("Building service %s", svc))
```
ifbyol: As discussed offline, I would create a new stage to represent the smart build check. As it is now, it would be very error prone because if we print something when the image is not cached, it could happen that we send messages to stages created previously and the experience is not good in the UI.

I would create a stage like `Checking if images were already built`, where we print the message for the images cached (we could even consider to print a messages with images that will be built).

And then, we have one stage `Building service ...` per image that needs to be built

cmd/build/v2/smartbuild/sequential_test.go
```diff
@@ -0,0 +1,479 @@
...
+
+			reference, err := strategy.GetImageDigestReferenceForServiceDeploy(tt.manifestName, tt.service, tt.buildInfo)
+
+			assert.Equal(t, tt.expectedReference, reference)
+			if tt.expectedError != nil {
+				assert.Error(t, err)
+				assert.Equal(t, tt.expectedError.Error(), err.Error())
+			} else {
+				assert.NoError(t, err)
+			}
```
ifbyol: Avoid conditionals within tests please

cmd/build/v2/smartbuild/sequential_test.go
```diff
@@ -0,0 +1,479 @@
...
+
+			tagger.AssertExpectations(t)
+			cacheProbe.AssertExpectations(t)
+		})
+	}
+}
+
+func TestNewSequentialCheckStrategy(t *testing.T) {
```
ifbyol: Nit: unit tests for New functions without special logic don't provide too much value

cmd/build/v2/build.go
```diff
@@ -201,179 +222,119 @@ func (ob *OktetoBuilder) Build(ctx context.Context, options *types.BuildOptions)
...
+			return fmt.Errorf("error checking images: %w", err)
+		}
+		sp.Stop()
+		ob.ioCtrl.Logger().Infof("Images cached: %v", cachedSvcs)
+		ob.ioCtrl.Logger().Infof("Images not cached: %v", notCachedSvcs)
+	}
 
-			repoHashDurationStart := time.Now()
+	for _, svcToBuild := range notCachedSvcs {
+		if options.EnableStages {
```
ifbyol, line 269: Should we also check this when we are creating the stage within `image_cache`?

ifbyol (approved):

PR #4805, opened 2025-10-23
improvement: Only add the args from dependencies in the build

jLopezbarb:

# Proposed changes

Fixes DEV-1172

- Checks if the previous environment variables are in the list of the dependents services. This reduces the unnecessary data we send to buildkit
- Introduced new tests for the AddArgs method in the Info struct to validate the handling of build arguments based on the depends_on configuration.
- Added a normalizeServiceName function to convert service names from environment variable format to the depends_on format.
- Implemented isServiceInDependencies method to check if a service is included in the depends_on list, enhancing the logic for managing build arguments.

## How to validate

1. Run okteto build with a service with depends on and use the variables in the dockerfile and check that it works

## CLI Quality Reminders 🔧

### DIFF
--- a/pkg/build/info.go
+++ b/pkg/build/info.go
@@ -17,12 +17,16 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strings"
 
 	"github.com/okteto/okteto/pkg/cache"
 	"github.com/okteto/okteto/pkg/env"
 	oktetoLog "github.com/okteto/okteto/pkg/log"
 )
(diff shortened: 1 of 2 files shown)


ifbyol (commented):
What was the reason of including the previous build environment variables in the next builds? Only the cases for `depends_on`? Or was there any other reason? Asking because other option to fix this but keeping the previous env vars is to try to generate a deterministic order when building images, so, given the same manifest, 2 consecutive builds would build images in the same order

PR #4829, opened 2025-11-21
feat: add feature flag for future implementation

jLopezbarb:
# Proposed changes

Fixes DEV-1226

This PR refactors the buildkit client initialization logic to prepare for introducing a queue-based build connector. The current implementation directly connects to buildkit, but we want to support an alternative queue-based approach controlled by a feature flag.

## Changes
**Package Restructuring**

We needed to move the code to a package that only manages the connection with buildkit, either if it is via direct connection or using a port forward to the pod. 

- Moved `pkg/build/buildkit/client.go` → `pkg/build/buildkit/connector/client.go`
- Moved `pkg/build/buildkit/wait.go` → `pkg/build/buildkit/connector/wait.go`
- Moved corresponding test files to maintain test coverage
- Created `pkg/build/buildkit/connector/direct.go` - new DirectConnector type that encapsulates direct buildkit connections

**Architectural Changes**

- Introduced `buildkitConnector` interface in `pkg/cmd/build/build.go` to abstract connection logic and use it depending on the connection done with buildkit
- Added `OKTETO_BUILD_QUEUE_ENABLED` environment variable (currently unused, preparation for queue implementation)

**Implementation Details**
- `DirectConnector` wraps `ClientFactory` and `Waiter`. The connector is the one in charge of doing the wait and providing the buildkit client so that's why I wrapped there
- `WaitUntilIsReady(ctx)` - waits for buildkit to be available (Same implementation as we had, only moved the code)
- `GetClientFactory()` - returns the client factory for build operations. We still need to provide the client in other parts of the code.
- `GetWaiter()` - returns the waiter instance. We are right now waiting in two different places, one before creating the client and another one before building, because it could happen that the buildkit was not available. That's why we need to return this
- Updated `buildWithOkteto()` to use connector methods instead of direct instantiation. This is the only part of the code we should do changes for this feature.

**Feature Flag**
- Added `OKTETO_BUILD_QUEUE_ENABLED` environment variable:
  - Default: `false` (uses direct connector)
  When `true`: Currently uses direct connector (queue implementation coming in future PR)
  Purpose: First test it on our end before shipping it to the rest of the users

## How to validate

1. Run okteto build with the `OKTETO_BUILD_QUEUE_ENABLED` set to "", "false", "true"

## CLI Quality Reminders 🔧

### DIFF
--- a/pkg/build/buildkit/connector/ingress.go
+++ b/pkg/build/buildkit/connector/ingress.go
@@ -26,13 +26,13 @@ type OktetoContextIface interface {
 	GetCurrentToken() string
 }
 
-type DirectConnector struct {
+type IngressConnector struct {
 	buildkitClientFactory *ClientFactory
 	waiter                *Waiter
 }
 
-// NewDirectConnector creates a new direct connector. It connects to the buildkit server directly.
-func NewDirectConnector(okCtx OktetoContextIface, ioCtrl *io.Controller) *DirectConnector {
+// NewIngressConnector creates a new ingress connector. It connects to the buildkit server via ingress.
+func NewIngressConnector(okCtx OktetoContextIface, ioCtrl *io.Controller) *IngressConnector {
 	buildkitClientFactory := NewBuildkitClientFactory(
 		okCtx.GetCurrentCertStr(),
 		okCtx.GetCurrentBuilder(),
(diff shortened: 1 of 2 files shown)


pkg/build/buildkit/connector/ingress.go
```diff
@@ -0,0 +1,63 @@
+// Copyright 2025 The Okteto Authors
```
ifbyol, line 1: As we talked offline, I don't know if this would be the best name. For me it is a bit misleading because I understood direct as the new way to connect, as we connect directly to the desired BuildKit pod, but it is the other way around, this is the current behavior. 

I don't have really good suggestions but brainstorming in case some of them like: External/internal (external would be the current one as we use external endpoint to connect), public/private.

Anyways, not a big deal, if the other approach has a meaningful name for it, "direct" could be good enough

pchico83: `ingress` vs `forward`?

PR #4840, opened 2025-12-03
  Unify CLI image configuration under a single OKTETO_CLI_IMAGE environment variable

pchico83:
# Summary

  This PR consolidates the fragmented CLI image configuration by unifying OKTETO_BIN and OKTETO_REMOTE_CLI_IMAGE into a single OKTETO_CLI_IMAGE environment variable. It also removes the redundant GetOktetoImage() function, ensuring consistent image tag resolution across all operations.

# Motivation

  Previously, we had multiple environment variables and functions to get the CLI image:
  - OKTETO_BIN - for okteto up start script
  - OKTETO_REMOTE_CLI_IMAGE - for remote deployments
  - GetBinImage() - returned okteto bin image
  - GetRemoteImage(versionString) - returned remote CLI image with version parameter
  - GetOktetoImage() - returned image for hybrid dev environments with "stable" fallback

  This fragmentation led to:
  - Inconsistent fallback behavior - different functions used different fallback tags ("master" vs "stable")
  - Redundant code - multiple functions doing essentially the same thing
  - Maintenance complexity - changes to image resolution logic needed updates in multiple places
  - Risk of version mismatches - the version parameter in GetRemoteImage() was always called with the same global config.VersionString anyway

# Changes

##  Unified Environment Variable

  - New: OKTETO_CLI_IMAGE - single source of truth for CLI image configuration
  - Backward compatible: Still reads OKTETO_BIN and OKTETO_REMOTE_CLI_IMAGE as fallbacks
  - Priority order:
    a. OKTETO_CLI_IMAGE
    b. OKTETO_BIN (deprecated)
    c. OKTETO_REMOTE_CLI_IMAGE (deprecated)
    d. CLI version from config.VersionString
    e. Fallback to "master" tag

##  Simplified Function Hierarchy

  - Removed: GetRemoteImage(versionString) - unnecessary parameter, always received config.VersionString
  - Removed: GetOktetoImage() - duplicated logic with inconsistent "stable" fallback
  - Unified: GetCliImage() - single function for all CLI image needs

##  Consistent Fallback Behavior

  All operations now consistently use "master" tag when the version string is invalid, ensuring we're always using a valid and recent image.

## Testing 

I enabled the `run-e2e` label to validate all e2e tests are passing after this change

### DIFF
--- a/pkg/config/image.go
+++ b/pkg/config/image.go
@@ -57,6 +57,7 @@ type ImageConfig struct {
 // Logger is the interface used to log messages
 type Logger interface {
 	Infof(format string, args ...interface{})
+	Warning(format string, args ...interface{})
 }
 
 // NewImageConfig creates a new ImageConfig instance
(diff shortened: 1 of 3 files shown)


pkg/config/image.go
```diff
@@ -67,47 +73,35 @@ func NewImageConfig(ioCtrl Logger) *ImageConfig {
...
+// This is used for all okteto operations including bin image for okteto up start script
+func (c *ImageConfig) GetCliImage() string {
+	// Check new unified env var first
+	cliImage := c.getEnv(oktetoCLIImageEnvVar)
+	if cliImage != "" {
+		c.ioCtrl.Infof("using okteto cli image (from OKTETO_CLI_IMAGE): %s", cliImage)
+		return cliImage
+	}
+
+	// Fall back to legacy OKTETO_BIN for backward compatibility
```
jLopezbarb, line 87: I'd add a warning indicating that this old env var are replaced with oktetoCLIImageEnvVar instead of just showing an info log message

pchico83: Screenshot 2025-12-04 at 05 57 03

Great point! I have updated the code, could you review the last commit?

jLopezbarb: Given that this part of the code is using the previous logger I'm accepting it but we should create a dedicated issue in order to remove all the mentions to the old logger

PR #4845, opened 2025-12-09
feat: add buildkit queue wait

jLopezbarb:
# Proposed changes

Fixes DEV-1223

This PR enhances the BuildKit port forwarding mechanism to handle scenarios where BuildKit pods are not immediately available. Instead of failing on the first attempt, the system now polls with exponential backoff and provides user feedback about queue position.

- Implemented polling loop with exponential backoff (1s → 2s → 4s → 8s → 10s max)
- Added 10-minute maximum timeout with actionable error message
- Display queue position and reason via spinner updates
- Added mutex protection for isActive field to prevent race conditions
- Changed oktetoClient to use types.OktetoInterface for better testability

Before
```mermaid
sequenceDiagram
    participant Client
    participant PortForwarder
    participant API as Okteto API
    participant K8s as Kubernetes

    Client->>PortForwarder: WaitUntilIsReady()
    PortForwarder->>API: GetLeastLoadedBuildKitPod()
    
    alt No pod available
        API-->>PortForwarder: Empty response
        PortForwarder-->>Client: ❌ Fails immediately
    else Pod available
        API-->>PortForwarder: Pod name
        PortForwarder->>K8s: Start port forward
        K8s-->>PortForwarder: Ready
        PortForwarder-->>Client: ✅ Success
    end
```

After

```mermaid
sequenceDiagram
    participant Client
    participant PortForwarder
    participant API as Okteto API
    participant K8s as Kubernetes

    Client->>PortForwarder: WaitUntilIsReady()
    PortForwarder->>PortForwarder: Start spinner
    
    loop Polling with backoff (1s → 10s max)
        PortForwarder->>API: GetLeastLoadedBuildKitPod()
        
        alt Pod available
            API-->>PortForwarder: Pod name
            PortForwarder->>K8s: Start port forward
            K8s-->>PortForwarder: Ready
            PortForwarder-->>Client: ✅ Success
        else In queue
            API-->>PortForwarder: Queue position
            PortForwarder->>PortForwarder: Update spinner with position
            Note over PortForwarder: Wait with backoff
        else Timeout (10 min)
            PortForwarder-->>Client: ❌ Timeout error
        end
    end
```

## How to validate

1. Set buildkit replicas to 0
1. Run okteto build
1. Set buildkit replicas to 1

## CLI Quality Reminders 🔧


ifbyol:
For now, I was just testing the main functionality (I didn't check the code yet), but I have found a couple of things to improve a bit.

First thing, in the message we are showing when the CLI is waiting to connect, we are just showing the reason as it comes from the backend, and I'm not sure that is very user friendly. We could map the status to a specific messages to show, as even the whole message might be different base of the status. I could see how the old pods are busy message might be different from the buildkit is scaling or any other. See attached screenshot below

Screenshot 2025-12-11 at 15 40 24

Second thing, when there is a timeout waiting for a position in the queue to execute the operation, it seems we are not handling correctly the spinner, and the error message is being displayed in the same line than the spinner. We should fix it. Also, I would review the messages with Cody (this can be done later, but we should review the messages we show to the users)

Screenshot 2025-12-11 at 16 28 55

I will keep testing it, and reviewing it, but in case we want to start tackling this

pkg/build/buildkit/connector/portforward.go
```diff
@@ -175,33 +191,81 @@ func getPortForwardK8sClient(oktetoClient *okteto.Client, okCtx PortForwarderOkt
...
+			select {
+			case <-pf.forwarder.readyChan:
+				pf.mu.Lock()
+				pf.isActive = true
+				pf.mu.Unlock()
+				pf.ioCtrl.Logger().Infof("port forward to buildkit pod %s is ready on 127.0.0.1:%d", response.PodName, pf.forwarder.localPort)
+				return nil
+			case <-ctx.Done():
+				pf.Stop()
+				return fmt.Errorf("context cancelled while waiting for port forward: %w", ctx.Err())
```
ifbyol, line 243: Could we extract this to its own function for readability?

jLopezbarb: I'll do it in the next PR. Adding it here would result in a conflict that will be difficult to see

ifbyol (approved):

PR #4847, opened 2025-12-12
feature(buildkit): add reconnection logic to port forward connector

jLopezbarb:

# Proposed changes

Fixes DEV-1224

## Problem

When a build request fails with a non-transient error (such as the BuildKit pod being evicted, network disruption, or pod failure), the build would fail permanently. The system needed to detect these scenarios and restart the build loop by requesting a new available BuildKit pod.

## Solution
The buildkitConnector interface now encapsulates the entire connection lifecycle:
```golang
type buildkitConnector interface {
    WaitUntilIsReady(ctx context.Context) error
    Stop()
    GetBuildkitClient(ctx context.Context) (*client.Client, error)
}
```

When a build fails, calling `Stop()` resets the connection state, allowing the next iteration to request a fresh BuildKit pod via `WaitUntilIsReady()`.

## Changes

- Modified buildkitConnector interface that manages the full lifecycle of BuildKit connections so instead of returning the waiter and client factory the connector is the one doing tha
- Add reconnection logic to request a new BuildKit pod when non-transient errors occur (e.g., connection lost, pod evicted)
- Refactor PortForwarder to track pod assignment and support graceful reconnection

## Diagram

```mermaid
sequenceDiagram
    participant Runner
    participant Connector
    participant Okteto API
    participant BuildKit Pod

    Runner->>Connector: WaitUntilIsReady()
    Connector->>Okteto API: GetLeastLoadedBuildKitPod()
    Okteto API-->>Connector: Pod A assigned
    Connector->>BuildKit Pod: Establish connection
    Connector-->>Runner: Ready

    Runner->>Connector: GetBuildkitClient()
    Connector-->>Runner: Client
    Runner->>BuildKit Pod: Execute build
    BuildKit Pod--xRunner: ❌ Non-transient error (pod evicted)

    Runner->>Connector: Stop()
    Note over Connector: Reset connection state

    Runner->>Connector: WaitUntilIsReady()
    Connector->>Okteto API: GetLeastLoadedBuildKitPod()
    Okteto API-->>Connector: Pod B assigned
    Connector->>BuildKit Pod: Establish connection
    Connector-->>Runner: Ready

    Runner->>Connector: GetBuildkitClient()
    Connector-->>Runner: Client
    Runner->>BuildKit Pod: Execute build
    BuildKit Pod-->>Runner: ✅ Success
```

## How to validate

1. Run okteto build on a several build services
1. Run kubectl delete pod buildkitPod --force --grace-period=0 -n okteto
1. See that it reconnects

## CLI Quality Reminders 🔧

### DIFF
--- a/pkg/build/buildkit/connector/ingress.go
+++ b/pkg/build/buildkit/connector/ingress.go
@@ -32,19 +32,6 @@ type IngressConnector struct {
 	waiter                *Waiter
 }
 
-// NoOpStarter is a no-op starter for the ingress connector
-type NoOpConnectionManager struct{}
-
-// Start is a no-op for the no-op connection starter
-func (n *NoOpConnectionManager) Start(ctx context.Context) error {
-	return nil
-}
-
-// Stop is a no-op for the no-op connection manager
-func (n *NoOpConnectionManager) Stop() {
-	// No-op: ingress connector doesn't maintain a persistent connection that needs to be closed
-}
-
 // NewIngressConnector creates a new ingress connector. It connects to the buildkit server via ingress.
 func NewIngressConnector(okCtx IngressOktetoContextInterface, ioCtrl *io.Controller) *IngressConnector {
 	buildkitClientFactory := NewBuildkitClientFactory(
--- a/pkg/build/buildkit/connector/portforward.go
+++ b/pkg/build/buildkit/connector/portforward.go
@@ -50,7 +50,7 @@ const (
 	initialPollIntervalPortForward = 1 * time.Second
 	maxPollIntervalPortForward     = 10 * time.Second
 	backoffMultiplierPortForward   = 2.0
-	buildkitPodWaitTimeoutEnvVar   = "OKTETO_BUILDKIT_POD_WAIT_TIMEOUT"
+	buildkitQueueWaitTimeoutEnvVar = "OKTETO_BUILDKIT_QUEUE_WAIT_TIMEOUT"
 )
 
 type PortForwarderOktetoContextInterface interface {
@@ -80,6 +80,7 @@ type PortForwarder struct {
 	isActive       bool
 	mu             sync.Mutex
 	buildkitClient *client.Client
+	waiter         *Waiter
 }
 
 // NewPortForwarder creates a new port forwarder. It forwards the port to the buildkit server.
@@ -100,7 +101,8 @@ func NewPortForwarder(ctx context.Context, okCtx PortForwarderOktetoContextInter
 	if err != nil {
 		return nil, fmt.Errorf("could not get available port: %w", err)
 	}
-	maxWaitTime := env.LoadTimeOrDefault(buildkitPodWaitTimeoutEnvVar, defaultMaxWaitTimePortForward)
+	maxWaitTime := env.LoadTimeOrDefault(buildkitQueueWaitTimeoutEnvVar, defaultMaxWaitTimePortForward)
+	waiter := NewBuildkitClientWaiter(ioCtrl)
 
 	pf := &PortForwarder{
 		sessionID:    sessionID,
@@ -110,11 +112,13 @@ func NewPortForwarder(ctx context.Context, okCtx PortForwarderOktetoContextInter
 		okCtx:        okCtx,
 		ioCtrl:       ioCtrl,
 		maxWaitTime:  maxWaitTime,
+		waiter:       waiter,
 		stopChan:     make(chan struct{}, 1),
 		readyChan:    make(chan struct{}, 1),
 		localPort:    port,
 	}
 
+	// We need to call it once in order to check if the buildkit pod endpoint is available
 	response, err := pf.oktetoClient.Buildkit().GetLeastLoadedBuildKitPod(ctx, pf.sessionID)
 	if err != nil {
 		return nil, fmt.Errorf("could not get least loaded buildkit pod: %w", err)
@@ -155,7 +159,7 @@ func (pf *PortForwarder) Start(ctx context.Context) error {
 		return err
 	}
 
-	if err := pf.waitUntilReady(ctx); err != nil {
+	if err := pf.waitUntilPortForwardIsReady(ctx); err != nil {
 		pf.ioCtrl.Logger().Infof("failed to wait until ready: %s", err)
 		return err
 	}
@@ -257,7 +261,7 @@ func (pf *PortForwarder) establishPortForward() error {
 }
 
 // waitUntilReady waits for the port forward to be ready or context to be cancelled
-func (pf *PortForwarder) waitUntilReady(ctx context.Context) error {
+func (pf *PortForwarder) waitUntilPortForwardIsReady(ctx context.Context) error {
 	pf.ioCtrl.Logger().Infof("waiting for port forward to be ready to pod %s", pf.podName)
 
 	select {
@@ -308,14 +312,15 @@ func getPortForwardK8sClient(oktetoClient *okteto.Client, okCtx PortForwarderOkt
 	return okteto.NewK8sClientProvider().Provide(portForwardCfg)
 }
 
-// WaitUntilIsReady waits for the buildkit server to be ready
+// WaitUntilIsReady waits for the buildkit server to be ready.
+// Any function calling this method must call Start() first and handle the Stop() call.
 func (pf *PortForwarder) WaitUntilIsReady(ctx context.Context) error {
 	if !pf.isActive {
 		if err := pf.Start(ctx); err != nil {
 			return fmt.Errorf("failed to start port forward: %w", err)
 		}
 	}
-	return NewBuildkitClientWaiter(pf.ioCtrl).WaitUntilIsUp(ctx, pf.GetBuildkitClient)
+	return pf.waiter.WaitUntilIsUp(ctx, pf.GetBuildkitClient)
 }
 
 // GetBuildkitClient returns the buildkit client. Start() must be called before this method.


pkg/build/buildkit/connector/portforward.go
```diff
@@ -48,9 +49,11 @@ type PortForwarderOktetoContextInterface interface {
 
 // forwarder represents a port forwarder for the buildkit server.
 type forwarder struct {
```
ifbyol: The fact that this struct and the "public" one have the same description is a smell of something odd, or not right. Why do we need a private struct that has some of the functionality that the public one would have?

pkg/build/buildkit/connector/portforward.go
```diff
@@ -100,16 +103,95 @@ func NewPortForwarder(ctx context.Context, okCtx PortForwarderOktetoContextInter
...
+		maxWaitTime         = 10 * time.Minute
+		initialPollInterval = 1 * time.Second
+		maxPollInterval     = 10 * time.Second
+		backoffMultiplier   = 2.0
+	)
+
+	startTime := time.Now()
+	pollInterval := initialPollInterval
+
+	sp := pf.ioCtrl.Out().Spinner("Waiting for BuildKit pod to become available...")
```
ifbyol, line 176: Nit. To review later with the general experience, but I don't know if this applies anymore. Maybe we don't even need to wait. We are just connecting at this point. Later it will have to wait or not

pkg/build/buildkit/connector/portforward.go
```diff
@@ -100,16 +103,95 @@ func NewPortForwarder(ctx context.Context, okCtx PortForwarderOktetoContextInter
...
+			return fmt.Errorf("timeout waiting for buildkit pod after %v: please contact your cluster administrator to increase the maximum number of BuildKit instances or adjust the metrics thresholds", maxWaitTime)
+		}
+
+		response, err := pf.oktetoClient.Buildkit().GetLeastLoadedBuildKitPod(ctx, pf.sessionID)
+		if err != nil {
+			return fmt.Errorf("could not get least loaded buildkit pod: %w", err)
+		}
+
+		if response.PodName != "" {
+			pf.forwarder.podName = response.PodName
```
ifbyol: Why do we need to keep an internal reference of the pod name in the struct? I don't think we are taking some advantage from it, as don't executing some logic if it is already calculated or things like that.

This could be perfectly a return of the function and we reduce the "state" of the forwarder struct if it is not being used

pkg/build/buildkit/runner.go
```diff
@@ -54,14 +54,24 @@ type runnerMetadata struct {
...
 	metadata                           *runnerMetadata
 	maxAttemptsBuildkitTransientErrors int
 }
+
+// buildkitConnector is the interface for the buildkit connector
+type buildkitConnector interface {
+	// WaitUntilIsReady waits for the buildkit server to be ready
+	WaitUntilIsReady(ctx context.Context) error
+	// Stop closes the connection to the buildkit server
+	Stop()
```
ifbyol, line 81: As we discussed offline, it seems odd from a client perspective that you offer a `Stop` function, but not having a `Start`, and that operation is done under the hood in `GetBuildkitClient`, which you wouldn't expect to. We should try to rethink this, and maybe expose the start function.

pkg/build/buildkit/connector/portforward.go
```diff
@@ -183,88 +276,24 @@ func getPortForwardK8sClient(oktetoClient *okteto.Client, okCtx PortForwarderOkt
...
-		case <-time.After(pollInterval):
-			pollInterval = time.Duration(float64(pollInterval) * backoffMultiplier)
-			if pollInterval > maxPollInterval {
-				pollInterval = maxPollInterval
-			}
-		case <-ctx.Done():
-			return fmt.Errorf("context cancelled while waiting for buildkit pod: %w", ctx.Err())
+// GetClientFactory returns the client factory
+func (pf *PortForwarder) GetBuildkitClient(ctx context.Context) (*client.Client, error) {
+	if !pf.isActive {
```
ifbyol, line 318: As we discussed offline, I think this `if` is not working as expected, as we are closing the connection between "sections" (build/deploy), this `isActive` would be false the second time. I tested it, and I had a wait between build and deploy to get a new BuildKit pod. Let's clarify if we changed the behavior due to some conversation, but I think it should not wait again

pkg/build/buildkit/connector/portforward.go
jLopezbarb: We will need it between builds and other commands in order to connect to the latest buildkit pod

pkg/build/buildkit/connector/portforward.go
jLopezbarb: Probably need to be review later, but we will need to have a message here in order to get the least loaded buildkit until we get one

pkg/build/buildkit/connector/portforward.go
jLopezbarb: fixed on 01efde7a8f2691f67a0e9848514b83d8b1d83196

pkg/build/buildkit/runner.go
jLopezbarb: fixed on: [d50d699](https://github.com/okteto/okteto/pull/4847/commits/d50d699d619bfbb629cbbf527c162ee69a6785b4)

pkg/build/buildkit/connector/wait.go
```diff
@@ -110,6 +122,13 @@ func (bw *Waiter) WaitUntilIsUp(ctx context.Context) error {
 			bw.waitingTime = bw.maxWaitTime
 			return fmt.Errorf("buildkit service not available for %v", bw.maxWaitTime)
 		default:
+			if err := bw.connectionManager.Start(ctx); err != nil {
+				bw.logger.Infof("Failed to start BuildKit service: %v\n", err)
```
ifbyol: Nit (or to be addressed later). We are not starting BuildKit service, we are just starting a  connection with Buildkit

pkg/build/buildkit/connector/portforward.go
```diff
@@ -190,93 +308,22 @@ func getPortForwardK8sClient(oktetoClient *okteto.Client, okCtx PortForwarderOkt
...
-				pollInterval = maxPollInterval
-			}
-		case <-ctx.Done():
-			return fmt.Errorf("context cancelled while waiting for buildkit pod: %w", ctx.Err())
+	if !pf.isActive {
+		if err := pf.Start(ctx); err != nil {
+			return fmt.Errorf("failed to start port forward: %w", err)
 		}
 	}
+	return NewBuildkitClientWaiter(pf.ioCtrl).WaitUntilIsUp(ctx, pf.GetBuildkitClient)
```
ifbyol: Nit: If this doesn't keep any state, we can include it as part of the forwarder as we do with the ingress one instead of creating it each time this function is called

pkg/build/buildkit/connector/portforward.go
```diff
@@ -190,93 +308,22 @@ func getPortForwardK8sClient(oktetoClient *okteto.Client, okCtx PortForwarderOkt
 
 // WaitUntilIsReady waits for the buildkit server to be ready
 func (pf *PortForwarder) WaitUntilIsReady(ctx context.Context) error {
```
ifbyol, line 317: Should we add a comment in this function for any client indicating that the start should be also called before, and they should take care of calling `stop`?

pkg/build/buildkit/connector/portforward.go
```diff
@@ -125,55 +227,71 @@ func (pf *PortForwarder) Start(ctx context.Context, podName string, remotePort i
...
+			pf.mu.Unlock()
+			pf.Stop()
 		}
 	}()
 
 	return nil
 }
 
+// waitUntilReady waits for the port forward to be ready or context to be cancelled
+func (pf *PortForwarder) waitUntilReady(ctx context.Context) error {
```
ifbyol: I would rename this function to make a clear distinction with `WaitUntilIsReady`. They have 2 different purposes and they generate confusion

pkg/build/buildkit/connector/ingress.go
```diff
@@ -31,33 +32,51 @@ type IngressConnector struct {
...
+
+// Start is a no-op for the no-op connection starter
+func (n *NoOpConnectionManager) Start(ctx context.Context) error {
+	return nil
+}
+
+// Stop is a no-op for the no-op connection manager
+func (n *NoOpConnectionManager) Stop() {
+	// No-op: ingress connector doesn't maintain a persistent connection that needs to be closed
+}
```
ifbyol: If I'm not wrong, this is not being used and it can be deleted

pkg/build/buildkit/connector/portforward.go
```diff
@@ -43,6 +45,14 @@ var waitReasonMessages = map[string]string{
 	"PODS_SCALING":      "BuildKit pods are starting up",
 }
 
+const (
+	defaultMaxWaitTimePortForward  = 10 * time.Minute
+	initialPollIntervalPortForward = 1 * time.Second
+	maxPollIntervalPortForward     = 10 * time.Second
+	backoffMultiplierPortForward   = 2.0
+	buildkitPodWaitTimeoutEnvVar   = "OKTETO_BUILDKIT_POD_WAIT_TIMEOUT"
```
ifbyol:
```suggestion
	buildkitPodWaitTimeoutEnvVar   = "OKTETO_BUILDKIT_QUEUE_WAIT_TIMEOUT"
```

As we are going to give more relevance to the queue (it is going to be even documented), I would be explicit about it in the var name

pkg/build/buildkit/runner.go
```diff
@@ -52,58 +55,121 @@ type runnerMetadata struct {
...
 	attempts := 0
 	var solveTime time.Duration
 	defer func() {
 		r.metadata.attempts = attempts
 		r.metadata.solveTime = solveTime
 	}()
+
+	var showBuildMessage bool
+	switch buildOptions.OutputMode {
+	case "test", "deploy", "destroy":
```
ifbyol, line 144: Nit: Didn't we have constants for these values?

jLopezbarb: yes, but it's on the build package and it would result on a cycle import

ifbyol: but they can be moved or created another ones here, can't they?

jLopezbarb: Will do in another PR

PR #4851, opened 2025-12-18
feat: add incluster buildkit connector

jLopezbarb:

# Proposed changes

Fixes DEV-1270

## Overview

This PR introduces InClusterConnector, a new BuildKit connector optimized for in-cluster execution scenarios. When Okteto CLI runs inside the cluster (remote deploys, pipeline installer, or managed namespace pods), it can now connect directly to BuildKit pods via their internal IP, bypassing the ingress layer.

## Problem

Previously, all BuildKit connections went through the Ingress (public endpoint), even when the CLI was running inside the same cluster:

```
┌─────────────────────────────────────────────────────────────────┐
│                        Kubernetes Cluster                       │
│  ┌──────────────┐                               ┌──────────────┐  │
│  │  Remote Pod  │ ──── Internet ────► Ingress ─►│  BuildKit    │   │
│  │  (okteto)    │      (unnecessary)           │  Pod         │          │
│  └──────────────┘                               └──────────────┘  │
└─────────────────────────────────────────────────────────────────┘
```

## Solution

The new InClusterConnector connects directly to BuildKit pods via internal IP:

```
┌─────────────────────────────────────────────────────────────────┐
│                        Kubernetes Cluster                        │
│  ┌──────────────┐         direct TCP          ┌──────────────┐  │
│  │  Remote Pod  │ ─────────────────────────► │  BuildKit    │  │
│  │  (okteto)    │       (pod IP:1234)         │  Pod         │  │
│  └──────────────┘                              └──────────────┘  │
└─────────────────────────────────────────────────────────────────┘
```

## How was it solved:

- Adds a new connector to the connector selection. This new connector is used when you are running inside an okteto pod (OKTETO_MANAGED_POD) in a remote execution or in an installer job
- We verify if the connection to the cluster is correct and if not retry the connection
- There is a graceful reconnection if the build fails with a retryable error
- Fallback to ingress if the API doesn't return a buildkit pod correctly

### Connector Selector

```mermaid
flowchart TB
    subgraph "Connector Selection"
        CLI[Okteto CLI] --> Check{Where am I running?}
        
        Check -->|"OKTETO_DEPLOY_REMOTE=true<br/>or OKTETO_IN_INSTALLER=true<br/>or OKTETO_MANAGED_POD=true"| IC[InClusterConnector]
        Check -->|"OKTETO_BUILD_QUEUE_ENABLED=true"| PF[PortForwarder]
        Check -->|"Default (local machine)"| IG[IngressConnector]
    end

    subgraph "Connection Methods"
        IC -->|"tcp://pod-ip:1234"| BK[BuildKit Pod]
        PF -->|"localhost:random → pod:1234"| BK
        IG -->|"tcp://buildkit.okteto.dev"| Ingress[Ingress] --> BK
    end

    style IC fill:#51cf66,stroke:#333
    style PF fill:#339af0,stroke:#333
    style IG fill:#ffd43b,stroke:#333
```

## Comparison:

| Connector           | Use Case                             | Connection Path                | Network Dependency     |
|---------------------|--------------------------------------|------------------------|-------------------------|
| InClusterConnector  | Remote commands, Installer, Managed pods | Direct to Pod IP             | Internal only          |
| PortForwarder       | Local CLI with queue enabled         | Port forward to Pod    | Requires K8s API       |
| IngressConnector    | Local CLI (default)                  | Via public Ingress           | External network       |

## How to validate

1. Using this CLI in an autocreate up session, run `okteto build -l debug` and check that is working 
1. Using this CLI in an autocreate up session, run `okteto build -l debug`, kill the pod in the middle of the execution and wait for the reconnection 

## CLI Quality Reminders 🔧

### DIFF
--- a/pkg/build/buildkit/connector/incluster_test.go
+++ b/pkg/build/buildkit/connector/incluster_test.go
@@ -20,64 +20,6 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func TestInClusterConnector_Stop(t *testing.T) {
-	tests := []struct {
-		name          string
-		initialPodIP  string
-		expectedPodIP string
-	}{
-		{
-			name:          "stop with podIP clears it",
-			initialPodIP:  "10.0.0.1",
-			expectedPodIP: "",
-		},
-		{
-			name:          "stop without podIP does nothing",
-			initialPodIP:  "",
-			expectedPodIP: "",
-		},
-	}
-
-	for _, tt := range tests {
-		t.Run(tt.name, func(t *testing.T) {
-			ic := &InClusterConnector{
-				podIP:  tt.initialPodIP,
-				ioCtrl: io.NewIOController(),
-			}
-
-			// Should not panic
-			require.NotPanics(t, func() {
-				ic.Stop()
-			})
-
-			require.Equal(t, tt.expectedPodIP, ic.podIP)
-		})
-	}
-}
-
-func TestInClusterConnector_Stop_MultipleCallsSafe(t *testing.T) {
-	ic := &InClusterConnector{
-		podIP:  "10.0.0.1",
-		ioCtrl: io.NewIOController(),
-	}
-
-	// First stop should work
-	require.NotPanics(t, func() {
-		ic.Stop()
-	})
-	require.Equal(t, "", ic.podIP)
-
-	// Second stop should also not panic (idempotent)
-	require.NotPanics(t, func() {
-		ic.Stop()
-	})
-
-	// Third stop should still not panic
-	require.NotPanics(t, func() {
-		ic.Stop()
-	})
-}
-
 func TestInClusterConnector_GetBuildkitClient_NoPodIP(t *testing.T) {
 	ic := &InClusterConnector{
 		podIP:  "",


pkg/cmd/build/build.go
```diff
@@ -91,23 +91,50 @@ type OktetoRegistryInterface interface {
...
+
+	return connector.NewIngressConnector(okCtx, logger)
+}
+
+// shouldUseInClusterConnector returns true when running inside an Okteto-managed environment
+// where we can connect directly to BuildKit via pod IP
+func shouldUseInClusterConnector() bool {
+	return env.LoadBoolean(constants.OktetoDeployRemote) || // Remote commands (deploy --remote, destroy --remote, test)
+		config.RunningInInstaller() || // Pipeline installer
+		env.LoadBoolean("OKTETO_MANAGED_POD") // Pods in managed namespaces
```
ifbyol: Would it make sense to extract a constant, too?

pkg/cmd/build/build.go
```diff
@@ -91,23 +91,50 @@ type OktetoRegistryInterface interface {
...
+}
+
+func newInClusterConnectorWithFallback(okCtx OktetoContextInterface, logger *io.Controller) BuildkitConnector {
+	conn, err := connector.NewInClusterConnector(context.Background(), okCtx, logger)
+	if err != nil {
+		logger.Infof("could not create in-cluster connector: %s, falling back to ingress", err)
+		logger.Out().Warning("Could not create in-cluster buildkit connector, falling back to ingress connector")
```
ifbyol: Nit: I don't think this customer face warning means something to the final user. The don't probably now what it means. To review later on

pkg/build/buildkit/connector/incluster.go
```diff
@@ -0,0 +1,245 @@
...
+	ic.mu.Lock()
+	hasPodIP := ic.podIP != ""
+	ic.mu.Unlock()
+
+	if !hasPodIP {
+		if err := ic.Start(ctx); err != nil {
+			return fmt.Errorf("failed to start in-cluster connection: %w", err)
+		}
+	}
+	return ic.waiter.WaitUntilIsUp(ctx, ic.GetBuildkitClient)
```
ifbyol: As we discussed offline, this could make the operation to be stuck waiting until timeout if the podIP changes between we get it, and this point. I think this connector wouldn't need this and just trust that the IP will be available. Then, we should make sure that any error trying to connect to BuildKit is retryable, so a new podIP is requested.

Try the different things we discussed, and see how it behaves

pkg/build/buildkit/connector/incluster.go
```diff
@@ -0,0 +1,226 @@
...
+				pollInterval = maxPollIntervalPortForward
+			}
+		case <-ctx.Done():
+			return "", fmt.Errorf("context cancelled while waiting for buildkit pod: %w", ctx.Err())
+		}
+	}
+}
+
+// createBuildkitClient creates a buildkit client with TLS certificates (like PortForwarder)
+func (ic *InClusterConnector) createBuildkitClient(ctx context.Context) (*client.Client, error) {
```
ifbyol, line 181: Nit: as we talked offline, possible optimization for the future. Make a singleton client, and detect if the `ic.podIP` is the same that the one used for the existent client, and if don't, create a new one

jLopezbarb: It's not trivial to get it from the Buildkit client so I will try to do it in another one

ifbyol: ok, sounds good!

PR #4852, opened 2025-12-19
feat: enhance buildkit connectors with metrics tracking

jLopezbarb:
# Proposed changes

Fixes DEV-1225

This PR adds comprehensive analytics tracking for BuildKit connectors to understand their behavior and performance. The metrics are sent to Mixpanel under the BuildkitConnection event.

## Tracked Metrics

| Property                         | Type   | Description                                                                 |
|----------------------------------|--------|-----------------------------------------------------------------------------|
| sessionId                        | string | Unique identifier for the connection session (for filtering in Mixpanel)   |
| connectorType                    | string | Type of connector used: ingress, portforward, or incluster                  |
| success                          | bool   | Whether the connection was established successfully                         |
| wasFallback                      | bool   | Whether this connector was used as a fallback (e.g., incluster failed and fell back to ingress) |
| wasQueued                        | bool   | Whether the user was placed in a queue at any point (queuePosition > 0)     |
| queueWaitDurationSeconds         | float  | Total time spent waiting for a BuildKit pod assignment                      |
| maxQueuePosition                 | int    | Highest queue position observed during the connection attempt               |
| queueReason                      | string | Last observed queue reason: QUEUE_POSITION, NO_PODS_AVAILABLE, ALL_PODS_BUSY, PODS_SCALING |
| serviceReadyDurationSeconds      | float  | Time spent waiting for BuildKit service to respond after connection         |
| podReused                        | bool   | Whether an existing pod was reused (same podName/podIP from previous connection) |
| waitingForPodTimedOut            | bool   | Whether the connection failed due to timeout while waiting for a pod        |

## What I tried to capture:

- Connector distribution: Which connector types are being used most frequently
- Queue behavior: How often users are queued, how long they wait, and why
- Fallback patterns: How often preferred connectors fail and fall back to ingress
- Performance: Connection times and service readiness latency
- Pod reuse efficiency: How often existing pods are being reused
- Timeout analysis: Correlation between timeouts and queue reasons

## CLI Quality Reminders 🔧

### DIFF
--- a/pkg/cmd/build/build.go
+++ b/pkg/cmd/build/build.go
@@ -124,7 +124,6 @@ func newInClusterConnectorWithFallback(okCtx OktetoContextInterface, logger *io.
 	conn, err := connector.NewInClusterConnector(context.Background(), okCtx, logger)
 	if err != nil {
 		logger.Infof("could not create in-cluster connector: %s, falling back to ingress", err)
-		logger.Out().Warning("Could not create in-cluster buildkit connector, falling back to ingress connector")
 		fallbackConn := connector.NewIngressConnector(okCtx, logger)
 		return fallbackConn
 	}


pkg/analytics/buildkit_connector_test.go
```diff
@@ -0,0 +1,223 @@
...
+			assert.Equal(t, tt.expected, props)
+		})
+	}
+}
+
+func TestConnectorTypeConstants(t *testing.T) {
+	assert.Equal(t, ConnectorType("ingress"), ConnectorTypeIngress)
+	assert.Equal(t, ConnectorType("portforward"), ConnectorTypePortForward)
+	assert.Equal(t, ConnectorType("incluster"), ConnectorTypeInCluster)
+}
```
ifbyol: Not sure this test provides any value. Maybe we could remove it?

pkg/build/buildkit/connector/incluster.go
```diff
@@ -218,9 +234,33 @@ func (ic *InClusterConnector) GetBuildkitClient(ctx context.Context) (*client.Cl
...
+	ic.mu.Lock()
+	hasPodIP := ic.podIP != ""
+	ic.mu.Unlock()
+
+	if !hasPodIP {
+		if err := ic.Start(ctx); err != nil {
+			return fmt.Errorf("failed to start in-cluster connection: %w", err)
+		}
+	}
+	return ic.waiter.WaitUntilIsUp(ctx, ic.GetBuildkitClient)
```
ifbyol: Was this a bug? Or why are we adding now the implementation of the function in the `incluster` connector?

pkg/build/buildkit/connector/incluster.go
```diff
@@ -218,9 +234,33 @@ func (ic *InClusterConnector) GetBuildkitClient(ctx context.Context) (*client.Cl
...
+	ic.mu.Lock()
+	defer ic.mu.Unlock()
+
+	if ic.podIP == "" {
+		ic.ioCtrl.Logger().Infof("in-cluster connection has no pod IP assigned")
+		return
+	}
+
+	ic.ioCtrl.Logger().Infof("clearing pod IP %s, will request new pod on next Start()", ic.podIP)
+	ic.podIP = ""
```
ifbyol: Same about this function. Why are we adding some implementation now with the metrics? Was it missing from the first implementation, or is it not intended to be added in this PR?

pkg/build/buildkit/connector/incluster.go
```diff
@@ -218,9 +234,33 @@ func (ic *InClusterConnector) GetBuildkitClient(ctx context.Context) (*client.Cl
...
+}
+
+// GetMetrics returns the connector metrics for external configuration
+func (ic *InClusterConnector) GetMetrics() *ConnectorMetrics {
+	return ic.metrics
```
ifbyol: Why creating this getter instead of just making metrics exposed publicly so it can be used outside? That is the golang way of doing it instead of doing these getters. I think it will be cleaner in Go.

ifbyol (commented):
I still need to review more things, but I wanted to post this comments for now as they are relevant questions or things to change

pkg/cmd/build/build.go
```diff
@@ -124,7 +124,10 @@ func newInClusterConnectorWithFallback(okCtx OktetoContextInterface, logger *io.
 	conn, err := connector.NewInClusterConnector(context.Background(), okCtx, logger)
 	if err != nil {
 		logger.Infof("could not create in-cluster connector: %s, falling back to ingress", err)
-		return connector.NewIngressConnector(okCtx, logger)
+		logger.Out().Warning("Could not create in-cluster buildkit connector, falling back to ingress connector")
```
ifbyol: I don't think this is relevant for the customers, or this will be customer-facing. The InCluster connector or the ingress connector is more like an internal thing, so I wouldn't show a warning in this case.

ifbyol (approved):

PR #4858, opened 2026-01-07
Support for helm4

pchico83:
This PR adds opt-in support for Helm 4 by:
- Support server-side apply in the k8s api server local proxy. Helm 4 uses server-side applies. This is needed to group resources in the Okteto UI, delete dev environments, ...
- Install Helm 4 in the okteto/okteto image on its own binary `helm4` to make it easier to give it a try

The proxy for server-side apply is based on YAML instead of JSON. Instead of suplicating the logic, the YAML translator converts between YAML and JSON and calls the JSON translator.

I also updated the protobuf translator to follow a similar approach.

### DIFF
--- a/pkg/deployable/proxy_test.go
+++ b/pkg/deployable/proxy_test.go
@@ -287,3 +287,55 @@ func (b *bytesBuffer) Write(p []byte) (n int, err error) {
 func int32Ptr(i int32) *int32 {
 	return &i
 }
+
+// TestPartialDeploymentServerSideApply tests that partial objects (minimal YAML)
+// can be decoded, modified, and encoded without errors
+func TestPartialDeploymentServerSideApply(t *testing.T) {
+	// Simulate a very minimal server-side apply payload
+	// This is the kind of partial object we might receive in a PATCH request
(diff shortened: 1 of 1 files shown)


jLopezbarb (approved):

pkg/deployable/proxy.go
```diff
@@ -183,9 +184,64 @@ func (p *Proxy) SetDivert(driver divert.Driver) {
...
+}
+
+func getEncoder(contentType string) runtime.Encoder {
+	switch contentType {
+	case "application/vnd.kubernetes.protobuf":
+		return protobuf.NewSerializer(scheme.Scheme, scheme.Scheme)
+	case "application/json", "application/apply-patch+yaml":
+		return json.NewSerializer(json.DefaultMetaFactory, scheme.Scheme, scheme.Scheme, false)
+	default:
+		return json.NewSerializer(json.DefaultMetaFactory, scheme.Scheme, scheme.Scheme, false)
```
ifbyol, line 244: These 2 branches are the same. I'm fine if we want to keep then separately for some reason. Mentioning it just in case

PR #4865, opened 2026-01-15
refactor: enhance buildkit connector metrics with error tracking

jLopezbarb:

# Proposed changes

- Removes ingress connector metrics entirely (default path needs no tracking)
- Adds `errReason` field to track specific failure causes:
  - `QueueTimeout` - timeout waiting for buildkit pod
  - `PortForwardCreation` - failed to establish port forward
  - `ContextCancelled` - context cancelled during operation
  - `IncompatibleBackend` - backend doesn't support `getLeastLoadedBuildKitPod`
  - `BackendInternalError` - internal error from backend API
- Removes redundant metric fields: `wasQueued`, `wasFallback`, `waitingForPodTimedOut`, `podReused`, `serviceReadyDuration`
- Configures waiter for exactly 3 retry attempts in portforward and incluster connectors
- Fixes duplicate metric tracking where success/failure could be tracked twice in same execution

## CLI Quality Reminders 🔧

### DIFF
--- a/pkg/build/buildkit/connector/portforward.go
+++ b/pkg/build/buildkit/connector/portforward.go
@@ -107,7 +107,7 @@ func NewPortForwarder(ctx context.Context, okCtx PortForwarderOktetoContextInter
 		return nil, fmt.Errorf("could not get available port: %w", err)
 	}
 	maxWaitTime := env.LoadTimeOrDefault(buildkitQueueWaitTimeoutEnvVar, defaultMaxWaitTimePortForward)
-	// Configure waiter for 3 attempts: 5s retry interval * 3 = 15s max wait time
+	// Configure waiter for 3 attempts: 1s retry interval * 4 = 4s max wait time
 	waiter := NewBuildkitClientWaiterWithConfig(ioCtrl, 4*time.Second, 1*time.Second)
 
 	pf := &PortForwarder{


pkg/build/buildkit/connector/incluster.go
```diff
@@ -64,7 +64,7 @@ func NewInClusterConnector(ctx context.Context, okCtx InClusterOktetoContextInte
 
 	sessionID := uuid.New().String()
 	maxWaitTime := env.LoadTimeOrDefault(buildkitQueueWaitTimeoutEnvVar, defaultMaxWaitTimePortForward)
-	waiter := NewBuildkitClientWaiter(ioCtrl)
+	waiter := NewBuildkitClientWaiterWithConfig(ioCtrl, 4*time.Second, 1*time.Second)
```
ifbyol, line 67: It seems we are using always the same values for the maxWaitTime and the retryInterval. Should we then use/initialize it internally in the function `NewBuildkitClientWaiterWithConfig`? Or is it not expected to have the same values?

pkg/build/buildkit/connector/portforward.go
```diff
@@ -106,7 +107,8 @@ func NewPortForwarder(ctx context.Context, okCtx PortForwarderOktetoContextInter
 		return nil, fmt.Errorf("could not get available port: %w", err)
 	}
 	maxWaitTime := env.LoadTimeOrDefault(buildkitQueueWaitTimeoutEnvVar, defaultMaxWaitTimePortForward)
-	waiter := NewBuildkitClientWaiter(ioCtrl)
+	// Configure waiter for 3 attempts: 5s retry interval * 3 = 15s max wait time
+	waiter := NewBuildkitClientWaiterWithConfig(ioCtrl, 4*time.Second, 1*time.Second)
```
ifbyol, line 111: Related to a previous comment. The values specified don't match with the comment. Or at least, It is not obvious. In the comment it says that the final max wait time is 15 seconds, but according to the function signature, that should be the second parameter, which is 4s. What is the expected value to be used?

ifbyol (approved):
Approving it as it looks good, but review the comment about the times used in the function `NewBuildkitClientWaiterWithConfig` before merge, please

PR #4875, opened 2026-01-21
fix multiple log with same value on buildkit

jLopezbarb:
# Proposed changes

Fixes DEV-1283

There were several logs duplicated. One per iteration on the waiting spininner which was resolved by just rewriting when the message change and another one for the build error message. This last one was fixed by not adding to the configmap the error and letting the error be written by the last error. I didn't want to affect other parts of the code so I added a new param to the function 

## How to validate

1. Saturate the queue
1. Run a okteto deploy that times out on the queue
1. See that there is no duplication

## CLI Quality Reminders 🔧

### DIFF
--- a/cmd/deploy/deploy.go
+++ b/cmd/deploy/deploy.go
@@ -441,7 +441,7 @@ func (dc *Command) Run(ctx context.Context, deployOptions *Options) error {
 	os.Setenv(constants.OktetoNameEnvVar, deployOptions.Name)
 
 	if err := dc.deployDependencies(ctx, deployOptions); err != nil {
-		if errStatus := dc.CfgMapHandler.UpdateConfigMap(ctx, cfg, data, err, true); errStatus != nil {
+		if errStatus := dc.CfgMapHandler.UpdateConfigMap(ctx, cfg, data, err); errStatus != nil {
 			return errStatus
 		}
 		return err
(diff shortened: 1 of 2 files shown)


cmd/deploy/deploy.go
```diff
@@ -452,7 +452,7 @@ func (dc *Command) Run(ctx context.Context, deployOptions *Options) error {
 	}
 
 	if err := buildImages(ctx, dc.Builder, dc.CfgMapHandler, deployOptions); err != nil {
-		if errStatus := dc.CfgMapHandler.UpdateConfigMap(ctx, cfg, data, err); errStatus != nil {
+		if errStatus := dc.CfgMapHandler.UpdateConfigMap(ctx, cfg, data, err, false); errStatus != nil {
```
ifbyol: I guess this si affecting to any error returned by the build operation, and I guess this was happening already for other errors (as we are doing the same in the rest of cases, log the error and then, return it). And most likely happens for other stages like DeployDependencies. If that is the case, I wouldn't fix it for this specific case. It is something that keeps happening for a long time, and the fix has to be more in the foundations on how we log things.

So, let's review if this happens for other stages too, and if it does, I wouldn't do this.

Also, If I'm not wrong, this would have an issue when the installer job is gone, as we would read the log from the configmap, and the error would be gone (but I'm not sure if the final error returned by the CLI is stored in the configmap or not)

jLopezbarb: If it's executed through the installer it will preserve, resulting in a duplication, but if you execute it. from the CLI it will be lost.

I'm reverting the change and only merging with the spinner fix

PR #4886, opened 2026-01-27
Add pod affinity for persistent volumes in translation logic

jLopezbarb:
- Introduced volume labels to enable pod affinity for shared persistent volumes, ensuring that pods sharing the same volume are scheduled on the same node.
- Updated the TranslationRule structure to include MainVolumeName and VolumeAccessMode for better handling of persistent volumes.
- Enhanced test cases to validate the new volume affinity behavior.

These changes improve the scheduling of pods that require access to the same persistent volume, addressing issues with ReadWriteOnce volume claims.

## CLI Quality Reminders 🔧

### DIFF
--- a/pkg/k8s/apps/translate.go
+++ b/pkg/k8s/apps/translate.go
@@ -115,7 +115,7 @@ func (tr *Translation) translate() error {
 	for _, rule := range tr.Rules {
 		devContainer := GetDevContainer(tr.DevApp.PodSpec(), rule.Container)
 		TranslateDevContainer(devContainer, rule)
-		TranslatePodSpec(tr.DevApp.PodSpec(), rule)
+		TranslatePodSpec(tr.DevApp.PodSpec(), rule, tr.MainDev.Name)
 		TranslateOktetoDevSecret(tr.DevApp.PodSpec(), tr.Dev.Name, rule.Secrets)
 
 		if rule.IsMainDevContainer() {
@@ -191,14 +191,14 @@ func TranslateDevContainer(c *apiv1.Container, rule *model.TranslationRule) {
 	TranslateContainerSecurityContext(c, rule.SecurityContext)
 }
 
-func TranslatePodSpec(podSpec *apiv1.PodSpec, rule *model.TranslationRule) {
+func TranslatePodSpec(podSpec *apiv1.PodSpec, rule *model.TranslationRule, devName string) {
 	TranslateOktetoVolumes(podSpec, rule)
 	TranslatePodSecurityContext(podSpec, rule.SecurityContext)
 	TranslatePodServiceAccount(podSpec, rule.ServiceAccount)
 	TranslatePodPriorityClassName(podSpec, rule.PriorityClassName)
 
 	TranslateOktetoNodeSelector(podSpec, rule.NodeSelector)
-	TranslateOktetoAffinity(podSpec, rule)
+	TranslateOktetoAffinity(podSpec, rule, devName)
 }
 
 // TranslateProbes translates the probes attached to a container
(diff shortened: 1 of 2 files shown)


pkg/k8s/apps/translate.go
```diff
@@ -101,6 +102,12 @@ func (tr *Translation) translate() error {
 		tr.DevApp.TemplateObjectMeta().Labels[model.DetachedDevLabel] = tr.getDevName()
 	}
 
+	// Add volume label to enable pod affinity for shared persistent volumes
+	if tr.MainDev.PersistentVolumeEnabled() {
+		volumeLabel := fmt.Sprintf("dev.okteto.com/volume-%s", tr.MainDev.GetVolumeName())
```
pchico83: This would create affinities to dev containers with the same name in other repos. It might be an edge case,  but I would add the dev environment name to the annotation to make sure it cannot happen

ifbyol: Be aware that the maximum length for name of the label is 63 chars. If we add the dev-environment we should add logic to short it if needed

pkg/k8s/apps/translate.go
```diff
@@ -101,6 +102,12 @@ func (tr *Translation) translate() error {
 		tr.DevApp.TemplateObjectMeta().Labels[model.DetachedDevLabel] = tr.getDevName()
 	}
 
+	// Add volume label to enable pod affinity for shared persistent volumes
+	if tr.MainDev.PersistentVolumeEnabled() {
+		volumeLabel := fmt.Sprintf("dev.okteto.com/volume-%s-%s", tr.MainDev.Name, tr.MainDev.GetVolumeName())
```
ifbyol: as we discussed offline, we should use the dev environment name and the dev container name. We should also be careful with the length of the name. We can truncate it or do a hash of the string

pkg/k8s/apps/translate.go
```diff
@@ -101,6 +104,12 @@ func (tr *Translation) translate() error {
 		tr.DevApp.TemplateObjectMeta().Labels[model.DetachedDevLabel] = tr.getDevName()
 	}
 
+	// Add volume label to enable pod affinity for shared persistent volumes
+	if tr.MainDev.PersistentVolumeEnabled() && len(tr.Rules) > 0 && tr.Rules[0].ManifestName != "" {
+		volumeLabel := getVolumeLabelKey(tr.Rules[0].ManifestName, tr.MainDev.GetVolumeName())
```
pchico83: Since we are computing a hash, it's less relevant, but it should be `tr.MainDev.Name` instead of `tr.MainDev.GetVolumeName()`. `tr.MainDev.GetVolumeName()` is the main container name + `-okteto`

pkg/k8s/apps/translate.go
```diff
@@ -101,6 +104,12 @@ func (tr *Translation) translate() error {
 		tr.DevApp.TemplateObjectMeta().Labels[model.DetachedDevLabel] = tr.getDevName()
 	}
 
+	// Add volume label to enable pod affinity for shared persistent volumes
+	if tr.MainDev.PersistentVolumeEnabled() && len(tr.Rules) > 0 && tr.Rules[0].ManifestName != "" {
```
ifbyol, line 108: Could it happen that the manifest name is empty? I mean, what would happen if the persistent volume is enabled but the rest of the condition is not met? Would it generate some problem where this doesn't work fine?

jLopezbarb: I can't think any scenario that the name would be empty. If it's empty we will use the name of the folder

PR #4898, opened 2026-02-11
Add compose endpoints httproute

jLopezbarb:
## Summary                                                                                                                                                                                                        
                                                                                                                                                                                                                    
  Implements automatic translation of Docker Compose endpoints to Gateway API HTTPRoute resources when gateway metadata is available in the cluster. This removes the need for a feature flag and makes the behavior
   automatic based on cluster configuration.                                                                                                                                                                        
                                                                                                                                                                                                                    
  Fixes DEV-1301                                                                                                                                                                                                    
                                                                                                                                                                                                                    
  ## Changes                                                                                                                                                                                                        
                                                                                                                                                                                                                    
  - **Automatic detection**: Added `shouldUseHTTPRoute()` function to check cluster metadata for gateway configuration                                                                                              
  - **Smart routing**: Automatically uses HTTPRoute when `GatewayName` and `GatewayNamespace` are present in cluster metadata                                                                                       
  - **Graceful fallback**: Falls back to Ingress when gateway is not configured                                                                                                                                     
  - **New package**: Added `pkg/k8s/httproutes/` with translation and CRUD operations for HTTPRoute resources                                                                                                       
  - **Dependency**: Added `sigs.k8s.io/gateway-api v1.2.1` for Gateway API support                                                                                                                                  
                                                                                                                                                                                                                    
  ## Behavior                                                                                                                                                                                                       
                                                                                                                                                                                                                    
  **No feature flag required** - automatic detection based on cluster metadata:                                                                                                                                     
                                                                                                                                                                                                                    
  - ✅ **When cluster has gateway configured**: Uses Gateway API HTTPRoute                                                                                                                                          
  - ✅ **When cluster has no gateway**: Uses Ingress (default, maintains backward compatibility)                                                                                                                    
                                                                                                                                                                                                                    
  ## Implementation Details                                                                                                                                                                                         
                                                                                                                                                                                                                    
  ### Files Modified                                                                                                                                                                                                
                                                                                                                                                                                                                    
  - `pkg/cmd/stack/deploy.go`: Added HTTPRoute deployment logic and automatic selection                                                                                                                             
  - `pkg/k8s/httproutes/crud.go`: HTTPRoute CRUD operations                                                                                                                                                         
  - `pkg/k8s/httproutes/translate.go`: Endpoint → HTTPRoute translation                                                                                                                                             
  - `go.mod` / `go.sum`: Added Gateway API dependency                                                                                                                                                               
  - `pkg/types/cluster_metadata.go`: Gateway metadata fields                                                                                                                                                        
  - `pkg/okteto/secrets.go`: Related GraphQL changes                                                                                                                                                                
                                                                                                                                                                                                                    
  ### Key Functions                                                                                                                                                                                                 
                                                                                                                                                                                                                    
  - `shouldUseHTTPRoute()`: Checks cluster metadata and returns gateway configuration                                                                                                                               
  - `deployK8sEndpointHTTPRoute()`: Deploys HTTPRoute resources for compose endpoints                                                                                                                               
  - `httproutes.Translate()`: Converts Okteto Endpoint model to Gateway API HTTPRoute   \

## Testing
- complete a deploy of a compose.yml

### DIFF
--- a/pkg/cmd/stack/deploy.go
+++ b/pkg/cmd/stack/deploy.go
@@ -139,7 +139,6 @@ func ShouldUseHTTPRoute() (bool, types.ClusterMetadata, error) {
 	return true, metadata, nil
 }
 
-
 func (sd *Stack) RunDeploy(ctx context.Context, s *model.Stack, options *DeployOptions) error {
 
 	analytics.TrackStackWarnings(s.Warnings.NotSupportedFields)


pkg/cmd/stack/deploy.go
```diff
@@ -86,6 +88,36 @@ const (
 	maxRestartsToConsiderFailed = 3
 )
 
+// shouldUseHTTPRoute determines if Gateway API HTTPRoute should be used instead of Ingress
+// Returns true if gateway info is available from cluster metadata
+func shouldUseHTTPRoute(ctx context.Context) (bool, types.ClusterMetadata) {
```
pchico83: We mentioned in the daily to support a feature flag envvar specifying if endpoints are exposed via ingress/gateway, and if it's ot present, defalt to ewhat this function does

pkg/cmd/stack/deploy.go
```diff
@@ -86,6 +88,36 @@ const (
...
+	// Only attempt to get metadata if running in Okteto context
+	if !okteto.GetContext().IsOkteto {
```
pchico83: Is this possible? "deploy" is always in an okteto context, right?

pkg/cmd/stack/deploy.go
```diff
@@ -86,6 +88,36 @@ const (
...
+	}
+
+	// Get cluster metadata to check for gateway configuration
+	oktetoClient, err := okteto.NewOktetoClient()
+	if err != nil {
+		oktetoLog.Debugf("failed to create okteto client for metadata retrieval: %s", err)
+		return false, types.ClusterMetadata{}
+	}
+
+	metadata, err := oktetoClient.User().GetClusterMetadata(ctx, okteto.GetContext().Namespace)
```
pchico83: Are we already querying metadata at the beginning of command execution? Could we reuse that call to avoid another network call?

pkg/cmd/stack/deploy.go
```diff
@@ -163,10 +195,27 @@ func deploy(ctx context.Context, s *model.Stack, c kubernetes.Interface, config
...
-			exit <- fmt.Errorf("error getting ingress client: %w", err)
-			return
+		// Check if we should use Gateway API HTTPRoute instead of Ingress
+		useHTTPRoute, clusterMetadata := shouldUseHTTPRoute(ctx)
+
+		var iClient *ingresses.Client
+		var httpRouteClient *httproutes.Client
+		var err error
+
+		if useHTTPRoute {
```
pchico83: Instead of if/then/else, let's create an interface to improve readability

pkg/k8s/httproutes/crud.go
```diff
@@ -0,0 +1,139 @@
...
+	gatewayclientset "sigs.k8s.io/gateway-api/pkg/client/clientset/versioned"
+)
+
+// Client represents the HTTPRoute client
+type Client struct {
+	gatewayClient gatewayclientset.Interface
+}
+
+// NewHTTPRouteClient creates a new HTTPRoute client from a rest config
+func NewHTTPRouteClient(config *rest.Config) (*Client, error) {
```
pchico83, line 34: In the backend, we have a dedicated package for [k8s-related clients](https://github.com/okteto/app/blob/master/backend/pkg/k8s/client/client.go). Should we follow a similar approach?

pkg/k8s/httproutes/crud.go
```diff
@@ -0,0 +1,139 @@
...
+	}, nil
+}
+
+// Get retrieves the HTTPRoute
+func (c *Client) Get(ctx context.Context, name, namespace string) (*gatewayv1.HTTPRoute, error) {
+	return c.gatewayClient.GatewayV1().HTTPRoutes(namespace).Get(ctx, name, metav1.GetOptions{})
+}
+
+// Create creates an HTTPRoute
+func (c *Client) Create(ctx context.Context, httpRoute *gatewayv1.HTTPRoute) error {
```
pchico83, line 50: Not sure what are the approach we follow in other packages, but I think we should revisit which methods are public/private. For example, looks like we expect the client to call `Deploy` instead of `Create`.
Also, in other similar packages in `pkg/k8s`, `c *Client` is a parameter of the function

pkg/k8s/httproutes/crud.go
jLopezbarb: I followed the same approach that we have for ingress https://github.com/okteto/okteto/blob/585df2a621a595b4073ff179555af78257758e9f/pkg/k8s/ingresses/crud.go#L45-L57 If we want to change it I'd do it as part of another PR and do it in all the other cases

pkg/k8s/httproutes/crud.go
jLopezbarb: Yes, but as the other comment, I followed the same approach as ingress. I'd agree that we should either go with struct or functions (struct are my preference because of dependency injection) but I'd do it as part of another PR.

cmd/context/create.go
```diff
@@ -289,6 +289,14 @@ func (c *Command) initOktetoContext(ctx context.Context, ctxOptions *Options) er
 	okteto.GetContext().CompanyName = clusterMetadata.CompanyName
 	okteto.GetContext().DivertCRDSEnabled = clusterMetadata.DivertCRDSEnabled
 
+	// Populate gateway metadata if available
+	if clusterMetadata.GatewayName != "" && clusterMetadata.GatewayNamespace != "" {
```
pchico83, line 293: Related to the commen here:
https://github.com/okteto/app/pull/9987/changes#r2803797563

I think it's more flexible if we use envvars only. That way, I can try the gatewya translation by running:
```
okteto deploy --var OKTETO_COMPOSE_ENDPOINTS_TYPE=gateway
```

pkg/cmd/stack/deploy.go
```diff
@@ -163,10 +256,37 @@ func deploy(ctx context.Context, s *model.Stack, c kubernetes.Interface, config
...
-		iClient, err := ingresses.GetClient(c)
-		if err != nil {
-			exit <- fmt.Errorf("error getting ingress client: %w", err)
-			return
+		// Check if we should use Gateway API HTTPRoute instead of Ingress and create deployer
+		useHTTPRoute, clusterMetadata := shouldUseHTTPRoute()
+
+		var deployer endpointDeployer
+
+		if useHTTPRoute {
```
pchico83: should be inject the endpoint deployer as a dependency of `sd *Stack`? WDYT?

pkg/cmd/stack/deploy.go
```diff
@@ -178,16 +298,16 @@ func deploy(ctx context.Context, s *model.Stack, c kubernetes.Interface, config
...
 			for _, ingressPort := range ingressPortsToDeploy {
 				ingressName := serviceName
-				// If more than one port, ingressName will have <serviceName>-<PORT>, each port will have an ingress
+				// If more than one port, ingressName will have <serviceName>-<PORT>, each port will have an ingress/httproute
 				if len(ingressPortsToDeploy) > 1 {
 					ingressName = fmt.Sprintf("%s-%d", serviceName, ingressPort.ContainerPort)
 				}
 
-				if err := deployK8sEndpoint(ctx, ingressName, serviceName, ingressPort, s, iClient); err != nil {
+				if err := deployer.DeployServiceEndpoint(ctx, ingressName, serviceName, ingressPort, s); err != nil {
```
pchico83: Something to consider: if the user switches from Ingress to Gateway, we should create the HTTPRoutes and also remove the existing Ingress objects. And viceversa

cmd/context/create.go
ifbyol: But we still need to have the gateway name and gateway namespace for the translations, right?

pchico83: Yes! But I would pick ingres vs gateway only based on `OKTETO_COMPOSE_ENDPOINTS_TYPE`

jLopezbarb: But if it's empty the gateway should we fallback to the ingress? Should we continue and get the error?

cmd/deploy/deploy.go
```diff
@@ -855,13 +856,38 @@ func (dc *Command) deployStack(ctx context.Context, opts *Options) error {
...
+		}
+		oktetoLog.Infof("Using Gateway API HTTPRoute for endpoints (gateway: %s/%s)", clusterMetadata.GatewayNamespace, clusterMetadata.GatewayName)
+		endpointDeployer = stack.NewHTTPRouteDeployer(httpRouteClient, composeSectionInfo.Stack.Name, composeSectionInfo.Stack.Namespace, clusterMetadata)
+	} else {
+		// Create Ingress deployer
+		ingressClient, err := ingresses.GetClient(c)
+		if err != nil {
+			return fmt.Errorf("error getting ingress client: %w", err)
+		}
+		endpointDeployer = stack.NewIngressDeployer(ingressClient, composeSectionInfo.Stack.Name, composeSectionInfo.Stack.Namespace)
```
ifbyol, line 880: I would log some info here indicating we are using the ingress for endpoints (similar to what we do for http route)

pkg/cmd/stack/deploy.go
```diff
@@ -80,12 +82,137 @@ type Stack struct {
...
+	DeployServiceEndpoint(ctx context.Context, name, serviceName string, port model.Port, stack *model.Stack) error
+	// DeployComposeEndpoint deploys an endpoint from the compose endpoints spec
+	DeployComposeEndpoint(ctx context.Context, name string, endpoint model.Endpoint, stack *model.Stack) error
+}
+
+// ShouldUseHTTPRoute determines if Gateway API HTTPRoute should be used instead of Ingress
+// Returns true if gateway should be used, cluster metadata, and an error if configuration is invalid
+// Priority order:
+// 1. OKTETO_COMPOSE_ENDPOINTS_TYPE (feature flag) - if set, takes absolute precedence
+// 2. OKTETO_DEFAULT_GATEWAY_TYPE (default) - only evaluated if feature flag is not set
```
ifbyol, line 102: What do we mean by "default"?

pkg/cmd/stack/deploy.go
```diff
@@ -80,12 +82,137 @@ type Stack struct {
...
+		}
+		oktetoLog.Infof("Using Gateway API HTTPRoute for endpoints (set by %s=gateway)", oktetoDefaultGatewayTypeEnvVar)
+		return true, metadata, nil
+	}
+
+	// Priority 3: Automatic detection based on gateway metadata in context
+	if metadata.GatewayName == "" || metadata.GatewayNamespace == "" {
+		oktetoLog.Infof("Gateway API is not configured in the cluster, using Ingress for endpoints")
+		return false, metadata, nil
+	}
```
ifbyol: Did we agree to rely on this if no environment variable was forced?

pkg/cmd/stack/deploy.go
```diff
@@ -80,12 +82,137 @@ type Stack struct {
...
+
+	return true, metadata, nil
+}
+
+// ingressDeployer deploys endpoints using Kubernetes Ingress
+type ingressDeployer struct {
```
ifbyol: Following the conversation we had yesterday, would it make sense to split each struct and functions to its own file for readability?

pkg/cmd/stack/deploy.go
```diff
@@ -178,16 +306,16 @@ func deploy(ctx context.Context, s *model.Stack, c kubernetes.Interface, config
 				exit <- err
 				return
 			}
-			// get the public ports from the compose service - this will be deployed into ingresses
+			// get the public ports from the compose service - this will be deployed into ingresses/httproutes
 			ingressPortsToDeploy := getSvcPublicPorts(serviceName, s)
 			for _, ingressPort := range ingressPortsToDeploy {
 				ingressName := serviceName
```
ifbyol, line 238: Nit: We should find a name to refer to this now. It doesn't make sense we keep using `ingress` to refer to something generic that could be an ingress or http route. Not mandatory to do it for this PR, but we should find it soon to avoid issues in the future

pkg/cmd/stack/deploy_test.go
```diff
@@ -1107,3 +1108,308 @@ func TestDeployK8sEndpoint(t *testing.T) {
...
+
+			if tt.expectedError {
+				assert.Error(t, err)
+			} else {
+				assert.NoError(t, err)
+				assert.Equal(t, tt.expectedUseRoute, useRoute)
+				assert.Equal(t, tt.expectedMetadata.GatewayName, metadata.GatewayName)
+				assert.Equal(t, tt.expectedMetadata.GatewayNamespace, metadata.GatewayNamespace)
+			}
+		})
```
ifbyol: We agreed to run away from conditionals in tests

pkg/k8s/httproutes/crud_test.go
```diff
@@ -0,0 +1,284 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "fake",
+			Namespace: "test",
+			Labels:    updatedLabels,
+		},
+	}
+	err := hrClient.Update(ctx, updatedHTTPRoute)
+	if err != nil {
+		t.Fatal(err)
+	}
```
ifbyol: We agreed to use `testify` library for test assertions

pkg/k8s/httproutes/translate_test.go
```diff
@@ -0,0 +1,418 @@
...
+			}
+
+			if !reflect.DeepEqual(result.Spec.Rules, tt.expectedHTTPRouteRules) {
+				t.Errorf("Wrong HTTPRoute rules: got '%v', expected '%v'", result.Spec.Rules, tt.expectedHTTPRouteRules)
+			}
+
+			if !reflect.DeepEqual(result.Spec.ParentRefs, tt.expectedParentRefs) {
+				t.Errorf("Wrong HTTPRoute ParentRefs: got '%v', expected '%v'", result.Spec.ParentRefs, tt.expectedParentRefs)
+			}
+		})
```
ifbyol: Same about using `testify` for test assertion

pkg/cmd/stack/deploy.go
ifbyol, line 102: I don't recall why we have 2 different environment variables for the same?

jLopezbarb: Default is the one that is configured as default in the cluster

pkg/cmd/stack/deploy.go
jLopezbarb: No, I thought that if no env var was set we should rely on this, but I'll remove it as agreeded

pkg/cmd/stack/deploy.go
jLopezbarb: We were having one as the feature flag and another one internal from the backend. We started just with the feature flag but we stated to have two in order to cercionate that the cluster is using one but we are using another one

cmd/deploy/deploy.go
```diff
@@ -855,13 +856,39 @@ func (dc *Command) deployStack(ctx context.Context, opts *Options) error {
...
+	}
+
+	var endpointDeployer stack.EndpointDeployer
+	if useHTTPRoute {
+		// Create HTTPRoute deployer
+		httpRouteClient, err := httproutes.NewHTTPRouteClient(cfg)
+		if err != nil {
+			return fmt.Errorf("error creating httproute client: %w", err)
+		}
+		oktetoLog.Infof("Using Gateway API HTTPRoute for endpoints (gateway: %s/%s)", clusterMetadata.GatewayNamespace, clusterMetadata.GatewayName)
```
ifbyol: Nit: The sentence `Using Gateway API HTTPRoute for endpoints` is already logged within `ShouldUseHTTPRoute` function, so it is logged twice:

```
INFO[0012] Using Gateway API HTTPRoute for endpoints (forced by OKTETO_COMPOSE_ENDPOINTS_TYPE=gateway) 
INFO[0012] Using Gateway API HTTPRoute for endpoints (gateway: ifbyol/envoy-gateway) 
```

My first impression was that we were using gateway from 2 different sources. I would log it only once, and put there the reason why that is being used, and the gateway configured. Something like: `Using HTTPRoute for endpoints with the configured gateway %s/%s (forced by ....)`

ifbyol (approved):
I added a minor comment, but it is not a blocker

cmd/deploy/deploy.go
jLopezbarb: I think they are two different things, one is for the gateway/ingress and the other one is telling the user which gateway will use. I'll rewrite the second message in order to debug and know where it's going to go

ifbyol: But the one within `ShouldUseHTTPRoute` is going to be logged each time that function is called, which probably is a couple of times per flow (if no more). Adding a log in an auxiliar function like that one might flood the debug log with repeated messages, and it might be misleading or complicate the comprehension of the rest of logs

PR #4906, opened 2026-02-19
Allow specific analytics events during deploy operations

jLopezbarb:
# Proposed changes

This PR fixes the blanket blocking of analytics events during deploy operations by introducing an allowlist-based approach.

Previously, when `okteto deploy` ran, it set `OKTETO_WITHIN_DEPLOY_COMMAND_CONTEXT=true`, which caused **all** analytics events to be blocked in the `track()` function. This prevented tracking of Build events, Management events, and Deploy Stack events that occur during deploy operations.

## Changes

- Added `eventsAllowedDuringDeploy` map in `pkg/analytics/track.go` with 12 events:
  - **Build events (6)**: Build, BuildTransientError, BuildPullError, BuildWithManifestVsDockerfile, imageBuild, BuildkitConnection
  - **Management events (5)**: Namespace, CreateNamespace, DeleteNamespace, Context, Contexts Deletion
  - **Deploy Stack (1)**: Deploy Stack

- Modified origin check in `track()` to consult allowlist before blocking events
- Added comprehensive test in `pkg/analytics/track_test.go` to verify allowlist contents

## How to validate

1. Set the deploy context environment variable:
   ```bash
   export OKTETO_WITHIN_DEPLOY_COMMAND_CONTEXT=true
   ```

2. Run commands that should be tracked (events in allowlist):
   ```bash
   okteto build
   okteto namespace create test-ns
   okteto context use
   ```
   - Verify events appear in Mixpanel with `origin: "okteto-deploy"`

3. Run commands that should NOT be tracked (events not in allowlist):
   ```bash
   okteto up
   okteto down
   ```
   - Verify these events do NOT appear in Mixpanel

4. Run tests to verify implementation:
   ```bash
   make test
   make lint
   ```
   - All tests should pass
   - Linter should report 0 issues

## CLI Quality Reminders 🔧

- [x] Scrutinize for potential regressions - Existing behavior preserved for non-allowlisted events
- [x] Ensure key automated tests are in place - Added `Test_eventsAllowedDuringDeploy`
- [x] Build the CLI and test using the validation steps - All analytics tests pass
- [x] Assess Developer Experience impact (log messages, performances, etc) - No user-facing changes
- [x] If too broad, consider breaking into smaller PRs - Single focused change
- [x] Adhere to our code style and code review guidelines - Linter passes with 0 issues


pkg/analytics/track_test.go
```diff
@@ -133,3 +133,53 @@ func Test_disabledInOktetoCluster(t *testing.T) {
...
+	}
+
+	// Verify Up/Down events are NOT in the allowlist (by design)
+	if eventsAllowedDuringDeploy[upEvent] {
+		t.Error("Up event should not be in allowlist")
+	}
+	if eventsAllowedDuringDeploy[downEvent] {
+		t.Error("Down event should not be in allowlist")
+	}
+}
```
ifbyol: As we agreed, let's use `testify` library for assertions in the tests, please

pkg/analytics/track.go
```diff
@@ -63,6 +63,28 @@ var (
...
+	// Management events
+	namespaceEvent:       true,
+	namespaceCreateEvent: true,
+	namespaceDeleteEvent: true,
+	contextEvent:         true,
+	deleteContexts:       true,
+
+	// Deploy stack
+	deployStackEvent: true,
+}
```
ifbyol: How can we know which events are not being triggered? This generates also some operational overhead as we have to remember to include here or not any new event to be sent or not. It would probably have to be the other way around, by default, all events should be tracked, and we should put in a list the ones we don't want/need. I prefer to send more events than losing them as we have done now with the buildkit nirvana events

I think we should re-think this. Does it make sense that some event is not triggered if the okteto command is being executed as part of a manifest deploy section. Wouldn't that hide some relevant metric/information?

PR #4963, opened 2026-03-31
feat: support dev mode on Argo Rollouts

aaronweisberg:
## Summary
- add dev-mode support for existing Argo Rollouts alongside Deployments and StatefulSets
- add Rollout app/crud integration, ReplicaSet pod resolution, and autodown coverage
- update tests and module deps for the Argo Rollouts typed client

## Testing
- go test ./...

### DIFF
--- a/cmd/up/autodown.go
+++ b/cmd/up/autodown.go
@@ -44,6 +44,7 @@ type autoDownRunner struct {
 	k8sLogger        *io.K8sLogger
 	analyticsTracker analyticsTrackerInterface
 	downCmd          downCmdRunner
+	getApp           func(ctx context.Context, dev *model.Dev, namespace string, c kubernetes.Interface, isRetry bool) (apps.App, bool, error)
 }
 
 type downCmdRunner interface {
@@ -62,6 +63,7 @@ func newAutoDown(ioCtrl *io.Controller, k8sLogger *io.K8sLogger, at analyticsTra
 		k8sLogger:        k8sLogger,
 		analyticsTracker: at,
 		downCmd:          downCmd,
+		getApp:           utils.GetApp,
 	}
 }
 
@@ -75,7 +77,12 @@ func (a *autoDownRunner) run(ctx context.Context, dev *model.Dev, namespace, man
 	sp := a.ioCtrl.Out().Spinner("Running okteto down...")
 	sp.Start()
 	defer sp.Stop()
-	app, _, err := utils.GetApp(ctx, dev, namespace, k8sClient, false)
+	getApp := a.getApp
+	if getApp == nil {
+		getApp = utils.GetApp
+	}
+
+	app, _, err := getApp(ctx, dev, namespace, k8sClient, false)
 	if err != nil {
 		if !oktetoErrors.IsNotFound(err) {
 			return err
--- a/cmd/up/autodown_test.go
+++ b/cmd/up/autodown_test.go
@@ -17,13 +17,19 @@ import (
 	"context"
 	"testing"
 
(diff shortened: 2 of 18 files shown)


codyjlandstrom:
Hey @aaronweisberg
Thanks for putting this together! Argo Rollouts support is something we've been thinking about, so it's great to see additional interest.
Before we dig into your code, I'd love to understand your team's setup a bit better so we can make sure we're solving this the right way.

A few questions:

1. What does your current dev workflow look like with Argo Rollouts today? When a developer on your team needs to iterate on a service that uses a Rollout, what do they actually do right now? (e.g., do they switch the resource to a Deployment for dev, use a separate tool, something else?)
2. What's the deployment strategy you're using? canary, blue-green, or a mix? We want to make sure we understand the edge cases around service annotation ownership and traffic routing, especially since blue-green and canary have pretty different implications for how dev mode needs to handle services.
3. Are you running Argo CD alongside Argo Rollouts, or just Rollouts? This matters for how we think about potential sync conflicts when okteto up modifies the rollout spec.
4. What was the specific pain point that pushed you to build this? I'm curious whether the main friction was around okteto up not recognizing Rollouts at all, or if there were other issues that were also blocking your team.
5. How many services in your stack use Rollouts vs. standard Deployments? Trying to understand whether this is a "one or two critical services" situation or more of a wholesale adoption across your infra.

We have some internal design work in progress on this that covers CLI, backend, and UI together, and there are some tricky edge cases (like what happens when a namespace is slept mid-dev-session with a rollout, or how to handle the managed-by-rollouts annotation on shared services) that we want to make sure we get right. Your context would really help us prioritize and sequence this well.

Thanks again for the contribution!

mdwn:
> Hey @aaronweisberg Thanks for putting this together! Argo Rollouts support is something we've been thinking about, so it's great to see additional interest. Before we dig into your code, I'd love to understand your team's setup a bit better so we can make sure we're solving this the right way.
> 
> A few questions:
> 
> 1. What does your current dev workflow look like with Argo Rollouts today? When a developer on your team needs to iterate on a service that uses a Rollout, what do they actually do right now? (e.g., do they switch the resource to a Deployment for dev, use a separate tool, something else?)
> 2. What's the deployment strategy you're using? canary, blue-green, or a mix? We want to make sure we understand the edge cases around service annotation ownership and traffic routing, especially since blue-green and canary have pretty different implications for how dev mode needs to handle services.
> 3. Are you running Argo CD alongside Argo Rollouts, or just Rollouts? This matters for how we think about potential sync conflicts when okteto up modifies the rollout spec.
> 4. What was the specific pain point that pushed you to build this? I'm curious whether the main friction was around okteto up not recognizing Rollouts at all, or if there were other issues that were also blocking your team.
> 5. How many services in your stack use Rollouts vs. standard Deployments? Trying to understand whether this is a "one or two critical services" situation or more of a wholesale adoption across your infra.
> 
> We have some internal design work in progress on this that covers CLI, backend, and UI together, and there are some tricky edge cases (like what happens when a namespace is slept mid-dev-session with a rollout, or how to handle the managed-by-rollouts annotation on shared services) that we want to make sure we get right. Your context would really help us prioritize and sequence this well.
> 
> Thanks again for the contribution!

I can answer a good bit of this, as I'm @aaronweisberg's coworker.

1. Right now we encourage users to make their dev environments as close to "real life" as possible, so right now they use actual Rollouts.
2. This is still emerging. Right now I would say "a mix" but we're trying to feel out new use cases and methods here.
3. ArgoCD is running alongside Rollouts.
4. The main issue was Okteto not recognizing Rollouts at all.
5. Right now I believe it's around 4, but we're looking to roll this out more widely (dozens of services).

Let me know if you have any other questions!

PR #4981, opened 2026-04-14
feat(build): add env var support for build secrets in Okteto manifest

jLopezbarb:
Fixes DEV-1369 | Part of PROD-461

## Proposed changes

Extends `build.secrets` in the Okteto manifest to support environment variable–based secrets in addition to the existing file-path form.

Previously only file paths were accepted as string values. This PR introduces a `Secret` struct with two mutually-exclusive fields (`file` and `env`), while keeping the shorthand string form backward-compatible:

```yaml
build:
  secrets:
    # Shorthand — backward-compatible, treated as file path
    my_cert: /path/to/cert.pem

    # Long form: explicit file
    server_cert:
      file: /path/to/server.cert

    # Long form: env var — new!
    npm_token:
      env: NPM_TOKEN
```

Validation errors now include a usage hint showing all valid forms. Unknown keys produce a clean user-facing error without leaking internal type names.

## How to validate

```dockerfile
# syntax=docker/dockerfile:1
FROM alpine:3.19
RUN --mount=type=secret,id=npm_token \
    echo "Secret content: $(cat /run/secrets/npm_token)"
```

**1. Shorthand (file path) — backward compat**
```
$ okteto build app -f okteto-shorthand.yml --no-cache --progress=plain
#7 0.159 Secret content: s3cr3t_value
 ✓  Image '...' successfully pushed
```

**2. Long form with `file:` key**
```
$ okteto build app -f okteto-file.yml --no-cache --progress=plain
#7 0.084 Secret content: s3cr3t_value
 ✓  Image '...' successfully pushed
```

**3. Env var secret (new)**
```
$ NPM_TOKEN=s3cr3t_from_env okteto build app -f okteto-env.yml --no-cache --progress=plain
#7 0.142 Secret content: s3cr3t_from_env
 ✓  Image '...' successfully pushed
```

**4. Error: file does not exist**
```
$ okteto build app -f okteto-missing-file.yml --no-cache --progress=plain
 x  Error building service 'app': failed to create build solver: failed to process build secrets: error creating the temp file with expanded values: open /tmp/secrets-test/does-not-exist.txt: no such file or directory
```

**5. Error: both `file` and `env` set**
```
$ okteto build app -f okteto-both-error.yml --no-cache --progress=plain
 x  Your Okteto Manifest is not valid, please check the following errors:
secret cannot specify both 'file' and 'env'

Secrets in your Okteto manifest must be valid file paths or environment variable names. Use one of these formats:

  # Short form (file path)
  secrets:
    my_secret: /path/to/secret/file

  # Explicit file key
  secrets:
    my_secret:
      file: /path/to/secret/file

  # Environment variable as secret value
  secrets:
    my_secret:
      env: MY_ENV_VAR
```

**6. Error: unknown key**
```
$ okteto build app -f okteto-neither-error.yml --no-cache --progress=plain
 x  Your Okteto Manifest is not valid, please check the following errors:
unknown secret key "something_else" — only 'file' and 'env' are allowed

Secrets in your Okteto manifest must be valid file paths or environment variable names. Use one of these formats:

  # Short form (file path)
  secrets:
    my_secret: /path/to/secret/file

  # Explicit file key
  secrets:
    my_secret:
      file: /path/to/secret/file

  # Environment variable as secret value
  secrets:
    my_secret:
      env: MY_ENV_VAR
```


codyjlandstrom:
@ekline-ai Work on the docs update for this

pkg/build/secret.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+	s.File = file
+	s.Env = env
+	return nil
+}
+
+// Secrets represents the secrets to be injected to the build of the image
+type Secrets map[string]Secret
+
+func (i *Info) expandSecrets() error {
```
ifbyol, line 91: Nit: Do we have some preference about having all the functions from a struct in the same file? I mean, `Info` struct I think it is define in `info.go` file. Wouldn't make more sense that struct and its functions are together in the same file?

pkg/build/secret.go
```diff
@@ -0,0 +1,113 @@
...
+		return fmt.Errorf("%w%s", err, secretFormatHint)
+	}
+
+	for k := range raw {
+		if k != "file" && k != "env" {
+			return fmt.Errorf("unknown secret key %q — only 'file' and 'env' are allowed%s", k, secretFormatHint)
+		}
+	}
+
+	file, env := raw["file"], raw["env"]
```
ifbyol, line 76: Nit: `env` is being shadowed, which can lead to issues in the future, we might change it to `envVal`, `envName` or something similar

pkg/build/secret.go
```diff
@@ -0,0 +1,113 @@
...
+	if file != "" && env != "" {
+		return fmt.Errorf("secret cannot specify both 'file' and 'env'%s", secretFormatHint)
+	}
+	if file == "" && env == "" {
```
ifbyol, line 80: Should we strim (I don't remember the name of the function) the values to make sure whitespaces don't pass this validation? I mean, If I specify a string like "    ", it would pass this validation

pkg/build/secret.go
```diff
@@ -0,0 +1,113 @@
+// Copyright 2023 The Okteto Authors
```
ifbyol, line 1:
```suggestion
// Copyright 2026 The Okteto Authors
```

pkg/build/secret.go
```diff
@@ -0,0 +1,113 @@
...
+
+	file, env := raw["file"], raw["env"]
+	if file != "" && env != "" {
+		return fmt.Errorf("secret cannot specify both 'file' and 'env'%s", secretFormatHint)
+	}
+	if file == "" && env == "" {
+		return fmt.Errorf("secret must specify either 'file' or 'env'%s", secretFormatHint)
+	}
+	s.File = file
+	s.Env = env
```
ifbyol, line 84: Should we add some validation to the env variable name? What would happen if this value contains have some values like "A,B", "A B" or things like that? Should we do a minimum validation about valid environment variable names?

ifbyol:
I have a doubt regarding the experience. When using the environment variable, should we validate if that env is set before sending it to BuildKit, and fail if not, or should we just let it through to BuildKit process? If I don't specify the value of the environment variable, an empty value will be sent to BuildKit, and the process might not even fail, but it will end up building an image with an invalid data, and users might take some time to figure it out.

WDYT @jLopezbarb @codyjlandstrom ? How do depot or docker do when that happens?

pkg/build/secret.go
jLopezbarb: I wouldn't do it given that other tools like docker will set it as an empty instead of failing

pkg/build/secret.go
jLopezbarb: For build we are doing it in each file. I don't know why we decided to go that way but check [info.go](https://github.com/okteto/okteto/blob/e8a390abda20f236e5cdc256b238e52729ef5d0d/pkg/build/info.go#L123)

ifbyol: I was checking, and we have all the functions related to `Info` struct within `info.go`, this is the only function "belonging" to `info` which is not in the info.go file

jLopezbarb:
> I have a doubt regarding the experience. When using the environment variable, should we validate if that env is set before sending it to BuildKit, and fail if not, or should we just let it through to BuildKit process? If I don't specify the value of the environment variable, an empty value will be sent to BuildKit, and the process might not even fail, but it will end up building an image with an invalid data, and users might take some time to figure it out.
> 
> WDYT @jLopezbarb @codyjlandstrom ? How do depot or docker do when that happens?

I tested it on several tools and all work in the same way. They accept malformed/empty env vars. I also think it's difficult to debug, but I wouldn't block the execution to match the common behavior. We could maybe show a warning  in order to debug in another PR. I think it's because you can pass an env var to buildkit but then not use it in the `Dockerfile`

ifbyol:
> > I have a doubt regarding the experience. When using the environment variable, should we validate if that env is set before sending it to BuildKit, and fail if not, or should we just let it through to BuildKit process? If I don't specify the value of the environment variable, an empty value will be sent to BuildKit, and the process might not even fail, but it will end up building an image with an invalid data, and users might take some time to figure it out.
> > WDYT @jLopezbarb @codyjlandstrom ? How do depot or docker do when that happens?
> 
> I tested it on several tools and all work in the same way. They accept malformed/empty env vars. I also think it's difficult to debug, but I wouldn't block the execution to match the common behavior. We could maybe show a warning in order to debug in another PR. I think it's because you can pass an env var to buildkit but then not use it in the `Dockerfile`

@codyjlandstrom What do you think about this?

pkg/build/secret.go
```diff
@@ -0,0 +1,113 @@
...
+	return nil
+}
+
+// Secrets represents the secrets to be injected to the build of the image
+type Secrets map[string]Secret
```
ifbyol, line 89: I think this change has an unexpected side effect in Smart Builds that we should review.  In the hasher file for smart builds, we are using the build secrets to build the smart build context. [This line](https://github.com/okteto/okteto/blob/16fcf6e8ae546e466cfc09482761d57fdfdc7bcf/cmd/build/v2/smartbuild/hasher.go#L152) is getting the value of the secret when doing `secrets = append(secrets, fmt.Sprintf("%s=%s", key, value))`. Before this change, `value` was always a string. Now, that value is of type `build.Secret`, and building the string using directly string templating would change the behavior, and it is not correct.

please, review it carefully

ifbyol:
If the value of an environment variable used as secret, Smart Builds cache is not being invalidated. We should invalidate it as we do with the file path secret

codyjlandstrom:
> > > I have a doubt regarding the experience. When using the environment variable, should we validate if that env is set before sending it to BuildKit, and fail if not, or should we just let it through to BuildKit process? If I don't specify the value of the environment variable, an empty value will be sent to BuildKit, and the process might not even fail, but it will end up building an image with an invalid data, and users might take some time to figure it out.
> > > WDYT @jLopezbarb @codyjlandstrom ? How do depot or docker do when that happens?
> > 
> > 
> > I tested it on several tools and all work in the same way. They accept malformed/empty env vars. I also think it's difficult to debug, but I wouldn't block the execution to match the common behavior. We could maybe show a warning in order to debug in another PR. I think it's because you can pass an env var to buildkit but then not use it in the `Dockerfile`
> 
> @codyjlandstrom What do you think about this?

I think having a warning is what we should do for our users. Don't block the build (matches Docker/Depot behavior), but emit a warn: secret 'npm_token' references env var NPM_TOKEN which is not set so users aren't confused

jLopezbarb:
> If the value of an environment variable used as secret, Smart Builds cache is not being invalidated. We should invalidate it as we do with the file path secret

After testing it, if you have the file on another folder and change it smartbuid doesn't invalidate the cache, thats why im not doing it for env vars to keep consistency between both

pkg/build/secret.go
```diff
@@ -0,0 +1,112 @@
...
+// UnmarshalYAML deserializes the secrets map and warns about any env var secrets
+// that are not set, since an empty secret may cause the build to fail.
+func (s *Secrets) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	raw := map[string]Secret{}
+	if err := unmarshal(&raw); err != nil {
+		return err
+	}
+	for name, secret := range raw {
+		if secret.Env != "" && os.Getenv(secret.Env) == "" {
+			oktetoLog.Warning("secret %q uses env var %q which is not set — the build secret will be empty and the build may fail", name, secret.Env)
```
ifbyol: As this is in the unmarshall function, wouldn't this be triggered every time the manifest is parsed, even if it is to execute `okteto test` or another command? Should we just do it on `OptsFromBuildInfo` for example?

I want to avoid something similar to the warning we get when a field is not supported on hybrid, which is displayed for any command

ifbyol:
From an implementation perspective it looks fine. @codyjlandstrom How do you see the user-facing messages? Do you think the error one is long or is it good?


Screenshot 2026-04-27 at 12 02 04

Screenshot 2026-04-27 at 12 06 58

codyjlandstrom:
> From an implementation perspective it looks fine. @codyjlandstrom How do you see the user-facing messages? Do you think the error one is long or is it good?

1. The content is good, but I think it feels just a bit too long for me. We could send them to the docs, but this is super clear. I'd say keep it.
2. I'd replace the em dash with a colon as that will render more consistently in the terminal. Otherwise looks good
> secret "npm_token" uses env var "NPM_TOKEN" which is not set: the build secret will be empty and the build may fail

ifbyol (approved):
Implementation looks fine, and there is only one minor thing related to the warning message to change, but approving it

PR #4997, opened 2026-04-22
fix(manifest): return compose error when filename matches compose pattern

jLopezbarb:
## Summary

When a file with a compose-style name (e.g. `docker-compose.yml`, `compose.yaml`, `okteto-stack.yml`) failed to parse as both an okteto manifest and a compose/stack file, the user received the okteto manifest error — the less relevant of the two. Users pointing at a compose file expect compose errors.

The fix checks whether the filename matches a known compose pattern via a new `discovery.IsComposeFilename` helper and, if so, returns the compose error instead.

## What changed

- **`pkg/discovery/compose.go`** — adds `IsComposeFilename(path string) bool`, which checks the base filename against the existing `possibleComposeManifests` list (the same list used by `GetComposePath`).
- **`pkg/model/manifest.go`** — in `getManifestFromFile`, when both okteto manifest and compose loading fail, returns the compose error when `IsComposeFilename` matches.
- **`pkg/discovery/compose_test.go`** — 15 cases covering all known compose filenames, full paths, `.okteto/` subdirectory paths, and non-compose names.

---

[Compound Engineering](https://github.com/EveryInc/compound-engineering-plugin)
Sonnet_4.6_(200K)-D97757?logo=claude&logoColor=white)

### DIFF
--- a/pkg/discovery/compose.go
+++ b/pkg/discovery/compose.go
@@ -47,6 +47,17 @@ var (
 	}
 )
 
+// IsComposeFilename reports whether the base name of path matches a known compose file name.
+func IsComposeFilename(path string) bool {
+	base := filepath.Base(path)
+	for _, parts := range possibleComposeManifests {
+		if parts[len(parts)-1] == base {
+			return true
+		}
+	}
+	return false
+}
+
 // GetComposePath returns a compose file if exists, error otherwise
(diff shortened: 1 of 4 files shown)


ifbyol (approved):
I was testing it and it looks good. I would add the test scenario mentioned by Copilot, but everything else looks good

PR #5002, opened 2026-04-27
docs(agents): add feature flag NoOp strategy guidelines

jLopezbarb:
## Summary

Add agent documentation for the team's agreed feature flag convention: new flags must always default to `false`, keeping the new code path as a NoOp until the implementation is explicitly promoted to production.

- New `.claude/context/feature-flags.md` covers the NoOp strategy rationale, rollout lifecycle, implementation pattern, helper function reference, naming conventions, a real codebase example, and a pre-merge checklist
- `.claude/CLAUDE.md` updated with a rule in the ALWAYS section and a link in Context References so agents load the convention automatically

## Test plan

- [ ] Verify `.claude/context/feature-flags.md` renders correctly on GitHub
- [ ] Confirm CLAUDE.md links resolve to the new file

---


.claude/context/feature-flags.md
```diff
@@ -0,0 +1,117 @@
+# Feature Flags
+
+> **See also**: [conventions.md](conventions.md) for general Go conventions and [architecture.md](architecture.md) for package structure.
+
+## The NoOp Strategy (Required)
+
+**Always default new feature flags to `false`.** The new code path must be a no-operation (NoOp) — preserving existing behavior — until the implementation is proven and ready.
```
ifbyol: The default value should depend on the meaning of the feature, and it is not always a boolean, for example, the gateway type feature flag has another possible values.

I would remove the default value and reinforce the idea of preserving the existing behavior

.claude/context/feature-flags.md
```diff
@@ -0,0 +1,117 @@
...
+
+### Why
+
+New features are developed incrementally. Defaulting to `false` ensures the new code path never activates for users until it has been explicitly validated and promoted. This makes feature flags safe to merge at any stage of development.
```
ifbyol: Same about default value.

ifbyol (commented):
I think we are mixing 2 things that, even they are related, they don't always happen together. I will try to explain myself

One thing is the feature flag itself. That is to split the flow in, at least, 2 different code paths. And ideally, we should follow an strategy pattern or similar to prevent to have the same check in several places, and create an struct or similar per strategy, so we create them once based on the flag.

The other thing is the NoOp strategy. That is to simplify the code when a specific behavior is only enabled when a feature flag has some specific value, but in any other case, it should do nothing. So, instead of doing something like this:

```
if featureFlagEnabled {
    callEnabledBehavior()
}
```

You would have this:

```
func createStrategy() strategyInterface {
    if featureFlagEnabled {
        return newBehavior{}
     }
     return noOpBehavior{}
}

....


strategy := createStrategy()
.....
strategy.Behavior() // internally each struct would define its own strategy
```

They don't happen always, but they are related.

So, I would change a bit the context for the feature flag and how it is framed.

Also, this is assuming that a feature flag has always boolean values, and its default is always `false`, but it is not the case. There might be cases where the default should be `true`, and they are not always boolean values

ifbyol (approved):
Now it looks good. Be aware that the diff shows a bigger diff than it should be, but the content of the no op and feature flag context is good!

PR #5016, opened 2026-05-11
feat(analytics): add PostHog analytics backend with image_build event

jLopezbarb:
# Proposed changes

Fixes # (issue)

This PR introduces a PostHog analytics backend alongside the existing Mixpanel integration, using a fan-out backend pattern. The first event covered is `image_build`, which is enriched with PostHog-specific properties including build context size, buildkit duration, context transfer duration, connection type, and error reason.

Key changes:

- **Backend abstraction**: introduces `analyticsBackend` interface and a `Tracker` that fans out to all registered backends (`mixpanelBackend` + `posthogBackend`). New events only need to be added to the interface and implemented per backend.
- **PostHog backend** (`pkg/analytics/posthog.go`): sends `image_build` events with common CLI properties (OS, arch, CLI version, machine ID, cluster ID, customer ID, agent type detection for Codex/Gemini/Claude/Cursor) and group membership. Routes through the Okteto-owned reverse proxy at `https://ph.okteto.com`.
- **Group identify**: `posthogBackend.IdentifyGroups()` sends `$groupidentify` calls for `cluster` and `customer` groups. `Tracker.IdentifyGroups()` is called from `cmd/context` after the okteto context is fully populated (ClusterID + CompanyName set).
- **ClusterInfo endpoint**: adds `GetClusterInfo()` REST call to the user client, fetching `ClusterVersion` and `ClusterID` from `/clusterinfo`. These fields are stored on the okteto context and used as PostHog common properties.
- **Build metadata propagation**: threads `BuildContextSize`, `BuildkitDuration`, `ContextTransferDuration`, `ConnectionType`, and `ErrorReason` from the buildkit solver through `BuildMetadata` up to `ImageBuildMetadata` so PostHog receives accurate per-build measurements.
- **Graceful shutdown**: `at.Close()` is called in `main()` before exit to flush pending PostHog events.
- **Connector tracker injection**: `GetBuildkitConnector` now accepts a `BuildkitConnectionTracker` interface so the analytics tracker is threaded in from call sites rather than constructed internally, eliminating the singleton dependency.

## How to validate

1. Build the CLI locally:
   ```
   make build
   ```
2. Set up a context pointing to an Okteto cluster:
   ```
   ./okteto context use https://<your-okteto-cluster>
   ```
3. Run an image build and observe that it completes without errors:
   ```
   ./okteto build
   ```
4. Verify analytics are dispatched by temporarily setting `OKTETO_ANALYTICS=true` (default) and checking the PostHog dashboard at `https://ph.okteto.com` for an `image_build` event with properties `service`, `duration_seconds`, `connection_type`, `cluster_id`, `customer_id`.
5. Run the unit tests to verify all backends and property mappings:
   ```
   go test ./pkg/analytics/... ./pkg/build/buildkit/... ./pkg/okteto/... ./cmd/context/... ./cmd/build/... -count=1
   ```
6. Run the linter:
   ```
   make lint
   ```

## CLI Quality Reminders 🔧


cmd/build/v2/build.go
```diff
@@ -290,16 +290,29 @@ func (ob *OktetoBuilder) Build(ctx context.Context, options *types.BuildOptions)
...
 			}
 		}
 		svcToBuild.SetBuildDuration(buildDuration, waitForBuildkitAvailable, err == nil)
+		svcToBuild.Metadata().BuildContextSize = buildContextSize
+		svcToBuild.Metadata().ConnectionType = connectionType
+		svcToBuild.Metadata().BuildkitDuration = buildkitDuration
+		svcToBuild.Metadata().ContextTransferDuration = contextTransferDuration
 
 		if err != nil {
+			svcToBuild.Metadata().ErrorReason = err.Error()
```
ifbyol, line 315: Nit: Agree with Copilot message. Is it expected that we send any error value here?

cmd/context/context.go
```diff
@@ -21,7 +21,7 @@ import (
 )
 
 // Context points okteto to a cluster.
-func Context() *cobra.Command {
+func Context(at groupsIdentifier) *cobra.Command {
```
ifbyol, line 24: Nit: What does `at` mean? Is it something random or does it have some meaning? if not, we might rename it to something more meaningful?

cmd/build/v2/build.go
```diff
@@ -290,16 +290,29 @@ func (ob *OktetoBuilder) Build(ctx context.Context, options *types.BuildOptions)
...
 		waitForBuildkitAvailable := time.Duration(0)
+		buildContextSize := int64(0)
+		connectionType := ""
+		buildkitDuration := time.Duration(0)
+		contextTransferDuration := time.Duration(0)
 		buildkitRunner, ok := ob.Builder.BuildRunner.(*buildCmd.OktetoBuilder)
 		if ok {
 			buildkitMetadata := buildkitRunner.GetMetadata()
 			if buildkitMetadata != nil {
 				waitForBuildkitAvailable = buildkitMetadata.WaitForBuildkitAvailableTime
```
ifbyol, line 301: I was checking, and I think we have an error here from a previous PR/change (not related to this), but I don't see where `WaitForBuildkitAvailableTime` within BuildKitMetadata is being set, so I think `waitForBuildkitAvailable` is always 0. Could you check it?

pkg/analytics/imagebuild.go
```diff
@@ -66,6 +71,44 @@ func (m *ImageBuildMetadata) toProps() map[string]interface{} {
...
+		"service":  m.Name,
+		"result":   m.Success,
+		"is_cache": m.CacheHit,
+	}
+	if d := int(m.BuildDuration.Seconds()); d > 0 {
+		props["duration_seconds"] = d
+	}
+	if d := int(m.WaitForBuildkitAvailable.Seconds()); d > 0 {
+		props["queue_duration_seconds"] = d
+	}
```
ifbyol, line 87: Related to a previous comment, I think this is always 0 as a result of a previous bug in the implementation

pkg/analytics/posthog_test.go
```diff
@@ -0,0 +1,252 @@
...
+		currentAnalytics = prevAnalytics
+		okteto.CurrentStore = prevStore
+		config.VersionString = prevVersion
+	}
+}
+
+func TestPostHogBackend_TrackImageBuild_NilClient(t *testing.T) {
+	b := &posthogBackend{client: nil}
+	// Must not panic — nil client is silently ignored
+	b.TrackImageBuild(context.Background(), &ImageBuildMetadata{Success: true})
```
ifbyol: Nit: Should we use `require.NoPanic` as in other tests

pkg/build/buildkit/runner_test.go
```diff
@@ -461,6 +463,26 @@ func TestNewBuildkitRunner(t *testing.T) {
...
+			return nil
+		},
+		metadata:               &runnerMetadata{build: &BuildMetadata{BuildContextSize: 1024}},
+		logger:                 io.NewIOController(),
+		okCtx:                  &fakeOktetoContext{},
+		fs:                     afero.NewMemMapFs(),
+		solveOptBuilderFactory: newFakeSolveOptBuilderFactory(&client.SolveOpt{}, nil),
+	}
+
+	assert.Equal(t, int64(1024), r.GetMetadata().BuildContextSize)
```
ifbyol: Nit:
```suggestion
	require.Equal(t, int64(1024), r.GetMetadata().BuildContextSize)
```

pkg/okteto/cluster_info_test.go
```diff
@@ -0,0 +1,127 @@
...
+				}),
+			}
+
+			c := &userClient{
+				httpClient: httpClient,
+				baseURL:    "https://okteto.example.com",
+			}
+
+			got, err := c.GetClusterInfo(context.Background())
+			if tt.expectedErr != "" {
```
ifbyol: Nit: Ideally, we should not use conditionals within tests

main.go
```diff
@@ -181,6 +181,7 @@ func main() {
 	root.AddCommand(pipeline.Pipeline(ctx))
 
 	err = root.Execute()
+	at.Close()
```
ifbyol, line 184: Would it make sense to call it within a `defer`? The main difference is that if a panic happens, we would have the events collected until the panic

ifbyol (approved):
I was testing it, and everything seems to be fine functionality wise. I defer to product team the validation of the information we send, name of paremeters, groups and all those things.

My comments are mainly nits, the only relevant comes from a previous bug where some data is not being set, which if it is something easy to fix, I'd as part of this PR

main.go
jLopezbarb: No, we can't because if there is an os.EXit and if there is an os call it will execute before the defer call so it won't be running

cmd/context/context.go
jLopezbarb: analyticsTracker. If it's not clear I can rename it

cmd/build/v2/build.go
jLopezbarb: We (product <> dev) agreed to send this event first and create the categories based on the errors that we send

PR #5029, opened 2026-05-26
feat(analytics): add okteto_up_started event and wire up PostHog backend

jLopezbarb:
# Proposed changes

Fixes # (issue)

This PR extends the PostHog analytics infrastructure introduced in #5016 (which covered `image_build` and the fan-out backend architecture) to add full PostHog coverage for the `up` command.

**`okteto_up_started` event** — fires after the dev name is resolved in `okteto up`, before build/start, so every attempt is captured including early failures. Records `service`, `namespace`, and `repo_url`.

**`up` event** — end-of-session event now dispatched to PostHog. Properties: `result`, `manifest_type`, `service`, `namespace`, `repo_url`, `is_interactive`, `is_build_executed`, `is_deploy_executed`, `has_build_section`, `has_deploy_section`, `is_reconnect`, `reconnect_count`, `is_auto_down`, duration fields, `reconnect_cause`, `error_reason`.

**`is_build_executed`** — set only after `Build()` returns successfully (was set before the call in the original code).

**Common properties** — `trigger_source` (`cli`/`github_action`/`gitlab_ci`) restored; `customer_name` and `cluster_name` added as human-readable aliases alongside `customer_id`/`cluster_id`.

## How to validate

1. `make build`
2. Run `okteto up` with an invalid service → confirm `okteto_up_started` still fires
3. `go test ./pkg/analytics/... ./cmd/up/... -count=1`
4. `make lint`

## CLI Quality Reminders 🔧


cmd/up/build.go
```diff
@@ -80,7 +83,13 @@ func (ub *upBuilder) build(ctx context.Context) error {
 		CommandArgs: svcsToBuild,
 		Manifest:    ub.manifest,
 	}
-	return ub.builder.Build(ctx, buildOptions)
+	if err := ub.builder.Build(ctx, buildOptions); err != nil {
+		return err
+	}
+	if ub.analyticsMeta != nil {
+		ub.analyticsMeta.HasRunBuild()
+	}
```
ifbyol: Open question: Why do we track only if it was successful? Shouldn't we track somehow that the okteto up failed due to an error building?

cmd/up/build.go
```diff
@@ -80,7 +83,13 @@ func (ub *upBuilder) build(ctx context.Context) error {
 		CommandArgs: svcsToBuild,
 		Manifest:    ub.manifest,
 	}
-	return ub.builder.Build(ctx, buildOptions)
+	if err := ub.builder.Build(ctx, buildOptions); err != nil {
+		return err
+	}
+	if ub.analyticsMeta != nil {
```
ifbyol: Why do we check if it is `nil`? In which cases this might be `nil`? Would it make sense to not make it a pointer to prevent this kind of situations? We don't tend to pre-validate struct properties as the approach is to make sure it is always set.

cmd/up/up.go
```diff
@@ -491,6 +498,12 @@ func (up *upContext) start() error {
 	up.analyticsMeta.ManifestProps(up.Manifest)
 	up.analyticsMeta.DevProps(up.Dev)
 	up.analyticsMeta.RepositoryProps(utils.IsOktetoRepo())
+	up.analyticsMeta.SetNamespace(up.Namespace)
+	repoURL, err := modelutils.GetRepositoryURL(up.Manifest.ManifestPath)
```
ifbyol: `modelutils.GetRepositoryURL` is already called earlier in the flow. Could we share it somehow, so it is not called twice?

pkg/analytics/posthog.go
```diff
@@ -34,6 +34,8 @@ const (
 	posthogEndpoint = "https://ph.okteto.com"
 
 	posthogImageBuildEvent = "image_build"
+	posthogUpEvent         = "up"
+	posthogUpStartedEvent  = "okteto_up_started"
```
ifbyol: Would it make sense to standarize these events somehow? I think they should be `okteto_up` and `okteto_up_started`, or `up` and `up_started`, but not a mix of them (I mean to always include `okteto` prefix or not). I don't know how they are defined in the document

pkg/analytics/posthog.go
```diff
@@ -77,8 +79,10 @@ func commonPostHogProperties() posthog.Properties {
 	agent := getAgent()
 	props := posthog.Properties{
 		// Common (all PostHog sources)
-		"customer_id":     ctx.CompanyName,
-		"cluster_id":      ctx.ClusterID,
+		"customer_id":   ctx.CompanyName,
```
ifbyol: Something to clarify with product and we discussed offline. I though we were not gonna include`customer_id` as it is the same than `customer_name`. In other events like `image_build` or `deploy_click` we are not including it.

Same for `cluster_name` vs `cluster_id`

pkg/analytics/tracker.go
```diff
@@ -13,12 +13,59 @@
...
+type closer interface {
+	Close()
+}
+
+// Tracker dispatches analytics events to all registered backends.
 type Tracker struct {
-	trackFn func(event string, success bool, props map[string]interface{})
+	trackFn  func(event string, success bool, props map[string]any)
+	backends []analyticsBackend
 }
```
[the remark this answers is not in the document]
ifbyol: @jLopezbarb Could you review this comment?

pkg/build/buildkit/runner.go
```diff
@@ -184,6 +184,7 @@ func (r *Runner) Run(ctx context.Context, buildOptions *types.BuildOptions, outp
 		}
 
 		// if buildkit is not available for 10 minutes, we should fail
+		waitStart = time.Now()
```
ifbyol: Why this? It seems unrelated to the changes targeted for this PR

pkg/analytics/posthog.go
```diff
@@ -168,6 +170,89 @@ func (b *posthogBackend) TrackImageBuild(ctx context.Context, m *ImageBuildMetad
...
+	maps.Copy(props, m.toPostHogProps())
+	namespace := m.namespace
+
+	b.wg.Add(1)
+	go func() {
+		defer b.wg.Done()
+		if b.nsResolver != nil && namespace != "" {
+			fetchCtx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+			defer cancel()
+			if uid, err := b.nsResolver.GetNamespaceUID(fetchCtx, namespace); err != nil {
```
ifbyol: Same comment I have about the build event in the base branch of this PR, take a look at it please

pkg/analytics/posthog.go
```diff
@@ -168,6 +170,89 @@ func (b *posthogBackend) TrackImageBuild(ctx context.Context, m *ImageBuildMetad
...
+		props["up_workflow_id"] = workflowID
+	}
+
+	b.wg.Add(1)
+	go func() {
+		defer b.wg.Done()
+		if b.nsResolver != nil && namespace != "" {
+			fetchCtx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+			defer cancel()
+			if uid, err := b.nsResolver.GetNamespaceUID(fetchCtx, namespace); err != nil {
```
ifbyol: Same about sharing somehow this piece of code on how we resolve the namespace UUID

cmd/up/build.go
jLopezbarb: The intent is that `is_build_executed` signals a build that ran and completed successfully. If the build fails, the `up` event already captures `result: error` and `error_reason`. Setting `is_build_executed: true` on failure would conflate "build failed" with "no build needed" (both would show `false`). Happy to add a separate `is_build_failed` flag if product wants to distinguish failed builds in the PostHog dashboard.

cmd/up/build.go
jLopezbarb: Removed. `upMeta` is always initialized with `analytics.NewUpMetricsMetadata()` before `newUpBuilder` is called, so the nil check was unnecessary. Updated `TestUpBuilder_Build` to always set `analyticsMeta`, following the pattern of always guaranteeing the field is set (commit 15894e6a).

cmd/up/up.go
jLopezbarb: `GetRepositoryURL` is called once (line 286) and the result is shared: passed to `TrackUpStarted` and stored via `upMeta.SetRepoURL`. This was fixed in commit `f75cfbdb` — the duplicate call was removed before this review.

pkg/analytics/posthog.go
jLopezbarb: The `okteto_` prefix was intentionally dropped (commit `d735ff1c`). Current event names are `image_build`, `up`, and `up_started` — consistent across all events (none have the prefix). If product wants to revisit the naming convention, happy to align, but it should be a consistent change across all events.

pkg/analytics/posthog.go
jLopezbarb: `customer_id` is not sent from the CLI — the comment in `commonPostHogProperties` notes it's derived server-side in PostHog. Only `customer_name` is sent as an explicit property. For cluster, we send `cluster_id` (UUID from ClusterInfo) and `cluster_url` (the context name). If the spec says differently, let me know which fields to drop.

pkg/analytics/tracker.go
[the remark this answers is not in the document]
jLopezbarb: In Go 1.18+, `any` is an alias for `interface{}` — `map[string]any` and `map[string]interface{}` are the same type. There is no compile-time mismatch. Confirmed: `go build ./...` passes without errors.

pkg/build/buildkit/runner.go
jLopezbarb: This change (`a7197dc0`) came from the `jlo/followup-build-event` base branch, which was merged into this PR. It removes an unnecessary `waitStart` assignment in the runner's `Run` method — minor housekeeping from the same work stream. Happy to revert it here if you prefer to keep this PR focused on the up event work.

pkg/analytics/posthog.go
jLopezbarb: Extracted into a `resolveNamespaceUID` helper shared by all three Track methods (commit 15894e6a). `TrackUpStarted` retains the fallback to the raw namespace name when the resolver is unavailable, which was the existing behavior.

pkg/analytics/posthog.go
jLopezbarb: Done — same commit 15894e6a.

cmd/up/build.go
ifbyol: > The intent is that is_build_executed signals a build that ran and completed successfully

According to the document, `is_build_executed` indicates `Runtime flag: did up trigger a build`. It means `triggers` not `succeed`. If not, I guess it would be called `is_build_succeed`?

For me, it is odd to see that an okteto up failed, with an error_reason indicating that the build failed, but with the property `is_build_executed` to `false`. Why the build failed if it wasn't executed?

> Setting is_build_executed: true on failure would conflate "build failed" with "no build needed" (both would show false)

I don't understand this. if `is_build_executed: true`, it cannot be also `false` in the case of `build failed`. 

> Happy to add a separate is_build_failed flag if product wants to distinguish failed builds in the PostHog dashboard.

Failed build should be obtained from the `error_reason` property, no need of another flag.

pkg/analytics/posthog.go
```diff
@@ -190,6 +192,43 @@ func (b *posthogBackend) TrackImageBuild(ctx context.Context, m *ImageBuildMetad
...
+	if b.client == nil {
+		return
+	}
+	if !analyticsEnabled() {
+		return
+	}
+	userID := okteto.GetContext().UserID
+	props := commonPostHogProperties()
+	maps.Copy(props, m.toPostHogProps())
+	b.enqueue(context.Background(), userID, posthogUpEvent, props)
```
ifbyol: Don't we need to pass the enricher `withNamespace` here?

pkg/analytics/up.go
```diff
@@ -197,7 +235,80 @@ func (u *UpMetricsMetadata) IsAutoDownEnabled(enabled bool) {
...
+		"is_deploy_executed": u.hasRunDeploy,
+		"has_build_section":  u.hasBuildSection,
+		"has_deploy_section": u.hasDeploySection,
+		"is_reconnect":       u.isReconnect,
+		"reconnect_count":    u.reconnectCount,
+		"is_auto_down":       u.isAutoDownEnabled,
+	}
+	if u.workflowID != "" {
+		props["up_workflow_id"] = u.workflowID
+	}
```
ifbyol: Don't we have to generate a random ID if there is no workflow id? If not, should at least send it empty?

pkg/analytics/up.go
```diff
@@ -197,7 +235,80 @@ func (u *UpMetricsMetadata) IsAutoDownEnabled(enabled bool) {
...
+	if u.service != "" {
+		props["service"] = u.service
+	}
+	if u.namespace != "" {
+		props["namespace"] = u.namespace
+	}
```
ifbyol: I think having this piece of code is dangerous. If we forget to add the `withNamespace` enricher (as it happens currently), we might be leaking the namespace name to the metric

pkg/analytics/posthog.go
ifbyol: Tested, this is not being applied

Image

jLopezbarb:
In the latest commit, the fields sent in the `up` and `up_started` events have been standardized: all fields (`service`, `repo_url`, `workflow_id`, durations, etc.) are now always included even when empty/zero, as agreed with product. The only two that remain conditional are `error_reason` (only when `result: false`) and `reconnect_cause` (only when `is_reconnect: true`), since sending them empty in the opposite cases adds no value.

ifbyol (approved):

PR #5030, opened 2026-05-26
feat(analytics): add deploy_completed PostHog event

jLopezbarb:
## Proposed changes

Adds the `deploy_completed` PostHog event to the CLI analytics backend.

This is part of the PostHog Phase 2 instrumentation (1.45 targets). The event fires on every `okteto deploy` invocation (success or failure) and covers UI clicks, CLI invocations, and CI sources.

**Properties sent:**

| Property | Notes |
|---|---|
| `result` | bool — success or failure |
| `deploy_type` | manifest archetype (`compose`, `okteto`, `manifest`, `pipeline`), defaults to `pipeline` |
| `is_preview` | bool |
| `is_redeploy` | bool |
| `has_dependencies_section` | config flag |
| `has_build_section` | config flag |
| `is_remote` | bool |
| `namespace` | conditional — hashed at ingestion via Hog transformation |
| `duration_seconds` | conditional — omitted if zero |
| `wait_for_deploy` | conditional — omitted if false |
| `error_reason` | conditional — normalized error category on failure (e.g. `timeout`, `command_failed`) |

Plus all common PostHog properties (via `commonPostHogProperties`) and `$groups`.

Depends on #5029.

## How to validate

1. Run `okteto deploy` against a real cluster
2. Confirm `deploy_completed` event appears in PostHog with the expected properties
3. Run unit tests: `go test ./pkg/analytics/...`

## CLI Quality Reminders 🔧


pkg/analytics/deploy.go
```diff
@@ -19,20 +19,55 @@ import (
...
+		"is_redeploy":              d.IsRedeploy,
+		"has_dependencies_section": d.HasDependenciesSection,
+		"has_build_section":        d.HasBuildSection,
+		"is_remote":                d.IsRemote,
+	}
+	if d.Namespace != "" {
+		props["namespace"] = d.Namespace
+	}
+	if d.WaitForDependencies {
+		props["wait_for_dependencies"] = true
```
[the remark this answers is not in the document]
ifbyol: @jLopezbarb This is very relevant, I think we have an error on the wait_for_dependencies flag in metrics, as we are passing the value of `--wait` flag which is not the same

cmd/deploy/deploy.go
```diff
@@ -281,9 +281,11 @@ $ okteto deploy --no-build=true`,
 				if options.Namespace == "" {
 					options.Namespace = okteto.GetContext().Namespace
 				}
+				k8sClient, _, provideErr := c.K8sClientProvider.ProvideWithLogger(okteto.GetContext().Cfg, c.K8sLogger)
+				isRedeploy := provideErr == nil && pipeline.IsDeployed(ctx, options.Name, options.Namespace, k8sClient)
 				err := c.Run(ctx, options)
```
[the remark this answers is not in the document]
ifbyol: @jLopezbarb This is relevant, this is not working correctly because the name is not set yet when the `IsDeployed` function is called

ifbyol (changes requested):
Regarding the `deploy_type`. In the PR description it says that the possible values are `compose`, `v2` and `pipeline`, but If I'm not wrong, that property map to the type `model.Archetype` which doesn't have the `v2` value, it has `manifest` or `okteto` instead. Could you review it?

pkg/analytics/deploy.go
```diff
@@ -19,20 +19,55 @@ import (
...
+	}
+	if secs := int(d.Duration.Seconds()); secs > 0 {
+		props["duration_seconds"] = secs
+	}
+	if !d.Success && d.Err != nil {
+		props["error_reason"] = d.Err.Error()
```
ifbyol: Could this error contain somehow PII like, `not able to access to image <image-name>` or `failed to read file /Users/<name>/.config/okteto/creds.yml`

jLopezbarb:
Fixed in the PR description — the correct values from `model.Archetype` are `compose`, `okteto`, `manifest`, and `pipeline`. `v2` was a typo. Also updated `error_reason` row to reflect the normalized categories introduced in the latest commit.

pkg/analytics/deploy.go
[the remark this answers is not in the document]
jLopezbarb: Fixed — `WaitForDependencies` now iterates `manifest.Dependencies` and is `true` only when at least one dependency has `wait: true` in the manifest. The `--wait` CLI flag is no longer involved.

cmd/deploy/deploy.go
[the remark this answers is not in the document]
jLopezbarb: Fixed — `isRedeploy` is now computed inside `Run()` immediately after `setDeployOptionsValuesFromManifest()` resolves the name, but before `TranslateConfigMapAndDeploy()` mutates state. Stored on `dc.isRedeploy` and read by `TrackDeploy()`, following the same pattern as `dc.PipelineType`.

pkg/analytics/deploy.go
jLopezbarb: Fixed — `error_reason` now uses normalized categories instead of the raw error string. Known errors map to codes like `timeout`, `command_failed`, `internal_server_error`, `user_error`, `no_deploy_commands`. Any unrecognized error omits the field entirely, so no PII reaches PostHog.

pkg/analytics/deploy.go
```diff
@@ -14,25 +14,141 @@
...
+	RepoURL           string
+	WorkflowID        string
+	ParentExecutionID string
+	IsPreview         bool
+	IsRedeploy        bool
+}
+
+func (d *DeployStartedMetadata) toPostHogProps() map[string]any {
+	props := map[string]any{
+		"is_within_preview": d.IsPreview,
```
ifbyol, line 132: I think this is wrong. `d.IsPreview` is initialized in some places with the function `isPreviewEnvironment`, which only checks a couple of environment variables, but those environment variables are not set if I execute `okteto ns use <preview> && okteto deploy`. This should be considered `is_within_preview` as `true` but it would be considered as false. Could you verify it, please?

cmd/preview/deploy.go
```diff
@@ -116,9 +116,14 @@ func (pw *Command) ExecuteDeployPreview(ctx context.Context, opts *DeployOptions
...
-			RepoURL:    opts.repository,
-			Preview:    opts.name,
+			WorkflowID:       opts.workflowID,
+			ParentWorkflowID: os.Getenv(constants.OktetoParentWorkflowIDEnvVar),
+			RepoURL:          opts.repository,
+			Preview:          opts.name,
+			IsWithinPreview: analytics.IsWithinPreview(ctx, func(ctx context.Context, ns string) error {
+				_, err := pw.okClient.Previews().Get(ctx, ns)
+				return err
+			}),
```
ifbyol, line 126: This doesn't make sense. As we discussed, for the preview event we should not include the `is_within_preview`. Review everything related to it

pkg/analytics/preview_triggered.go
```diff
@@ -17,11 +17,13 @@ import "context"
...
-	RepoURL    string
-	Preview    string
-	UIElement  string
-	IsRedeploy bool
+	WorkflowID       string
+	ParentWorkflowID string
+	RepoURL          string
+	Preview          string
+	UIElement        string
+	IsWithinPreview  bool
```
ifbyol, line 25: As discussed , this can be deleted

ifbyol (approved):

PR #5042, opened 2026-06-05
feat(analytics): harden PostHog image_build event and simplify group analytics

jLopezbarb:
# Proposed changes

Fixes # (issue)

This PR makes several improvements to the PostHog analytics pipeline, focused on the `image_build` event and common properties:

- **PII protection**: `repo_url` is now SHA-256-hashed before being sent to PostHog, preventing accidental leakage of private repository URLs.
- **Namespace UID enrichment**: `TrackImageBuild` resolves the Kubernetes namespace UID (via a new `UIDResolver` with a `sync.Map` cache) and attaches it as the `namespace` property. Resolution happens in a background goroutine with a 5-second timeout so the caller is never blocked; the event is sent regardless of whether resolution succeeds.
- **Remove group analytics from the CLI**: `IdentifyGroups` / `$groupidentify` calls are removed from `posthogBackend`, `Tracker`, and `cmd/context`. Group analytics will be computed server-side, eliminating a round-trip on every context switch.
- **Richer common properties**: `cluster_url` (the full context URL, formerly `cluster_name`) and `trigger_source: "cli"` are added to every PostHog event. A comment clarifies that `customer_id` is generated server-side.
- **Rename `cluster_name` → `cluster_url`**: `ctx.Name` holds a URL, so the field name is corrected.

## How to validate

1. Build the CLI from this branch and set up a working Okteto context:
   ```
   make build
   ./bin/okteto context use <your-okteto-url>
   ```
2. Run a build and confirm no `$groupidentify` events are fired (check PostHog debugger or add a temporary log to `Enqueue`):
   ```
   ./bin/okteto build
   ```
3. Inspect the resulting `image_build` PostHog event and verify:
   - `repo_url` is a 64-character hex string (SHA-256), not a raw URL.
   - `namespace` contains a UUID matching the Kubernetes namespace UID (`kubectl get ns <your-ns> -o jsonpath='{.metadata.uid}'`).
   - `cluster_url` contains the Okteto context URL (not a display name).
   - `trigger_source` equals `"cli"`.
4. Run unit tests to confirm coverage of the new resolver and updated posthog backend:
   ```
   go test ./pkg/analytics/... ./pkg/k8s/namespaces/... ./cmd/context/...
   ```
5. Confirm the context command works without the removed `groupsIdentifier` argument:
   ```
   ./bin/okteto context use <your-okteto-url>
   ./bin/okteto context list
   ```

## CLI Quality Reminders 🔧


pkg/k8s/namespaces/uid_resolver.go
```diff
@@ -0,0 +1,62 @@
...
+	"context"
+	"fmt"
+	"sync"
+
+	"github.com/okteto/okteto/pkg/okteto"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// UIDResolver fetches Kubernetes namespace UIDs and caches results by namespace name.
+type UIDResolver struct {
```
ifbyol, line 26: Shouldn't this be "uuid"? (apply to all the file)

pkg/analytics/posthog.go
```diff
@@ -112,82 +130,51 @@ func getAgent() string {
...
+	namespace := m.Namespace
+
+	b.wg.Add(1)
+	go func() {
+		defer b.wg.Done()
+		if b.nsResolver != nil && namespace != "" {
+			fetchCtx, cancel := context.WithTimeout(ctx, 5*time.Second)
+			defer cancel()
+			if uid, err := b.nsResolver.GetNamespaceUID(fetchCtx, namespace); err != nil {
+				oktetoLog.Infof("analytics: failed to get namespace UID: %s", err)
```
ifbyol: If there is an error, should we include an empty value for namespace property? It might look odd that there could be events without namespace set

pkg/analytics/posthog.go
```diff
@@ -112,82 +130,51 @@ func getAgent() string {
...
-	}
+	namespace := m.Namespace
+
+	b.wg.Add(1)
+	go func() {
+		defer b.wg.Done()
+		if b.nsResolver != nil && namespace != "" {
+			fetchCtx, cancel := context.WithTimeout(ctx, 5*time.Second)
+			defer cancel()
+			if uid, err := b.nsResolver.GetNamespaceUID(fetchCtx, namespace); err != nil {
```
ifbyol: I'm not a big fan of having this here. I'm not sure this place should be the responsible of resolving the namespace UUID, but also, we would need to duplicate this on every function that sends some event to PostHog. 

I think that we can make our own wrapper for `Enqueue` event where we include the namespace UUID, and potentially, it could enrich the event in a generic way. So, each `Track...` function would call to an internal `track` or `enqueue` which calls the resolver. That doesn't resolve the problem of this component being the responsible of resolving the namespace UUID, but at least, removes duplicated code. WDYT?

jLopezbarb: The wrapper approach assumes namespace is always relevant to every event, but that's not necessarily true, some events are CLI-level (manifest information, config changes...) and have no meaningful namespace context.
                                                                                         
  If namespace ends up being truly universal, I think that the right place is commonProperties,  which is already the mechanism for enriching all events consistently. Adding it there is explicit and doesn't hide the enrichment inside the transport layer.                   
                                                                                      
That said, I agree duplicating the resolver call in every `Track...` function isn't ideal, happy to refactor it into `commonProperties` if that makes sense to you.

pkg/analytics/posthog.go
jLopezbarb: We are doing that for other properties such as `error_reason`, but I agree that is different with namespace. @codyjlandstrom WDYT? Should we send empty namespace or not send anything at all if we fail to resolve the UID of the namespace

pkg/k8s/namespaces/uid_resolver.go
jLopezbarb: The Kubernetes API field is `.metadata.uid` with type `types.UID`. Even though the value is UUID-formatted, Kubernetes officially calls it `UID`, not `UUID`. Renaming to "uuid" would be less accurate. The current naming follows Kubernetes conventions. WDYT

pkg/analytics/posthog.go
ifbyol: But one of the advantages of the wrapper is that you can call it....or not from the `Track...` function. That is up to the event you want to track. You don't have to use it always. We don't have to call it wrapper, maybe other name fits better

pkg/k8s/namespaces/uid_resolver.go
ifbyol: Hmm very good point! let's leave `uid` then, makes sense to use that

pkg/analytics/posthog.go
codyjlandstrom: I think we should leave it empty on failure. That will make it super straightforward for handling in posthog.

I'll have to make sure we have this noted in case we start getting many of these, how to handle in a dashboard - but lets start getting the data first.

pkg/analytics/posthog.go
```diff
@@ -112,82 +135,66 @@ func getAgent() string {
...
-				Set("customer_name", ctx.CompanyName),
-		}); err != nil {
-			oktetoLog.Infof("failed to send posthog group identify (customer): %s", err)
+// withNamespace returns an enricherFn that resolves the namespace UID and sets
+// it as the "namespace" property. It is a no-op when namespace is empty or the
+// resolver is unavailable.
+func (b *posthogBackend) withNamespace(namespace string) enricherFn {
+	return func(ctx context.Context, props posthog.Properties) {
+		if namespace == "" || b.nsResolver == nil {
+			return
```
ifbyol, line 165: We should set the namespace to empty value as Cody mentioned in a previous comment

pkg/analytics/posthog.go
```diff
@@ -112,82 +135,66 @@ func getAgent() string {
...
 }
 
-// Close flushes pending events and shuts down the PostHog client.
+// Close waits for any in-flight goroutines to finish enqueuing, then flushes
+// and shuts down the PostHog client.
 func (b *posthogBackend) Close() {
 	if b.client == nil {
 		return
 	}
+	b.wg.Wait()
```
ifbyol, line 199: We have to bear in mind that this could potentially block the CLI execution termination up to 5 seconds. Did you experience it in any of your tests?

jLopezbarb: Nope, I tested with several ones but never got stuck

pkg/analytics/posthog.go
```diff
@@ -112,82 +135,72 @@ func getAgent() string {
...
+func (b *posthogBackend) withNamespace(namespace string) enricherFn {
+	return func(ctx context.Context, props posthog.Properties) {
+		if namespace == "" {
+			props["namespace"] = ""
+			return
+		}
+		if b.nsResolver == nil {
+			props["namespace"] = ""
+			return
+		}
```
ifbyol: Nit:

```suggestion
		if namespace == "" || b.nsResolver == nil {
			props["namespace"] = ""
			return
		}
```

ifbyol (approved):
Minor comment about combining 2 conditions in one single if, but the rest looks good

PR #5052, opened 2026-06-11
feat(analytics): add deploy_pipeline_triggered and deploy_preview_triggered PostHog events

jLopezbarb:
## Proposed changes

Adds two new PostHog analytics events — `deploy_pipeline_triggered` and `deploy_preview_triggered` — that fire at the moment the CLI triggers a pipeline or preview deployment. A `workflowID` UUID is generated at the command layer and threaded through both the analytics event and the corresponding GraphQL mutation, so the front-end event and the back-end pipeline run can be correlated.

Key changes:

- **`pkg/analytics`** — Two new files (`pipeline_triggered.go`, `preview_triggered.go`) define the metadata structs and `Tracker` methods for the new events. `posthog.go` gains `TrackDeployPipelineTriggered` and `TrackDeployPreviewTriggered` implementations, a `withPreview` enricher that resolves the preview namespace UID, and the exported helper `IsWithinPreview()`. The Mixpanel backend gets no-op stubs to satisfy the updated `analyticsBackend` interface.
- **`pkg/okteto/pipeline.go` / `preview.go`** — `workflowID` is propagated as a GraphQL argument in all mutation variants. Backward-compatible fallback logic is extended so that if the connected Okteto cluster does not yet support the `workflowID` argument, the CLI retries without it and gracefully degrades.
- **`pkg/types`** — `PipelineDeployOptions.WorkflowID` field added; `PreviewInterface.DeployPreview` signature updated to include `workflowID`.
- **`cmd/pipeline` / `cmd/preview`** — `workflowID` stored on `DeployOptions`, analytics events fired inside `ExecuteDeployPipeline` and `ExecuteDeployPreview` before the GraphQL call. `IsRedeploy` uses `!oktetoErrors.IsNotFound(getErr)` to correctly distinguish missing previews from transient errors.
- **`cmd/deploy`, `cmd/destroy`, `cmd/test`, `cmd/up`** — All call sites of `pipelineCMD.NewCommand()` and `previewCMD.NewCommand()` updated to pass the analytics tracker through.

## How to validate

```bash
# Build
make build

# Unit tests
go test ./pkg/analytics/... ./pkg/okteto/... ./cmd/pipeline/... ./cmd/preview/...

# Lint
make lint
```

Manual validation:
1. Trigger a pipeline deployment and confirm `deploy_pipeline_triggered` appears in PostHog with `workflow_id`, `is_redeploy`, `namespace`, `repo_url`
2. Redeploy the same pipeline and confirm `is_redeploy: true`
3. Trigger a preview deployment and confirm `deploy_preview_triggered` appears in PostHog with `preview` (namespace UID)

## CLI Quality Reminders 🔧


cmd/pipeline/deploy.go
```diff
@@ -85,9 +87,10 @@ type DeployOptions struct {
 	RedeployDependencies bool
 	IsDependency         bool
 	DependenciesIsSet    bool
+	WorkflowID           string
```
ifbyol, line 90: Just curiosity, why storing it as part of the `DeployOptions`? Is it the only struct we have to pass the information down to the rest of the flow?

cmd/preview/deploy.go
```diff
@@ -103,6 +105,19 @@ okteto preview deploy --wait=false`,
...
+	opts.workflowID = uuid.New().String()
+	if pw.analyticsTracker != nil {
+		pw.analyticsTracker.TrackDeployPreviewTriggered(ctx, analytics.DeployPreviewTriggeredMetadata{
+			WorkflowID:      opts.workflowID,
+			RepoURL:         opts.repository,
+			Preview:         opts.name,
+			IsWithinPreview: analytics.IsWithinPreview(),
+			IsRedeploy:      !oktetoErrors.IsNotFound(getErr),
+		})
+	}
```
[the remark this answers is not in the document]
ifbyol: @jLopezbarb I think both things mentioned here are important, take a look at them please

pkg/analytics/posthog.go
```diff
@@ -190,6 +213,49 @@ func (b *posthogBackend) TrackImageBuild(ctx context.Context, m *ImageBuildMetad
...
 
+// TrackDeployPipelineTriggered sends a deploy_pipeline_triggered event to PostHog.
+func (b *posthogBackend) TrackDeployPipelineTriggered(ctx context.Context, m DeployPipelineTriggeredMetadata) {
+	if b.client == nil || !analyticsEnabled() {
+		return
+	}
+
+	userID := okteto.GetContext().UserID
+	props := commonPostHogProperties()
+	props["workflow_id"] = m.WorkflowID
```
ifbyol, line 233: Nit: In other events we are calling it `up_workflow_id`. Would it make sense to call this `deploy_workflow_id`? should we just call it `workflow_id` for all the events?

pkg/analytics/posthog.go
```diff
@@ -190,6 +213,49 @@ func (b *posthogBackend) TrackImageBuild(ctx context.Context, m *ImageBuildMetad
...
+	if m.RepoURL != "" {
+		props["repo_url"] = hashString(m.RepoURL)
+	}
+	b.enqueue(ctx, userID, posthogDeployPreviewTriggeredEvent, props, b.withPreview(m.Preview))
+}
+
+// IsWithinPreview reports whether the current execution context is inside a
+// preview environment (set by the Okteto platform via env var).
+func IsWithinPreview() bool {
+	return os.Getenv(constants.OktetoIsPreviewEnvVar) == "true"
```
ifbyol: This might not be enough. If I use `okteto ns use <preview-name>`, and then, I execute `okteto pipeline deploy`, it should have `is_within_preview` to `true`, but using this check, it would be `false`

pkg/okteto/preview_test.go
```diff
@@ -225,13 +225,58 @@ func TestDeployPreview(t *testing.T) {
...
+
+	t.Run("workflowID not supported with labels - falls back to legacy mutation", func(t *testing.T) {
+		client := &fakeGraphQLMultipleCallsClient{
+			errs:           []error{errors.New(`Unknown argument "workflowID" on field "deployPreview" of type "Mutation"`), nil},
+			mutationResult: []interface{}{nil, successResult},
+		}
+		pc := previewClient{client: client, namespaceValidator: newNamespaceValidator()}
+		response, err := pc.DeployPreview(context.Background(), "test", "", "", "", "", "", "", nil, []string{"key=val"}, false)
+		assert.NoError(t, err)
+		assert.Equal(t, expectedResponse, response)
```
ifbyol: nit: let's use `require` instead of `assert`

ifbyol (commented):
If a `okteto deploy` triggers a pipeline deploy due to dependencies. Do they share the same workflow id? or each pipeline/dependency would have a different workflow if than the "parent" one?

pkg/analytics/posthog.go
jLopezbarb: We decided to call it workflow_id for deploy. In up we ended up calling  wokflow_id so I guess it's already done :) https://github.com/okteto/okteto/blob/fdde5c44edd898bf59190455aa28c2b7db59dcbc/pkg/analytics/up.go#L272

cmd/pipeline/deploy.go
jLopezbarb: We could be passing it on all the params but given the amount of functions that we had. I ended up setting it here because it's something that we need to send it via graphql  and given that the other properties are here I think it belongs here

cmd/preview/deploy.go
```diff
@@ -103,6 +106,25 @@ okteto preview deploy --wait=false`,
...
+		opts.workflowID = envWorkflowID
+	}
+	if pw.analyticsTracker != nil {
+		_, getErr := pw.okClient.Previews().Get(ctx, opts.name)
+		pw.analyticsTracker.TrackDeployPreviewTriggered(ctx, analytics.DeployPreviewTriggeredMetadata{
+			WorkflowID:       opts.workflowID,
+			ParentWorkflowID: os.Getenv(constants.OktetoParentWorkflowIDEnvVar),
+			RepoURL:          opts.repository,
+			Preview:          opts.name,
+			IsWithinPreview: analytics.IsWithinPreview(ctx, func(ctx context.Context, ns string) error {
```
ifbyol: Do `isWithinPreview` make sense for the deploy_preview_triggered event? There is no concept of preview within another preview. I think this should be omitted

cmd/preview/deploy.go
```diff
@@ -103,6 +106,25 @@ okteto preview deploy --wait=false`,
...
+				_, err := pw.okClient.Previews().Get(ctx, ns)
+				return err
+			}),
+			IsRedeploy: getErr == nil,
```
ifbyol: Does this cover any kind of error? If there is a transient error getting the preview, does it count as redeploy?

cmd/preview/deploy.go
```diff
@@ -103,6 +106,25 @@ okteto preview deploy --wait=false`,
...
+	if pw.analyticsTracker != nil {
+		_, getErr := pw.okClient.Previews().Get(ctx, opts.name)
+		pw.analyticsTracker.TrackDeployPreviewTriggered(ctx, analytics.DeployPreviewTriggeredMetadata{
+			WorkflowID:       opts.workflowID,
+			ParentWorkflowID: os.Getenv(constants.OktetoParentWorkflowIDEnvVar),
+			RepoURL:          opts.repository,
+			Preview:          opts.name,
+			IsWithinPreview: analytics.IsWithinPreview(ctx, func(ctx context.Context, ns string) error {
+				_, err := pw.okClient.Previews().Get(ctx, ns)
+				return err
```
ifbyol: I added a previous comment indicating that I believe this is not needed. But if it is, in line 114 we already got the error getting the preview, no need to execute, this could be just

```
IsWithinPreview: analytics.IsWithinPreview(ctx, func(ctx context.Context, ns string) error {
    return getErr
}
```

But again, I think this is not needed at all

pkg/analytics/posthog.go
```diff
@@ -112,15 +115,22 @@ func commonPostHogProperties() posthog.Properties {
 		"arch":               runtime.GOARCH,
 		"machine_id":         get().MachineID,
 		"measurement_source": "cli",
-		"trigger_source":     "cli",
+		"trigger_source":     config.GetDeployOrigin(),
 		"is_agent":           agent != "",
+		"is_ci":              isCI(),
+		"is_automation":      !isCI() && agent == "",
```
ifbyol: Is this correct? Wouldn't this criteria indicate that something is an automation if I trigger something manually?

pkg/analytics/posthog.go
```diff
@@ -192,6 +222,65 @@ func (b *posthogBackend) TrackImageBuild(ctx context.Context, m *ImageBuildMetad
...
+// TrackDeployPreviewTriggered sends a deploy_preview_triggered event to PostHog.
+func (b *posthogBackend) TrackDeployPreviewTriggered(ctx context.Context, m DeployPreviewTriggeredMetadata) {
+	if b.client == nil || !analyticsEnabled() {
+		return
+	}
+
+	userID := okteto.GetContext().UserID
+	props := commonPostHogProperties()
+	props["workflow_id"] = m.WorkflowID
+	props["is_within_preview"] = m.IsWithinPreview
```
ifbyol: As other comment in another file, I believe this is not correct for the deploy preview event

pkg/analytics/posthog.go
```diff
@@ -192,6 +222,65 @@ func (b *posthogBackend) TrackImageBuild(ctx context.Context, m *ImageBuildMetad
...
+	props["is_redeploy"] = m.IsRedeploy
+	if m.ParentWorkflowID != "" {
+		props["parent_workflow_id"] = m.ParentWorkflowID
+	}
```
ifbyol: Preview would never have parent workflow, at least, no that we could calculate easily, right? This is not like the case of a pipeline triggering another pipeline because of a dependency. So, I think this is another candidate to delete

cmd/preview/deploy.go
jLopezbarb: Agreed, removed `is_within_preview` from the deploy_preview_triggered event in 98278d5d.

cmd/preview/deploy.go
jLopezbarb: Done in 98278d5d: dropped the whole `is_within_preview` block, which also removes the redundant second `Previews().Get()` call.

pkg/analytics/posthog.go
jLopezbarb: Removed `is_within_preview` from the preview event in 98278d5d.

pkg/analytics/posthog.go
jLopezbarb: Right, removed `parent_workflow_id` from the preview event (and its metadata field) in 98278d5d.

cmd/preview/deploy.go
jLopezbarb: Good catch. Changed in de77fd11 to `IsRedeploy: !oktetoErrors.IsNotFound(getErr)`: only a NotFound means a brand-new preview, so a transient error no longer gets misreported as a first-time deploy (it stays as redeploy=true rather than silently flipping to a fresh deploy).

pkg/analytics/posthog.go
jLopezbarb: You're right, it was inverted: a manual human run was being flagged as automation. Fixed in de77fd11 to `is_automation = isCI() || agent != ""`, so it's true only for CI or AI-agent actors and false for manual runs.

pkg/analytics/posthog.go
jLopezbarb: We decided to drop `is_within_preview` entirely in de77fd11 (the detection was unreliable for exactly this `okteto ns use <preview>` case), so this concern no longer applies.

jLopezbarb:
> If a `okteto deploy` triggers a pipeline deploy due to dependencies, do they share the same workflow id?

No, each pipeline/dependency gets its **own** `workflow_id`. What ties them together is a separate **parent** id:

- `okteto deploy` generates a single `parentWorkflowID` (UUID) and sends it as a GraphQL argument on every dependency deploy, so the backend can correlate all dependencies of that deploy with their parent (see `cmd/deploy/deploy.go:720-751`).
- Each `pipeline deploy` (parent or dependency) generates its own `workflow_id` via `uuid.New()`, unless the backend injects `OKTETO_WORKFLOW_ID`.
- For the analytics event specifically, `parent_workflow_id` is sourced only from the backend-injected `OKTETO_PARENT_WORKFLOW_ID` env var (not the GraphQL arg) — that distinction is documented in `cmd/pipeline/deploy.go:240-242`.

So: different `workflow_id` per pipeline, shared `parent_workflow_id` for correlation.

pkg/okteto/pipeline_test.go
```diff
@@ -250,6 +253,128 @@ func TestDeployPipeline(t *testing.T) {
...
+	}
+
+	t.Run("workflowId not supported - falls back to legacy mutation", func(t *testing.T) {
+		client := &fakeGraphQLMultipleCallsClient{
+			errs:           []error{errors.New(`Unknown argument "workflowId" on field "deployGitRepository" of type "Mutation"`), nil},
+			mutationResult: []interface{}{nil, successResult},
+		}
+		pc := pipelineClient{client: client}
+		response, err := pc.Deploy(context.Background(), types.PipelineDeployOptions{Name: "test"})
+		assert.NoError(t, err)
```
ifbyol: Nit: Change `assert.` by `require.`

ifbyol (approved):

PR #5064, opened 2026-06-23
feat(stack): native runtime cloud access for Compose services via x-okteto-identity-token

ifbyol:
# Proposed changes

This PR adds a new per-service Okteto Compose directive, `x-okteto-identity-token`, that
projects an audience-scoped, kubelet-refreshed Kubernetes ServiceAccount token into the
runtime pod. Applications can then perform OIDC / AWS STS web-identity federation **without
static credentials**.

The directive is cloud-agnostic: the CLI injects no environment variables — the user sets
`AWS_ROLE_ARN`, `AWS_REGION` and `AWS_WEB_IDENTITY_TOKEN_FILE` (or any provider equivalent)
themselves. It is opt-in by the presence of the field; there is no feature flag, and services
that do not declare it are unaffected.

Jira: PROD-497

## Manifest shape

```yaml
services:
  api:
    image: my-api:latest
    environment:
      - AWS_ROLE_ARN=arn:aws:iam::<acct>:role/api
      - AWS_REGION=us-east-1
      - AWS_WEB_IDENTITY_TOKEN_FILE=/var/run/secrets/tokens/aws/token
    x-okteto-identity-token:
      audience: sts.amazonaws.com
      mount_path: /var/run/secrets/tokens/aws
      # expiration_seconds: 3600   # optional; literal int or ${ENV}
```

## What changed

- **Model parsing & validation** (`pkg/model/stack.go`, `pkg/model/stack_serializer.go`):
  - New `ServiceIdentityToken` type and a typed `x-okteto-identity-token` field on the service
    that bypasses `validateExtensions` (so the `x-` directive is accepted rather than rejected).
  - `validateIdentityToken`: `audience` required, `mount_path` required and POSIX-absolute,
    `expiration_seconds` (when set) must be ≥ 600s (the kubelet minimum).
  - New `IdentityTokenExpiration` type with a custom `UnmarshalYAML` that accepts either a YAML
    integer (the common case) or a numeric string, so values supplied through env-var expansion
    (e.g. `expiration_seconds: ${EXP}`) parse correctly.
  - `mergeServices` carries the directive across manifest merges.
- **Kubernetes translation** (`pkg/cmd/stack/translate.go`):
  - Builds a projected `serviceAccountToken` volume (`okteto-identity-token`) with the configured
    audience, projected path `token`, and expiration (default 3600s when unset).
  - Mounts it read-only at `mount_path`.
  - Wired into Deployment, StatefulSet and Job translation, including explicit wiring into
    `translateDeployment`, which does not use the shared volume helpers.

## Notes / scope

- Pods use the namespace `default` ServiceAccount. A dedicated per-service SA was reviewed and
  deliberately deferred: per-service role assignment already works via each service's own
  `AWS_ROLE_ARN`, and isolation is tunable via per-service `audience` values.
- Documenting the cluster issuer (IRSA) trust setup is a follow-up.

## How to validate

1. Build the CLI and run the unit tests for the affected packages:
   ```bash
   make build
   go test ./pkg/model/... -run IdentityToken -v
   go test ./pkg/cmd/stack/... -run IdentityToken -v
   ```
1. Validation rules — confirm a malformed directive is rejected. Create `compose-bad.yml`:
   ```yaml
   services:
     api:
       image: nginx:latest
       x-okteto-identity-token:
         mount_path: var/run/secrets   # not absolute, and no audience
   ```
   then run `okteto deploy -f compose-bad.yml` and observe it fails with
   `invalid 'x-okteto-identity-token' for service 'api': 'audience' is required`
   (and `'mount_path' must be an absolute path` once an audience is added).
1. End-to-end — deploy a valid compose file and inspect the rendered workload. Create
   `docker-compose.yml`:
   ```yaml
   services:
     app:
       image: nginx:latest
       ports:
         - 80
       x-okteto-identity-token:
         audience: sts.amazonaws.com
         mount_path: /var/run/okteto/identity
   ```
   then:
   ```bash
   okteto deploy
   kubectl get deployment app -o yaml
   ```
   Observe a projected volume `okteto-identity-token` with a `serviceAccountToken` projection
   (`audience: sts.amazonaws.com`, `path: token`, `expirationSeconds: 3600`) and a read-only
   volume mount at `/var/run/okteto/identity` on the app container. This is covered by the e2e
   test `integration/deploy/compose_test.go::TestDeployComposeWithIdentityToken`.

## CLI Quality Reminders 🔧


ifbyol:
For testing, I created a simple go web service that list all the s3 buckets available. I didn't create a repository, but mainly this was the app code:

```golang
package main

import (
	"context"
	"encoding/json"
	"log"
	"net/http"
	"os"
	"time"

	"github.com/aws/aws-sdk-go-v2/config"
	"github.com/aws/aws-sdk-go-v2/service/s3"
)

// s3API is the subset of the S3 client we use, so it can be mocked in tests.
type s3API interface {
	ListBuckets(ctx context.Context, params *s3.ListBucketsInput, optFns ...func(*s3.Options)) (*s3.ListBucketsOutput, error)
}

type server struct {
	s3 s3API
}

type bucket struct {
	Name         string    `json:"name"`
	CreationDate time.Time `json:"creationDate"`
}

type listBucketsResponse struct {
	Buckets []bucket `json:"buckets"`
}

type errorResponse struct {
	Error string `json:"error"`
}

func (s *server) handleListBuckets(w http.ResponseWriter, r *http.Request) {
	ctx, cancel := context.WithTimeout(r.Context(), 15*time.Second)
	defer cancel()

	out, err := s.s3.ListBuckets(ctx, &s3.ListBucketsInput{})
	if err != nil {
		log.Printf("failed to list buckets: %v", err)
		writeJSON(w, http.StatusBadGateway, errorResponse{Error: "failed to list buckets: " + err.Error()})
		return
	}

	resp := listBucketsResponse{Buckets: make([]bucket, 0, len(out.Buckets))}
	for _, b := range out.Buckets {
		item := bucket{}
		if b.Name != nil {
			item.Name = *b.Name
		}
		if b.CreationDate != nil {
			item.CreationDate = *b.CreationDate
		}
		resp.Buckets = append(resp.Buckets, item)
	}

	writeJSON(w, http.StatusOK, resp)
}

func handleHealth(w http.ResponseWriter, _ *http.Request) {
	writeJSON(w, http.StatusOK, map[string]string{"status": "ok"})
}

func writeJSON(w http.ResponseWriter, status int, body any) {
	w.Header().Set("Content-Type", "application/json")
	w.WriteHeader(status)
	if err := json.NewEncoder(w).Encode(body); err != nil {
		log.Printf("failed to encode response: %v", err)
	}
}

func main() {
	ctx := context.Background()

	// LoadDefaultConfig walks the standard AWS credential chain. With
	// AWS_ROLE_ARN + AWS_WEB_IDENTITY_TOKEN_FILE set (as they are for IRSA /
	// Okteto federated identity), it transparently assumes the role via STS
	// using the projected web identity token.
	cfg, err := config.LoadDefaultConfig(ctx)
	if err != nil {
		log.Fatalf("unable to load AWS config: %v", err)
	}

	srv := &server{s3: s3.NewFromConfig(cfg)}

	mux := http.NewServeMux()
	mux.HandleFunc("/buckets", srv.handleListBuckets)
	mux.HandleFunc("/healthz", handleHealth)

	port := os.Getenv("PORT")
	if port == "" {
		port = "8080"
	}

	httpServer := &http.Server{
		Addr:              ":" + port,
		Handler:           mux,
		ReadHeaderTimeout: 10 * time.Second,
	}

	log.Printf("listening on :%s", port)
	if err := httpServer.ListenAndServe(); err != nil && err != http.ErrServerClosed {
		log.Fatalf("server error: %v", err)
	}
}

```

And I have verified that setting the environment variables and the `x-okteto-identity-token` in the compose the application is able to connect to AWS

jLopezbarb (approved):

PR #5081, opened 2026-07-07
feat(analytics): add wake_triggered PostHog event

jLopezbarb:
# Proposed changes

Jira: [DEV-1435](https://okteto.atlassian.net/browse/DEV-1435)

Adds a new `wake_triggered` PostHog analytics event, emitted whenever a namespace or preview environment is woken through the GraphQL `Wake` mutation. The event is fired from the three call sites that trigger a wake, and only when the wake actually succeeds:

- `okteto namespace wake` (explicit)
- `okteto preview wake` (explicit)
- the implicit wake performed by `okteto up` on a sleeping namespace

The event carries two event-specific properties: `namespace` (resolved to its Kubernetes UID via the existing `withNamespace` enricher) and `is_preview`. All common properties (`customer_name`, `cluster_id`, `cluster_url`, `cluster_version`, `user_id`, `measurement_source`, `trigger_source`, `is_agent`, `agent_type`, `machine_id`, etc.) are attached automatically by `commonPostHogProperties()`, so they are not set per event. `workflow_id` and `ui_element` from the product spec are intentionally omitted (`workflow_id` is not generated yet; `ui_element` is UI-only).

`is_preview` is `true` for preview wakes, computed via `analytics.IsWithinPreview` for the implicit `up` wake (matching the `deploy` command), and `false` for namespace wakes.

Following the existing two-backend pattern, the event is implemented on the PostHog backend and is a no-op on the legacy Mixpanel backend.

## How to validate

1. Build the CLI: `make build`.
1. Put a namespace to sleep and wake it with debug logging, confirming no `failed to send posthog analytics` error appears: `./bin/okteto namespace sleep <ns>` then `./bin/okteto --log-level=debug namespace wake <ns>`.
1. Wake a preview and confirm the event is emitted with `is_preview=true`: `./bin/okteto --log-level=debug preview wake <preview>`.
1. Run `./bin/okteto up` on a sleeping namespace and confirm the implicit wake emits the event.
1. Run the unit tests: `go test ./pkg/analytics/... ./cmd/namespace/... ./cmd/preview/... ./cmd/up/...`.

## CLI Quality Reminders 🔧

[DEV-1435]: https://okteto.atlassian.net/browse/DEV-1435?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cmd/namespace/namespace.go
+++ b/cmd/namespace/namespace.go
@@ -18,11 +18,17 @@ import (
 
 	contextCMD "github.com/okteto/okteto/cmd/context"
 	"github.com/okteto/okteto/cmd/utils"
+	"github.com/okteto/okteto/pkg/analytics"
 	"github.com/okteto/okteto/pkg/log/io"
 	"github.com/okteto/okteto/pkg/okteto"
 	"github.com/okteto/okteto/pkg/types"
 	"github.com/spf13/cobra"
 )
+
+// wakeAnalyticsTracker tracks the wake_triggered event.
+type wakeAnalyticsTracker interface {
+	TrackWakeTriggered(ctx context.Context, m analytics.WakeTriggeredMetadata)
+}
 
 // Command has all the namespaces subcommands
 type Command struct {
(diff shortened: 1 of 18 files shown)


ifbyol (approved):
We should review how we are calculating if a namespace is a preview or not. It doesn't scale to add a call to Kubernetes API for each event we add that might need the property. And we should review it as soon as we can

PR #5085, opened 2026-07-11
Assign SSH port right before opening it to reduce conflict window with other processes competing for ports

pchico83:
`okteto up` picks its dynamic ports (SSH tunnel + syncthing forwards) with a select-then-bind pattern that has a TOCTOU gap: another process can grab the port between choosing it and binding it. Under parallel load (e.g. e2e) this flakes as local "port N is already in-use".

Changes:
- Wrap dynamic-port collisions in a new transient `errDynamicPortInUse`; the existing `activate()` retry loop then reruns with freshly allocated ports.
- Split `LoadRemote`: the SSH pubkey secret stays early (needed for pod translation); the SSH port is now assigned in sshForwards, right before binding, so it's re-picked on each retry.

### DIFF
--- a/cmd/up/forwards.go
+++ b/cmd/up/forwards.go
@@ -90,6 +90,11 @@ func (up *upContext) sshForwards(ctx context.Context) error {
 		return err
 	}
 
+	// assign the SSH tunnel port right before creating the forward, to minimize
+	// the window between choosing a free port and binding it. If it is already
+	// assigned (user-pinned or a previous connection), the same port is reused.
+	up.Dev.AssignRemotePort()
+
 	oktetoLog.Infof("starting SSH port forwards")
 	f := forwardk8s.NewPortForwardManager(ctx, up.Dev.Interface, restConfig, k8sClient, up.Namespace)
 	if err := f.Add(forward.Forward{Local: up.Dev.RemotePort, Remote: up.Dev.SSHServerPort}); err != nil {
@@ -109,14 +114,14 @@ func (up *upContext) sshForwards(ctx context.Context) error {
 		return err
 	}
 
-	if err := ssh.AddEntry(up.Dev.Name, up.Dev.Interface, up.Dev.RemotePort); err != nil {
-		oktetoLog.Infof("failed to add entry to your SSH config file: %s", err)
-		return fmt.Errorf("failed to add entry to your SSH config file")
-	}
-
 	err = up.Forwarder.Start(up.Pod.Name, up.Namespace)
 	if err != nil {
 		return err
+	}
+
+	if err := ssh.AddEntry(up.Dev.Name, up.Dev.Interface, up.Dev.RemotePort); err != nil {
+		oktetoLog.Infof("failed to add entry to your SSH config file: %s", err)
+		return fmt.Errorf("failed to add entry to your SSH config file")
 	}
 
 	if isNeededGlobalForwarder(up.Manifest.GlobalForward) {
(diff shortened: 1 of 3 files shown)


pkg/model/dev.go
```diff
@@ -693,19 +693,26 @@ func (dev *Dev) validateSecurityContext() error {
...
+// the tunnel is opened. It must only be called when the user has not pinned the
+// port (via the manifest 'remote' field or the --remote flag).
+func (dev *Dev) AssignRemotePort() error {
+	// a failure here means no local socket can be opened (e.g. an invalid
+	// interface); any bind on the same interface would fail too, so it is
+	// returned as a fatal error.
+	p, err := GetAvailablePort(dev.Interface)
+	if err != nil {
+		return fmt.Errorf("failed to get a free local port for the SSH connection: %w", err)
 	}
```
ifbyol, line 711: Before we were defaulting to `oktetoDefaultSSHServerPort` when there was an error getting an available port, but I think that logic is now missing, and we hard fail

cmd/up/forwards.go
ifbyol, line 77: Should we pass this through `asDynamicPortErr`? Why don't we do it here, but we do it in the call to `Start` within `sshForwards`?

cmd/up/forwards.go
```diff
@@ -90,19 +102,30 @@ func (up *upContext) sshForwards(ctx context.Context) error {
 		return err
 	}
 
+	// assign the SSH tunnel port right before binding it; on a retry a fresh
+	// port is picked. A user-pinned port is kept as is.
+	if up.autoAssignRemotePort {
+		if err := up.Dev.AssignRemotePort(); err != nil {
+			return err
```
ifbyol: Can't `GetAvailablePort` return a problem binding a port? Why don't we get the `asDynamicPorError` call here?

pkg/k8s/forward/manager.go
```diff
@@ -109,7 +124,7 @@ func (p *PortForwardManager) Add(f forward.Forward) error {
 				return fmt.Errorf("local port %d is privileged. Try running \"sudo setcap 'cap_net_bind_service=+ep' /usr/local/bin/okteto\" and try again", f.Local)
 			}
 		}
-		return fmt.Errorf("local port %d is already in-use in your local machine", f.Local)
+		return fmt.Errorf("local port %d is already in-use in your local machine: %w", f.Local, oktetoErrors.ErrPortAlreadyAllocated)
```
ifbyol: I think this would affect to the GlobalForward logic where users specify the port. Is this protected with the `autoAssignRemotePort` flag?

ifbyol (commented):
Apart of the comments, I think that any transient error now would provoke a change in the port chosen to be used. This might lead to problems is someone is using ssh session outside Okteto

pchico83:
@ifbyol I simplified the approach to assign the SSH dynamic port right before creating the port forwards, reducing the conflict window.
The previous logic didn't eliminate the risk of conflicting ports in all cases, and it was riskier.
I think this change is enough to reduce the port conflicts in e2e noticeably, and it's low risk

pkg/model/dev.go
ifbyol: This comment doesn't apply anymore. Resolving it

ifbyol (approved):

ifbyol:
> @ifbyol I simplified the approach to assign the SSH dynamic port right before creating the port forwards, reducing the conflict window. The previous logic didn't eliminate the risk of conflicting ports in all cases, and it was riskier. I think this change is enough to reduce the port conflicts in e2e noticeably, and it's low risk

@pchico83 Looks good to me, I think it is going to improve the tests stability for sure

PR #5121, opened 2026-08-13
ci: shorten release tag regex for CircleCI config compiler

jLopezbarb:
# Proposed changes

CircleCI is shipping a new config compiler (rolling out **2026-09-21**) that restricts the regular-expression engine used to evaluate workflow tag filters. Among other changes, it rejects overly long patterns with `Invalid regular expression: ... Pattern too long`, which breaks our `release` workflow because the `&release-regex` alias used the full ~180-character semver.org regex.

This PR replaces that regex with a short pattern that matches **exactly** the release tags we cut, as documented in `scripts/ci/release.sh`:

- stable: `MAJOR.MINOR.PATCH`
- beta: `MAJOR.MINOR.PATCH-beta.N`

```
before: /^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(?:-(...))?(?:\+(...))?$/   (~180 chars)
after:  /^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(-(beta)\.(0|[1-9]\d*))?$/    (~68 chars)
```

The new pattern is both short (avoids the compiler limit) and strict: it disallows leading zeros and only accepts the real release channels, so malformed tags like `01.02.03`, `1.2.3-.` or `1.2.3-beta..1` can no longer trigger the release workflow. The alias is used by the `release` workflow (`build-binaries`, `push-image-tag`, `release-job`) and as a tag `ignore` filter for the unit-test jobs, so the same set of tags continues to be filtered identically.

## How to validate

1. Run `circleci config validate .circleci/config.yml` and `circleci config validate .circleci/config.yml --next` — both report the config as valid.
1. Confirm the pattern matches every real release tag and rejects malformed ones. Checked against all 582 existing tags: only the two malformed strays (`V2.4.0-beta.1`, `version`) are excluded; every valid stable/beta/rc/dev release matches.
1. Confirm the `release` workflow tag filters still resolve against real semver tags.

## CLI Quality Reminders 🔧

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,9 +6,9 @@ aliases:
     echo $GCLOUD_SERVICE_KEY | gcloud auth activate-service-account --key-file=-
     gcloud --quiet config set project ${GOOGLE_PROJECT_ID}
     gcloud --quiet config set compute/zone ${GOOGLE_COMPUTE_ZONE}
-  # semver tags: X.Y.Z with optional prerelease/build metadata (e.g. 3.22.0, 3.22.0-beta.3).
+  # semver release tags: stable X.Y.Z and prerelease X.Y.Z-{beta,rc,dev}.N (see scripts/ci/release.sh).
   # Kept short because CircleCI's new config compiler (2026-09-21) rejects overly long patterns.
-  - &release-regex /^\d+\.\d+\.\d+(-[0-9A-Za-z.-]+)?(\+[0-9A-Za-z.-]+)?$/
+  - &release-regex /^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(-(beta|rc|dev)\.\d+)?$/
   - &release-branch-regex /^release-\d+\.\d+$/
   - &docker-login echo "$DOCKER_PASS" | docker login -u "$DOCKER_USER" --password-stdin
   - &github-packages-login echo "$GITHUB_PACKAGES_OKTETOBOT_TOKEN" | docker login ghcr.io -u "oktetobot" --password-stdin
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,8 +6,9 @@ aliases:
     echo $GCLOUD_SERVICE_KEY | gcloud auth activate-service-account --key-file=-
     gcloud --quiet config set project ${GOOGLE_PROJECT_ID}
     gcloud --quiet config set compute/zone ${GOOGLE_COMPUTE_ZONE}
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -6,8 +6,9 @@ aliases:
     echo $GCLOUD_SERVICE_KEY | gcloud auth activate-service-account --key-file=-
     gcloud --quiet config set project ${GOOGLE_PROJECT_ID}
     gcloud --quiet config set compute/zone ${GOOGLE_COMPUTE_ZONE}
-  # standard semver regex as defined in: https://semver.org/
-  - &release-regex /^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(?:-((?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$/
+  # semver release tags: stable X.Y.Z and prerelease X.Y.Z-{beta,rc,dev}.N (see scripts/ci/release.sh).
+  # Kept short because CircleCI's new config compiler (2026-09-21) rejects overly long patterns.
```
ifbyol: I would delete the date from the comment. It would become stale soon

.circleci/config.yml
```diff
@@ -6,8 +6,9 @@ aliases:
...
+  - &release-regex /^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(-(beta|rc|dev)\.\d+)?$/
```
ifbyol: As far as I know, `rc` and `dev` are no longer used/valid tags, right? Would it make sense to delete them?

.circleci/config.yml
ifbyol: Done in https://github.com/okteto/okteto/pull/5121/changes/acbf98f9b1a369c687e8247ee8f08262bb38867f

.circleci/config.yml
ifbyol: Done in https://github.com/okteto/okteto/pull/5121/changes/acbf98f9b1a369c687e8247ee8f08262bb38867f

ifbyol:
In the last commit I drop the support of `-rc` and `-dev` tags as they are not being used. There are some legacy tags with `-rc` tags, but they belong to version 2.X, more than 1 year ago

ifbyol (approved):

PR #5140, opened 2026-09-08
perf(analytics): compute is_preview once per command + dedupe up-metrics I/O

jLopezbarb:
# Proposed changes

Jira: [DEV-1442](https://okteto.atlassian.net/browse/DEV-1442)

Follow-up to #5081, raised in review by @ifbyol.

`is_preview` / `is_within_preview` was recomputed independently by **every**
analytics event that needs it. When `OKTETO_WITHIN_PREVIEW` is not set,
`analytics.IsWithinPreview` falls back to an Okteto API call (`Previews().Get`)
to decide whether the active namespace is a preview. In a single `okteto up`
this triggered **up to 3 separate preview detections** (`deploy_started`, `up`,
`wake_triggered`) — each a potential API round-trip — for a value that does not
change within a command run.

This PR fixes that (the ticket) and removes two more per-event redundancies found
in the same analytics hot path:

- **Preview check (DEV-1442, main change):** memoize `is_preview` per namespace
  *inside* `IsWithinPreview`, via a package-singleton `previewResolver`
  (`sync.Map` + `singleflight`). The API detection now runs **at most once per
  command run**, concurrent callers collapse into one call, and — because it
  lives inside `IsWithinPreview` — every call site (`cmd/up`, `deploy`,
  `pipeline`, `preview`, `namespace`) and any future event reuses it
  automatically, with no metadata threading and no call-site changes.
- **Config disk reads:** the analytics config file was re-read from disk on every
  event (10+ reads/session, for both the Mixpanel and PostHog trackers). It is
  now memoized behind a mutex and read at most once per session; a failed read is
  not cached.
- **`isCI()`:** computed once per event instead of twice.
- **Namespace UID:** concurrent namespace-UID lookups now collapse into a single
  `GET namespace` via `singleflight`; errors are not cached, so a failed lookup
  can be retried.
- **Cleanup:** the byte-identical `withNamespace` / `withPreview` enrichers are
  merged into one parametrized helper.

**Behavior / schema:** unchanged, with one intentional, documented exception — on
a *transient* preview-check API error the `up` / `deploy` / `pipeline` events now
consistently report `is_preview=false` instead of potentially diverging, making
the metric deterministic under error conditions.

**Out of scope** (considered and rejected in review): touching `okteto.Context`,
and snapshotting the common-property set / `repo_url` hash (YAGNI). No new
dependencies.

Acceptance criteria (DEV-1442): ✅ at most one preview detection per command run,
✅ all current and future events reuse the same value, ✅ no change to the emitted
`is_preview` / `is_within_preview` values.

## How to validate

1. **Unit tests** (prove the dedup and no regression):
   `go test -race ./pkg/analytics/... ./pkg/k8s/namespaces/...`
   — includes tests asserting the preview check and namespace-UID lookup each run
   only once for repeated/concurrent calls.
1. **Lint:** `make lint` (or `golangci-lint run`) → 0 issues.
1. **Build the CLI:** `make build` → produces `./bin/okteto`.
1. **Manual check (behavior unchanged):** with a context pointing at your cluster,
   run `./bin/okteto up --log-level=debug` in **a preview namespace** and confirm
   the session behaves exactly as before (and analytics carry `is_preview=true`);
   repeat in **a regular namespace** and confirm `is_preview=false`. The value is
   correct in both cases, but the preview check — and the redundant disk / k8s
   work — no longer repeat per analytics event within the run.

## CLI Quality Reminders 🔧

[DEV-1442]: https://okteto.atlassian.net/browse/DEV-1442?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/analytics/config.go
+++ b/pkg/analytics/config.go
@@ -83,9 +83,9 @@ func deprecatedFileExists() bool {
 }
 
 // get returns the analytics config, memoizing the first successful load so later
-// calls skip the disk read. The shared *Analytics is read lock-free, safe only
-// because Enable/Disable/Init never run concurrently with tracking. Disabled
-// fallbacks (missing file / read error) are returned fresh, not cached.
+// calls skip the disk read. Access to the shared *Analytics is guarded by
+// analyticsMu, so concurrent callers are safe. Disabled fallbacks (missing file
+// / read error) are returned fresh, not cached.
 func get() *Analytics {
 	analyticsMu.Lock()
 	defer analyticsMu.Unlock()
--- a/pkg/analytics/config.go
+++ b/pkg/analytics/config.go
@@ -148,12 +148,13 @@ func (a *Analytics) save() error {
 	return nil
 }
 
-// Disable disables analytics. It mutates the shared *Analytics from get(), safe
-// only because Enable/Disable/Init never run concurrently with tracking.
+// Disable disables analytics. trackDisable must run before flipping Enabled:
+// get() returns the shared *Analytics that the tracking path also reads, so
+// disabling first would make analyticsEnabled() short-circuit and drop the event.
 func Disable() error {
 	a := get()
-	a.Enabled = false
 	trackDisable(true)
+	a.Enabled = false
 	return a.save()
 }
 
--- a/pkg/analytics/config.go
+++ b/pkg/analytics/config.go
@@ -17,6 +17,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"sync"
 
 	"github.com/denisbrodbeck/machineid"
 	"github.com/okteto/okteto/pkg/config"
@@ -28,6 +29,9 @@ var (
 	EnterpriseContext = "Enterprise"
 	KubernetesContext = "Kubernetes"
 	currentAnalytics  *Analytics
+	// analyticsMu guards currentAnalytics so the memoized load in get() and the
+	// write in save() never race.
+	analyticsMu sync.Mutex
 )
 
 // Analytics contains the analytics configuration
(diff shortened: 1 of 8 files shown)


pkg/analytics/config.go
```diff
@@ -99,13 +110,16 @@ func get() *Analytics {
 		return &Analytics{Enabled: false, MachineID: ""}
 	}
 
-	return result
+	currentAnalytics = result
+	return currentAnalytics
```
ifbyol, line 114: This change makes that the command `okteto analytics --disable` doesn't trigger an event anymore. This would be a regression. As we return the shared `currentAnalytics`, and we mutate the `Enabled` flag before `trackDisabled`, it will never trigger the event anymore.

I think the disable logic looks weird anyways. We should review it

pkg/analytics/preview_resolver.go
```diff
@@ -0,0 +1,53 @@
...
+func (r *previewResolver) isWithinPreview(ctx context.Context, ns string, checkPreview func(context.Context, string) error) bool {
+	if v, ok := r.cache.Load(ns); ok {
+		return v.(bool)
+	}
+
+	v, _, _ := r.sg.Do(ns, func() (any, error) {
+		if v, ok := r.cache.Load(ns); ok {
+			return v.(bool), nil
+		}
+		result := checkPreview(ctx, ns) == nil
```
ifbyol: AFAIK, this call directly to the previews API, and it returns any error it gets. This mean that a transient error like a timeout, it would be cached as false, affecting to other calls in the same flow. Before this wasn't relevant, because in case of a transient error, any other check would call again to the API, having the chance of changing the value, but since this change, it would cache a transient error as false.

I think we should handle the error, and check the reason before caching it, to just cache in case of not found error

ifbyol (approved):

PR #5150, opened 2026-09-13
fix(env): trim variable key whitespace and validate key in env.Parse

### LINKED ISSUE #4029: Remote okteto deploy broken when `var` has whitespace
**Describe the bug**
If you have a remote deploy, and are trying to run it with `--var key="something with whitespaces"` you'll get an error like

`x  Services args are can only be used while trying to deploy a compose`

**To Reproduce**

1. cd into https://github.com/okteto/movies
1. `okteto deploy --remote --var test="white space"`

You should something like this error
```
 i  Using agustinramirodiaz @ product.okteto.dev as context
 x   x  Services args are can only be used while trying to deploy a compose
 x  Error building image 'okteto/pipeline-runner:1.0.3': build failed: failed to solve: process "/bin/sh -c mkdir -p $HOME/.ssh && echo \"UserKnownHostsFile=/run/secrets/known_hosts\" >> $HOME/.ssh/config &&   okteto deploy --log-output=json --server-name=\"$INTERNAL_SERVER_NAME\" --name \"movies\" --var test=white space --timeout 5m0s" did not complete successfully: exit code: 1
```

:point_up: here we can see the problem, when okteto crafts the Dockerfile for the remote pipeline, it's not quoting the variable's value in  `--var test=white space`, thus miss-configuring the command

**Expected behavior**
The variable should get passed correctly to the remote pipline

- Version `2.21.0` and `master`

Tyagiquamar:
Fixes #4029. Trims leading and trailing whitespace from variable keys when parsing environment variables in \nv.Parse\, and ensures empty key names are rejected with an error.


jLopezbarb:
Thanks for the contribution, @Tyagiquamar! A couple of concerns before we can move forward:

 1. #4029 is already fixed in code — the issue is just still open. That bug is about a remote deploy breaking when a `--var` value contains whitespace (e.g. `okteto deploy --remote --var test="white space"`): the value wasn't being quoted when okteto crafts the remote command (see #4045). The issue simply never got closed. This PR trims the key in env.Parse, which doesn't touch that code path, so it doesn't change the behavior described in #4029. If you're still hitting the whitespace problem on master, could you share a reproduction? Otherwise the issue can just be closed.

 2. We'd rather not reject empty keys. Parse backs the `--var` flag for deploy/destroy, and rejecting an empty key removes the only way to pass an empty variable, which is a use case we want to keep supporting. So `if key == "" { return ... }` would be a regression for us.

 Could you tell us more about the concrete use case that motivated the key trimming and the empty-key validation? That'll help us understand what problem you're actually hitting.

Tyagiquamar:
Thanks for the detailed context. I rechecked #4029 and #4045, and you're right — I had mapped the whitespace-in-value issue to `env.Parse`, but the actual bug was in how the remote deploy command was constructed, and #4045 already fixes that path.

I don't have a current reproduction on master, or a separate concrete case that requires trimming variable keys. The empty-key validation was also based on that same incorrect assumption, so if empty keys are intentionally supported, I agree we shouldn't change that behavior.

Given that, this PR isn't the right fix for #4029. Thanks for catching the mismatch and explaining the intended semantics.

Tyagiquamar:
Thanks for the detailed context. I rechecked #4029 and #4045, and you're right  I had mapped the whitespace-in-value issue to `env.Parse`, but the actual bug was in how the remote deploy command was constructed, and #4045 already fixes that path.

I don't have a current reproduction on master, or a separate concrete case that requires trimming variable keys. The empty-key validation was also based on that same incorrect assumption, so if empty keys are intentionally supported, I agree we shouldn't change that behavior.

Given that, this PR isn't the right fix for #4029. Thanks for catching the mismatch and explaining the intended semantics.

